Issue #11916, opened 2022-10-13
[BUG] Memcheck error in ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/22


davidwendt:
Found an out-of-bounds memory write in the `PARQUET_TEST` at `ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/22`
The error does not occur for any other sizes 1,24 -- only for 22.

Partial output from compute-sanitizer:
```
========= COMPUTE-SANITIZER
Note: Google Test filter = ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/22
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ParquetDictionaryTest/ParquetSizedTest
[ RUN      ] ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/22
========= Invalid __global__ write of size 1 bytes
=========     at 0xdfb0 in void cudf::io::parquet::gpu::gpuEncodePages<(int)128>(cudf::device_span<cudf::io::parquet::gpu::EncPage, (unsigned long)18446744073709551615>, cudf::device_span<cudf::device_span<const unsigned char, (unsigned long)18446744073709551615>, (unsigned long)18446744073709551615>, cudf::device_span<cudf::device_span<unsigned char, (unsigned long)18446744073709551615>, (unsigned long)18446744073709551615>, cudf::device_span<cudf::io::compression_result, (unsigned long)18446744073709551615>)
=========     by thread (92,0,0) in block (158,0,0)
=========     Address 0x7fd632539b50 is out of bounds
=========     and is 1 bytes after the nearest allocation at 0x7fd62c000000 of size 106142544 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x305c18]
=========                in /usr/lib/x86_64-linux-gnu/libcuda.so.1
=========     Host Frame: [0x1402c]
=========                in /conda/envs/rapids/lib/libcudart.so.11.0
=========     Host Frame:cudaLaunchKernel [0x68808]
=========                in /conda/envs/rapids/lib/libcudart.so.11.0
[...]

PR #11962, opened 2022-10-20
Fix maximum page size estimate in Parquet writer

vuule:
## Description
Closes https://github.com/rapidsai/cudf/issues/11916

cuda memcheck reports an OOB write in one of the tests. The root cause is an underallocated buffer for encoded pages.
This PR fixes the computation of the maximum size of data pages (RLE encoded) when dictionary encoding is used.
Other changes:
Refactored max RLE page size computation to avoid code repetition.
Use actual dictionary index width instead of (outdated) worst case.

## Checklist

### DIFF
--- a/cpp/src/io/parquet/page_enc.cu
+++ b/cpp/src/io/parquet/page_enc.cu
@@ -337,7 +337,7 @@ __global__ void __launch_bounds__(128)
       __syncwarp();
       uint32_t fragment_data_size =
         (ck_g.use_dictionary)
-          ? frag_g.num_leaf_values * 2  // Assume worst-case of 2-bytes per dictionary index
+          ? frag_g.num_leaf_values * util::div_rounding_up_unsafe(ck_g.dict_rle_bits, 8)
           : frag_g.fragment_data_size;
       // TODO (dm): this convoluted logic to limit page size needs refactoring
       size_t this_max_page_size = (values_in_page * 2 >= ck_g.num_values)   ? 256 * 1024
--- a/cpp/src/io/parquet/page_enc.cu
+++ b/cpp/src/io/parquet/page_enc.cu
@@ -217,6 +217,14 @@ __global__ void __launch_bounds__(128)
   if (frag_id < num_fragments_per_column and lane_id == 0) groups[column_id][frag_id] = *g;
 }
 
+constexpr uint32_t max_RLE_page_size(uint8_t value_bit_width, uint32_t num_values)
+{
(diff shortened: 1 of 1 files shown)


cpp/src/io/parquet/page_enc.cu
```diff
@@ -343,8 +351,8 @@ __global__ void __launch_bounds__(128)
           (values_in_page > 0 && (page_size + fragment_data_size > this_max_page_size)) ||
           rows_in_page >= max_page_size_rows) {
         if (ck_g.use_dictionary) {
-          page_size =
-            1 + 5 + ((values_in_page * ck_g.dict_rle_bits + 7) >> 3) + (values_in_page >> 8);
+          // Additional byte to store entry bit width
+          page_size = 1 + max_RLE_page_size(ck_g.dict_rle_bits, values_in_page);
```
vuule, line 355: Based on [Apache Parquet format docs](https://parquet.apache.org/docs/file-format/data-pages/encodings/#dictionary-encoding-plain_dictionary--2-and-rle_dictionary--8):
> Data page format: the bit width used to encode the entry ids stored as 1 byte (max bit width = 32), followed by the values encoded using RLE/Bit packed described above (with the given bit width).

cpp/src/io/parquet/page_enc.cu
```diff
@@ -217,6 +217,14 @@ __global__ void __launch_bounds__(128)
   if (frag_id < num_fragments_per_column and lane_id == 0) groups[column_id][frag_id] = *g;
 }
 
+constexpr uint32_t max_RLE_page_size(uint8_t value_bit_width, uint32_t num_values)
```
bdice, line 220: Nice refactoring here.

cpp/src/io/parquet/page_enc.cu
bdice: This is the bug fix and the rest of the diff is refactoring, right? Is a C++ test or Python comparison to another reader/writer needed?

davidwendt: Technically there is already a gtest for this. The error was an OOB write and only manifested with a memcheck without an rmm pool which is tested nightly.

davidwendt (approved):

cpp/src/io/parquet/page_enc.cu
bdice: Good call. Approving.

bdice (approved):

vuule:
@gpucibot merge

Issue #11280, opened 2022-07-15
[BUG] Decompressing an orc file with ZSTD can fail due to an inaccurate uncompress block size


jbrennan333:
Decompressing an orc file can lead to a crash in the nvcomp-2.3.2 ZSTD decompression code.
The problem is that the nvcomp ZSTD decompressor requires an accurate calculation of the total decompress size, and this heuristic used in `gpuParseCompressedStripeData` can under-estimate the size for small blocks.

```
      // TBD: For some codecs like snappy, it wouldn't be too difficult to get the actual
      // uncompressed size and avoid waste due to block size alignment For now, rely on the max
      // compression ratio to limit waste for the most extreme cases (small single-block streams)
      uncompressed_size = (is_uncompressed)                         ? block_len
                          : (block_len < (block_size >> log2maxcr)) ? block_len << log2maxcr
                                                                    : block_size;
```

In the case of zstd, log2maxcr is set to 11, which corresponds to a max compression ratio of 2048:1

This was found while testing tpc-ds queries in spark using the spark-rapids plugin.  When run with Orc/ZSTD, 11 queries were producing incorrect output (it wasn't actually crashing).   While investigating one of them, I was able simplify the query to one that caused a crash and then I dumped the orc file that was passed to cudf (attached).   Here is part of the output log from spark-shell, while running under compute-sanitizer and reading the orc file:

```
22/07/15 22:04:18 INFO MultiFileOrcPartitionReader: Writing split data for path: file:///home/jimb/spark/dump-files-shell-orc-zstd-0714/2035750964.orc, range: 0-82178, partition values: [empty row] to /home/jimb/spark/dump-files-shell-orc-zstd/566502374.orc
ZSTD: compinfo[0]: num_compressed_blocks: 1, num_uncompressed_blocks: 0 compressed_data_size: 81973 max_uncompressed_size: 262144 max_uncompressed_block_size: 262144
ZSTD: compinfo[1]: num_compressed_blocks: 1, num_uncompressed_blocks: 0 compressed_data_size: 48 max_uncompressed_size: 92160 max_uncompressed_block_size: 92160
ZSTD: compinfo[2]: num_compressed_blocks: 0, num_uncompressed_blocks: 0 compressed_data_size: 6 max_uncompressed_size: 3 max_uncompressed_block_size: 3
[...]

PR #11288, opened 2022-07-18
Workaround for nvcomp zstd overwriting blocks for orc due to underestimate of sizes

jbrennan333:
This is a possible workaround for issue #11280.  We have a goal to support NVCOMP ZSTD in 22.08, so a short-term fix is desired.

By forcing the log2 decompression ratio to 24 for ZSTD when calling `ParseCompressedStripeData`, it effectively disables the heuristic for estimating the block size when` blk_len < (block_size << log2maxcr)`.  I have verified that I can run nds2.0 power queries with this change, and they succeed with the correct output.

I'm not sure if this is the best workaround.  We could use 31 for the max instead of 24, or we could put this workaround in orc.cpp where `m_log2MaxRatio` is set to `11` for ZSTD.    For this PR, I put it in `decompress_stripe_data()` to narrow the scope of the workaround to where it is used.

I am definitely open to other suggestions.
cc: @devavret, @vuule

PR #11964, opened 2022-10-21
Remove uncompressed block size estimate from ORC reader

vuule:
## Description
closes https://github.com/rapidsai/cudf/issues/11280

ORC reader estimates the uncompressed block size when the compressed block is much smaller than the maximum uncompressed block size (64KB-256KB). This estimate is based on compression ratio and has shown to be a bit flaky (see https://github.com/rapidsai/cudf/pull/11288).

This PR removes this logic so now reader allocates max block size (typically not larger than 256KB) for each block, regardless of the compressed size. 

In theory this impacts the memory usage when file contains many small blocks. However, there is no observed increase in peak memory usage when reading files with a thousand columns that each contain ~512 bytes of data.

## Checklist


vuule:
Closing, will open a new PR to refactor/enable for ZSTD

Issue #11729, opened 2022-09-21
The segmented_sort.cu should align with sort_column.cu implementation


davidwendt:
There are some inconsistencies in the sort logic used by `sort_column.cu` and `segmented_sort.cu` both use a radix-sort optimization but apply different logic for handling nulls and fixed-width types. Overall the logic should match to provide predictable results. One such occurrence appears in #11630 where NaNs are handled incorrectly. This logic was already correct in the `sort_column.cu` implementation.
Upon fixing #11630 in #11703 we found other discrepancies like handling of fixed-point types. It also appears that chrono-types are not handled in the same was as well.
Using the `is_relationally_comparable` trait should help align these two implementations after #11703 is merged.

Additional reference: https://github.com/rapidsai/cudf/pull/11703#issuecomment-1253674563

davidwendt:
Looks like the `sort_column.cu`'s `is_radix_sort_supported()` should match the one in `segmented_sort.cu` (in PR #11703). The `sort_column.cu` logic uses `thrust::sort_by_key` to call the CUB radix-sort routines. But thrust catches the chrono types will not work and automatically funnels those types to thrust's internal merge-sort function instead.

davidwendt:
Here is how the current sorting.hpp functions and are implemented for a single column.

| API | Implementation |
| ---- | ---- |
| [cudf::sorted_order](https://docs.rapids.ai/api/libcudf/stable/group__column__sort.html#ga234cbe8106e60ecbbaf5704e1da0762c) |  copies input column, creates sequence column and calls `thrust::sort_by_key` (fast-sort) or `thrust::sort` (with comparator) |
| [cudf::stable_sorted_order](https://docs.rapids.ai/api/libcudf/stable/group__column__sort.html#ga35d35d83d9966ee78b9e1a09c130c28a) | copies input column, creates sequence column and calls `thrust::stable_sort_by_key` (fast-sort) or `thrust::stable_sort` (with comparator) |
| [cudf::sort](https://docs.rapids.ai/api/libcudf/stable/group__column__sort.html#ga63ae5bdac795a9829eda093339cf5a6d) | calls [cudf::detail::sorted_order](https://github.com/rapidsai/cudf/blob/branch-22.12/cpp/src/sort/sort_column.cu#L124) and `cudf::detail::gather` |
[...]

PR #11969, opened 2022-10-21
Accelerate libcudf segmented sort with CUB segmented sort

davidwendt:
## Description
Moves the CUB segmented sort acceleration code logic from `cudf::lists::segmented_sorted_order` to `cudf::segmented_sorted_order` so these two functions are now aligned in behavior and performance.
This change allows `cudf::lists::segmented_sorted_order` is to use the `cudf::detail::segmented_sorted_order` for all cases and simplifies the implementation there. This is also improves the performance of `cudf::segmented_sorted_order` since appropriate columns can now use CUB's optimized segmented sort.
No function has been changed and the existing tests are sufficient -- source has only been refactored.
Added a segmented-sort benchmark using nvbench as well.

Reference #11729 

## Checklist


davidwendt:
The benchmark was useful in finding where the fast CUB segmented sort was indeed faster. There were a few places where a large number of wide-ish segments would not perform so well. So the code logic checks for this and bypasses the fast sort. I verified the behavior on an RTX A6000 as well as a Quadro GV100. Here is the final result.

```
|  size  |  row    |            |            |               |         |
|  bytes |  width  |   Ref Time |   Cmp Time |          Diff |   %Diff |
|--------|---------|------------|------------|---------------|---------|
|  2^16  |    16   | 267.526 us |  49.285 us |   -218.241 us | -81.58% |
|  2^18  |    16   | 294.013 us |  50.009 us |   -244.004 us | -82.99% |
|  2^20  |    16   | 353.535 us |  64.992 us |   -288.543 us | -81.62% |
|  2^22  |    16   |   1.181 ms | 317.454 us |   -863.895 us | -73.13% |
|  2^24  |    16   |   4.008 ms |   1.119 ms |  -2888.984 us | -72.09% |
|  2^28  |    16   |  65.237 ms |  13.417 ms | -51820.400 us | -79.43% |
|  2^16  |   128   | 286.055 us |  57.291 us |   -228.764 us | -79.97% |
|  2^18  |   128   | 313.091 us | 174.742 us |   -138.350 us | -44.19% |
|  2^20  |   128   | 379.024 us | 375.419 us |     -3.605 us |  -0.95% |
|  2^22  |   128   |   1.276 ms |   1.287 ms |     11.095 us |   0.87% |
|  2^24  |   128   |   4.283 ms |   4.295 ms |     12.402 us |   0.29% |
|  2^28  |   128   |  65.661 ms |  66.135 ms |    474.111 us |   0.72% |
|  2^16  |  1024   | 339.134 us |  38.492 us |   -300.642 us | -88.65% |
|  2^18  |  1024   | 368.825 us |  38.886 us |   -329.939 us | -89.46% |
|  2^20  |  1024   | 430.233 us | 427.226 us |     -3.007 us |  -0.70% |
|  2^22  |  1024   |   1.404 ms |   1.418 ms |     13.610 us |   0.97% |
|  2^24  |  1024   |   4.601 ms |   4.607 ms |      5.853 us |   0.13% |
|  2^28  |  1024   |  71.912 ms |  72.513 ms |    600.250 us |   0.83% |
```

These are only the results for test input with no nulls though the benchmark covers both nulls and no-nulls. The fast sort is not used with nulls and so the results did not change.

cpp/src/sort/segmented_sort.cu
```diff
@@ -34,6 +39,195 @@ namespace {
...
+/**
+ * @brief Functor performs faster sort on eligible columns
+ */
+struct column_fast_sort_fn {
+  /**
+   * @brief Run-time check for faster sort on an eligible column
+   */
+  static bool is_fast_sort_supported(column_view const& col)
+  {
+    return !col.has_nulls() and cudf::is_numeric(col.type()) and
```
bdice, line 51: I'm a little unsettled that we are using `is_numeric and is_floating_point` at runtime, but `is_integral` at compile time. Can we add a trait `cudf::is_integral(data_type)` that acts like `std::is_integral`, to align with how we have `cudf::is_floating_point`?

Note that `cuda::std::is_integral` will return true for int128, but `std::is_integral` will not.

More broadly, we need to decide if we want to handle decimals using this same code path. In theory that should be fine, right? What about other types using integral representations, like timestamps or durations? Should they get the fast sort? To my knowledge there is no reason (or special representation for those types) that would prohibit the fast sort.

Because we are now able to define this in a single location, we should ensure that we cover all the appropriate types with the fastest method available.

cpp/src/sort/segmented_sort.cu
```diff
@@ -34,6 +39,195 @@ namespace {
...
+                 bool ascending,
+                 rmm::cuda_stream_view stream)
+  {
+    // CUB's segmented sort functions cannot accept iterators.
+    // We create a temporary column here for it to use.
+    auto temp_col =
+      cudf::detail::allocate_like(input, input.size(), mask_allocation_policy::NEVER, stream);
+    mutable_column_view output_view = temp_col->mutable_view();
+
+    if (ascending) {
```
bdice, line 157: We should reduce a lot of this code by using a lambda with perfect forwarding to select the CUB function. All the input parameters and temp storage appear to be the same. Example: https://godbolt.org/z/3sY39EMYz

With that, you can eliminate `fast_sort_ascending/fast_sort_descending` and keep the implementation unified in `fast_sort`.

bdice (changes requested):
Awesome work so far, and the benchmarks look great. I have a few suggestions.

cpp/src/sort/segmented_sort.cu
davidwendt: Unfortunately, CUB `DeviceSegmentedSort` does not seem to like `timestamp` and `duration` types. 
Also it does not seem to like `int128` either so this means the current implementation will fail for DECIMAL128 columns.

bdice: Is it well-defined to sort the `timestamp` or `duration` types "as if" they're integral, e.g. with a type cast?

Totally fine for this skip `int128` -- but should still be viable for 32 and 64 bit decimals, I'd think.

cpp/src/sort/segmented_sort.cu
davidwendt: I find this a bit frustrating. There is technically not a lot of code here. It only appears that way because of the clang-formatting. Your suggestion actually only reduces the number of lines of code from 12 to 10. The `std::forward` helps to reduce the source text lines count (dramatically) because the parameter list does not need to be expanded by clang-format. So I feel this is more of a solution to get around code formatting than to make the code more clear.

bdice: Maybe I'm missing something, but I think my proposal would reduce the common logic by a substantial amount, irrespective of clang-formatting? I can push a commit to my fork if you want to see exactly what I'm thinking (but probably tomorrow).

davidwendt: I pushed a change. Let me know what you think. Maybe I misunderstood.

cpp/include/cudf/utilities/traits.hpp
```diff
@@ -261,6 +261,30 @@ constexpr inline bool is_signed_iterator()
   return std::is_signed_v<typename std::iterator_traits<Iterator>::value_type>;
 }
 
+/**
+ * @brief Indicates whether the type `T` is an integral type.
```
bdice, line 265: Should we note that `int128` is _not_ considered integral by this definition? (Either way is fine but that subtlety between `std::` and `cuda::std::` is not always obvious.)

cpp/src/sort/segmented_sort.cu
davidwendt: > Is it well-defined to sort the `timestamp` or `duration` types "as if" they're integral, e.g. with a type cast? (I've asked this before and can't quite recall the answer.)

I don't that is allowed. Maybe I don't understand what you are asking. There is no equivalent dispatch-storage-type for the chrono types. The chrono types have their own comparators and perhaps that is why `DeviceSegmentedSort` cannot handle them.

cpp/src/sort/segmented_sort.cu
bdice: This change is great! It reduced the number of times that arguments were specified to 1 time instead of 4 times (for each ascending/descending, for both calls to the function). Thank you.

bdice (approved):

cpp/src/sort/segmented_sort.cu
bdice: Okay, thanks. I always forget that they have their own comparators.

cpp/include/cudf/utilities/traits.hpp
davidwendt: I think this one should be `cuda::std::is_integral<T>`. I just need some time for the world to recompile to test if this is ok. If not, I'll add the comment.

cpp/src/sort/segmented_sort.cu
davidwendt: I'm getting this error in CI but not locally
```
21:14:43 $SRC_DIR/cpp/src/sort/segmented_sort.cu(85): error #177-D: parameter "args" was declared but never referenced
```
I could use some help figuring out why this is happening and how to fix it.
https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/cudf/job/prb/job/cudf-cpu-cuda-build/CUDA=11.5/13331/console

davidwendt: Adding `[[maybe_unused]]` to the declaration seems to have fixed it.

cpp/src/sort/segmented_sort.cu
```diff
@@ -89,8 +218,42 @@ std::unique_ptr<column> segmented_sorted_order_common(
...
+  constexpr cudf::size_type MAX_AVG_LIST_SIZE_FOR_FAST_SORT{100};
+  // the maximum row count for which to prefer fast sort
+  constexpr cudf::size_type MAX_LIST_SIZE_FOR_FAST_SORT{1 << 18};
+
+  // fast-path for single column sort:
+  // - single-column table
+  // - not stable-sort
+  // - no nulls and allowable fixed-width type
+  // - size and width are limited -- based on benchmark results
+  if (keys.num_columns() == 0 and sorting == sort_method::UNSTABLE and
```
karthikeyann, line 248: `keys.num_columns() == 0`
equals zero?
How are unit tests and benchmarks not catching this?

davidwendt: Good catch. This is a mistake and was meant only to capture the baseline benchmarks..

cpp/src/sort/segmented_sort.cu
```diff
@@ -25,6 +28,8 @@
 #include <thrust/binary_search.h>
 #include <thrust/iterator/counting_iterator.h>
 
+#include <cub/cub.cuh>
```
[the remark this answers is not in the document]
davidwendt: This include is what the documentation says to use for these APIs:
https://nvlabs.github.io/cub/struct_device_segmented_sort.html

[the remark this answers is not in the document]
karthikeyann: In the same doc page, Snippet has
```
#include <cub/cub.cuh>
// or equivalently <cub/device/device_segmented_sort.cuh>
```
nit: Hence I recommended it to reduce the includes. (IWYU - Include What You Use).

davidwendt:
@gpucibot merge

PR #11971, opened 2022-10-21
Purge nonempty nulls from byte_cast list outputs.

### LINKED ISSUE #11754: [BUG] cudf::byte_cast can return non-empty null values.
**Describe the bug**
For short, int, and long values that are nulls the results will not point to an empty list.  Technically this does not violate the Arrow spec, but with changes going into CUDF for sorting of nested types, this can potentially cause incorrect results.

Not sure how critical this is, just because the list sorting is not done, and it may only really be a problem for lists of stricts, but I just don't know for sure.

Call cudf::byte_cast for an int column with `[null, 5]` in it. The result will look correct, but the first row will point to 4 bytes of data.

**Expected behavior**
The first column would point to the same offset for both the start and the end. (in this case 0).

bdice:
## Description
Resolves #11754. The `byte_cast` function is creating unsanitized lists from null inputs, which is a bug. [This logic](https://github.com/rapidsai/cudf/blob/9c06330363db4da99803a3728b8bf44f9829f0b9/cpp/src/reshape/byte_cast.cu#L66-L81) copies nonzero bytes even if the input element is null. The input's null mask is copied onto the output parent list column, but the null children are nonempty. This PR fixes the bug by calling `cudf::purge_nonempty_nulls` on the result before returning, if there are any nulls to be purged.

Depends on:
 * https://github.com/rapidsai/cudf/pull/13099

## Checklist


ttnghia:
This is fixing for lists. What about [strings](https://github.com/rapidsai/cudf/blob/9c06330363db4da99803a3728b8bf44f9829f0b9/cpp/src/reshape/byte_cast.cu#L94)?

cpp/src/reshape/byte_cast.cu
```diff
@@ -80,13 +81,20 @@ struct byte_list_conversion {
...
-                             std::move(null_mask),
-                             stream,
-                             mr);
+    auto result = make_lists_column(input_column.size(),
+                                    std::move(offsets_column),
+                                    std::move(byte_column),
+                                    input_column.null_count(),
+                                    std::move(null_mask),
+                                    stream,
+                                    mr);
```
karthikeyann, line 90: I had this doubt for a while. Must all temporaries be created with default memory resource only or is it okay to create temporaries using `mr` too sometimes?

hyperbolic2346:
Would it be better to not copy the null data to begin with instead of making a new pass to remove the data that was copied? Is there a reason that wouldn't work?

cpp/src/reshape/byte_cast.cu
bdice: This isn't a temporary. We need to use `mr` if the result can be returned to the user.

The `result` column _may_ be returned as-is, if no nulls need to be purged. If there are nulls to purge, then the purged column uses `mr` and is returned (leaving `result` to be deallocated when it goes out of scope).

bdice:
> This is fixing for lists. What about [strings](https://github.com/rapidsai/cudf/blob/9c06330363db4da99803a3728b8bf44f9829f0b9/cpp/src/reshape/byte_cast.cu#L94)?

Very good catch @ttnghia. I'll address this soon.

ttnghia:
> Would it be better to not copy the null data to begin with instead of making a new pass to remove the data that was copied? Is there a reason that wouldn't work?

We need to know the exact target location to copy to. If we want to generate empty lists directly,we have to recompute offsets before copying. I believe that it is not cheaper while more complicated.

hyperbolic2346 (approved):

cpp/src/reshape/byte_cast.cu
```diff
@@ -80,13 +81,20 @@ struct byte_list_conversion {
...
     rmm::device_buffer null_mask = detail::copy_bitmask(input_column, stream, mr);
 
-    return make_lists_column(input_column.size(),
-                             std::move(offsets_column),
-                             std::move(byte_column),
-                             input_column.null_count(),
-                             std::move(null_mask),
-                             stream,
-                             mr);
+    auto result = make_lists_column(input_column.size(),
```
ttnghia, line 84: FYI: I did a fix for `make_lists_column` in https://github.com/rapidsai/cudf/pull/12370 thus I think this PR is no longer needed.

ttnghia: Sorry for another opposite idea: From discussions, we may want to change the strategy to deal with list sanitization. By the new approach, I'll reverse https://github.com/rapidsai/cudf/pull/12370 and call `purge_nonempty_nulls` in the APIs generating output lists column. As such, this PR is still relevant.

ttnghia: From our offline discussion, the course of approaches was reversed again: keep calling `purge_nonempty_nulls` in `make_lists_column` (but that factory func will add a boolean flag allowing to bypass calling `purge_nonempty_nulls`).

So please pause this, waiting for more clarification/implementation.

ttnghia: FYI: From our final discussion, we will continue to address non-empty nulls in each individual APIs. Please resume this :fire:

cpp/src/reshape/byte_cast.cu
```diff
@@ -80,13 +81,20 @@ struct byte_list_conversion {
...
+                                    input_column.null_count(),
+                                    std::move(null_mask),
+                                    stream,
+                                    mr);
+
+    // If any nulls are present, the corresponding lists must be purged so that
+    // the result is sanitized.
+    if (result->null_count() > 0) {
+      return cudf::detail::purge_nonempty_nulls(lists_column_view{*result}, stream, mr);
+    }
```
ttnghia, line 96: From my previous issue with performance regression due to `purge_nonempty_nulls`, please use this:
```suggestion
    if (auto const result_cv = result->view(); cudf::detail::has_nonempty_nulls(result_cv)) {
      return cudf::detail::purge_nonempty_nulls(result_cv, stream, mr);
    }
```

cpp/src/reshape/byte_cast.cu
```diff
@@ -31,36 +32,53 @@
...
+};
+
+template <typename T>
+static constexpr bool is_supported()
+{
+  return cudf::is_numeric<T>() || std::is_same_v<T, cudf::string_view>;
+}
+
+template <typename T>
+struct byte_list_conversion_fn<
```
hyperbolic2346, line 68: Should this one be up with the other and not straddling the dispatcher?

cpp/src/reshape/byte_cast.cu
```diff
@@ -31,36 +32,53 @@
...
+  T,
+  std::enable_if_t<is_supported<T>() && !std::is_same_v<T, cudf::string_view>>> {
+  static std::unique_ptr<column> invoke(column_view const& input,
+                                        flip_endianness configuration,
+                                        rmm::cuda_stream_view stream,
+                                        rmm::mr::device_memory_resource* mr)
+  {
+    size_type num_bytes = input.size() * sizeof(T);
```
hyperbolic2346, line 76:
```suggestion
    size_type const num_bytes = input.size() * sizeof(T);
```

cpp/src/reshape/byte_cast.cu
```diff
@@ -74,80 +92,95 @@ struct byte_list_conversion {
...
-                             mr);
+template <typename T>
+struct byte_list_conversion_fn<T, std::enable_if_t<std::is_same_v<T, cudf::string_view>>> {
+  static std::unique_ptr<column> invoke(column_view const& input,
+                                        flip_endianness,
+                                        rmm::cuda_stream_view stream,
+                                        rmm::mr::device_memory_resource* mr)
+  {
+    strings_column_view input_strings(input);
+    auto strings_count = input_strings.size();
```
hyperbolic2346, line 126:
```suggestion
    auto const strings_count = input_strings.size();
```

hyperbolic2346 (changes requested):
In general this is looking good to me. Just a couple of questions and suggestions.

cpp/src/reshape/byte_cast.cu
ttnghia: Sorry we still have to dispatch, since there are many types in `cudf::is_numeric`.

cpp/tests/reshape/byte_cast_tests.cpp
```diff
@@ -360,3 +328,34 @@ TEST_F(ByteCastTest, StringValues)
...
+    auto output =
+      cudf::test::strings_column_wrapper(
+        {"", "The quick", " brown fox...", "!\"#$%&\'()*+,-./", "0123456789:;<=>?@", "[\\]^_`{|}~"})
+        .release();
+
+    // Set nulls by `set_null_mask` so the output column will have non-empty nulls.
+    // This is intentional.
+    auto const null_iter = cudf::test::iterators::nulls_at({2, 4});
+    output->set_null_mask(
+      cudf::test::detail::make_null_mask(null_iter, null_iter + output->size()));
```
davidwendt, line 344:
```suggestion
      cudf::test::detail::make_null_mask(null_iter, null_iter + output->size()), 2);
```
Set the null count too instead of relying on the default.

davidwendt (changes requested):
Looks good. 
Can there be tests for an empty input column too?
And perhaps tests for any `throw` conditions as well? Or maybe I missed those.

ttnghia:
> Can there be tests for an empty input column too?

Thanks. Having the suggested tests, I discovered bug(s) when the input is empty/all nulls. Now they're fixed.

ttnghia:
Since this also uses `lists::make_empty_lists_column`, it depends on:
 * https://github.com/rapidsai/cudf/pull/13099

davidwendt (approved):
LGTM

cpp/src/reshape/byte_cast.cu
```diff
@@ -31,123 +33,160 @@
...
-   */
-  template <typename T>
-  std::enable_if_t<!std::is_integral_v<T> and !is_floating_point<T>(), std::unique_ptr<column>>
-  operator()(column_view const&,
-             flip_endianness,
-             rmm::cuda_stream_view,
-             rmm::mr::device_memory_resource*) const
+
+// Data type of the output data column after conversion.
+data_type constexpr output_type{data_type{type_id::UINT8}};
```
vyasr, line 43: We don't really use "east constexpr" anywhere in our codebase. Many of the reasons that east const makes sense don't really make sense for constexpr, particularly when it comes to the confusion around pointers. For instance `data_type constexpr* output_type` does not in fact mean "pointer to constant data_type", it means "constant pointer to data_type". For instance, check out this example: https://godbolt.org/z/o4YaEhrsE.

cpp/src/reshape/byte_cast.cu
```diff
@@ -31,123 +33,160 @@
...
+    auto const d_out = byte_column->mutable_view().data<char>();
 
     if (configuration == flip_endianness::YES) {
       thrust::for_each(rmm::exec_policy(stream),
                        thrust::make_counting_iterator(0),
                        thrust::make_counting_iterator(num_bytes),
-                       [d_chars, d_data, mask] __device__(auto index) {
-                         d_chars[index] = d_data[index + mask - ((index & mask) << 1)];
+                       [d_inp, d_out] __device__(auto index) {
+                         auto constexpr mask = static_cast<size_type>(sizeof(T) - 1);
```
vyasr, line 92: Same note about constexpr.

cpp/src/reshape/byte_cast.cu
```diff
@@ -31,123 +33,160 @@
...
+    if (input.size() == input.null_count()) {
+      return cudf::lists::detail::make_all_nulls_lists_column(
+        input.size(), output_type, stream, mr);
+    }
+
+    auto col_content     = std::make_unique<column>(input, stream, mr)->release();
+    auto chars_contents  = col_content.children[strings_column_view::chars_column_index]->release();
+    auto const num_chars = chars_contents.data->size();
+    auto uint8_col       = std::make_unique<column>(
+      output_type, num_chars, std::move(*(chars_contents.data)), rmm::device_buffer{}, 0);
```
vyasr, line 141: Nice this section is cleaner now without the old null mask clutter.

vyasr (approved):
I love that we had tests explicitly noting that we had unsanitized data and working around it.

One non-blocking note about east vs west constexpr, otherwise good to ship from me.

hyperbolic2346 (approved):

cpp/src/reshape/byte_cast.cu
```diff
@@ -80,13 +81,20 @@ struct byte_list_conversion {
...
+                                    std::move(null_mask),
+                                    stream,
+                                    mr);
+
+    // If any nulls are present, the corresponding lists must be purged so that
+    // the result is sanitized.
+    if (result->null_count() > 0) {
+      return cudf::detail::purge_nonempty_nulls(lists_column_view{*result}, stream, mr);
+    }
+    return result;
```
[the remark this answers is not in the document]
karthikeyann: You may have to use `std::move(result)`.

PR #11986, opened 2022-10-25
Move protobuf compilation to CMake

vyasr:
## Description

We currently compile a proto file into a Python file inside setup.py by overriding a certain setuptool (scikit-build) stage (`build_ext`). However, depending on the exact means by which we are building the package (specifically, in the case of building wheels) we may occasionally bypass that stage. Putting this logic into the CMake guarantees that it is always run.
 
## Checklist

### DIFF
--- a/python/cudf/cmake/Modules/ProtobufHelpers.cmake
+++ b/python/cudf/cmake/Modules/ProtobufHelpers.cmake
@@ -25,6 +25,8 @@ function(codegen_protoc)
   foreach(_proto_path IN LISTS ARGV)
     string(REPLACE "\.proto" "_pb2\.py" pb2_py_path "${_proto_path}")
     set(pb2_py_path "${CMAKE_CURRENT_SOURCE_DIR}/${pb2_py_path}")
+    # Note: If we ever need to process larger numbers of protobuf files we should consider switching
+    # to protobuf_generate_python from the FindProtobuf module.
     execute_process(
       COMMAND ${protoc_COMMAND} --python_out=. "${_proto_path}"
       WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} COMMAND_ERROR_IS_FATAL ANY


python/cudf/setup.py
```diff
@@ -84,48 +80,7 @@ def get_cuda_version_from_header(cuda_include_dir, delimeter=""):
...
-                        + os.linesep
-                    )
-                    src.seek(0)
-                    src.write(new_src_content)
-
-        # Run original Cython build_ext command
-        super().run()
-
-
 cmdclass = versioneer.get_cmdclass()
```
bdice, line 83: Let's move this down. Set `cmdclass=versioneer.get_cmdclass()` in the kwargs of `setup(...)`.

bdice (approved):
LGTM, with one suggestion.

python/cudf/cmake/Modules/ProtobufHelpers.cmake
```diff
@@ -0,0 +1,49 @@
...
+    )
+    file(APPEND "${pb2_py_path}" "${pb2_py}")
+    file(
+      APPEND "${pb2_py_path}"
+      [=[
+# fmt: on
+]=]
+    )
+  endforeach()
+endfunction()
```
wence-, line 49: Minor (more for understanding): does this do partial rebuild tracking appropriately? That is, if I only change a single `.proto` file does only that one get rebuilt?

wence- (approved):
Modulo @bdice's minor quibble

python/cudf/cmake/Modules/ProtobufHelpers.cmake
bdice: Currently it doesn't matter, because there is only one `.proto` file.
https://github.com/rapidsai/cudf/blob/branch-22.12/python/cudf/cudf/utils/metadata/orc_column_statistics.proto

I'm not sure what would happen if multiple files were present, but it may not be worth worrying about.

vyasr: Yeah this will just rebuild everything naively. We can rethink the strategy if we ever need to scale up, but right now the usage is super limited to a single ORC statistics file so it's not worth overengineering yet.

vyasr:
@gpucibot merge

PR #11980, opened 2022-10-24
Replace most of preprocessor usage in nvcomp adapter with `constexpr`

vuule:
## Description
C++17's "constexpr if" provides the same functionality as `#if` directive, as used in the nvcomp adapter.
This PR replaces macros with `constexpr` variables and uses them as conditions in "constexpr if" statements.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #11999, opened 2022-10-25
Revert "Replace most of preprocessor usage in nvcomp adapter with `constexpr`"

vuule:
Reverts rapidsai/cudf#11980

The PR was made under the assumption that `if constexpr` branches can contain invalid code, if the branch is not taken. However, this only holds for templates.

### DIFF
--- a/cpp/src/io/comp/nvcomp_adapter.cpp
+++ b/cpp/src/io/comp/nvcomp_adapter.cpp
@@ -31,30 +31,46 @@
 #include NVCOMP_ZSTD_HEADER
 #endif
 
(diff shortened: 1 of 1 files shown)


ttnghia (approved):
:cold_face:

bdice (approved):

vuule:
@gpucibot merge

bdice:
For posterity:

The symbols in the `if constexpr` body may have to exist, even if they're supposed to be compiled out. https://godbolt.org/z/h3a7vazzc

Older nvcomp versions do not have the symbol `nvcompBatchedZstdCompressGetTempSize` defined, which led to an error for some users.

PR #12000, opened 2022-10-26
[REVIEW] Ignore python docs build artifacts

galipremsagar:
## Description
This PR gitignores some of the python docs build artifcats that keep showing up in `git status`

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -165,3 +165,6 @@ dask-worker-space/
 # Sphinx docs & build artifacts
 docs/cudf/source/api_docs/generated/*
 docs/cudf/source/api_docs/api/*
+docs/cudf/source/user_guide/example_files/*
+docs/cudf/source/user_guide/example_output/*
+docs/cudf/source/user_guide/cudf.*Dtype.*.rst


.gitignore
```diff
@@ -165,3 +165,6 @@ dask-worker-space/
 # Sphinx docs & build artifacts
 docs/cudf/source/api_docs/generated/*
 docs/cudf/source/api_docs/api/*
+docs/cudf/source/user_guide/example_files/*
```
bdice, line 168: I'd like to modify the `10min.ipynb` to put this in `example_outputs` so that all files are in the same output directory.

.gitignore
```diff
@@ -165,3 +165,6 @@ dask-worker-space/
...
+docs/cudf/source/user_guide/example_output/*
+docs/cudf/source/user_guide/cudf.*Dtype.*.rst
```
bdice, line 170: What does this rule ignore? Are there generated files / where do they come from?

.gitignore
galipremsagar: Updated.

.gitignore
galipremsagar: There are files generated after build for every attribute & method:
Grayed out ones are the build artifacts
Screen Shot 2022-10-26 at 3 59 09 PM

bdice: @galipremsagar Do you know what script generates those files or how that is defined? This doesn't seem like a Sphinx feature that I'm familiar with.

bdice (approved):
LGTM, see question in thread.

.gitignore
galipremsagar: That is the expected behavior and happens in all of our sphinx-build. Those files are generated in `docs/cudf/source/api_docs/api/` folder because we specify `toctree: api/`.

The part of code that is responsible for this is everywhere in our code where we do `HACK --...` in comments for example:
```
..
   HACK -- the point here is that we don't want this to appear in the output, but the autosummary should still generate the pages.
   .. autosummary::
      :toctree:
      {% for item in all_methods %}
      {%- if not item.startswith('_') or item in ['__call__'] %}
      {{ name }}.{{ item }}
      {%- endif -%}
      {%- endfor %}
```

This is basically being done to mimic panadas & generate dedicated html pages for each property and function.

bdice: Thanks for the explanation!

galipremsagar:
@gpucibot merge

PR #12003, opened 2022-10-26
Remove default parameters for cudf::strings::detail functions

davidwendt:
## Description
Removes default parameters from the `cudf::strings::detail` functions. Most of these were unintentional the rest were for allowing for the default memory-resource which was easily fixed. Most of the detail functions are not used outside of strings and the default parameters were not actually necessary there.

Hopefully this will help with #11967 

## Checklist

### DIFF
--- a/cpp/src/strings/filling/fill.cu
+++ b/cpp/src/strings/filling/fill.cu
@@ -19,7 +19,8 @@
 #include <cudf/detail/valid_if.cuh>
 #include <cudf/null_mask.hpp>
 #include <cudf/scalar/scalar_device_view.cuh>
-#include <cudf/strings/combine.hpp>
+//#include <cudf/strings/combine.hpp>
+#include <cudf/strings/detail/fill.hpp>
 #include <cudf/strings/detail/utilities.cuh>
 #include <cudf/strings/detail/utilities.hpp>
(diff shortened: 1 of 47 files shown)


bdice (approved):

cpp/src/strings/filling/fill.cu
```diff
@@ -19,7 +19,8 @@
 #include <cudf/detail/valid_if.cuh>
 #include <cudf/null_mask.hpp>
 #include <cudf/scalar/scalar_device_view.cuh>
-#include <cudf/strings/combine.hpp>
+//#include <cudf/strings/combine.hpp>
```
PointKernel, line 22:
```suggestion
```
Can this be removed?

PointKernel (approved):
One minor question about the comment-out header inclusion.

LGTM. Thanks!

cpp/src/strings/filling/fill.cu
bdice: Agreed. I'll commit this so that CI can re-run, since we have two approvals.

davidwendt: Oops. Yes. Thanks.

davidwendt:
rerun tests

davidwendt:
@gpucibot merge

PR #11447, opened 2022-08-03
[REVIEW] Refactor the `Buffer` class

madsbk:
## Description
This PR introduces factory functions to create `Buffer` instances, which makes it possible to change the returned buffer type based on a configuration option in a follow-up PR.

Beside simplifying the code base a bit, this is motivated by the spilling work in https://github.com/rapidsai/cudf/pull/10746. We would like to introduce a new spillable Buffer class that requires minimal changes to the existing code and is only used when enabled explicitly. This way, we can introduce spilling in cuDF as an experimental feature with minimal risk to the existing code.

@shwina and I discussed the possibility to let `Buffer.__new__` return different class type instances instead of using factory functions but we concluded that having `Buffer()` return anything other than an instance of `Buffer` is simply too surprising :)

**Notice**, this is breaking because it removes unused methods such as `Buffer.copy()` and `Buffer.nbytes`. 
~~However, we still support creating a buffer directly by calling `Buffer(obj)`. AFAIK, this is the only way `Buffer` is created outside of cuDF, which [a github search seems to confirm](https://github.com/search?l=&q=cudf.core.buffer+-repo%3Arapidsai%2Fcudf&type=code).~~
This PR doesn't change the signature of `Buffer.__init__()` anymore.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #4164, opened 2020-02-15
Add Buffer "constructor-kwargs" header

jakirkham:
Part of `Buffer`'s serializers' work includes handling subclasses (like cuML's `Array`). However it doesn't handle any keyword arguments those subclass constructors may need. This changes that by adding an optional (by default trivial) "constructor-kwargs" header. Subclasses can then call their parents' serializer and pack this header with relevant content afterwards. This way during deserialization this content can be passed into the subclass' constructor.

cc @dantegd

PR #12009, opened 2022-10-27
Rollback of `DeviceBufferLike`

madsbk:
This PR replaces `DeviceBufferLike` with `Buffer` and clear the way for a spillable sub-class of `Buffer`.

#### Context 
The introduction of the [`DeviceBufferLike`](https://github.com/rapidsai/cudf/pull/11447) protocol was motivated by [the spilling work](https://github.com/rapidsai/cudf/pull/11553), which we initially thought would have to be implemented in Cython. However, it can be done in pure Python, which makes `DeviceBufferLike` an unneeded complexity.  

#### Review notes 

- In order to introduce a spillable-buffer in the future, we still use a factory function, `as_buffer()`, to create Buffers. 
- `buffer.py` is moved into the submodule `core.buffer` to ease organization when adding the spillable-buffer and spilling manager. 

#### Breaking
This PR breaks external use of `Buffer` e.g. `Buffer.__init__` raise an exception now and the `"constructor-kwargs"` header from #4164 has been removed. 
Submitted a PR to fix this in cuml: https://github.com/rapidsai/cuml/pull/4965

#### Checklist

cc. @shwina @vyasr


madsbk:
rerun tests

python/cudf/cudf/core/abc.py
```diff
@@ -132,18 +122,10 @@ def device_deserialize(cls, header, frames):
...
-        assert all(
-            (type(f._owner) is rmm.DeviceBuffer)
-            if c
-            else (type(f) is memoryview)
-            for c, f in zip(header["is-cuda"], frames)
-        )
-        obj = typ.deserialize(header, frames)
-
-        return obj
+        return typ.deserialize(header, frames)
```
vyasr, line 128: Is removing the assertion just cleanup?

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -0,0 +1,231 @@
+# Copyright (c) 2020-2022, NVIDIA CORPORATION.
```
vyasr, line 1: From a quick scan it looks like the only changes in buffer.py (aside from the move) are the removal of the `DeviceBufferLike` and `as_device_buffer_like` (which is now just `as_buffer` in utils.py) right? Just want to make sure I'm not missing anything.

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        buffer protocol. If `obj` represents host memory, data will
+        be copied.
+
+    Return
+    ------
+    Buffer
+        A buffer instance that represents the device memory of `obj`.
+    """
+
+    if isinstance(obj, Buffer):
```
vyasr, line 36: This function is pretty bare right now, but I assume that we expect more logic here when we introduce `SpillableBuffer` so it's valuable to have this, correct? Otherwise we could just stick this logic in the `Buffer` constructor and get rid of this entirely.

python/cudf/cudf/core/column/column.py
```diff
@@ -1281,7 +1281,7 @@ def column_empty(
         data = None
         children = (
             build_column(
-                data=as_device_buffer_like(
+                data=as_buffer(
```
vyasr, line 1284: General question: normally, in cases where we _know_ that we're passing an argument to `as_buffer` that is not a `Buffer` (aka where we are explicitly constructing an `rmm.DeviceBuffer`) I would recommend that we should use that knowledge and just call the `Buffer` constructor directly. However, I assume that in the next step we're going to use this function to also dispatch to `SpillableBuffer` vs `Buffer` right? If so, I wonder if we should protect ourselves further against instantiating the wrong object by maybe disabling the constructor and instead implementing a factory. What do you think?

vyasr: CC @shwina 
The case that I'm concerned about is someone manually doing `Buffer(...)` somewhere that should use `as_buffer` in a `SpillableBuffer`-enabled world.

vyasr (changes requested):
Some small questions to address before we finalize this.

python/cudf/cudf/core/abc.py
madsbk: Yes, and we do not enforce `type(f._owner) is rmm.DeviceBuffer` anywhere.

python/cudf/cudf/core/column/column.py
shwina: >  disabling the constructor and instead implementing a factory.

Agree

python/cudf/cudf/core/buffer/utils.py
madsbk: Exactly, this is for the coming `SpillableBuffer`

python/cudf/cudf/core/buffer/buffer.py
madsbk: Also introduction of `_init_from_host_memory()` and clean up of `deserialize()`

python/cudf/cudf/core/column/column.py
madsbk: > If so, I wonder if we should protect ourselves further against instantiating the wrong object by maybe disabling the constructor and instead implementing a factory. What do you think?

I think this would be great, cuML use to call  `Buffer(...)` directly but not anymore AFAICS.

madsbk: Do we need deprecation warning?

shwina:
rerun tests

madsbk:
Thanks for the review @vyasr. I think it is a good idea to use the `as_buffer` factory function exclusively.  
Now, `as_buffer` creates a Buffer instance through `Buffer.from_device_memory()` and `Buffer.from_host_memory()`.  

This also enables `SpillableBuffer` to overwrite specific parts of the buffer constructor.

python/cudf/cudf/core/column/column.py
madsbk: This is done now, which also makes this PR `breaking`.

python/cudf/cudf/core/buffer/buffer.py
madsbk: Instead of `_init_from_host_memory()`, we now have `from_device_memory()` and `from_host_memory()`

madsbk:
rerun tests

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -0,0 +1,270 @@
...
+    owner : object
+        Python object to which the lifetime of the memory allocation is tied.
+    """
+
+    _ptr: int
+    _size: int
+    _owner: object
+
+    def __init__(self, ptr: int, size: int, owner: object):
+        if size < 0:
```
vyasr, line 41: Actually I'm proposing something even more radical: change `Buffer`'s constructor to raise an Exception. The factories can do `cls.__new__(cls)` and then initialize members (looks like only really necessary in `from_device_memory`). I think both `from_device_memory` and `from_host_memory` should be internal (prefixed with an underscore). The _only_ publicly available method for creating `Buffer`s should be `as_buffer` (the factories are available for internal use, aka inside `as_buffer`).

python/cudf/cudf/core/column/column.py
vyasr: https://github.com/rapidsai/cudf/pull/12009#discussion_r1009694262

vyasr: This is the main outstanding item left for me, then I think this PR will get a :heavy_check_mark: from me.

python/cudf/cudf/core/buffer/buffer.py
madsbk: Done

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -0,0 +1,298 @@
...
+import rmm
+
+import cudf
+from cudf.core.abc import Serializable
+from cudf.utils.string import format_bytes
+
+T = TypeVar("T", bound="Buffer")
+
+
+def cuda_array_interface_wrapper(ptr: int, size: int, owner: object = None):
```
vyasr, line 21: Maybe out of scope for this PR, but we use the `SimpleNamespace.__cuda_array_interface__` trick in multiple other places as well (mostly in `column` code IIRC) and I wonder if we could use the same helper function in multiple places. It's not immediately obvious why one would create this object so having all the places that do so call to a single, well-documented helper function could be nice. Happy to push that to a downstream refactoring PR though.

vyasr (approved):
One small suggestion, otherwise LGTM!

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -0,0 +1,298 @@
...
+            An object that represens host memory.
+
+        Returns
+        -------
+        Buffer
+            Buffer representing a copy of `data`.
+        """
+
+        # Extract pointer and size
+        ptr, size = get_ptr_and_size(numpy.asarray(data).__array_interface__)
```
shwina, line 124: Just to be sure: are there any situations in which `numpy.asarray()` creates an unnecessary copy of `data`? Also, is it useful to use `ascontiguousarray()` here and not raise in `get_ptr_and_size`?

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -0,0 +1,298 @@
...
+
+        Returns
+        -------
+        Buffer
+            Buffer representing the same device memory as `data`
+        """
+
+        # Bypass `__init__` and initialize attributes manually
+        ret = cls.__new__(cls)
+        ret._owner = data
```
shwina, line 91: If `data` is a `Buffer`, do we just want to grab its `_owner` and keep a reference to that? Or does doing so break our spilling logic somehow?

```suggestion
        if isinstance(data, Buffer):
            ret._owner = data._owner
        else:
            ret._owner = data
```

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    If `data` isn't a buffer already, a new buffer that points to the memory of
+    `data` is created. If `data` represents host memory, it is copied to a new
+    `rmm.DeviceBuffer` device allocation. Otherwise, the memory of `data` is
+    **not** copied, instead the new buffer keeps a reference to `data` in order
+    to retain its lifetime.
+
+    Parameters
+    ----------
+    data : int or buffer-like or array-like
+        An integer representing a pointer to device memory or a buffer-like
```
shwina, line 27: > pointer to device memory

Can it also be a pointer to host memory? If so, we should clarify. If not, the docs are a bit confusing, because previously we say:

>  If `data` represents host memory, it is copied to a new `rmm.DeviceBuffer` device allocation.

python/cudf/cudf/core/buffer/buffer.py
madsbk: Done

python/cudf/cudf/core/buffer/buffer.py
madsbk: > Just to be sure: are there any situations in which numpy.asarray() creates an unnecessary copy of data?

No unnecessary, but an object itself might make a copy when exposing its array or buffer interface. Also, objects like `list` will be copied. Anyway, this is a good point since we need to keep the converted numpy array alive until _after_ the copy to device memory. Fixed.

> Also, is it useful to use `ascontiguousarray()` here and not raise in `get_ptr_and_size`?

I am not sure, if the buffer isn't contiguous, it is probably a bug? Might be best to force the user to convert the input manually before calling `as_buffer()`?

python/cudf/cudf/core/buffer/utils.py
madsbk: Added: 
```
If `data` is an integer, it is assumed to point to device memory.
```

python/cudf/cudf/core/buffer/buffer.py
madsbk: It will not break the spilling logic but `data` should never be a `Buffer` instance. `as_buffer()` return as-is if the inpurt is a `Buffer`.

madsbk:
rerun tests

shwina (approved):

shwina:
@gpucibot merge

PR #11314, opened 2022-07-20
Added 'crosstab' and 'pivot_table' features

shaswat-indian:
Resolves https://github.com/rapidsai/cudf/issues/5944, resolves https://github.com/rapidsai/cudf/issues/1214

This PR adds support for basic use cases of  `crosstab` and `pivot_table` functions.

PR #12015, opened 2022-10-27
Add DataFrame.pivot_table.

bdice:
## Description
This PR adds the method `DataFrame.pivot_table` to enhance pandas API compatibility. It uses the exact same arguments as `cudf.pivot_table` but automatically supplies the first argument (a DataFrame).

Related: #11314

## Checklist

### DIFF
--- a/python/cudf/cudf/tests/test_reshape.py
+++ b/python/cudf/cudf/tests/test_reshape.py
@@ -596,6 +596,42 @@ def test_pivot_table_simple(data, aggfunc, fill_value):
     assert_eq(expected, actual, check_dtype=False)
 
 
+@pytest.mark.parametrize(
+    "data",
+    [
+        {
+            "A": ["one", "one", "two", "three"] * 6,
+            "B": ["A", "B", "C"] * 8,
+            "C": ["foo", "foo", "foo", "bar", "bar", "bar"] * 4,
+            "D": np.random.randn(24),
+            "E": np.random.randn(24),
+        }
+    ],
+)
+@pytest.mark.parametrize(
+    "aggfunc", ["mean", "count", {"D": "sum", "E": "count"}]
+)
+@pytest.mark.parametrize("fill_value", [0])
+def test_dataframe_pivot_table_simple(data, aggfunc, fill_value):
+    pdf = pd.DataFrame(data)
+    expected = pdf.pivot_table(
+        values=["D", "E"],
+        index=["A", "B"],
+        columns=["C"],
+        aggfunc=aggfunc,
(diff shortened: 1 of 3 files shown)


python/cudf/cudf/tests/test_reshape.py
```diff
@@ -596,6 +596,42 @@ def test_pivot_table_simple(data, aggfunc, fill_value):
...
+            "D": np.random.randn(24),
+            "E": np.random.randn(24),
+        }
+    ],
+)
+@pytest.mark.parametrize(
+    "aggfunc", ["mean", "count", {"D": "sum", "E": "count"}]
+)
+@pytest.mark.parametrize("fill_value", [0])
+def test_dataframe_pivot_table_simple(data, aggfunc, fill_value):
```
mroeschke, line 615: What the difference between this test and `test_pivot_table_simple`? I can't immediately tell

mroeschke:
Should all the arguments that aren't supported and raise `NotImplementedError` be tested as well? https://github.com/rapidsai/cudf/pull/11314/files#diff-a3d6b9fab46a4b2a1de7c903eab52bda471251b40452b1ac172d9d02a35a8f2bR1341

python/cudf/cudf/tests/test_reshape.py
bdice: The previous test uses `cudf.pivot_table`, and the new test uses `DataFrame.pivot_table`. That is the only difference. I am open to alternatives here. I initially decided against a parametrization of the previous test but could be swayed to do whatever you think is right.

bdice:
> Should all the arguments that aren't supported and raise `NotImplementedError` be tested as well? https://github.com/rapidsai/cudf/pull/11314/files#diff-a3d6b9fab46a4b2a1de7c903eab52bda471251b40452b1ac172d9d02a35a8f2bR1341

Since this code is a pure wrapper around the module-level function, I didn’t add any tests beyond some API parametrizations. We could do that for completeness but code coverage won’t change.

python/cudf/cudf/tests/test_reshape.py
mroeschke: Ah I see it now, thanks. Yeah I don't think it's worth parameterizing here.

mroeschke:
> Since this code is a pure wrapper around the module-level function, I didn’t add any tests beyond some API parametrizations. We could do that for completeness but code coverage won’t change.

SGTM

mroeschke (approved):

shwina (approved):

shwina:
@gpucibot merge

Issue #8973, opened 2021-08-05
[BUG] Can't name subset of columns with read_csv


efajardo-nv:
**Describe the bug**
Our framework uses a wrapper around cuDF `read_csv` to read CSV file input into our data pipelines. Ideally, we should be able read in any CSV just using a single call to `read_csv` with the appropriate arguments. We have run into an issue where `read_csv` returns an error when trying to name columns selected via the `usecols` argument. The `pandas` equivalent works.

**Steps/Code to reproduce bug**
```
import cudf
import pandas as pd

filename = 'foo.csv'
lines = [
  "num1,datetime,text",
  "123,2018-11-13T12:00:00,abc",
  "456,2018-11-14T12:35:01,def",
  "789,2018-11-15T18:02:59,ghi"
]

with open(filename, 'w') as fp:
    fp.write('\n'.join(lines)+'\n')
```
cuDF:
```
>>> cudf.read_csv(filename, skiprows=1, header=None, usecols=[2], names=['renamed_text_col'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/envs/rapids/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/opt/conda/envs/rapids/lib/python3.8/site-packages/cudf/io/csv.py", line 70, in read_csv
    return libcudf.csv.read_csv(
  File "cudf/_lib/csv.pyx", line 393, in cudf._lib.csv.read_csv
RuntimeError: basic_string::_M_construct null not valid
```
pandas:
```
>>> pd.read_csv(filename, skiprows=1, header=None, usecols=[2], names=['renamed_text_col'])
  renamed_text_col
0              abc
1              def
2              ghi
```

**Expected behavior**
Same result as pandas

**Environment overview (please complete the following information)**
 - Environment location: Docker
 - Method of cuDF install: Docker
   - docker pull rapidsai/rapidsai-nightly:21.08-cuda11.0-runtime-ubuntu18.04-py3.8

**Environment details**
<details><summary>Click here to see environment details</summary><pre>
     
     **git***
     commit 29b5f9ac6d24c64163349f1a5b2b5b5ef049769e (HEAD -> branch-21.10, origin/branch-21.10, origin/HEAD)
     Author: David Wendt <45795991+davidwendt@users.noreply.github.com>
     Date:   Wed Aug 4 19:11:03 2021 -0400
     
[...]

PR #12018, opened 2022-10-27
Fix issues when both `usecols` and `names` options are used in `read_csv`

vuule:
## Description
closes https://github.com/rapidsai/cudf/issues/8973
CSV reader has a few gaps in the logic for column selection and user specified column names:
1. Users cannot only specify the names of selected columns;
2. Reader fails in unpredictable ways when only a subset of column names is passed (w/o column selection);

This PR fixes the issues above. Users can now specify column names (can be lower than the actual number of columns) or names of columns selected via their indices (must match the number of indices). If selection via indices is used, the number of column names has to match either the actual number of columns, or the number of selected columns.

Also fixed test an error that went unnoticed due to issues above.

## Checklist


vuule:
rerun tests

galipremsagar (approved):
🔥 Awesome

vuule:
CC @jlowe in case you want to check if the new behavior matches the expectations

vyasr:
I started trying to give this a quick review but I don't understand the existing logic well enough. It might be best if we had an additional reviewer familiar with this part of the code; I could figure it out, but it may not be the most time-effective method here and I don't want to hold up progress. If there isn't a better person though let me know and I can give this another shot, I'll just need to familiarize myself much more with how the reader works.

jlowe:
The behavior validated in the new tests seems OK to me, but I'm wondering what happens when the specified schema is not a subset of the file?  For example, if the user says the file schema is `a: string, b: string, c: string, d: string` and asks to load columns b and d, what happens if there's only two columns in the file data?  Spark would load the second column's data for b and nulls for d in that case.

python/cudf/cudf/utils/ioutils.py
```diff
@@ -948,7 +948,8 @@
     the column names: if no names are passed, header=0;
     if column names are passed explicitly, header=None.
 names : list of str, default None
-    List of column names to be used.
+    List of column names to be used. Needs to include names of all column in
+    the file, or names of all columns selected using `usecols` (indices only).
```
karthikeyann, line 952: indices only? pytest example has column names too.

cpp/tests/io/csv_test.cpp
```diff
@@ -2260,4 +2260,64 @@ TEST_F(CsvReaderTest, CsvDefaultOptionsWriteReadMatch)
   EXPECT_EQ(new_table_and_metadata.metadata.column_names[1], "1");
 }
 
+TEST_F(CsvReaderTest, UseColsValidation)
+{
+  std::string buffer = "1,2,3";
```
karthikeyann, line 2265:
```suggestion
  const std::string buffer = "1,2,3";
```

cpp/tests/io/csv_test.cpp
```diff
@@ -2260,4 +2260,64 @@ TEST_F(CsvReaderTest, CsvDefaultOptionsWriteReadMatch)
...
+  cudf::io::csv_reader_options bad_name_options =
+    cudf::io::csv_reader_options::builder(cudf::io::source_info{buffer.c_str(), buffer.size()})
+      .names({"a", "b", "c"})
+      .use_cols_names({"nonexistent_name"});
+  EXPECT_THROW(cudf::io::read_csv(bad_name_options), cudf::logic_error);
+}
+
+TEST_F(CsvReaderTest, CropColumns)
+{
+  std::string csv_in{"12,9., 10\n34,8., 20\n56,7., 30"};
```
karthikeyann, line 2288:
```suggestion
  const std::string csv_in{"12,9., 10\n34,8., 20\n56,7., 30"};
```

cpp/tests/io/csv_test.cpp
```diff
@@ -2260,4 +2260,64 @@ TEST_F(CsvReaderTest, CsvDefaultOptionsWriteReadMatch)
   EXPECT_EQ(new_table_and_metadata.metadata.column_names[1], "1");
 }
 
+TEST_F(CsvReaderTest, UseColsValidation)
+{
+  std::string buffer = "1,2,3";
+
+  cudf::io::csv_reader_options idx_cnt_options =
```
karthikeyann, line 2267:
```suggestion
  const cudf::io::csv_reader_options idx_cnt_options =
```
use const at other places too.

cpp/tests/io/csv_test.cpp
```diff
@@ -2260,4 +2260,64 @@ TEST_F(CsvReaderTest, CsvDefaultOptionsWriteReadMatch)
...
+TEST_F(CsvReaderTest, CropColumns)
+{
+  std::string csv_in{"12,9., 10\n34,8., 20\n56,7., 30"};
+
+  cudf::io::csv_reader_options in_opts =
+    cudf::io::csv_reader_options::builder(cudf::io::source_info{csv_in.c_str(), csv_in.size()})
+      .dtypes(std::vector<data_type>{dtype<int32_t>(), dtype<float>()})
+      .names({"a", "b"})
+      .header(-1);
+  auto result = cudf::io::read_csv(in_opts);
```
karthikeyann, line 2295:
```suggestion
  const auto result = cudf::io::read_csv(in_opts);
```

python/cudf/cudf/utils/ioutils.py
vuule: the requirement to have the same number of elements as `usecols` is limited to the case when `usecols` contains indices.
I included the failing cases in the C++ test `UseColsValidation`, the options object names indicate the reason why it should fail.
I can add comments there to make it clearer. Also, let me know if this comment should be worded differently to be clearer.

karthikeyann: `(indices only)` is only confusing. does `indices only`  mean `usecols` uses indices, but not column names?

Rest all looks good.

vuule:
> The behavior validated in the new tests seems OK to me, but I'm wondering what happens when the specified schema is not a subset of the file? For example, if the user says the file schema is `a: string, b: string, c: string, d: string` and asks to load columns b and d, what happens if there's only two columns in the file data? Spark would load the second column's data for b and nulls for d in that case.

Pandas does not support this, but cuDF does (that feels odd to type about `read_csv` :D). User can pass additional column names and data types and null columns of the given type will be returned.

The one limitation we have here is that you cannot select columns by index if passing these additional names that don't exist in the file (unless selecting all columns in the input schema). Selection via column names works fine.
This matches what Pandas does, but I think I can remove the constraint. @jlowe Let me know if  this is a case we need to support.

python/cudf/cudf/utils/ioutils.py
vuule: expanded the related comments, I hope they make sense now.

cpp/src/io/csv/reader_impl.cu
```diff
@@ -678,25 +678,29 @@ table_with_metadata read_csv(cudf::io::datasource* source,
...
+    (
+      // no data to detect (the number of) columns
+      detected_column_names.empty() or
+      // number of user specified names matches what is detected
+      reader_opts.get_names().size() == detected_column_names.size() or
+      // Columns are not selected by indices; read first reader_opts.get_names().size() columns
+      unique_use_cols_indexes.empty());
+  auto column_names = opts_have_all_col_names ? reader_opts.get_names() : detected_column_names;
+
+  auto num_actual_columns = static_cast<int32_t>(column_names.size());
```
karthikeyann, line 698: should this be `const`?

cpp/src/io/csv/reader_impl.cu
```diff
@@ -757,24 +761,40 @@ table_with_metadata read_csv(cudf::io::datasource* source,
...
 
-    for (const auto& name : reader_opts.get_use_cols_names()) {
-      const auto it = std::find(column_names.begin(), column_names.end(), name);
-      if (it != column_names.end()) {
-        auto curr_it = it - column_names.begin();
-        if (column_flags[curr_it] == column_parse::disabled) {
-          column_flags[curr_it] = column_parse::enabled | column_parse::inferred;
-          num_active_columns++;
-        }
+  if (not reader_opts.get_use_cols_names().empty()) {
```
karthikeyann, line 790: nit: `if` case above uses `unique_use_cols_indexes.empty()` condition. This one uses `reader_opts.get_use_cols_names().empty())`, not the `unique_use_cols_names.empty()`. Is there a reason for using `reader_opts` here in the condition? would it make sense to use `unique_use_cols_names.empty()` here?

karthikeyann (approved):
LGTM 👍

Minor nitpicks. Look good.

nvdbaranec (commented):
One general comment.  There's a number of blocks of mildly complex logic that would be well-served with some simple comments on what they do.  Examples:  Lines, 702, 773, 789.  Descriptions of what each block is intended for would help with the overall flow of reading.

cpp/src/io/csv/reader_impl.cu
```diff
@@ -677,25 +677,29 @@ table_with_metadata read_csv(cudf::io::datasource* source,
...
-  } else {
-    column_names = get_column_names(
-      header, parse_opts.view(), reader_opts.get_header(), reader_opts.get_prefix());
-
-    num_actual_columns = num_active_columns = column_names.size();
-
-    column_flags.resize(num_actual_columns, column_parse::enabled | column_parse::inferred);
-
+  auto num_records                   = std::max(row_offsets.size(), 1ul) - 1;
+  auto const unique_use_cols_indexes = std::set<int>(reader_opts.get_use_cols_indexes().cbegin(),
```
vyasr, line 681: CTAD should work here right?
```suggestion
  auto const unique_use_cols_indexes = std::set(reader_opts.get_use_cols_indexes().cbegin(),
```

cpp/src/io/csv/reader_impl.cu
```diff
@@ -677,25 +677,29 @@ table_with_metadata read_csv(cudf::io::datasource* source,
...
+
+  auto const detected_column_names =
+    get_column_names(header, parse_opts.view(), reader_opts.get_header(), reader_opts.get_prefix());
+  auto const opts_have_all_col_names =
+    not reader_opts.get_names().empty() and
+    (
+      // no data to detect (the number of) columns
+      detected_column_names.empty() or
+      // number of user specified names matches what is detected
+      reader_opts.get_names().size() == detected_column_names.size() or
```
vyasr, line 692: Probably a stupid edge case, but what happens if the user specifies the right number of column names but the names don't match (e.g. they're misspelled)? It looks like the current code will silently just return all columns in the file. Should there be some sort of error or log indicating that the user requested some nonexistent names?

cpp/src/io/csv/reader_impl.cu
```diff
@@ -756,24 +760,40 @@ table_with_metadata read_csv(cudf::io::datasource* source,
   }
 
   // User can specify which columns should be parsed
-  if (!reader_opts.get_use_cols_indexes().empty() || !reader_opts.get_use_cols_names().empty()) {
+  auto const unique_use_cols_names = std::unordered_set<std::string>(
```
vyasr, line 763: CTAD?
```suggestion
  auto const unique_use_cols_names = std::unordered_set(
```

cpp/src/io/csv/reader_impl.cu
```diff
@@ -756,24 +760,40 @@ table_with_metadata read_csv(cudf::io::datasource* source,
...
+    reader_opts.get_use_cols_names().cbegin(), reader_opts.get_use_cols_names().cend());
+  auto const is_column_selection_used =
+    not unique_use_cols_names.empty() or not unique_use_cols_indexes.empty();
+
+  if (is_column_selection_used) {
     std::fill(column_flags.begin(), column_flags.end(), column_parse::disabled);
+    num_active_columns = 0;
```
vyasr, line 770: What makes a column "active"? I'm not sure why this is set to 0 here.

cpp/src/io/csv/reader_impl.cu
```diff
@@ -756,24 +760,40 @@ table_with_metadata read_csv(cudf::io::datasource* source,
...
-        auto curr_it = it - column_names.begin();
-        if (column_flags[curr_it] == column_parse::disabled) {
-          column_flags[curr_it] = column_parse::enabled | column_parse::inferred;
-          num_active_columns++;
-        }
+  if (not unique_use_cols_names.empty()) {
+    for (auto const& name : unique_use_cols_names) {
+      auto const it = std::find(column_names.cbegin(), column_names.cend(), name);
+      CUDF_EXPECTS(it != column_names.end(), "Nonexistent column selected");
+      auto const curr_it = it - column_names.begin();
```
vyasr, line 793:
```suggestion
      auto const curr_it = std::distance(it, column_names.begin());
```

cpp/src/io/csv/reader_impl.cu
```diff
@@ -677,25 +677,29 @@ table_with_metadata read_csv(cudf::io::datasource* source,
...
-    column_names = reader_opts.get_names();
-  } else {
-    column_names = get_column_names(
-      header, parse_opts.view(), reader_opts.get_header(), reader_opts.get_prefix());
-
-    num_actual_columns = num_active_columns = column_names.size();
-
-    column_flags.resize(num_actual_columns, column_parse::enabled | column_parse::inferred);
-
+  auto num_records                   = std::max(row_offsets.size(), 1ul) - 1;
```
ttnghia, line 680: `const`? 
In addition, this variable is used very late in this function thus it should be defined as late as possible, right before usage.

cpp/src/io/csv/reader_impl.cu
```diff
@@ -677,25 +677,29 @@ table_with_metadata read_csv(cudf::io::datasource* source,
...
+  auto const opts_have_all_col_names =
+    not reader_opts.get_names().empty() and
+    (
+      // no data to detect (the number of) columns
+      detected_column_names.empty() or
+      // number of user specified names matches what is detected
+      reader_opts.get_names().size() == detected_column_names.size() or
+      // Columns are not selected by indices; read first reader_opts.get_names().size() columns
+      unique_use_cols_indexes.empty());
+  auto column_names = opts_have_all_col_names ? reader_opts.get_names() : detected_column_names;
```
ttnghia, line 695: `const`?

cpp/src/io/csv/reader_impl.cu
```diff
@@ -677,25 +677,29 @@ table_with_metadata read_csv(cudf::io::datasource* source,
...
+
+  auto const num_actual_columns = static_cast<int32_t>(column_names.size());
+  auto num_active_columns       = num_actual_columns;
+  auto column_flags             = std::vector<column_parse::flags>(
```
ttnghia, line 699: `const` if possible?

cpp/src/io/csv/reader_impl.cu
ttnghia: `_it` implies iterator. If this is distance value then let's call `curr_dist`.

ttnghia: In addition, `std::distance(it, column_names.begin());` will yield negative value. Should be the opposite order: `std::distance(column_names.begin(), it);`.

vyasr: Whoops sorry, good catch. And yes, I agree, I think the current name is misleading and I meant to comment on that but forgot while writing the std::distance suggestiong.

cpp/src/io/csv/reader_impl.cu
vuule: The names specified through `names` option (accessor `get_names()` used here) are meant to override the names in the file (if the header is present). This option does not mean column selection, names in `get_names` don't need to match the names in the file. 
Maybe you mixed it up with `use_cols_names` option (accessor `get_use_cols_names()`), which does column selection by their names.

cpp/src/io/csv/reader_impl.cu
vuule: Can't, names are potentially modified in a few places later in the code (empty names -> "Unnamed: col_index", mangle duplicates, apply `names` to selected column when index-based column selection is used).

cpp/src/io/csv/reader_impl.cu
vuule: Active means that it's selected and included in the output dataframe. If we don't do column selection, all columns are active (we read all columns). If column selection is used, we clear active columns (i.e. `fill` in the line above) and then re-set flags for the selected columns. `num_active_columns` is updated to match `column_flags`.

cpp/src/io/csv/reader_impl.cu
vyasr: Oh I see, that makes sense. OK cool.

cpp/src/io/csv/reader_impl.cu
vuule: renamed to `col_idx` and used `distance`

cpp/src/io/csv/reader_impl.cu
vuule: moved to the point of use + const.

cpp/src/io/csv/reader_impl.cu
vuule: not possible, flags are modified if column selection is used.

vyasr (approved):

ttnghia (approved):

nvdbaranec (approved):

vuule:
rerun tests

vuule:
@gpucibot merge

PR #12021, opened 2022-10-28
fix bug in csv_reader_options construction in cython

karthikeyann:
## Description
Fixes bug in csv_reader_options construction in cython
The false values for csv were not passed to the csv_reader_options during construction in cython code. This is fixed and a unit test is added.

## Checklist

### DIFF
--- a/python/cudf/cudf/_lib/csv.pyx
+++ b/python/cudf/cudf/_lib/csv.pyx
@@ -304,7 +304,7 @@ cdef csv_reader_options make_csv_reader_options(
 
     if false_values is not None:
         c_false_values.reserve(len(false_values))
-        for fv in c_false_values:
+        for fv in false_values:
             c_false_values.push_back(fv.encode())
         csv_reader_options_c.set_false_values(c_false_values)
 
--- a/python/cudf/cudf/tests/test_csv.py
+++ b/python/cudf/cudf/tests/test_csv.py
@@ -766,6 +766,43 @@ def test_csv_reader_bools(tmpdir, names, dtypes, data, trues, falses):
     assert_eq(df_out, out)
 
 
+def test_csv_reader_bools_NA():
+    names = ["text", "int"]
+    dtypes = ["str", "int"]
+    trues = ["foo"]
+    falses = ["bar"]
+    lines = [
+        ",".join(names),
+        "true,true",
+        "false,false",
+        "foo,foo",
+        "bar,bar",
+        "qux,qux",
+    ]
+
+    buffer = "\n".join(lines)
+
+    df = read_csv(
+        StringIO(buffer),
+        names=names,
+        dtype=dtypes,
+        skiprows=1,
+        true_values=trues,
+        false_values=falses,
+    )
+    assert len(df.columns) == 2
+    assert df["text"].dtype == np.dtype("object")
+    assert df["int"].dtype == np.dtype("int64")
+    expected = pd.DataFrame(
+        {
+            "text": ["true", "false", "foo", "bar", "qux"],
+            "int": [1.0, 0.0, 1.0, 0.0, np.nan],
+        }
+    )
+    # breaking behaviour is np.nan for qux
+    assert_eq(df, expected)
+
+
 def test_csv_quotednumbers(tmpdir):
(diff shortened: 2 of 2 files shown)


python/cudf/cudf/tests/test_csv.py
```diff
@@ -766,6 +766,43 @@ def test_csv_reader_bools(tmpdir, names, dtypes, data, trues, falses):
     assert_eq(df_out, out)
 
 
+def test_csv_reader_bools_NA():
```
galipremsagar, line 769: Pandas csv reader has `false_values` support, can we not write this pytest by comparing against pandas's csv reader output rather than hand-coding the comparison?

karthikeyann: Unknown values `"qux"` returns null in our reader, but pandas fails to parse. That's is why comparison against pandas was not done in this test.

python/cudf/cudf/tests/test_csv.py
```diff
@@ -766,6 +766,43 @@ def test_csv_reader_bools(tmpdir, names, dtypes, data, trues, falses):
...
+        true_values=trues,
+        false_values=falses,
+    )
+    assert len(df.columns) == 2
+    assert df["text"].dtype == np.dtype("object")
+    assert df["int"].dtype == np.dtype("int64")
+    expected = pd.DataFrame(
+        {
+            "text": ["true", "false", "foo", "bar", "qux"],
+            "int": [1.0, 0.0, 1.0, 0.0, np.nan],
```
karthikeyann, line 799:
```suggestion
            "int": [1, 0, 1, 0, 0],
```

python/cudf/cudf/tests/test_csv.py
galipremsagar: In that case can we have two separate tests? One validating against pandas and one validating for `qux`?

bdice: Testing against pandas / splitting this test would be great.

vyasr (commented):
The fix here looks fine. I'll let the other reviewers follow up on the question of testing.

python/cudf/cudf/tests/test_csv.py
```diff
@@ -766,6 +766,80 @@ def test_csv_reader_bools(tmpdir, names, dtypes, data, trues, falses):
...
+    assert df["int"].dtype == np.dtype("int64")
+    expected = pd.read_csv(
+        StringIO(buffer),
+        names=names,
+        dtype=dtypes,
+        skiprows=1,
+        true_values=trues,
+        false_values=falses,
+    )
+    assert_eq(df, expected)
```
galipremsagar, line 803:
```suggestion
    expected = pd.read_csv(
        StringIO(buffer),
        names=names,
        dtype=dtypes,
        skiprows=1,
        true_values=trues,
        false_values=falses,
    )
    assert_eq(df, expected, check_dtype=True)
```

galipremsagar: This should would be equivalent of `check_dtype=True`

galipremsagar (approved):

karthikeyann:
Pandas behaviour:
bool literals `true`, `false`, user passed `true_values`, `false_values`, are not considered as valid literal while parsing as integer.
Also, `"0"` is not considered as false while parsing as bool. similarly for `"1"`. 

Our CSV parser:
bool literals are converted to int. 
`"0"` considered as false, `"1"` considered as true.

These are differences between our parsers and pandas. We should decide where we should deviate from pandas parsers.

python/cudf/cudf/tests/test_csv.py
karthikeyann: done. There are differences between our parses and pandas.

python/cudf/cudf/_lib/csv.pyx
```diff
@@ -297,7 +297,7 @@ cdef csv_reader_options make_csv_reader_options(
 
     if false_values is not None:
         c_false_values.reserve(len(false_values))
-        for fv in c_false_values:
+        for fv in false_values:
```
wence-, line 300: Oops :)

wence- (approved):

karthikeyann:
@gpucibot merge

PR #11898, opened 2022-10-11
Trim quotes for non-string values in nested json parsing

karthikeyann:
## Description
Trim quotes for non-string values in nested json parsing
Added corner cases for unquoted and quoted literals. (Review the unit test)
Fixes old json reader to treat `"null"` as string instead of NULL.

closes #11817

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12022, opened 2022-10-28
Null element for parsing error in numeric types in JSON, CSV reader

karthikeyann:
## Description
This PR will change behaviour of nested json reader to return null element on parsing error of numeric types.
- Uses `std::optional` for indicating parsing failure in `parse_numeric`
- Cleanup
  - Removed `decode_value` as it only gives only specialization for timestamp and duration types, rest of types are passthrough.
  - Unified `decode_digit`

Breaking behaviour in read_csv, and read_json:
 - old behaviour: parsing error returns quiet NAN for that type as result for that row.
 - new behaviour: parsing error returns null for that row.

unit tests added.

## Checklist

### DIFF
--- a/cpp/src/io/utilities/parsing_utils.cuh
+++ b/cpp/src/io/utilities/parsing_utils.cuh
@@ -117,8 +117,9 @@ struct parse_options {
 };
 
 /**
- * @brief Returns the numeric value of an ASCII/UTF-8 character. Specialization
- * for integral types. Handles hexadecimal digits, both uppercase and lowercase.
+ * @brief Returns the numeric value of an ASCII/UTF-8 character.
+ * Handles hexadecimal digits, both uppercase and lowercase
+ * for integral types and only decimal digits for floating point types.
  * If the character is not a valid numeric digit then `0` is returned and
  * valid_flag is set to false.
  *
@@ -127,31 +128,14 @@ struct parse_options {
  *
  * @return uint8_t Numeric value of the character, or `0`
  */
-template <typename T, CUDF_ENABLE_IF(std::is_integral_v<T>)>
+template <typename T, bool as_hex = false>
 constexpr uint8_t decode_digit(char c, bool* valid_flag)
 {
   if (c >= '0' && c <= '9') return c - '0';
-  if (c >= 'a' && c <= 'f') return c - 'a' + 10;
-  if (c >= 'A' && c <= 'F') return c - 'A' + 10;
-
-  *valid_flag = false;
-  return 0;
-}
-
-/**
- * @brief Returns the numeric value of an ASCII/UTF-8 character. Specialization
- * for non-integral types. Handles only decimal digits. If the character is not
- * a valid numeric digit then `0` is returned and valid_flag is set to false.
- *
- * @param c ASCII or UTF-8 character
- * @param valid_flag Set to false if input is not valid. Unchanged otherwise.
- *
- * @return uint8_t Numeric value of the character, or `0`
- */
-template <typename T, CUDF_ENABLE_IF(!std::is_integral_v<T>)>
-constexpr uint8_t decode_digit(char c, bool* valid_flag)
-{
-  if (c >= '0' && c <= '9') return c - '0';
+  if constexpr (as_hex and std::is_integral_v<T>) {
+    if (c >= 'a' && c <= 'f') return c - 'a' + 10;
+    if (c >= 'A' && c <= 'F') return c - 'A' + 10;
+  }
 
   *valid_flag = false;
   return 0;
@@ -194,13 +178,13 @@ constexpr bool is_infinity(char const* begin, char const* end)
  * @return The parsed and converted value
  */
 template <typename T, int base = 10>
-constexpr T parse_numeric(char const* begin,
-                          char const* end,
-                          parse_options_view const& opts,
-                          T error_result = std::numeric_limits<T>::quiet_NaN())
+__host__ __device__ std::optional<T> parse_numeric(char const* begin,
+                                                   char const* end,
+                                                   parse_options_view const& opts)
 {
   T value{};
   bool all_digits_valid = true;
+  constexpr bool as_hex = (base == 16);
 
   // Handle negative values if necessary
   int32_t sign = (*begin == '-') ? -1 : 1;
@@ -223,7 +207,7 @@ constexpr T parse_numeric(char const* begin,
     } else if (base == 10 && (*begin == 'e' || *begin == 'E')) {
       break;
     } else if (*begin != opts.thousands && *begin != '+') {
-      value = (value * base) + decode_digit<T>(*begin, &all_digits_valid);
+      value = (value * base) + decode_digit<T, as_hex>(*begin, &all_digits_valid);
     }
     ++begin;
   }
@@ -237,7 +221,7 @@ constexpr T parse_numeric(char const* begin,
         break;
       } else if (*begin != opts.thousands && *begin != '+') {
         divisor /= base;
-        value += decode_digit<T>(*begin, &all_digits_valid) * divisor;
+        value += decode_digit<T, as_hex>(*begin, &all_digits_valid) * divisor;
       }
       ++begin;
     }
@@ -248,12 +232,12 @@ constexpr T parse_numeric(char const* begin,
       if (*begin == '-' || *begin == '+') { ++begin; }
       int32_t exponent = 0;
       while (begin < end) {
-        exponent = (exponent * 10) + decode_digit<T>(*(begin++), &all_digits_valid);
+        exponent = (exponent * 10) + decode_digit<T, as_hex>(*(begin++), &all_digits_valid);
       }
       if (exponent != 0) { value *= exp10(double(exponent * exponent_sign)); }
     }
   }
-  if (!all_digits_valid) { return error_result; }
+  if (!all_digits_valid) { return {}; }
 
   return value * sign;
 }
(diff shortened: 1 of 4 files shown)


vyasr:
This PR looks like it contains changes from both #12021 and #11898 so I'll hold off reviewing until those (or at least #11898, #12021 is a trivial change other than the test additions) are merged.

davidwendt:
Does this violate the [libcudf does not introspect data](https://github.com/rapidsai/cudf/blob/branch-22.12/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#libcudf-does-not-introspect-data) rule?
I would like to consider that we do not do this when parsing invalid data in other libcudf APIs.
Perhaps this has already been discussed. 
Is there any performance decrease to add this breaking change? We probably don't want to penalize the performance for good data.

Maybe we need a parse function that only validates the data without building the output so the caller can optionally build a null-mask? I'm thinking another parse option that is `validate-only` for example. And returns bool column(s) identifying invalid data.

revans2:
> Does this violate the [libcudf does not introspect data](https://github.com/rapidsai/cudf/blob/branch-22.12/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#libcudf-does-not-introspect-data) rule? I would like to consider that we do not do this when parsing invalid data in other libcudf APIs. Perhaps this has already been discussed. Is there any performance decrease to add this breaking change? We probably don't want to penalize the performance for good data.
> 
> Maybe we need a parse function that only validates the data without building the output so the caller can optionally build a null-mask? I'm thinking another parse option that is `validate-only` for example. And returns bool column(s) identifying invalid data.

For Parquet and ORC we do some minimal error checking. In those cases generally if the input file is corrupted an exception is thrown for the entire file instead of trying to return a single value as a null. The problem with enforcing the rule of no introspection on I/O, specifically the input part of I/O, is that the data is inherently untrusted. Even if the file was written by CUDF bit rot is a real world problem and we cannot trust that what we wrote is what will be read. All CPU implementations will do some form of error checking, so if we say you have to parse it twice, once to look for errors and once to read the data, there is no situation I know of where we would not want the error checking, so it would slow down all GPU implementations.

That said I personally think that we might want to split the processing for text based formats like JASON and CSV at a different level. For Spark we ask CUDF to just return strings and we handle doing the parsing and error checking. Part of this is because the Spark parsing/error checking is not consistent with C++ and python parsing. But this is still going to run into issues when the structure of the nested data does not match the requested structure. What happens when we expect a LIST, but see a single value? Should the list be null? Should there be a boolean column to indicate that an error happened? Should it return a list with a single value? What happens if CUDF decides to do something differently than what Spark wants?

To me I want to see a few different public APIs.  One that tokenizes the raw inputs and lets the user decide how to handle the tokens. Just because there will be corner cases that everyone wants to handle differently. The other should be a reference/default implementation that does some form of error checking that matches what a general user would expect.

karthikeyann:
> What happens when we expect a LIST, but see a single value? Should the list be null? Should there be a boolean column to indicate that an error happened? Should it return a list with a single value? What happens if CUDF decides to do something differently than what Spark wants?

The value is treated as null now.

Right now, cudf json parsers does not report _type parsing errors_. If this error should to be reported, it will be a significant breaking behavior. Existing workflows might break. If this is desired, a detailed discussion is needed.

For spark, `cudf::detail` APIs for JSON tokenization can be added in future. The tree structure need to be traversed by Spark. (created an issue https://github.com/rapidsai/cudf/issues/12051).

cpp/src/io/utilities/parsing_utils.cuh
```diff
@@ -670,13 +597,15 @@ struct ConvertFunctor {
                                                       parse_options_view const& opts,
                                                       bool as_hex)
   {
+    // TODO what's invalid input
+    bool all_digits_valid = true;
     static_cast<device_storage_type_t<T>*>(out_buffer)[row] =
-      [&opts, output_type, begin, end]() -> device_storage_type_t<T> {
+      [&opts, &all_digits_valid, output_type, begin, end]() -> device_storage_type_t<T> {
```
vyasr, line 603: Should `all_digits_valid` be getting used somewhere? I'm not seeing anything that could set it to false right now.

cpp/src/strings/json/json_path.cu
```diff
@@ -570,9 +570,10 @@ class path_state : private parser {
...
               op.expected_type = OBJECT;
             } else {
-              op.type  = path_operator_type::CHILD_INDEX;
-              op.index = cudf::io::parse_numeric<int>(
-                op.name.data(), op.name.data() + op.name.size_bytes(), json_opts, -1);
+              op.type          = path_operator_type::CHILD_INDEX;
+              auto const value = cudf::io::parse_numeric<int>(
+                op.name.data(), op.name.data() + op.name.size_bytes(), json_opts);
+              op.index = value.value_or(-1);
               CUDF_EXPECTS(op.index >= 0, "Invalid numeric index specified in JSONPath");
```
vyasr, line 577: Minor nit, rather than assigning and checking you could also do `CUDF_EXPECTS(value.has_value...); op.index = *value;`. Not sure which you think is cleaner.

karthikeyann: I made minimal change. `has_value` should be cleaner.

cpp/src/io/utilities/parsing_utils.cuh
karthikeyann: It's not used now. It should be used when we decide what is a parsing error for decimal, timestamp and duration types. Right now, it is not decided, just a placeholder. Should I replace it with `true`?

cpp/src/io/utilities/parsing_utils.cuh
```diff
@@ -670,13 +597,15 @@ struct ConvertFunctor {
...
       return strings::detail::parse_decimal<device_storage_type_t<T>>(
         begin, end, output_type.scale());
     }();
 
-    return true;
+    return all_digits_valid;
```
karthikeyann, line 608:
```suggestion
    return true;
```

cpp/src/io/utilities/parsing_utils.cuh
```diff
@@ -670,13 +597,15 @@ struct ConvertFunctor {
                                                       parse_options_view const& opts,
                                                       bool as_hex)
   {
+    // TODO what's invalid input
+    bool all_digits_valid = true;
```
karthikeyann, line 601:
```suggestion
```

cpp/src/io/utilities/parsing_utils.cuh
karthikeyann:
```suggestion
      [&opts, output_type, begin, end]() -> device_storage_type_t<T> {
```

karthikeyann:
rerun tests

cpp/src/strings/json/json_path.cu
karthikeyann: what if a negative number is given in`op.name`.

vyasr: You're right, that would have been a problem. The current solution looks good.

cpp/tests/io/json_test.cpp
```diff
@@ -1451,4 +1452,87 @@ TEST_F(JsonReaderTest, JsonNestedDtypeSchema)
...
+  std::string data = (test_opt == json_test_t::json_lines_row_orient) ? row_orient : record_orient;
+
+  auto make_validity = [](std::vector<int> const& validity) {
+    return cudf::detail::make_counting_transform_iterator(
+      0, [=](auto i) -> bool { return static_cast<bool>(validity[i]); });
+  };
+
+  constexpr int iNA    = -1;
+  constexpr double dNA = std::numeric_limits<double>::quiet_NaN();
+  constexpr bool bNA   = false;
```
vyasr, line 1492: I know it makes things longer, but I would prefer these names spelled out at least to the normal type names (e.g. int_NA, probably integer_NA is overlong and adds no info). I didn't immediately grasp what the names meant.

vyasr (approved):
Looks good. One request, but feel free to ignore if you disagree with the verbosity it adds.

cpp/src/io/utilities/parsing_utils.cuh
```diff
@@ -248,12 +232,12 @@ constexpr T parse_numeric(char const* begin,
...
       int32_t exponent = 0;
       while (begin < end) {
-        exponent = (exponent * 10) + decode_digit<T>(*(begin++), &all_digits_valid);
+        exponent = (exponent * 10) + decode_digit<T, as_hex>(*(begin++), &all_digits_valid);
       }
       if (exponent != 0) { value *= exp10(double(exponent * exponent_sign)); }
     }
   }
-  if (!all_digits_valid) { return error_result; }
+  if (!all_digits_valid) { return {}; }
```
vuule, line 240: Let's be more explicit here. Current code could be mistaken for returning a zero.
```suggestion
  if (!all_digits_valid) { return std::nullopt; }
```

vuule:
CC @galipremsagar for viz and Python perspective on this behavior change

vuule (commented):
Few small suggestions.
It's not obvious at first glance what the desired behavior is, especially since Pandas uses `NaN` (for it's own reasons). The main benefit of using `null` here is that we used to return zeroes for invalid integers :( Not a great stand-in for `NaN`.

python/cudf/cudf/tests/test_csv.py
```diff
@@ -834,7 +834,7 @@ def test_csv_reader_bools_NA():
     expected = pd.DataFrame(
         {
             "text": ["true", "false", "foo", "bar", "qux"],
-            "int": [1, 0, 1, 0, 0],
+            "int": [1.0, 0.0, 1.0, 0.0, np.nan],
```
galipremsagar, line 837: So WDPD(What does pandas do)? Error here saying `qux` isn't a decodable value?

cpp/tests/io/json_test.cpp
```diff
@@ -1451,4 +1452,87 @@ TEST_F(JsonReaderTest, JsonNestedDtypeSchema)
...
+
+  std::vector<int> const validity_int   = {1, 1, 0, 1, 1, 0, 0, 0, 0, 1, 1, 0, 1, 1, 0, 0};
+  std::vector<int> const validity_float = {1, 1, 0, 1, 1, 0, 0, 0, 0, 1, 1, 0, 1, 1, 0, 0};
+  std::vector<int> const validity_bool  = {1, 1, 0, 1, 1, 0, 0, 0, 0, 1, 1, 0, 1, 1, 0, 0};
```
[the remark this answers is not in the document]
karthikeyann: That's right. 👍 . need to cleanup.

cpp/src/io/utilities/parsing_utils.cuh
karthikeyann: Compiles locally, but CI build gives this error 
`04:37:13 $SRC_DIR/cpp/src/io/utilities/parsing_utils.cuh(240): error #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed`

The constructor from nullopt_t arg might be a host function. not constexpr.

karthikeyann:
0e45d133ce3631ac8b74d8188dc1c9d3361ee717
This commit reverts change of behaviour of nested json reader to return null element on parsing error of numeric types.
Breaking behaviour in read_csv, and read_json:
 - old behaviour: parsing error returns quiet NAN for that type as result for that row.
 - new behaviour: parsing error returns null for that row.

pushing this breaking change to next release.

galipremsagar (approved):

python/cudf/cudf/tests/test_csv.py
galipremsagar: Discussed offline and decided to push these changes to later.

cpp/src/io/utilities/parsing_utils.cuh
```diff
@@ -670,6 +597,7 @@ struct ConvertFunctor {
                                                       parse_options_view const& opts,
                                                       bool as_hex)
   {
+    // TODO what's invalid input
```
vuule, line 600: Can you please make these TODOs clearer if we're planning to merge them?
I looked into the code and don't see that invalid input is handled in a robust way in `parse_decimal` and `to_timestamp`. Is that what this TODO needs to remind us of?

vuule (approved):
Nice improvement.

cpp/src/io/utilities/parsing_utils.cuh
karthikeyann: Yes. this `TODO` is a reminder for future and also to remind reviewers to discuss and decide what is invalid input for this type.

karthikeyann:
@gpucibot merge

cpp/src/io/utilities/parsing_utils.cuh
newjavaer: @karthikeyann Hi,I also encountered this problem, did you solve this problem?

karthikeyann: Yes. By callling default constructor of std::optional
`if (!all_digits_valid) { return std::optional<T>{}; }`

PR #12024, opened 2022-10-28
Add device buffer datasource

### LINKED ISSUE #11077: [FEA] Add device buffer datasource
There is a use case for JSON and CSV reader where the input data is already on the GPU (in the form of string column, issue https://github.com/rapidsai/cudf/issues/10688). One part of support for this kind of input is to add a new datasource type that takes a device memory buffer.

PointKernel:
## Description
Closes #11077 

This PR adds device buffer datasource.

Parquet benchmark results show that `DEVICE_BUFFER` is about 10~25% more efficient than `HOST_BUFFER` and in ORC reader benchmarks, `DEVICE_BUFFER` is about 10% to 3.5x faster.

## Checklist


cpp/include/cudf/io/datasource.hpp
```diff
@@ -104,13 +104,21 @@ class datasource {
...
    */
   static std::unique_ptr<datasource> create(host_buffer const& buffer);
 
+  /**
+   * @brief Creates a source from a device memory buffer.
+   *
+   * @param buffer Device buffer object
+   * @return Constructed datasource object
+   */
+  static std::unique_ptr<datasource> create(cudf::device_span<uint8_t> buffer);
```
vuule, line 120: should be able to be const
```suggestion
  static std::unique_ptr<datasource> create(cudf::device_span<uint8_t const> buffer);
```

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +199,47 @@ class direct_read_source : public file_source {
...
+/**
+ * @brief Implementation class for reading from a device buffer source
+ */
+class device_buffer_source : public datasource {
+ public:
+  explicit device_buffer_source(cudf::device_span<uint8_t> d_buffer) : _d_buffer{d_buffer} {}
+
+  size_t host_read(size_t offset, size_t size, uint8_t* dst) final
+  {
+    CUDF_FAIL("Host read shouldn't be used with device_buffer_source");
```
vuule, line 211: we can copy to host when required

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +199,47 @@ class direct_read_source : public file_source {
...
+  }
+
+  std::unique_ptr<buffer> host_read(size_t offset, size_t size) final
+  {
+    CUDF_FAIL("Host read shouldn't be used with device_buffer_source");
+  }
+
+  [[nodiscard]] bool supports_device_read() const final { return true; }
```
vuule, line 219: Need to use `override` for all functions defined here. Without `virtual` or `override`, these are not virtual functions, just overloads of the parent functions.

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +199,47 @@ class direct_read_source : public file_source {
...
+
+  size_t device_read(size_t offset, size_t size, uint8_t* dst, rmm::cuda_stream_view stream) final
```
vuule, line 221: You have to implement `device_read_async` as well.
This one's a bit tricky. I'm trying to figure out if we can just do an async D2D copy and return a finished task.

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +199,47 @@ class direct_read_source : public file_source {
...
+  {
+    CUDF_FAIL("TODO: should this be invoked?");
+    // TODO: to be finished with deep copy
```
vuule, line 224: darn, forgot that we generally read into a pre-allocated buffer. Not really a no-op, but should still be much faster than H2D.
I guess this one should be implemented as sync D2D copy on the given stream.

vuule (commented):
initial review to hopefully help with debugging

cpp/src/io/utilities/datasource.cpp
PointKernel: Somehow I remember `final` implies `override`. Corrected.

cpp/tests/io/parquet_test.cpp
```diff
@@ -1087,7 +1087,7 @@ TEST_F(ParquetWriterTest, MultiIndex)
   cudf::test::expect_metadata_equal(expected_metadata, result.metadata);
 }
 
-TEST_F(ParquetWriterTest, HostBuffer)
+TEST_F(ParquetWriterTest, BufferSource)
```
PointKernel, line 1090: I can also leave the current test as it is and add another `DeviceBuffer` test case. The main reason to update existing tests is to minimize duplicates.

cpp/include/cudf/io/types.hpp
```diff
@@ -169,6 +171,25 @@ struct host_buffer {
   host_buffer(const char* data, size_t size) : data(data), size(size) {}
 };
 
+/**
+ * @brief Non-owning view of a device memory buffer
+ *
+ * Used to describe buffer input in `source_info` objects.
+ */
+// TODO: to be replaced by `device_span`?
```
jrhemstad, line 179: This was precisely what I was going to ask. Why not just use `device_span<byte>`?

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +199,62 @@ class direct_read_source : public file_source {
...
+ * @brief Implementation class for reading from a device buffer source
+ */
+class device_buffer_source final : public datasource {
+ public:
+  explicit device_buffer_source(device_buffer const& d_buffer) : _d_buffer{d_buffer} {}
+
+  size_t host_read(size_t offset, size_t size, uint8_t* dst) override
+  {
+    auto const count = std::min(size, this->size() - offset);
+    cudaMemcpy(dst, _d_buffer._data + offset, count, cudaMemcpyDeviceToHost);
```
jrhemstad, line 212: Error checking

cpp/include/cudf/io/types.hpp
PointKernel: Use `device_span` instead.

cpp/include/cudf/io/types.hpp
```diff
@@ -156,8 +157,9 @@ struct table_with_metadata {
  *
  * Used to describe buffer input in `source_info` objects.
  */
+// TODO: to be replaced by `host_span`?
```
PointKernel, line 160: Probably a followup PR for this

vuule: yup. If we do go to `host_span<byte>` (instead of `host_span<byte>`, which is equivalent to `host_buffer`) it will require some wider changes.

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +200,67 @@ class direct_read_source : public file_source {
...
+class device_buffer_source final : public datasource {
+ public:
+  explicit device_buffer_source(cudf::device_span<std::byte const> d_buffer) : _d_buffer{d_buffer}
+  {
+  }
+
+  size_t host_read(size_t offset, size_t size, uint8_t* dst) override
+  {
+    auto const count = std::min(size, this->size() - offset);
+    CUDF_CUDA_TRY(cudaMemcpy(dst, _d_buffer.data() + offset, count, cudaMemcpyDeviceToHost));
```
vuule, line 215: Does this mean we need a stream parameter in these functions?

@jrhemstad is there a downside to not using `cudaMemcpyAsync` here?

vuule: Yup, understood. I'm trying to determine if we should add the stream parameter.

vuule:
Benchmark results:
|      io       | compression | cardinality | run_length | Samples | CPU Time | Noise | GPU Time | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|---------------|-------------|-------------|------------|---------|----------|-------|----------|-------|------------------|-------------------|-------------------|
|   HOST_BUFFER |      SNAPPY |           0 |          1 |      5x |  2.467 s | 0.01% |  2.467 s | 0.01% |        217595814 |         1.490 GiB |       493.950 MiB |
|   HOST_BUFFER |      SNAPPY |        1000 |          1 |      5x |  2.027 s | 0.30% |  2.027 s | 0.30% |        264806029 |         1.167 GiB |       161.238 MiB |
|   HOST_BUFFER |      SNAPPY |           0 |         32 |      5x |  1.616 s | 0.31% |  1.616 s | 0.31% |        332223805 |         1.056 GiB |        49.703 MiB |
|   HOST_BUFFER |      SNAPPY |        1000 |         32 |      5x |  1.608 s | 0.21% |  1.608 s | 0.21% |        333848972 |         1.031 GiB |        23.415 MiB |
|   HOST_BUFFER |        NONE |           0 |          1 |      5x |  2.341 s | 0.26% |  2.341 s | 0.26% |        229310026 |         1.490 GiB |       501.137 MiB |
|   HOST_BUFFER |        NONE |        1000 |          1 |      6x |  1.982 s | 0.46% |  1.982 s | 0.46% |        270856822 |         1.167 GiB |       169.774 MiB |
|   HOST_BUFFER |        NONE |           0 |         32 |      5x |  1.609 s | 0.31% |  1.609 s | 0.31% |        333655158 |         1.056 GiB |        56.544 MiB |
|   HOST_BUFFER |        NONE |        1000 |         32 |     10x |  1.572 s | 0.57% |  1.572 s | 0.57% |        341613126 |         1.031 GiB |        30.410 MiB |
| DEVICE_BUFFER |      SNAPPY |           0 |          1 |      5x |  1.855 s | 0.05% |  1.855 s | 0.05% |        289469282 |         1.490 GiB |       493.950 MiB |
| DEVICE_BUFFER |      SNAPPY |        1000 |          1 |      5x |  1.852 s | 0.32% |  1.852 s | 0.32% |        289851516 |         1.167 GiB |       161.238 MiB |
| DEVICE_BUFFER |      SNAPPY |           0 |         32 |      5x |  1.581 s | 0.31% |  1.581 s | 0.31% |        339647647 |         1.056 GiB |        49.703 MiB |
| DEVICE_BUFFER |      SNAPPY |        1000 |         32 |      5x |  1.606 s | 0.26% |  1.606 s | 0.26% |        334317660 |         1.031 GiB |        23.415 MiB |
| DEVICE_BUFFER |        NONE |           0 |          1 |      5x |  1.698 s | 0.03% |  1.698 s | 0.03% |        316169265 |         1.490 GiB |       501.137 MiB |
| DEVICE_BUFFER |        NONE |        1000 |          1 |      5x |  1.773 s | 0.24% |  1.773 s | 0.24% |        302837708 |         1.167 GiB |       169.774 MiB |
| DEVICE_BUFFER |        NONE |           0 |         32 |      5x |  1.548 s | 0.24% |  1.548 s | 0.24% |        346863247 |         1.056 GiB |        56.544 MiB |
| DEVICE_BUFFER |        NONE |        1000 |         32 |      5x |  1.551 s | 0.30% |  1.551 s | 0.30% |        346181357 |         1.031 GiB |        30.410 MiB |

cpp/src/io/utilities/datasource.cpp
PointKernel: I would vote for adding the stream parameter and using asynchronous APIs.

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +200,67 @@ class direct_read_source : public file_source {
...
+  {
+    return device_read_async(offset, size, dst, stream).get();
+  }
+
+  std::unique_ptr<buffer> device_read(size_t offset,
+                                      size_t size,
+                                      rmm::cuda_stream_view stream) override
+  {
+    return std::make_unique<non_owning_buffer>(
+      reinterpret_cast<uint8_t*>(const_cast<std::byte*>(_d_buffer.data() + offset)), size);
```
vuule, line 254: You shouldn't have to `const_cast` here, that's a bug in `non_owning_buffer` implementation. It should take `uint8_t const* data` and data members should be 
```
    uint8_t const*  _data{nullptr};
    size_t _size{0};
```
If would be good to fix this error in this PR instead of having a `const_cast` here.

cpp/tests/io/parquet_test.cpp
```diff
@@ -1105,12 +1105,32 @@ TEST_F(ParquetWriterTest, HostBuffer)
...
+    cudf::test::expect_metadata_equal(expected_metadata, result.metadata);
+  }
+
+  // device buffer
+  {
+    auto const d_input = cudf::detail::make_device_uvector_sync(
+      cudf::host_span<uint8_t const>{reinterpret_cast<uint8_t const*>(out_buffer.data()),
+                                     out_buffer.size()},
+      cudf::get_default_stream());
+    auto const d_buffer = cudf::device_span<std::byte const>(
```
vuule, line 1125: Might not be required
```suggestion
    auto const d_buffer = cudf::device_span(
```

vuule (changes requested):
Looks pretty good, some minor suggestions.
Thanks for updating the benchmarks as well!

cpp/benchmarks/io/cuio_common.cpp
```diff
@@ -37,15 +39,26 @@ std::string random_file_in_dir(std::string const& dir_path)
...
-    case io_type::HOST_BUFFER: return cudf::io::source_info(buffer.data(), buffer.size());
+    case io_type::HOST_BUFFER: return cudf::io::source_info(h_buffer.data(), h_buffer.size());
+    case io_type::DEVICE_BUFFER: {
+      auto const stream = cudf::get_default_stream();
+      d_buffer.resize(h_buffer.size(), stream);
+      CUDF_CUDA_TRY(cudaMemcpyAsync(
+        d_buffer.data(), h_buffer.data(), h_buffer.size(), cudaMemcpyDefault, stream.value()));
+
+      auto const d_source = cudf::device_span<std::byte>(d_buffer);
+      return cudf::io::source_info(d_source);
```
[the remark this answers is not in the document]
PointKernel: Good point. Fixed.

cpp/tests/io/parquet_test.cpp
PointKernel: Still needed due to the CTAD failure:
```bash
/home/yunsongw/dev/rapids/cudf/cpp/tests/io/parquet_test.cpp:1126:73: error: class template argument deduction failed:
 1126 |       reinterpret_cast<std::byte const*>(d_input.data()), d_input.size());
      |                                                                         ^

```

cpp/benchmarks/io/parquet/parquet_reader_input.cpp
```diff
@@ -126,8 +126,9 @@ using d_type_list = nvbench::enum_type_list<data_type::INTEGRAL,
                                             data_type::LIST,
                                             data_type::STRUCT>;
 
-using io_list =
-  nvbench::enum_type_list<cudf::io::io_type::FILEPATH, cudf::io::io_type::HOST_BUFFER>;
+using io_list = nvbench::enum_type_list<cudf::io::io_type::FILEPATH,
+                                        cudf::io::io_type::HOST_BUFFER,
+                                        cudf::io::io_type::DEVICE_BUFFER>;
```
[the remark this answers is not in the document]
PointKernel: Done.

`DEVICE_BUFFER` is about 10% to 3.5x faster in ORC reader benchmarks which is way higher than the parquet numbers. Is this expected?

cpp/benchmarks/io/cuio_common.cpp
```diff
@@ -55,7 +68,8 @@ cudf::io::sink_info cuio_source_sink_pair::make_sink_info()
   switch (type) {
     case io_type::VOID: return cudf::io::sink_info(&void_sink);
     case io_type::FILEPATH: return cudf::io::sink_info(file_name);
-    case io_type::HOST_BUFFER: return cudf::io::sink_info(&buffer);
+    case io_type::HOST_BUFFER: [[fallthrough]];
+    case io_type::DEVICE_BUFFER: return cudf::io::sink_info(&h_buffer);
```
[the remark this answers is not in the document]
PointKernel: Added a line in the doc: https://github.com/PointKernel/cudf/blob/cce529469fc62e35cda0c984877787015ecf0916/cpp/benchmarks/io/cuio_common.hpp#L72

cpp/benchmarks/io/parquet/parquet_reader_input.cpp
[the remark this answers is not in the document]
vuule: in which case(s) is it 3x faster?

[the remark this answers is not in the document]
PointKernel: 2 cases where the device buffer is about 3x faster:
|      io       | compression | cardinality | run_length | Samples |  CPU Time  | Noise |  GPU Time  | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|---------------|-------------|-------------|------------|---------|------------|-------|------------|-------|------------------|-------------------|-------------------|
|   HOST_BUFFER |      SNAPPY |           0 |          1 |      5x | 895.869 ms | 0.12% | 895.867 ms | 0.12% |        599275360 |         1.344 GiB |       482.814 MiB |
|   HOST_BUFFER |        NONE |           0 |          1 |      5x | 886.743 ms | 0.10% | 886.740 ms | 0.10% |        605443592 |         1.314 GiB |       484.294 MiB |
| DEVICE_BUFFER |      SNAPPY |           0 |          1 |      5x | 256.281 ms | 0.25% | 256.275 ms | 0.25% |       2094901991 |         1.344 GiB |       482.814 MiB |
| DEVICE_BUFFER |        NONE |           0 |          1 |      5x | 245.159 ms | 0.13% | 245.153 ms | 0.13% |       2189945956 |         1.314 GiB |       484.294 MiB |

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +200,67 @@ class direct_read_source : public file_source {
...
+
+  std::future<size_t> device_read_async(size_t offset,
+                                        size_t size,
+                                        uint8_t* dst,
+                                        rmm::cuda_stream_view stream) override
+  {
+    auto const count = std::min(size, this->size() - offset);
+    CUDF_CUDA_TRY(cudaMemcpyAsync(
+      dst, _d_buffer.data() + offset, count, cudaMemcpyDeviceToDevice, stream.value()));
+    return std::async(std::launch::async, [count] { return count; });
```
vuule, line 238: If I understand `std::async` correctly, we might be better off with `std::launch::deferred` here, to avoid spawning a new thread. Since the work in the `std::async` is trivial, we should be fine just deferring it to `future.get()` call time.
```suggestion
    return std::async(std::launch::deferred, [count] { return count; });
```

vuule (changes requested):
Looks good, just a few small suggestions.

cpp/benchmarks/io/parquet/parquet_reader_input.cpp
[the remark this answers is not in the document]
vuule: Discussed this offline: while the magnitude of performance improvement is unexpected, it does happen in cases where we expect the highest improvement (largest files). The results looks valid.
Thanks for running the benchmarks @PointKernel!

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +200,66 @@ class direct_read_source : public file_source {
...
+  {
+    return std::make_unique<non_owning_buffer>(
+      reinterpret_cast<uint8_t const*>(_d_buffer.data() + offset), size);
+  }
+
+  [[nodiscard]] size_t size() const override { return _d_buffer.size(); }
+
+ private:
+  cudf::device_span<std::byte const>
+    _d_buffer;  ///< A non-owning buffer to the existing device data
```
vuule, line 260:
```suggestion
    _d_buffer;  ///< A non-owning view of the existing device data
```

vuule (approved):
looks good. Just a small suggestion
Thanks for taking this up @PointKernel !

cpp/include/cudf/io/types.hpp
davidwendt: This comment's position may mess up doxygen generation. Perhaps move it to some other line in the code?

cpp/benchmarks/io/cuio_common.cpp
```diff
@@ -16,6 +16,8 @@
 
 #include <benchmarks/io/cuio_common.hpp>
 
+#include <cudf/detail/utilities/vector_factories.hpp>
```
davidwendt, line 19: Is this include needed?

PointKernel: Good catch. It should be a leftover.

Removed.

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +200,65 @@ class direct_read_source : public file_source {
...
+class device_buffer_source final : public datasource {
+ public:
+  explicit device_buffer_source(cudf::device_span<std::byte const> d_buffer) : _d_buffer{d_buffer}
+  {
+  }
+
+  size_t host_read(size_t offset, size_t size, uint8_t* dst) override
+  {
+    auto const count = std::min(size, this->size() - offset);
+    CUDF_CUDA_TRY(cudaMemcpy(dst, _d_buffer.data() + offset, count, cudaMemcpyDefault));
```
bdice, line 215: Should this use `cudaMemcpyAsync` with a stream? (Even if it must be synced immediately -- it might mean that it's only a stream sync instead of a device sync.)

See previous conversation with @vuule: https://github.com/rapidsai/cudf/pull/12314#discussion_r1060934999

> I default to `cudaMemcpyAsync` because it takes a stream, and we seem to be completely avoiding `cudaMemcpy` in libcudf (presumably because of the use of default stream?).

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +200,65 @@ class direct_read_source : public file_source {
...
+    return count;
+  }
+
+  std::unique_ptr<buffer> host_read(size_t offset, size_t size) override
+  {
+    auto const count = std::min(size, this->size() - offset);
+    std::vector<std::byte> h_data(count);
+    CUDF_CUDA_TRY(cudaMemcpy(h_data.data(), _d_buffer.data() + offset, count, cudaMemcpyDefault));
```
bdice, line 223: Same question here about `cudaMemcpyAsync` -- and similarly anywhere else in this PR.

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +200,65 @@ class direct_read_source : public file_source {
...
+  {
+    auto const count = std::min(size, this->size() - offset);
+    CUDF_CUDA_TRY(cudaMemcpy(dst, _d_buffer.data() + offset, count, cudaMemcpyDefault));
+    return count;
+  }
+
+  std::unique_ptr<buffer> host_read(size_t offset, size_t size) override
+  {
+    auto const count = std::min(size, this->size() - offset);
+    std::vector<std::byte> h_data(count);
```
bdice, line 222: Should this use a [vector factory](https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/include/cudf/detail/utilities/vector_factories.hpp)?

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +200,65 @@ class direct_read_source : public file_source {
...
+
+  std::future<size_t> device_read_async(size_t offset,
+                                        size_t size,
+                                        uint8_t* dst,
+                                        rmm::cuda_stream_view stream) override
+  {
+    auto const count = std::min(size, this->size() - offset);
+    CUDF_CUDA_TRY(
+      cudaMemcpyAsync(dst, _d_buffer.data() + offset, count, cudaMemcpyDefault, stream.value()));
+    return std::async(std::launch::deferred, [count] { return count; });
```
bdice, line 237: How do we know this future is stream-synced when calling `.get()`? If that doesn't matter, then why use a future?

cpp/src/io/utilities/datasource.cpp
PointKernel: > Should this use cudaMemcpyAsync with a stream? 

Yes, this has been discussed in https://github.com/rapidsai/cudf/pull/12024#discussion_r1066192873. Adding the stream parameter requires changes to all classes derived from the base `datasource` thus would be done in a separate PR.

bdice: I'm suggesting something different that doesn't require a new parameter. Replace:
```cpp
CUDF_CUDA_TRY(cudaMemcpy(...));
```
with
```cpp
auto const stream = cudf::get_default_stream();
CUDF_CUDA_TRY(cudaMemcpyAsync(..., stream);
stream.synchronize();
```

From my understanding, `cudaMemcpy` syncs the whole device and not just the stream that libcudf is using. If that's incorrect, please let me know!

PointKernel: Got it. Make sense to me.

cpp/benchmarks/io/cuio_common.cpp
```diff
@@ -37,15 +37,25 @@ std::string random_file_in_dir(std::string const& dir_path)
...
 }
 
 cudf::io::source_info cuio_source_sink_pair::make_source_info()
 {
   switch (type) {
     case io_type::FILEPATH: return cudf::io::source_info(file_name);
-    case io_type::HOST_BUFFER: return cudf::io::source_info(buffer.data(), buffer.size());
+    case io_type::HOST_BUFFER: return cudf::io::source_info(h_buffer.data(), h_buffer.size());
+    case io_type::DEVICE_BUFFER: {
+      auto const stream = cudf::get_default_stream();
```
ttnghia, line 52: Why are we going to use the default stream? Is there any input stream stored?

ttnghia: Because we're returning `source_info` constructed from `d_buffer`. If the default stream is used to copy data to `d_buffer` but some other stream is used to access it then ......

cpp/benchmarks/io/cuio_common.hpp
```diff
@@ -81,7 +81,8 @@ class cuio_source_sink_pair {
   static temp_directory const tmpdir;
 
   io_type const type;
-  std::vector<char> buffer;
+  std::vector<char> h_buffer;
+  rmm::device_uvector<std::byte> d_buffer;
```
ttnghia, line 85: Why they are using different data types (`char` vs `std::byte`)? Can we use the same?

cpp/include/cudf/io/datasource.hpp
```diff
@@ -318,8 +327,8 @@ class datasource {
     [[nodiscard]] uint8_t const* data() const override { return _data; }
 
    private:
-    uint8_t* const _data{nullptr};
-    size_t const _size{0};
+    uint8_t const* _data{nullptr};
+    size_t _size{0};
```
ttnghia, line 331: Why removing `const` for `_size`?

cpp/src/io/utilities/datasource.cpp
```diff
@@ -199,6 +200,65 @@ class direct_read_source : public file_source {
...
+    std::vector<std::byte> h_data(count);
+    CUDF_CUDA_TRY(cudaMemcpy(h_data.data(), _d_buffer.data() + offset, count, cudaMemcpyDefault));
+    return std::make_unique<owning_buffer<std::vector<std::byte>>>(std::move(h_data));
+  }
+
+  [[nodiscard]] bool supports_device_read() const override { return true; }
+
+  std::future<size_t> device_read_async(size_t offset,
+                                        size_t size,
+                                        uint8_t* dst,
```
ttnghia, line 231: I see that we are using mixed `uint8_t`, `char` and `std::byte`. Can we just use one standard type like `std::byte` or `char` (out-of-scope of this PR)?

cpp/src/io/utilities/datasource.cpp
PointKernel: The naming is a bit misleading. The "async" here refers to the fact that device reads are launched asynchronously from the host side. Other data sources seem to launch multiple device reads (operations more comlplex than just invoking `cudaMemcpy`) and then sync to get the final data. For the case of device buffer, this operation is one single invocation of `cudaMemcpy` so we don't really need the `std::future`. Returning a future is not necessary in this case and we are doing so simply because the base virtual function we are overriding is defined in this way.

cpp/benchmarks/io/cuio_common.cpp
bdice: This is a case that will need to be fixed in order to make libcudf a fully stream friendly library. However, it is fine for now to make this use the  cudf default stream.

cpp/src/io/utilities/datasource.cpp
PointKernel: Agreed. We should unify all byte types by using `std::byte` consistently. cc @vuule

ttnghia (approved):
I'm not completely satisfied with the current code (such as using the default stream), but understand that we cannot fix everything at once thus I'm fine with this.

cpp/src/io/utilities/datasource.cpp
vuule: yes. we should be using std::byte for "raw data" buffers (https://github.com/rapidsai/cudf/issues/6029#issuecomment-1302662162).
It's a big task and I haven't decided on how best to approach it.
If there's interest, I can open an issue and we can discuss there.

cpp/src/io/utilities/datasource.cpp
vuule: Because GDS did not have an async API at the time of our integration (they do now, but it's not stream-ordered), we have odd semantics of the async reads. Async here means "once the future returns, data is available in the given stream". Sometimes the future part does not matter (like here), and sometimes the stream part might not matter. But calling read_async(..., stream), then get() on the returned future, then stream.synchronize() on the passed stream should always mean that the data is at the destination from that point forward.

vuule:
Opened https://github.com/rapidsai/cudf/issues/12576 to replace host_buffer.

cpp/src/io/utilities/datasource.cpp
karthikeyann: +1  for using `std::byte`

Issue #11780, opened 2022-09-27
[FEA] Remove `**kwargs` in all IO readers and writers


galipremsagar:
**Is your feature request related to a problem? Please describe.**
We will need to remove usage of `**kwargs` in all IO readers and writers, having them is leading to problems where we are silently skipping the incorrectly named parameters that users pass. We have removed the usage of `kwargs` in CSV readers and writers in https://github.com/rapidsai/cudf/pull/11762.

**Describe the solution you'd like**
Do the same for 
- [ ] `parquet`
- [ ] `orc`
- [ ] `json`

bdice:
@galipremsagar I started work on this a while back. See here for the branch: https://github.com/rapidsai/cudf/compare/branch-22.10...bdice:cudf:io-kwargs

galipremsagar:
> @galipremsagar I started work on this a while back. See here for the branch: [branch-22.10...bdice:cudf:io-kwargs](https://github.com/rapidsai/cudf/compare/branch-22.10...bdice:cudf:io-kwargs)

I'll build on top of this

Issue #6916, opened 2020-12-04
[FEA] Expose `storage_options` kwarg to readers/writers


ayushdg:
**Is your feature request related to a problem? Please describe.**
cuDF supports reading/writing to multiple data locations aided by `fsspec`. Reading data from remote source like cloud storage often requires additional kwargs passed to fsspec via the `storage_options` argument which is not exposed directly in cuDF but is a hidden kwarg. 

Since v1.2, pandas has transitioned to using fsspec for a lot of file handling operations and now exposes the `storage_options` argument.

**Describe the solution you'd like**
cuDF can therefore expose this argument rather than using it as a hidden kwarg.

**Describe alternatives you've considered**
N/A

**Additional context**
Here are a couple of examples for pandas exposing storage_options from their docs: [read_csv](https://pandas.pydata.org/docs/dev/reference/api/pandas.read_csv.html#pandas.read_csv) , [to_parquet](https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.to_parquet.html#pandas.DataFrame.to_parquet).

ayushdg:
Can tackle once pandas requirements are updated (being worked on in #7375).

vuule:
The prerequisite PR has been merged. 
@ayushdg, should I move this to 0.20?

ayushdg:
Thanks for the ping @vuule. Pushing to 0.20 seems fine. Shouldn't impact workflows, since we already support it when passed via `kwargs`, but don't explicitly have it as an argument.

vuule:
Moving to 21.12 since the linked PR is still a draft.

chinmaychandak:
This will be really useful! What is the current status for this?

ayushdg:
@chinmaychandak passing storage_options is already supported today. It just isn't exposed explicitly as a param, which this issue is meant to address. 
Is storage_options not working for you for certain Io methods?

chinmaychandak:
No, I meant exposing this as an explicit param would be really useful. Not a blocker by any means.

PR #12025, opened 2022-10-28
[REVIEW] Reduce/Remove reliance on `**kwargs` and `*args` in `IO` readers & writers

galipremsagar:
## Description
Resolves: #11780 

This PR:

- [x] Reduces reliance on `args` & `kwargs` for readers and writers when `cudf` engine is selected. However, these will have to stay for the purpose of other engines we support in few readers & writers such as `pandas` & `pyarrow` engines.
- [x] Fixes some bugs where dead parameters were still being used.
- [x] Fixes some bugs where parameters weren't being passed until the cython later in the first place.
- [x] Updates docs related to newly exposed parameters.

## Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -6019,11 +6019,51 @@ def select_dtypes(self, include=None, exclude=None):
...
+        metadata_file_path=None,
+        int96_timestamps=False,
+        row_group_size_bytes=None,
+        row_group_size_rows=None,
+        max_page_size_bytes=None,
+        max_page_size_rows=None,
+        storage_options=None,
+        return_metadata=False,
+        *args,
+        **kwargs,
```
bdice, line 6041: Can you give a little more context on what arguments might fall into this category? You mentioned it might vary by engine? Do we need both `*args` and `**kwargs` here?

galipremsagar: Had to keep those because other engines by themselves support params like `kwargs`: https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_to_dataset.html
`args` is likely due to dask upstream API passing it.

python/cudf/cudf/io/parquet.py
```diff
@@ -666,7 +739,15 @@ def to_parquet(
             row_group_size_rows=row_group_size_rows,
             max_page_size_bytes=max_page_size_bytes,
             max_page_size_rows=max_page_size_rows,
-            **kwargs,
+            partitions_info=list(
+                zip(
+                    partition_offsets,
+                    np.roll(partition_offsets, -1) - partition_offsets,
```
bdice, line 745: Try:
```python
list(zip(partition_offsets, np.diff(partition_offsets)))
```

Example:
```python
import numpy as np
offsets = [0, 10, 15, 25, 37]
print(list(zip(offsets, np.roll(offsets, -1) - offsets))[:-1])
print(list(zip(offsets, np.diff(offsets))))
```
Both lines print:
```
[(0, 10), (10, 5), (15, 10), (25, 12)]
```

python/cudf/cudf/io/parquet.py
```diff
@@ -666,7 +739,15 @@ def to_parquet(
...
+                )
+            )[:-1]
+            if partition_offsets is not None
+            else partition_offsets,
```
bdice, line 749: Being explicit in the case of `partitions_offsets is None` might help here.
```suggestion
            else None,
```

python/cudf/cudf/io/parquet.py
```diff
@@ -872,6 +955,13 @@ class ParquetDatasetWriter:
     file_name_prefix : str
         This is a prefix to file names generated only when
         `max_file_size` is specified.
+    storage_options : dict, optional, default None
+        Extra options that make sense for a particular storage connection,
+        e.g. host, port, username, password, etc. For HTTP(S) URLs the
+        key-value pairs are forwarded to ``urllib.request.Request`` as
+        header options. For other URLs (e.g. starting with “s3://”, and
+        “gcs://”) the key-value pairs are forwarded to ``fsspec.open``.
```
bdice, line 963: Careful when copying from rendered docs online. Sphinx will generate smart quotes from regular quotes by default, which shows up on the online docs, but we don't want to use smart quotes in our source code.
```suggestion
        header options. For other URLs (e.g. starting with "s3://", and
        "gcs://") the key-value pairs are forwarded to ``fsspec.open``.
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -43,12 +43,20 @@
...
     If not None, only these columns will be read.
 skiprows : int, default None
     If not None, the number of rows to skip from the start of the file.
 num_rows : int, default None
     If not None, the total number of rows to read.
+storage_options : dict, optional, default None
+    Extra options that make sense for a particular storage connection,
+    e.g. host, port, username, password, etc. For HTTP(S) URLs the key-value
+    pairs are forwarded to ``urllib.request.Request`` as header options.
+    For other URLs (e.g. starting with “s3://”, and “gcs://”) the key-value
```
bdice, line 57:
```suggestion
    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -132,6 +140,13 @@
     Parser engine to use.
 columns : list, default None
     If not None, only these columns will be read.
+storage_options : dict, optional, default None
+    Extra options that make sense for a particular storage connection,
+    e.g. host, port, username, password, etc. For HTTP(S) URLs the key-value
+    pairs are forwarded to ``urllib.request.Request`` as header options.
+    For other URLs (e.g. starting with “s3://”, and “gcs://”) the key-value
```
bdice, line 147:
```suggestion
    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -253,9 +275,21 @@
 max_page_size_rows: integer or None, default None
     Maximum number of rows of each page of the output.
     If None, 20000 will be used.
-**kwargs
+storage_options : dict, optional, default None
+    Extra options that make sense for a particular storage connection,
+    e.g. host, port, username, password, etc. For HTTP(S) URLs the key-value
+    pairs are forwarded to ``urllib.request.Request`` as header options.
+    For other URLs (e.g. starting with “s3://”, and “gcs://”) the key-value
```
bdice, line 282:
```suggestion
    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -253,9 +275,21 @@
...
+    pairs are forwarded to ``fsspec.open``. Please see ``fsspec`` and
+    ``urllib`` for more details.
+return_metadata : bool, default False
+    Return parquet metadata for written data. Returned metadata will
+    include the file-path metadata (relative to `root_path`).
```
bdice, line 287: Not sure if this is copied from external docs but I'd remove the hyphen in "file path."
```suggestion
    include the file path metadata (relative to `root_path`).
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -392,7 +426,20 @@
     If True, Arrow-backed PythonFile objects will be used in place of fsspec
     AbstractBufferedFile objects at IO time. This option is likely to improve
     performance when making small reads from larger ORC files.
-kwargs are passed to the engine
+storage_options : dict, optional, default None
+    Extra options that make sense for a particular storage connection,
+    e.g. host, port, username, password, etc. For HTTP(S) URLs the key-value
+    pairs are forwarded to ``urllib.request.Request`` as header options.
+    For other URLs (e.g. starting with “s3://”, and “gcs://”) the key-value
```
bdice, line 433:
```suggestion
    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -444,6 +492,20 @@
     A list of column names which should be written as map type in the ORC file.
     Note that this option only affects columns of ListDtype. Names of other
     column types will be ignored.
+storage_options : dict, optional, default None
+    Extra options that make sense for a particular storage connection,
+    e.g. host, port, username, password, etc. For HTTP(S) URLs the key-value
+    pairs are forwarded to ``urllib.request.Request`` as header options.
+    For other URLs (e.g. starting with “s3://”, and “gcs://”) the key-value
```
bdice, line 499:
```suggestion
    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -613,6 +675,13 @@
     If `True`, any string values are read literally (and wrapped in an
     additional set of quotes).
     If `False` string values are parsed into Python strings.
+storage_options : dict, optional, default None
+    Extra options that make sense for a particular storage connection,
+    e.g. host, port, username, password, etc. For HTTP(S) URLs the key-value
+    pairs are forwarded to ``urllib.request.Request`` as header options.
+    For other URLs (e.g. starting with “s3://”, and “gcs://”) the key-value
```
bdice, line 682:
```suggestion
    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -1238,6 +1307,13 @@
     compressed file (upper 48 bits).
     The start offset points to the first byte to be read, the end offset points
     one past the last byte to be read.
+storage_options : dict, optional, default None
+    Extra options that make sense for a particular storage connection,
+    e.g. host, port, username, password, etc. For HTTP(S) URLs the key-value
+    pairs are forwarded to ``urllib.request.Request`` as header options.
+    For other URLs (e.g. starting with “s3://”, and “gcs://”) the key-value
```
bdice, line 1314: I keep thinking I'm done fixing smart quotes... I won't add any more suggestions for this. Please apply to any places I missed. 😅 
```suggestion
    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -1511,6 +1582,22 @@ def get_reader_filepath_or_buffer(
...
+        For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
+        pairs are forwarded to ``fsspec.open``. Please see ``fsspec`` and
+        ``urllib`` for more details, and for more examples on storage options
+        refer `here <https://pandas.pydata.org/docs/user_guide/io.html?
+        highlight=storage_options#reading-writing-remote-files>`__.
+    bytes_per_thread : int, default None
+        Determines the number of bytes to be allocated per thread to read the
+        files in parallel. When there is a file of large size, we get slightly
+        better throughput by decomposing it and transferring multiple "blocks"
+        in parallel (using a python thread pool). Default allocation is
```
bdice, line 1598: Capitalize Python (applies throughout).
```suggestion
        in parallel (using a Python thread pool). Default allocation is
```

python/dask_cudf/dask_cudf/io/parquet.py
```diff
@@ -292,24 +292,45 @@ def write_partition(
...
+                df=df,
+                root_path=path,
+                compression=compression,
                 filename=filename,
                 partition_cols=partition_on,
                 fs=fs,
                 preserve_index=preserve_index,
                 return_metadata=return_metadata,
-                **kwargs,
+                statistics=kwargs.get("statistics", "ROWGROUP"),
```
bdice, line 303: If these defaults match those of the `write_to_dataset` function, it's perfectly fine to pass `**kwargs` through here rather than re-define the defaults. We just want to guard against unrecognized/unsupported names in `**kwargs` making their way all the way down the call stack and being silently ignored (without warnings/errors) if the user expects them to do something.

Alternatively you can use `kwargs.pop("key", "default")` to extract these values before calling `write_to_dataset` and assert that all the `**kwargs` were used -- it should be empty after popping all the known keys. 😉

bdice (changes requested):
Great work on this PR, @galipremsagar. I recognize these changes are difficult! I started down this route earlier before realizing I didn't know enough about the meaning of all the arguments to make good decisions on how to handle them. Thank you for this!

python/cudf/cudf/io/avro.py
```diff
@@ -11,26 +14,33 @@ def read_avro(
...
-        path_or_data=filepath_or_buffer, compression=None, **kwargs
+        path_or_data=filepath_or_buffer,
+        compression=None,
+        storage_options=storage_options,
     )
     if compression is not None:
         ValueError("URL content-encoding decompression is not supported")
 
     if engine == "cudf":
+        warnings.warn(
```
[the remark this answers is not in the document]
galipremsagar: Oops, forgot to add it. Thanks for adding it !

python/cudf/cudf/io/parquet.py
galipremsagar: Yikes

python/cudf/cudf/io/orc.py
```diff
@@ -339,7 +342,10 @@ def read_orc(
             path_or_data=source,
             compression=None,
             use_python_file_object=use_python_file_object,
-            **kwargs,
+            storage_options=storage_options,
+            bytes_per_thread=256_000_000
+            if bytes_per_thread is None
+            else bytes_per_thread,
```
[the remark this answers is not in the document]
galipremsagar: > if bytes_per_thread is None:
    bytes_per_thread = 256_000_000

done 👍 



> Alternatively, should we specify `bytes_per_thread=256_000_000` in the function's default parameters?

I don't think so, because `bytes_per_thread` can only be passed if `use_python_file_object=False`, we have an active check for this something like:
```python
if use_python_file_object and bytes_per_thread is not None:
        raise ValueError(
            "bytes_per_thread is only supported when "
            "`use_python_file_object=False`"
        )
```

python/cudf/cudf/io/csv.py
```diff
@@ -61,6 +61,9 @@ def read_csv(
             "`use_python_file_object=False`"
         )
 
+    if bytes_per_thread is None:
+        bytes_per_thread = 256_000_000
```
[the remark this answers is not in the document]
wence-: Aside: if this is meant to be 256MB, do we want  `256 * 1024 * 1024` here?

python/cudf/cudf/io/parquet.py
```diff
@@ -136,25 +145,51 @@ def write_to_dataset(
...
     return_metadata : bool, default False
         Return parquet metadata for written data. Returned metadata will
         include the file-path metadata (relative to `root_path`).
-    **kwargs : dict,
-        kwargs for to_parquet function.
+    int96_timestamps : bool, default False
+        If ``True``, write timestamps in int96 format. This will convert
+        timestamps from timestamp[ns], timestamp[ms], timestamp[s], and
+        timestamp[us] to the int96 format, which is the number of Julian
+        days and the number of nanoseconds since midnight. If ``False``,
```
wence-, line 168: Number of Julian days since when?

python/cudf/cudf/io/parquet.py
```diff
@@ -166,31 +201,50 @@ def write_to_dataset(
...
-            grouped_df,
-            full_paths,
+            df=grouped_df,
+            path=full_paths,
+            compression=compression,
             index=preserve_index,
             partition_offsets=part_offsets,
-            **kwargs,
+            storage_options=storage_options,
+            metadata_file_path=metadata_file_paths
```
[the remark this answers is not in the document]
galipremsagar: Dpme

python/cudf/cudf/io/parquet.py
```diff
@@ -431,19 +489,22 @@ def read_parquet(
...
+    for source in filepath_or_buffer:
         tmp_source, compression = ioutils.get_reader_filepath_or_buffer(
             path_or_data=source,
             compression=None,
             fs=fs,
             use_python_file_object=use_python_file_object,
             open_file_options=open_file_options,
-            **kwargs,
+            storage_options=storage_options,
+            bytes_per_thread=256_000_000
```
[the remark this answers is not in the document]
galipremsagar: Done, we can actually do `bytes_per_thread=_PARQUET_BYTES_PER_THREAD_DEFAULT` and docs will render as `bytes_per_thread= 256000000`

python/cudf/cudf/io/parquet.py
wence-: Or:
```python
[(i, j - i) for i, j in zip(offsets, offsets[1:])]
```
If you don't want to invoke numpy

python/cudf/cudf/utils/ioutils.py
```diff
@@ -43,12 +43,20 @@
     (such as builtin `open()` file handler function or `BytesIO`).
 engine : ['cudf'], default 'cudf'
     Parser engine to use.
+    This Parameter is deprecated.
```
wence-, line 49: Do we already have a deprecation FutureWarning in place for this?
```suggestion
    This parameter is deprecated.
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -392,7 +426,20 @@
...
+    pairs are forwarded to ``urllib.request.Request`` as header options.
+    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
+    pairs are forwarded to ``fsspec.open``. Please see ``fsspec`` and
+    ``urllib`` for more details.
+bytes_per_thread : int, default None
+    Determines the number of bytes to be allocated per thread to read the
+    files in parallel. When there is a file of large size, we get slightly
+    better throughput by decomposing it and transferring multiple "blocks"
+    in parallel (using a python thread pool). Default allocation is
+    256_000_000 bytes.
```
wence-: Do we want power of two bytes as default?

python/cudf/cudf/utils/ioutils.py
```diff
@@ -444,6 +492,20 @@
...
+    pairs are forwarded to ``urllib.request.Request`` as header options.
+    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
+    pairs are forwarded to ``fsspec.open``. Please see ``fsspec`` and
+    ``urllib`` for more details.
+index : bool, default None
+    If ``True``, include the dataframe's index(es) in the file output. If
+    ``False``, they will not be written to the file. If ``None``, the
+    engine's default behavior will be used. However, instead of being saved
+    as values, the ``RangeIndex`` will be stored as a range in the metadata
+    so it doesn’t require much space and is faster. Other indexes will
```
wence-, line 514: It's somewhat unclear under what circumstances `RangeIndex` is stored in compact form. I think:

- `True` => `RangeIndex` expanded, all indices written
- `False` => No indices written
- `None` => `RangeIndex` compact, all indices written

But this wording leaves open the idea that the engine default might be to not write any indices. Is this intentional?

python/cudf/cudf/io/parquet.py
galipremsagar: Works 👍

python/cudf/cudf/_lib/orc.pyx
```diff
@@ -129,15 +129,15 @@ cpdef read_orc(object filepaths_or_buffers,
...
+    actual_index_names, col_names, is_range_index, reset_index_name, \
+        range_idx = _get_index_from_metadata(c_result.metadata.user_data,
+                                             names,
+                                             skip_rows,
+                                             num_rows)
 
     data, index = data_from_unique_ptr(
         move(c_result.tbl),
-        names,
+        col_names if columns is None else names,
```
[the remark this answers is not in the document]
galipremsagar: Exactly, added coverage for this in `test_orc_columns_and_index_param`

python/cudf/cudf/io/avro.py
```diff
@@ -11,26 +14,33 @@ def read_avro(
...
+            "`engine` parameter is deprecated and will be removed in a "
```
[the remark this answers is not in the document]
galipremsagar: For updating their code, The message should be self-explanatory no? i.e., drop `engine` from their function calls..

python/cudf/cudf/io/json.py
```diff
@@ -117,6 +137,8 @@ def read_json(
                 lines=lines,
                 dtype=dtype,
                 compression=compression,
+                storage_options=storage_options,
+                orient=orient,
                 *args,
                 **kwargs,
             )
```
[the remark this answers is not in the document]
galipremsagar: Done

python/cudf/cudf/_lib/orc.pyx
```diff
@@ -247,7 +247,8 @@ cpdef write_orc(table,
                 object stripe_size_bytes=None,
                 object stripe_size_rows=None,
                 object row_index_stride=None,
-                object cols_as_map_type=None):
+                object cols_as_map_type=None,
+                object index=None):
     """
     Cython function to call into libcudf API, see `write_orc`.
```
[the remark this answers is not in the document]
galipremsagar: Updated the docstring to correctly point to `cudf::io::write_orc`

python/cudf/cudf/utils/ioutils.py
galipremsagar: Fixed it

python/cudf/cudf/utils/ioutils.py
galipremsagar: > I think:
> 
> * `True` => `RangeIndex` expanded, all indices written
> * `False` => No indices written
> * `None` => `RangeIndex` compact, all indices written

This is correct and that is the intended behavior. However, the docstring has been copied from [pandas](https://pandas.pydata.org/pandas-docs/version/1.1/reference/api/pandas.DataFrame.to_parquet.html), let me know if you have a suggestion to improve the wording here we could definitely do that here.

python/cudf/cudf/io/csv.py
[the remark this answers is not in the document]
galipremsagar: Done 👍

python/cudf/cudf/io/json.py
```diff
@@ -17,22 +18,23 @@
...
-        warnings.warn(
-            "passing 'dtype' as list is deprecated, instead pass "
-            "a dict of column name and types key-value paris."
-            "in future versions 'dtype' can only be a dict or bool",
-            FutureWarning,
+    if dtype is not None and not isinstance(dtype, (abc.Mapping, bool)):
+        raise TypeError(
+            "'dtype' parameter only support "
+            "a dict of column name and types key-value paris, "
+            f"or a bool or None. Got {type(dtype)}"
```
bdice, line 37:
```suggestion
            "'dtype' parameter only supports "
            "a dict of column names and types as key-value pairs, "
            f"or a bool, or None. Got {type(dtype)}"
```

python/cudf/cudf/io/parquet.py
bdice: I think one instance was fixed, but not all of the cases. c96c5ab893e956d8b7f975c6940b7d7d3abeb5a8 missed this one.

python/cudf/cudf/io/parquet.py
```diff
@@ -166,31 +201,48 @@ def write_to_dataset(
...
-        if return_metadata:
-            kwargs["metadata_file_path"] = filename
-        metadata = df.to_parquet(full_path, index=preserve_index, **kwargs)
+
+        metadata = df.to_parquet(
+            path=full_path,
+            compression=compression,
+            index=preserve_index,
+            storage_options=storage_options,
+            metadata_file_path=filename if return_metadata else None,
```
bdice, line 238: I would prefer to write `metadata_file_path = filename if return_metadata else None` above and then use it here.

python/cudf/cudf/io/parquet.py
```diff
@@ -600,16 +670,28 @@ def to_parquet(
...
     *args,
     **kwargs,
 ):
     """{docstring}"""
 
     if engine == "cudf":
+        if kwargs:
+            raise ValueError(
+                "cudf engine doesn't support the "
+                f"following keyword arguments: {list(kwargs.keys())}"
```
bdice, line 688: This error message was changed to `keyword arguments` but all the other instances still say `parameters`.

python/cudf/cudf/io/parquet.py
```diff
@@ -666,7 +738,12 @@ def to_parquet(
             row_group_size_rows=row_group_size_rows,
             max_page_size_bytes=max_page_size_bytes,
             max_page_size_rows=max_page_size_rows,
-            **kwargs,
+            partitions_info=list(
+                zip(partition_offsets, np.diff(partition_offsets))
```
bdice, line 742: Can we revisit this and apply @wence-'s suggestion:

https://github.com/rapidsai/cudf/pull/12025#discussion_r1009484064

There are two places in this file where NumPy is being used, and these can both be replaced by `[(i, j - i) for i, j in zip(offsets, offsets[1:])]` to eliminate the dependence on `import numpy as np`.

python/cudf/cudf/utils/ioutils.py
```diff
@@ -1043,15 +1120,15 @@
...
+    For other URLs (e.g. starting with "s3://", and "gcs://") the key-value
     pairs are forwarded to ``fsspec.open``. Please see ``fsspec`` and
     ``urllib`` for more details.
 bytes_per_thread : int, default None
     Determines the number of bytes to be allocated per thread to read the
     files in parallel. When there is a file of large size, we get slightly
     better throughput by decomposing it and transferring multiple "blocks"
     in parallel (using a python thread pool). Default allocation is
-    256_000_000 bytes.
+    {bytes_per_thread_bytes} bytes.
```
bdice, line 1131: The unit is already clear from the variable name. To avoid redundancy let's call this `bytes_per_thread`. (Applies in several places.)
```suggestion
    {bytes_per_thread} bytes.
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -1511,6 +1591,22 @@ def get_reader_filepath_or_buffer(
...
+        pairs are forwarded to ``fsspec.open``. Please see ``fsspec`` and
+        ``urllib`` for more details, and for more examples on storage options
+        refer `here <https://pandas.pydata.org/docs/user_guide/io.html?
+        highlight=storage_options#reading-writing-remote-files>`__.
+    bytes_per_thread : int, default None
+        Determines the number of bytes to be allocated per thread to read the
+        files in parallel. When there is a file of large size, we get slightly
+        better throughput by decomposing it and transferring multiple "blocks"
+        in parallel (using a Python thread pool). Default allocation is
+        256_000_000 bytes.
```
bdice, line 1608: Is this supposed to be `_BYTES_PER_THREAD_DEFAULT`, injected by a format operation? (The value is different.)

bdice (changes requested):
Additional cleanup from previous round of edits. This is looking a lot better!

python/cudf/cudf/utils/ioutils.py
wence-: The pandas version is:

> If True, include the dataframe’s index(es) in the file output. If False, they will not be written to the file. _If None, similar to True the dataframe’s index(es) will be saved._ However, instead of being saved as values, the RangeIndex will be stored as a range in the metadata so it doesn’t require much space and is faster. Other indexes will be included as columns in the file output.

So no talk of "the engine's default". Which I think would clarify things. I would also write:

", however, instead of being saved as values, _any_ RangeIndex will be stored ..."

Two changes: "however" follows on from previous sentence, "any" rather than "the" `RangeIndex`

wence-: Concretely:

```
     If ``True``, include the dataframe's index(es) in the file output. If
     ``False``, they will not be written to the file. If ``None``, similar to ``True`` 
     the dataframe's index(es) will be saved, however, instead of being saved
     as values any ``RangeIndex`` will be stored as a range in the metadata
     so it doesn’t require much space and is faster.
```

python/cudf/cudf/io/parquet.py
galipremsagar: Apologies @wence- I couldn't see this comment earlier. Now it should be resolved fully.

python/cudf/cudf/io/parquet.py
galipremsagar: Fixed.

python/cudf/cudf/utils/ioutils.py
galipremsagar: Done

galipremsagar:
@bdice Should we ready for another round of review.

ayushdg:
This pr also fixes #6916

python/cudf/cudf/utils/ioutils.py
galipremsagar: Sounds good to me 👍 Updated the docstring.

python/cudf/cudf/_lib/orc.pyx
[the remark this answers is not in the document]
wence-: Thanks.

python/cudf/cudf/_lib/parquet.pyx
```diff
@@ -312,7 +312,7 @@ cpdef write_parquet(
         object statistics="ROWGROUP",
         object metadata_file_path=None,
         object int96_timestamps=False,
-        object row_group_size_bytes=None,
+        object row_group_size_bytes=134217728,
```
wence-, line 315: ioutils._ROW_GROUP_SIZES_BYTES_DEFAULT?

galipremsagar: Changed it 👍

wence- (approved):
Two very minor niggles, but otherwise looks great, thanks for this!

python/cudf/cudf/utils/ioutils.py
```diff
@@ -208,15 +234,15 @@
...
-    engine's default behavior will be used. However, instead of being saved
-    as values, the ``RangeIndex`` will be stored as a range in the metadata
-    so it doesn’t require much space and is faster. Other indexes will
-    be included as columns in the file output.
+    If ``True``, include the dataframe's index(es) in the file output.
+    If ``False``, they will not be written to the file.
+    If ``None``, similar to ``True`` the dataframe's index(es) will
+    be saved, however, instead of being saved as values any
+    ``RangeIndex`` will be stored as a range in the metadata so it
+    doesn’t require much space and is faster.
```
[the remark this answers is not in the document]
galipremsagar: No worries, we are good. Updated it 👍

python/dask_cudf/dask_cudf/io/parquet.py
galipremsagar: So the thing here is, the method signature is not guaranteed to match to that of `write_to_dataset`, because the upstream API signature is:
```
def write_partition(cls, df, path, fs, filename, **kwargs):
```

To perform validation will be a blocker here because dask sometimes sends a few parameters like `custom_metadata`, etc which aren't relevant to us. Hence doing this process of filtration at this level.

bdice: Sounds like you have more context here -- I'll defer to your judgment.

python/cudf/cudf/io/parquet.py
```diff
@@ -136,25 +144,51 @@ def write_to_dataset(
...
     root_path : string,
         The root directory of the dataset
+    compression : {'snappy', 'ZSTD', None}, default 'snappy'
+        Name of the compression to use. Use ``None`` for no compression.
     filename : string, default None
         The file name to use (within each partition directory). If None,
         a random uuid4 hex string will be used for each file name.
+    partition_cols : list,
+        Column names by which to partition the dataset
+        Columns are partitioned in the order they are given
```
bdice, line 154:
```suggestion
        Column names by which to partition the dataset.
        Columns are partitioned in the order they are given.
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -25,6 +25,9 @@
     fsspec_parquet = None
 
 
+_BYTES_PER_THREAD_DEFAULT = 256 * 1024 * 1024
+_ROW_GROUP_SIZE_BYTES_DEFAULT = 134217728
```
bdice, line 29:
```suggestion
_ROW_GROUP_SIZE_BYTES_DEFAULT = 128 * 1024 * 1024
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -208,14 +234,15 @@
...
-    ``False``, they will not be written to the file. If ``None``, the
-    engine's default behavior will be used. However, instead of being saved
-    as values, the ``RangeIndex`` will be stored as a range in the metadata
-    so it doesn’t require much space and is faster. Other indexes will
+    If ``True``, include the dataframe's index(es) in the file output.
+    If ``False``, they will not be written to the file.
+    If ``None``, similar to ``True`` the dataframe's index(es) will
+    be saved, however, instead of being saved as values any
+    ``RangeIndex`` will be stored as a range in the metadata so it
+    doesn’t require much space and is faster. Other indexes will
```
bdice, line 245:
```suggestion
    doesn't require much space and is faster. Other indexes will
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -239,11 +266,11 @@
...
     timestamp[us] to the int96 format, which is the number of Julian
-    days and the number of nanoseconds since midnight. If ``False``,
-    timestamps will not be altered.
-row_group_size_bytes: integer or None, default None
+    days and the number of nanoseconds since midnight of 1970-01-01.
+    If ``False``, timestamps will not be altered.
+row_group_size_bytes: integer, default {row_group_size_bytes_val}
     Maximum size of each stripe of the output.
-    If None, 134217728 (128MB) will be used.
+    If None, {row_group_size_bytes_val} (128MB) will be used.
```
bdice, line 273: This formatted value isn't guaranteed to stay in sync with the text `(128MB)`. We might remove the parenthesized text, or make it match. Something like:
```suggestion
    If None, {row_group_size_bytes_val} ({row_group_size_bytes_val/1024/1024} MB) will be used.
```

python/cudf/cudf/utils/ioutils.py
```diff
@@ -1247,6 +1334,66 @@
...
+    Type of compression algorithm for the content
+mode : str
+    Mode in which file is opened
+iotypes : (), default (BytesIO)
+    Object type to exclude from file-like check
+use_python_file_object : boolean, default False
+    If True, Arrow-backed PythonFile objects will be used in place
+    of fsspec AbstractBufferedFile objects.
+open_file_options : dict, optional
+    Optional dictionary of key-word arguments to pass to
```
bdice, line 1356:
```suggestion
    Optional dictionary of keyword arguments to pass to
```

bdice (approved):
A few minor suggestions, otherwise LGTM. Thank you for the iterations!

galipremsagar:
@gpucibot merge

galipremsagar:
Thanks for patiently reviewing this PR @bdice & @wence-, I know this was a tedious one.

PR #12034, opened 2022-11-01
Enable building against the libarrow contained in pyarrow

vyasr:
## Description

This feature is a prerequisite for wheels. There is no real good reason to do this except to provide interop with a pyarrow wheel, so this option is marked as advanced. In the process of implementing this feature, I have also done some cleanup of `get_arrow.cmake` to try and simplify its logic. 

## Checklist


vyasr:
@robertmaynard @trxcllnt I'd appreciate some feedback from you two on the various TODO comments that I added to `get_arrow.cmake` to help figure out what we can do to simplify the logic there. I'm not sure how much of the complexity is unavoidable and how much we could simplify.

cpp/src/interop/to_arrow.cu
```diff
@@ -398,9 +398,9 @@ std::shared_ptr<arrow::Table> to_arrow(table_view input,
     arrays.end(),
     metadata.begin(),
     std::back_inserter(fields),
-    [](auto const& array, auto const& meta) { return arrow::field(meta.name, array->type()); });
+    [](auto const& array, auto const& meta) { return std::make_shared<arrow::Field>(meta.name, array->type()); });
```
vyasr, line 401: @trxcllnt should we upstream this issue to the arrow devs indicating that the library contained in the wheels isn't exporting all the symbols that we would expect?

trxcllnt: I made this change before we realized the ABI was different... maybe the symbol really is there? Can we revert this and see if it still breaks?

vyasr: Hmm I thought that I did that, but maybe that was also before I was correctly forwarding along the ABI flags from the Python to the C++ build here. Let me test again to be sure.

vyasr: Yup I guess I was wrong. This is in fact not necessary :tada:

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -20,43 +20,109 @@
...
+        TRUE
+        PARENT_SCOPE
+    )
+    set(ARROW_LIBRARIES
+        arrow_shared
+        PARENT_SCOPE
+    )
+
+    # When using the libarrow inside a wheel we must build libcudf with the old ABI because
+    # pyarrow's `libarrow.so` is compiled for manylinux2014 (centos7 toolchain) which uses the old
```
bdice, line 76: I'm not sure if I follow where this manylinux2014 is coming from. It looks like the PyArrow wheels are for `manylinux_2_17` (glibc 2.17)?

https://pypi.org/project/pyarrow/9.0.0/#files

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -20,43 +20,109 @@
...
+    # ABI. Note that these flags will often be redundant because we build wheels in manylinux
+    # containers that actually have the old libc++ anyway, but setting them explicitly ensures
+    # correct and consistent behavior in all other cases such as aarch builds on newer manylinux or
+    # testing builds in newer containers. TODO: Tests will not build successfully now without also
+    # propagating these options to builds of GTest. Similarly, benchmarks will not work without
+    # updating GBench (and possibly NVBench) builds. Do we care about that since we don't anticipate
```
bdice, line 82: I don't think I care. Just change this TODO so it documents the expected failures.

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -125,61 +200,73 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
...
-          ${ARROW_LIBRARY}
-          INTERFACE "$<BUILD_INTERFACE:${Arrow_SOURCE_DIR}/cpp/src>"
-                    "$<BUILD_INTERFACE:${Arrow_SOURCE_DIR}/cpp/src/generated>"
-                    "$<BUILD_INTERFACE:${Arrow_SOURCE_DIR}/cpp/thirdparty/hadoop/include>"
-                    "$<BUILD_INTERFACE:${Arrow_SOURCE_DIR}/cpp/thirdparty/flatbuffers/include>"
-        )
-      endforeach()
     endif()
+    #
+    # This shouldn't be necessary!
```
bdice, line 248: I think this comment indicates something slightly different than your intent. I read this as "this code can be removed" but you seem to mean "this is a hack." I would just remove this statement so the comments below can clarify what is happening and why.

cpp/cmake/thirdparty/get_arrow.cmake
vyasr: 2_17 and 2014 are the same thing, just aliases.

cpp/cmake/thirdparty/get_arrow.cmake
vyasr: Git is confused by the changes to indentation introduced in this PR (I removed an outer if-else). This comment is not new. That said, I am trying to do some cleanup in this PR so yes I can remove this.

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -125,61 +200,73 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
...
-  # Arrow_ADDED: set if CPM downloaded Arrow from Github Arrow_DIR:   set if CPM found Arrow on the
-  # system/conda/etc.
-  if(Arrow_ADDED OR Arrow_DIR)
-    if(BUILD_STATIC)
-      list(APPEND ARROW_LIBRARIES arrow_static)
-    else()
-      list(APPEND ARROW_LIBRARIES arrow_shared)
-    endif()
+  if(BUILD_STATIC)
+    set(ARROW_LIBRARIES
```
robertmaynard, line 210: This isn't doing what you expect. When you set a variable with `PARENT_SCOPE` it becomes set once you return to the parent scope. That means the below code that iterates `ARROW_LIBRARIES` won't work since the local variable `ARROW_LIBRARIES` doesn't exist. 

This breaks builds with arrow built from source. Instead you should use the previous behavior of building a local variable and setting it up to the parent scope

robertmaynard (changes requested):

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -20,43 +20,109 @@
...
+    set(CUDF_CUDA_FLAGS
+        "${CUDF_CUDA_FLAGS}"
+        PARENT_SCOPE
+    )
+
+    return()
+  endif()
+
+  # TODO: How would the targets have been created before this function is called? Is this just to
+  # ensure that the function is idempotent? If so, why do we need to set these variables again?
```
robertmaynard, line 99: Correct they could be brought in by another project that cpm'ed arrow before cudf did. So we need to set the variable back up again so cudf is idempotent.

cpp/cmake/thirdparty/get_arrow.cmake
vyasr: I see, I did not realize that. In the case of `ARROW_FOUND` I think the old behavior was actually causing problems because the variable wasn't being propagated up in certain cases, but for `ARROW_LIBRARIES` it makes sense to stick to the older behavior then.

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -69,8 +135,13 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
     set(ARROW_BUILD_STATIC ON)
     set(ARROW_BUILD_SHARED OFF)
     # Turn off CPM using `find_package` so we always download and make sure we get proper static
-    # library
+    # library TODO: Can we use `CPM_DOWNLOAD_<PackageName>` (aka `CPM_DOWNLOAD_ARROW`) instead? This
```
vyasr, line 138: @robertmaynard thoughts on this?****

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -69,8 +135,13 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
...
+    # is a big hammer, although I guess it is effectively limited which packages will be affected by
+    # the scope of the variable here.
```
vyasr, line 140: @robertmaynard thoughts on this?

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -20,43 +20,109 @@
...
+    BUILD_EXPORT_SET cudf-exports
+    INSTALL_EXPORT_SET cudf-exports
+    HEADER_NAMES arrow/python/arrow_to_pandas.h
+  )
+
+  find_package(Arrow ${PYARROW_VERSION} MODULE REQUIRED GLOBAL)
+  add_library(arrow_shared ALIAS Arrow::Arrow)
+
+  # TODO: Should these both be here? I think so, the `rapids_find_generate_module` doesn't seems to
+  # create the necessary `find_dependency` calls in the config files.
```
vyasr, line 51: I'm not certain but these both currently seem necessary.

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -90,8 +161,12 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
   endif()
 
   rapids_cpm_find(
-    Arrow ${VERSION}
-    GLOBAL_TARGETS arrow_shared parquet_shared arrow_dataset_shared
+    Arrow
+    ${VERSION}
+    # TODO: Should we set the list of global targets conditionally based on whether we're building
+    # shared/static and whether parquet is enabled?
```
vyasr, line 167: This mostly feels like a cleanup task, I don't think it's super important

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -125,61 +200,73 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
...
-        if(NOT Parquet_DIR)
-          # Set this to enable `find_package(Parquet)`
-          set(Parquet_DIR "${Arrow_DIR}")
-        endif()
-        # Set this to enable `find_package(ArrowDataset)`
-        set(ArrowDataset_DIR "${Arrow_DIR}")
-        find_package(ArrowDataset REQUIRED QUIET)
+  # Arrow_DIR:   set if CPM found Arrow on the system/conda/etc.
+  if(Arrow_DIR)
+    # TODO: Under what circumstances do we need a `find_package` _after_ `rapids_cpm_find`??
```
vyasr, line 223: I'm really not sure why this is needed here, it seems entirely redundant.

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -230,6 +317,9 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
       BUILD Arrow
       VERSION ${VERSION}
       EXPORT_SET arrow_targets
+      # TODO: What about all the other GLOBAL_TARGETS specified in the `rapids_cpm_find` command?
+      # Also, should this be constructed conditionally based on whether we're using shared or static
+      # libs for arrow?
```
vyasr, line 322: As above, the conditional construction seems redundant, but mismatch with `rapids_cpm_find` above seems odd.

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -279,6 +369,12 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
     endif()
   endif()
   # We generate the arrow-configfiles when we built arrow locally, so always do `find_dependency`
+  # TODO: Isn't the second call here redundant? The above `rapids_export` command will also encode a
+  # `find_dependency` in the INSTALL export set. It's less obvious whether we should still include
+  # the first `rapids_export_package` below since that will translate to a `find_dependency` call in
+  # the build config and `rapids-export` will put a `CPMFindPackage` there instead; is our goal to
+  # add `find_dependency`? Since it'll happen after `CPMFindPackage` I think it'll just be
+  # redundant.
```
vyasr, line 377: Thoughts on this one?

python/cudf/CMakeLists.txt
```diff
@@ -62,6 +78,14 @@ if(NOT cudf_FOUND)
   install(TARGETS cudf DESTINATION ${cython_lib_dir})
 endif()
 
+if(USE_LIBARROW_FROM_PYARROW)
+  # TODO: It would be nice to avoid setting these globally, but we'd need to manually set this for
+  # each of the rapids-cython targets otherwise which seems undesirable. For now I'm setting this
+  # after the `add_subdirectory(cpp)` call to limit the scope, maybe that's good enough.
```
vyasr, line 84: Can you think of any cleaner way to set up these variables rather than modifying the `CMAKE_*_FLAGS` variables directly? I think it might be OK here since it should only affect things compiled below this which are just our own extension modules, it just feels like a slightly more extensive modification than I would hope.

vyasr (commented):
@robertmaynard I left comments to help clarify exactly where I think there are still open questions that I could use some help with.

python/cudf/CMakeLists.txt
robertmaynard: Yeah we should be able to attach these as INTERFACE target_compile_options to the `${ARROW_LIBRARIES}` the same we do for include directories. That will isolate the flag to only targets that link to arrow.

cpp/cmake/thirdparty/get_arrow.cmake
robertmaynard: You can have a superset of targets for all configurations in the global target set as only those that exist will be promoted. It was design for this use-case of not wanting to handle configuration permutations/combinations

cpp/cmake/thirdparty/get_arrow.cmake
robertmaynard: Yes we should use `CPM_DOWNLOAD_Arrow` to limit the download

cpp/cmake/thirdparty/get_arrow.cmake
robertmaynard: The superset in the rapids_cpm_find is needed when you locate an existing built arrow. In that case all the optional arrow components like ArrowDataset are found in this single call so we need to mark all the targets as global.

Likewise when we are generating the components we mark the targets from that subcomponent as global to a have a clean mapping of components to targets.

python/cudf/CMakeLists.txt
vyasr: Good call, this is much nicer.

trxcllnt (approved):

cpp/cmake/thirdparty/get_arrow.cmake
trxcllnt: Can't remember why we needed this, but it does look redundant so we can probably remove it?

vyasr: It looks like we need this because `rapids_cpm_find` doesn't propagate up all the variables that `find_package` sets for Arrow.

robertmaynard (approved):

vyasr:
@gpucibot merge

PR #12040, opened 2022-11-01
Move `_label_encoding` from Series to Column

### LINKED ISSUE #9474: [BUG] cudf uses deprecated `Series.label_encoding` function internally
There are multiple places where cuDF internally calls the `Series.label_encoding` method, which was deprecated in #9289:
- https://github.com/rapidsai/cudf/blob/branch-21.12/python/cudf/cudf/core/column/column.py#L968
- https://github.com/rapidsai/cudf/blob/branch-21.12/python/cudf/cudf/core/algorithms.py#L62
- https://github.com/rapidsai/cudf/blob/branch-21.12/python/cudf/cudf/core/dataframe.py#L3248
Note that the latter is the `DataFrame` implementation of `label_encoding`.

The first two should be rewritten to not use the deprecated function, while the third should also be deprecated.

shwina:
Maybe closes #9474.

The `Series._label_encoding` function is used to integer encode columns. It makes much more sense for this to be a method of `Column` rather than `Series` (the index is irrelevant). Thus, I've moved the implementation to `Column`,


python/cudf/cudf/core/algorithms.py
```diff
@@ -58,9 +58,7 @@ def factorize(values, sort=False, na_sentinel=-1, size_hint=None):
 
     cats = values._column.dropna().unique().astype(values.dtype)
 
-    name = values.name  # label_encoding mutates self.name
     labels = values._label_encoding(cats=cats, na_sentinel=na_sentinel).values
```
vyasr, line 61: I'm confused, how was this working before? It looks like `values` is a `Column` here, not a `Series`, but the `Column` method didn't exist before this PR right? Is this dead code somehow?

vyasr: Oh sorry I missed that the line above assigned the output to `cats`, not back to `values`. Never mind.

python/cudf/cudf/core/column/column.py
```diff
@@ -1212,6 +1212,45 @@ def _with_type_metadata(self: ColumnBase, dtype: Dtype) -> ColumnBase:
         """
         return self
 
+    def _label_encoding(self, cats, dtype=None, na_sentinel=-1):
```
vyasr, line 1215: Can we add docs here?

python/cudf/cudf/core/column/column.py
```diff
@@ -1212,6 +1212,45 @@ def _with_type_metadata(self: ColumnBase, dtype: Dtype) -> ColumnBase:
...
+
+        def _return_sentinel_column():
+            return cudf.core.column.full(
+                size=len(self), fill_value=na_sentinel, dtype=dtype
+            )
+
+        if dtype is None:
+            dtype = min_scalar_type(max(len(cats), na_sentinel), 8)
+
+        cats = as_column(cats)
```
vyasr, line 1226: This is an internal method. Do we not already know at this point whether `cats` is a column or not?

vyasr (commented):
I agree that it looks like we can't actually remove the method. Should we rename it to more accurately reflect its current role rather than the name of the former public API that it reflects (I don't have any ideas, asking in case you do).

python/cudf/cudf/core/column/column.py
shwina: Done

python/cudf/cudf/core/column/column.py
shwina: I made it so that we always pass a column.

vyasr (approved):

vyasr (approved):

vyasr:
@gpucibot merge

Issue #11934, opened 2022-10-17
[FEA] cudf.DataFrame.from_dict


mattf:
**Is your feature request related to a problem? Please describe.**
rewriting code from pandas into cudf, trying to use `import cudf as pd`

**Describe the solution you'd like**
`cudf.DataFrame.from_dict` matching https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.from_dict.html

rjzamora:
+1 to this feature (and to it's `dask_cudf` counterpart).

quasiben:
From a call with @shwina.  We can already do something dict things like:


```python
In [4]: cudf.DataFrame({'a': cudf.Series([1,2,3]), 'b': cudf.Series([4,5,6], index=[9,10,11])})
Out[4]:
       a     b
0      1  <NA>
1      2  <NA>
2      3  <NA>
9   <NA>     4
10  <NA>     5
11  <NA>     6
```

Hmm, actually the DataFrame constructor is pretty robust:


```python
In [19]: cudf.DataFrame({'c': ['a', 'b', 'd'], 'b': cudf.Series([1, 2, 3]), 'd': pd.Series([10, 11, 12])})
Out[19]:
   c  b   d
0  a  1  10
1  b  2  11
2  d  3  12

In [23]: cudf.DataFrame({'a': cudf.Series([1,2,3]), 'b': cudf.Series([4,5,6], index=[9,10,11]),'c': pd.Series(['q', 'r', 's'])})
Out[23]:
       a     b     c
0      1  <NA>     q
1      2  <NA>     r
2      3  <NA>     s
9   <NA>     4  <NA>
10  <NA>     5  <NA>
11  <NA>     6  <NA>
```
@mattf can you, as a test, check if the standard constructor works for your usecase ?  If so, maybe this is as easy as redirecting `from_dict` to `DataFrame(...)`

rjzamora:
@quasiben - I also suspect that `cudf.from_dict` can simply wrap `cudf.DataFrame` - Probably raising a `NotImplementedError` for `orient="index"` and `orient="tight"`.  In the long run, it may be nice to support the same `orient` options as `pandas`. However,  the most important thing is probably that the API *exists*. 

[...]

PR #10381, opened 2022-03-02
Consolidate some Frame APIs

vyasr:
This PR contributes to #9038 by combining common code for `append`, `astype`, `drop`, and `explode`. It's not a small LOC change, but most of the changed lines are just moving code around. Some minor rewrites of `astype` in this PR do significantly speed up those calls (2x faster for <1M rows). Additionally, I've deprecated `append` since it is now deprecated in pandas. Note that it has not been removed in pandas, so we should not remove it until they do.

PR #12048, opened 2022-11-02
[REVIEW] Add support for `DataFrame.from_dict`\`to_dict` and `Series.to_dict`

galipremsagar:
## Description

Resolves: #11934 

- [x] Adds `DataFrame.from_dict` and `DataFrame.to_dict`
- [x] Adds `Series.to_dict`

## Checklist


vyasr (commented):
This adds a lot more logic than I had anticipated. Out of curiosity, I copied this implementation into my current branch, then added a second method `from_dict2` with an identical signature that is just defined as:
```
return cls.from_pandas(pd.DataFrame.from_dict(data, orient, dtype, columns))
```

Here are some quick benchmark results:
```
In [9]: d = {f'i': np.random.rand(1000) for i in range(20)}

In [10]: %timeit cudf.DataFrame.from_dict2(d)
279 µs ± 1.51 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [11]: %timeit cudf.DataFrame.from_dict(d)
315 µs ± 1.26 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [12]: d = {f'i': np.random.rand(10) for i in range(20)}

In [13]: %timeit cudf.DataFrame.from_dict2(d)
277 µs ± 345 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [14]: %timeit cudf.DataFrame.from_dict(d)
307 µs ± 253 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [15]: d = {f'i': np.random.rand(10000) for i in range(50)}

In [16]: %timeit cudf.DataFrame.from_dict2(d)
310 µs ± 322 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [17]: %timeit cudf.DataFrame.from_dict(d)
316 µs ± 120 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

Is there any benefit to the current, much more complex approach rather than just using the pandas fallback in this case? The complexity of our constructor is high enough that it seems like bypassing it using `from_pandas` recoups more than enough performance, and this way we really don't have to think about the semantics much at all.

galipremsagar:
> Is there any benefit to the current, much more complex approach rather than just using the pandas fallback in this case? The complexity of our constructor is high enough that it seems like bypassing it using `from_pandas` recoups more than enough performance, and this way we really don't have to think about the semantics much at all.


So, yes this is for sure quicker for CPU-backed data..I had to vendor the entire logic from pandas and do a special handling to support `cudf.Series` as values in the dict inputs, like going through pandas API will fail because of an implicit cast:
```python
In [1]: import cudf

In [2]: def v2(data):
   ...:     return cudf.from_pandas(pd.DataFrame.from_dict(data))
   ...: 

In [3]: import pandas as pd

In [4]: data = {"a": cudf.Series([1, 2, 3, 4]), "B": cudf.Series([10, 11, 12, 13])}

In [5]: v2(data)
TypeError                                 Traceback (most recent call last)
Cell In [5], line 1
----> 1 v2(data)

Cell In [2], line 2, in v2(data)
      1 def v2(data):
----> 2     return cudf.from_pandas(pd.DataFrame.from_dict(data))

File /nvme/0/pgali/envs/cudfdev/lib/python3.9/site-packages/pandas/core/frame.py:1762, in DataFrame.from_dict(cls, data, orient, dtype, columns)
   1756     raise ValueError(
   1757         f"Expected 'index', 'columns' or 'tight' for orient parameter. "
   1758         f"Got '{orient}' instead"
   1759     )
   1761 if orient != "tight":
-> 1762     return cls(data, index=index, columns=columns, dtype=dtype)
   1763 else:
   1764     realdata = data["data"]

File /nvme/0/pgali/envs/cudfdev/lib/python3.9/site-packages/pandas/core/frame.py:662, in DataFrame.__init__(self, data, index, columns, dtype, copy)
    656     mgr = self._init_mgr(
    657         data, axes={"index": index, "columns": columns}, dtype=dtype, copy=copy
    658     )
    660 elif isinstance(data, dict):
    661     # GH#38939 de facto copy defaults to False only in non-dict cases
--> 662     mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
    663 elif isinstance(data, ma.MaskedArray):
    664     import numpy.ma.mrecords as mrecords

File /nvme/0/pgali/envs/cudfdev/lib/python3.9/site-packages/pandas/core/internals/construction.py:493, in dict_to_mgr(data, index, columns, dtype, typ, copy)
    489     else:
    490         # dtype check to exclude e.g. range objects, scalars
    491         arrays = [x.copy() if hasattr(x, "dtype") else x for x in arrays]
--> 493 return arrays_to_mgr(arrays, columns, index, dtype=dtype, typ=typ, consolidate=copy)

File /nvme/0/pgali/envs/cudfdev/lib/python3.9/site-packages/pandas/core/internals/construction.py:123, in arrays_to_mgr(arrays, columns, index, dtype, verify_integrity, typ, consolidate)
    120         index = ensure_index(index)
    122     # don't force copy because getting jammed in an ndarray anyway
--> 123     arrays = _homogenize(arrays, index, dtype)
    124     # _homogenize ensures
    125     #  - all(len(x) == len(index) for x in arrays)
    126     #  - all(x.ndim == 1 for x in arrays)
   (...)
    129 
    130 else:
    131     index = ensure_index(index)

File /nvme/0/pgali/envs/cudfdev/lib/python3.9/site-packages/pandas/core/internals/construction.py:617, in _homogenize(data, index, dtype)
    614             val = dict(val)
    615         val = lib.fast_multiget(val, oindex._values, default=np.nan)
--> 617     val = sanitize_array(
    618         val, index, dtype=dtype, copy=False, raise_cast_failure=False
    619     )
    620     com.require_length_match(val, index)
    622 homogenized.append(val)

File /nvme/0/pgali/envs/cudfdev/lib/python3.9/site-packages/pandas/core/construction.py:616, in sanitize_array(data, index, dtype, copy, raise_cast_failure, allow_2d)
    613 # materialize e.g. generators, convert e.g. tuples, abc.ValueView
    614 if hasattr(data, "__array__"):
    615     # e.g. dask array GH#38645
--> 616     data = np.array(data, copy=copy)
    617 else:
    618     data = list(data)

File /nvme/0/pgali/envs/cudfdev/lib/python3.9/site-packages/cudf/core/frame.py:447, in Frame.__array__(self, dtype)
    446 def __array__(self, dtype=None):
--> 447     raise TypeError(
    448         "Implicit conversion to a host NumPy array via __array__ is not "
    449         "allowed, To explicitly construct a GPU matrix, consider using "
    450         ".to_cupy()\nTo explicitly construct a host matrix, consider "
    451         "using .to_numpy()."
    452     )

TypeError: Implicit conversion to a host NumPy array via __array__ is not allowed, To explicitly construct a GPU matrix, consider using .to_cupy()
To explicitly construct a host matrix, consider using .to_numpy().
```

python/dask_cudf/dask_cudf/backends.py
```diff
@@ -474,10 +474,9 @@ def from_dict(data, npartitions, orient="columns", **kwargs):
 
             if orient != "columns":
                 raise ValueError(f"orient={orient} is not supported")
-            # TODO: Use cudf.from_dict
-            # (See: https://github.com/rapidsai/cudf/issues/11934)
+
```
galipremsagar, line 477: cc: @rjzamora for review

python/dask_cudf/dask_cudf/backends.py
```diff
@@ -474,10 +474,9 @@ def from_dict(data, npartitions, orient="columns", **kwargs):
 
             if orient != "columns":
                 raise ValueError(f"orient={orient} is not supported")
```
rjzamora, line 476: Cant we remove this check (and remove `orient` from the function signature) now that we have a proper `cudf.DataFrame.from_dict` implementation?

It should also be fair game to officially add `from_dict` to `dask_cudf`, and simply invoke that function here. E.g.

```python
        @staticmethod
        def from_dict(data, **kwargs):
            from dask_cudf import from_dict

            return from_cudf(data, **kwargs)
```

galipremsagar: Done, also made an upstream PR that is needed to support this change: https://github.com/dask/dask/pull/9628

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,250 @@ def _make_operands_and_index_for_binop(
...
+        """  # noqa: E501
+        index = None
+        orient = orient.lower()
+        if orient == "index":
+
+            if len(data) > 0:
+                data_list = list(data.values())
+                if isinstance(data_list[0], (pd.Series, Series, dict)):
+                    if isinstance(data_list[0], Series):
+                        data = {key: val.to_pandas() for key, val in data}
```
wence-, line 2090:
```suggestion
                        data = {key: val.to_pandas() for key, val in data.items()}
```

Which suggests this code path is not tested?

wence-: Seems unfortunate that you need to come back to pandas just to iterate over things.

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,250 @@ def _make_operands_and_index_for_binop(
...
+                    data = _from_nested_dict(data)
+                else:
+                    index = list(data.keys())
+                    data = data_list
+            df = pd.DataFrame(data=data, index=index, columns=columns)
+            df = cudf.from_pandas(df)
```
wence-, line 2096: Is there really no way to do this on the GPU if we start with GPU dataframes?

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,250 @@ def _make_operands_and_index_for_binop(
...
+                    f"cannot use columns parameter with orient='{orient}'"
+                )
+        else:
+            raise ValueError(
+                f"Expected 'index', 'columns' or 'tight' for orient "
+                f"parameter. Got '{orient}' instead"
+            )
+
+        if orient != "tight":
+            return cls(data, index=index, columns=columns, dtype=dtype)
```
wence-, line 2112: `columns` must be `None` here?

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,250 @@ def _make_operands_and_index_for_binop(
...
+        else:
+            realdata = data["data"]
```
wence-, line 2114: I _think_ this is the `orient == "columns"` case. It would be clearer if the logic didn't do multiple rounds of matching, I think.

python/cudf/cudf/core/dataframe.py
```diff
@@ -7444,3 +7688,11 @@ def _reassign_categories(categories, cols, col_idxs):
...
                 size=cols[name].size,
             )
+
+
+def _from_nested_dict(data) -> defaultdict:
+    new_data: defaultdict = defaultdict(dict)
+    for index, s in data.items():
+        for col, v in s.items():
+            new_data[col][index] = v
+    return new_data
```
wence-, line 7698: I really don't like this name, since it doesn't say anything either about what it expects as input, nor what it delivers as output. I think what you're doing is transposing a dictionary?

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,250 @@ def _make_operands_and_index_for_binop(
...
+        index = None
+        orient = orient.lower()
+        if orient == "index":
+
+            if len(data) > 0:
+                data_list = list(data.values())
+                if isinstance(data_list[0], (pd.Series, Series, dict)):
+                    if isinstance(data_list[0], Series):
+                        data = {key: val.to_pandas() for key, val in data}
+                    data = _from_nested_dict(data)
```
wence-, line 2091: As below, this name is note very enlightening. Given this is the only use, would it not make more sense to just inline the implementation?

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,250 @@ def _make_operands_and_index_for_binop(
...
+        a  b   1  3
+           c   2  4
+        """  # noqa: E501
+        index = None
+        orient = orient.lower()
+        if orient == "index":
+
+            if len(data) > 0:
+                data_list = list(data.values())
+                if isinstance(data_list[0], (pd.Series, Series, dict)):
```
wence-, line 2088: This assumes that the user passed in homogeneous data. Is that guaranteed? Should we instead do:
```python
try:
    typ, = set(map(type, data.values()))
except ValueError:
    raise ValueError("For orient='index' all values must have same type") # or something
if typ is Series:
    data = {k: v.to_pandas() for k, v in data}
```
WDYT?

wence- (changes requested):
Wondering if we can make the `from_dict` logic at all clearer.

python/cudf/cudf/core/dataframe.py
galipremsagar: Done

python/cudf/cudf/core/dataframe.py
galipremsagar: Why do you think so? This is for `orient != 'tight'`. `columns` is fetched from `data` when `orient=='tight'`

python/cudf/cudf/core/dataframe.py
galipremsagar: Done, we can achieve it using transpose.

python/cudf/cudf/core/dataframe.py
galipremsagar: Remove it

python/cudf/cudf/core/dataframe.py
wence-: Ah, ok.

python/cudf/cudf/core/dataframe.py
wence-: orient can be "columns", "index", or "tight"; if `orient == "index"` we have already returned by this point. Then on line 2103 you check if `orient == "columns" or orient == "tight"` (must succeed) and then on line 2104 if `columns is not None` we raise a ValueError. So at this point `orient != "tight"` means `orient == "columns"` and columns must have been `None` otherwise we would have bailed out with a `ValueError` on line 2105.

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,253 @@ def _make_operands_and_index_for_binop(
...
+                    f"cannot use columns parameter with orient='{orient}'"
+                )
+        else:
+            raise ValueError(
+                f"Expected 'index', 'columns' or 'tight' for orient "
+                f"parameter. Got '{orient}' instead"
+            )
+
+        if orient != "tight":
+            return cls(data, index=index, columns=columns, dtype=dtype)
```
galipremsagar, line 2115:
```suggestion
            return cls(data, index=index, dtype=dtype)
```

python/cudf/cudf/core/dataframe.py
galipremsagar: Oh yes, you are right. Removed passing of columns.

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,253 @@ def _make_operands_and_index_for_binop(
...
+        index = None
+        orient = orient.lower()
+        if orient == "index":
+
+            if len(data) > 0:
+                data_list = list(data.values())
+                if isinstance(data_list[0], Series):
+                    data = DataFrame.from_dict(data=data).T
+                elif isinstance(data_list[0], (pd.Series, dict)):
+                    new_data: defaultdict = defaultdict(dict)
```
vyasr, line 2091: mypy should be able to infer the type here, no?

galipremsagar: Yup, that was not needed.

galipremsagar, line 2091:
```suggestion
                    new_data = defaultdict(dict)
```

galipremsagar: Removed

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,253 @@ def _make_operands_and_index_for_binop(
...
+        >>> df.to_dict('records', into=dd)
+        [defaultdict(<class 'list'>, {'col1': 1, 'col2': 0.5}),
+         defaultdict(<class 'list'>, {'col1': 2, 'col2': 0.75})]
+        """  # noqa: E501
+        orient = orient.lower()
+
+        if orient == "series":
+            # Special case needed because we don't want
+            # convert the values as pd.Series, and want
+            # them as cudf.Series
```
vyasr, line 2236:
```suggestion
            # Special case needed to avoid converting
            # cudf.Series objects into pd.Series
```

vyasr (approved):
It would be nice if 1) we could leverage the pandas method directly in cases where we know that it would be faster, and 2) if we didn't have to handle so many edge cases before devolving to the constructor, but I don't see how to do (1) without doing introspection that will slow down other code paths and I think (2) seems necessary. If other reviewers have more ideas I would welcome them, but I think I'm good with this code as is.

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,253 @@ def _make_operands_and_index_for_binop(
...
+        index = None
+        orient = orient.lower()
+        if orient == "index":
+
+            if len(data) > 0:
+                data_list = list(data.values())
+                if isinstance(data_list[0], Series):
+                    data = DataFrame.from_dict(data=data).T
+                elif isinstance(data_list[0], (pd.Series, dict)):
+                    new_data = defaultdict(dict)
```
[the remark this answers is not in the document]
vyasr: Ah well that's too bad. I'm guessing the defaultdict type needs type stubs added to the built in typeshed. Can look into it later.

wence-:
> It would be nice if 1) we could leverage the pandas method directly in cases where we know that it would be faster, and 2) if we didn't have to handle so many edge cases before devolving to the constructor, but I don't see how to do (1) without doing introspection that will slow down other code paths and I think (2) seems necessary. If other reviewers have more ideas I would welcome them, but I think I'm good with this code as is.

How about this rewrite:
```python
    @staticmethod
    def __create_index(indexlist, namelist, library):
        try:
            name, = namelist
            return library.Index(indexlist, name=name)
        except ValueError:
            return library.MultiIndex.from_tuples(
                indexlist, names=namelist
            )
            
    @classmethod
    def from_dict(cls, data, orient="columns", dtype=None, index=None, columns=None):
        if orient == "index":
            if len(data) > 0 and isinstance(next(iter(data.values())), cudf.Series):
                data = cls(data).T
                return cls(data, index=None, columns=columns, dtype=dtype)
            else:
                return cls.from_pandas(
                    pd.DataFrame.from_dict(
                        data=data,
                        orient=orient,
                        dtype=dtype,
                        index=index,
                        columns=columns,
                    )
                )
        elif orient == "columns":
            if columns is not None:
                raise ValueError("Cannot use columns parameter with orient='columns'")
            return cls(data, index=index, columns=None, dtype=dtype)
        elif orient == "tight":
            if columns is not None:
                raise ValueError("Cannot use columns parameter with orient='right'")
            index = cls.__create_index(data["index"], data["index_names"], cudf)
            columns = cls.__create_index(data["columns"], data["column_names"], pd)
            return cls(data["data"], index=index, columns=columns, dtype=dtype)
        else:
            raise ValueError(f"Expected 'index', 'columns', or 'tight' got {orient=}")
```

Which (personally) I find a bit clearer. It simplifies the dict transpose in the case where the values are not cudf.Series by just deferring to pandas. This may, however, be bad if the values are provided as cupy arrays (I think).

Other changes (might be worthwhile anyway): handle each `orient` case separately with early returns (which I think makes the logic a bit easier to read); define `__create_index` as a `staticmethod` since it's not closing over any variables.

jakirkham:
> This may, however, be bad if the values are provided as cupy arrays (I think).

Think this is a use case we should prepare for (and ideally test against)

galipremsagar:
@wence- @jakirkham Addressed your reviews. Could you please re-review/approve when you get chance?

galipremsagar:
Also cc: @vyasr since code has changed tiny bit

python/cudf/cudf/core/frame.py
```diff
@@ -321,15 +321,19 @@ def __len__(self):
...
             else:
-                result[col_name] = col.copy() if copy else col
+                result_data[col_name] = col.copy() if copy else col
 
-        return result
+        return ColumnAccessor._create_unsafe(
+            data=result_data,
+            multiindex=self._data.multiindex,
+            level_names=self._data.level_names,
+        )
```
wence-, line 336: Why does `Frame.astype` return a `ColumnAccessor`? Should this not return `self._from_data(data=ColumnAccessor)` ?

python/cudf/cudf/core/dataframe.py
```diff
@@ -1992,6 +1992,247 @@ def _make_operands_and_index_for_binop(
...
+
+        orient = orient.lower()
+        if orient == "index":
+            if len(data) > 0 and isinstance(
+                next(iter(data.values())), (cudf.Series, cupy.ndarray)
+            ):
+                result = cls(data).T
+                result.columns = columns
+                if dtype is not None:
+                    result = result.astype(dtype)
```
wence-, line 2091: Would be nice to have coverage for this case in a test.

python/cudf/cudf/core/frame.py
galipremsagar: @vyasr was this a conscious decision to return dict here in: https://github.com/rapidsai/cudf/pull/10381?

python/cudf/cudf/core/dataframe.py
galipremsagar: Done

python/cudf/cudf/core/frame.py
wence-: Think this is because `IndexedFrame.astype` calls this to build the `data` argument to `_from_data`. Since that just takes a `MutableMapping` either worked I guess, though previously multiindices were not passed in?

vyasr: It was a conscious decision because `IndexedFrame.astype` and `GenericIndex.astype` both use the returned value to construct the relevant objects, but in the case of `GenericIndex` we have to construct the appropriate subclass which means we have to use `_index_from_data` rather than `self._from_data`. Ideally `Frame.astype` would take the `data` it's currently returning and instead return `self._from_data(data)`, but that makes the code harder to reuse in `GenericIndex` and forces extra creation of cudf objects so I chose not to do that. In hindsight, it would probably have been better to rename the method to `Frame._astype` or something similar to indicate that it's a helper function that doesn't return exactly the same thing as all the public `astype` APIs. I would support such a rename, or a restructuring of the code if you could find a solution that reduced duplication further in the subclasses.

galipremsagar: I'd be +1 for renaming to the internal method `Frame._astype`. I can't think of a better refactor than it is currently, and don't want to make a widespread refactor at end of a release. WDYT @wence- ?

wence-: Let's raise an issue for 23.0x and leave as is here in the interrsts of making progress

wence- (approved):

python/cudf/cudf/core/frame.py
galipremsagar: done: https://github.com/rapidsai/cudf/issues/12140

galipremsagar:
@gpucibot merge

Issue #11949, opened 2022-10-20
[FEA][JNI] Track GPU memory usage at process and at local level


abellina:
This issue tracks the cuDF side of https://github.com/NVIDIA/spark-rapids/issues/6745.

Essentially, we'd like to be able to track the maximum amount of memory used when calling a set of cuDF functions from Spark. This would be something we enable during debug, as new code is getting implemented at least, but I could also see the Spark code having a mode where it will run slower, but it will help diagnose where memory is going as we call into cuDF.

The proposal is to add a global watermark for the whole process (maximum GPU memory used at any point in time) and also add a "local" maximum, that can be reset by the user as a debug tool.

abellina:
The watermark API added in #11950 is useful but we want to be able to factor in our spillable cache in the spark-rapids plugin, which means so far that we need finer grained optional callbacks on every allocate and free. This isn't high performance, but it doesn't seem to be the end of the world. As such, I am working on patch that should ultimately close this particular issue.

PR #11950, opened 2022-10-20
Add gpu memory watermark apis to JNI

abellina:
This PR addresses https://github.com/rapidsai/cudf/issues/11949.

We are adding methods to get the current memory usage watermarks at the whole process level and adding a "scoped" maximum, where the user can reset the initial value, run cuDF functions, and then call the API to get what happened since the reset.

For the scoped maximum, the `getScopedMaximumOutstanding` could have somewhat surprising results. If the scoped maximum is reset to 0 for example, and we only see frees for allocations done before the reset, we are going to see that the scoped maximum returned is 0. This is because our memory usage is literally negative in this scenario.

The APIs here assume that the caller process is using a single thread to call into the GPU (for Spark it would be 1 concurrent task).

Note I assume `Rmm.initialize` has been called, otherwise this doesn't track allocations done before that.

PR #12054, opened 2022-11-03
[JNI] Add debug-only onAllocated/onDeallocated to RmmEventHandler

abellina:
This adds `onAllocated` and `onDeallocated` to `RmmEventHandler` as debug callbacks. If the event handler is installed with debug enabled (in `Rmm.setEventHandler`) these callbacks will be invoked when an allocation or deallocation finishes.

It also fixes a bug with https://github.com/rapidsai/cudf/pull/11950 where the initial allocated amount was not getting set appropriately. It was getting set to 0, but instead it should be set to the new initial value/maximum.

Closes https://github.com/rapidsai/cudf/issues/11949.


java/src/main/java/ai/rapids/cudf/Rmm.java
```diff
@@ -185,14 +185,30 @@ public static void resetScopedMaximumBytesAllocated() {
...
    */
   public static void setEventHandler(RmmEventHandler handler) throws RmmException {
+    setEventHandler(handler, false);
+  }
+
+  /**
+   * Sets the event handler to be called on RMM events (e.g.: allocation failure) and
+   * optionally enable debug mode (callbacks on every allocate and deallocate)
+   *
+   * NOTE: Only use this while debugging, code will run slower!
```
jlowe, line 195: This note implies the entire method should not be used or code will run much slower which is incorrect.
```suggestion
   * NOTE: Only enable debug mode when necessary, as code will run much slower!
```

java/src/main/java/ai/rapids/cudf/Rmm.java
```diff
@@ -185,14 +185,30 @@ public static void resetScopedMaximumBytesAllocated() {
    * @throws RmmException if an active handler is already set
```
jlowe, line 185: 2022 copyrights

java/src/main/java/ai/rapids/cudf/RmmEventHandler.java
```diff
@@ -32,6 +32,18 @@ default boolean onAllocFailure(long sizeRequested) {
         "Unexpected invocation of deprecated onAllocFailure without retry count.");
   }
 
+  /**
+   * Invoked on every memory allocation when installed with debug mode enabled.
```
jlowe, line 36:
```suggestion
   * Invoked after every memory allocation when debug mode is enabled.
```

java/src/main/java/ai/rapids/cudf/RmmEventHandler.java
```diff
@@ -32,6 +32,18 @@ default boolean onAllocFailure(long sizeRequested) {
...
+   * @param size number of bytes allocated
+   */
+  default void onAllocated(long size) {}
+
+  /**
+   * Invoked on every memory deallocation when installed with debug mode enabled.
```
jlowe, line 42:
```suggestion
   * Invoked after every memory deallocation when debug mode is enabled.
```

java/src/main/native/src/RmmJni.cpp
```diff
@@ -308,7 +320,15 @@ class java_event_handler_memory_resource final : public device_memory_resource {
       }
     }
   }
+  bool supports_get_mem_info() const noexcept override { return resource->supports_get_mem_info(); }
```
jlowe, line 323:
```suggestion

    bool supports_get_mem_info() const noexcept override { return resource->supports_get_mem_info(); }
```

java/src/main/native/src/RmmJni.cpp
```diff
@@ -349,19 +369,50 @@ class java_event_handler_memory_resource final : public device_memory_resource {
...
+    }
+  }
 
-  std::pair<size_t, size_t> do_get_mem_info(rmm::cuda_stream_view stream) const override {
-    return resource->get_mem_info(stream);
+  void on_deallocated_callback(void *p, std::size_t size, rmm::cuda_stream_view stream) {
+    JNIEnv *env = cudf::jni::get_jni_env(jvm);
+    env->CallVoidMethod(handler_obj, on_deallocated_method, size);
+    if (env->ExceptionCheck()) {
+      env->ExceptionDescribe();
```
jlowe, line 384: Accidentally left in for debugging?

java/src/main/native/src/RmmJni.cpp
```diff
@@ -349,19 +369,50 @@ class java_event_handler_memory_resource final : public device_memory_resource {
...
+      throw std::runtime_error("onDeallocated handler threw an exception");
```
jlowe, line 385: Not sure it's safe to throw here, as this is likely called within a destructor. Throwing in a destructor will likely cause the C++ runtime to terminate the application harshly.

java/src/main/native/src/RmmJni.cpp
```diff
@@ -349,19 +369,50 @@ class java_event_handler_memory_resource final : public device_memory_resource {
                                  on_dealloc_threshold_method, "onDeallocThreshold", total_after);
   }
 
-  bool supports_get_mem_info() const noexcept override { return resource->supports_get_mem_info(); }
+  void on_allocated_callback(std::size_t num_bytes, rmm::cuda_stream_view stream) {
```
jlowe, line 372: It's a bit odd that these methods are part of this class, and this class shouldn't have to risk looking up Java methods it has no intentions of calling.  This logic belongs in the derived class because it's specific to that class.

abellina: I was wondering if we should have a class that is the "event handler wrapper". All it does is expose to the C++ JNI code the methods of the handler, and so it can get passed to the production java handler or the debug java handler in order to call those functions. Does this sound reasonable?

java/src/main/native/src/RmmJni.cpp
abellina: nice catch

java/src/main/native/src/RmmJni.cpp
abellina: Sounds good. I won't call `ExceptionCheck` since there's nothing for me to handle. I expect the java code would see an exception after returning from the java-side of the JNI code and then throw, but it wouldn't throw within cuDF.

abellina: Yeah confirmed we get an exception on the java side when not handled here, which seems fine.

java/src/main/native/src/RmmJni.cpp
```diff
@@ -348,20 +360,62 @@ class java_event_handler_memory_resource final : public device_memory_resource {
...
-  bool supports_get_mem_info() const noexcept override { return resource->supports_get_mem_info(); }
+class java_debug_event_handler_memory_resource final : public java_event_handler_memory_resource {
+public:
+  java_debug_event_handler_memory_resource(JNIEnv *env, jobject jhandler,
+                                           jlongArray jalloc_thresholds,
+                                           jlongArray jdealloc_thresholds,
+                                           device_memory_resource *resource_to_wrap)
+      : java_event_handler_memory_resource(env, jhandler, jalloc_thresholds, jdealloc_thresholds,
+                                           resource_to_wrap) {
+    jclass cls = env->GetObjectClass(jhandler);
```
jlowe, line 373: Need to check for failure to lookup class (very unlikely, but should still check).

abellina:
@jlowe does this look ok?  Also, I upmerged in hopes of resolving some odd Ci issues.

jlowe (approved):

abellina:
@gpucibot merge

PR #12056, opened 2022-11-03
Allow falling back to `shim_60.ptx` by default in `strings_udf`

brandon-b-miller:
In the context of distributed, `strings_udf` needs to import and set itself up without creating a CUDA context, as this can interfere with up the way the network is being set up. In this situation it can't use it's normal mechanism (which requires a context) to query the compute capability of the device, and it falls back on an environment variable `STRINGS_UDF_CC` that it needs to be passed from dask instead. A user can set this and their code will work no problem, but we also need some default configuration that just works when someone builds their code. Without knowing their setup beforehand this can be problematic, as such I originally added the default value of `cc=52` when the environment variable isn't set. This was however not exactly correct for a few reasons:

- It should be 60 I think since pascal is the oldest arch supported by rapids
- we don't always build shim_60.ptx especially in local mode.

This PR fixes this problem.

### DIFF
--- a/python/strings_udf/cpp/CMakeLists.txt
+++ b/python/strings_udf/cpp/CMakeLists.txt
@@ -92,6 +92,10 @@ endfunction()
 # Create the shim library for each architecture.
 set(SHIM_CUDA_FLAGS --expt-relaxed-constexpr -rdc=true)
 
+# always build a default PTX file in case RAPIDS_NO_INITIALIZE is set and the device cc can't be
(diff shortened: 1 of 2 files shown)


galipremsagar (approved):

vyasr (approved):

brandon-b-miller:
rerun tests

brandon-b-miller:
@gpucibot merge

PR #12059, opened 2022-11-03
Mark nvcomp zstd compression stable

jbrennan333:
## Description
NVCOMP zstd compression was added in 22.10, but marked experimental, meaning you have to define the environment variable `LIBCUDF_NVCOMP_POLICY=ALWAYS` to enable it.  After completing validation testing using the spark rapids plugin as documented here: https://github.com/NVIDIA/spark-rapids/issues/3037, we believe that we can now change the zstd compression status to stable, which will enable it in cudf by default.  `LIBCUDF_NVCOMP_POLICY=STABLE` is the default value.
## Checklist
  I made local modifications to parquet/orc tests/benchmarks to test with ZSTD.  Currently we don't test with ZSTD by default because it's possible to build with a version of nvcomp that does not support it.


ttnghia (approved):

davidwendt (approved):

davidwendt:
Does this require a documentation change somewhere else? 
Seems that if we are changing a default that this should be a _breaking_ change?

vuule:
> Does this require a documentation change somewhere else? Seems that if we are changing a default that this should be a _breaking_ change?

The docs at 
https://github.com/rapidsai/cudf/blob/f19bdbca559a29d08b5a8ea9a8ddf9d4723676ac/python/cudf/cudf/utils/ioutils.py#L237
https://github.com/rapidsai/cudf/blob/f19bdbca559a29d08b5a8ea9a8ddf9d4723676ac/python/cudf/cudf/utils/ioutils.py#L487
already include ZSTD. I don't see any other places where we list supported compression types.

> Seems that if we are changing a default that this should be a breaking change?

We're not really changing the default here; this PR basically enables a feature.

vuule:
rerun tests

vuule:
rerun tests

ttnghia:
Need to merge upstream to fix.

jbrennan333:
rerun tests

jbrennan333:
@vuule can you please merge if this looks good?

vuule (approved):

vuule:
@gpucibot merge

PR #12060, opened 2022-11-03
Fixes List offset bug in Nested JSON reader

karthikeyann:
## Description
Fixes List offset end last item write condition bug
If there is a list row followed by empty list in next row, the previous row's end is not written to offsets.

## Checklist

### DIFF
--- a/cpp/src/io/json/json_column.cu
+++ b/cpp/src/io/json/json_column.cu
@@ -532,7 +532,8 @@ void make_device_json_column(device_span<SymbolT const> input,
             .child_offsets[row_offsets[parent_node_id]] = ordered_row_offsets[i];
         }
         // TODO: verify if this code is right. check with more test cases.
-        if (i == num_nodes - 1 || (col_ids[i] != col_ids[i + 1])) {
+        if (i == num_nodes - 1 ||
+            (col_ids[i] != col_ids[i + 1] or ordered_parent_node_ids[i + 1] != parent_node_id)) {
           // last value of list child_offset is its size.
           d_columns_data[original_col_ids[parent_node_id]]
             .child_offsets[row_offsets[parent_node_id] + 1] = ordered_row_offsets[i] + 1;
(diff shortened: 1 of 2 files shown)


cpp/src/io/json/json_column.cu
```diff
@@ -532,7 +532,8 @@ void make_device_json_column(device_span<SymbolT const> input,
             .child_offsets[row_offsets[parent_node_id]] = ordered_row_offsets[i];
         }
         // TODO: verify if this code is right. check with more test cases.
-        if (i == num_nodes - 1 || (col_ids[i] != col_ids[i + 1])) {
+        if (i == num_nodes - 1 ||
+            (col_ids[i] != col_ids[i + 1] or ordered_parent_node_ids[i + 1] != parent_node_id)) {
```
vyasr, line 536: Let's not mix `||` and `or` in the same line. Pick one.

vyasr (approved):
Based on my limited understanding of the JSON code and the contents of the test the fix looks correct. One small stylistic comment.

cpp/src/io/json/json_column.cu
karthikeyann: fixed!

cpp/src/io/json/json_column.cu
```diff
@@ -525,14 +525,15 @@ void make_device_json_column(device_span<SymbolT const> input,
...
       if (parent_node_id != parent_node_sentinel and node_categories[parent_node_id] == NC_LIST) {
         // unique item
-        if (i == 0 ||
+        if (i == 0 or
             (col_ids[i - 1] != col_ids[i] or ordered_parent_node_ids[i - 1] != parent_node_id)) {
           // scatter to list_offset
           d_columns_data[original_col_ids[parent_node_id]]
             .child_offsets[row_offsets[parent_node_id]] = ordered_row_offsets[i];
         }
         // TODO: verify if this code is right. check with more test cases.
```
bdice, line 534: Is this TODO still needed? Or does this PR cover enough test cases that we now trust its correctness?

bdice (approved):
Fix looks fine, and thanks for the new test. I have one question about removing a `TODO`.

cpp/src/io/json/json_column.cu
karthikeyann: will keep the TODO for now. 
This PR has test cases to test the conditions. But it needs a few more test cases to be thorough.

karthikeyann:
rerun tests

karthikeyann:
rerun tests

karthikeyann:
@gpucibot merge

PR #12070, opened 2022-11-04
Fix ingest_raw_data performance issue in Nested JSON reader due to RVO

karthikeyann:
## Description
Issue is that `json::experimental::ingest_raw_data` took double the time of `json::ingest_raw_data` for same data.

After replacing tertiary operator with `if` `else`, runtime for 500 MB file is same as `json::ingest_raw_data`
I suspect, RVO (copy elision) is skipped while using tertiary operator.

## Checklist

### DIFF
--- a/cpp/src/io/json/experimental/read_json.cpp
+++ b/cpp/src/io/json/experimental/read_json.cpp
@@ -19,27 +19,49 @@
 #include <io/comp/io_uncomp.hpp>
 #include <io/json/nested_json.hpp>
 
+#include <cudf/detail/nvtx/ranges.hpp>
 #include <cudf/utilities/error.hpp>
 
 #include <numeric>
 
 namespace cudf::io::detail::json::experimental {
 
-std::vector<uint8_t> ingest_raw_input(host_span<std::unique_ptr<datasource>> sources,
-                                      compression_type compression)
+size_t sources_size(host_span<std::unique_ptr<datasource>> const sources,
+                    size_t range_offset,
+                    size_t range_size)
 {
-  auto const total_source_size =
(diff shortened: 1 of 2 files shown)


cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -19,38 +19,64 @@
...
+      bytes_read += source->host_read(range_offset, data_size, destination);
+    }
   }
 
-  return (compression == compression_type::NONE) ? buffer : decompress(compression, buffer);
+  if (compression == compression_type::NONE) {
+    return buffer;
+  } else {
+    return decompress(compression, buffer);
+  }
```
ttnghia, line 64: Interesting, I also didn't know that ternary expression could prevent copy elision
Ref: https://stackoverflow.com/questions/22078029/why-does-the-ternary-operator-prevent-return-value-optimization

If possible, can you make a simple example in goldbolt (class printing to stdout when constructed/copied) to demonstrate that, please?

ttnghia (approved):

cpp/src/io/json/experimental/read_json.cpp
karthikeyann: https://godbolt.org/z/1vzPqqzs4
Here is a sample code. 
tertiary operator calls `vector(vector const&)` for buffer return
but `if() return buffer;` calls `vector(vector&&)`.

Reason for this could be - for tertial operator, one is lvalue (`buffer`), and the other is `rvalue` (return value of another function).
So, for lvalue, it makes call to copy constructor.
For Rvalue, it called move constructor.

elstehle (approved):

mythrocks (approved):
That's a brilliant catch.

karthikeyann:
@gpucibot merge

PR #12075, opened 2022-11-04
Fix singleton-range `__setitem__` edge case

### LINKED ISSUE #12073: [BUG] Missing data with length-one sliced-based `__setitem__` and singleton array
**Describe the bug**

Trying to set a length-one slice of a series with a length-one array (rather than a scalar) loses the value.

```python
import cudf
import cupy
s = cudf.Series([1, 2, 3], dtype=cupy.int64)
s.iloc[:1] = cupy.asarray([7], dtype=cupy.int64)
print(s)
# 0    1 <------ expecting this to be 7
# 1    2
# 2    3
# dtype: int64
```

**Expected behavior**

Setting a length-one range with a length-one array should work.

***Why does this happen?***

There's an off-by-one in copying.copy_range:

```python
    if abs(target_end - target_begin) <= 1: # should be `<`.
        return target_column
```

wence-:
## Description

When trying to set a length-one range with a length-one array, an off-by-one error in `copying.copy_range` meant that the value was discarded. Fix that, and tidy up the semantics of `copy_range` a little while we're here. Closes #12073.

## Checklist

### DIFF
--- a/python/cudf/cudf/_lib/copying.pyx
+++ b/python/cudf/cudf/_lib/copying.pyx
@@ -144,7 +144,7 @@ def copy_range(Column input_column,
     target_column from target_begin to target_end
     """
 
-    if abs(target_end - target_begin) <= 1:
+    if abs(target_end - target_begin) < 1:
         return target_column
 
     if target_begin < 0:
--- a/python/cudf/cudf/_lib/copying.pyx
+++ b/python/cudf/cudf/_lib/copying.pyx
@@ -132,36 +132,37 @@ def _copy_range(Column input_column,
     return Column.from_unique_ptr(move(c_result))
 
 
-def copy_range(Column input_column,
+def copy_range(Column source_column,
                Column target_column,
-               size_type input_begin,
-               size_type input_end,
+               size_type source_begin,
+               size_type source_end,
                size_type target_begin,
                size_type target_end,
                bool inplace):
     """
-    Copy input_column from input_begin to input_end to
-    target_column from target_begin to target_end
+    Copy a contiguous range from a source to a target column
+
+    Notes
+    -----
+    Expects the source and target ranges to have been sanitised to be
+    in-range for the source and target column respectively. For
+    example via ``slice.indices``.
     """
 
-    if abs(target_end - target_begin) < 1:
+    assert (
+        source_end - source_begin == target_end - target_begin,
+        "Source and target ranges must be same length"
+    )
+    if target_end >= target_begin and inplace:
+        # FIXME: Are we allowed to do this when inplace=False?
         return target_column
 
-    if target_begin < 0:
-        target_begin = target_begin + target_column.size
-
-    if target_end < 0:
-        target_end = target_end + target_column.size
-
-    if target_begin > target_end:
-        return target_column
-
-    if inplace is True:
-        _copy_range_in_place(input_column, target_column,
-                             input_begin, input_end, target_begin)
+    if inplace:
+        _copy_range_in_place(source_column, target_column,
+                             source_begin, source_end, target_begin)
     else:
-        return _copy_range(input_column, target_column,
-                           input_begin, input_end, target_begin)
+        return _copy_range(source_column, target_column,
+                           source_begin, source_end, target_begin)
 
 
 def gather(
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/_lib/copying.pyx
```diff
@@ -132,31 +132,32 @@ def _copy_range(Column input_column,
...
+        source_end - source_begin == target_end - target_begin,
+        "Source and target ranges must be same length"
+    )
+    if target_end >= target_begin and inplace:
+        # FIXME: Are we allowed to do this when inplace=False?
         return target_column
 
-    if inplace is True:
+    if inplace:
         _copy_range_in_place(input_column, target_column,
```
wence-, line 161: screwed up the renaming, sorry.

python/cudf/cudf/tests/test_setitem.py
```diff
@@ -108,6 +108,12 @@ def test_series_set_item(psr, arg):
     assert_eq(psr, gsr)
 
 
+def test_series_setitem_singleton_range():
+    sr = cudf.Series([1, 2, 3], dtype=np.int64)
+    sr.iloc[:1] = np.asarray([7], dtype=np.int64)
+    assert_eq(sr, cudf.Series([7, 2, 3], dtype=np.int64))
```
galipremsagar, line 114:
```suggestion
    sr = cudf.Series([1, 2, 3], dtype=np.int64)
    psr = sr.to_pandas()
    np_arr = np.asarray([7], dtype=np.int64)
    sr.iloc[:1] = np_arr
    psr.iloc[:1] = np_arr
    assert_eq(sr, psr)
```

galipremsagar: Better to have a pandas comparison in the test?

wence-: That just asserts we are bug-for-bug compatible with pandas though.

python/cudf/cudf/_lib/copying.pyx
```diff
@@ -132,36 +132,37 @@ def _copy_range(Column input_column,
...
+    example via ``slice.indices``.
+    """
 
-    if target_begin > target_end:
+    assert (
+        source_end - source_begin == target_end - target_begin,
+        "Source and target ranges must be same length"
+    )
+    if target_end >= target_begin and inplace:
+        # FIXME: Are we allowed to do this when inplace=False?
```
wence-, line 157: - [ ]TODO: address this comment.

wence-: I'm happy to do so by just opening an issue to track in service of trying to get this in.

python/cudf/cudf/tests/test_setitem.py
```diff
@@ -108,6 +108,16 @@ def test_series_set_item(psr, arg):
     assert_eq(psr, gsr)
 
 
+def test_series_setitem_singleton_range():
+    sr = cudf.Series([1, 2, 3], dtype=np.int64)
+    psr = sr.to_pandas()
+    value = np.asarray([7], dtype=np.int64)
+    sr.iloc[:1] = value
+    psr.iloc[:1] = value
+    assert_eq(sr, cudf.Series([7, 2, 3], dtype=np.int64))
```
galipremsagar, line 117:
```suggestion
```
This is now redundant, right?

wence-: Well, only if we believe pandas is bug-free (I suppose it probably is here).

galipremsagar: In that case we could check for dtype explicitly with `check_dtype=True`:
```python
assert_eq(sr, psr, check_dtype=True)
```

wence-: Done.

python/cudf/cudf/_lib/copying.pyx
galipremsagar: @wence- could you open a separate issue to fix this todo?

galipremsagar (approved):

galipremsagar:
@gpucibot merge

python/cudf/cudf/_lib/copying.pyx
wence-: https://github.com/rapidsai/cudf/issues/12135

PR #12077, opened 2022-11-04
Fix alignment of compressed blocks in ORC writer

### LINKED ISSUE #11812: [BUG] ORC ZLIB tests fail with nvCOMP 2.4
nvCOMP introduced changes to the DEFLATE compress/decompress (for example, decompress chunk size is limited to 64KB). ORC tests now fail, further triage is required to root cause.

vuule:
## Description
Closes #11812
Fixed alignment of compressed blocks in ORC writer - impacted ZLIB compression.
Re-enabled nvCOMP DEFLATE compression in ORC - nvCOMP 2.5+ only.

Refactored the nvCOMP feature status(enabled/disabled in cuDF) checks to include reason why features are not enabled (if not enabled).
Refactored call sites to return the detailed error message if an operation fails because of nvCOMP integration config.
Refactored nvCOMP adapter macros to allow mocking of the parameters that determine if an nvCOMP feature is enabled (env var, GPU compute capability, nvCOMP version).
Added tests to verify the logic of the newly refactored feature status checks (allowed by the mocking above).
Fix a Parquet test that was calling ORC reader/writer :grimacing: 

## Checklist


cpp/src/io/orc/writer_impl.cu
```diff
@@ -2187,9 +2188,9 @@ void writer::impl::write(table_view const& table)
     auto const max_compressed_block_size =
       max_compression_output_size(compression_kind_, compression_blocksize_);
     auto const padded_max_compressed_block_size =
-      util::round_up_unsafe<size_t>(max_compressed_block_size, uncomp_block_align);
+      util::round_up_unsafe<size_t>(max_compressed_block_size, compressed_block_align);
     auto const padded_block_header_size =
-      util::round_up_unsafe<size_t>(block_header_size, uncomp_block_align);
+      util::round_up_unsafe<size_t>(block_header_size, compressed_block_align);
```
vuule, line 2193: Actual fix for the alignment. Everything else is to safely re-enable DEFLATE compression.

cpp/tests/io/comp/decomp_test.cpp
```diff
@@ -166,4 +170,57 @@ TEST_F(BrotliDecompressTest, HelloWorld)
...
+  EXPECT_FALSE(comp_disabled(compression_type::DEFLATE, {2, 5, 0, true, true, 0}));
+  // version 2.5 required
+  EXPECT_TRUE(comp_disabled(compression_type::DEFLATE, {2, 4, 0, true, true, 0}));
+  // all integrations enabled required
+  EXPECT_TRUE(comp_disabled(compression_type::DEFLATE, {2, 5, 0, false, true, 0}));
+
+  EXPECT_FALSE(comp_disabled(compression_type::ZSTD, {2, 4, 0, true, true, 0}));
+  EXPECT_FALSE(comp_disabled(compression_type::ZSTD, {2, 4, 0, false, true, 0}));
+  // 2.3 version required
+  EXPECT_TRUE(comp_disabled(compression_type::ZSTD, {2, 2, 0, false, true, 0}));
```
jbrennan333, line 187: Shouldn't this be 2.4 required?

jbrennan333 (commented):
Looks good!  I just had a few minor comments.

cpp/tests/io/comp/decomp_test.cpp
```diff
@@ -166,4 +170,57 @@ TEST_F(BrotliDecompressTest, HelloWorld)
...
+  EXPECT_TRUE(decomp_disabled(compression_type::ZSTD, {2, 2, 0, true, true, 7}));
+  // stable integrations enabled required
+  EXPECT_TRUE(decomp_disabled(compression_type::ZSTD, {2, 4, 0, false, false, 7}));
+  // 2.4.0 disabled on Pascal
+  EXPECT_TRUE(decomp_disabled(compression_type::ZSTD, {2, 4, 0, true, true, 6}));
+
+  EXPECT_FALSE(decomp_disabled(compression_type::SNAPPY, {2, 4, 0, true, true, 7}));
+  EXPECT_FALSE(decomp_disabled(compression_type::SNAPPY, {2, 3, 0, false, true, 7}));
+  // stable integrations enabled required
+  EXPECT_TRUE(decomp_disabled(compression_type::SNAPPY, {2, 2, 0, false, false, 7}));
```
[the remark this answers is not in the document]
vuule: Thank you for the review! Added a check, is that what you had in mind?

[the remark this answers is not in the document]
jbrennan333: Yes.  Thanks for adding it.

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -390,18 +359,109 @@ void batched_compress(compression_type compression,
...
   switch (compression) {
-    case compression_type::DEFLATE:
-      // See https://github.com/rapidsai/cudf/issues/11812
-      return false;
-    case compression_type::SNAPPY: return detail::nvcomp_integration::is_stable_enabled();
-    case compression_type::ZSTD:
-      return NVCOMP_HAS_ZSTD_COMP and detail::nvcomp_integration::is_stable_enabled();
-    default: return false;
+    case compression_type::DEFLATE: {
+      if (not NVCOMP_HAS_DEFLATE(
```
bdice, line 380: I suppose this expands to a valid runtime expression as well as being useful at compile time in `#if` macros? Not super familiar with using macros like this.

cpp/src/io/comp/nvcomp_adapter.hpp
```diff
@@ -30,14 +32,52 @@ namespace cudf::io::nvcomp {
...
+  int lib_major_version;
+  int lib_minor_version;
+  int lib_patch_version;
+  bool are_all_integrations_enabled;
+  bool are_stable_integrations_enabled;
+  int compute_capability;
+
+  feature_status_parameters();
+  feature_status_parameters(
+    int major, int minor, int patch, bool all_enabled, bool stable_enabled, int cc)
```
bdice, line 47: Compute capabilities are not really integral, there are lots of values like `8.6` or `6.1`. Does this get some kind of normalized integral value like `86`? How does this interact with `params.compute_capability == 6`?

bdice (commented):
I'm glad we have tests now. Do we actually cover a good portion of the version/hardware cases in our testing (manual or CI)?

cpp/src/io/comp/nvcomp_adapter.cpp
hyperbolic2346: It is simply a compile time text substitution, so this one will translate to `if (not (MAJOR > 2 or (MAJOR == 2 and MINOR >= 5)))`

vuule:
> Do we actually cover a good portion of the version/hardware cases in our testing (manual or CI)?

I don't know about HW, but we definitely only cover a single nvCOMP version.

cpp/src/io/comp/nvcomp_adapter.hpp
vuule: Looks like my naming is bad here. 
This is just the major capability number.
`cudaDeviceGetAttribute(&compute_capability, cudaDevAttrComputeCapabilityMajor, device))` returns an int so this should be fine. 
I'll fix the name to avoid confusion.

cpp/src/io/comp/nvcomp_adapter.cpp
vuule: > Not super familiar with using macros like this.

Yeah, this is tech from the dark ages, thankfully not commonly used any more. I unfortunately could not avoid it here :(

I think you have the right idea, this is pretty much a type-unsafe constexpr function that can be used in `#if` conditions. By the means of text substitution, as Mike said :D

cpp/src/io/comp/nvcomp_adapter.hpp
vuule: Renamed to reflect that this is the `major` component of compute capability.

bdice (approved):

vuule:
rerun tests

vuule:
@gpucibot merge

PR #12079, opened 2022-11-04
Workaround groupby aggregate thrust::copy_if overflow

### LINKED ISSUE #12058: [BUG] size_type overflow in `cudf::groupby::detail::hash::extract_populated_keys`
We look to be overflowing `size_type` while in `extract_populated_keys`. The iterator on `rmm::device_uvector` is signed so I think we need to go to unsigned in order to be able to group by up to the 2B row limit. 

I believe the issue is actually when calling `populated_keys.resize` where the `std::distance` may be negative, but the stack trace I got (on a non-debug build) points at `copy_if`. Either way it definitely looks to be an issue triggered in this function.

```
ai.rapids.cudf.Rmm.initialize(0, null, 0)

// create a vector with 1.2B rows with distinct keys
val cvs = new scala.collection.mutable.ArrayBuffer[ai.rapids.cudf.ColumnVector]()
cvs += ai.rapids.cudf.ColumnVector.fromInts((0 until 1200000).toArray:_*)
(1 until 1000).foreach(ix => cvs += cvs(ix-1).add(ai.rapids.cudf.Scalar.fromInt(1200000)))
val concat = ai.rapids.cudf.ColumnVector.concatenate(cvs:_*)
cvs.foreach(_.close)
cvs.clear()

// group by count on column 0
val tbl = new ai.rapids.cudf.Table(concat)
val agg =tbl.groupBy(0).aggregate(ai.rapids.cudf.GroupByAggregation.sum().onColumn(0))
```

C++ repro:

```
#include <cudf/types.hpp>
#include <cudf/table/table.hpp>
#include <cudf/aggregation.hpp>
#include <cudf/groupby.hpp>

  auto values = thrust::make_counting_iterator(0);
  std::size_t num_rows = 1200000000L;
  auto columns = std::vector<std::unique_ptr<cudf::column>>();
  columns.push_back(cudf::make_fixed_width_column(cudf::data_type{cudf::type_id::INT32}, num_rows));
  thrust::copy(rmm::exec_policy(rmm::cuda_stream_default), values, values + num_rows, columns[0]->mutable_view().begin<uint32_t>());
  auto tbl = cudf::table(std::move(columns));

  cudf::groupby::groupby grouper(tbl, 
    cudf::null_policy::INCLUDE, 
    cudf::sorted::NO, 
    std::vector<cudf::order>{}, 
    std::vector<cudf::null_order>{});

  std::vector<cudf::groupby::aggregation_request> requests;
  std::vector<std::unique_ptr<cudf::groupby_aggregation>> aggs;
  aggs.push_back(cudf::make_sum_aggregation<cudf::groupby_aggregation>());
  requests.emplace_back(cudf::groupby::aggregation_request { tbl.get_column(0), std::move(aggs) });

  auto result = grouper.aggregate(requests);
}
```

Stack trace:

```
java.lang.OutOfMemoryError: Could not allocate native memory: std::bad_alloc: out_of_memory: std::bad_alloc: out_of_memory: CUDA error at: include/rmm/mr/device/cuda_memory_resource.hpp:70: cudaErrorMemoryAllocation out of memory. Attempted size: 18446744073696393216 attempted at: #0 in rmm::detail::stack_trace::stack_trace() from /tmp/cudf5256769593043486578.so
#1 in /tmp/cudf5256769593043486578.so(+0x1170e6a) [0x7f73c1cbfe6a]
#2 in int* thrust::cuda_cub::copy_if<thrust::detail::execute_with_allocator<rmm::mr::thrust_allocator<char>, thrust::cuda_cub::execute_on_stream_base>, thrust::transform_iterator<__nv_dl_wrapper_t<__nv_dl_tag<rmm::device_uvector<int> (*)(concurrent_unordered_map<int, int, cudf::experimental::row::hash::device_row_hasher<cudf::detail::MurmurHash3_32, cudf::nullate::DYNAMIC>, cudf::experimental::row::equality::device_row_comparator<cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>, default_allocator<thrust::pair<int, int> > > const&, int, rmm::cuda_stream_view), &cudf::groupby::detail::hash::(anonymous namespace)::extract_populated_keys, 1u>>, thrust::pair<int, int>*, thrust::use_default, thrust::use_default>, int*, __nv_dl_wrapper_t<__nv_dl_tag<rmm::device_uvector<int> (*)(concurrent_unordered_map<int, int, cudf::experimental::row::hash::device_row_hasher<cudf::detail::MurmurHash3_32, cudf::nullate::DYNAMIC>, cudf::experimental::row::equality::device_row_comparator<cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>, default_allocator<thrust::pair<int, int> > > const&, int, rmm::cuda_stream_view), &cudf::groupby::detail::hash::(anonymous namespace)::extract_populated_keys, 2u>, int> >(thrust::cuda_cub::execution_policy<thrust::detail::execute_with_allocator<rmm::mr::thrust_allocator<char>, thrust::cuda_cub::execute_on_stream_base> >&, thrust::transform_iterator<__nv_dl_wrapper_t<__nv_dl_tag<rmm::device_uvector<int> (*)(concurrent_unordered_map<int, int, cudf::experimental::row::hash::device_row_hasher<cudf::detail::MurmurHash3_32, cudf::nullate::DYNAMIC>, cudf::experimental::row::equality::device_row_comparator<cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>, default_allocator<thrust::pair<int, int> > > const&, int, rmm::cuda_stream_view), &cudf::groupby::detail::hash::(anonymous namespace)::extract_populated_keys, 1u>>, thrust::pair<int, int>*, thrust::use_default, thrust::use_default>, thrust::transform_iterator<__nv_dl_wrapper_t<__nv_dl_tag<rmm::device_uvector<int> (*)(concurrent_unordered_map<int, int, cudf::experimental::row::hash::device_row_hasher<cudf::detail::MurmurHash3_32, cudf::nullate::DYNAMIC>, cudf::experimental::row::equality::device_row_comparator<cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>, default_allocator<thrust::pair<int, int> > > const&, int, rmm::cuda_stream_view), &cudf::groupby::detail::hash::(anonymous namespace)::extract_populated_keys, 1u>>, thrust::pair<int, int>*, thrust::use_default, thrust::use_default>, int*, __nv_dl_wrapper_t<__nv_dl_tag<rmm::device_uvector<int> (*)(concurrent_unordered_map<int, int, cudf::experimental::row::hash::device_row_hasher<cudf::detail::MurmurHash3_32, cudf::nullate::DYNAMIC>, cudf::experimental::row::equality::device_row_comparator<cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>, default_allocator<thrust::pair<int, int> > > const&, int, rmm::cuda_stream_view), &cudf::groupby::detail::hash::(anonymous namespace)::extract_populated_keys, 2u>, int>) from /tmp/cudf5256769593043486578.so
#3 in cudf::groupby::detail::hash::(anonymous namespace)::extract_populated_keys(concurrent_unordered_map<int, int, cudf::experimental::row::hash::device_row_hasher<cudf::detail::MurmurHash3_32, cudf::nullate::DYNAMIC>, cudf::experimental::row::equality::device_row_comparator<cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>, default_allocator<thrust::pair<int, int> > > const&, int, rmm::cuda_stream_view) from /tmp/cudf5256769593043486578.so
#4 in cudf::groupby::detail::hash::(anonymous namespace)::groupby(cudf::table_view const&, cudf::host_span<cudf::groupby::aggregation_request const, 18446744073709551615ul>, cudf::detail::result_cache*, bool, cudf::null_policy, rmm::cuda_stream_view, rmm::mr::device_memory_resource*) from /tmp/cudf5256769593043486578.so
#5 in cudf::groupby::detail::hash::groupby(cudf::table_view const&, cudf::host_span<cudf::groupby::aggregation_request const, 18446744073709551615ul>, cudf::null_policy, rmm::cuda_stream_view, rmm::mr::device_memory_resource*) from /tmp/cudf5256769593043486578.so
#6 in cudf::groupby::groupby::dispatch_aggregation(cudf::host_span<cudf::groupby::aggregation_request const, 18446744073709551615ul>, rmm::cuda_stream_view, rmm::mr::device_memory_resource*) from /tmp/cudf5256769593043486578.so
#7 in cudf::groupby::groupby::aggregate(cudf::host_span<cudf::groupby::aggregation_request const, 18446744073709551615ul>, rmm::mr::device_memory_resource*) from /tmp/cudf5256769593043486578.so
#8 in Java_ai_rapids_cudf_Table_groupByAggregate from /tmp/cudf5256769593043486578.so
#9 in [0x7f796d017de7]

  at ai.rapids.cudf.Table.groupByAggregate(Native Method)
  at ai.rapids.cudf.Table.access$3000(Table.java:41)
  at ai.rapids.cudf.Table$GroupByOperation.aggregate(Table.java:3657)
  ... 47 elided
```

davidwendt:
## Description
Workaround for limitation in `thrust::copy_if` which fails if the input-iterator spans more than int-max.
The `thrust::copy_if` hardcodes the iterator distance type to be an int
https://github.com/NVIDIA/thrust/blob/dbd144ed543b60c4ff9d456edd19869e82fe8873/thrust/system/cuda/detail/copy_if.h#L699-L708

This calls the `copy_if` in chunks if the iterator can span greater than int-max.

Closes #12058 

## Checklist

### DIFF
--- a/cpp/src/groupby/hash/groupby.cu
+++ b/cpp/src/groupby/hash/groupby.cu
@@ -512,18 +512,29 @@ rmm::device_uvector<size_type> extract_populated_keys(map_type const& map,
 {
   rmm::device_uvector<size_type> populated_keys(num_keys, stream);
 
-  auto get_key    = [] __device__(auto const& element) { return element.first; };  // first = key
-  auto get_key_it = thrust::make_transform_iterator(map.data(), get_key);
-  auto key_used   = [unused = map.get_unused_key()] __device__(auto key) { return key != unused; };
+  auto get_key  = [] __device__(auto const& element) { return element.first; };  // first = key
+  auto key_itr  = thrust::make_transform_iterator(map.data(), get_key);
+  auto key_used = [unused = map.get_unused_key()] __device__(auto key) { return key != unused; };
 
-  auto end_it = thrust::copy_if(rmm::exec_policy(stream),
-                                get_key_it,
-                                get_key_it + map.capacity(),
-                                populated_keys.begin(),
-                                key_used);
+  // thrust::copy_if has a bug where it cannot iterate over int-max values
+  // so if map.capacity() > int-max we'll call thrust::copy_if in chunks instead
+  auto const map_size =
+    std::min(map.capacity(), static_cast<std::size_t>(std::numeric_limits<size_type>::max()));
+  auto key_end      = key_itr + map.capacity();
+  auto pop_keys_itr = populated_keys.begin();
 
-  populated_keys.resize(std::distance(populated_keys.begin(), end_it), stream);
+  std::size_t output_size = 0;
+  while (key_itr < key_end) {
+    auto copy_end = key_itr + map_size < key_end ? key_itr + map_size : key_end;
+    auto end_it =
+      thrust::copy_if(rmm::exec_policy(stream), key_itr, copy_end, pop_keys_itr, key_used);
+    auto copied = std::distance(pop_keys_itr, end_it);
+    pop_keys_itr += copied;
+    output_size += copied;
+    key_itr = copy_end;
+  }
 
+  populated_keys.resize(output_size, stream);
   return populated_keys;
 }
 


cpp/src/groupby/hash/groupby.cu
```diff
@@ -512,18 +512,29 @@ rmm::device_uvector<size_type> extract_populated_keys(map_type const& map,
...
-                                key_used);
-
-  populated_keys.resize(std::distance(populated_keys.begin(), end_it), stream);
+  auto get_key  = [] __device__(auto const& element) { return element.first; };  // first = key
+  auto key_itr  = thrust::make_transform_iterator(map.data(), get_key);
+  auto key_used = [unused = map.get_unused_key()] __device__(auto key) { return key != unused; };
+
+  // thrust::copy_if has a bug where it cannot iterate over int-max values
+  // so if map.capacity() > int-max we'll call thrust::copy_if in chunks instead
+  auto const map_size =
```
abellina, line 521: nit, could this be changed to something like "copy_size"? "map_size" seems to indicate that it is the size of the actual map.

davidwendt: It is the size of the map. We are iterating over the entire map to copy out the keys.

cpp/src/groupby/hash/groupby.cu
```diff
@@ -512,18 +512,29 @@ rmm::device_uvector<size_type> extract_populated_keys(map_type const& map,
...
-  auto end_it = thrust::copy_if(rmm::exec_policy(stream),
-                                get_key_it,
-                                get_key_it + map.capacity(),
-                                populated_keys.begin(),
-                                key_used);
-
-  populated_keys.resize(std::distance(populated_keys.begin(), end_it), stream);
+  auto get_key  = [] __device__(auto const& element) { return element.first; };  // first = key
+  auto key_itr  = thrust::make_transform_iterator(map.data(), get_key);
+  auto key_used = [unused = map.get_unused_key()] __device__(auto key) { return key != unused; };
```
ttnghia, line 517: `auto const` if possible.

cpp/src/groupby/hash/groupby.cu
```diff
@@ -512,18 +512,29 @@ rmm::device_uvector<size_type> extract_populated_keys(map_type const& map,
...
+  // thrust::copy_if has a bug where it cannot iterate over int-max values
+  // so if map.capacity() > int-max we'll call thrust::copy_if in chunks instead
+  auto const map_size =
+    std::min(map.capacity(), static_cast<std::size_t>(std::numeric_limits<size_type>::max()));
+  auto key_end      = key_itr + map.capacity();
+  auto pop_keys_itr = populated_keys.begin();
+
+  std::size_t output_size = 0;
+  while (key_itr < key_end) {
+    auto copy_end = key_itr + map_size < key_end ? key_itr + map_size : key_end;
```
ttnghia, line 528: I think we should not compare `<` for the iterator. Just `==` or `!=`, so we should use something like `thrust::distance(key_itr, key_end) > map_size`.

cpp/src/groupby/hash/groupby.cu
davidwendt: I'm waiting for @abellina to verify this works before putting this up for review.

cpp/src/groupby/hash/groupby.cu
abellina: I agree that this is the map size in the regular scenario, where `map.capacity()` is less than or equal to `std::numeric_limits<size_type>::max()`, but in the case where it is larger, it is the the chunk of the map we are going to copy, not the whole size.

abellina (approved):
This LGTM, I will be running this today (building things) but I do not think my run is a blocker.

abellina:
Update: I ran with this patch on an 80GB GPU, and while on the aggregate I see a `cudaErrorIllegalAddress`. I am in the precess of running under `compute-sanitizer`.

abellina:
OK, @davidwendt by itself this seems to be working fine in the 80GB gpu. I believe I am hitting a second issue that is not related to this one for the join that comes later. I see issues with `gather` that will have to be investigated separately. +1 still

cpp/src/groupby/hash/groupby.cu
```diff
@@ -512,18 +512,31 @@ rmm::device_uvector<size_type> extract_populated_keys(map_type const& map,
...
+  // thrust::copy_if has a bug where it cannot iterate over int-max values
+  // so if map.capacity() > int-max we'll call thrust::copy_if in chunks instead
+  auto const map_size =
+    std::min(map.capacity(), static_cast<std::size_t>(std::numeric_limits<size_type>::max()));
+  auto const key_end = key_itr + map.capacity();
+  auto pop_keys_itr  = populated_keys.begin();
+
+  std::size_t output_size = 0;
+  while (key_itr != key_end) {
+    auto const copy_end = static_cast<std::size_t>(std::distance(key_itr, key_end)) <= map_size
```
bdice, line 528: Does this need `thrust::distance` instead of `std::distance`? It looks like it is working on Thrust iterators but I don’t know if this matters.

bdice: Same for `copied` below.

bdice (commented):
Could this bug manifest in other places `thrust::copy_if` is called?

cpp/src/groupby/hash/groupby.cu
davidwendt: We prefer `std` in CPU code. I try to use `thrust::distance` only in device code. The `std::distance()` does work correctly here and since it does not need to call any device code.

PointKernel:
Do we have a test exercising this overflow issue?
This algorithm will be refactored by using cuco hash maps and `thrust::copy_if` may not be needed at all. I just want to make sure that we have tests covering this thus we won't reintroduce the same issue into the code during refactoring.

davidwendt:
> Do we have a test exercising this overflow issue? This algorithm will be refactored by using cuco hash maps and `thrust::copy_if` may not be needed at all. I just want to make sure that we have tests covering this thus we won't reintroduce the same issue into the code during refactoring.

A test would not be practical because it would require too much memory for the CI machines. Even with the fix, this runs out of memory on my 48GB GPU. This is really a thrust bug so I feel like the test should go there.

cpp/src/groupby/hash/groupby.cu
```diff
@@ -512,18 +512,31 @@ rmm::device_uvector<size_type> extract_populated_keys(map_type const& map,
...
-
-  populated_keys.resize(std::distance(populated_keys.begin(), end_it), stream);
+  auto get_key  = [] __device__(auto const& element) { return element.first; };  // first = key
+  auto key_itr  = thrust::make_transform_iterator(map.data(), get_key);
+  auto key_used = [unused = map.get_unused_key()] __device__(auto key) { return key != unused; };
+
+  // thrust::copy_if has a bug where it cannot iterate over int-max values
+  // so if map.capacity() > int-max we'll call thrust::copy_if in chunks instead
+  auto const map_size =
+    std::min(map.capacity(), static_cast<std::size_t>(std::numeric_limits<size_type>::max()));
```
robertmaynard, line 522: I would prefer this to be limits<int> instead as that makes very clear what our max value would be

cpp/src/groupby/hash/groupby.cu
robertmaynard: I agree that the name should coney this is the max length we can pass to thrust::copy_if

robertmaynard (approved):

cpp/src/groupby/hash/groupby.cu
ttnghia: Remind again for this.

davidwendt:
@gpucibot merge

PR #12080, opened 2022-11-05
Throw an error when libcudf is built without cuFile and `LIBCUDF_CUFILE_POLICY` is set to `"ALWAYS"`

vuule:
## Description
Currently, creating a cufile `datasource` or `data_sink` silently fails if libcudf was built without the cuFile headers. This is expected behavior when the `LIBCUDF_CUFILE_POLICY` is not set, or is set to and value other than "ALWAYS". However, with "ALWAYS", there should be no fallback from GDS.
This PR adds a check to fail loudly when `LIBCUDF_CUFILE_POLICY=="ALWAYS"` cannot be enforced because of missing dependency (cuFile).

## Checklist

### DIFF
--- a/cpp/src/io/utilities/file_io_utilities.cpp
+++ b/cpp/src/io/utilities/file_io_utilities.cpp
@@ -269,6 +269,10 @@ std::unique_ptr<cufile_input_impl> make_cufile_input(std::string const& filepath
       if (cufile_integration::is_always_enabled()) throw;
     }
   }
+#else
+  CUDF_EXPECTS(not cufile_integration::is_always_enabled(),
+               "Unable to enforce the use of GDS requested through the LIBCUDF_CUFILE_POLICY "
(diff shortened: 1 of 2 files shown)


cpp/src/io/utilities/file_io_utilities.cpp
```diff
@@ -269,6 +269,10 @@ std::unique_ptr<cufile_input_impl> make_cufile_input(std::string const& filepath
       if (cufile_integration::is_always_enabled()) throw;
     }
   }
+#else
```
davidwendt, line 272: Is the try-catch still required? 
Perhaps whatever is thrown in `std::make_unique<cufile_input_impl>(filepath)` should just pass through to the caller.

davidwendt (approved):

hyperbolic2346 (approved):

hyperbolic2346:
build

vuule:
@gpucibot merge

Issue #11943, opened 2022-10-18
[FEA] Ensure that streams passed to libcudf APIs are properly forwarded through


vyasr:
**Is your feature request related to a problem? Please describe.**
libcudf currently runs entirely on CUDA's default stream. Internally, this is managed using an `rmm::cuda_stream_view` stored as a global variable used everywhere. We would like to move towards enabling stream-ordered libcudf APIs. In order to enable this safely, we need to ensure that any API that accepts a stream correctly forwards this stream down rather than passing `cudf::default_stream_value` through (or anything else like `rmm::cuda_stream_default`. 

**Describe the solution you'd like**
I would like to create and maintain a new feature branch of libcudf. On this branch, we can converting public libcudf APIs to take stream arguments a bit at a time. We should leverage the library written for #11942 to verify that the new streams are correctly being passed all the way down the call stack. To do so, requires the following:
1. Define a new function `cudf::test::get_default_stream` that returns `cudf::get_default_stream` and is used everywhere in the tests where a stream argument is required.
2. Include an overload of `cudf::test::get_default_stream` in `identify_stream_usage` that, instead of returning `cudf::get_default_stream`, will return a different stream (can be easily managed using a function-local static).
3. Modify `identify_stream_usage` so that instead of throwing an error when the default stream is observed, it throws an error whenever anything other than `cudf::test::get_default_stream()` is observed. This feature should be set as a compile-time flag so that we can compile `identify_stream_usage` twice, once where it expects to not see any of CUDA's default streams and once where it expects to only see `cudf::test::get_default_stream()`.

Once these steps are completed, we could work towards exposing streams by transitioning APIs one at a time on a feature branch. This approach requires maintaining an ongoing feature branch, but it saves us the need to do multiple rounds of converting tests to use detail functions and then back. It also allows us to accomplish even more precisely what #11942 intends since we can directly check for the usage of the desired stream rather than the current roundabout approach of checking for the default stream.

**Describe alternatives you've considered**
I considered trying to find some way to overload all cudf API calls in the tests to instead use some sort of patched version. The biggest issue here (aside from the technical challenges of finding a way to do it) is the fact that not every single cudf API actually accepts a stream, so the logic would need to be sophisticated enough to check the corresponding APIs. I do not think that level of complexity is desirable or warranted when the above solution would be fairly easy to maintain going forward.

**Additional context**
[...]

PR #11875, opened 2022-10-07
Add tests ensuring that cudf's default stream is always used

vyasr:
## Description

This PR ensures that cudf's default stream is properly passed to all kernel launches so that nothing implicitly runs on the CUDA default stream. It adds a small library that is built during the tests and overloads CUDA functions to throw an exception when usage of the default stream is detected. It also fixes all remaining usage of anything other than cudf's default stream (I fixed most of the issues in previous PRs, but I found a few others when finalizing this one).

Resolves #11929
Resolves #11942 

### Important notes for reviewers:
- **The changeset is deceptively large.** The vast majority of the changes are just a global find-and-replace of `cudf::get_default_stream()` for `cudf::default_stream_value`, as well as a few smaller fixes such as missing `CUDF_TEST_PROGRAM_MAIN` in a couple of tests and usage of `rmm::cuda_stream_default`. The meaningful changes are:
    - The new default stream getter/setter in `default_stream.[hpp|cpp]`
    - The addition of `cpp/tests/utilities/identify_stream_usage`
    - The changes to the base testing fixture in `cpp/include/cudf_test/base_fixture.hpp` to inject the custom stream.
    - The changes to CI in `ci/gpu/build.sh` to build and use the new library.
- This PR is a breaking change because it moves the default stream into the detail namespace. Going forward the default stream may only be accessed using the public accessor `cudf::get_default_stream()`. I have added a corresponding setter, but it is also in the detail namespace since I do not want to publicly support changing the default stream yet, only for the purpose of testing. Reviewers, please leave comments if you disagree with those choices.
- I have made getting and setting the default stream thread-safe, but there is still only a single stream. In multi-threaded applications we may want to support a stream per thread so that users could manually achieve PTDS with more fine-tuned control. Is this worthwhile? Even if it is, I'm inclined to wait for a subsequent PR to implement this unless someone feels strongly otherwise.
- I'm currently only overloading `cudaLaunchKernel`. I can add overloads for other functions as well, but I didn't want to go through the effort of overloading every possible API. If reviewers have a minimal set that they'd like to see overloaded, let me know. [I've included links to all the relevant pages of the CUDA runtime API in the identify_stream_usage.cu file](https://github.com/rapidsai/cudf/pull/11875/files#diff-0b2762207c27c080acd2114475c7a1c06377a7c18c4e9c3de60ecbdc82a4dc61R99) if someone wants to look through them.

## Checklist
[...]

PR #12089, opened 2022-11-08
Enable detection of undesired stream usage

vyasr:
## Description

This PR builds on #11875 and partially addresses #11943. This PR allows us to run all tests on a precise stream (the newly introduced `cudf::test::get_default_stream()`) and then verify that all CUDA APIs end up invoked on that stream. This implements the feature required in #11943, but to apply it universally across libcudf will require the API changes that will expose streams so I plan to make those changes incrementally after this PR is merged.

The preload library is now compiled twice, once to overload `cudf::get_default_stream` and once to overload `cudf::test::get_default_stream`. For now there is still some manual coordination associated with determining which one should be used with a given test, but once #12451 is merged and we start running all tests via ctest instead of direct invocation of the test executables we can start encoding this information in the CMake configuration of the tests by associating the require environment variables directly with the test executable using `set_tests_properties`.

## Checklist


vyasr:
An example of how this will be used can be seen in #12090. Interested reviewers can diff that branch against this one (for now it just reworks a single API and is very minimal), pull it down locally, and try to run the tests.

java/src/main/native/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2019-2022, NVIDIA CORPORATION.
+# Copyright (c) 2019-2023, NVIDIA CORPORATION.
```
jlowe, line 2: Copyright update with no other changes to the file?

vyasr: Hmm must have been a pre-commit hook, looks like there were a few. Let me revert.

vyasr: Removed, thanks.

cpp/CMakeLists.txt
```diff
@@ -731,10 +731,11 @@ add_library(cudf::cudf ALIAS cudf)
 
 if(CUDF_BUILD_TESTUTIL)
   add_library(
-    cudftestutil STATIC
+    cudftestutil SHARED
```
robertmaynard, line 734: I am curious on why we changed this to shared

vyasr: I had a TODO comment to make sure reviewers were OK with this change earlier and then removed it at some point since I didn't see an alternative. We need this to be a shared library in order to be able to overload the symbols in a preload library. Specifically, we need to be able to redefine `cudf::test::get_default_stream` at load time.

vyasr: I guess the original question does still apply though, are we OK with this cost to the change?

robertmaynard: I don't think it will be an issue to switch this to a shared library.

robertmaynard (approved):

cpp/CMakeLists.txt
vyasr: Added a comment explaining why this is a dynamic lib in case someone wonders why this is a static lib in the future.

raydouglass (approved):

cpp/include/cudf_test/default_stream.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace cudf {
+namespace test {
+
+/**
+ * @brief Get the default stream to use for tests.
+ *
+ * @return The default stream to use for tests.
+ */
+rmm::cuda_stream_view const get_default_stream();
```
ttnghia, line 29: Sorry I don't get this. Why do we need this alias function to `cudf::get_default_stream()`? If we indeed need it, please add more comment to clarify.

cpp/include/cudf_test/stream_checking_resource_adaptor.hpp
```diff
@@ -131,25 +138,44 @@ class stream_checking_resource_adaptor final : public rmm::mr::device_memory_res
...
     }
   }
 
-  Upstream* upstream_;  // the upstream resource used for satisfying allocation requests
+  Upstream* upstream_;            // the upstream resource used for satisfying allocation requests
+  bool error_on_invalid_stream_;  // If true, throw an exception when the wrong stream is detected.
+                                  // If false, simply print to stdout.
+  bool check_default_stream_;  // If true, throw an exception when the default stream is observed.
+                               // If false, throw an exception when anything other than
+                               // cudf::test::get_default_stream() is observed.
```
ttnghia, line 178: Our convention (and common usage in libcudf) is using an underscore as prefix, not suffix, for class private members.
```suggestion
  Upstream* _upstream;            // the upstream resource used for satisfying allocation requests
  bool _error_on_invalid_stream;  // If true, throw an exception when the wrong stream is detected.
                                  // If false, simply print to stdout.
  bool _check_default_stream;  // If true, throw an exception when the default stream is observed.
                               // If false, throw an exception when anything other than
                               // cudf::test::get_default_stream() is observed.
```

cpp/include/cudf_test/stream_checking_resource_adaptor.hpp
```diff
@@ -33,7 +35,12 @@ class stream_checking_resource_adaptor final : public rmm::mr::device_memory_res
    *
    * @param upstream The resource used for allocating/deallocating device memory
    */
-  stream_checking_resource_adaptor(Upstream* upstream) : upstream_{upstream}
+  stream_checking_resource_adaptor(Upstream* upstream,
+                                   bool error_on_invalid_stream,
+                                   bool check_default_stream)
+    : upstream_{upstream},
```
ttnghia, line 41: Why do we still use raw pointer? We should favor smart pointer, shouldn't we?

ttnghia: If we cannot claim ownership of the pointer but have to use raw pointer, please add `CUDF_EXPECT` to check for nullptr.

cpp/tests/utilities/default_stream.cpp
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+#include <cudf/utilities/default_stream.hpp>
+
+#include <cudf_test/default_stream.hpp>
+
+namespace cudf {
+namespace test {
+
+rmm::cuda_stream_view const get_default_stream() { return cudf::get_default_stream(); }
```
ttnghia, line 24: Will this cause double definition of `cudf::test::get_default_stream()`? This is defined here, and also defined again in `identify_stream_usage.cpp`.

ci/test_cpp.sh
```diff
@@ -8,9 +8,10 @@ trap "EXITCODE=1" ERR
 set +e
 
 # Get library for finding incorrect default stream usage.
-STREAM_IDENTIFY_LIB="${CONDA_PREFIX}/lib/libcudf_identify_stream_usage.so"
+STREAM_IDENTIFY_LIB_MODE_CUDF="${CONDA_PREFIX}/lib/libcudf_identify_stream_usage_mode_cudf.so"
+STREAM_IDENTIFY_LIB_MODE_TESTING="${CONDA_PREFIX}/lib/libcudf_identify_stream_usage_mode_testing.so"
```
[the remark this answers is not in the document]
vyasr: Not in this PR, but it will be used in immediate follow-ups like #12090 and it's conceptually easier to explain why it's necessary as part of this PR because this PR is where the library has been modified to be compiled twice.

cpp/include/cudf_test/default_stream.hpp
vyasr: No worries, this is a good question and the answer is not obvious. The purpose of this function is to provide a symbol that may be overloaded with a preloaded library, which is what all the stream verification is designed to leverage. The idea is that if all tests pass `cudf::test::get_default_stream` as an argument to functions that accept a stream, we can overload the behavior of the function at runtime to provide a specific stream and then track all CUDA API calls to ensure that everything runs on that new stream instead of on e.g. the CUDA default stream (which would indicate that somewhere we are failing to pass the stream all the way down the call stack).

cpp/include/cudf_test/stream_checking_resource_adaptor.hpp
vyasr: That is true, but I chose to do it this way to follow the conventions of all rmm adaptors that I have seen (such as [the aligned adaptor](https://github.com/rapidsai/rmm/blob/branch-23.04/include/rmm/mr/device/aligned_resource_adaptor.hpp#L210) or [the logging adaptor](https://github.com/rapidsai/rmm/blob/branch-23.04/include/rmm/mr/device/logging_resource_adaptor.hpp#L304)). I can change it if you feel strongly about it, but I have a weak preference to leave it as is.

cpp/include/cudf_test/stream_checking_resource_adaptor.hpp
vyasr: Smart pointers imply information about ownership semantics. When you want to be able to work with a pointer to some object regardless of who owns it, it's pretty normal to pass around raw pointers. One could argue that a reference would be better here to ensure that the parameter passed is non-null, but again I chose to follow the pattern used for other rmm adaptors. @jrhemstad or @harrism might have more insight into why a pointer was preferred to a reference here, I wouldn't expect adaptors to perform any of the types of modifications of the upstream where a pointer is preferred but I could be wrong.

cpp/tests/utilities/default_stream.cpp
vyasr: It would if they were both compiled into the same library/executable, but the purpose of these is explicitly _not_ to be put into the same TU. The idea is that this file is compiled as part of libcudf's test utils, so it is the "official" definition of `cudf::test::get_default_stream`. When we want to ensure that all libcudf APIs are properly passing streams to everything they call (e.g. if I call `cudf::foo(..., stream)`, internally we want to make sure that `cudf::foo` calls `cudf::detail::foo(..., stream)` and doesn't forget to pass the stream through), we will run tests as `LD_PRELOAD=/path/to/identify_stream_usage TEST_EXECUTABLE`, which will result in the definition of `cudf::test::get_default_stream` in `identify_stream_usage` being used instead of the definition in `default_stream`. The definition in `identify_stream_usage` uses a custom stream defined as a function-local static, so we can then overload all CUDA APIs (like `cudaLaunchKernel`) to make sure that the stream that `cudaLaunchKernel` was called with is the same stream as `cudf::test::get_default_stream`. If it isn't, that implies that somewhere in libcudf we accidentally call a function without a stream (or we explicitly pass `cudf::get_default_stream`) instead of correctly forwarding the user-provided stream through.

ttnghia: Humnn, this is the first time I know about `LD_PRELOAD`. I was confused in the first place as I didn't know it so I had no idea why we have repeated definitions. Now I understand it :)

However, this seems like a trick rather than an "official" technique when we overshadow a standard symbol by "preload" another version of that symbol contained in different libraries. But it seems the trick doesn't produce undefined behavior thus I'm fine with this.

ttnghia (approved):

cpp/tests/utilities/default_stream.cpp
vyasr: Yes, you could certainly think of it as a trick in some sense. I wouldn't advocate using it in production in 99% of cases. It turns out to be a perfect tool for what we want to validate with streams, though!

Issue #9639, opened 2021-11-09
[Story] Support string UDFs via Numba 


brandon-b-miller:
**Is your feature request related to a problem? Please describe.**
Currently we can't use string columns inside UDFs. This is for a number of reasons. Firstly, there is limited support for strings in general in Numba, which forms the basis of our UDF framework. Secondly even if strings were supported in numba, we would still need to extend numba for it to be able to properly generate kernels that work as we expect on the buffers containing our string data. Lastly, there are special memory considerations on the GPU that complicate the situation further.

**Describe the solution you'd like**
Recently @davidwendt has experimented with a c++ class which solves many of the nuances around handling single strings that live on the device inside UDFs. @gmarkall subsequently wrote a proof of concept showing how simple string functions such as `len` can be overloaded using numba to map to the methods contained in that c++ class and baked into a kernel. We would like to plumb this machinery through cuDF. This roughly consists of the following steps:

1. Make it so that when cuDF is built, the c++ string class and its methods are precompiled and made available as a blob of PTX or similar that we can link to when building a kernel in python.
2. Create the pipeline in python that writes, links, compiles and executes the correct kernels that can leverage the aformentioned PTX blobs at runtime.
3. Create numba typing and lowering that overloads calls to common string functions in python and maps them to the corresponding methods of the c++ class. Ideally we'd do all of them although some may be more complex than others due to memory considerations. Thats 43 functions:
- `capitalize`
- `casefold`
- `center`
- `count`
- `encode`
- `endswith`
- `expandtabs`
- `find`
- `format`
- `format_map`
- `index`
- `isalnum`
- `isalpha`
- `isascii`
- `isdecimal`
- `isdigit`
- `islower`
- `isprintable`
- `isspace`
- `istitle`
- `isupper`
- `join`
- `ljust`
- `lower`
- `lstrip`
- `maketrans`
- `removeprefix`
- `removesuffix`
- `replace`
- `rfind`
- `rindex`
- `rjust`
- `rpartition`
- `rsplit`
- `rstrip`
- `split`
- `splitlines`
- `startswith`
- `swapcase`
- `title`
- `translate`
- `upper`
- `zfill`

Concretely, when we encounter a UDF that is written like this for example:

```python
def f(row):
    return len(row['str_field'])
```

Our code should
- Detect a declaration of `len` that we will write which expects a `MaskedType(string)` and returns a `MaskedType(int64)`
[...]

PR #12091, opened 2022-11-08
Support `strip`, `lstrip`, and `rstrip` in `strings_udf`

brandon-b-miller:
This PR adds support for the following three functions in `strings_udf`:

- `str.strip(other)`
- `str.lstrip(other)`
- `str.rstrip(other)`


python/strings_udf/cpp/src/strings/udf/shim.cu
```diff
@@ -227,3 +228,45 @@ extern "C" __device__ int udf_string_from_string_view(int* nb_retbal,
 
   return 0;
 }
+
+extern "C" __device__ int strip(int* nb_retval,
```
vyasr, line 232: In the interest of expediency I'm fine merging this as is, but in the long run I suggest that we use some templating or macros to reduce the boilerplate in this file. There are a lot of functions with nearly identical signatures.

vyasr: To be clear, I recognize that the functions that numba is going to call need to use the C ABI, but we can still generate them with a macro or have them internally call a templated function. Not necessary to do in this PR though.

python/strings_udf/strings_udf/_typing.py
```diff
@@ -229,6 +229,7 @@ def resolve_count(self, mod):
     "isnumeric",
     "istitle",
 ]
+string_binary_funcs = ["strip", "lstrip", "rstrip"]
```
vyasr, line 232: This name is somewhat misleading. There are other string binary functions that we already have implemented, including operators and things like `find` and `contains`. Is this list meant to contain binary string functions that also return a string, or is it even more specific than that?

python/strings_udf/strings_udf/lowering.py
```diff
@@ -162,17 +171,40 @@ def deco(cuda_func):
...
+                    cuda_func,
+                    nb_signature(retty, _STR_VIEW_PTR, _STR_VIEW_PTR),
+                    (lhs_ptr, rhs_ptr),
+                )
+                return result
+            else:
+                # binary function of two strings yielding a new string
+                # example: str.strip(other) -> str
+                # shim functions can not return a struct due to C linkage
+                # so we operate on an extra void ptr and throw away nb_retval
```
vyasr, line 193: Don't reference variables that don't exist here, it's hard to understand.
```suggestion
                # shim functions can not return a struct due to C linkage
                # so we create a new udf_string and pass a pointer to it
                # for the shim function to write the output to. The return
                # value of compile_internal is therefore discarded (although
                # this may change in the future if we need to return error
                # codes, for instance).
```

vyasr (approved):
Some minor suggestions for improvement. Feel free to merge once you've addressed, I don't need to look again unless something new comes up (in which case feel free to re-request).

vyasr:
Are tests failing because there's somewhere extra that needs to have the new method attributes added?
```
E                   numba.core.errors.TypingError: Failed in cuda mode pipeline (step: nopython frontend)
E                   #x1B[1m#x1B[1mUnknown attribute 'strip' of type Masked(Poison<#x1B[91m
E                   strings_udf library required for usage of string dtypes inside user defined functions.
E                   #x1B[0m>)
E                   #x1B[1m
E                   File "tests/test_udf_masked_ops.py", line 882:#x1B[0m
E                   #x1B[1m    def func(row):
E                   #x1B[1m        return row["str_col"].strip(strip_char)
E                   #x1B[0m        #x1B[1m^#x1B[0m#x1B[0m
E                   #x1B[0m
```

brandon-b-miller:
> Are tests failing because there's somewhere extra that needs to have the new method attributes added?
I think this is just because those tests need to be marked `@string_udf_test`.

python/strings_udf/cpp/src/strings/udf/shim.cu
brandon-b-miller: Going to resolve this conversation for now but experiment with this a bit for next release 👍

python/strings_udf/strings_udf/_typing.py
brandon-b-miller: Renaming this to `string_return_attrs`

brandon-b-miller:
@davidwendt I'm getting one failing test locally for the following string using:

```python
>>> import cudf
>>> sr = cudf.Series(['.123a'])
>>> def f(st):
...     return st.lstrip('a')
... 
>>> sr.apply(f)
0    .123
dtype: object
```
It looks like this `rstrip`'ed the string rather than `lstrip`'ed it. Interestingly the reverse is also a fail case:
```python
>>> import cudf
>>> sr = cudf.Series(['.123a'])
>>> def f(st):
...     return st.rstrip('.')
... 
>>> sr.apply(f)
0    123a
dtype: object
>>> 
```
Here it looks like this was `lstrip`'ed instead of `rstrip`'ed.

davidwendt:
> Here it looks like this was `lstrip`'ed instead of `rstrip`'ed.

I'm not seeing this error in my unit tests
```
  auto input = cudf::string_view(".123a",5);
  auto achr = cudf::string_view("a",1);
  auto dot = cudf::string_view(".",1);

  d_str = cudf::strings::udf::strip( input, achr, cudf::strings::side_type::LEFT );
  printf("lstrip(.123a,a)=[%s]\n", d_str.data());
  d_str = cudf::strings::udf::strip( input, dot, cudf::strings::side_type::LEFT );
  printf("lstrip(.123a,.)=[%s]\n", d_str.data());
  d_str = cudf::strings::udf::strip( input, achr, cudf::strings::side_type::RIGHT );
  printf("rstrip(.123a,a)=[%s]\n", d_str.data());
  d_str = cudf::strings::udf::strip( input, dot, cudf::strings::side_type::RIGHT );
  printf("rstrip(.123a,.)=[%s]\n", d_str.data());
```

```
lstrip(.123a,a)=[.123a]
lstrip(.123a,.)=[123a]
rstrip(.123a,a)=[.123]
rstrip(.123a,.)=[.123a]
```

brandon-b-miller:
@davidwendt you're completely right, with https://github.com/rapidsai/cudf/pull/12091/commits/02167d390e930c188ab84ee87139d745563bf2b1 the tests pass.

brandon-b-miller:
@gpucibot merge

PR #12094, opened 2022-11-08
Allow setting malloc heap size in string udfs

brandon-b-miller:
Adds a mechanism for setting the default cuda malloc heap size for string UDFs, with 2gb default.


python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,26 @@ def _get_ptx_file():
         return regular_result[1]
 
 
+default_heap_size = int(2e6)
```
davidwendt, line 91: I feel like this needs a big comment on why this is default. This number seems a bit high considering 10M rows of 100 bytes each would only be 1GB. I suppose it  makes sense since it is the largest strings column in bytes possible.

Would it be too strange to make this configurable through an environment variable perhaps?
At least for this release until we can come up with a good metric for setting it.

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,26 @@ def _get_ptx_file():
...
+    """
+    Heap size control for strings_udf, size in bytes.
+    """
+    global heap_size
+    if size == heap_size:
+        return
+    else:
+        (ret,) = cudart.cudaDeviceSetLimit(cudart.cudaLimit(2), size)
+        if ret.value != 0:
+            breakpoint()
```
mroeschke, line 105: Guessing this if a leftover breakpoint

python/strings_udf/strings_udf/__init__.py
brandon-b-miller: I introduced an environment variable to control this. I'm still open to discussion here as to what to set the value as. Indeed I did base it on the size of a strings column.

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,29 @@ def _get_ptx_file():
...
 
+# Maximum size of a string column is 2gb
+STRINGS_UDF_DEFAULT_HEAP_SIZE = int(2e9)
+STRINGS_UDF_HEAP_SIZE = os.environ.get(
+    "STRINGS_UDF_HEAP_SIZE", STRINGS_UDF_DEFAULT_HEAP_SIZE
+)
+heap_size = 0
+
+
+def set_malloc_heap_size(size=STRINGS_UDF_DEFAULT_HEAP_SIZE):
```
davidwendt, line 99: Does this use the environment variable setting?

brandon-b-miller: nice catch! Maybe this is a sign I have too many distinct variables here. Maybe there's a better way of doing this?

davidwendt (approved):

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,29 @@ def _get_ptx_file():
         return regular_result[1]
 
 
+# Maximum size of a string column is 2gb
+STRINGS_UDF_DEFAULT_HEAP_SIZE = int(2e9)
```
bdice, line 92: This doesn't match the maximum `size_type`. It is slightly short. Let's avoid problems arising from that.
```suggestion
STRINGS_UDF_DEFAULT_HEAP_SIZE = int(2 * 1024**3)
```

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,29 @@ def _get_ptx_file():
...
+
+
+def set_malloc_heap_size(size=STRINGS_UDF_HEAP_SIZE):
+    """
+    Heap size control for strings_udf, size in bytes.
+    """
+    global heap_size
+    if size == heap_size:
+        return
+    else:
```
bdice, line 106: Simplify.
```suggestion
    if size != heap_size:
```

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,29 @@ def _get_ptx_file():
...
+        (ret,) = cudart.cudaDeviceSetLimit(cudart.cudaLimit(2), size)
```
bdice, line 107: I think this is what you need. Access the enum value by name, rather than explicitly referring to its integer equivalent.
```suggestion
        (ret,) = cudart.cudaDeviceSetLimit(cudart.cudaLimit.cudaLimitMallocHeapSize, size)
```

bdice (changes requested):

python/strings_udf/strings_udf/__init__.py
bdice: Or better yet:
```suggestion
STRINGS_UDF_DEFAULT_HEAP_SIZE = 2**31
```

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,29 @@ def _get_ptx_file():
         return regular_result[1]
 
 
+# Maximum size of a string column is 2gb
+STRINGS_UDF_DEFAULT_HEAP_SIZE = 2**31
```
bdice, line 92: Should these names be module-scope public or private?

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,29 @@ def _get_ptx_file():
...
+STRINGS_UDF_HEAP_SIZE = os.environ.get(
+    "STRINGS_UDF_HEAP_SIZE", STRINGS_UDF_DEFAULT_HEAP_SIZE
+)
+heap_size = 0
+
+
+def set_malloc_heap_size(size=STRINGS_UDF_HEAP_SIZE):
```
bdice, line 99: Why should this have a default, rather than reading the env var when it is lazily allocated?

I’m not convinced `STRINGS_UDF_HEAP_SIZE` needs to be defined.

python/cudf/cudf/core/udf/__init__.py
```diff
@@ -47,7 +47,14 @@
         utils.JIT_SUPPORTED_TYPES |= STRING_TYPES
         _supported_masked_types |= {string_view, udf_string}
 
-        utils.launch_arg_getters[cudf_str_dtype] = column_to_string_view_array
+        def column_to_string_view_array_init_heap(col):
+            # lazily allocate heap only when a string needs to be returned
```
bdice, line 51: The structure of this code makes it look like idempotency is its main goal, not laziness.

`strings_udf.set_malloc_heap_size()` should only ever be called once, I think?

Maybe this function or module in cudf (and not strings_udf) should own the “is it allocated already?” logic via a cache or global of some kind, and then the strings_udf logic doesn’t need to read env vars for the allocation size, etc.

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,29 @@ def _get_ptx_file():
         return regular_result[1]
 
 
+# Maximum size of a string column is 2gb
+STRINGS_UDF_DEFAULT_HEAP_SIZE = 2**31
+STRINGS_UDF_HEAP_SIZE = os.environ.get(
+    "STRINGS_UDF_HEAP_SIZE", STRINGS_UDF_DEFAULT_HEAP_SIZE
+)
+heap_size = 0
```
vyasr, line 96: I would prefix these with underscores unless you want them to be visible. I could see arguments for either of the first two being visible, but they will also be settable so it's safer to hide them and provide getters if needed later.

python/strings_udf/strings_udf/__init__.py
vyasr, line 99: This default parameter feels sort of weird. If (for whatever reason) you go in later and say `STRINGS_UDF_HEAP_SIZE = some_random_number` it won't change the default parameter for this function, which is encoded at definition time. Is that what you intend/do we care?

brandon-b-miller: The pool is _very_ small if unset - something like 8mb I believe. Perhaps it'd be better if it said something like `_STRINGS_UDF_INITIAL_HEAP_SIZE`?

python/strings_udf/strings_udf/__init__.py
vyasr: Very similar to @bdice's query above.

vyasr:
I realized halfway through my review that my screen hadn't refreshed to include @bdice's latest comments, which largely overlapped with mine. I'll resolve my comments for now so that we just focus on his, I can post again if I have questions about the resolution.

python/cudf/cudf/core/udf/__init__.py
davidwendt: This is not an allocation. There is currently a malloc-heap-size by default (set to about 8MB on the GPUs I've tested on). This updates the heap size to a new value so "already allocated" is not the goal.

vyasr: Is `set_malloc_heap_size` intended to be a public API that can be called by a user, or is it an internal function? It's not entirely clear what the public API is here because this file doesn't define an `__all__`. If the function is just for internal use, why does it need to be called every time someone accesses a string column rather than just calling `cudaDeviceSetLimit` once when the package is imported (or after we can guarantee that the CUDA context/etc is initialized if that is the requirement that we're currently addressing by doing this "lazily")?

python/strings_udf/strings_udf/__init__.py
vyasr: I think @bdice is just asking why there's a default function parameter at all if this is set by an environment variable. This sort of relates to the other question about why we call this lazily rather than at initialization. If it needs to be done lazily for a good reason, then perhaps a better alternative would be to have the default `size=None` and read the environment variable inside the function if `size==None` instead of storing it in a module-scope function. That shouldn't lead to repeated reads since after the first invocation with the current state of the code `heap_size` will always be equal to `size` so it won't go into the conditional.

davidwendt: The lazy initialization is kind of a placeholder in case we have an algorithm that can help guess the size based on the input size for example. I cannot speak to the env var access since I'm not familiar with the scoping here as it relates to udf kernel launches. Ideally if it could be set before each kernel launch? Though I understand the env var may not have changed.

python/cudf/cudf/core/udf/__init__.py
brandon-b-miller: I think what we want to happen is:
- The first time a user runs a string_udf, set the heap size to either default or what is retrieved from their environment. 
- The next time a user runs a string_udf, do not set the heap size.
- Reset the heap the next time the user runs a string_udf iff the user sets the value of `STRING_UDF_HEAP_SIZE` to something else during the session.
Does this seems reasonable?

davidwendt: Seems we have 2 comment chains asking about the same thing.
https://github.com/rapidsai/cudf/pull/12094#discussion_r1024471017

python/strings_udf/strings_udf/__init__.py
brandon-b-miller: +1 on what @davidwendt mentioned about initialization. For the logic discussion, what do you think of the changes in https://github.com/rapidsai/cudf/pull/12094/commits/af7cd9f85b2afa0a6b92f53e4abffbb538c9d02f ?

python/cudf/cudf/core/udf/__init__.py
bdice: > - The first time a user runs a string_udf, set the heap size to either default or what is retrieved from their environment. 

Yes.

> - The next time a user runs a string_udf, do not set the heap size. Or at least early return.

The "set" function should not be called. Avoid designs that try to set but exit early.

> - Reset the heap the next time the user runs a string_udf iff the user sets the value of `STRING_UDF_HEAP_SIZE` to something else during the session.

No. The `STRING_UDF_HEAP_SIZE` should not be modifiable (I argued separately that that variable should not exist at all). If the user needs to be able to set a specific size, then the "set" function should be public and the user should be able to call it explicitly with immediate effect (not waiting until the next UDF execution, because memory demands can change in the meantime).

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,31 @@ def _get_ptx_file():
...
+_STRINGS_UDF_DEFAULT_HEAP_SIZE = 2**31
+heap_size = 0
+
+
+def set_malloc_heap_size(size=None):
+    """
+    Heap size control for strings_udf, size in bytes.
+    """
+    global heap_size
+    if size == None:
```
bdice, line 101:
```suggestion
    if size is None:
```

python/strings_udf/strings_udf/__init__.py
```diff
@@ -87,6 +88,31 @@ def _get_ptx_file():
...
+        size = os.environ.get(
```
bdice, line 102: Generally environment variables take effect at load time. I don't think we should allow/encourage the user to modify the environment of the process while it is running in order to change this.

I think you want to define it on import:
```python
_STRINGS_UDF_DEFAULT_HEAP_SIZE = os.environ.get("STRINGS_UDF_HEAP_SIZE", 2**31)
```

python/cudf/cudf/core/udf/__init__.py
brandon-b-miller: With the changes in https://github.com/rapidsai/cudf/commit/af7cd9f85b2afa0a6b92f53e4abffbb538c9d02f I still have `heap_size` available as a module level attribute of `strings_udf`, what do you think about checking if `heap_size == 0` within the getter to determine if the function need be called or not?

bdice: I would use a cached function for this. Following ideas from https://mail.python.org/archives/list/python-ideas@python.org/thread/5OR3LJO7LOL6SC4OOGKFIVNNH4KADBPG/

```python
from functools import lru_cache

# Only executes the body of the function on the first call
once = lru_cache(maxsize=None)

@once
def set_initial_malloc_heap_size():
    strings_udf.set_malloc_heap_size()

def column_to_string_view_array_init_heap(col):
    set_initial_malloc_heap_size()
    return column_to_string_view_array(col)
```

brandon-b-miller: thanks this worked perfectly!

brandon-b-miller:
rerun tests

python/cudf/cudf/core/udf/__init__.py
```diff
@@ -21,6 +23,13 @@
 )
 _STRING_UDFS_ENABLED = False
 cudf_str_dtype = dtype(str)
+
+
+@lru_cache(maxsize=None)
+def set_initial_malloc_heap_size():
```
bdice, line 29: Why is this defined twice? Delete this in favor of line 60?

bdice (approved):
Looks like a function is defined twice. Otherwise LGTM.

brandon-b-miller:
@gpucibot merge

PR #12036, opened 2022-11-01
Force black exclusions for pre-commit.

bdice:
## Description
This PR fixes an issue where `black` would not respect the exclusions in `pyproject.toml` when run via `pre-commit`. When executing the hook, `pre-commit` explicitly lists all changed files and provides them as positional arguments to `black`. When `black` receives explicit filenames, it does not respect the `exclude` settings but _does_ respect the `force-exclude` settings added in `black 20.8b1`. See issue: https://github.com/psf/black/issues/438 and PR: https://github.com/psf/black/pull/1032.

cc: @csadorf for cuML
 
## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12097, opened 2022-11-09
Add codespell as a linter

benfred:
## Description

This adds codespell as a linter to the pre-commit config, and fixes various spelling errors it identifies  (https://github.com/codespell-project/codespell)

## Checklist


setup.cfg
```diff
@@ -53,3 +53,13 @@ exclude = (?x)(
   | dask_cudf/dask_cudf/tests/
   # This close paren cannot be in column zero otherwise the config parser barfs
  )
+
+[codespell]
+skip = ./.git,./.github,./cpp/build,.*egg-info.*,versioneer.py,./.mypy_cache,./cpp/tests,./python/cudf/cudf/tests
+ignore-words-list = inout,indext,offsett,dout,unparseable,incase,afterall,fiter,trings
```
bdice, line 59: This list is not that long! 💯

bdice (approved):
Nice @benfred! I'd support including this hook.

setup.cfg
benfred: I'm cheating a bit and ignoring the tests =) 

(The tests have some test data in there that was messing it up - but wasn't really typos)

bdice: I think that’s reasonable. We can always expand later if we wish. The errors fixed in this PR are worthwhile even if not covering the whole repo.

galipremsagar (approved):
Love this code spell linter addition ❤️

setup.cfg
davidwendt: You probably could add `./java/src/test` to this list and then remove `trings`
And add `./cpp/include/cudf_test/cxxopts.hpp` and then remove `fiter`.
I'm not finding `indext` or `afterall` in the code.
Is `offsett` in here because of `offsetted`? Maybe that can be corrected.

davidwendt:
ok to test

setup.cfg
davidwendt: Also, should the `skip` list include `./python/cudf/_skbuild` and `./python/strings_udf/_skbuild` as well?

ttnghia (approved):

setup.cfg
benfred: It doesn't seem like its possible to get codespell to automatically exclude files from .gitignore  yet =( https://github.com/codespell-project/codespell/issues/1887 

Running codespell through pre-commit won't run on any files that aren't being tracked by git though, which should get around having to mirror the .gitignore file.   I've changed the codespell pre-commit exclude regex to exclude test files and versioneer.py files - which should make codespell pass when run with `pre-commit run --all-files`

The skips in setup.cfg are for running codespell in its interactive mode, and will be ignored when used with pre-commit - because of basically the same issue in https://github.com/rapidsai/cudf/pull/12036  (pre-commit passes an explicit list of files to codespell, which the `skip` section provided here doesn't override).  We could perhaps entirely remove this skip section , if we don't want to support running directly codespell from the command line rather than from pre-commit.

I've also added java/src/test to the skips, and removed trings, afterall and indext.

`offsett` is because of template params like https://github.com/rapidsai/cudf/blob/7d173c9d144a64c5e1a0467d2a5eb4181854f25e/cpp/src/io/fst/agent_dfa.cuh#L79-L84

wence- (approved):
Yes, thanks!

setup.cfg
davidwendt: Can we fix the `fiter` one? It only appears in `cxxopts.hpp` and seems like a typo:
https://github.com/rapidsai/cudf/blob/4497ed6bcd2347ba7a77f4b8d9e7f867c10e1d42/cpp/include/cudf_test/cxxopts.hpp#L1417-L1434

wence-: > `fitr` would work, right

I would be inclined to just call this `iter` or `it` ?

wence-: > `offsett` is because of template params like

Add `\b[a-zA-Z][[:wordlike:]]T\b` to the ignore regex?

bdice: I believe `cxxopts.hpp` is vendored and should be ignored rather than edited. I believe that specific name `fiter` is supposed to mean “format iterator” and is not a typo.

davidwendt: Then it should not be linted at all.
If you look at the history for `cxxopts.hpp` it has been modified it several times.
I realize `fiter` is not a typo but it seems odd to create an exception for it for the entire repo.
Either we should not lint this file or we should fix it.

bdice: I agree with David that we should not lint this file. (My previous comment may have been unclear.)

benfred: I've removed the cxxopts.hpp from linting - and applied @wence- suggestion about the ignore regex to exclude template params types. The ignore-words list is down to just `inout,unparseable` now, and I think those are both valid words myself.

bdice: Perfect! That's ideal.

benfred:
Rerun tests

benfred:
Rerun tests

setup.cfg
```diff
@@ -53,3 +53,13 @@ exclude = (?x)(
...
   # This close paren cannot be in column zero otherwise the config parser barfs
  )
+
+[codespell]
+# note: pre-commit passes explicit lists of files here, which this skip file list doesn't override -
+# this is only to allow you to run codespell interactively
+skip = ./.git,./.github,./cpp/build,.*egg-info.*,versioneer.py,./.mypy_cache,./cpp/tests,./python/cudf/cudf/tests,./java/src/test,./cpp/include/cudf_test/cxxopts.hpp
+# ignore short words, and typename parameters like OffsetT
+ignore-regex = \b(.{1,4}|[A-Z]\w*T)\b
+ignore-words-list = inout,unparseable
```
karthikeyann, line 63: ignore list is so small.
In cspell, I had 74 words for cudf, 117 words for C++14 keywords.
code-spell is non-dictionary based. misspellings checker!

karthikeyann (approved):
Awesome work! 💯 
I used cspell to run spell check locally. But too many special cases, own dictionary to avoid many C++14, cudf, data analytics specific names. It is hard to automate it as pre-commit hook.
code-spell is great! misspellings checker.

Couple of points:
- This runs on CHANGELOG.md as well. Now Change log is autogenerated by taking PR's title. After every release, the new branch may fail this hook.
- Mention this check in pre-commit section in `CONTRIBUTING.md` or `docs/cudf/source/developer_guide/contributing_guide.md`

setup.cfg
benfred: thanks @karthikeyann !   I've applied your suggestions in the last commit.

One reason I like codespell is because of the relatively small number of false positives, meaning the exclusions is relatively small compared to other tools -

CONTRIBUTING.md
```diff
@@ -377,6 +377,11 @@ Now code linters and formatters will be run each time you commit changes.
 
 You can skip these checks with `git commit --no-verify` or with the short version `git commit -n`.
 
+cuDF also uses [codespell](https://github.com/codespell-project/codespell) to find spelling
+mistakes, and this check is run as part of the pre-commit hook. To apply the suggested spelling
+fixes, you can run  `codespell -i 3 -w .` from the commandline in the cuDF root directory.
```
davidwendt, line 382:
```suggestion
fixes, you can run  `codespell -i 3 -w .` from the command-line in the cuDF root directory.
```
I've not seen command-line as one word before.
https://en.wikipedia.org/wiki/Command-line_interface

karthikeyann:
rerun tests

benfred:
Rerun tests

davidwendt:
I wondering if this perhaps can be moved to 23.02 only because I think there is a risk to 22.12 hotfixes (?). Technically this PR would have to be the very last PR to be built/merged on the latest 22.12 code to ensure all files in 22.12 have been checked.
For example, there may be current 22.12 PRs that would not pass this check that could get merged. If this PR is then merged on its current state then any future hotfixes could fail.
If it is moved to 23.02 then any such discrepancies can just be new PRs or updates to existing PRs.

bdice:
I support moving to 23.02 to minimize risk as @davidwendt mentioned. I can migrate it and fix the merge conflicts.

benfred:
rerun tests

bdice:
I changed (and reset) the target branch so that the Branch Checker CI would be unblocked. It looks like CI will need to re-run, then this can be merged.

bdice:
@gpucibot merge

galipremsagar:
rerun tests

ttnghia:
> [Fix typo in ChunkedReaderJni.cpp](https://github.com/rapidsai/cudf/pull/12097/commits/dde04c559f61e35aeb81d4ecaf0e79bd937a82c9)

OMG I can't know why I've made this typo. I swear that I stared at that comment many times before merging :joy:

PR #12098, opened 2022-11-09
Fix reading of CSV files with blank second row

vuule:
## Description
There are two options to get the names of columns in a CSV file - header or the first row. In case the first row is used, names are generated, and the only part of the row that is used is the number of detected columns.

This PR fixes the corner case where a blank line after the first (non-header) row causes the reader to detect an additional column (and return an additional column of nulls). 
The fix is to break when there is a terminator character within the first row; this only happens with blank row(s) after the first data row. The reader already does this when reading column names from a header, this PR just removes this difference in behavior that was causing the bug.

## Checklist


davidwendt (approved):
LGTM

vuule:
@gpucibot merge

PR #12099, opened 2022-11-09
Support `upper` and `lower` in `strings_udf`

brandon-b-miller:
This PR adds support for the following two functions in `strings_udf`:

- `str.upper()`
- `str.lower()`


python/strings_udf/cpp/src/strings/udf/shim.cu
```diff
@@ -227,3 +229,72 @@ extern "C" __device__ int udf_string_from_string_view(int* nb_retbal,
...
+
+  cudf::strings::udf::chars_tables tables;
+  tables.flags_table                = flags_table_ptr;
+  tables.cases_table                = cases_table_ptr;
+  tables.special_case_mapping_table = special_table_ptr;
+
+  *udf_str_ptr = to_upper(tables, *st_ptr);
+
+  return 0;
+}
```
davidwendt, line 300: Is the `lower()` definition missing here?

brandon-b-miller: Added 👍

python/cudf/cudf/tests/test_udf_masked_ops.py
```diff
@@ -876,7 +876,6 @@ def func(row):
     run_masked_udf_test(func, str_udf_data, check_dtype=False)
 
 
-@string_udf_test
```
galipremsagar, line 879: Was this removal needed? or a typo?

galipremsagar (approved):

python/cudf/cudf/tests/test_udf_masked_ops.py
```diff
@@ -876,7 +876,6 @@ def func(row):
...
 @pytest.mark.parametrize("strip_char", ["1", "a", "12", " ", "", ".", "@"])
```
galipremsagar, line 879:
```suggestion
@string_udf_test
@pytest.mark.parametrize("strip_char", ["1", "a", "12", " ", "", ".", "@"])
```

galipremsagar (changes requested):
Looks like the CI failures are around this change.

python/cudf/cudf/tests/test_udf_masked_ops.py
brandon-b-miller: 👍

python/strings_udf/strings_udf/_lib/tables.pyx
```diff
@@ -1,14 +1,26 @@
...
+
+
+def get_character_cases_table_ptr():
+    cdef const uint16_t* tbl_ptr = cpp_get_character_cases_table()
+    return np.uint64(<uintptr_t>tbl_ptr)
+
+
+def get_special_case_mapping_table_ptr():
+    cdef const void* tbl_ptr = cpp_get_special_case_mapping_table()
+    return np.uint64(<uintptr_t>tbl_ptr)
```
wence-, line 26: You want [`numpy.uintp`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.uintp) not `uint64`

python/strings_udf/cpp/src/strings/udf/shim.cu
```diff
@@ -270,3 +271,57 @@ extern "C" __device__ int rstrip(int* nb_retval,
 
   return 0;
 }
+
+extern "C" __device__ int upper(int* nb_retval,
+                                void* udf_str,
+                                void* const* st,
+                                std::int64_t flags_table,
+                                std::int64_t cases_table,
+                                std::int64_t special_table)
```
wence-, line 280: These are pointers being represented by integers I think. So you want `std::uintptr_t` not `int64_t` (although it is only optional in C++11 and later, I think it is provided on all sane platforms).

This comment applies throughout (and indeed in existing code where you accept pointers as integers and then call `reinterpret_cast<some_ptr*>(some_integer)`)

davidwendt, line 280: Just curious why these are not `void*` as well?

python/strings_udf/cpp/src/strings/udf/shim.cu
```diff
@@ -270,3 +271,57 @@ extern "C" __device__ int rstrip(int* nb_retval,
...
+  tables.special_case_mapping_table = special_table_ptr;
+
+  *udf_str_ptr = to_upper(tables, *st_ptr);
+
+  return 0;
+}
+
+extern "C" __device__ int lower(int* nb_retval,
+                                void* udf_str,
+                                void* const* st,
```
wence-, line 304: Does this need to be a `void **` or can it just be a `void *` (I note that inside you cast to `string_view *` and then dereference, so I think you can strip a `*` everywhere).

python/strings_udf/cpp/src/strings/udf/udf_apis.cu
```diff
@@ -42,7 +42,8 @@ namespace {
 struct udf_string_to_string_view_transform_fn {
   __device__ cudf::string_view operator()(cudf::strings::udf::udf_string const& dstr)
   {
-    return cudf::string_view{dstr.data(), dstr.size_bytes()};
+    return dstr.data() == nullptr ? cudf::string_view{"", 0}
+                                  : cudf::string_view{dstr.data(), dstr.size_bytes()};
```
wence-, line 46:
```suggestion
    return dstr.data() == nullptr ? cudf::string_view{}
                                  : cudf::string_view{dstr.data(), dstr.size_bytes()};
```

I think `string_view` has a default empty constructor.

galipremsagar (approved):

python/strings_udf/cpp/src/strings/udf/shim.cu
brandon-b-miller: Let me make sure all is right here. Each of the functions that return a pointer to an input mapping table returns its own type of pointer. For the flags table its `uint8_t*`, for the cases table its `uint16_t`. For the special case mapping table it's a `special_case_mapping*`. Would the correct thing to do in this case be to receive each of these as a `uintptr_t` and then carry them through the python into the lowering as a `np.uintp`? Then these shim functions could accept a `uintptr_t` here.

python/strings_udf/strings_udf/lowering.py
```diff
@@ -324,6 +345,74 @@ def id_func_impl(context, builder, sig, args):
...
+        @cuda_lower(id_func, string_view)
+        def id_func_impl(context, builder, sig, args):
+            str_ptr = builder.alloca(args[0].type)
+            builder.store(args[0], str_ptr)
+
+            # Lookup table required for conversion functions
+            # must be resolved at runtime after context initialization,
+            # therefore cannot be a global variable
+            flags_tbl_ptr = context.get_constant(
+                types.int64, character_flags_table_ptr
```
wence-, line 365: Probably `types.uintp` throughout

python/strings_udf/cpp/src/strings/udf/shim.cu
davidwendt: I think it would be more correct these to be pointer types than `int64_t` types.

python/strings_udf/cpp/src/strings/udf/shim.cu
```diff
@@ -270,3 +271,57 @@ extern "C" __device__ int rstrip(int* nb_retval,
...
+    reinterpret_cast<cudf::strings::detail::character_flags_table_type*>(flags_table);
+  auto cases_table_ptr =
+    reinterpret_cast<cudf::strings::detail::character_cases_table_type*>(cases_table);
+  auto special_table_ptr =
+    reinterpret_cast<cudf::strings::detail::special_case_mapping*>(special_table);
+
+  cudf::strings::udf::chars_tables tables;
+  tables.flags_table                = flags_table_ptr;
+  tables.cases_table                = cases_table_ptr;
+  tables.special_case_mapping_table = special_table_ptr;
```
wence-, line 322: Can we not do: `cudf::strings::udf::chars_tables tables{flags_table_ptr, cases_table_ptr, special_table_ptr};` ?

python/strings_udf/cpp/src/strings/udf/shim.cu
```diff
@@ -270,3 +271,57 @@ extern "C" __device__ int rstrip(int* nb_retval,
...
 }
+
+extern "C" __device__ int upper(int* nb_retval,
+                                void* udf_str,
+                                void* const* st,
+                                std::int64_t flags_table,
+                                std::int64_t cases_table,
+                                std::int64_t special_table)
+{
+  auto udf_str_ptr = new (udf_str) udf_string;
```
wence-, line 282: This calls the default constructor, but this object is never read anywhere, and is then overwritten on line 297 (`to_upper` returns a `udf_string`). So what is going on here?

Does this want to disappear and line 297 be `*reinterpret_cast<udf_string *>(udf_str) = to_upper(tables, *st_ptr);` ?

wence- (changes requested):
A bunch of nits around cross-calling ABI and type-punning pointers and integers.

python/strings_udf/cpp/src/strings/udf/shim.cu
wence-: If you can interact with ctypes I think you can carry around a `ctypes.cvoidp`.

python/strings_udf/cpp/src/strings/udf/udf_apis.cu
davidwendt: > I think `string_view` has a default empty constructor.

You are correct.

python/strings_udf/cpp/src/strings/udf/shim.cu
davidwendt: No the object must be initialized by a constructor in order for it to be valid. Otherwise the `operator=` will crash if the target has garbage in it.

python/strings_udf/cpp/src/strings/udf/shim.cu
brandon-b-miller: I am not sure I follow here. This follows the pattern from the rest of the shim functions when a `string_view` is an arg. IIUC `st` is pointing directly to the struct, so only one level of pointing right?

python/strings_udf/strings_udf/lowering.py
```diff
@@ -324,6 +345,74 @@ def id_func_impl(context, builder, sig, args):
     return deco
 
 
+def create_upper_or_lower(id_func):
+    """
+    Provide a wrapper around numba's low-level extension API which
+    produces the boilerplate needed to implement a unary function
+    of a string.
```
wence-, line 352: Docstring wants changed?

python/strings_udf/cpp/src/strings/udf/shim.cu
wence-: Thanks!

python/strings_udf/cpp/src/strings/udf/shim.cu
wence-: Most functions here (e.g. https://github.com/rapidsai/cudf/blob/418cbc69b336f34b06e1de21ae2c016b40f1e6fc/python/strings_udf/cpp/src/strings/udf/shim.cu#L186-L193) that take a point to `string_view` take the argument as `void const *`, not `void* const *`. Why is this different?

brandon-b-miller:
rerun tests

wence- (approved):
Thanks @brandon-b-miller

davidwendt (approved):
Approving C++ code.

brandon-b-miller:
some test failures here that seem to be related to a previous PR that I am looking into now.

brandon-b-miller:
I am thinking these CI failures _might_ be transient. Going to `rerun tests` just to make sure, as things seem to pass locally for me.

brandon-b-miller:
rerun tests

brandon-b-miller:
@gpucibot merge

Issue #12093, opened 2022-11-08
[FEA] cudf::column needs a set_stream() function.


nvdbaranec:
`cudf::column` uses `rmm::device_buffer` for internal storage. `rmm::device_buffer` internally stores the stream it was created on.  This can create issues when creating columns on temporary worker-style streams and then returning them to a primary stream.  When the column gets destroyed, rmm throws exceptions about invalid device contexts.

A `set_stream(rmm::cuda_stream_view)` function that recurses through all children would make it easier to hand off columns between streams. 

We initially encountered this issue with `cudf::io::column_buffer` which has a similar problem but it will also be an issue for columns themselves.

bdice:
The `rmm::device_buffer` has a [set_stream](https://docs.rapids.ai/api/librmm/nightly/classrmm_1_1device__buffer.html#ab271ced85f304e3061a3ff72526dbc37) method. This proposed API would call that method of the underlying buffer.

This seems reasonable from what I can tell, but its utility (transferring ownership between streams) will be limited until we expose streams publicly, aside from the internal I/O methods that @hyperbolic2346 is working on. Perhaps we should wait to add this to the public API until we have a broader discussion of adding streams to libcudf APIs?

nvdbaranec:
I think that's totally fine.  The stuff Mike is working on is just experimental stuff as is, and doesn't actually require this from `column` itself.

hyperbolic2346:
I added a PR showing what we are talking about here. I don't know if this is the best way to expose this, but it hopefully clears it up enough to have a solid conversation about it.

vyasr:
@nvdbaranec @hyperbolic2346 now that #13744 is closed, do you want to revisit this issue?

bdice:
https://github.com/rapidsai/cudf/issues/19118 was opened as a duplicate of this issue. @tgujar Feel free to share more about your use case here.

tgujar:
Our use case is also for handing off columns between streams. 
```
task t1 {...};
task t2 {...};
rmm::cuda_stream s1, s2;
{
    std::shared_ptr<cudf::column> c {...};
    t1.add_dependency(c);
    t2.add_dependency(c);
}
t1.execute(s1); t2.execute(s2);
```

[...]

PR #12100, opened 2022-11-09
Adding set_stream for column and column_buffer

hyperbolic2346:
## Description
As described in https://github.com/rapidsai/cudf/issues/12093, I ran into some issues with streams when I was doing some prototyping. I don't know if we are ready to do this change, but I wanted to put it up so people could see what it is we may need to do in the future in some form.

## Checklist


cpp/include/cudf/column/column.hpp
```diff
@@ -347,6 +347,18 @@ class column {
...
+   * @brief Sets the stream to be used for deallocation of internal buffers
+   */
+  void set_stream(rmm::cuda_stream_view stream) noexcept
+  {
+    _data.set_stream(stream);
+    _null_mask.set_stream(stream);
+    for (auto& child : _children) {
+      child->set_stream(stream);
+    }
+  }
```
davidwendt, line 360: Would prefer the function definition be put into `cpp/src/column/column.cu` instead of in the header.

cpp/src/io/utilities/column_buffer.hpp
```diff
@@ -108,6 +108,19 @@ struct column_buffer {
...
+   */
+  void set_stream(rmm::cuda_stream_view stream) noexcept
+  {
+    if (_strings) _strings->set_stream(stream);
+    _data.set_stream(stream);
+    _null_mask.set_stream(stream);
+    for (auto& child : children) {
+      child.set_stream(stream);
+    }
+  }
```
davidwendt, line 122: Same here. Move the function definition into `cpp/src/io/utilities/column_buffer.cpp` if possible.

davidwendt (changes requested):

ttnghia:
Why do we have this to change the underlying stream for these buffers? Can we construct these buffers with the desired stream from the beginning?

nvdbaranec:
> Why do we have this to change the underlying stream for these buffers? Can we construct these buffers with the desired stream from the beginning?

It's applicable for certain kinds of worker-stream type situations.  The overlapped parquet work that Mike is doing uses two streams to overlap IO work and decoding work in a double-buffering kind of way.  The decode step (including memory allocation) happens in a separate stream from the top level one that the reader is called with so you explicitly tell the underlying rmm device buffers that they belong to a different stream than the one they were created on.

ttnghia:
Is this still relevant?

hyperbolic2346:
I believe so. This will become more and more of a thing as we expand our use of streams in cuDF. I'll update this PR.

cpp/include/cudf/column/column.hpp
hyperbolic2346: This has changed to a templated function, but I pulled out the child functions.

cpp/src/io/utilities/column_buffer.hpp
hyperbolic2346: Done

davidwendt: `set_stream()` still seems to be defined in the header `column_buffer.hpp`?

hyperbolic2346: I'm not sure how that happened. VS Code shows the file without the definition, but the local filesystem still had it. Updated via my old friend, vi.

vyasr:
At this point I think we probably want to recreate this PR from scratch. I'll add some relevant information to the linked issue, though.

PR #12017, opened 2022-10-27
byte_range support for JSON Lines format

karthikeyann:
## Description
This PR adds support for byte_range to be used in nested JSON parser for JSON Lines format (newline delimited JSON http://ndjson.org/)
The record delimiter "New lines" are only expected at the end of each record. Newlines in middle of record or within quotes are not expected and will lead to unknown behaviour. The record delimiters are not context aware in this PR.

This PR provides libcudf APIs, Cython APIs and python tests to enable byte range support. This will allow dask to do distributed/segmented parsing of JSON.

No Dask changes

Addresses part of https://github.com/rapidsai/cudf/issues/11843
Depends on  #12060

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12101, opened 2022-11-09
Expose engine argument in dask_cudf.read_json

rjzamora:
## Description
Exposes the `engine` argument in `dask_cudf.read_json`, enabling `dask_cudf.read_json(... engine="cudf_experimental")` for nested json data.

TODO (~maybe this PR?~):

- [ ] (**EDIT**: This should be done in a separate PR) Add simple/optimized code path to leverage the `byte_range` parameter for local storage (similar to what is done in [`dask_cudf.read_csv`](https://github.com/rapidsai/cudf/blob/7535f31cfaf7e01578c413bb3ba46b03d2014806/python/dask_cudf/dask_cudf/io/csv.py#L72)). This would depend on #12017 for nested json data.

## Checklist

### DIFF
--- a/python/dask_cudf/dask_cudf/backends.py
+++ b/python/dask_cudf/dask_cudf/backends.py
@@ -2,6 +2,7 @@
 
 import warnings
 from collections.abc import Iterator
+from functools import partial
 
 import cupy as cp
 import numpy as np
@@ -493,11 +494,15 @@ try:
             )
 
         @staticmethod
-        def read_json(*args, engine=None, **kwargs):
+        def read_json(*args, engine="cudf", **kwargs):
             return _default_backend(
                 dd.read_json,
                 *args,
-                engine=cudf.read_json,
+                engine=(
+                    partial(cudf.read_json, engine=engine)
+                    if isinstance(engine, str)
+                    else engine
+                ),
                 **kwargs,
             )
 
--- a/python/dask_cudf/dask_cudf/backends.py
+++ b/python/dask_cudf/dask_cudf/backends.py
@@ -2,7 +2,6 @@
 
 import warnings
 from collections.abc import Iterator
-from functools import partial
 
 import cupy as cp
 import numpy as np
@@ -494,17 +493,10 @@ try:
             )
 
         @staticmethod
-        def read_json(*args, engine="cudf", **kwargs):
-            return _default_backend(
-                dd.read_json,
-                *args,
-                engine=(
-                    partial(cudf.read_json, engine=engine)
-                    if isinstance(engine, str)
-                    else engine
-                ),
-                **kwargs,
-            )
+        def read_json(*args, **kwargs):
+            from dask_cudf.io.json import read_json
+
+            return read_json(*args, **kwargs)
 
         @staticmethod
         def read_orc(*args, **kwargs):
--- a/python/dask_cudf/dask_cudf/io/json.py
+++ b/python/dask_cudf/dask_cudf/io/json.py
@@ -6,4 +6,17 @@ import dask
 
 import cudf
 
-read_json = partial(dask.dataframe.read_json, engine=cudf.read_json)
+
+def read_json(path, engine="auto", **kwargs):
+    # Wrap `dd.read_json` with special engine handling
+
+    # TODO: Add optimized code path to leverage the
+    # `byte_range` argument in `cudf.read_json` for
+    # local storage (see `dask_cudf.read_csv`)
+    if isinstance(engine, str):
+        # Pass `str` engine argument to `cudf.read_json``
+        engine = partial(cudf.read_json, engine=engine)
+    elif not callable(engine):
+        raise ValueError(f"Unsupported engine option: {engine}")
+    # Pass `callable` engine argument to `dd.read_json`
+    return dask.dataframe.read_json(path, engine=engine, **kwargs)
--- a/python/dask_cudf/dask_cudf/io/json.py
+++ b/python/dask_cudf/dask_cudf/io/json.py
@@ -6,17 +6,66 @@ import dask
 
 import cudf
 
+from dask_cudf.backends import _default_backend
 
-def read_json(path, engine="auto", **kwargs):
-    # Wrap `dd.read_json` with special engine handling
+
+def read_json(url_path, engine="auto", **kwargs):
+    """Create a dask_cudf DataFrame collection from JSON data
+
+    This function wraps ``dask.dataframe.read_json``, and passes
+    ``engine=partial(cudf.read_json, engine="auto")`` by default.
+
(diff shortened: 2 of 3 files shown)


python/dask_cudf/dask_cudf/backends.py
```diff
@@ -493,11 +494,15 @@ def read_parquet(*args, engine=None, **kwargs):
...
+        def read_json(*args, engine="cudf", **kwargs):
             return _default_backend(
                 dd.read_json,
                 *args,
-                engine=cudf.read_json,
+                engine=(
+                    partial(cudf.read_json, engine=engine)
+                    if isinstance(engine, str)
+                    else engine
+                ),
```
wence-, line 505: So a simpler patch would just be passing `engine=engine or cudf.read_json` through. Though I suppose that this is a slightly easier to use.

Would be useful to add a short docstring describing the `engine` parameter (since it now differs from `dask.dataframe.read_json`.

Question: what happens if the user passes `pd.read_json` as engine here?

python/dask_cudf/dask_cudf/io/json.py
```diff
@@ -6,4 +6,17 @@
 
 import cudf
 
-read_json = partial(dask.dataframe.read_json, engine=cudf.read_json)
+
+def read_json(path, engine="auto", **kwargs):
+    # Wrap `dd.read_json` with special engine handling
```
wence-, line 11: Does it make sense to again add a short docstring here?

python/dask_cudf/dask_cudf/io/json.py
```diff
@@ -6,4 +6,17 @@
...
+    # TODO: Add optimized code path to leverage the
+    # `byte_range` argument in `cudf.read_json` for
+    # local storage (see `dask_cudf.read_csv`)
+    if isinstance(engine, str):
+        # Pass `str` engine argument to `cudf.read_json``
+        engine = partial(cudf.read_json, engine=engine)
+    elif not callable(engine):
+        raise ValueError(f"Unsupported engine option: {engine}")
+    # Pass `callable` engine argument to `dd.read_json`
+    return dask.dataframe.read_json(path, engine=engine, **kwargs)
```
wence-, line 22: I think I perhaps don't understand the backend hookup properly. Why do we have to do this both here and in the `CudfBackendEntryPoint`?

python/dask_cudf/dask_cudf/io/tests/test_json.py
```diff
@@ -71,3 +71,22 @@ def test_read_json_lines(lines):
...
+    # Check that `engine="cudf_experimental"` can
+    # be used to support nested data
+    df = pd.DataFrame(
+        {
+            "a": [{"y": 2}, {"y": 4}, {"y": 6}, {"y": 8}],
+            "b": [[1, 2, 3], [4, 5], [6], [7]],
+            "c": [1, 3, 5, 7],
+        }
+    )
+    lines = dict(orient="records", lines=True)
```
wence-, line 87:
```suggestion
    kwargs = dict(orient="records", lines=True)
```
?

python/dask_cudf/dask_cudf/io/tests/test_json.py
```diff
@@ -71,3 +71,22 @@ def test_read_json_lines(lines):
...
+    with tmpfile("json") as f:
```
wence-, line 88: Use the `tmp_path` fixture to get a temporary `pathlib.Path` object that pytest manages and will clean up afterwards (but leaves lying around in the case that the test fails, which can be useful):

```python
def test_read_json_nested_experimental(tmp_path):
    ...
    with tmp_path / "data.json" as f:
         ...
```

wence- (changes requested):
Very minor queries. Overall looks good!

python/dask_cudf/dask_cudf/backends.py
rjzamora: Thanks for the quick review @wence- !

>So a simpler patch would just be passing engine=engine or cudf.read_json through. Though I suppose that this is a slightly easier to use.

Not sure this is true?

Just to clarify what is intended here: Before this change, `dd`/`dask_cudf` was expecting a callable argument for `engine` (like `cudf.read_json` or `pd.read_json`).  The `cudf` version, however, expects a string argument (like "cudf", "cudf_experimental", or "pandas"). The purpose of this change is to let the user pass in either to `dask_cudf` (or to the "cudf"-dispatched version of `dd.read_json`)

>Would be useful to add a short docstring describing the engine parameter (since it now differs from dask.dataframe.read_json.

I was hoping to avoid this originally, but I completely agree.

>Question: what happens if the user passes pd.read_json as engine here?

The user will get a pandas-backed Dask-DataFrame. The "confusing" result is if the user passes in "pandas", because `cudf.read_json` will take this to mean: "use pandas to read the data, but then convert to cudf".

python/dask_cudf/dask_cudf/io/json.py
rjzamora: Yeah - I'll add a new docstring now that the engine logic is more expansive than `dd.read_json`

python/dask_cudf/dask_cudf/io/tests/test_json.py
rjzamora: Ah - lazy copy-pasting on my end. Thanks for catching this!

python/dask_cudf/dask_cudf/backends.py
wence-: > Thanks for the quick review @wence- !
> 
> > So a simpler patch would just be passing engine=engine or cudf.read_json through. Though I suppose that this is a slightly easier to use.
> 
> Not sure this is true?

Ah sorry, I meant:

```python
def read_json(..., engine=None, ...):
    return whatever.read_json(..., engine=engine or cudf.read_json, ...)
```
And if the user wants to provide a default to `cudf.read_json` then they would say:
```python
read_json(..., engine=partial(cudf.read_json, engine="cudf_experimental"), ...)
```

But this is not very nice UX, so this is basically a long way of saying I approve of this change.

 [...]

> > Would be useful to add a short docstring describing the engine parameter (since it now differs from dask.dataframe.read_json.
> 
> I was hoping to avoid this originally, but I completely agree.
> 
> > Question: what happens if the user passes pd.read_json as engine here?
> 
> The user will get a pandas-backed Dask-DataFrame. The "confusing" result is if the user passes in "pandas", because `cudf.read_json` will take this to mean: "use pandas to read the data, but then convert to cudf".

Thanks, I think the docstring clarifies nicely what is going to happen.

wence- (approved):
Thanks!

rjzamora:
@gpucibot merge

Issue #11540, opened 2022-08-16
[FEA] Remove usages of `pyorc` where not necessary


galipremsagar:
**Is your feature request related to a problem? Please describe.**
Writing a pandas dataframe to an orc file using `pyorc` is a bit of a complex operation. Until now we have been using `pyorc` as a reference writer because we had no other choice, with the introduction of `pyarrow`'s orc writer we should be making a switch from `pyorc` and this should remove a lot of complex handling that needs to be done for nested dtypes.

**Describe the solution you'd like**
Drop `pyorc` usages to almost none - Though we will keep it probably for a few basic dtype tests to validate compatibilty. But fuzz-testing and the rest of pytests should make the switch.

**Describe alternatives you've considered**
The FEA itself is a better alternative to `pyorc` 😉 

**Additional context**
https://arrow.apache.org/docs/python/generated/pyarrow.orc.write_table.html

https://github.com/rapidsai/cudf/issues/11540#tasklist-block-36d8af1d-36b8-447e-9227-e4ad3da9d7e1

galipremsagar:
cc: @GregoryKimball @vuule

vuule:
Is the main difference in the PyORC's requirement to pass in a schema?
Would it be possible to try this out in fuzz tests to verify that pyarrow is robust?

galipremsagar:
> Is the main difference in the PyORC's requirement to pass in a schema?
That + while the writer was an internal only API it lacked the stripe_size support which was a limiting factor to use that internal version of pyarrow's orc writer. 

> Would it be possible to try this out in fuzz tests to verify that pyarrow is robust?

Yup

vuule:
I definitely like the suggestion, pyarrow API looks very clean and... comprehensive (more so than ours 😬).

galipremsagar:
https://github.com/rapidsai/cudf/pull/12103 solves part of the problem. However we will need to wait until pyarrow can write complex nested data-types to an orc file.

vyasr:
This was completed in #14323

PR #12103, opened 2022-11-09
[REVIEW] First pass of `pd.read_orc` changes in tests

galipremsagar:
## Description
This PR changes calls going via `pyarrow` and then `to_pandas` to directly call `pd.read_orc`. How-ever since `pd.read_orc` was added in pandas 1.0, we will need to version the call to this constructor. This PR does that.

Partially contributes to #11540 
## Checklist

### DIFF
--- a/python/cudf/cudf/tests/test_orc.py
+++ b/python/cudf/cudf/tests/test_orc.py
@@ -18,7 +18,6 @@ import cudf
 from cudf.io.orc import ORCWriter
 from cudf.testing import assert_frame_equal
 from cudf.testing._utils import (
-    _pandas_read_orc,
     assert_eq,
     gen_rand_series,
     supported_numpy_dtypes,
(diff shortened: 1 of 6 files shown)


python/cudf/cudf/tests/test_orc.py
```diff
@@ -1696,7 +1607,7 @@ def test_orc_writer_rle_stream_size(datadir, tmpdir):
 
     # Segfaults when RLE stream sizes don't account for varint length
     pa_out = pa.orc.ORCFile(reencoded).read()
-    assert_eq(df.to_pandas(), pa_out)
+    assert df.to_arrow().equals(pa_out)
```
vuule, line 1610: looks like this was a different kind of redundant casting?

vuule (approved):
Like the change!

Can we expect noticeably lower test execution time with this change? I don't have a good sense of how efficient the conversion is(was).

python/cudf/cudf/tests/test_orc.py
galipremsagar: Yup. We were going to pandas for comparison previously. Now we just do arrow level comparison.

wence- (approved):
Thanks! My only comment is whether we need to support pandas pre 1.0 at all?

galipremsagar:
> Thanks! My only comment is whether we need to support pandas pre 1.0 at all?

Good catch. Rapids only support pandas >=1.0, so dropped the pre 1.0 code.

galipremsagar:
@gpucibot merge

PR #12110, opened 2022-11-09
Create an `int8` column in `read_csv` when all elements are missing

vuule:
## Description
CSV reader creates int8 columns when all elements are null. However, when all elements in a 
column are missing (e.g. `names` option includes more columns than the CSV file), CSV reader creates an `int64` column. Such columns take up a lot more device memory.

This PR changes the behavior so that all columns with no valid elements are created as `int8`.

## Checklist


hyperbolic2346 (approved):

cpp/tests/io/csv_test.cpp
```diff
@@ -2244,6 +2244,27 @@ TEST_F(CsvReaderTest, CsvDefaultOptionsWriteReadMatch)
...
+      .header(-1);
+  // More elements in `names` than in the file; additional columns are filled with nulls
+  auto result = cudf::io::read_csv(in_opts);
+
+  const auto result_table = result.tbl->view();
+  ASSERT_EQ(result_table.num_columns(), 4);
+  // All columns should contain only nulls; expect INT8 type to use as little memory as possible
+  for (auto& column : result_table) {
+    ASSERT_EQ(column.type(), data_type{type_id::INT8});
+    ASSERT_EQ(column.null_count(), 2);
```
PointKernel, line 2264: I was wondering what is the difference between `EXPECT_EQ` and `ASSERT_EQ` since the latter is more seen in our tests. It looks like the two are really the same according to [user guide](http://google.github.io/googletest/reference/assertions.html#EXPECT_EQ).

PointKernel (approved):
Nice small fix. Thanks!

cpp/src/io/utilities/column_type_histogram.hpp
```diff
@@ -33,6 +33,11 @@ struct column_type_histogram {
   cudf::size_type positive_small_int_count{};
   cudf::size_type big_int_count{};
   cudf::size_type bool_count{};
+  auto total_count() const
```
harrism, line 36: Perhaps for the future -- seems like this struct would be more versatile and more easily extendible (and totals easier to calculate in a bug-free way) if you store a `std::array` (or similar) of counts and index it using constant names for indices. Then this line could just be a `std::accumulate` (or similar).

harrism (approved):
Looks good.

vuule:
@gpucibot merge

PR #12113, opened 2022-11-10
[Java] Add JNI for `substring` without 'end' parameter.

firestarman:
cudf `substring(column, int, int)` has its own behavior when `end` is -1 (a valid position pointing to the last char), which is different from that in JNI. So JNI should not override it.

Besides, JNI's behavior does not match its doc when `end` is "-1". According to the doc, the last char should not be included in the result substring. But currently it does. 

Signed-off-by: Liangcai Li <liangcail@nvidia.com>

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/ColumnView.java
+++ b/java/src/main/java/ai/rapids/cudf/ColumnView.java
@@ -3985,7 +3985,7 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
   private static native long substring(long columnView, int start, int end) throws CudfException;
 
   /**
-   * Native method to calculate substring from a given string column. 0 indexing.
+   * Native method to extract substrings from a given strings column.
    * @param columnView native handle of the cudf::column_view being operated on.
    * @param start      first character index to begin the substrings (inclusive).
    */
--- a/java/src/main/java/ai/rapids/cudf/ColumnView.java
+++ b/java/src/main/java/ai/rapids/cudf/ColumnView.java
@@ -2632,12 +2632,13 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
 
   /**
    * Returns a new strings column that contains substrings of the strings in the provided column.
-   * Overloading subString to support if end index is not provided. Appending -1 to indicate to
-   * read until end of string.
+   * The character positions to retrieve in each string are `[start, <the string end>)`..
+   *
    * @param start first character index to begin the substring(inclusive).
    */
   public final ColumnVector substring(int start) {
-    return substring(start, -1);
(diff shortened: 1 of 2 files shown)


firestarman:
NOTE: This is a breaking change, and I have filed PR https://github.com/NVIDIA/spark-rapids/pull/7040 for the plugin.

firestarman:
> Could you please describe the behaviour you want in the PR description? Besides, as this is a breaking change (not a bug), could you please explain why you can't do the change in spark-plugin, but in cudf?

The doc of `substring` is what I want. I think it is a bug in JNI, so I changed it in JNI.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -3983,6 +3984,13 @@ private static native long stringSplitRecord(long nativeHandle, String pattern,
    */
   private static native long substring(long columnView, int start, int end) throws CudfException;
 
+  /**
+   * Native method to calculate substring from a given string column. 0 indexing.
```
ttnghia, line 3988:
```suggestion
   * Native method to extract substrings from a given strings column.
```

java/src/main/native/src/ColumnViewJni.cpp
```diff
@@ -1397,6 +1397,18 @@ JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_binaryOpVS(JNIEnv *env, j
   CATCH_STD(env, 0);
 }
 
+JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_substringS(JNIEnv *env, jclass,
+                                                                  jlong column_view, jint start) {
+  JNI_NULL_CHECK(env, column_view, "column is null", 0);
+  try {
+    cudf::jni::auto_set_device(env);
+    cudf::column_view *cv = reinterpret_cast<cudf::column_view *>(column_view);
+    cudf::strings_column_view scv(*cv);
```
ttnghia, line 1406:
```suggestion
JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_substringS(JNIEnv *env, jclass,
                                                                  jlong cv_handle, jint start) {
  JNI_NULL_CHECK(env, cv_handle, "column is null", 0);
  try {
    cudf::jni::auto_set_device(env);
    auto const cv = reinterpret_cast<cudf::column_view const *>(cv_handle);
    auto const scv = cudf::strings_column_view{*cv};
```

revans2 (approved):
looks good to me but it would be nice to address the comments by @ttnghia

ttnghia (approved):

firestarman:
@gpucibot merge

PR #12117, opened 2022-11-10
Support `+` in `strings_udf`

brandon-b-miller:
This PR adds support for the following operator `strings_udf`:

- `st + other`


brandon-b-miller:
Note for reviewers: one might ask themselves, where is the lowering for this operator for `MaskedType`? and how does it work without it? Well, `+` is part of our `binary_ops` and in our normal lowering for `MaskedType` within `masked_lowering.py`, `+` is already registered between instances of `MaskedType`. Numba hooks into this lowering, and then ends up finding the `+` operator already registered between the two `.value_type`s and just calls that. So, no need for explicit lowering for concat here.

brandon-b-miller:
rerun tests

davidwendt (approved):
Approving C++ code.

python/strings_udf/strings_udf/_typing.py
```diff
@@ -159,8 +159,13 @@ def generic(self, args, kws):
 register_stringview_binaryop(operator.gt, types.boolean)
 register_stringview_binaryop(operator.le, types.boolean)
 register_stringview_binaryop(operator.ge, types.boolean)
+
+# st in other
 register_stringview_binaryop(operator.contains, types.boolean)
 
+# st + other
+register_stringview_binaryop(operator.add, udf_string)
```
bdice, line 167: Does this work for reverse addition? Would `concat_char + row["str_col"]` dispatch differently?

python/strings_udf/strings_udf/tests/test_string_udfs.py
```diff
@@ -302,3 +302,11 @@ def func(st):
         return st.rstrip(strip_char)
 
     run_udf_test(data, func, "str")
+
+
+@pytest.mark.parametrize("concat_char", ["1", "a", "12", " ", "", ".", "@"])
+def test_string_udf_concat(data, concat_char):
```
bdice, line 308: Maybe add tests for reversed concatenation like `concat_char + st`?

bdice (approved):
LGTM. Possibly needs another test but I'll let you make that call.

brandon-b-miller:
@gpucibot merge

PR #12125, opened 2022-11-10
[JNI] Adds an EventHandler to Java MemoryBuffer to be invoked on close

abellina:
This PR adds an EventHandler to `MemoryBuffer` with a single method `onClosed`. This is invoked during the `close` call, but after the `refCount` has been updated.

I am also making `getRefCount` public in this PR. Spill code in the RAPIDS Accelerator for Spark could likely assert/require that refCount==1 when taking in a new buffer to be spillable. This last change is a nice to have.

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/MemoryBuffer.java
+++ b/java/src/main/java/ai/rapids/cudf/MemoryBuffer.java
@@ -1,6 +1,6 @@
 /*
  *
- *  Copyright (c) 2019-2020, NVIDIA CORPORATION.
+ *  Copyright (c) 2019-2022, NVIDIA CORPORATION.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


revans2 (approved):

jbrennan333 (approved):
+1 lgtm

gerashegalov:
Should we start exploring Java Instrumentation API such that we can keep auxiliary monitoring code out the production codebase similar to CUPTI

abellina:
> Should we start exploring Java Instrumentation API such that we can keep auxiliary monitoring code out the production codebase similar to CUPTI

In this scenario this is production code. We expect to use this in the spill framework.

java/src/main/java/ai/rapids/cudf/MemoryBuffer.java
```diff
@@ -193,13 +209,35 @@ public final void copyFromMemoryBufferAsync(
...
   public abstract MemoryBuffer slice(long offset, long len);
 
+  /**
+   * Set an event handler for this buffer. This method can be invoked with null
+   * to unset the handler.
+   */
+  public synchronized void setEventHandler(EventHandler handler) {
+    if (this.eventHandler != null && handler != null) {
+      throw new IllegalStateException("EventHandler is already set for this buffer");
+    }
```
jlowe, line 219: IMO this should not check or throw.  Instead this should return the old handler and let the caller decide if they want to assert or do other logic.

jlowe (approved):

abellina:
@gpucibot merge

PR #12127, opened 2022-11-10
Support string indexing in `strings_udf`

brandon-b-miller:
This PR adds support for the following operator `strings_udf`:

- `st[idx]`


brandon-b-miller:
Something I didn't fully consider when first mocking that especially with string indexing, the error cases are really important. In python for instance we get an out of range index error for indices beyond the length of the string. A UDF in pandas encountering this would just throw during the loop over the data, but in our case though since we're executing the equivalent code at the thread level and can't raise in the same way. 

I think this could be a little dangerous as it could hide logic errors in the UDF that pandas highlights by failing and the user could have their UDF compile and run and just get garbage values for strings that would have thrown in python. 

We could maybe fix this using a similar approach to what we started discussing in https://github.com/rapidsai/cudf/issues/8774. Until then we're going to be missing some important functionality here.

vyasr:
We're going to want to revisit this soon since we can pick up work on more of #9639 after #18453 is in, but those are pretty massive changes so it will make sense to start fresh.

PR #12131, opened 2022-11-11
Floor division uses integer division for integral arguments

### LINKED ISSUE #12120: [BUG] incorrect floor division in libcudf binarop (coming from CUDF cython)
**Describe the bug**

Floor division of large integers loses precision because it looks like it goes through floating point division somewhere.

```python
import cudf
import cupy
from cudf._lib.binaryop import binaryop
import numpy as np
s = cudf.Series([12], dtype="int64")
v = np.int64(1513393355500000000)
divided = cupy.asarray(binaryop(v, s._column, "__floordiv__", np.int64))
assert divided[0] == (v // 12) # => false
assert divided[0] == (v / 12).astype(np.int64) # => true, hmm suspicious
```

**Expected behavior**

Floor division should work with integers and not pass through a floating point representation.

wence-:
## Description

Previously (since c8b897ccd) floor division for all pairs of arguments
was implemented by promoting both arguments to double, dividing, and
then taking the floor.

This produces results when both operands are integral that sometimes
don't match the behaviour of normal floor division in any of the
target languages that cuDF supports. Specifically for int64 operands,
any time an operand is larger than $2^{53}$, the conversion to double is
not injective and so information is lost.

To address this, implement SFINAE overloads for the FloorDiv device
functor that use standard C++ type promotion semantics for division
with mixed type operands. In the case of integral types with same
sign, this is then just standard division; if the signs are mixed, we
round towards negative infinity; if either type is floating, we
perform floating point division and then take the floor.

These operations are now explicitly not implemented for Decimal and
duration dtypes (rather than implicitly not implemented in the wrapper
layers).

If the calling library's requirements for type promotion differ from
those of C++, it is the now the responsibility of the caller to
promote the operands to floor division before calling into libcudf.

Closes #12120.

## Checklist

### DIFF
--- a/cpp/src/binaryop/compiled/operation.cuh
+++ b/cpp/src/binaryop/compiled/operation.cuh
@@ -99,7 +99,13 @@ struct FloorDiv {
                       is_duration<TypeLhs>())>* = nullptr>
   __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> decltype(x / y)
   {
-    return x / y;
+    if ((x ^ y) >= 0) {
+      return x / y;
+    } else {
+      const auto quotient  = x / y;
+      const auto remainder = x % y;
+      return quotient - !!remainder;
+    }
   }
 
   template <
--- a/cpp/src/binaryop/compiled/operation.cuh
+++ b/cpp/src/binaryop/compiled/operation.cuh
@@ -98,7 +98,7 @@ struct FloorDiv {
   __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> decltype(x / y)
   {
     auto const quotient          = x / y;
-    auto const nonzero_remainder = !!(x % y);
+    auto const nonzero_remainder = (x % y) != 0;
     auto const mixed_sign        = (x ^ y) < 0;
     return quotient - mixed_sign * nonzero_remainder;
   }
--- a/cpp/src/binaryop/compiled/operation.cuh
+++ b/cpp/src/binaryop/compiled/operation.cuh
@@ -91,12 +91,10 @@ struct TrueDiv {
 };
 
 struct FloorDiv {
-  template <
-    typename TypeLhs,
-    typename TypeRhs,
-    std::enable_if_t<(std::is_integral_v<std::common_type_t<TypeLhs, TypeRhs>> or
-                      (cudf::is_fixed_point<TypeLhs>() and std::is_same_v<TypeLhs, TypeRhs>) or
-                      is_duration<TypeLhs>())>* = nullptr>
+  template <typename TypeLhs,
+            typename TypeRhs,
+            std::enable_if_t<(std::is_integral_v<std::common_type_t<TypeLhs, TypeRhs>> and
+                              std::is_signed_v<std::common_type_t<TypeLhs, TypeRhs>>)>* = nullptr>
   __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> decltype(x / y)
   {
     if ((x ^ y) >= 0) {
@@ -106,6 +104,15 @@ struct FloorDiv {
       const auto remainder = x % y;
       return quotient - !!remainder;
     }
+  }
+
+  template <typename TypeLhs,
+            typename TypeRhs,
+            std::enable_if_t<(std::is_integral_v<std::common_type_t<TypeLhs, TypeRhs>> and
+                              !std::is_signed_v<std::common_type_t<TypeLhs, TypeRhs>>)>* = nullptr>
+  __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> decltype(x / y)
+  {
+    return x / y;
   }
 
   template <
--- a/cpp/src/binaryop/compiled/operation.cuh
+++ b/cpp/src/binaryop/compiled/operation.cuh
@@ -100,8 +100,8 @@ struct FloorDiv {
     if ((x ^ y) >= 0) {
       return x / y;
     } else {
-      const auto quotient  = x / y;
-      const auto remainder = x % y;
+      auto const quotient  = x / y;
+      auto const remainder = x % y;
       return quotient - !!remainder;
     }
   }
--- a/cpp/src/binaryop/compiled/operation.cuh
+++ b/cpp/src/binaryop/compiled/operation.cuh
@@ -97,13 +97,10 @@ struct FloorDiv {
                               std::is_signed_v<std::common_type_t<TypeLhs, TypeRhs>>)>* = nullptr>
   __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> decltype(x / y)
   {
-    if ((x ^ y) >= 0) {
-      return x / y;
-    } else {
-      auto const quotient  = x / y;
-      auto const remainder = x % y;
-      return quotient - !!remainder;
-    }
+    auto const quotient          = x / y;
+    auto const nonzero_remainder = !!(x % y);
+    auto const mixed_sign        = (x ^ y) < 0;
+    return quotient - mixed_sign * nonzero_remainder;
   }
 
   template <typename TypeLhs,
--- a/cpp/src/binaryop/compiled/operation.cuh
+++ b/cpp/src/binaryop/compiled/operation.cuh
@@ -91,11 +91,33 @@ struct TrueDiv {
 };
 
 struct FloorDiv {
-  template <typename T1, typename T2>
-  __device__ inline auto operator()(T1 const& lhs, T2 const& rhs)
-    -> decltype(floor(static_cast<double>(lhs) / static_cast<double>(rhs)))
+  template <
+    typename TypeLhs,
+    typename TypeRhs,
+    std::enable_if_t<(std::is_integral_v<std::common_type_t<TypeLhs, TypeRhs>> or
+                      (cudf::is_fixed_point<TypeLhs>() and std::is_same_v<TypeLhs, TypeRhs>) or
+                      is_duration<TypeLhs>())>* = nullptr>
+  __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> decltype(x / y)
   {
-    return floor(static_cast<double>(lhs) / static_cast<double>(rhs));
+    return x / y;
+  }
+
+  template <
+    typename TypeLhs,
+    typename TypeRhs,
+    std::enable_if_t<(std::is_same_v<std::common_type_t<TypeLhs, TypeRhs>, float>)>* = nullptr>
+  __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> float
+  {
+    return floorf(x / y);
+  }
+
+  template <
+    typename TypeLhs,
+    typename TypeRhs,
+    std::enable_if_t<(std::is_same_v<std::common_type_t<TypeLhs, TypeRhs>, double>)>* = nullptr>
+  __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> double
+  {
+    return floor(x / y);
   }
 };
 
--- a/cpp/tests/binaryop/binop-compiled-test.cpp
+++ b/cpp/tests/binaryop/binop-compiled-test.cpp
@@ -27,6 +27,7 @@
 #include <cudf_test/type_lists.hpp>
 
 #include <cudf/utilities/error.hpp>
+#include <limits>
 #include <tests/binaryop/assert-binops.h>
 #include <tests/binaryop/binop-fixture.hpp>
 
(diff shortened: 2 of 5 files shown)


wence-:
TODO:
- [ ] Add C++ tests
- [ ] Add Python tests
- [ ] Confirm this is right for SPARK

wence-:
I can spin a4704c62b off into a separate PR if that's useful.

wence-:
This will probably need to sit on top of #12074

cpp/src/binaryop/compiled/operation.cuh
```diff
@@ -91,11 +91,33 @@ struct TrueDiv {
...
-    -> decltype(floor(static_cast<double>(lhs) / static_cast<double>(rhs)))
+  template <
+    typename TypeLhs,
+    typename TypeRhs,
+    std::enable_if_t<(std::is_integral_v<std::common_type_t<TypeLhs, TypeRhs>> or
+                      (cudf::is_fixed_point<TypeLhs>() and std::is_same_v<TypeLhs, TypeRhs>) or
+                      is_duration<TypeLhs>())>* = nullptr>
+  __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> decltype(x / y)
+  {
+    return x / y;
```
wence-, line 102: This is not right for negative values, since in C++ integer division rounds towards zero, rather than -infinity (as required in Python).

This would do the trick:

```suggestion
    if ((x ^ y) >= 0) {
      return x / y;
    } else {
       auto quot = x / y;
       return quot - !!(x - quot * b);
    }
```

We should then probably narrow the type-overload to only do this in the case of signed types (although GCC can optimise to `x / y` in the case that both types are unsigned (for which `(x^y) >= 0` is always true)).

Question: is floor division like this for division of integer columns actually what is wanted in SPARK/Java. Java at least has the same semantics as C++ (round to zero)

wence-:
@revans2 This will change the behaviour for SPARK cases (I _think_ correcting some previous bugs), but would like some confirmation (or some advice as to check that libcudf still does the right thing here).

cpp/src/binaryop/compiled/operation.cuh
```diff
@@ -91,11 +91,46 @@ struct TrueDiv {
...
+            typename TypeRhs,
+            std::enable_if_t<(std::is_integral_v<std::common_type_t<TypeLhs, TypeRhs>> and
+                              std::is_signed_v<std::common_type_t<TypeLhs, TypeRhs>>)>* = nullptr>
+  __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> decltype(x / y)
+  {
+    if ((x ^ y) >= 0) {
+      return x / y;
+    } else {
+      const auto quotient  = x / y;
+      const auto remainder = x % y;
```
PointKernel, line 104:
```suggestion
      auto const quotient  = x / y;
      auto const remainder = x % y;
```
nit: we prefer east const throughout libcudf

cpp/src/binaryop/compiled/operation.cuh
```diff
@@ -91,11 +91,46 @@ struct TrueDiv {
...
+      return quotient - !!remainder;
```
PointKernel, line 105: For my own education, what is going on here? :exploding_head:

cpp/src/binaryop/compiled/operation.cuh
```diff
@@ -91,11 +91,46 @@ struct TrueDiv {
...
+    return x / y;
+  }
+
+  template <
+    typename TypeLhs,
+    typename TypeRhs,
+    std::enable_if_t<(std::is_same_v<std::common_type_t<TypeLhs, TypeRhs>, float>)>* = nullptr>
+  __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> float
+  {
+    return floorf(x / y);
```
PointKernel, line 124:
```suggestion
    return std::floorf(x / y);
```

cpp/src/binaryop/compiled/operation.cuh
```diff
@@ -91,11 +91,46 @@ struct TrueDiv {
...
+  }
+
+  template <
+    typename TypeLhs,
+    typename TypeRhs,
+    std::enable_if_t<(std::is_same_v<std::common_type_t<TypeLhs, TypeRhs>, double>)>* = nullptr>
+  __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> double
   {
-    return floor(static_cast<double>(lhs) / static_cast<double>(rhs));
+    return floor(x / y);
```
PointKernel, line 133:
```suggestion
    return std::floor(x / y);
```

cpp/tests/binaryop/binop-compiled-test.cpp
```diff
@@ -27,6 +27,7 @@
 #include <cudf_test/type_lists.hpp>
 
 #include <cudf/utilities/error.hpp>
+#include <limits>
```
PointKernel, line 30:
```suggestion
```
This should be replaced right before `#include <type_traits>` without a blank line. They both belong to the STL header group.

cpp/tests/binaryop/util/operation.h
```diff
@@ -143,9 +143,48 @@ struct TrueDiv {
...
+            std::enable_if_t<(std::is_integral_v<std::common_type_t<LhsT, RhsT>> and
+                              std::is_signed_v<std::common_type_t<LhsT, RhsT>>)>* = nullptr>
   TypeOut operator()(TypeLhs lhs, TypeRhs rhs)
   {
-    return static_cast<TypeOut>(floor(static_cast<double>(lhs) / static_cast<double>(rhs)));
+    if ((lhs ^ rhs) >= 0) {
+      return lhs / rhs;
+    } else {
+      const auto quotient  = lhs / rhs;
+      const auto remainder = lhs % rhs;
```
PointKernel, line 157:
```suggestion
      auto const quotient  = lhs / rhs;
      auto const remainder = lhs % rhs;
```
east const

cpp/tests/binaryop/util/operation.h
```diff
@@ -143,9 +143,48 @@ struct TrueDiv {
...
+    return lhs / rhs;
+  }
+
+  template <typename OutT                                                              = TypeOut,
+            typename LhsT                                                              = TypeLhs,
+            typename RhsT                                                              = TypeRhs,
+            std::enable_if_t<(std::is_same_v<std::common_type_t<LhsT, RhsT>, float>)>* = nullptr>
+  TypeOut operator()(TypeLhs lhs, TypeRhs rhs)
+  {
+    return static_cast<TypeOut>(floorf(lhs / rhs));
```
PointKernel, line 178:
```suggestion
    return static_cast<TypeOut>(std::floorf(lhs / rhs));
```

cpp/tests/binaryop/util/operation.h
```diff
@@ -143,9 +143,48 @@ struct TrueDiv {
...
+  }
+
+  template <typename OutT                                                               = TypeOut,
+            typename LhsT                                                               = TypeLhs,
+            typename RhsT                                                               = TypeRhs,
+            std::enable_if_t<(std::is_same_v<std::common_type_t<LhsT, RhsT>, double>)>* = nullptr>
+  TypeOut operator()(TypeLhs lhs, TypeRhs rhs)
+  {
+    return static_cast<TypeOut>(floor(lhs / rhs));
```
PointKernel, line 187:
```suggestion
    return static_cast<TypeOut>(std::floor(lhs / rhs));
```

cpp/src/binaryop/compiled/operation.cuh
bdice: Careful - we may want the CUDA functions instead of std:: functions.

PointKernel: Ah, really good point!

PointKernel: @bdice Great catch 👍. We should use built-in CUDA math functions on the device as opposed to STL. I will discard all similar change requests.

cpp/tests/binaryop/binop-compiled-test.cpp
wence-: I wonder why clang-format did not do this for me (or at least complain)

wence-: Ah, it only sorts in blocks.

cpp/src/binaryop/compiled/operation.cuh
wence-: `!!` is a function $!! : \mathbb{N} \to {0, 1}$ with behaviour

$$
!! x = \begin{cases} 0 &\text{ if $ x = 0$} \\
1 & \text{ otherwise}
 \end{cases}
$$

If we have mixed sign arguments for floor division, if we have a non-zero remainder from the division we need to round towards negative infinity by subtracting 1 (this is a convenient branch-free way of getting 1).

cpp/src/binaryop/compiled/operation.cuh
wence-, line 104: It might be better to implement this branch-free:

```c++
auto const quotient = x / y;
auto const nonzero_remainder = !!(x % y);
auto const mixed_sign = (x ^ y) < 0;
return quotient - mixed_sign * nonzero_remainder;
```

python/cudf/cudf/tests/test_binops.py
```diff
@@ -881,12 +880,27 @@ def gen_df():
...
+                reason="https://github.com/rapidsai/cudf/issues/12162"
+            ),
+        ),
+        pytest.param(
+            "rfloordiv",
+            marks=pytest.mark.xfail(
+                reason="https://github.com/rapidsai/cudf/issues/12162"
+            ),
+        ),
+    ],
```
wence-, line 900: Not changing the behaviour of the test except to loudly xfail, rather than silently skip.

python/cudf/cudf/tests/test_binops.py
```diff
@@ -895,19 +909,11 @@ def test_binop_bool_uint(func, rhs):
...
     "divisor_dtype",
     (
-        pytest.param(
-            np.bool_,
-            marks=pytest_xfail(
-                reason=(
-                    "Pandas handling of division by zero-bool is too strange"
-                )
-            ),
-        ),
```
wence-, line 910: Moving the tests of bool operands into a separate test to xfail in a more fine-grained manner.

cpp/src/binaryop/compiled/operation.cuh
```diff
@@ -91,11 +91,46 @@ struct TrueDiv {
...
+  __device__ inline auto operator()(TypeLhs x, TypeRhs y) -> decltype(x / y)
+  {
+    if ((x ^ y) >= 0) {
+      return x / y;
+    } else {
+      auto const quotient  = x / y;
+      auto const remainder = x % y;
+      return quotient - !!remainder;
+    }
+  }
```
wence-, line 107: A branch-free implementation looks like it performs better across the board:

```
gbench/tools/compare.py -a --no-utest benchmarks cudf/cpp/build/benchmarks/branchy.json cudf/cpp/build/benchmarks/nonbranchy.json  
Comparing cudf/cpp/build/benchmarks/branchy.json to cudf/cpp/build/benchmarks/nonbranchy.json
Benchmark                                                                                          Time             CPU      Time Old      Time New       CPU Old       CPU New
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/10000/manual_time_mean                      -0.0282         -0.0244            43            41            63            62
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/10000/manual_time_median                    -0.0286         -0.0249            43            41            63            62
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/10000/manual_time_stddev                    +0.2685         +0.4944             0             0             0             0
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/100000/manual_time_mean                     -0.0365         -0.0300            46            44            66            64
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/100000/manual_time_median                   -0.0365         -0.0300            46            44            66            64
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/100000/manual_time_stddev                   -0.0921         -0.1937             0             0             0             0
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/1000000/manual_time_mean                    -0.0682         -0.0562            80            74           100            95
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/1000000/manual_time_median                  -0.0683         -0.0562            80            74           100            95
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/1000000/manual_time_stddev                  -0.0394         -0.1046             0             0             0             0
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/10000000/manual_time_mean                   -0.0862         -0.0828           425           388           446           410
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/10000000/manual_time_median                 -0.0862         -0.0827           425           388           446           410
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/10000000/manual_time_stddev                 +0.7530         -0.0782             0             0             0             0
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/100000000/manual_time_mean                  -0.1000         -0.1000          3722          3350          3763          3387
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/100000000/manual_time_median                -0.1000         -0.1000          3722          3350          3763          3386
COMPILED_BINARYOP/FLOOR_DIV_int64_t_int64_t_int64_t/100000000/manual_time_stddev                +0.0632         +0.1715             0             0             1             1
```

cpp/src/binaryop/compiled/operation.cuh
bdice: Nice trick. A comment in the code would be appropriate here.

wence-: I've reworked this entire function to be branch-free and moved to less tricksy code (`remainder != 0`) which compiles to the same PTX.

cpp/src/binaryop/compiled/operation.cuh
wence-: It was.

cpp/src/binaryop/compiled/operation.cuh
PointKernel: Thread divergence could easily be the number one performance killer for such a small kernel. The new branchless algorithm looks great.

cpp/tests/binaryop/util/operation.h
```diff
@@ -143,9 +143,48 @@ struct TrueDiv {
...
   TypeOut operator()(TypeLhs lhs, TypeRhs rhs)
   {
-    return static_cast<TypeOut>(floor(static_cast<double>(lhs) / static_cast<double>(rhs)));
+    if ((lhs ^ rhs) >= 0) {
+      return lhs / rhs;
+    } else {
+      auto const quotient  = lhs / rhs;
+      auto const remainder = lhs % rhs;
+      return quotient - !!remainder;
+    }
```
PointKernel, line 159: Will the host implementation also benefit from the branchless algorithm?

cpp/tests/binaryop/util/operation.h
```diff
@@ -143,9 +143,48 @@ struct TrueDiv {
...
+    return lhs / rhs;
+  }
+
+  template <typename OutT                                                              = TypeOut,
+            typename LhsT                                                              = TypeLhs,
+            typename RhsT                                                              = TypeRhs,
+            std::enable_if_t<(std::is_same_v<std::common_type_t<LhsT, RhsT>, float>)>* = nullptr>
+  TypeOut operator()(TypeLhs lhs, TypeRhs rhs)
+  {
+    return static_cast<TypeOut>(std::floor(lhs / rhs));
```
PointKernel, line 178:
```suggestion
    return static_cast<TypeOut>(std::floorf(lhs / rhs));
```

cpp/tests/binaryop/util/operation.h
wence-: This is only a test utility, that really should be expunged completely (see #12241), so I'm not worried about performance.

cpp/tests/binaryop/util/operation.h
wence-: Per https://en.cppreference.com/w/cpp/numeric/math/floor I think there's an overload for `std::floor` with a float argument?

python/cudf/cudf/tests/test_binops.py
wence-: Moved this into a separate branch.

python/cudf/cudf/tests/test_binops.py
wence-: Moved this into a separate branch.

cpp/tests/binaryop/util/operation.h
PointKernel: Yeah, my intention is to use a more descriptive name so people know it's dedicated to `float` operations. Not necessarily a change request.

PointKernel (approved):
CPP changes look good to me.

cpp/src/binaryop/compiled/operation.cuh
harrism: What are you comparing against? The old version doesn't appear to have branches either.

wence-: The previous implementation on this branch: 

https://github.com/rapidsai/cudf/pull/12131/commits/bd5bc8895c4256760b06fe6778485bddab4388e9#diff-783a8c11b115684b22107cfc1316116a7ea17b6e5df50042f31627ff1869d34eR102

bdice (approved):
Looks good to me! Thanks @wence-.

hyperbolic2346 (approved):

wence-:
@gpucibot merge

PR #12138, opened 2022-11-14
Safely allocate `udf_string` pointers in `strings_udf`

brandon-b-miller:
In `strings_udf`, functions that return strings are built around c++ methods that return a `cudf::strings::udf::udf_string` object. However due to requiring external `C` linkage, our shim functions need to work by accepting a pointer to a preallocated `udf_string` object and setting the result into the memory it points to before returning. 

This piece of memory is allocated based off our `UDFString` extension class datamodel and while it is set up to be the right size, simply allocating it does not actually call the underlying `udf_string` default constructor so the memory isn't necessarily initialized in the same way a proper `udf_string` would initialize it. 

This can lead to some unsafe behavior when we try and assign the result. This PR changes it so that whenever we need to allocate a `udf_string` and pass its pointer to a shim function, we first zfill that memory.

### DIFF
--- a/python/strings_udf/cpp/src/strings/udf/shim.cu
+++ b/python/strings_udf/cpp/src/strings/udf/shim.cu
@@ -223,7 +223,7 @@ extern "C" __device__ int udf_string_from_string_view(int* nb_retbal,
                                                       void* udf_str)
 {
   auto str_view_ptr = reinterpret_cast<cudf::string_view const*>(str);
-  auto udf_str_ptr  = reinterpret_cast<udf_string*>(udf_str);
+  auto udf_str_ptr  = new (udf_str) udf_string;
   *udf_str_ptr      = udf_string(*str_view_ptr);
 
   return 0;
@@ -236,7 +236,7 @@ extern "C" __device__ int strip(int* nb_retval,
 {
   auto to_strip_ptr  = reinterpret_cast<cudf::string_view const*>(to_strip);
   auto strip_str_ptr = reinterpret_cast<cudf::string_view const*>(strip_str);
-  auto udf_str_ptr   = reinterpret_cast<udf_string*>(udf_str);
+  auto udf_str_ptr   = new (udf_str) udf_string;
 
   *udf_str_ptr = strip(*to_strip_ptr, *strip_str_ptr);
 
@@ -250,7 +250,7 @@ extern "C" __device__ int lstrip(int* nb_retval,
 {
   auto to_strip_ptr  = reinterpret_cast<cudf::string_view const*>(to_strip);
   auto strip_str_ptr = reinterpret_cast<cudf::string_view const*>(strip_str);
-  auto udf_str_ptr   = reinterpret_cast<udf_string*>(udf_str);
+  auto udf_str_ptr   = new (udf_str) udf_string;
 
   *udf_str_ptr = strip(*to_strip_ptr, *strip_str_ptr, cudf::strings::side_type::LEFT);
 
@@ -264,7 +264,7 @@ extern "C" __device__ int rstrip(int* nb_retval,
 {
   auto to_strip_ptr  = reinterpret_cast<cudf::string_view const*>(to_strip);
   auto strip_str_ptr = reinterpret_cast<cudf::string_view const*>(strip_str);
-  auto udf_str_ptr   = reinterpret_cast<udf_string*>(udf_str);
+  auto udf_str_ptr   = new (udf_str) udf_string;
 
   *udf_str_ptr = strip(*to_strip_ptr, *strip_str_ptr, cudf::strings::side_type::RIGHT);
 
--- a/python/strings_udf/strings_udf/lowering.py
+++ b/python/strings_udf/strings_udf/lowering.py
@@ -115,7 +115,9 @@ def cast_string_literal_to_string_view(context, builder, fromty, toty, val):
 @cuda_lowering_registry.lower_cast(string_view, udf_string)
 def cast_string_view_to_udf_string(context, builder, fromty, toty, val):
     sv_ptr = builder.alloca(default_manager[fromty].get_value_type())
-    udf_str_ptr = builder.alloca(default_manager[toty].get_value_type())
+    udf_str_ptr = cgutils.alloca_once(
+        builder, default_manager[toty].get_value_type(), zfill=True
+    )
     builder.store(val, sv_ptr)
     _ = context.compile_internal(
         builder,
@@ -196,8 +198,10 @@ def create_binary_string_func(binary_func, retty):
                 # value of compile_internal is therefore discarded (although
                 # this may change in the future if we need to return error
                 # codes, for instance).
-                udf_str_ptr = builder.alloca(
-                    default_manager[udf_string].get_value_type()
+                udf_str_ptr = cgutils.alloca_once(
+                    builder,
+                    default_manager[udf_string].get_value_type(),
+                    zfill=True,
                 )
 
                 _ = context.compile_internal(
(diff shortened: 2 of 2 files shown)


gmarkall:
> > This piece of memory is allocated based off our UDFString extension class datamodel and while it is set up to be the right size, simply allocating it does not actually call the underlying udf_string default constructor so the memory isn't necessarily initialized in the same way a proper udf_string would initialize it.
> 
> Simply zeroing the bits isn't valid. You should use [`placement new`](https://www.geeksforgeeks.org/placement-new-operator-cpp/) to construct the object into already allocated memory.

My understanding is that the default constructor would initialize `m_data`, `m_bytes`, and `m_capacity` to `nullptr`, `0`, and `0` respectively, so because the `udf_string` is created on the Python side, this implements the same operations as the default constructor on the Python side by effectively zeroing the fields to get the same result. An alternative could also be to create a struct proxy for the `string_udf` instance then set the fields in the Numba-generated code like:

```python
udf_str.data = context.get_constant(types.int64, 0)
udf_str.length = context.get_constant(types.int32, 0)
udf_str.bytes = context.get_constant(types.int32, 0)
```

and perhaps this would be a bit more clear what's going on at a high level, yet should result in the same generated code - maybe this would be better from a clarity / maintainability point of view anyway.

davidwendt:
> My understanding is that the default constructor would initialize `m_data`, `m_bytes`, and `m_capacity` to `nullptr`, `0`, and `0` respectively, so because the `udf_string` is created on the Python side, this implements the same operations as the default constructor on the Python side by effectively zeroing the fields to get the same result.

This is a bit fragile since the default constructor logic and members could/will change in the future and we'd have to keep the Python logic in sync somehow. It would be better to use the actual constructor on the existing memory using placement new.

wence-:
> `udf_str.data = context.get_constant(types.int64, 0)`

 My language-lawyering is always a bit weak, but if you're trying to initialise (from python) a struct member that has type `void *` to `nullptr` this looks dubious (in C I would say that at the very least you should be using `uintptr_t` rather than `int64`).

gmarkall:
> This is a bit fragile since the default constructor logic and members could/will change in the future and we'd have to keep the Python logic in sync somehow. It would be better to use the actual constructor on the existing memory using placement new.

I sympathise with this argument. A motivation for doing it in the generated Python would be that the initialization of these members might be elided by Numba in other scenarios (e.g. if something writes to the members immediately following the allocation / initialization of the instance) but it's probably not an optimization worth worrying about at this point.

> > `udf_str.data = context.get_constant(types.int64, 0)`
> 
> My language-lawyering is always a bit weak, but if you're trying to initialise (from python) a struct member that has type `void *` to `nullptr` this looks dubious (in C I would say that at the very least you should be using `uintptr_t` rather than `int64`).

Whoops, yes, I should have used `types.uintp` there. WRT language-lawyering, I'm not sure what law there is that governs "compiling Python and CUDA C++ and linking them together" other than the law of "this is what works" :slightly_smiling_face:.

brandon-b-miller:
I'm torn here. On one hand, it might be easier to debug things if we go the route where we do the equivalent zeroing of the members in the lowering. On the other hand, placement new seems to be there for this kind of case specifically and would make shim functions a little less sensitive to mistakes in the lowering.

I maybe lean a little towards the latter? I am not a c++ expert, but as it is that our shim functions expect void pointers perhaps its best practice that the shim functions don't _require_ that a default constructed `udf_string` lives in the pointed-to memory (in contrast to if we had a real `udf_string*` as an arg...this _implies_ that what's being pointed to is valid right?)

gmarkall:
> I am not a c++ expert, but as it is that our shim functions expect void pointers perhaps its best practice that the shim functions don't _require_ that a default constructed `udf_string` lives in the pointed-to memory

That seems pretty logically sound so I'd also lean towards that.

brandon-b-miller:
Any reason I should keep the `zfill` anyways here?

gmarkall:
> Any reason I should keep the `zfill` anyways here?

I don't think so - I don't see that it will accomplish anything not already accomplished with the placement new.

wence-:
> I maybe lean a little towards the latter? I am not a c++ expert, but as it is that our shim functions expect void pointers perhaps its best practice that the shim functions don't _require_ that a default constructed `udf_string` lives in the pointed-to memory (in contrast to if we had a real `udf_string*` as an arg...this _implies_ that what's being pointed to is valid right?)

IIUC, you have _pre-compiled_ shim functions that are `extern "C"` with some defined ABI (which you control). You then link against them using code generated from numba and so need to advertise the calling convention to numba. There's actually remarkably little safety that the tooling can give you here I think, but I might be wrong.

But in any case, if your function advertises that it takes a `udf_string *x`, usual RAII would suggest that if you dereference `x` you get a correctly initialised object.

So if your shim functions just take a `void *` (which is a pointer to some externally managed heap that you promise is big enough but not initialised) and some information about how to initialise the `udf_string` then I too would do:

```c++
extern "C" void my_shim(arguments, void *data)
{
   udf_string str = new (data) ...;
   call_to_unshimmed_function(str, ...);
}
```

gmarkall (approved):

python/strings_udf/cpp/src/strings/udf/shim.cu
```diff
@@ -223,7 +223,7 @@ extern "C" __device__ int udf_string_from_string_view(int* nb_retbal,
                                                       void* udf_str)
 {
   auto str_view_ptr = reinterpret_cast<cudf::string_view const*>(str);
-  auto udf_str_ptr  = reinterpret_cast<udf_string*>(udf_str);
+  auto udf_str_ptr  = new (udf_str) udf_string;
   *udf_str_ptr      = udf_string(*str_view_ptr);
 
   return 0;
```
wence-, line 229: I think I don't understand what is going on here (and, _mutatis mutandis_, the rest of the patch).

`udf_string{string_view}` constructs a new `udf_string` from a `string_view` by calling `udf_string{sv.data(), sv.size_bytes()}` which sets up `m_bytes`, `m_capacity` with `size_bytes`, does `malloc` of the capacity and then a memcpy of the data into the `m_data` field.

So we did a malloc + memcpy to make the rvalue on line 227.

Separately we use placement new to build a pointer to some heap allocated space which default-initialises `udf_str_ptr` to be (say) zero-initialised.

Finally, the assignment `*udf_str_ptr = udf_string(*str_view_ptr)` calls `operator=` which does a move-construction (via `udf_string::assign` ?).

Why this level of indirection, why not:
```
auto udf_str_ptr = new (udf_str) udf_string;
(*udf_str_ptr)::assign(*reinterpret_cast<string_view *>(str));
```

What am I missing?

brandon-b-miller: That describes the process as I understand it as well, is your comment more about this being too verbose or is there something off about what I'm intending to happen here?

wence-: No sorry, I got confused by all the `operator=` overloading. I was worried that there was an extra copy happening, but I realise no.

python/strings_udf/strings_udf/lowering.py
```diff
@@ -199,7 +199,6 @@ def binary_func_impl(context, builder, sig, args):
                 udf_str_ptr = builder.alloca(
                     default_manager[udf_string].get_value_type()
                 )
-
```
wence-, line 202: Doesn't really need a python review for this change :)

wence- (approved):

brandon-b-miller:
@gpucibot merge

Issue #12081, opened 2022-11-06
[BUG] Building from source fails as deps are not fetched


ccoulombe:
First, :clap:  great work into making cudf and other RAPIDSAI libraries more easy to build from source!!! A lot of work has been put into it.
This is the continuity of #2770.

**Describe the bug**

Building v22.12 from source fails as some dependencies are missing or not included properly.

1. Linking fails from internal linker error.
```
/cvmfs/soft.computecanada.ca/gentoo/2020/usr/x86_64-pc-linux-gnu/binutils-bin/2.33.1/ld.gold: error: /home/coulombc/cudf/cpp/build/fatbin.ld: SECTIONS seen after other input files; try -T/--script
/cvmfs/soft.computecanada.ca/gentoo/2020/usr/x86_64-pc-linux-gnu/binutils-bin/2.33.1/ld.gold: internal error in write_sections, at /cvmfs/soft.computecanada.ca/gentoo/2020/var/tmp/portage/sys-devel/binutils-2.33.1-r1/work/binutils-2.33.1/gold/reloc.cc:790
```
2. `dlpack` is fetched and used for `libcudf` but fails for the python bindings.
```
/home/coulombc/cudf/python/cudf/_skbuild/linux-x86_64-3.10/cmake-build/cudf/_lib/interop.cxx:893:10: fatal error: dlpack/dlpack.h: No such file or directory
  893 | #include "dlpack/dlpack.h"
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
```

**Steps/Code to reproduce bug**
```
export CPATH=$CPATH:$RMM_ROOT/include
git clone --recursive https://github.com/rapidsai/cudf.git && cd cudf
PARALLEL_LEVEL=16 bash build.sh |& tee build.out
```

**Expected behavior**
Succeeds to build `cudf` from source, and not using conda!

**Environment overview (please complete the following information)**
 - Environment location: bare metal, centos
 - Method of cuDF install: source

**Solutions**
1. I can overcome the internal failure by building with `--cmake-args=\"-DCMAKE_SHARED_LINKER_FLAGS=\'-fuse-ld=bfd\'\"`
2. I can insall `dlpack` centrally and therefore it is found and included.
But since it is pulled and used when building `libcudf`, the `CMakeLists.txt` for the python binding of `cudf` should re-use it.

GregoryKimball:
[...]

PR #12139, opened 2022-11-14
Ensure dlpack include is provided to cudf interop lib

robertmaynard:
## Description
As brought up in #12081 it is possible to have python build failures due to no include paths to dlpack being provided. This fixes the issue by ensure that the DLPACK_INCLUDE_DIR is propagated down to the interop target.

We don't run into this issue with conda, since the dlpack headers are inside the conda include dir which is already being provided to the compiler.

## Checklist

### DIFF
--- a/cpp/cmake/thirdparty/get_dlpack.cmake
+++ b/cpp/cmake/thirdparty/get_dlpack.cmake
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2020-2021, NVIDIA CORPORATION.
+# Copyright (c) 2020-2022, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
--- a/cpp/cmake/thirdparty/get_dlpack.cmake
+++ b/cpp/cmake/thirdparty/get_dlpack.cmake
@@ -18,7 +18,9 @@ function(find_and_configure_dlpack VERSION)
   include(${rapids-cmake-dir}/find/generate_module.cmake)
   rapids_find_generate_module(DLPACK HEADER_NAMES dlpack.h)
 
+  include(${rapids-cmake-dir}/cpm/init.cmake)
   include(${rapids-cmake-dir}/cpm/find.cmake)
+  rapids_cpm_init()
   rapids_cpm_find(
     dlpack ${VERSION}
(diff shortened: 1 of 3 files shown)


vyasr (approved):
I'm fine with this solution in order to get it fixed for 22.12. In the long run, @robertmaynard how do you feel about C++ packages (libcudf, but also other RAPIDS C++ packages) making their public link interfaces more configurable via (possibly advanced) options that could be leveraged by our Python builds. It would be nice to avoid double-cloning/building something in cases where we know that the C++ build is being invoked from a Python build and we have complete control over the reuse. I'm also thinking about this from the perspective of libraries like cuml/cugraph using raft and how they reuse parts of what raft builds when placed in the public link interface.

robertmaynard:
> I'm fine with this solution in order to get it fixed for 22.12. In the long run, @robertmaynard how do you feel about C++ packages (libcudf, but also other RAPIDS C++ packages) making their public link interfaces more configurable via (possibly advanced) options that could be leveraged by our Python builds. It would be nice to avoid double-cloning/building something in cases where we know that the C++ build is being invoked from a Python build and we have complete control over the reuse. I'm also thinking about this from the perspective of libraries like cuml/cugraph using raft and how they reuse parts of what raft builds when placed in the public link interface.

Valid. I have captured this request for dlpack in a feature request ( https://github.com/rapidsai/cudf/issues/12175 ).
It would be good to see if other dependencies are re-cloned for python since that shouldn't be occuring with how we have setup libcudf.

vyasr:
I agree that this probably not much of an issue in libcudf right now, it's just top of mind because it's definitely an issue for raft/cuml/cugraph.

We need to include `rapids-find` before including `get_dlpack.cmake`. Would you prefer to do that in the CMakeLists.txt file or would you rather scope it more tightly and only include `find/generate_module.cmake` inside `get_dlpack.cmake`?

robertmaynard:
> I agree that this probably not much of an issue in libcudf right now, it's just top of mind because it's definitely an issue for raft/cuml/cugraph.
> 
> We need to include `rapids-find` before including `get_dlpack.cmake`. Would you prefer to do that in the CMakeLists.txt file or would you rather scope it more tightly and only include `find/generate_module.cmake` inside `get_dlpack.cmake`?

scope it tightly and only include `find/generate_module.cmake` inside `get_dlpack.cmake`

vyasr (approved):

vyasr:
@gpucibot merge

Issue #11636, opened 2022-08-31
[BUG] Binary ops fail unexpectedly for Decimal inputs


bdice:
**Describe the bug**
cuDF 22.08 raises an `UnboundLocalError` without much helpful context when attempting to compute certain operators like `__pow__` or `__floordiv__` on decimal columns.

**Steps/Code to reproduce bug**
```python
>>> from decimal import Decimal
>>> import cudf
>>> cudf.Series([Decimal("3.")]) ** 1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bdice/mambaforge/envs/rapids-22.08/lib/python3.9/site-packages/cudf/core/mixins/mixin_factory.py", line 11, in wrapper
    return method(self, *args1, *args2, **kwargs1, **kwargs2)
  File "/home/bdice/mambaforge/envs/rapids-22.08/lib/python3.9/site-packages/cudf/core/indexed_frame.py", line 2984, in _binaryop
    ColumnAccessor(type(self)._colwise_binop(operands, op)),
  File "/home/bdice/mambaforge/envs/rapids-22.08/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/home/bdice/mambaforge/envs/rapids-22.08/lib/python3.9/site-packages/cudf/core/frame.py", line 1707, in _colwise_binop
    else getattr(operator, fn)(left_column, right_column)
  File "/home/bdice/mambaforge/envs/rapids-22.08/lib/python3.9/site-packages/cudf/core/mixins/mixin_factory.py", line 11, in wrapper
    return method(self, *args1, *args2, **kwargs1, **kwargs2)
  File "/home/bdice/mambaforge/envs/rapids-22.08/lib/python3.9/site-packages/cudf/core/column/decimal.py", line 107, in _binaryop
    return result
UnboundLocalError: local variable 'result' referenced before assignment
```

**Expected behavior**
Either the operation will be computed successfully OR a helpful/informative error message is shown.

**Environment overview (please complete the following information)**
 - Environment location: Bare-metal
 - Method of cuDF install: conda

**Additional context**
Discovered as part of #10178.

PR #12142, opened 2022-11-15
[REVIEW] Fix decimal binary operations

galipremsagar:
## Description
Fixes: #11636 
This PR:

- [x] Fixes an `UnboundLocalError` error.
- [x] Fixes reflected binary operations and added tests for the same.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/column/decimal.py
+++ b/python/cudf/cudf/core/column/decimal.py
@@ -81,10 +81,8 @@ class DecimalBaseColumn(NumericalBaseColumn):
         # are computed outside of libcudf
         try:
             if op in {"__add__", "__sub__", "__mul__", "__div__"}:
-                output_type = _get_decimal_type(self.dtype, other.dtype, op)
-                result = libcudf.binaryop.binaryop(
-                    self, other, op, output_type
-                )
+                output_type = _get_decimal_type(lhs.dtype, rhs.dtype, op)
+                result = libcudf.binaryop.binaryop(lhs, rhs, op, output_type)
                 # TODO:  Why is this necessary? Why isn't the result's
                 # precision already set correctly based on output_type?
                 result.dtype.precision = output_type.precision
@@ -96,7 +94,12 @@ class DecimalBaseColumn(NumericalBaseColumn):
                 "__le__",
                 "__ge__",
             }:
-                result = libcudf.binaryop.binaryop(self, other, op, bool)
+                result = libcudf.binaryop.binaryop(lhs, rhs, op, bool)
+            else:
+                raise NotImplementedError(
+                    f"{op} not supported for types following dtypes:"
+                    f"{self.dtype}, {other.dtype}"
+                )
         except RuntimeError as e:
             if "Unsupported operator for these types" in str(e):
                 raise NotImplementedError(
--- a/python/cudf/cudf/core/column/decimal.py
+++ b/python/cudf/cudf/core/column/decimal.py
@@ -97,7 +97,7 @@ class DecimalBaseColumn(NumericalBaseColumn):
                 result = libcudf.binaryop.binaryop(lhs, rhs, op, bool)
             else:
                 raise NotImplementedError(
-                    f"{op} not supported for types following dtypes:"
+                    f"{op} not supported for the following dtypes: "
                     f"{self.dtype}, {other.dtype}"
                 )
         except RuntimeError as e:
--- a/python/cudf/cudf/core/column/decimal.py
+++ b/python/cudf/cudf/core/column/decimal.py
@@ -79,6 +79,10 @@ class DecimalBaseColumn(NumericalBaseColumn):
 
         # Binary Arithmetics between decimal columns. `Scale` and `precision`
         # are computed outside of libcudf
+        unsupported_msg = (
+            f"{op} not supported for the following dtypes: "
+            f"{self.dtype}, {other.dtype}"
+        )
         try:
             if op in {"__add__", "__sub__", "__mul__", "__div__"}:
                 output_type = _get_decimal_type(lhs.dtype, rhs.dtype, op)
@@ -96,15 +100,10 @@ class DecimalBaseColumn(NumericalBaseColumn):
             }:
                 result = libcudf.binaryop.binaryop(lhs, rhs, op, bool)
             else:
-                raise NotImplementedError(
-                    f"{op} not supported for the following dtypes: "
-                    f"{self.dtype}, {other.dtype}"
-                )
+                raise TypeError(unsupported_msg)
         except RuntimeError as e:
             if "Unsupported operator for these types" in str(e):
-                raise NotImplementedError(
-                    f"{op} not supported for types with different bit-widths"
-                ) from e
+                raise TypeError(unsupported_msg) from e
             raise
 
         return result
--- a/python/cudf/cudf/tests/test_binops.py
+++ b/python/cudf/cudf/tests/test_binops.py
@@ -2126,6 +2126,74 @@ def test_binops_decimal(args):
     "args",
     [
         (
+            "radd",
+            ["1.5", "2.0"],
+            cudf.Decimal64Dtype(scale=2, precision=3),
+            ["1.5", "2.0"],
+            cudf.Decimal64Dtype(scale=2, precision=3),
+            ["3.0", "4.0"],
+            cudf.Decimal64Dtype(scale=2, precision=4),
+        ),
+        (
+            "rsub",
+            ["100", "200"],
+            cudf.Decimal64Dtype(scale=-2, precision=10),
+            ["0.1", "0.2"],
+            cudf.Decimal64Dtype(scale=6, precision=10),
+            ["-99.9", "-199.8"],
+            cudf.Decimal128Dtype(scale=6, precision=19),
+        ),
+        (
+            "rmul",
+            ["1000", "2000"],
+            cudf.Decimal64Dtype(scale=-3, precision=4),
+            ["0.343", "0.500"],
+            cudf.Decimal64Dtype(scale=3, precision=3),
+            ["343.0", "1000.0"],
+            cudf.Decimal64Dtype(scale=0, precision=8),
+        ),
+        (
+            "rtruediv",
+            ["1.5", "0.5"],
+            cudf.Decimal64Dtype(scale=3, precision=6),
+            ["1.5", "2.0"],
+            cudf.Decimal64Dtype(scale=3, precision=6),
+            ["1.0", "4.0"],
+            cudf.Decimal64Dtype(scale=10, precision=16),
+        ),
+    ],
+)
+def test_binops_reflect_decimal(args):
+    op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype = args
+
+    a = utils._decimal_series(lhs, l_dtype)
+    b = utils._decimal_series(rhs, r_dtype)
+    expect = (
+        utils._decimal_series(expect, expect_dtype)
+        if isinstance(
+            expect_dtype,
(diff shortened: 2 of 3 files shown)


python/cudf/cudf/core/column/decimal.py
```diff
@@ -96,7 +94,12 @@ def _binaryop(self, other: ColumnBinaryOperand, op: str):
                 "__le__",
                 "__ge__",
             }:
-                result = libcudf.binaryop.binaryop(self, other, op, bool)
+                result = libcudf.binaryop.binaryop(lhs, rhs, op, bool)
+            else:
+                raise NotImplementedError(
+                    f"{op} not supported for types following dtypes:"
```
bdice, line 100: Needs a space. But what does “types following dtypes” mean? Can we just say “dtypes”?
```suggestion
                    f"{op} not supported for the following dtypes: "
```

python/cudf/cudf/tests/test_binops.py
```diff
@@ -2122,6 +2122,74 @@ def test_binops_decimal(args):
     utils.assert_eq(expect, got)
 
 
+@pytest.mark.parametrize(
+    "args",
```
bdice, line 2126:
```suggestion
    "op,lhs,l_dtype,rhs,r_dtype,expect,expect_dtype",
```

python/cudf/cudf/tests/test_binops.py
```diff
@@ -2122,6 +2122,74 @@ def test_binops_decimal(args):
...
+            ["1.5", "0.5"],
+            cudf.Decimal64Dtype(scale=3, precision=6),
+            ["1.5", "2.0"],
+            cudf.Decimal64Dtype(scale=3, precision=6),
+            ["1.0", "4.0"],
+            cudf.Decimal64Dtype(scale=10, precision=16),
+        ),
+    ],
+)
+def test_binops_reflect_decimal(args):
```
bdice, line 2166: List args explicitly as above.

python/cudf/cudf/tests/test_binops.py
```diff
@@ -2122,6 +2122,74 @@ def test_binops_decimal(args):
...
+    op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype = args
+
+    a = utils._decimal_series(lhs, l_dtype)
+    b = utils._decimal_series(rhs, r_dtype)
+    expect = (
+        utils._decimal_series(expect, expect_dtype)
+        if isinstance(
```
bdice, line 2173: I think there is a helper that checks for decimal dtypes.

python/cudf/cudf/core/column/decimal.py
galipremsagar: The intention was to show dtypes to the user rather than leaving then confused about which op isn't supported between which two types.

python/cudf/cudf/tests/test_binops.py
galipremsagar: Cleaned up this a bit, but didn't find any utility on those lines.

python/cudf/cudf/tests/test_binops.py
```diff
@@ -2122,6 +2122,68 @@ def test_binops_decimal(args):
...
+)
+def test_binops_reflect_decimal(
+    op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype
+):
+
+    a = utils._decimal_series(lhs, l_dtype)
+    b = utils._decimal_series(rhs, r_dtype)
+    expect = utils._decimal_series(expect, expect_dtype)
+
+    got = getattr(a, op)(b)
```
bdice, line 2174: This test is operating under a false premise, I think. Normal operator dispatch will never use the reverse operators here because it should default to the forward direction operator between two elements of the same type. How is this code path reachable under normal usage?

galipremsagar: We are not accessing the dunder methods directly, we are going through the API like: https://pandas.pydata.org/docs/reference/api/pandas.Series.rtruediv.html

python/cudf/cudf/tests/test_binops.py
```diff
@@ -2122,6 +2122,68 @@ def test_binops_decimal(args):
     utils.assert_eq(expect, got)
 
 
+@pytest.mark.parametrize(
+    "op,lhs,l_dtype,rhs,r_dtype,expect,expect_dtype",
+    [
+        (
+            "radd",
```
bdice, line 2129: Why are these `radd` and not `__radd__`?

python/cudf/cudf/core/column/decimal.py
```diff
@@ -96,7 +94,12 @@ def _binaryop(self, other: ColumnBinaryOperand, op: str):
                 "__le__",
                 "__ge__",
             }:
-                result = libcudf.binaryop.binaryop(self, other, op, bool)
+                result = libcudf.binaryop.binaryop(lhs, rhs, op, bool)
+            else:
+                raise NotImplementedError(
+                    f"{op} not supported for the following dtypes: "
```
bdice, line 100: This seems to compete with the logic on lines 103-107 which catch libcudf's `RuntimeError` and re-raise a different error about bit widths (too specific?). Shouldn't we be following a single failure path for all these cases, and simply passing through libcudf's `RuntimeError`? I don't like catching that and re-raising `NotImplementedError` here because I don't think we are consistent about our errors across all binary ops and dtypes. I'm content with passing through the `RuntimeError` but not happy with inconsistency if we aren't doing the same for unsupported operators with strings, datetimes, durations, etc.

bdice (changes requested):

python/cudf/cudf/tests/test_binops.py
bdice: TIL. I did not realize those methods existed. Apologies. We should be testing the dispatch as well, if possible.

galipremsagar: No worries, I'll add the dunder tests too.

python/cudf/cudf/core/column/decimal.py
galipremsagar: Discussed offline and decided to have one common error message here.

python/cudf/cudf/tests/test_binops.py
galipremsagar: Added the tests

python/cudf/cudf/tests/test_binops.py
```diff
@@ -2103,10 +2139,16 @@ def test_binops_with_NA_consistent(dtype, op):
...
     ],
 )
-def test_binops_decimal(args):
-    op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype = args
-
-    a = utils._decimal_series(lhs, l_dtype)
+def test_binops_decimal(op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype):
+
+    if l_dtype is None:
+        # cudf doesn't support binops with between
```
bdice, line 2145:
```suggestion
        # cudf doesn't support binops between
```

python/cudf/cudf/tests/test_binops.py
```diff
@@ -2103,10 +2139,16 @@ def test_binops_with_NA_consistent(dtype, op):
         ),
     ],
 )
-def test_binops_decimal(args):
-    op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype = args
-
-    a = utils._decimal_series(lhs, l_dtype)
+def test_binops_decimal(op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype):
+
+    if l_dtype is None:
```
bdice, line 2144: Can we change this to just interpret scalar values as `Scalar`?
```suggestion
    if isinstance(lhs, (int, float)):
```
Then define `l_dtype` as `cudf.Decimal64Dtype(scale=2, precision=6)` or similar for the scalar values, instead of using `None`.

Then the comment below is unnecessary and the result is `cudf.Scalar(2, cudf.Decimal64Dtype(scale=2, precision=6))` or similar.

bdice (changes requested):
A couple suggestions for simplification.

python/cudf/cudf/tests/test_binops.py
```diff
@@ -2103,10 +2139,16 @@ def test_binops_with_NA_consistent(dtype, op):
...
-
-    a = utils._decimal_series(lhs, l_dtype)
+def test_binops_decimal(op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype):
+
+    if isinstance(lhs, (int, float)):
+        # cudf doesn't support binops between
+        # arbitrary int/float and decimal column,
+        # hence need to construct a decimal scalar
+        # of same bit-width
+        a = cudf.Scalar(lhs, r_dtype)
```
galipremsagar, line 2149:
```suggestion
        a = cudf.Scalar(lhs, l_dtype)
```

python/cudf/cudf/tests/test_binops.py
```diff
@@ -1885,6 +1912,15 @@ def test_binops_with_NA_consistent(dtype, op):
             ["56.77", "1.79"],
             cudf.Decimal128Dtype(scale=13, precision=25),
         ),
+        (
+            operator.truediv,
+            20,
+            None,
```
galipremsagar, line 1918:
```suggestion
            cudf.Decimal128Dtype(scale=2, precision=6),
```

python/cudf/cudf/tests/test_binops.py
```diff
@@ -1858,6 +1876,15 @@ def test_binops_with_NA_consistent(dtype, op):
             ["343.0", "1000.0"],
             cudf.Decimal64Dtype(scale=0, precision=8),
         ),
+        (
+            operator.mul,
+            200,
+            None,
```
galipremsagar, line 1882:
```suggestion
            cudf.Decimal64Dtype(scale=3, precision=6),
```

python/cudf/cudf/tests/test_binops.py
```diff
@@ -1831,6 +1840,15 @@ def test_binops_with_NA_consistent(dtype, op):
             ["99.9", "199.8"],
             cudf.Decimal128Dtype(scale=6, precision=19),
         ),
+        (
+            operator.sub,
+            2,
+            None,
```
galipremsagar, line 1846:
```suggestion
            cudf.Decimal64Dtype(scale=3, precision=4),
```

python/cudf/cudf/tests/test_binops.py
```diff
@@ -1786,6 +1786,15 @@ def test_binops_with_NA_consistent(dtype, op):
             ["3.0", "4.0"],
             cudf.Decimal64Dtype(scale=2, precision=4),
         ),
+        (
+            operator.add,
+            2,
+            None,
```
galipremsagar, line 1792:
```suggestion
            cudf.Decimal64Dtype(scale=2, precision=3),
```

python/cudf/cudf/tests/test_binops.py
```diff
@@ -2103,10 +2139,16 @@ def test_binops_with_NA_consistent(dtype, op):
...
-    op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype = args
-
-    a = utils._decimal_series(lhs, l_dtype)
+def test_binops_decimal(op, lhs, l_dtype, rhs, r_dtype, expect, expect_dtype):
+
+    if isinstance(lhs, (int, float)):
+        # cudf doesn't support binops between
+        # arbitrary int/float and decimal column,
+        # hence need to construct a decimal scalar
+        # of same bit-width
```
galipremsagar, line 2148:
```suggestion
```

python/cudf/cudf/tests/test_binops.py
galipremsagar: Done 👍

bdice (approved):

galipremsagar:
@gpucibot merge

Issue #12149, opened 2022-11-15
[BUG] Pandas Rolling Var Differences Between 1.4.0 and 1.5.1


quasiben:
I'm seeing a small failure with cuDF when using Pandas 1.4.0 and it seems like it's because rolling variance changed between Pandas 1.4.0 and 1.5.1:

```python
In [2]: import pandas as pd

In [3]: pd.__version__
Out[3]: '1.4.0'

In [4]: s = pd.Series([1e8, 5, 5, 5])

In [5]: s.rolling(3).var()
Out[5]:
0             NaN
1             NaN
2    3.333333e+15
3   -5.000000e-01
dtype: float64

In [6]:
```

```python
In [2]: import pandas as pd

In [3]: pd.__version__
Out[3]: '1.5.1'

In [4]: s = pd.Series([1e8, 5, 5, 5])

In [5]: s.rolling(3).var()
Out[5]:
0             NaN
1             NaN
2    3.333333e+15
3    0.000000e+00
dtype: float64

In [6]:
```

cuDF meanwhile returns back the 1.5.1 result:

```python
In [7]: cudf.from_pandas(s).rolling(3).var()
Out[7]:
0            <NA>
1            <NA>
2    3.333333e+15
3             0.0
dtype: float64
```

@galipremsagar should we formally push users to Pandas 1.5.1 ?

galipremsagar:
We can `xfail` the test conditionally for 1.4.0

quasiben:
Good idea!  That's a lot better than pushing everyone to 1.5+

quasiben:
Upgrading env to 1.5.1

PR #12002, opened 2022-10-26
Add GitHub Actions Workflows

ajschmidt8:
## Description

This PR adds GitHub Actions workflows to `cudf`.

### Task list

Coverage required for this PR:
- [x] C++ tests
- [x] Python tests for cudf (first success: c3bdbce)
- [x] Python tests for custreamz, cudf_kafka
- [x] Java tests
- [x] Check dask configuration
- [x] `strings_udf`
- [x] Benchmarks
- [x] Notebooks
- [x] Codecov
- [x] Enable style checks script (`ci/check_style.sh`)
- [x] Consolidate [existing checks job](https://github.com/rapidsai/cudf/blob/c37d33c86f86bac13cb78b1e4643b05be81115df/.github/workflows/dependency-files.yml#L7-L12) into the `pr.yaml` workflow in this PR.

Future PRs:
- [ ] Build Metrics Report
- [ ] `libidentify_stream_usage`
- [ ] JUnit output annotations? https://github.com/marketplace/actions/junit-report-action
- [ ] Reduce exit code handling verbosity (use global variable with `trap` function)
- [ ] Setup `rapids-dependency-file-generator` pre-commit hook (also setup for others repos: `rmm`, `rapids-cmake`, `cuspatial`, etc.)
- [ ] Docs builds

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12150, opened 2022-11-15
fix/xfail two small tests for pandas 1.4.1

quasiben:
Fixes https://github.com/rapidsai/cudf/issues/12149

### DIFF
--- a/python/cudf/cudf/tests/test_rolling.py
+++ b/python/cudf/cudf/tests/test_rolling.py
@@ -8,12 +8,7 @@ import pandas as pd
 import pytest
 
 import cudf
-from cudf.core._compat import (
-    PANDAS_GE_110,
-    PANDAS_GE_130,
-    PANDAS_GE_150,
-    PANDAS_LT_140,
-)
+from cudf.core._compat import PANDAS_GE_110, PANDAS_GE_150, PANDAS_LT_140
 from cudf.testing._utils import _create_pandas_series, assert_eq
 from cudf.testing.dataset_generator import rand_dataframe
 
@@ -220,7 +215,7 @@ def test_rolling_var_std_large(agg, ddof, center, seed, window_size):
 
 
 @pytest.mark.xfail(
-    condition=not PANDAS_GE_130,
+    condition=not PANDAS_LT_140,
     reason="https://github.com/pandas-dev/pandas/issues/37051",
 )
 def test_rolling_var_uniform_window():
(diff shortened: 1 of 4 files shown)


python/cudf/cudf/tests/test_setitem.py
```diff
@@ -247,7 +247,7 @@ def test_categorical_setitem_invalid():
         # Following workaround is needed because:
         # https://github.com/pandas-dev/pandas/issues/46646
         with pytest.raises(
-            ValueError,
```
galipremsagar, line 250: Should this change exist? I see our code not changing in this PR.

python/cudf/cudf/tests/test_rolling.py
```diff
@@ -220,7 +215,7 @@ def test_rolling_var_std_large(agg, ddof, center, seed, window_size):
 
 
 @pytest.mark.xfail(
-    condition=not PANDAS_GE_130,
+    condition=not PANDAS_LT_140,
```
bdice, line 218: This is changing the version _and_ the direction of the bounds. Is this correct? Before: `not >=1.3.0` (so `< 1.3.0`). After: `not <1.4.0` (so `>=1.4.0`). Does this break compatibility with pandas 1.2, 1.3, ...?

More broadly, do we have a commitment to support all pandas versions 1.0+ in our tests? I think we should re-evaluate our pandas pinning strategy in that light and attempt to improve test coverage in that direction. It should be much more straightforward to test multiple dependency versions after the switch to GitHub Actions (#12002).

galipremsagar: It was actually fixed in 1.3.0 and broken in 1.4.0 until 1.5.0, so the condition has to be:
```
condition=not PANDAS_GE_130 or (PANDAS_GE_140 and not PANDAS_GE_150),
```

galipremsagar: @bdice is also right, It might be worth revisiting our back-word compatible pytest execution once github actions are enabled. But yeah not in scope of this PR.

quasiben: Hmm, given these statements and also finding more issue I think I should just move to 1.5.1 and close this PR.   What do you all think ?

galipremsagar: If you are in a dev environment, yes upgrading to 1.5.1 would be ideal. We can only guarantee 1.4 behavior for when our max support was 1.4 i.e., Jan 2022.

quasiben: Ok.  Closing the PR.  Sorry for the noise!

PR #11330, opened 2022-07-22
Fix performance issue and add a new code path to `cudf::detail::contains`

ttnghia:
The current implementation of `cudf::detail::contains` can process input with arbitrary nested types. However, it was reported to have severe performance issue when the input tables have many duplicate rows (https://github.com/rapidsai/cudf/issues/11299). In order to fix the issue, https://github.com/rapidsai/cudf/pull/11310 and https://github.com/rapidsai/cudf/pull/11325 was created. 

Unfortunately, https://github.com/rapidsai/cudf/pull/11310 is separating semi-anti-join from `cudf::detail::contains`, causing duplicate implementation. On the other hand, https://github.com/rapidsai/cudf/pull/11325 can address the issue https://github.com/rapidsai/cudf/issues/11299 but semi-anti-join using it still performs worse than the previous semi-anti-join implementation.

The changes in this PR include the following:
 * Fix the performance issue reported in https://github.com/rapidsai/cudf/issues/11299 for the current `cudf::detail::contains` implementation that support nested types.
 * Add a separate code path into `cudf::detail::contains` such that:
     * Input without having lists column (at any nested level) will be processed by the code path that is the same as the old implementation of semi-anti-join. This is to make sure the performance of semi-anti-join will remain the same as before.
     * Input with nested lists column, or NaNs compared as unequal, will be processed by another code path that supports nested types and different NaNs behavior. This will make sure support for nested types will not be dropped.

Closes https://github.com/rapidsai/cudf/issues/11299.

PR #10770, opened 2022-05-02
Update `groupby::hash` to use new row operators for keys

PointKernel:
Related to #8039 and #10181 

Contributes to #10186 

This PR updates `groupby::hash` to use new row operators. It gets rid of the current "flattened nested column" logic and allows `groupby::hash` to handle `LIST` and `STRUCT` keys. The work also involves small cleanups like getting rid of unnecessary template parameters and removing unused arguments.

It becomes a breaking PR since the updated `groupby::hash` will treat inner nulls as equal when top-level nulls are excluded 
 while the current behavior treats inner nulls as **unequal**.

PR #12159, opened 2022-11-16
Updating `stream_compaction/unique` to use new row comparators

divyegala:
## Checklist

As per conversation with @ttnghia, we decided the best way forward was to pre-sort the inputs to the `unique` benchmarks. However, the primitive types don't do that, but I can add the sorting as I have with list types

Edit: For some reason the pre-commit hook changed the copyright years of a lot of files untouched by this commit which ended up requesting many reviewers


divyegala:
run benchmarks

randerzander:
run benchmarks

ttnghia:
Your code failed style check.

randerzander:
run benchmarks

DaceT:
run benchmarks

ttnghia:
> run benchmarks

What is the purpose of that command? I never know that we have such github command.

DaceT:
> > run benchmarks
> 
> What is the purpose of that command? I never know that we have such github command.

__ This isn't a github command. However, we are attempting to test the dagster sensor work, which initiates a run (which includes building cudf and running the benchmarks) when the phrase `run benchmarks` is commented.

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

ttnghia:
We've experience performance regression in the past (https://github.com/rapidsai/cudf/pull/11330) and had to implement two separate code paths for the input with and w/o lists.

Thus, please provide a benchmark comparison for the changes before and after this PR. The benchmark input should include basic types and also complex nested lists.

divyegala:
Screen Shot 2022-11-17 at 6 16 48 PM
@ttnghia here's a screenshot of nvbench comparisons between the old row comparators and the new ones for primitive types

ttnghia:
Yeah the diffs in the margin of 5% is expected for primitive types. For nested types, it will be huge differences.

divyegala:
> Yeah the diffs in the margin of 5% is expected for primitive types. For nested types, it will be huge differences.

@ttnghia I don't think I understand, what is the comparison difference when it comes to nested types since we don't have a baseline for them?

ttnghia:
> what is the comparison difference when it comes to nested types since we don't have a baseline for them?

We don't have baseline for lists, as I said above. We can still compare structs.

divyegala:
@ttnghia any benchmark that you can recommend that constructs structs so I can use that as a reference?

cpp/benchmarks/stream_compaction/unique.cpp
```diff
@@ -77,3 +80,47 @@ NVBENCH_BENCH_TYPES(nvbench_unique, NVBENCH_TYPE_AXES(data_type, keep_option))
   .set_name("unique")
   .set_type_axes_names({"Type", "KeepOption"})
   .add_int64_axis("NumRows", {10'000, 100'000, 1'000'000, 10'000'000});
+
+template <typename Type, cudf::duplicate_keep_option Keep>
+void nvbench_unique_list(nvbench::state& state, nvbench::type_list<Type, nvbench::enum_type<Keep>>)
+{
+  if constexpr (Keep != cudf::duplicate_keep_option::KEEP_FIRST) {
+    state.skip("Skip unwanted benchmarks.");
```
PointKernel, line 88: Out of curiosity, why do we exclude `KEEP_FIRST` here?

divyegala: From what I understood from reading the implementation, `KEEP_ANY` and `KEEP_FIRST` do the same thing for `unique`

PointKernel: Ah, you are right.
https://github.com/rapidsai/cudf/blob/dec8bde1d5cc7462e52535a4e26f2c1be507a237/cpp/src/stream_compaction/unique.cu#L57-L58
Probably worth adding a comment explaining the reason of skipping `KEEP_FIRST`.

divyegala: I copied this from the primitive types benchmark, but we are skipping everything *but* `KEEP_FIRST`. I should probably modify both the benchmarks and skip `KEEP_ANY`

PointKernel: Works for me. Thanks!

ttnghia:
> @ttnghia any benchmark that you can recommend that constructs structs so I can use that as a reference?

You can have a look at this one: https://github.com/rapidsai/cudf/blob/762def46a77ea7500502d4bc12d71de686d7c456/cpp/benchmarks/groupby/group_struct_values.cpp#L30.

For producing duplicate, use the approach that I mentioned:
 * Generate random template table using the reference code above (the template table has size `M`)
 * Generate a gather map with size `N` (`N` can be `k` times `M`, such as `N = 10M`).
 * Gather the template table using that gather map. Doing so you will have input with approximately `k` times duplicates.

PointKernel (approved):
LGTM

IMO We don't need to investigate the performance impact on structs between experimental operators and existing ones in this PR since similar work has been performed already (e.g. https://github.com/rapidsai/cudf/pull/10770#issuecomment-1135336359).

cpp/benchmarks/stream_compaction/unique.cpp
```diff
@@ -62,9 +66,12 @@ void nvbench_unique(nvbench::state& state, nvbench::type_list<Type, nvbench::enu
   auto input_column = source_column->view();
   auto input_table  = cudf::table_view({input_column, input_column, input_column, input_column});
 
+  auto const sort_order = cudf::sorted_order(input_table);
+  auto const sort_table = cudf::gather(input_table, *sort_order);
+
   state.set_cuda_stream(nvbench::make_cuda_stream_view(cudf::get_default_stream().value()));
   state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
-    auto result = cudf::unique(input_table, {0}, Keep, cudf::null_equality::EQUAL);
+    auto result = cudf::unique(*sort_table, {0}, Keep, cudf::null_equality::EQUAL);
```
ttnghia, line 74: I think we can do better by benchmarking both with and w/o sorting input. This is is what I was expecting, and we can achieve that very easily.
```
auto const run_bench = [&] (auto const input){
  state.set_cuda_stream();
  state.exec(...);
};
auto const sorting = static_cast<bool>(state.get_int64("Sort"));
if(sorting) {
   // do the benchmark with sorted table
   run_bench(sort_table);
} else {
   // do the benchmark with the random generated input table
   run_bench(rnd_table);
}
```

cpp/benchmarks/stream_compaction/unique.cpp
```diff
@@ -77,3 +84,48 @@ NVBENCH_BENCH_TYPES(nvbench_unique, NVBENCH_TYPE_AXES(data_type, keep_option))
...
+    // number of unique rows. The max list size is 4 and the number of unique values in the
+    // list's child is 5. So the number of unique rows in the list = 1 + 5 + 5^2 + 5^3 + 5^4 = 781
+    // We want this column to also have 781 unique values.
+    builder.distribution(dtype, distribution_id::UNIFORM, 0, 781);
+  }
+
+  auto const table = create_random_table(
+    {dtype}, table_size_bytes{static_cast<size_t>(size)}, data_profile{builder}, 0);
+  auto const sort_order = cudf::sorted_order(*table);
+  auto const sort_table = cudf::gather(*table, *sort_order);
```
ttnghia, line 118: Similar to above.

cpp/benchmarks/stream_compaction/unique.cpp
```diff
@@ -62,9 +66,12 @@ void nvbench_unique(nvbench::state& state, nvbench::type_list<Type, nvbench::enu
   auto input_column = source_column->view();
   auto input_table  = cudf::table_view({input_column, input_column, input_column, input_column});
 
+  auto const sort_order = cudf::sorted_order(input_table);
```
ttnghia, line 69: This comment is not for this line but for the line 61 above. The data gen profile should based on the input size. For example, the line 
```
distribution_id::UNIFORM, 0, 100
```
should be replaced by 
```
distribution_id::UNIFORM, 0, num_rows / 100
```

By doing so, we will have more realistic data with approx. constant ratio of duplicates. Otherwise we will have very skewed data. For example, imagine that we have 100 millions rows with values in `[0, 100)`, the duplicate factor is huge and this benchmark may give very strange output.

ttnghia (approved):

PR #12160, opened 2022-11-16
Fix data corruption when reading ORC files with empty stripes

### LINKED ISSUE #12155: [BUG] CUDF crashes trying to  read a valid ORC file with a single null entry in it
**Describe the bug**
The file that causes this error is attached below

Try to read this particular file using CUDF.  It will throw an exception like

```
cuDF failure at: .../cpp/src/io/orc/reader_impl.cu:1081: Expected streams data within stripe
```

I am not an expert in ORC, but I tried to debug this a little bit at the point in which the error is being thrown we are trying to read two STRING columns (I assume the key and value for the map).

**Expected behavior**
I should be able to read the file and get the null out, like I can in Spark.

```
scala> spark.read.orc("503819761.orc").show()
+---------+
|query_map|
+---------+
|     null|
+---------+

```
or pandas

```
>>> pd.read_orc("503819761.orc")
  query_map
0      None
```

### LINKED ISSUE #12156: [BUG] CUDF crashes when trying to read a valid ORC file with 100 null entries in it.
**Describe the bug**
This is related to https://github.com/rapidsai/cudf/issues/12155.  I was trying to produce a synthetic file with the same error just so I could share it without getting permission.  Well I was able to make a crash happen, but it was not the same crash. I am not sure if this is a bug that will show up as a part of #12155 or not, but I wanted to document it.  Just so you know I wrote this file using CUDF.

Read the orc file inside of [other_error.zip](https://github.com/rapidsai/cudf/files/10016760/other_error.zip) using CUDF it will fail with an error something like 

```
cuDF failure at: cudf/cpp/src/io/orc/reader_impl.cu:333: No decompressible data found
```
Please ignore the line number I added in some debug print statements and so the lines are off a bit.

**Expected behavior**
It is read correctly like with Spark.
```
scala> spark.read.orc("part-00000-f7e595c6-ab11-46d0-9619-1f71cc41deac-c000.snappy.orc").show()
+----+
|   m|
+----+
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
|null|
+----+
only showing top 20 rows

```

or pandas

```
>>> pd.read_orc("part-00000-f7e595c6-ab11-46d0-9619-1f71cc41deac-c000.snappy.orc")
       m
0   None
1   None
2   None
3   None
4   None
..   ...
95  None
96  None
97  None
98  None
99  None

[100 rows x 1 columns]
```

vuule:
## Description
Closes #12155 
Closes #12156

Two fixes:

- Fixes the reader logic that used to mark entire (nesting) level of columns as having no data when one or more stripes have no data.
- Removes the assert that failed when a non-struct column has no data in a stripe. There are several corner cases where this is valid input, some of which are not cheap to check.

## Checklist


revans2:
This fixes the one file I uploaded, but it does not work when there are other stripes in the file that do have data. I'll see if I can come up with a synthetic file that shows this same issue.

revans2:
[combined_error.zip](https://github.com/rapidsai/cudf/files/10023395/combined_error.zip) holds an ORC file that has more rows in it and still is showing the problem that we had, and fits more with what our customer is actually seeing.

harrism (approved):

galipremsagar (approved):

PointKernel (approved):

revans2 (approved):

jolorunyomi (approved):

PR #12164, opened 2022-11-16
Don't rely on GNU find in headers_test.sh

wence-:
## Description

`-printf` is a GNU find extension, so `headers_test.sh` fails on systems where binutils is a BSD toolchain.

To get around this, use sed to obtain the effect of `-printf`.

## Checklist

### DIFF
--- a/ci/checks/headers_test.sh
+++ b/ci/checks/headers_test.sh
@@ -10,7 +10,7 @@ DIRNAMES="cudf cudf_test"
 
 # existence tests for lib${LIBNAME}
 for DIRNAME in ${DIRNAMES[@]}; do
-    HEADERS=`cd cpp && find include/${DIRNAME}/ -type f \( -iname "*.h" -o  -iname "*.hpp" \) -printf "        - test -f \\\$PREFIX/%p\n" | sort`
+    HEADERS=`cd cpp && find include/${DIRNAME} -type f \( -iname "*.h" -o  -iname "*.hpp" \) -print | sed 's|^|        - test -f $PREFIX/|' | sort`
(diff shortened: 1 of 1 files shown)


bdice (approved):

ci/checks/headers_test.sh
```diff
@@ -10,7 +10,7 @@ DIRNAMES="cudf cudf_test"
 
 # existence tests for lib${LIBNAME}
 for DIRNAME in ${DIRNAMES[@]}; do
-    HEADERS=`cd cpp && find include/${DIRNAME}/ -type f \( -iname "*.h" -o  -iname "*.hpp" \) -printf "        - test -f \\\$PREFIX/%p\n" | sort`
+    HEADERS=`cd cpp && find include/${DIRNAME} -type f \( -iname "*.h" -o  -iname "*.hpp" \) -print | sed 's|^|        - test -f $PREFIX/|' | sort`
```
vyasr, line 13: Should the $PREFIX still be escaped? Unless I'm misunderstanding, the old code would output `\$PREFIX` whereas the new code just prints `$PREFIX`. The meta.yaml files don't contain the backslash though so I suspect that I just don't know the escaping syntaxes here all too well.

wence-: single quotes as I use here inhibit shell expansion, which is what is wanted, AFAICT.

vyasr: Ah I didn't see that you had changed the type of quotes. That makes sense then.

vyasr (approved):

vyasr:
@gpucibot merge

raydouglass (approved):

PR #12181, opened 2022-11-17
Support regex EOL where the string ends with a new-line character

### LINKED ISSUE #11979: [FEA] Make line terminator sequences in regular expression using `$` a configurable option when `MULTILINE` flag is enabled
In regular expression multiline mode, currently the `$` matches at the position right before a newline character `\n` (a line terminator) in cuDF. In Python, this behavior makes sense and is consistent with the Python implementation. However, Apache Spark uses the JDK (Java) implementation, and line terminator sequences are a bit more complex. The JDK regular expression library utilizes either of newline (`\n`), carriage return (`\r`), carriage return followed by newline (`\r\n`), and 3 other Unicode newline variants as a line terminator.

It would be useful if we could configure the concept of line terminator sequences in cuDF. Ideally, this could be an optional parameter that would support a simple array of strings for line terminator sequences. Another alternative could be another flag for line terminators or another type of MULTILINE flag

Currently, [spark-rapids](https://github.com/NVIDIA/spark-rapids) handles `$` by doing a heavy translation from a JDK regular expression to another regular expression supported by cuDF that handles the multiple possible line terminator sequences that the JDK uses. It cannot use cuDF MULTILINE mode because only the newline is handled there.  With this translation, we are limited to only using the `$` in simple scenarios at the end of the regular expression, we cannot use them in choice `|` right now among other constructions because of the complexity.

davidwendt:
## Description

Support regex EOL where the string ends with a new-line character.
This matches the behavior for the EOL anchor `$` described here: https://www.regular-expressions.info/refanchors.html
Additional gtests are included.
The doxygen for cudf regex support is also updated.

Close #11979

## Checklist


bdice (approved):

vyasr (approved):

davidwendt:
@gpucibot merge

PR #12130, opened 2022-11-11
Reduce page fragment size in Parquet writer

vuule:
## Description
Parquet writer creates pages from fragments of fixed number of rows. With deep nesting or long strings this way of creating pages can lead to pages that are much larger than the target size.
This PR alleviates the issue by reducing the number of rows in each fragment.

Performance impact:
5% throughput increase on average in the writer benchmark. The benefit is very uneven, with numeric types having almost 10% throughput loss and list type getting up to 110% faster.
15% throughput increase on average in the reader benchmark, with ~10% throughput loss with numeric types and up to 220% gain with lists.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #12182, opened 2022-11-17
Fix page size calculation in Parquet writer

etseidl:
## Description
When calculating page boundaries, the current Parquet writer does not take into account storage needed per page for repetition and definition level data.  As a consequence pages may sometimes exceed the specified limit, which in turn impacts the ability to compress these pages with codecs that have a maximum buffer size.  This PR fixes the page size calculation to take repetition and definition levels into account.  

~~This also incorporates the fragment size reduction from 5000 to 1000 that was suggested in #12130~~

## Checklist


GPUtester:
Can one of the admins verify this patch?

Admins can comment `ok to test` to allow this one PR to run or `add to allowlist` to allow all future PRs from the same author to run.

etseidl:
I ran the NDS conversion from csv to parquet/zstd at scale=100.  The total output size was 27G, consistent with the output before this change.  Individual files sizes are:
```
12K     income_band
12K     reason
12K     ship_mode
16K     warehouse
20K     call_center
20K     web_site
28K     household_demographics
36K     web_page
40K     store
44K     promotion
484K    time_dim
788K    catalog_page
956K    date_dim
5.0M    customer_demographics
15M     customer_address
26M     item
79M     customer
426M    web_returns
879M    catalog_returns
1.3G    store_returns
1.6G    inventory
3.8G    web_sales
7.8G    catalog_sales
11G     store_sales
```
compared to the following provided by @jbrennan333 
```
16K	./gpu_parquet_sf100_zstd/income_band
16K	./gpu_parquet_sf100_zstd/reason
16K	./gpu_parquet_sf100_zstd/ship_mode
20K	./gpu_parquet_sf100_zstd/warehouse
24K	./gpu_parquet_sf100_zstd/call_center
24K	./gpu_parquet_sf100_zstd/web_site
32K	./gpu_parquet_sf100_zstd/household_demographics
40K	./gpu_parquet_sf100_zstd/web_page
44K	./gpu_parquet_sf100_zstd/store
48K	./gpu_parquet_sf100_zstd/promotion
488K	./gpu_parquet_sf100_zstd/time_dim
796K	./gpu_parquet_sf100_zstd/catalog_page
960K	./gpu_parquet_sf100_zstd/date_dim
5.0M	./gpu_parquet_sf100_zstd/customer_demographics
16M	./gpu_parquet_sf100_zstd/customer_address
27M	./gpu_parquet_sf100_zstd/item
82M	./gpu_parquet_sf100_zstd/customer
435M	./gpu_parquet_sf100_zstd/web_returns
887M	./gpu_parquet_sf100_zstd/catalog_returns
1.3G	./gpu_parquet_sf100_zstd/store_returns
1.6G	./gpu_parquet_sf100_zstd/inventory
3.8G	./gpu_parquet_sf100_zstd/web_sales
7.8G	./gpu_parquet_sf100_zstd/catalog_sales
11G	./gpu_parquet_sf100_zstd/store_sales
```

etseidl:
@vuule the benchmarks really are a mixed bag. I'll have to test with 5000 again, but maybe we should hold off on the fragment size change in favor of trying the per-column approach.

vuule:
This is my summary of results with just the fragment size change:
> Performance impact:
5% throughput increase on average in the writer benchmark. The benefit is very uneven, with numeric types having almost 10% throughput loss and list type getting up to 110% faster.
15% throughput increase on average in the reader benchmark, with ~10% throughput loss with numeric types and up to 220% gain with lists.

Are you seeing something similar? This is without ZSTD (de)compression, so maybe you're seeing a different picture.

etseidl:
My number are just a bit more extreme.  I'm seeing more like a 13-15% slowdown on integral/time types and as much as 50% increase for list.  This is on my workstation w/ an A6000, so I should probably try repeating on the A100s at work.  That and I should try benchmarking with some different datasets.  (Also need to dump to CSV so I can get harder performance numbers).

vuule:
As you suggested earlier, let's take the L on the fragment size and just fix the page size for now :)

etseidl:
@vuule the fix is mostly ready for review then...do you think there are any unit tests to add?  I can't think of anything beyond write a file then read all the page headers to confirm that all uncompressed sizes are below the threshold.  I don't know if there's any value there.

cpp/src/io/parquet/page_enc.cu
```diff
@@ -359,6 +359,11 @@ __global__ void __launch_bounds__(128)
       // override this_max_page_size if the requested size is smaller
       this_max_page_size = min(this_max_page_size, max_page_size_bytes);
 
+      // subtract size of rep and def level vectors
+      auto num_vals = values_in_page + frag_g.num_values;
+      this_max_page_size -= max_RLE_page_size(col_g.num_def_level_bits(), num_vals) +
+                            max_RLE_page_size(col_g.num_rep_level_bits(), num_vals);
+
```
ttnghia, line 366: Please make sure to merge upstream with the latest changes in this file.

cpp/src/io/parquet/page_enc.cu
```diff
@@ -359,6 +359,11 @@ __global__ void __launch_bounds__(128)
       // override this_max_page_size if the requested size is smaller
       this_max_page_size = min(this_max_page_size, max_page_size_bytes);
 
+      // subtract size of rep and def level vectors
+      auto num_vals = values_in_page + frag_g.num_values;
+      this_max_page_size -= max_RLE_page_size(col_g.num_def_level_bits(), num_vals) +
```
vuule, line 364: Maybe we should be using `long` for `this_max_page_size` so it doesn't underflow in some odd case.

etseidl: Ok, I'll take a look at that.

etseidl: @vuule so use signed rather than `size_t` so it will underflow rather than wrap around?  Underflow would just guarantee the test immediately following would evaluate to `true`.

vuule: But it should evaluate to `true`, right? If `this_max_page_size` somehow becomes negative, it means that we can't fit the def and rep levels in the target page size, even without the data.
With `size_t`, `this_max_page_size` will be a huge positive number instead and we'll keep building the page until other conditions are met (e.g. `num_rows` limit).

etseidl: Yes, I wasn't clear.  Underflow will evaluate true, which is what we want.  Overflow will result in an even larger page, which we don't want.  I think we're in agreement 😄

vuule (approved):

cpp/src/io/parquet/page_enc.cu
```diff
@@ -352,13 +352,19 @@ __global__ void __launch_bounds__(128)
           ? frag_g.num_leaf_values * util::div_rounding_up_unsafe(ck_g.dict_rle_bits, 8)
           : frag_g.fragment_data_size;
       // TODO (dm): this convoluted logic to limit page size needs refactoring
-      size_t this_max_page_size = (values_in_page * 2 >= ck_g.num_values)   ? 256 * 1024
-                                  : (values_in_page * 3 >= ck_g.num_values) ? 384 * 1024
-                                                                            : 512 * 1024;
+      long this_max_page_size = (values_in_page * 2 >= ck_g.num_values)   ? 256 * 1024
```
ttnghia, line 355: You may also need to modify the line 360 below to avoid comparing between `size_t` and `long`:
```suggestion
if(this_max_page_size > static_cast<long>(max_page_size_bytes)) {
  this_max_page_size = max_page_size_bytes;
}
```

ttnghia: The should also be similar issue with comparing `this_max_page_size` with other variables below but I can't figure out an efficient way to fix all.

vuule: Huh. Doesn't compiler complain about signed-unsigned comparison?

ttnghia: I have no idea. Sometimes it is very aggressive. Some other times it just silently ignored.

etseidl: tbh I expected the compiler to yell at me about this...didn't realize `min()` was overloaded so much (unlike `std::min`).  Fixed.

ttnghia: Yeah I've been working with cuda code/`nvcc` long enough time and no longer trust in its ability to yell at me for non-critical warnings.

ttnghia: I believe that the remaining usage of `this_max_page_size` is at line 369. Can you please check for signed/unsigned comparison there (and fix if necessary)?

etseidl: I think the comparison at line 369 is ok since it's checking whether the sum of two 32-bit unsigneds is greater than a 64-bit signed.  

Ofc that raises the question of whether using 32 bits for page size is sufficient.  But since pages really should be in the kilobytes, it shouldn't be an issue.  Maybe we should just change the `max_page_size_bytes` parameter to be a `size_type` instead?  But that's getting a little out-of-scope maybe.

vuule: `size_type` is meant to be used for the number of rows (I'm sure there are many places where it's misused). I prefer to use 64-bit types for byte counting, as much as I agree that page size should never be over 2GB :D
Agreed that this would be out-of-scope for this, very focused, PR.

etseidl: I take that back...if the addition overflows, then the test at 369 might never evaluate true 😮 

Maybe there should be a follow-on PR to clean this up.

ttnghia (approved):

vuule:
@gpucibot merge

PR #12190, opened 2022-11-17
Workaround thrust-copy-if limit in json get_tree_representation

davidwendt:
## Description
Workaround in json's get_tree_representation due to limitation in `thrust::copy_if` which fails if the input-iterator spans more than int-max.

Found existing thrust issue: https://github.com/NVIDIA/thrust/issues/1302
This calls the thrust::copy_if in chunks if the iterator can span greater than int-max.

Found while working on https://github.com/rapidsai/cudf/pull/12079

## Checklist


elstehle (approved):
Thanks for the workaround. This looks good. There should be effectively "zero" overhead in cases when we're running for less than `INT_MAX` items. 

nit: would we want to put this into some common ground for other places to use, where we could run into scenarios that exceed `INT_MAX` problem sizes (e.g., the wordpiece tokenizer)?

davidwendt:
> nit: would we want to put this into some common ground for other places to use, where we could run into scenarios that exceed `INT_MAX` problem sizes (e.g., the wordpiece tokenizer)?

I was considering that. I think the other 2 places this is occurs will be fixed differently in the future (not require the copy-if). 
The only other one I found was this one here (which also used the sentinel overload) as so I thought it best to keep it isolated until we found another case. I could easily be persuaded to move this to a detail header.

cpp/src/io/json/json_tree.cu
```diff
@@ -58,6 +58,50 @@
...
 namespace cudf::io::json {
 namespace detail {
+namespace {
+
+/**
+ * @brief Utility for calling thrust::copy_if
+ *
+ * Workaround for thrust::copy_if bug (https://github.com/NVIDIA/thrust/issues/1302)
+ * where it cannot iterate over int-max values `distance(first,last) > int-max`
+ * This calls thrust::copy_if in 2B chunks instead.
```
hyperbolic2346, line 68: 2G

hyperbolic2346 (approved):
Small not on comment typo, but otherwise looks good.

cpp/src/io/json/json_tree.cu
```diff
@@ -58,6 +58,50 @@
...
+}
+
+template <typename InputIterator, typename OutputIterator, typename Predicate>
+OutputIterator thrust_copy_if(rmm::exec_policy policy,
+                              InputIterator first,
+                              InputIterator last,
+                              OutputIterator result,
+                              Predicate pred)
+{
+  return thrust_copy_if(policy, first, last, first, result, pred);
```
ttnghia, line 102: Wow, I assume that you're going to extract this out into a common utility file and call it across many places.

davidwendt: https://github.com/rapidsai/cudf/pull/12190#issuecomment-1332799678

davidwendt:
@gpucibot merge

Issue #11993, opened 2022-10-25
[FEA] CUDF cannot read binary encoded decimal values in parquet


revans2:
**Describe the bug**
The parquet spec says that you should be able to encode any decimal type as a binary (like a string). When I try to do this CUDF throws an exception saying it is not supported.

[tmp.zip](https://github.com/rapidsai/cudf/files/9862988/tmp.zip)

```
cudf/cpp/src/io/parquet/reader_impl_helpers.cu:124: Invalid representation of decimal type
```

**Steps/Code to reproduce bug**
The attached zip file above has two parquet files in it. One has a single column and the other has 3 columns.

**Expected behavior**
Ideally we would like to have CUDF parse the values and return a DECIMAL column that is sized based off of the size of the precision.  Precision <= 9 is DECIMAL32
Precision <= 18 && > 9 is DECIMAL64
Precision <= 38 && > 18 is DECIMAL128
Precision > 38 is an error

Also when parsing the data it is possible to have binary data that is larger than the given type will hold. To be compatible with Spark ideally if the higher order bytes are zero, then they are ignored and the data is returned. If the resulting data is too large to fit then that should also throw an exception. But this should really be rare.

revans2:
If you need another file https://github.com/daveoshinsky/drill/blob/d2372a892216269a71425175c2b2ad9b02941261/exec/java-exec/src/test/resources/parquet/varlenDecimal.parquet is a file created for the same kind of issue on a different project.

If you need me to include the data in these files I am happy to do it.

GregoryKimball:
Thanks @revans2 for raising this issue. It looks like we added the `parquet::FIXED_LEN_BYTE_ARRAY` physical type for decimal during the `DECIMAL128` push.

For context, here is the description of [binary physical types for decimal](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal) in the parquet spec. Where can we find out more about how this representation works? How many bits go to precision? 

I suppose we ought to change the [error](https://github.com/rapidsai/cudf/blob/b7d011528af766294e84e4dafd483c650fb5901b/cpp/src/io/parquet/reader_impl.cu#L145) from `"Invalid representation of decimal type"` to `Not implemented`...

PR #12205, opened 2022-11-18
Add support to read binary encoded decimals in parquet

PointKernel:
Issue #11993 

## Checklist

### DIFF
--- a/cpp/src/io/parquet/page_data.cu
+++ b/cpp/src/io/parquet/page_data.cu
@@ -760,6 +760,24 @@ inline __device__ void gpuOutputInt64Timestamp(volatile page_state_s* s, int src
 }
 
 /**
+ * @brief Output a byte array as int.
+ *
+ * @param[in] ptr Pointer to the byte array
+ * @param[in] len Byte array length
+ * @param[out] dst Pointer to row output data
+ */
+template <typename T>
+__device__ void gpuOutputByteArrayAsInt(char const* ptr, int32_t len, T* dst)
+{
+  T unscaled = 0;
+  for (auto i = 0; i < len; i++) {
+    uint32_t v = ptr[i];
+    unscaled   = (unscaled << 8) | v;
+  }
+  *dst = unscaled;
+}
+
+/**
  * @brief Output a fixed-length byte array as int.
  *
  * @param[in,out] s Page state input/output
@@ -1831,6 +1849,12 @@ __global__ void __launch_bounds__(block_size) gpuDecodePageData(
           s->page.nesting[leaf_level_index].data_out + static_cast<size_t>(dst_pos) * dtype_len;
         if (dtype == BYTE_ARRAY) {
           gpuOutputString(s, val_src_pos, dst);
+          if (s->col.converted_type == DECIMAL) {
+            auto* char_data = static_cast<string_index_pair*>(dst);
+            auto* ptr       = char_data->first;
+            auto const len  = char_data->second;
+            gpuOutputByteArrayAsInt(ptr, len, static_cast<__int128_t*>(dst));
+          }
         } else if (dtype == BOOLEAN) {
           gpuOutputBoolean(s, val_src_pos, static_cast<uint8_t*>(dst));
         } else if (s->col.converted_type == DECIMAL) {
--- a/cpp/src/io/parquet/parquet_gpu.hpp
+++ b/cpp/src/io/parquet/parquet_gpu.hpp
@@ -184,7 +184,7 @@ struct ColumnChunkDesc {
                            int8_t codec_,
                            int8_t converted_type_,
                            LogicalType logical_type_,
-                           int8_t decimal_scale_,
+                           int8_t decimal_precision_,
                            int32_t ts_clock_rate_,
                            int32_t src_col_index_,
                            int32_t src_col_schema_)
@@ -207,7 +207,7 @@ struct ColumnChunkDesc {
       codec(codec_),
       converted_type(converted_type_),
       logical_type(logical_type_),
-      decimal_scale(decimal_scale_),
+      decimal_precision(decimal_precision_),
       ts_clock_rate(ts_clock_rate_),
       src_col_index(src_col_index_),
       src_col_schema(src_col_schema_)
@@ -236,7 +236,7 @@ struct ColumnChunkDesc {
   int8_t codec;                               // compressed codec enum
   int8_t converted_type;                      // converted type enum
   LogicalType logical_type;                   // logical type
-  int8_t decimal_scale;                       // decimal scale pow(10, -decimal_scale)
+  int8_t decimal_precision;                   // decimal precision
   int32_t ts_clock_rate;  // output timestamp clock frequency (0=default, 1000=ms, 1000000000=ns)
 
   int32_t src_col_index;   // my input column index
--- a/cpp/src/io/parquet/parquet_gpu.hpp
+++ b/cpp/src/io/parquet/parquet_gpu.hpp
@@ -184,7 +184,6 @@ struct ColumnChunkDesc {
                            int8_t codec_,
                            int8_t converted_type_,
                            LogicalType logical_type_,
-                           int8_t decimal_precision_,
                            int32_t ts_clock_rate_,
                            int32_t src_col_index_,
                            int32_t src_col_schema_)
@@ -207,7 +206,6 @@ struct ColumnChunkDesc {
       codec(codec_),
       converted_type(converted_type_),
       logical_type(logical_type_),
-      decimal_precision(decimal_precision_),
       ts_clock_rate(ts_clock_rate_),
       src_col_index(src_col_index_),
       src_col_schema(src_col_schema_)
@@ -236,7 +234,6 @@ struct ColumnChunkDesc {
   int8_t codec;                               // compressed codec enum
   int8_t converted_type;                      // converted type enum
   LogicalType logical_type;                   // logical type
-  int8_t decimal_precision;                   // decimal precision
   int32_t ts_clock_rate;  // output timestamp clock frequency (0=default, 1000=ms, 1000000000=ns)
 
   int32_t src_col_index;   // my input column index
--- a/cpp/src/io/parquet/reader_impl_helpers.cpp
+++ b/cpp/src/io/parquet/reader_impl_helpers.cpp
@@ -77,15 +77,11 @@ type_id to_type_id(SchemaElement const& schema,
   parquet::Type const physical            = schema.type;
   parquet::LogicalType const logical_type = schema.logical_type;
   parquet::ConvertedType converted_type   = schema.converted_type;
-  int32_t decimal_precision               = schema.decimal_precision;
 
   // Logical type used for actual data interpretation; the legacy converted type
   // is superceded by 'logical' type whenever available.
   auto const inferred_converted_type = logical_type_to_converted_type(logical_type);
   if (inferred_converted_type != parquet::UNKNOWN) converted_type = inferred_converted_type;
-  if (inferred_converted_type == parquet::DECIMAL) {
-    decimal_precision = schema.logical_type.DECIMAL.precision;
-  }
 
   switch (converted_type) {
     case parquet::UINT_8: return type_id::UINT8;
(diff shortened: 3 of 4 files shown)


cpp/src/io/parquet/page_data.cu
```diff
@@ -759,6 +759,24 @@ inline __device__ void gpuOutputInt64Timestamp(volatile page_state_s* s, int src
...
+__device__ void gpuOutputByteArrayAsInt(char const* ptr, int32_t len, T* dst)
+{
+  T unscaled = 0;
+  for (auto i = 0; i < len; i++) {
+    uint8_t v = ptr[i];
+    unscaled  = (unscaled << 8) | v;
+  }
+  *dst = unscaled;
+}
+
```
wence-, line 779: Right now, this function is only called with an `__int128_t *` `dst`. Does it have to be templated?

Relatedly, presumably there are some restrictions on valid parquet files? What if `len > 16` so that the left shift would overflow?

cpp/src/io/parquet/reader_impl_helpers.cpp
```diff
@@ -116,6 +119,10 @@ type_id to_type_id(SchemaElement const& schema,
           return type_id::DECIMAL128;
         }
       }
+      if (physical == parquet::BYTE_ARRAY) {
+        CUDF_EXPECTS(decimal_precision < max_decimal128_precision, "Invalid decimal precision");
```
wence-, line 123: Ah ok, we're asserting that `__int128_t` is big enough here.

cpp/src/io/parquet/reader_impl_helpers.cpp
```diff
@@ -77,14 +79,15 @@ type_id to_type_id(SchemaElement const& schema,
...
   // Logical type used for actual data interpretation; the legacy converted type
   // is superceded by 'logical' type whenever available.
   auto const inferred_converted_type = logical_type_to_converted_type(logical_type);
-  if (inferred_converted_type != parquet::UNKNOWN) converted_type = inferred_converted_type;
-  if (inferred_converted_type == parquet::DECIMAL && decimal_scale == 0)
-    decimal_scale = schema.logical_type.DECIMAL.scale;
+  if (inferred_converted_type != parquet::UNKNOWN) { converted_type = inferred_converted_type; }
+  if (inferred_converted_type == parquet::DECIMAL) {
+    decimal_precision = schema.logical_type.DECIMAL.precision;
+  }
```
wence-, line 90: precision not scale + clang-format

cpp/src/io/parquet/reader_impl_helpers.cpp
wence-: Is it the case that we always need to send back a `DECIMAL128` here, or could the same handling as for fixed length byte arrays specialise to a narrower type here too?

wence- (commented):
Python changes approved, then I had some minor queries on the C++ parts.

cpp/src/io/parquet/page_data.cu
revans2: Actually the parquet spec has no limit on the size of these arrays.

https://github.com/apache/parquet-format/blob/master/LogicalTypes.md

>    * fixed_len_byte_array: precision is limited by the array size. Length n can store <= floor(log_10(2^(8*n - 1) - 1)) base-10 digits
>   * binary: precision is not limited, but is required. The minimum number of bytes to store the unscaled value should be used.

We are imposing a 128-bit limit ourselves. That said I don't think it is common to have anything over that, so not making it templated should be fine.

cpp/src/io/parquet/reader_impl_helpers.cpp
revans2: Should this not be `<=`? We want to support a precision of 38 when reading byte arrays.

revans2 (changes requested):

cpp/src/io/parquet/reader_impl_helpers.cpp
revans2: > could the same handling as for fixed length byte arrays specialise to a narrower type here too?

The narrower type would be nice, as the type returned would only depend on the precision in the file and not the physical type it was stored as. But for me it is not critical.

cpp/src/io/parquet/page_data.cu
PointKernel: Good catch. I started with template since I tried narrow type in the first place. i.e. using the smallest decimal type based on its required precision. It didn't work out due to accuracy losses (still not sure why)

cpp/src/io/parquet/reader_impl_helpers.cpp
```diff
@@ -23,6 +23,8 @@ namespace cudf::io::detail::parquet {
 
 namespace {
 
+auto constexpr max_decimal128_precision = 38;
```
wence-, line 26: Perhaps comment where this number comes from?

$$
\lfloor \log_{10}{(2^{8 n - 1} - 1)} \rfloor
$$

with $n = 16$

Aside; for all $n \ge 1$ this approximation

$$
\lfloor (8 n - 1) \log_{10}{2} \rfloor
$$

is actually exact.

So one could (although arguably it's not really any clearer) use:
```c++
  auto constexpr max_supported_decimal_width = (sizeof(__int128_t) * CHAR_BIT) / 8;
  /* log10(2) = 0.3010299956639812 */
  std::uint32_t constexpr max_decimal_precision = (8 * max_supported_decimal_width - 1) * 0.3010299956639812;
```

cpp/src/io/parquet/page_data.cu
```diff
@@ -759,6 +759,24 @@ inline __device__ void gpuOutputInt64Timestamp(volatile page_state_s* s, int src
...
 
+/**
+ * @brief Output a byte array as int.
+ *
+ * @param[in] ptr Pointer to the byte array
+ * @param[in] len Byte array length
+ * @param[out] dst Pointer to row output data
+ */
+template <typename T>
+__device__ void gpuOutputByteArrayAsInt(char const* ptr, int32_t len, T* dst)
```
revans2, line 770: It looks like negative values are not being handled properly with this.

`0xFF` by itself should result in a `-1` being returned. But we are treating it the same as `0x00, 0xFF`. The sign of the Most Significant Byte should be extended through the rest of `dst`.

cpp/src/io/parquet/reader_impl_helpers.cpp
PointKernel: I'm using pre-commit hooks so the code should have been formatted before committing.

cpp/src/io/parquet/reader_impl_helpers.cpp
PointKernel: Added

cpp/src/io/parquet/page_data.cu
PointKernel: Fixed

cpp/src/io/parquet/reader_impl_helpers.cpp
PointKernel: "Narrower type" supported

cpp/src/io/parquet/page_data.cu
```diff
@@ -971,6 +996,11 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
         s->dtype_len = s->dtype_len <= sizeof(int32_t)   ? sizeof(int32_t)
                        : s->dtype_len <= sizeof(int64_t) ? sizeof(int64_t)
                                                          : sizeof(__int128_t);
+      } else if (s->col.converted_type == DECIMAL && data_type == BYTE_ARRAY) {
+        auto const decimal_precision = s->col.decimal_precision;
+        s->dtype_len = decimal_precision <= MAX_DECIMAL32_PRECISION   ? sizeof(int32_t)
+                       : decimal_precision <= MAX_DECIMAL64_PRECISION ? sizeof(int64_t)
+                                                                      : sizeof(__int128_t);
```
vuule, line 1003: could this be done in the switch above? The only different I see is in the value of `s->dtype_len_in`, which is probably irrelevant for byte arrays.

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -39,6 +39,12 @@ namespace cudf::io::parquet {
 
 using cudf::io::detail::string_index_pair;
 
+// Max decimal precisions according to the parquet spec:
+// https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal
+auto constexpr MAX_DECIMAL32_PRECISION  = 9;
+auto constexpr MAX_DECIMAL64_PRECISION  = 18;
+auto constexpr MAX_DECIMAL128_PRECISION = 38;  // log10(2^(sizeof(int128_t) * 8 - 1) - 1)
```
vuule, line 46: Parquet writer has these values hardcoded at https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/src/io/parquet/writer_impl.cu#L437
Would be nice to use these constants there as well.

cpp/src/io/parquet/reader_impl_helpers.cpp
```diff
@@ -77,14 +77,15 @@ type_id to_type_id(SchemaElement const& schema,
...
 
   // Logical type used for actual data interpretation; the legacy converted type
   // is superceded by 'logical' type whenever available.
   auto const inferred_converted_type = logical_type_to_converted_type(logical_type);
-  if (inferred_converted_type != parquet::UNKNOWN) converted_type = inferred_converted_type;
-  if (inferred_converted_type == parquet::DECIMAL && decimal_scale == 0)
-    decimal_scale = schema.logical_type.DECIMAL.scale;
+  if (inferred_converted_type != parquet::UNKNOWN) { converted_type = inferred_converted_type; }
+  if (inferred_converted_type == parquet::DECIMAL) {
+    decimal_precision = schema.logical_type.DECIMAL.precision;
```
vuule, line 87: `decimal_scale` was unused in data decode?

vuule (commented):
Looks pretty good. Couple of questions/suggestions.

cpp/src/io/parquet/reader_impl_helpers.cpp
wence-: Sorry, this was just a comment as to what had changed, nothing to do here.

cpp/src/io/parquet/reader_impl_helpers.cpp
PointKernel: yeah it's not used. It should be a leftover.

wence- (approved):

revans2 (approved):
looks great all of my tests pass

vuule (approved):

cpp/src/io/parquet/page_data.cu
```diff
@@ -959,7 +981,16 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
           [[fallthrough]];
         case DOUBLE: s->dtype_len = 8; break;
         case INT96: s->dtype_len = 12; break;
-        case BYTE_ARRAY: s->dtype_len = sizeof(string_index_pair); break;
+        case BYTE_ARRAY:
+          if (s->col.converted_type == DECIMAL) {
+            auto const decimal_precision = s->col.decimal_precision;
+            s->dtype_len = decimal_precision <= MAX_DECIMAL32_PRECISION   ? sizeof(int32_t)
+                           : decimal_precision <= MAX_DECIMAL64_PRECISION ? sizeof(int64_t)
+                                                                          : sizeof(__int128_t);
```
davidwendt, line 989: Would recommend using if-statements instead of nested ternaries like
https://github.com/rapidsai/cudf/pull/12205/files#diff-d42b9912ad4c2fb482d2c15ee3eb2d6935d878a17a436f7d39aa8c0ab2dcedc5R122
and
https://github.com/rapidsai/cudf/pull/12205/files#diff-672b345ca18d5957ee39bd7802bbf67f454cb941ab78174fbab3dc1ddaa048b3R1867

ttnghia (approved):
LGTM :+1:

cpp/src/io/parquet/page_data.cu
PointKernel: Good point. Done

PointKernel:
@gpucibot merge

GregoryKimball:
Nice work @PointKernel !

PR #12208, opened 2022-11-18
Adds in new java APIs for appending byte arrays to host columnar data

revans2:
## Description
Adds in a few new java APIs that can speed up appending lists of bytes to host data before it goes to the GPU.

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/HostColumnVector.java
+++ b/java/src/main/java/ai/rapids/cudf/HostColumnVector.java
@@ -1362,7 +1362,7 @@ public final class HostColumnVector extends HostColumnVectorCore {
       assert type.isBackedByByte();
       assert srcOffset >= 0;
       assert length >= 0;
-      assert value.length + srcOffset <= length;
+      assert length + srcOffset <= value.length;
 
       if (length > 0) {
         growFixedWidthBuffersAndRows(length);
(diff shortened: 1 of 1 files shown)


java/src/main/java/ai/rapids/cudf/HostColumnVector.java
```diff
@@ -1344,6 +1355,41 @@ public ColumnBuilder appendUTF8String(byte[] value, int srcOffset, int length) {
...
     }
 
+    /**
+     * Append multiple non-null byte values.
+     */
+    public ColumnBuilder append(byte[] value, int srcOffset, int length) {
+      assert type.isBackedByByte();
+      assert srcOffset >= 0;
+      assert length >= 0;
+      assert value.length + srcOffset <= length;
```
jbrennan333, line 1365: Should this be: `assert srcOffset + length <= value.length` ?

revans2: Yes (face palm) I copied and pasted that bug from another part of the code, that I think I wrote a long time ago. Thanks for catching it.

ttnghia: One question I'm thinking about most of the time with such code is: whether we should use `if(...) throw new exception()` or `assert`?

revans2: We try to favor asserts for things that actually touch the data because it can reduce the overhead. The asserts are enabled during tests so we can verify that things are working as expected, but disabled in production so we don't have the performance impact.

jbrennan333 (approved):
+1 lgtm

revans2:
@gpucibot merge

PR #12207, opened 2022-11-18
Support `replace` in `strings_udf`

brandon-b-miller:
This PR adds support for the following function in `strings_udf`:

- `str.replace`

Part of https://github.com/rapidsai/cudf/issues/9639

PR #12210, opened 2022-11-18
Add cudf::strings:udf::replace function

davidwendt:
## Description
Add udf function for replacing all occurrences of a target string within a source string with a replacement string.
This is to support #12207 

## Checklist

### DIFF
--- a/python/strings_udf/cpp/include/cudf/strings/udf/replace.cuh
+++ b/python/strings_udf/cpp/include/cudf/strings/udf/replace.cuh
@@ -37,19 +37,20 @@ __device__ inline udf_string replace(string_view source,
   if (target.empty()) return udf_string(source);
 
   udf_string result;
-
+  auto const tgt_length   = target.length();
+  auto const src_length   = source.length();
   size_type last_position = 0;
-  auto position           = source.find(target, 0);
+  size_type position      = 0;
   while (position != string_view::npos) {
-    auto left = source.substr(last_position, position - last_position);
-    result.append(left);
-    result.append(replacement);
-    last_position = position + target.length();
-    position      = source.find(target, last_position);
+    position = source.find(target, last_position);
+    if (position != string_view::npos) {
+      result.append(source.substr(last_position, position - last_position));
+      result.append(replacement);
+      last_position = position + tgt_length;
+    }
   }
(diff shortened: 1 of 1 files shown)


DaceT:
run benchmarks

python/strings_udf/cpp/include/cudf/strings/udf/replace.cuh
```diff
@@ -0,0 +1,61 @@
...
+ * @param source Source string to search
+ * @param target String to match within source
+ * @param replacement String to replace the target within the source
+ * @return Resulting string
+ */
+__device__ inline udf_string replace(string_view source,
+                                     string_view target,
+                                     string_view replacement)
+{
+  if (target.empty()) return udf_string(source);
```
bdice, line 37: Does this fit Python’s replace logic for empty targets? Or does pandas differ from this behavior / do we have an alternative function?
```python
>>> print('test'.replace('', 'a'))
ataeasata
```

brandon-b-miller (changes requested):
I tried patching https://github.com/rapidsai/cudf/pull/12207 locally with this code and I found that this function passes my tests, _except_ for the cases @bdice mentioned where an empty string is being replaced by a nonempty character or sequence. 

there, my shim function around this is
```c++
extern "C" __device__ int replace(
  int* nb_retval, void* udf_str, void* const src, void* const to_replace, void* const replacement)
{
  auto src_ptr         = reinterpret_cast<cudf::string_view const*>(src);
  auto to_replace_ptr  = reinterpret_cast<cudf::string_view const*>(to_replace);
  auto replacement_ptr = reinterpret_cast<cudf::string_view const*>(replacement);

  auto udf_str_ptr = new (udf_str) udf_string;
  udf_string result;

  result = replace(*src_ptr, *to_replace_ptr, *replacement_ptr);

  *udf_str_ptr = result;

  return 0;
}

bdice (approved):
LGTM if the tests for empty targets are passing. I suppose the tests aren't supposed to be in this PR, but another one (#12207).

brandon-b-miller (approved):
Tests pass in my PR now with these changes.

davidwendt:
@gpucibot merge

PR #12141, opened 2022-11-14
[WIP] Selectively use dictionary encoding in Parquet writer

etseidl:
## Description
The current implementation of Zstd compression in nvcomp is limited to 64k chunk sizes.  In some cases, column chunks will not be compressed because the dictionary page exceeds this size, but the data pages do not.  This PR attempts to fix this by selectively disabling dictionary encoding when the dictionary page is too large.  Dictionary encoding will still be used if the data reduction from encoding exceeds a 4:1 ratio (4:1 is chosen rather arbitrarily and needs refinement).

This is a clumsy first attempt to gather feedback and suggestions for other approaches to try.  

## Checklist
- [X] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #12211, opened 2022-11-19
Selectively use dictionary encoding in Parquet writer

etseidl:
## Description
With the addition of Zstandard compression, sometimes it is not desirable to perform dictionary encoding.  This PR adds a `dictionary_policy` setting to the parquet writer which allows either completely disabling dictionary encoding, setting a `max_dictionary_size` (like the [parquet-mr](https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/README.md) `parquet.dictionary.page.size` parameter), or using the default behavior before this PR (i.e. always use dictionaries). 

This PR also adds a `max_page_fragment_size` setting as a stop gap until an adaptive fragment size calculation can be implemented.

Continuation of #12141.

## Checklist


GPUtester:
Can one of the admins verify this patch?

Admins can comment `ok to test` to allow this one PR to run or `add to allowlist` to allow all future PRs from the same author to run.

etseidl:
Adding ability to control the page fragment size, nvbench will now run with ZSTD enabled (some tests were getting out of memory errors with nvcomp 2.5).
```
## parquet_write_io_compression

### [0] NVIDIA RTX A6000

|    io    | compression | cardinality | run_length | Samples | CPU Time | Noise  | GPU Time | Noise  | bytes_per_second | peak_memory_usage | encoded_file_size |
|----------|-------------|-------------|------------|---------|----------|--------|----------|--------|------------------|-------------------|-------------------|
| FILEPATH |      SNAPPY |           0 |          1 |      1x |  8.825 s |   inf% |  8.825 s |   inf% |         60838455 |         1.641 GiB |       520.394 MiB |
| FILEPATH |      SNAPPY |        1000 |          1 |      3x |  4.880 s |   inf% |  4.880 s |   inf% |        110009979 |         2.679 GiB |       170.059 MiB |
| FILEPATH |      SNAPPY |           0 |         32 |      5x |  1.342 s |  0.26% |  1.342 s |  0.26% |        400071465 |         2.675 GiB |        51.197 MiB |
| FILEPATH |      SNAPPY |        1000 |         32 |      5x |  1.327 s |  0.48% |  1.327 s |  0.48% |        404429732 |         2.679 GiB |        24.210 MiB |
| FILEPATH |        ZSTD |           0 |          1 |      3x |  3.900 s |   inf% |  3.900 s |   inf% |        137658926 |         8.089 GiB |       508.155 MiB |
| FILEPATH |        ZSTD |        1000 |          1 |      3x |  3.360 s |   inf% |  3.360 s |   inf% |        159762240 |         4.932 GiB |       166.636 MiB |
| FILEPATH |        ZSTD |           0 |         32 |      7x |  1.317 s | 24.31% |  1.317 s | 24.31% |        407760452 |        12.328 GiB |        44.184 MiB |
| FILEPATH |        ZSTD |        1000 |         32 |      5x |  1.083 s |  0.43% |  1.083 s |  0.43% |        495599009 |        14.225 GiB |        18.543 MiB |
| FILEPATH |        NONE |           0 |          1 |      6x |  2.370 s | 76.13% |  2.370 s | 76.13% |        226500420 |         1.641 GiB |       527.750 MiB |
| FILEPATH |        NONE |        1000 |          1 |      3x |  3.662 s |   inf% |  3.662 s |   inf% |        146587947 |         2.679 GiB |       178.888 MiB |
| FILEPATH |        NONE |           0 |         32 |      6x |  2.620 s |  3.81% |  2.620 s |  3.81% |        204890707 |         2.675 GiB |        58.172 MiB |
| FILEPATH |        NONE |        1000 |         32 |      8x |  1.093 s |  3.07% |  1.093 s |  3.07% |        491183652 |         2.679 GiB |        31.367 MiB |
```

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -321,40 +343,64 @@ static void batched_compress_async(compression_type compression,
...
 
+// since some compressors have extreme temp memory requirements, scale back the number
+// of chunks to process until the request for temp memory succeeds.
+std::pair<rmm::device_buffer, size_t> compress_temp_buffer(compression_type compression,
+                                                           size_t num_chunks,
+                                                           size_t max_uncomp_chunk_size,
+                                                           rmm::cuda_stream_view stream)
+{
+  // give up after 4 attempts
+  for (int i = 0; i < 4; i++) {
```
vuule, line 359: Should we give up here (outside of failing to allocate temp buffer for a single chunk)?

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -321,40 +343,64 @@ static void batched_compress_async(compression_type compression,
...
 
-bool is_aligned(void const* ptr, std::uintptr_t alignment) noexcept
+constexpr bool is_aligned(void const* ptr, std::uintptr_t alignment) noexcept
 {
   return (reinterpret_cast<std::uintptr_t>(ptr) % alignment) == 0;
 }
 
+// since some compressors have extreme temp memory requirements, scale back the number
+// of chunks to process until the request for temp memory succeeds.
+std::pair<rmm::device_buffer, size_t> compress_temp_buffer(compression_type compression,
```
vuule, line 353: This will be fun to implement with GetTempSizeEx (takes total uncompressed size instead of max chunk size).
Looks good for the current GetTempSize API.

cpp/include/cudf/io/parquet.hpp
```diff
@@ -967,6 +1026,60 @@ class parquet_writer_options_builder {
...
+    options.set_max_dictionary_size(val);
+    return *this;
+  }
+
+  /**
+   * @brief Sets the maximum page fragment size, in rows.
+   *
+   * Files with nested schemas or very long strings may need a page fragment size
+   * smaller than the default value of 5000 to ensure a single fragment will not
+   * exceed the desired maximum page size in bytes.
```
vuule, line 1072: Very needed documentation ❤️

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1042,7 +1059,7 @@ auto build_chunk_dictionaries(hostdevice_2dvector<gpu::EncColumnChunk>& chunks,
          col_desc[chunk.col_desc_id].physical_type == Type::BYTE_ARRAY)) {
       chunk.use_dictionary = false;
     } else {
-      chunk.use_dictionary = true;
+      chunk.use_dictionary = dict_policy != dictionary_policy::NEVER;
```
vuule, line 1062: Can we short-circuit `build_chunk_dictionaries` when `dictionary_policy == NEVER`?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1077,6 +1094,18 @@ auto build_chunk_dictionaries(hostdevice_2dvector<gpu::EncColumnChunk>& chunks,
       auto dict_enc_size = ck.uniq_data_size + rle_byte_size;
       if (ck.plain_data_size <= dict_enc_size) { return {false, 0}; }
 
+      // don't use dictionary if it gets too large for the given compression codec.
+      // but in the case where the dictionary encoding is giving large savings, prefer
+      // that over compression.
+      if (dict_policy == dictionary_policy::DEFAULT) {
+        constexpr int DICT_THRESHOLD = 4;  // FIXME: need a defensible value here
```
vuule, line 1101: Any ideas on how to pick a value for this?

vuule (commented):
I like all ideas in this PR, just unclear on some details.

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -178,6 +178,24 @@ void batched_decompress(compression_type compression,
   update_compression_results(nvcomp_statuses, actual_uncompressed_data_sizes, results, stream);
 }
 
+// FIXME(ets): workaround bug in zstd temp size calc. remove once it's fixed.
```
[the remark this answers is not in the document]
etseidl: You can ask Eric for the deets, but compilers can do some funny implicit conversions sometimes 😅

cpp/benchmarks/io/parquet/parquet_writer.cpp
```diff
@@ -197,8 +199,9 @@ using io_list = nvbench::enum_type_list<cudf::io::io_type::FILEPATH,
                                         cudf::io::io_type::HOST_BUFFER,
                                         cudf::io::io_type::VOID>;
 
-using compression_list =
-  nvbench::enum_type_list<cudf::io::compression_type::SNAPPY, cudf::io::compression_type::NONE>;
+using compression_list = nvbench::enum_type_list<cudf::io::compression_type::SNAPPY,
+                                                 cudf::io::compression_type::ZSTD,
```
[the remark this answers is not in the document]
etseidl: Ah, ok. I just put it in to get some comparison numbers...easy to remove.

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: Not sure what to do.  Originally I just had a `while (num_chunks > 1)`, and could switch back. I just didn't want to thrash *too* much, but then the allocate should fail pretty fast.

cpp/src/io/parquet/writer_impl.cu
etseidl: I'll investigate.  I haven't done the deep dive into `initialize_chunk_hash_maps` and `populate_chunk_hash_maps` to see if there are other things being initialized beyond the dictionaries.

cpp/src/io/parquet/writer_impl.cu
etseidl: A good ballpark estimate of the "average" compression achieved?  If the dictionary is doing well, you'd think the compressor would too, but then the dictionary is over an entire column chunk, while the compressor just gets individual pages.  Also, I added this before adding the `max_dictionary_size` param to the writer, so maybe stop trying to be clever here and just go back to a hard cutoff...if the dictionary is too big don't use it.

It would be nice if we already knew the compressed size before deciding (which I think is how compression is decided later on...aren't all pages compressed, and then compression kept only when the compressed size is smaller?). But then we'd have to encode each page twice to (once with plain encoding and once with dict) to do that comparison, so likely not worth it.

etseidl: I think this also made more sense when the max was 64KB.  With snappy unconstrained and Zstd good up to 16MB now, maybe it's better to remove this test.

etseidl:
Thanks for the great feedback @vuule :-)

cpp/src/io/parquet/writer_impl.cu
etseidl: I added the short-circuit and it doesn't appear to hurt anything. Still want to look a little closer and make sure.

cpp/src/io/comp/nvcomp_adapter.cpp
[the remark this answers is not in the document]
etseidl: Good news, the latest RC fixes this, so I'm removing it.

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -321,40 +323,66 @@ static void batched_compress_async(compression_type compression,
...
+    try {
+      auto const temp_size =
+        batched_compress_temp_size(compression, num_chunks, max_uncomp_chunk_size);
+      rmm::device_buffer buf(temp_size, stream);
+      return std::pair(std::move(buf), num_chunks);
+    } catch (rmm::bad_alloc& ba) {
+      // don't loop forever...if num_chunks is already 1, the following divide will also
+      // yield 1
+      if (num_chunks == 1) { break; }
+      num_chunks = (num_chunks + 1) / 2;
```
etseidl, line 348: @vuule rather than halving each time, do you think it would be better to divide num_chunks by an increasing factor?  So for one example, 3 15GB passes rather than 4 12.5GB passes.

cpp/src/io/comp/nvcomp_adapter.cpp
vuule: The way I see it: If we give up early the `write_parquet` call will throw. If we keep trying smaller batches, it might succeed, albeit with pretty bad performance. The second option seems preferable to me in most cases.

etseidl: It's back to a while loop now

cpp/src/io/comp/nvcomp_adapter.cpp
vuule: I think this mostly depends on the performance of failed allocations. We're trading off more failed allocations for larger batch size.
Followed up on this for rmm, and learned that rmm performance in this case depends on the fragmentation, as it need to iterate over all free blocks before finding there are no blocks that are large enough.
That said, I would guess that it's still faster to increase the factor linearly, as you suggest.

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: Easy enough to break out of the loop at a given scale so performance doesn't tank.  Suggestions for a sensible number of tries welcome :)

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: Is there a way to find the largest possible allocation in advance?  Then we could tune num_chunks to match that without actually trying the `allocate()` call.

vuule: Good point. Yes, it would be better to throw `bad_alloc` so caller can programmatically get the reason for failure and act accordingly (e.g. spilling).

vuule:
Marked as `non-breaking`, but unsure if that's right. @etseidl please correct if you think this could break any valid uses of the parquet writer.

etseidl:
The temp buffer rollback won't.  At worst it will eventually throw the `bad_alloc` that would have been thrown in the first place.  Now there is the question of whether handling of the OOM should happen in the library or let the user deal with it, but for my use case, where I'm using the chunked writer, it isn't always the first chunk that blows up.  Simply catching and trying the partial write again with a smaller slice of the table results in some metadata issues, so the real solution is to start over in that case.  As a user, I'd prefer the write just worked if possible. :)

The parameter changes might be breaking.  I can see turning off the dictionary encoding leading to much larger uncompressed pages that might trigger an OOM somewhere down the line.  That's why I put in the logic to prefer dictionary encoding over compression when the dictionary gives a big enough benefit.  (The definition of "big enough" is still open to debate).  The param changes can also result in larger (or smaller) files than currently, but I don't know if that constitutes a breaking change as you all define it.

cpp/include/cudf/io/parquet.hpp
```diff
@@ -967,6 +1026,60 @@ class parquet_writer_options_builder {
...
+   * compression for that column.
+   *
+   * @param val policy for dictionary use
+   * @return this for chaining
+   */
+  parquet_writer_options_builder& dictionary_policy(dictionary_policy val)
+  {
+    options.set_dictionary_policy(val);
+    return *this;
+  }
```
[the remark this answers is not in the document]
davidwendt: That would be great. We should have host definitions in .cpp files. There are exceptions for things like templated functions. If you feel this is out-of-scope for this PR then I can do that in my 12372 PR.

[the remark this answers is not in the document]
etseidl: No, totally in-scope.  I'll just wait for your PR to merge (I *think* that will result in fewer merge conflicts on my end), and then follow suit here.  Thanks!

cpp/benchmarks/io/parquet/parquet_writer.cpp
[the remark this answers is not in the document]
etseidl: removed for now

cpp/src/io/parquet/writer_impl.cu
etseidl: Even though this was the original intent of the PR, I'm going to punt for now because I think the other changes are more important to get Zstd compression working with nvcomp 2.5.  Without this, one of my tests still throws a memory error, but I can work around that by either setting dictionary policy to `ALWAYS` or bumping up the max dictionary size above the default 1MB.  I do want to revisit this though.

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: Modifed to re-throw the `bad_alloc` when giving up.

cpp/include/cudf/io/parquet.hpp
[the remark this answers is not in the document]
etseidl: Done, if you want to take a quick peek @davidwendt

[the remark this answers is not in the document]
davidwendt: Looks good to me. Thanks.

cpp/tests/io/parquet_test.cpp
```diff
@@ -4809,6 +4809,147 @@ TEST_F(ParquetWriterTest, SingleValueDictionaryTest)
...
+      .dictionary_policy(cudf::io::dictionary_policy::DEFAULT);
+  cudf::io::write_parquet(out_opts);
+
+  cudf::io::parquet_reader_options default_in_opts =
+    cudf::io::parquet_reader_options::builder(cudf::io::source_info{filepath});
+  auto const result = cudf::io::read_parquet(default_in_opts);
+
+  CUDF_TEST_EXPECT_TABLES_EQUAL(expected, result.tbl->view());
+
+  // make sure dictionary was used
```
hyperbolic2346, line 4885: this one is a mixed bag, right?

hyperbolic2346 (approved):
Small nit about a test comment, but looks great to me! Thanks!

cpp/tests/io/parquet_test.cpp
etseidl: ah, cut and paste 😅 I'll clean that up

cpp/include/cudf/io/parquet.hpp
```diff
@@ -487,6 +489,12 @@ class parquet_writer_options {
   size_type _max_page_size_rows = default_max_page_size_rows;
   // Maximum size of min or max values in column index
   int32_t _column_index_truncate_length = default_column_index_truncate_length;
+  // When to use dictionary encoding for data
+  dictionary_policy _dictionary_policy = dictionary_policy::ADAPTIVE;
```
vuule, line 493: Is this in line with parquet-mr default behavior? Their docs are a bit vague on the impact of `parquet.dictionary.page.size`.

etseidl: parquet-mr does limit dictionaries to `parquet.dictionary.page.size`, but I don't believe they have any coupling between dictionary encoding and compression like this. I also think they'll still use the dictionary until it gets too big, and then do plain encoding for the rest of the chunk.

vuule:
This should have been multiple PRs, there's so much going on :sweat_smile:

I feel like 1MB limit is too low, given that we support 24bit dictionaries. Maybe we should change the default limit to 16MB (matches the 24 bits if you don't think about it :D).
I'm not sure about the default. I don't see good impact on our benchmarks, do you have a case where this option helps significantly (outside of comp limit)?

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -321,40 +324,68 @@ static void batched_compress_async(compression_type compression,
...
+// since some compressors have extreme temp memory requirements, scale back the number
+// of chunks to process until the request for temp memory succeeds.
+std::pair<rmm::device_buffer, size_t> compress_temp_buffer(compression_type compression,
+                                                           size_t num_chunks,
+                                                           size_t max_uncomp_chunk_size,
+                                                           rmm::cuda_stream_view stream)
+{
+  auto scaled_num_chunks = num_chunks;
+  size_t scale           = 1;
+  while (scaled_num_chunks > 0) {
```
vuule, line 341: This is a potential perf regression for Spark, since they have an option to spill and make more memory available. If we end up compressing a single chunk at a time, performance could be a lot worse than spilling. CC @revans2 
Maybe we should bail at `sqrt(num_chunks)`? It makes the minimum number of concurrent chunks the same as the number of retries ¯\_(ツ)_/¯

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -321,40 +324,68 @@ static void batched_compress_async(compression_type compression,
...
+      rmm::device_buffer buf(temp_size, stream);
+      return std::pair(std::move(buf), scaled_num_chunks);
+    } catch (rmm::bad_alloc& ba) {
+      // don't loop forever...if scaled_num_chunks is already 1, the following divide will
+      // also yield 1
+      if (scaled_num_chunks == 1) { throw ba; }
+      scaled_num_chunks = util::div_rounding_up_safe(num_chunks, ++scale);
+    }
+  }
+  CUDF_FAIL("Cannot allocate temp buffer for compression");
```
vuule, line 354: We should never get to this point. Does the compiler complain without this line?

vuule (commented):
Few nitpicks and one more design decision to make :)

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1077,6 +1102,16 @@ auto build_chunk_dictionaries(hostdevice_2dvector<gpu::EncColumnChunk>& chunks,
       auto dict_enc_size = ck.uniq_data_size + rle_byte_size;
       if (ck.plain_data_size <= dict_enc_size) { return {false, 0}; }
 
+      // don't use dictionary if it gets too large for the given compression codec.
+      // but in the case where the dictionary encoding is giving large savings, prefer
+      // that over compression.
```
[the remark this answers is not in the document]
etseidl: nice catch. :)

etseidl:
> This should have been multiple PRs, there's so much going on 😅

Sorry, this series of PRs has suffered from feature creep, but at least I removed some of the more speculative parts 😅 

> I feel like 1MB limit is too low, given that we support 24bit dictionaries. Maybe we should change the default limit to 16MB (matches the 24 bits if you don't think about it :D). I'm not sure about the default. I don't see good impact on our benchmarks, do you have a case where this option helps significantly (outside of comp limit)?

I've got to go back and rerun some tests...total amnesia from before the holidays.  parquet-mr uses 1MB, and my gut says 16MB is too big, so maybe something in between?  Then again, I'm a parameter fiddler and don't mind tuning to my needs, so I could go either way.  The spark team will have seen a lot more cases than me, so maybe @revans2 can weigh in on this as well 🙏

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: I've been all over the place on this and don't have my heart set on anything. I'm wondering, though, if the spill would happen on file writing, or just during query execution?  Or does Spark spill to parquet files?

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: I will try compiling without it with debug on (that seems to have pickier options set).  I would hope it warns about not returning a value (unless the compiler is really smart and can tell that the function either returns or throws `bad_alloc`). I just want some kind of annotation that that line should not be reached.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1077,6 +1102,16 @@ auto build_chunk_dictionaries(hostdevice_2dvector<gpu::EncColumnChunk>& chunks,
...
+      if (dict_policy == dictionary_policy::ADAPTIVE) {
+        auto const unique_size = static_cast<size_t>(ck.uniq_data_size);
+        if (unique_size > max_page_bytes(compression, std::min(unique_size, max_dict_size))) {
```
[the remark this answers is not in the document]
etseidl: This must have made sense at some point, but you're right, it doesn't now.

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: yep, compile fails without that line

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: After running some tests, I'm wondering if this should go back to a geometric scaling rather than linear.  In one case It tried 41 times until succeeding 😮

vuule: wow! What was the initial number of chunks?
I'm okay with geometric :+1:

cpp/src/io/comp/nvcomp_adapter.cpp
vuule: Alright, a load-bearing unreachable line :)

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: > wow! What was the initial number of chunks?

Around 2400.  Oddly, I get worse execution time with geometric scaling since it overshoots most of the time, resulting in more compression batches than necessary.

vuule: Hmm. I assume that's doubling at each step. Have you tried geometric with lower ratio?

cpp/src/io/comp/nvcomp_adapter.cpp
vyasr: You can, in principle, use [`CUDF_UNREACHABLE`](https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/include/cudf/detail/utilities/assert.cuh#L57) for this use case. In practice we do prefer `CUDF_FAIL` unless this is in device code since the main benefit of the macro is to help the compiler optimize out those code paths and we're unlikely to ever care about that in host code.

cpp/src/io/comp/nvcomp_adapter.cpp
etseidl: `sqrt(2)` worked pretty well.  Got 41 iterations down to 11, and still got good performance.  How's that for a compromise?

vuule: Sounds good!

etseidl:
@vuule now that nvcomp 2.5 is out and being used in builds, do you think this PR is ready to go?  Could you ok this to test?

vuule:
> @vuule now that nvcomp 2.5 is out and being used in builds, do you think this PR is ready to go? Could you ok this to test?

@etseidl Spark team wants to test/benchmark this PR, mainly because of the OOM mitigation changes.
Thoughts on raising the minimum chunk count when allocating the temp buffer? If your workflow really requires backing off to a single chunk, maybe the minimum count should be configurable.

etseidl:
> Thoughts on raising the minimum chunk count when allocating the temp buffer? If your workflow really requires backing off to a single chunk, maybe the minimum count should be configurable.

For parquet, a chunk is a page (not a column chunk, ahhh overloaded terminology) so if there's a case that requires a single chunk something is *very* wrong.  I can't imagine (although I'm sure there are cases) wanting pages much larger than 1MB, so we should still be able to get hundreds of chunks in one go.  And with 2.6 imminent, maybe much more.  I'd assume it's similar in ORC?  I'm fine with adding a minimum chunk size parameter with a default around 500.  Would have to come up with a good name for it to not get confused...maybe `minimum_compression_batch_size`?

So if the minimum is reached, still throw `bad_alloc` right?

vuule:
ORC compresses blocks that are not tied to encode units (i.e. pages) so there are no rogue 10MB blocks, like dict pages in Parquet :) Block are normally ~100KB.
`minimum_compression_batch_size` sounds good.

> So if the minimum is reached, still throw bad_alloc right?
Yes.

> with a default around 500
Should we do something like 512MB/page_size?

> Also, should I retarget for nvcomp 2.6?
I think this one should be 2.6 agnostic. I'm starting today on a PR that makes use of nvCOMP 2.6. I think testing can cover 2.6 with my PR, with and without your PR.

etseidl:
> with a default around 500 Should we do something like 512MB/page_size?

Oh, good point.  I've been testing with 64KB pages, so maybe 500 is too big.  Will have to test with bigger pages.  Maybe use the number of SMs as a lower bound?  I need to ask Eric what grid sizes he uses? Looking at nsys profiles I'm not seeing a 1:1 page:thread_block.

Anyway, I'll start on adding the parameter.  Thanks for the feedback!

etseidl:
Just tested with nvcomp 2.6, and I'm thinking we should punt on the memory scaling for now.  The temp size requirements are way smaller now. How about just doing the fragment size and dictionary policy for this PR, and then revisit the scaling if need be?

vuule:
> Just tested with nvcomp 2.6, and I'm thinking we should punt on the memory scaling for now. The temp size requirements are way smaller now. How about just doing the fragment size and dictionary policy for this PR, and then revisit the scaling if need be?

Torn on this, but I guess it's fine, we can always revisit later. Also, the new 2.6 API impacts this part so it will need to be adapted or limited to <2.6.

etseidl:
Reverted. Now this is just the dictionary policy and fragment size changes, with defaults being the old behavior.

cpp/src/io/parquet/writer_impl.hpp
```diff
@@ -214,6 +218,9 @@ class writer::impl {
   size_t max_page_size_bytes           = default_max_page_size_bytes;
   size_type max_page_size_rows         = default_max_page_size_rows;
   statistics_freq stats_granularity_   = statistics_freq::STATISTICS_NONE;
+  dictionary_policy dict_policy_       = dictionary_policy::ADAPTIVE;
```
vuule, line 221: should this be ALWAYS?

etseidl: arghhh.  yes.

vuule (approved):
I think this is finally it.
Thank you for the numerous changes made during review.

etseidl:
Thanks for all the help @vuule.  Is this ok to test then?

vuule:
ok to test

etseidl:
I think it needs a '/ok' now 😅

vuule:
rerun tests

vuule:
OMG

etseidl:
that should have worked 😮

etseidl:
one more time, please 😭

Issue #11975, opened 2022-10-24
[FEA] cudf.Series.hasnans


mattf:
**Is your feature request related to a problem? Please describe.**
rewriting code from pandas to cudf, trying to use `import cudf as pd`

**Describe the solution you'd like**
`cudf.Series.hasnans` matching https://pandas.pydata.org/docs/reference/api/pandas.Series.hasnans.html (`bool(cudf.Series.isna().any())`)

PR #12214, opened 2022-11-21
[REVIEW] Add `Series.hasnans` and `Index.hasnans`

galipremsagar:
## Description
Resolves: #11975 

This PR adds support for `Series.hasnans` and `Index.hasnans`

## Checklist

### DIFF
--- a/python/cudf/cudf/core/index.py
+++ b/python/cudf/cudf/core/index.py
@@ -281,6 +281,11 @@ class RangeIndex(BaseIndex, BinaryOperand):
 
     @property  # type: ignore
     @_cudf_nvtx_annotate
+    def hasnans(self):
+        return False
+
+    @property  # type: ignore
+    @_cudf_nvtx_annotate
     def _data(self):
         return cudf.core.column_accessor.ColumnAccessor(
             {self.name: self._values}
@@ -1390,6 +1395,11 @@ class GenericIndex(SingleColumnFrame, BaseIndex):
 
     def is_interval(self):
         return False
+
+    @property  # type: ignore
+    @_cudf_nvtx_annotate
+    def hasnans(self):
+        return self._column.has_nulls(include_nan=True)
 
     @_cudf_nvtx_annotate
     def argsort(
--- a/python/cudf/cudf/core/series.py
+++ b/python/cudf/cudf/core/series.py
@@ -671,6 +671,48 @@ class Series(SingleColumnFrame, IndexedFrame, Serializable):
         """
         return [self.index]
 
+    @property  # type: ignore
+    @_cudf_nvtx_annotate
+    def hasnans(self):
+        """
+        Return True if there are any NaNs.
+
+        Returns
+        -------
+        out : bool
+            If Series has at least one NaN value, return True, if not
+            return False.
+
+        Examples
+        --------
+        >>> import cudf
+        >>> import numpy as np
+        >>> series = cudf.Series([1, 2, np.nan, 3, 4], nan_as_null=False)
+        >>> series
+        0    1.0
+        1    2.0
+        2    NaN
+        3    3.0
+        4    4.0
+        dtype: float64
+        >>> series.hasnans
+        True
+
+        `hasnans` returns `True` for the presence of any `NA` values:
+
+        >>> series = cudf.Series([1, 2, 3, None, 4])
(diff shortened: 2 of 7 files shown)


python/cudf/cudf/core/index.py
```diff
@@ -1391,6 +1396,11 @@ def is_categorical(self):
     def is_interval(self):
         return False
 
+    @property  # type: ignore
+    @_cudf_nvtx_annotate
+    def hasnans(self):
+        return self._column.has_nulls(include_nan=True)
```
bdice, line 1402: This looks like it returns "has NaNs OR nulls." How do we want to handle this? If we are including nulls in the definition of `hasnans`, we need to explicitly document that.

python/cudf/cudf/tests/test_index.py
```diff
@@ -2795,3 +2795,21 @@ def test_index_error_list_index():
...
+        [np.nan, 10, 15, 16],
+        range(0, 10),
+        [np.nan, None, 10, 20],
+        ["ab", "zx", "pq"],
+        ["ab", "zx", None, "pq"],
+    ],
+)
+def test_index_hasnans(data):
+    gs = cudf.Index(data, nan_as_null=False)
+    ps = gs.to_pandas()
```
bdice, line 2813: Why does the other test use `nullable=True` but not this one?

galipremsagar: We are yet to add this support of creating a nullable dtypes in pandas Index: https://github.com/rapidsai/cudf/issues/12216

python/cudf/cudf/core/series.py
```diff
@@ -671,6 +671,48 @@ def axes(self):
...
+        """
+        Return True if there are any NaNs.
+
+        Returns
+        -------
+        out : bool
+            If Series has at least one NaN value, return True, if not
+            return False.
+
+        Examples
```
galipremsagar, line 686:
```suggestion
        Return True if there are any NaNs or nulls.

        Returns
        -------
        out : bool
            If Series has at least one NaN or null value, return True,
            if not return False.

        Examples
```

python/cudf/cudf/core/index.py
galipremsagar: Updated the docstring to be explicit rather than mentioning just in an example.

python/cudf/cudf/tests/test_index.py
bdice: Can you add a note in that issue to follow up and fix this?

python/cudf/cudf/core/index.py
bdice: Thanks!

bdice (approved):

galipremsagar:
rerun tests

python/cudf/cudf/core/series.py
```diff
@@ -671,6 +671,48 @@ def axes(self):
...
+        0       1
+        1       2
+        2       3
+        3    <NA>
+        4       4
+        dtype: int64
+        >>> series.hasnans
+        True
+        """
+        return self._column.has_nulls(include_nan=True)
```
ttnghia, line 714: Why there is calling `has_nulls` in the `has_nans` function?

galipremsagar: Because pandas counts `nan` & `null` values in `hasnans`

ttnghia: > Return True if there are any NaNs or nulls.

If we're checking both nulls and NaNs then the function name should be `has_null_or_nan` instead. Calling `has_nans` but the output is `true` if having nulls is very misleading.

galipremsagar: It is an established naming convention for pandas, for example, `isna`, `isnull` all account for `nan` & `null` values. We will need to have `hasnan` for parity with pandas similar to the way we do for `isna` & `isnull`. I can add an alias `has_null_or_nan` if you want, but that'll raise the question for what about the rest of the APIs?

ttnghia: Keeping the names on par with pandas this way would be really annoying. But I don't want to mess up with this so I'm fine with the name for now :(

galipremsagar:
@gpucibot merge

PR #12220, opened 2022-11-22
Clean up of `test_spilling.py`

madsbk:
Based on @vyasr's [review](https://github.com/rapidsai/cudf/pull/12106#pullrequestreview-1183334305), I am cleaning up the spilling tests. Hopefully making them more readable. 

Notice, I am targeting `branch-23.02` because of code freeze but maybe we should target `branch-22.12` since it implements spilling?

### DIFF
--- a/python/cudf/cudf/tests/test_spilling.py
+++ b/python/cudf/cudf/tests/test_spilling.py
@@ -45,11 +45,11 @@ if get_global_manager() is not None:
     )
 
 
-def std_df(target="gpu") -> cudf.DataFrame:
+def single_column_df(target="gpu") -> cudf.DataFrame:
     """Create a standard single column dataframe used for testing
 
-    Use `std_df_data`, `std_df_base_data`, `gen_df_data_nbytes`
-    for easy access to the buffer of the column.
+    Use `single_column_df_data`, `single_column_df_base_data`,
+    `gen_df_data_nbytes` for easy access to the buffer of the column.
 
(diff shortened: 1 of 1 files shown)


vyasr:
No need to target 22.12 unless test cleanup reveals serious bugs in the spilling code.

madsbk:
rerun tests

python/cudf/cudf/tests/test_spilling.py
```diff
@@ -45,17 +45,47 @@
     )
 
 
-def gen_df(target="gpu") -> cudf.DataFrame:
+def std_df(target="gpu") -> cudf.DataFrame:
```
shwina, line 48:
```suggestion
def single_column_df(target="gpu") -> cudf.DataFrame:
```

shwina: And ditto below?

shwina: Perhaps we can also consider changing `target="gpu"` to `spilled=True | False` -- but that is less important.

madsbk: Renamed `std_df` => `single_column_df` but I like to keep `target="gpu"`. When we add spilling-to-disk, we introduce `target="disk"`.

madsbk:
@shwina, thanks for the review. Do you have anything else?

shwina (approved):

shwina:
@gpucibot merge

PR #9543, opened 2021-10-27
[JAVA] Adds cudaProfilerStart/cudaProfilerStop in JNI api

abellina:
Closes #9399 

This PR adds support for `cudaProfilerStart` and `cudaProfilerStop`. I did not add a test for this case as I don't think it makes sense, given that a user could be legitimately capturing a profile of the tests (so a start/stop could change things).

The way one can use this with `nsys` is:

```
nsys profile -c cudaProfilerApi --capture-range-end=repeat
```

Where there are other options for what to do on capture end according to [the docs](https://docs.nvidia.com/nsight-systems/UserGuide/index.html). The command above stops collecting, and restarts again when `cudaProfilerStart` is invoked. The nsys profiled process doesn't need to be restarted, which is useful if one is trying to capture several experiments back-to-back.

There is a way to do this with nvtx ranges triggering the capture, but I could not get it to work with named ranges. I was told to try registering the name of the nvtx range, and we can still try this, but I felt the `cudaProfiler*` api was useful on its own (and for our purposes).

PR #12221, opened 2022-11-22
Use CUDF_JNI_ENABLE_PROFILING to conditionally enable profiling support.

bdice:
## Description
The conda cudatoolkit packages do not include `<cuda_profiler_api.h>`. This PR makes CUDA profiling (introduced in #9543) optional in the JNI layer. By default, profiling will be enabled. In GitHub Actions (#12002), profiling will be disabled.

## Checklist


java/src/main/native/CMakeLists.txt
```diff
@@ -87,6 +88,10 @@ if(CUDF_USE_PER_THREAD_DEFAULT_STREAM)
   add_compile_definitions(CUDA_API_PER_THREAD_DEFAULT_STREAM CUDF_USE_PER_THREAD_DEFAULT_STREAM)
 endif()
 
+if(CUDF_JNI_ENABLE_PROFILING)
+  target_compile_definitions(cudfjni PUBLIC CUDF_JNI_ENABLE_PROFILING)
```
bdice, line 92: @rapidsai/rapids-cmake-codeowners Can someone look at this CMake? Do I need a value here or something like `"CUDF_JNI_ENABLE_PROFILING=1"`? I don't feel super confident in this.

abellina:
The JNI side of this looks great, I am not too sure on the CMake side of things either.

java/src/main/native/CMakeLists.txt
jbrennan333: I'm not a CMake expert, but this looks correct to me.

jbrennan333 (approved):
+1 this looks good to me

abellina:
JNI build failed on:

> 08:13:45      [exec] CMake Error at CMakeLists.txt:92 (target_compile_definitions):
08:13:45      [exec]   Cannot specify compile definitions for target "cudfjni" which is not built
08:13:45      [exec]   by this project.

java/src/main/native/CMakeLists.txt
jbrennan333: Based on the build failure, I think this should be `add_compile_definitions` like the code just above.

bdice: I am going to push a fix for this shortly. I think this CMake file has some issues. We _should_ prefer `target_compile_definitions` to keep the scope limited, but `add_library(cudfjni)` occurs lower in the script. All `target_compile_definitions` must happen after the library is defined. This is also an issue for `USE_NVTX`, which I was using as an example. I am guessing that we have not recently tested the build with that option turned off.

robertmaynard: both `CUDF_JNI_ENABLE_PROFILING` and `CUDF_JNI_ENABLE_PROFILING=1` are valid. It depends on if you want to define a name or a name with value.

```cpp
#define CUDF_JNI_ENABLE_PROFILING 
// versus
#define CUDF_JNI_ENABLE_PROFILING 1
```

robertmaynard: But yes this needs to go after the construction of the target `cudfjni`

robertmaynard (changes requested):

java/src/main/native/CMakeLists.txt
```diff
@@ -152,6 +148,15 @@ add_library(
   src/check_nvcomp_output_sizes.cu
 )
 
+# Disable NVTX if necessary
+if(NOT USE_NVTX)
+  target_compile_definitions(cudfjni PUBLIC NVTX_DISABLE)
+endif()
+
+if(CUDF_JNI_ENABLE_PROFILING)
+  target_compile_definitions(cudfjni PUBLIC CUDF_JNI_ENABLE_PROFILING)
```
bdice, line 157:
```suggestion
  target_compile_definitions(cudfjni PRIVATE CUDF_JNI_ENABLE_PROFILING)
```

java/src/main/native/CMakeLists.txt
```diff
@@ -152,6 +148,15 @@ add_library(
   src/check_nvcomp_output_sizes.cu
 )
 
+# Disable NVTX if necessary
```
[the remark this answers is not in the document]
bdice: Discussed offline. `CUDF_JNI_ENABLE_PROFILING` should be private. `NVTX_DISABLE` probably needs to stay public (it's also public in the cudf target).

robertmaynard (approved):

bdice:
@gpucibot merge

PR #12227, opened 2022-11-22
Add cudf::strings::reverse function

davidwendt:
## Description
Adds `cudf::strings::reverse` function.
This is to support https://github.com/NVIDIA/spark-rapids/issues/6885

## Checklist


cpp/tests/strings/reverse_tests.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <cudf/strings/strings_column_view.hpp>
+
+#include <vector>
+
+struct StringsReverseTest : public cudf::test::BaseFixture {
+};
+
+TEST_F(StringsReverseTest, Reverse)
+{
+  cudf::test::strings_column_wrapper input({"abcdef", "12345", "", "", "aébé", "A é Z"});
```
bdice, line 32: Can we include a null value in some test?

davidwendt: Yes, I have more tests coming. This is a draft PR.

cpp/src/strings/reverse.cu
```diff
@@ -0,0 +1,89 @@
...
+}  // namespace
+
+std::unique_ptr<column> reverse(strings_column_view const& input,
+                                rmm::cuda_stream_view stream,
+                                rmm::mr::device_memory_resource* mr)
+{
+  if (input.is_empty()) { return make_empty_column(type_id::STRING); }
+
+  // copy the column; replace data in the chars column
+  auto result = std::make_unique<column>(input.parent(), stream, mr);
```
hyperbolic2346, line 65: Is this making a full copy of the string data and then replacing it in the functor?

davidwendt: Yes, that was the intention of the comment. Should've been worded better?

robertmaynard (approved):
Approving CMakes changes

cpp/benchmarks/string/reverse.cpp
```diff
@@ -0,0 +1,57 @@
...
+  auto const num_rows  = static_cast<cudf::size_type>(state.get_int64("num_rows"));
+  auto const row_width = static_cast<cudf::size_type>(state.get_int64("row_width"));
+
+  if (static_cast<std::size_t>(num_rows) * static_cast<std::size_t>(row_width) >=
+      static_cast<std::size_t>(std::numeric_limits<cudf::size_type>::max())) {
+    state.skip("Skip benchmarks greater than size_type limit");
+  }
+
+  data_profile const table_profile = data_profile_builder().distribution(
+    cudf::type_id::STRING, distribution_id::NORMAL, 0, row_width);
```
ttnghia, line 38: I'm not sure if generating benchmark data this way will have 2-bytes characters?

cpp/src/strings/reverse.cu
ttnghia: I see that the char child of the result column will be overwritten. Should we just copy the offsets over?

cpp/src/strings/reverse.cu
```diff
@@ -0,0 +1,89 @@
...
+  column_device_view const d_strings;
+  offset_type const* d_offsets;
+  char* d_chars;
+
+  __device__ void operator()(size_type idx)
+  {
+    if (d_strings.is_null(idx)) { return; }
+    auto const d_str = d_strings.element<string_view>(idx);
+    // pointer to the end of the output string
+    auto d_output = d_chars + d_offsets[idx] + d_str.size_bytes();
```
ttnghia, line 48: Is this `d_offsets` always zero-based? There is a test with sliced column below so I think it should be 0-based and just want to ask to confirm.

cpp/src/strings/reverse.cu
davidwendt: This one step does everything we need. It builds the output column and normalizes the offsets if the input is sliced.

cpp/benchmarks/string/reverse.cpp
davidwendt: The `create_random_table` does generate multibyte UTF-8 characters for `STRING` columns.

cpp/src/strings/reverse.cu
davidwendt: The copy-column step ensures the offsets here will always be 0-based here.

ttnghia (approved):

cpp/tests/strings/reverse_tests.cpp
```diff
@@ -0,0 +1,54 @@
...
+struct StringsReverseTest : public cudf::test::BaseFixture {
+};
+
+TEST_F(StringsReverseTest, Reverse)
+{
+  auto input = cudf::test::strings_column_wrapper(
+    {"abcdef", "12345", "", "", "aébé", "A é Z", "X", "é"}, {1, 1, 1, 0, 1, 1, 1, 1});
+  auto results  = cudf::strings::reverse(cudf::strings_column_view(input));
+  auto expected = cudf::test::strings_column_wrapper(
+    {"fedcba", "54321", "", "", "ébéa", "Z é A", "X", "é"}, {1, 1, 1, 0, 1, 1, 1, 1});
```
cwharris, line 37: Are any of these multiple byte characters? Should we have multiple byte characters in the test? I don't recall if we support them or not.

cwharris (approved):
really nice!

cpp/tests/strings/reverse_tests.cpp
davidwendt: All of the accented ones (`é`) are multi-byte UTF-8 characters.

AyodeAwe (approved):

bdice (approved):
LGTM.

One interesting thing that came up while thinking hard about reversed strings: [Emoji ZWJ sequences](https://emojipedia.org/emoji-zwj-sequence/). I don't think we need special handling here, just wanted to share my findings for curiosity.

An emoji like [Woman Shrugging, Medium Skin Tone](https://emojipedia.org/woman-shrugging-medium-skin-tone/) 🤷🏽‍♀️ is composed of a sequence of 5 code points:
[🤷 U+1F937](https://emojipedia.org/emoji/%F0%9F%A4%B7/)
[🏽 U+1F3FD](https://emojipedia.org/emoji/%F0%9F%8F%BD/)
U+200D
[♀ U+2640](https://emojipedia.org/emoji/%E2%99%80/)
U+FE0F

Python reverses the sequence literally, with the code points in the opposite order, thereby no longer making a single emoji but a string that renders like `️♀‍🏽🤷`. I would expect this PR to  behave similarly because each Unicode code point is handled individually, so I don't think additional testing or work is needed.

davidwendt:
@gpucibot merge

PR #12229, opened 2022-11-22
Cover parsing to decimal types in `read_json` tests

### LINKED ISSUE #9791: Add decimal128 tests to CSV/JSON
- [x] CSV reader
- [x] CSV writer
- [ ] JSON reader

vuule:
## Description
Closes #9791

Adds `read_json` tests for decimal types. 
Tests are equivalent to existing `read_csv` tests for decimal types.
Tests invoke both JSON parsers.

## Checklist


cpp/tests/io/json_test.cpp
```diff
@@ -215,6 +216,48 @@ std::string to_records_orient(std::vector<std::map<std::string, std::string>> co
   return (result + suffix);
 }
 
+template <typename DecimalType>
+struct JsonFixedPointReaderTest : public JsonReaderTest {
+  void run_test(std::vector<std::string> const& reference_strings,
+                numeric::scale_type scale,
+                bool use_experimental_parser)
+  {
+    cudf::test::strings_column_wrapper strings(reference_strings.begin(), reference_strings.end());
```
ttnghia, line 225: Please add `const` whenever possible. Everywhere from now.

cpp/tests/io/json_test.cpp
```diff
@@ -215,6 +216,48 @@ std::string to_records_orient(std::vector<std::map<std::string, std::string>> co
...
+    auto input_column = cudf::strings::to_fixed_point(cudf::strings_column_view(strings),
```
ttnghia, line 226: I see this seems to be `expected` instead of `input`.

cpp/tests/io/json_test.cpp
```diff
@@ -215,6 +216,48 @@ std::string to_records_orient(std::vector<std::map<std::string, std::string>> co
...
+                                                      data_type{type_to_id<DecimalType>(), scale});
+
+    std::string buffer =
```
mythrocks, line 229: Since `c_str()` and `size()` are `const` functions, I suspect `buffer` can be `const`.

cpp/tests/io/csv_test.cpp
```diff
@@ -95,7 +95,6 @@ struct CsvReaderNumericTypeTest : public CsvReaderTest {
 using SupportedNumericTypes = cudf::test::Types<int64_t, double>;
 TYPED_TEST_SUITE(CsvReaderNumericTypeTest, SupportedNumericTypes);
 
-// Typed test to be instantiated for numeric::decimal32 and numeric::decimal64
```
mythrocks, line 98: Was this deletion intentional?

cpp/tests/io/json_test.cpp
```diff
@@ -1563,4 +1606,23 @@ TEST_P(JsonReaderParamTest, JsonDtypeParsing)
   }
 }
 
+TYPED_TEST(JsonFixedPointReaderTest, SingleColumnNegativeScale)
+{
+  this->run_tests({"1.23", "876e-2", "5.43e1", "-0.12", "0.25", "-0.23", "-0.27", "0.00", "0.00"},
```
mythrocks, line 1611: Shouldn't we add a test for empty strings?

mythrocks (changes requested):

cpp/tests/io/json_test.cpp
vuule: Having an empty value in these tests does not work well because the reference `to_fixed_point` converts them to zero, old JSON reader reads them as null, and the new reader throws.
Added a separate test for this case.

ttnghia (approved):

mythrocks (approved):

vuule:
@gpucibot merge

PR #12165, opened 2022-11-16
[REVIEW] Pin `dask` and `distributed` for release

galipremsagar:
## Description
This PR pins `dask` and `distributed` to `2022.11.1` for `22.12` release.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12231, opened 2022-11-22
Make pinning >=.

vyasr:
## Description

This PR relaxes the pinning of dask and distributed for 22.12 so that it can be updated in lockstep with other packages in #12165. The current strict pinning causes problems for all downstream RAPIDS packages that have cudf as a pip dependency.

## Checklist

### DIFF
--- a/python/dask_cudf/setup.py
+++ b/python/dask_cudf/setup.py
@@ -8,8 +8,8 @@ from setuptools import find_packages, setup
 cuda_suffix = os.getenv("RAPIDS_PY_WHEEL_CUDA_SUFFIX", default="")
 
 install_requires = [
-    "dask==2022.9.2",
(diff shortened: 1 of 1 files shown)


vyasr:
I'm going to reenable the wheel building process to ensure that wheels build successfully with these changes before we merge.

vyasr:
Wheels build successfully and at least locally the new cudf wheels resolve the dask/distributed dependency resolution issues I ran into during other wheel builds. This PR should be good to merge unless @galipremsagar has a different preference on how we should handle this pinning/unpinning in conjunction with #12165.

galipremsagar (approved):
Thanks @vyasr ! Lets admin merge this now since actual dask pinning is going to happen on this month's end.

shwina (approved):

Issue #12167, opened 2022-11-16
Rework specific cudf::strings and nvtext APIs to use make_strings_children utility


davidwendt:
The fix for https://github.com/rapidsai/cudf/issues/12087 is expected to go into the `cudf::strings::detail::make_strings_children` utility which will throw an error if it detects the output column will be too large.

There are currently 86 `cudf::strings` APIs with 55 of those creating new strings though only 30 of those can possibly create larger strings than its input. 18 of those 30 use the `make_strings_children` utility and will be automatically covered by the fix for #12087. And 2 of the 30 are already have builtin logic to handle any possible overflow. Similarly, `nvtext` has 3 APIs not covered.

This means there are 13 APIs that should either use the `make_strings_children` utility or apply the same fix. This issue is to keep track of the work for fixing these APIs.

- [x] `cudf::strings::from_booleans` https://github.com/rapidsai/cudf/pull/12365
- [x] `cudf::strings::from_timestamps` https://github.com/rapidsai/cudf/pull/12317
- [x] `cudf::strings::from_durations` https://github.com/rapidsai/cudf/pull/12434
- [x] `cudf::strings::from_fixed_point` https://github.com/rapidsai/cudf/pull/12434
- [x] `cudf::strings::from_floats` https://github.com/rapidsai/cudf/pull/12434
- [x] `cudf::strings::from_integers` https://github.com/rapidsai/cudf/pull/12434
- [x] `cudf::strings::integers_to_ipv4` https://github.com/rapidsai/cudf/pull/12401
- [x] `cudf::strings::url_encode` https://github.com/rapidsai/cudf/pull/12385
- [x] `cudf::strings::pad`
- [x] `cudf::strings::zfill`
- [x] `nvtext::generate_character_ngrams` https://github.com/rapidsai/cudf/pull/12480
- [x] `nvtext::ngrams_tokenize` https://github.com/rapidsai/cudf/pull/12540
- [x] `nvtext::detokenize`

PR #12238, opened 2022-11-23
Rework cudf::strings::pad and zfill to use make_strings_children

davidwendt:
## Description
Rework `cudf::strings::pad` and `cudf::strings::zfill` to use the internal `make_strings_children`. This simplifies the code and allows the operation to throw an error if it the output would exceed the size limit of a column.
No function has been added, removed, or changed.

Reference #12167 

## Checklist


cpp/src/strings/padding.cu
```diff
@@ -29,142 +26,135 @@
...
+    : d_column(d_column), width(width), fill_char_size(fill_char_size)
+  {
+  }
 
-  __device__ size_type operator()(size_type idx)
+  __device__ void operator()(size_type idx) const
+  {
+    if (d_column.is_null(idx)) {
+      if (!d_chars) d_offsets[idx] = 0;
+    }
```
ttnghia, line 61: Should this also call `else return`?

cpp/src/strings/padding.cu
```diff
@@ -29,142 +26,135 @@
...
+  auto children = [&] {
+    if (side == side_type::LEFT) {
+      auto fn = pad_fn<side_type::LEFT>{*d_strings, width, fill_char_size, d_fill_char};
+      return make_strings_children(fn, input.size(), stream, mr);
+    } else if (side == side_type::RIGHT) {
+      auto fn = pad_fn<side_type::RIGHT>{*d_strings, width, fill_char_size, d_fill_char};
+      return make_strings_children(fn, input.size(), stream, mr);
+    }
+    auto fn = pad_fn<side_type::BOTH>{*d_strings, width, fill_char_size, d_fill_char};
+    return make_strings_children(fn, input.size(), stream, mr);
```
ttnghia, line 123: Nit: Make `auto const fn = ...`.

davidwendt: This cannot be made `const` since it is modified by `make_strings_children`

cpp/src/strings/padding.cu
davidwendt: Oops. Good catch.

cpp/src/strings/padding.cu
mythrocks: Isn't the functor `fn` passed by copy to `make_strings_children()`? Perhaps I've misunderstood why this can't be const.

ttnghia: Yeah it should be passed by copying.

ttnghia: I just tested that. There is no moving: https://godbolt.org/z/s73nvcjnq.

davidwendt: I think making this `const` incorrectly communicates its usage. Is changing this required for approving the PR?

ttnghia: Nah just a nit. Coding style BTW.

ttnghia (approved):

mythrocks (approved):

davidwendt:
@gpucibot merge

PR #12230, opened 2022-11-22
Rename `cudf::structs::detail::superimpose_parent_nulls` APIs

ttnghia:
There are several overloads of these functions which work differently. They are classified into 2 groups:
 * `superimpose_parent_nulls(null_mask, null_count, column)`: Performs superimposing nulls from somewhere else into the input column,
 * `superimpose_parent_nulls(column_view/table_view)`: Perform superimposing nulls of the input column(s) into their children columns-the input root column(s) are not affected.

That is confusing. They should have different names to reflect their purposes. This PR renames these groups into more meaningful names: `superimpose_nulls` and `push_down_nulls`. No implementation has been changed.

This also supports https://github.com/rapidsai/cudf/issues/12027.

Issue #12027, opened 2022-10-28
[BUG] `cudf::structs::detail::superimpose_parent_nulls` does not purge nulls for the children


ttnghia:
As the title said, the current implementation of `cudf::structs::detail::superimpose_parent_nulls` only set null mask for the children columns. If a child column is a lists/structs/strings/dictionary column, it also needs to be sanitized by `purge_nonempty_nulls`. 

Reference: 
 * https://github.com/rapidsai/cudf/blob/07eb7235338699662cddf242ca8877abdf5ca383/cpp/src/structs/utilities.cpp#L214-L218
 * https://github.com/rapidsai/cudf/blob/07eb7235338699662cddf242ca8877abdf5ca383/cpp/src/structs/utilities.cpp#L261-L262

PR #12239, opened 2022-11-23
Purge non-empty nulls for `superimpose_nulls` and `push_down_nulls`

ttnghia:
The current implementation of `cudf::structs::detail::superimpose_nulls` and `cudf::structs::detail::push_down_nulls` does not do null sanitization. In particular, they only apply the given null mask on the input columns, or push down the parent null mask into the children columns. If there are lists/strings being superimposed by a null bit, they are not sanitized (i.e., they still remain non-empty).

This PR fixes that behavior, sanitizing all non-empty nulls for the output column(s) of these functions. Since there are some changes in the function signatures, this PR is flagged as breaking change.

No new unit tests are needed because the existing tests are already capable to discover the sanitization issue. They were just using `CUDF_TEST_EXPECT_COLUMNS_EQUIVALENT` which allows non-empty nulls to be compared as equal to empty nulls. Changing to `CUDF_TEST_EXPECT_COLUMNS_EQUAL` in these existing tests should be sufficient.

Last but not least, the breaking changes in this PR alter the output results of the `cudf::make_lists_column` function, causing breaking in some other tests such as row bit count tests.

Closes:
 * https://github.com/rapidsai/cudf/issues/12027

Depends on:
 * https://github.com/rapidsai/cudf/pull/12230


karthikeyann:
@ttnghia  How does this change affect the benchmarks?

cpp/src/structs/utilities.cpp
```diff
@@ -207,73 +210,124 @@ flattened_table flatten_nested_columns(table_view const& input,
...
+ * For performance reason, no non-empty null will actually be checked. Instead, this function
+ * performs conservative checking to see if the input column contains any lists or strings column
+ * that is nulllable. As such, it may return false positive answer.
+ *
+ * @param col The input column to check
+ */
+auto may_need_sanitize(column_view const& col)
 {
-  if (input.type().id() == cudf::type_id::EMPTY) {
+  switch (col.type().id()) {
```
davidwendt, line 227: Use type-dispatcher instead of switch statement on types.

ttnghia: Sure.

ttnghia:
> @ttnghia How does this change affect the benchmarks?

As I am aware of, non of the existing benchmarks should be affected. The changes in this PR afffects structs column creation and group by/reduction scan on structs input.

Anyway, I'm going to add a new reduction scan benchmark with structs input to see how it is affected.

ttnghia:
Alright, this is a benchmark showing the affect of this PR on performance of `cudf::detail::scan_inclusive` on structs of 10 members (5 ints + 5 strings):
```
# structs_scan

## [0] Quadro RTX 6000

|  null_probability  |  null_policy  |  data_size  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |         Diff |   %Diff |  Status  |
|--------------------|---------------|-------------|------------|-------------|------------|-------------|--------------|---------|----------|
|         0          |       0       |    10000    | 508.366 us |      53.23% | 546.552 us |      36.61% |    38.186 us |   7.51% |   PASS   |
|        0.1         |       0       |    10000    |   1.676 ms |      13.18% |   1.702 ms |      12.33% |    26.671 us |   1.59% |   PASS   |
|        0.5         |       0       |    10000    |   1.566 ms |      12.52% |   1.582 ms |      12.48% |    16.864 us |   1.08% |   PASS   |
|        0.9         |       0       |    10000    |   1.474 ms |      12.92% |   1.605 ms |      25.59% |   131.019 us |   8.89% |   PASS   |
|         0          |       1       |    10000    | 481.908 us |      22.82% | 514.612 us |      31.67% |    32.703 us |   6.79% |   PASS   |
|        0.1         |       1       |    10000    |   1.716 ms |      11.89% |   1.791 ms |      14.70% |    75.207 us |   4.38% |   PASS   |
|        0.5         |       1       |    10000    |   1.623 ms |      14.17% |   1.726 ms |      16.72% |   103.438 us |   6.37% |   PASS   |
|        0.9         |       1       |    10000    |   1.534 ms |      14.33% |   1.707 ms |      22.41% |   172.439 us |  11.24% |   PASS   |
|         0          |       0       |   100000    | 588.484 us |      45.28% | 601.128 us |      20.75% |    12.644 us |   2.15% |   PASS   |
|        0.1         |       0       |   100000    |   2.003 ms |      14.43% |   1.975 ms |      10.88% |   -27.678 us |  -1.38% |   PASS   |
|        0.5         |       0       |   100000    |   1.786 ms |      14.39% |   1.746 ms |      13.08% |   -39.362 us |  -2.20% |   PASS   |
|        0.9         |       0       |   100000    |   1.710 ms |      14.26% |   1.604 ms |      11.60% |  -105.551 us |  -6.17% |   PASS   |
|         0          |       1       |   100000    | 603.780 us |      22.57% | 590.989 us |      20.04% |   -12.791 us |  -2.12% |   PASS   |
|        0.1         |       1       |   100000    |   1.961 ms |      13.42% |   1.959 ms |      11.54% |    -1.386 us |  -0.07% |   PASS   |
|        0.5         |       1       |   100000    |   1.862 ms |      16.27% |   1.769 ms |      10.07% |   -93.009 us |  -5.00% |   PASS   |
|        0.9         |       1       |   100000    |   1.641 ms |      11.37% |   1.674 ms |      13.77% |    32.446 us |   1.98% |   PASS   |
|         0          |       0       |   1000000   |   2.570 ms |       8.98% |   2.626 ms |      14.57% |    56.095 us |   2.18% |   PASS   |
|        0.1         |       0       |   1000000   |   5.994 ms |       7.55% |   6.031 ms |       6.65% |    36.536 us |   0.61% |   PASS   |
|        0.5         |       0       |   1000000   |   3.997 ms |      10.58% |   3.843 ms |       7.44% |  -154.106 us |  -3.86% |   PASS   |
|        0.9         |       0       |   1000000   |   3.300 ms |       8.47% |   3.314 ms |       7.45% |    13.486 us |   0.41% |   PASS   |
|         0          |       1       |   1000000   |   2.583 ms |       8.51% |   2.665 ms |       8.85% |    81.936 us |   3.17% |   PASS   |
|        0.1         |       1       |   1000000   |   5.068 ms |       6.05% |   5.128 ms |      10.93% |    59.390 us |   1.17% |   PASS   |
|        0.5         |       1       |   1000000   |   3.899 ms |      10.48% |   4.005 ms |      10.95% |   105.328 us |   2.70% |   PASS   |
|        0.9         |       1       |   1000000   |   3.558 ms |      12.17% |   3.417 ms |       8.87% |  -140.072 us |  -3.94% |   PASS   |
|         0          |       0       |  10000000   |  23.600 ms |       2.74% |  24.506 ms |       5.20% |   906.078 us |   3.84% |   FAIL   |
|        0.1         |       0       |  10000000   |  41.350 ms |       3.63% |  43.320 ms |       5.40% |     1.971 ms |   4.77% |   FAIL   |
|        0.5         |       0       |  10000000   |  28.677 ms |       6.83% |  27.385 ms |       1.99% | -1291.848 us |  -4.50% |   FAIL   |
|        0.9         |       0       |  10000000   |  22.663 ms |       3.34% |  22.769 ms |       3.77% |   106.140 us |   0.47% |   PASS   |
|         0          |       1       |  10000000   |  24.244 ms |       4.12% |  24.358 ms |       4.29% |   114.620 us |   0.47% |   PASS   |
|        0.1         |       1       |  10000000   |  37.078 ms |       2.58% |  37.932 ms |       4.46% |   854.250 us |   2.30% |   PASS   |
|        0.5         |       1       |  10000000   |  28.196 ms |       4.66% |  29.114 ms |       5.40% |   917.740 us |   3.25% |   PASS   |
|        0.9         |       1       |  10000000   |  23.306 ms |       4.90% |  24.178 ms |       4.59% |   871.050 us |   3.74% |   PASS   |
```

The comparison shows performance changes in the interval of `[-5%, +5%]`.

cpp/src/structs/utilities.cpp
```diff
@@ -288,59 +370,107 @@ std::tuple<cudf::column_view, std::vector<rmm::device_buffer>> push_down_nulls(
...
-                                     ret_children),
-                         std::move(ret_validity_buffers));
+  return std::pair{column_view(input.type(),
+                               input.size(),
+                               nullptr,
+                               input.null_mask(),
+                               input.null_count(),  // Alternatively, postpone.
+                               input.offset(),
+                               ret_children),
+                   std::move(ret_nullable_data)};
```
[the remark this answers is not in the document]
ttnghia: The purpose of such functions is to fix a `column_view` or `table_view` for intermediate computation, such as sorting, so we only need to generate a new `column_view`/`table_view` with all nulls are empty. If we return new `column`(s) from the input `column_view`(s), we always have to copy (a lot of) data from the input to the output.

The current approach is trying to avoid copying as much as possible. Only in the cases when we want to purge non-empty nulls then we will do copying.

cpp/benchmarks/reduction/scan_structs.cpp
```diff
@@ -0,0 +1,74 @@
...
+    row_count{size},
+    profile);
+  auto [null_mask, null_count] = create_random_null_mask(size, null_probability);
+  auto const input             = cudf::make_structs_column(
+    size, std::move(data_table->release()), null_count, std::move(null_mask));
+
+  auto const agg         = cudf::make_min_aggregation<cudf::scan_aggregation>();
+  auto const null_policy = static_cast<cudf::null_policy>(state.get_int64("null_policy"));
+
+  state.exec(nvbench::exec_tag::sync, [&](nvbench::launch& launch) {
```
davidwendt, line 55: Recommend setting the stream for the nvbench state:
```
```
Also, it would be good if it is possible to provide some throughput metrics.
Example: https://github.com/rapidsai/cudf/blob/ec45c18c0f88acbf8f518a3ac87c3181da93ee2c/cpp/benchmarks/string/like.cpp#L87-L89
(The metrics are not a requirement)

cpp/tests/transform/row_bit_count_test.cu
```diff
@@ -292,7 +294,8 @@ std::pair<std::unique_ptr<cudf::column>, std::unique_ptr<cudf::column>> build_st
   auto size_iter =
     cudf::detail::make_counting_transform_iterator(0, [&strings, &struct_validity](int i) {
       return (sizeof(float) * CHAR_BIT) + 1 + (sizeof(int16_t) * CHAR_BIT) + 1 +
-             (static_cast<cudf::size_type>(strings[i].size()) * CHAR_BIT) +
+             ((struct_validity[i] ? static_cast<cudf::size_type>(strings[i].size()) : 0) *
+              CHAR_BIT) +
              (sizeof(cudf::offset_type) * CHAR_BIT) + 1 + 1;
```
davidwendt, line 299: Perhaps these values should be hardcoded instead of calculated. It seems to be duplicating code logic for `cudf::row_bit_count()`. This can be considered outside the scope of this PR.

cpp/src/reductions/scan/scan_inclusive.cu
```diff
@@ -264,10 +264,21 @@ std::unique_ptr<column> scan_inclusive(
...
+    auto const num_rows   = output->size();
+    auto const null_count = output->null_count();
+    auto content          = output->release();
+
+    // Build new children columns.
+    const auto current_mask = reinterpret_cast<bitmask_type const*>(content.null_mask->data());
+    std::for_each(
+      content.children.begin(), content.children.end(), [current_mask, stream, mr](auto& child) {
+        child = structs::detail::superimpose_nulls(
+          current_mask, cudf::UNKNOWN_NULL_COUNT, std::move(child), stream, mr);
```
davidwendt, line 276:
```suggestion
          current_mask, null_count, std::move(child), stream, mr);
```
The `current_mask == content.null_mask` so is this is not `null_count`?
I'd like to minimize further reliance on `UNKNOWN_NULL_COUNT` if possible.

davidwendt (changes requested):

cpp/benchmarks/reduction/scan_structs.cpp
ttnghia: Thanks. The metrics are not trivial to compute in this case, as the input here is a complex structs column.

cpp/tests/transform/row_bit_count_test.cu
ttnghia: I thought about that when fixing the tests in this file. After a couple of times I wrongly calculated the values, I gave up and decided to just continue using the code :smile:

cpp/src/reductions/scan/scan_inclusive.cu
ttnghia: Oh yes, `null_count` should be used. Thanks.

cpp/tests/transform/row_bit_count_test.cu
davidwendt: More tech debt then.

cpp/tests/transform/row_bit_count_test.cu
```diff
@@ -389,14 +393,34 @@ build_nested_and_expected_column(std::vector<bool> const& struct_validity)
...
+    });
+
   // wrap in a list
   std::vector<int> outer_offsets{0, 1, 1, 3, 6, 7, 8};
   cudf::test::fixed_width_column_wrapper<int> outer_offsets_col(outer_offsets.begin(),
                                                                 outer_offsets.end());
   auto const size = static_cast<cudf::column_view>(outer_offsets_col).size() - 1;
 
-  cudf::test::fixed_width_column_wrapper<cudf::size_type> expected_sizes{
-    276, 32, 520, 572, 212, 212};
```
davidwendt, line 399: Could this be resurrected and not perpetuate the problem of keeping the code logic matching here?

ttnghia: These numbers are computed using the expected behavior of the API. Previously these numbers were computed with non-empty nulls (which was the expected behavior), now they are recomputed based on empty nulls (new expected behavior, which is believed to be correct). In the future, if we have new breaking changes (such as changing the offset type from int32 to int64, although it is very unlikely) then this could be resurrected again.

davidwendt: This appears to be now replaced with a recoding of the expected behavior instead of ground-truth results. Would prefer checking against ground-truth.

davidwendt: I disagree. We should prefer to check against ground-truth instead of recoding the internal logic in the unit test.

davidwendt (approved):

mythrocks:
> If there are lists/strings being superimposed by a null bit, they are not sanitized (i.e., they still remain non-empty)... This PR fixes that behavior, 

This behaviour was by design (at the time), because the sanitization would involve the non-trivial expense of reseating the child columns. I'm glad to see that you've added benchmarks to confirm that the operation is not too expensive.

cpp/include/cudf/detail/structs/utilities.hpp
```diff
@@ -31,6 +31,30 @@ enum class column_nullability {
...
+   * @brief Store the newly generated columns with new null masks (if any).
+   */
+  std::vector<std::unique_ptr<column>> new_columns;
+
+  /**
+   * @brief Take over the content of another instance and append it into the internal data.
+   *
+   * @param other The other instance to take over its content
+   */
+  void emplace_back(temporary_nullable_data&& other);
```
mythrocks, line 55: :+1:

cpp/include/cudf/scalar/scalar.hpp
```diff
@@ -867,8 +867,24 @@ class struct_scalar : public scalar {
  private:
   table _data;
 
-  void init(bool is_valid, rmm::cuda_stream_view stream, rmm::mr::device_memory_resource* mr);
-  void superimpose_nulls(rmm::cuda_stream_view stream, rmm::mr::device_memory_resource* mr);
+  /**
+   * @brief Check if all the input columns constructing this struct scalar have valid size.
+   */
+  void check_size();
```
mythrocks, line 873: This is where we assert a class invariant. It would better named as `assert_valid_size()`.

cpp/src/groupby/sort/group_scan_util.cuh
```diff
@@ -226,8 +226,8 @@ struct group_scan_functor<K,
     // column to them.
     if (values.has_nulls()) {
       for (std::unique_ptr<column>& child : scanned_children) {
-        structs::detail::superimpose_nulls(
-          values.null_mask(), values.null_count(), *child, stream, mr);
+        child = structs::detail::superimpose_nulls(
+          values.null_mask(), values.null_count(), std::move(child), stream, mr);
```
mythrocks, line 230: Ah, this is elegant.

cpp/src/structs/utilities.cpp
```diff
@@ -207,73 +213,149 @@ flattened_table flatten_nested_columns(table_view const& input,
...
+ */
+struct may_need_sanitize_dispatch {
+  template <typename T>
+  bool operator()(column_view const& col) const
+  {
+    return std::any_of(col.child_begin(), col.child_end(), [](auto const& child) {
+      return type_dispatcher(child.type(), may_need_sanitize_dispatch{}, child);
+    });
+  }
+};
```
mythrocks, line 234: I think this might need rephrasing, because I'm not sure we have defined what "sanitize" means in this context. I agree that you and I know it, but it might not be clear to the lay reader.
"Purge" is a clearer term (as you've used in the PR's title), but "sanitize" is more handy. If we defined "sanitize == purge nested contents for null rows" somewhere, that would be useful.

cpp/src/structs/utilities.cpp
```diff
@@ -207,73 +213,149 @@ flattened_table flatten_nested_columns(table_view const& input,
...
+  return col.nullable();
+}
+
+template <>
+bool may_need_sanitize_dispatch::operator()<cudf::list_view>(column_view const& col) const
+{
+  // A lists column may have empty data (i.e., having child column of type EMPTY).
+  // Thus, it needs to be sanitized only if the child column is not empty.
+  return col.nullable() &&
+         col.child(lists_column_view::child_column_index).type().id() != type_id::EMPTY;
```
mythrocks, line 248: Wow, why is this required? Under what scenario is `typeid::EMPTY` possible?

ttnghia: I hit list of empty type in json column creation. Without handling empty type here, json tests will fail.

mythrocks (approved):
LGTM. A couple of minor questions/suggestions.

cpp/src/structs/utilities.cpp
ttnghia: Thanks. I'll rename it to `may_contain_non_empty_nulls` which is clearer.

cpp/benchmarks/reduction/scan_structs.cpp
```diff
@@ -0,0 +1,73 @@
...
+ * limitations under the License.
+ */
+
+#include <benchmarks/common/generate_input.hpp>
+#include <benchmarks/fixture/rmm_pool_raii.hpp>
+
+#include <nvbench/nvbench.cuh>
+
+#include <cudf/column/column_factories.hpp>
+#include <cudf/detail/scan.hpp>
```
PointKernel, line 23:
```suggestion
#include <cudf/column/column_factories.hpp>
#include <cudf/detail/scan.hpp>

#include <nvbench/nvbench.cuh>
```
nit: nvbench header is further than cudf ones

cpp/src/table/row_operators.cu
```diff
@@ -398,13 +398,13 @@ std::shared_ptr<preprocessed_table> preprocessed_table::create(table_view const&
 {
   check_eq_compatibility(t);
 
-  auto [null_pushed_table, null_masks] = structs::detail::push_down_nulls(t, stream);
-  auto struct_offset_removed_table     = remove_struct_child_offsets(null_pushed_table);
-  auto [verticalized_lhs, _, __, ___]  = decompose_structs(struct_offset_removed_table);
+  auto [null_pushed_table, nullable_data] = structs::detail::push_down_nulls(t, stream);
+  auto struct_offset_removed_table        = remove_struct_child_offsets(null_pushed_table);
+  auto [verticalized_lhs, _, __, ___]     = decompose_structs(struct_offset_removed_table);
```
PointKernel, line 403:
```suggestion
  auto verticalized_lhs = std::get<0>(decompose_structs(struct_offset_removed_table));
```
Retrieving the first element only seems cleaner to me

ttnghia:
@gpucibot merge

PR #12242, opened 2022-11-24
Replace column/table test utilities with macros

PointKernel:
w.r.t. #11007

This PR moves `cudf::test::expect_column/table*` utilities to the `detail` namespace and replace their use cases with corresponding macros when possible.

## Checklist

### DIFF
--- a/cpp/tests/utilities_tests/column_utilities_tests.cpp
+++ b/cpp/tests/utilities_tests/column_utilities_tests.cpp
@@ -566,9 +566,7 @@ TEST_F(ColumnUtilitiesStructsTest, Values)
 
   // equivalent, but not equal
   CUDF_TEST_EXPECT_COLUMNS_EQUIVALENT(s_col0, s_col1);
-  EXPECT_EQ(
-    cudf::test::detail::expect_columns_equal(s_col0, s_col1, cudf::test::debug_output_level::QUIET),
-    false);
+  EXPECT_FALSE(cudf::test::detail::expect_columns_equal(s_col0, s_col1, cudf::test::debug_output_level::QUIET));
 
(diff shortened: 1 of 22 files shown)


davidwendt (approved):
LGTM

cpp/tests/utilities_tests/column_utilities_tests.cpp
```diff
@@ -559,8 +566,9 @@ TEST_F(ColumnUtilitiesStructsTest, Values)
 
   // equivalent, but not equal
   CUDF_TEST_EXPECT_COLUMNS_EQUIVALENT(s_col0, s_col1);
-  EXPECT_EQ(cudf::test::expect_columns_equal(s_col0, s_col1, cudf::test::debug_output_level::QUIET),
-            false);
+  EXPECT_EQ(
+    cudf::test::detail::expect_columns_equal(s_col0, s_col1, cudf::test::debug_output_level::QUIET),
+    false);
```
bdice, line 571: Instead of "expect equal to false", just use "expect false". This applies in lots of places in this file.
```suggestion
  EXPECT_FALSE(cudf::test::detail::expect_columns_equal(s_col0, s_col1, cudf::test::debug_output_level::QUIET));
```

bdice (approved):

PointKernel:
@gpucibot merge

karthikeyann:
@PointKernel Thank you for the cleanup. This cpp API usage still happens.
is there a way we can avoid using the cpp API directly instead of macro?
One idea is to leave a not in the API's documentation.

PointKernel:
> Is there a way we can avoid using the cpp API directly instead of macro?

I moved those utilities to `detail` namespace mainly for this reason.

> One idea is to leave a not in the API's documentation.

Good point. I will put up a PR updating docs and comments for this.

Issue #12133, opened 2022-11-11
[BUG] off-by-one errors in `cudf.date_range`


wence-:
**Describe the bug**

[As part of attempting to get to XPASS-zero in the test suite]

If the date range is long enough, and for some frequencies, `cudf.date_range` has a fencepost error in the number of dates it produces.

**Steps/Code to reproduce bug**
```python
import cudf
import pandas as pd
start = "1831-05-08 15:23:21"
end = "1996-11-21 04:05:30"
freq = "110546789L"

cr = cudf.date_range(start=start, end=end, freq=freq)
pr = pd.date_range(start=start, end=end, freq=freq)

assert len(cr) == len(pr) # => False, len(cr) == len(pr) + 1
print(cr[-1])
# => 1996-11-21T14:14:21.984000000
# Which is _after_ the specified end
```

**Expected behavior**

No fencepost error.

Issue #12132, opened 2022-11-11
[BUG] `cudf.concat` with `sort=True` does not match pandas behaviour when concatenating a single object


wence-:
**Describe the bug**

The pandas documentation for [`pandas.concat`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.concat.html) says, of `sort`:

> Sort non-concatenation axis if it is not already aligned when join is ‘outer’. This has no effect when join='inner', which already preserves the order of the non-concatenation axis.

So if we pass `axis=0` then we expect axis-1 to be sorted (and vice versa).

Aside: `sort=True` applies in all cases (irrespective of whether `join="inner"` or `"outer"`) in pandas as far as I can tell (filed https://github.com/pandas-dev/pandas/issues/49646 to check)

However, if we are only concatenating a singleton list of dataframes, this logic is backwards

https://github.com/rapidsai/cudf/blob/d335aa3dcf6a3c55dc7dbd95b94ca8e4b409daee/python/cudf/cudf/core/reshape.py#L278

We want to sort on `1 - axis`.

**Steps/Code to reproduce bug**

```python
import cudf
df = cudf.DataFrame({"b": [1, 2], "a": [3, 4]}, index=["C", "A"])

df2 = cudf.concat([df], sort=True, join="inner")
print(df2)
#    b  a
# A  2  4
# C  1  3

# Expected output
#    a  b
# C  3  1
# A  4  2
```

**Expected behavior**

concat should sort the appropriate axis.

Issue #12243, opened 2022-11-25
[BUG] read_parquet/to_parquet don't handle empty dataframes correctly


wence-:
**Describe the bug**

[This is (possibly) related to https://github.com/pandas-dev/pandas/issues/37897]

An empty dataframe cannot be round-tripped through `to_parquet/read_parquet`:

```python
import cudf
from io import BytesIO

df = cudf.DataFrame(index=cudf.RangeIndex(0, 10))

buf = BytesIO()
df.to_parquet(buf, index=True)
df_read = cudf.read_parquet(buf) # => IndexError
File parquet.pyx:123, in cudf._lib.parquet.read_parquet()

File parquet.pyx:145, in cudf._lib.parquet.read_parquet()

File utils.pyx:52, in cudf._lib.io.utils.make_source_info()

IndexError: Out of bounds on buffer access (axis 0)
```

If I try and read the buffer in pandas I also get an issue

```python
import cudf
import pandas as pd
from io import BytesIO

df = cudf.DataFrame(index=cudf.RangeIndex(0, 10))

buf = BytesIO()
df.to_parquet(buf, index=True)
df_read = pd.read_parquet(buf) # => ArrowInvalid: Could not open Parquet input source '<Buffer>': Parquet file size is 0 bytes
```

If I create the buffer from pandas,  Iget a bad allocation in libcudf
```python
import cudf
import pandas as pd
from io import BytesIO

df = pd.DataFrame(index=pd.RangeIndex(0, 10))

buf = BytesIO()
df.to_parquet(buf, index=True)
df_read = cudf.read_parquet(buf)

File parquet.pyx:123, in cudf._lib.parquet.read_parquet()

File parquet.pyx:162, in cudf._lib.parquet.read_parquet()

MemoryError: std::bad_alloc
```

**Expected behavior**

~cuDF should not segfault on invalid inputs. It should also produce a valid parquet file in this case.~

cuDF should be able to round-trip empty dataframes through a parquet file.

wence-:
In current 23.02 the segmentation faults are gone (good!). But it continues to be the case that none of these examples completes successfully.

It appears that `read_parquet` and `to_parquet` don't play well with empty dataframes.

wence-:
@GregoryKimball I think this probably needs some handling in libcudf, although I _guess_ we could special-case empty dataframes in python, it feels hacky.

wence-:
An update, with 24.02:

Case 1: segfaults
[...]

PR #12244, opened 2022-11-25
Turn on xfail_strict=true

### LINKED ISSUE #7386: Remove xfail condition from now passing tests
Several tests in the cuDF Python codebase were marked `xfail` during initial development but now pass. We should remove the `xfail` condition from these tests.

wence-:
## Description

Systematically work through all currently XPASSing tests and remove
them. Mostly being more careful with the set of parameters that are
marked as XFAIL. In many cases the bug was already fixed!

Furthermore, since programmatic pytest.xfail doesn't really do
anything useful (subsequent code is not run, so we never notice if
something is fixed), add advice in the developer docs, and expunge all
use of programmatic pytest.xfail (enforced via pre-commit). This leads
to slightly worse code in some places in the test suite, but on
balance I think it is worth it since we will now complain loudly when
a previously xfailing feature is fixed.

Close #7386.

## Checklist


python/cudf/cudf/core/reshape.py
```diff
@@ -273,9 +273,11 @@ def concat(objs, axis=0, join="outer", ignore_index=False, sort=None):
                 if isinstance(obj, cudf.Series) and obj.name is None:
                     # If the Series has no name, pandas renames it to 0.
                     data[0] = data.pop(None)
-                result = cudf.DataFrame._from_data(data)
+                result = cudf.DataFrame._from_data(
+                    data, index=obj.index.copy(deep=True)
+                )
 
-        return result.sort_index(axis=axis) if sort else result
+        return result.sort_index(axis=(1 - axis)) if sort else result
```
wence-, line 280: This fixes #12132, I'll spin it out.

wence-: https://github.com/rapidsai/cudf/pull/12247

python/cudf/cudf/core/tools/datetimes.py
```diff
@@ -894,13 +894,6 @@ def date_range(
             # end == start, return exactly 1 timestamp (start)
             periods = 1
 
-    # We compute `end_estim` (the estimated upper bound of the date
-    # range) below, but don't always use it.  We do this to ensure
-    # that the appropriate OverflowError is raised by Pandas in case
-    # of overflow.
-    # FIXME: when `end_estim` is out of bound, but the actual `end` is not,
-    # we shouldn't raise but compute the sequence as is. The trailing overflow
-    # part should get trimmed at the end.
```
wence-, line 903: This comment still applies, so I should reinstate it.

python/cudf/cudf/core/tools/datetimes.py
```diff
@@ -971,18 +964,18 @@ def _offset_to_nanoseconds_lower_bound(offset: DateOffset) -> int:
...
-        + kwds.get("hours", 0) * 3600 * 1e9
-        + kwds.get("minutes", 0) * 60 * 1e9
-        + kwds.get("seconds", 0) * 1e9
-        + kwds.get("milliseconds", 0) * 1e6
-        + kwds.get("microseconds", 0) * 1e3
+        + kwds.get("hours", 0) * 3600 * 10**9
+        + kwds.get("minutes", 0) * 60 * 10**9
+        + kwds.get("seconds", 0) * 10**9
+        + kwds.get("milliseconds", 0) * 10**6
+        + kwds.get("microseconds", 0) * 10**3
```
wence-, line 978: I thought these would fix #12133, but they did not, so I should spin them out I think.

python/cudf/cudf/testing/_utils.py
```diff
@@ -339,10 +339,6 @@ def does_not_raise():
     yield
 
 
-def xfail_param(param, **kwargs):
-    return pytest.param(param, marks=pytest.mark.xfail(**kwargs))
```
wence-, line 343: Never used anywhere.

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -70,35 +70,34 @@ def test_ufunc_index(ufunc):
...
+        # so that we can easily identify these methods. As of this
         # writing, the only missing methods are isnat and heaviside.
         if "module 'cupy' has no attribute" in str(e):
-            pytest.xfail(reason="Operation not supported by cupy")
-        raise
+            with pytest.raises(AttributeError):
+                raise e
+            return
+        else:
+            raise
```
wence-, line 82: This is a pattern I have to introduce in a bunch of places, which I don't _really_ like, but it's better than xfail which is kind of "wrong"?

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -70,35 +70,34 @@ def test_ufunc_index(ufunc):
...
-        # the method so that we can easily identify these methods. As of this
+        # We check if an attribute error is raised so that if we don't
+        # have an explicit dispatch and cupy doesn't have the method
+        # so that we can easily identify these methods. As of this
         # writing, the only missing methods are isnat and heaviside.
         if "module 'cupy' has no attribute" in str(e):
-            pytest.xfail(reason="Operation not supported by cupy")
-        raise
+            with pytest.raises(AttributeError):
+                raise e
```
wence-, line 79: Always true, since we caught attibuteerror, so
```suggestion
```

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -70,35 +70,34 @@ def test_ufunc_index(ufunc):
...
+
+    with context:
         if ufunc.nout > 1:
             for g, e in zip(got, expect):
                 assert_eq(g, e, check_exact=False)
         else:
             assert_eq(got, expect, check_exact=False)
-    except AssertionError as e:
-        # TODO: This branch can be removed when
-        # https://github.com/rapidsai/cudf/issues/10178 is resolved
```
wence-, line 90: It was, but we didn't notice!

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -70,35 +70,34 @@ def test_ufunc_index(ufunc):
...
+    if (
+        fname in {"bitwise_and", "bitwise_or", "bitwise_xor"}
+        and not PANDAS_GE_150
+    ):
+        # https://github.com/pandas-dev/pandas/issues/46769
+        context = pytest.raises(AssertionError)
+    else:
+        context = nullcontext()
+
+    with context:
```
wence-, line 95: We're expecting an `AssertionError` in pandas < 1.5 rather than `xfailing` programmatically; this pattern repeats...

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -269,21 +263,29 @@ def test_binary_ufunc_series_array(ufunc, has_nulls, indexed, reflect):
...
             else:
-                assert_eq(g, e, check_exact=False)
-    else:
-        if has_nulls:
-            expect[mask] = np.nan
-        if reflect:
-            assert (cp.asnumpy(got) == expect).all()
-        else:
-            assert_eq(got, expect, check_exact=False)
+                assert_eq(got, expect, check_exact=False)
```
wence-, line 288: Just an indentation change due to the contextmanager.

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -365,17 +358,57 @@ def test_ufunc_dataframe(ufunc, has_nulls, indexed):
...
-        # writing, the only missing methods are isnat and heaviside.
+        # We check if an attribute error is raised so that if we don't
+        # have an explicit dispatch and cupy doesn't have the method
+        # so that we can easily identify these methods. As of this
+        # writing, the only missing methods are isnat, heaviside, and spacing.
         if "module 'cupy' has no attribute" in str(e):
-            pytest.xfail(reason="Operation not supported by cupy")
-        raise
+            with pytest.raises(AttributeError):
+                raise e
```
wence-, line 367:
```suggestion
```

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -365,17 +358,57 @@ def test_ufunc_dataframe(ufunc, has_nulls, indexed):
...
+            return
+        else:
+            raise
 
     with _hide_ufunc_warnings(ufunc):
         expect = ufunc(*(arg.to_pandas() for arg in pandas_args))
 
-    try:
+    if indexed and fname in {
```
wence-, line 375: It turns out _some_ of the functions work in pandas when indexed...

python/cudf/cudf/tests/test_binops.py
```diff
@@ -910,19 +909,11 @@ def test_binop_bool_uint(func, rhs):
 
 
 @pytest.mark.parametrize(
-    "series_dtype", (np.bool_, np.int8, np.uint8, np.int64, np.uint64)
```
wence-, line 913: Spinning the bool case out into a separate test.

python/cudf/cudf/tests/test_udf_masked_ops.py
```diff
@@ -488,11 +488,12 @@ def func(x):
     # Just a single column -> result will be all NA
     data = cudf.Series([1, 2, cudf.NA])
     if constant is cudf.NA and op in {operator.pow, operator.ipow}:
-        # in pandas, 1**NA == 1. In cudf, 1**NA == 1.
-        with pytest.xfail():
+        # in pandas, 1**NA == 1. In cudf, 1**NA == NA. See
+        # https://github.com/rapidsai/cudf/issues/7478
```
wence-, line 492: Pandas is more correct if one is to treat NA like nan.

python/cudf/cudf/tests/test_udf_masked_ops.py
```diff
@@ -503,16 +504,15 @@ def func(x):
...
     data = cudf.Series([1, 2, cudf.NA])
-    if (
-        constant is not cudf.NA
-        and constant == 1
-        and op in {operator.pow, operator.ipow}
-    ):
-        # in pandas, 1**NA == 1. In cudf, 1**NA == 1.
-        with pytest.xfail():
+    # Using in {1} since bool(NA == 1) raises a TypeError since NA is
+    # neither truthy nor falsy
```
wence-, line 508: and flake8 complains if you write `constant is 1`

python/dask_cudf/dask_cudf/tests/test_groupby.py
```diff
@@ -82,28 +84,29 @@ def test_groupby_cumulative(aggregation, pdf, series):
...
 
     if series:
         gdf_grouped = gdf_grouped.xx
         ddf_grouped = ddf_grouped.xx
 
-    a = getattr(gdf_grouped, aggregation)()
-    b = getattr(ddf_grouped, aggregation)()
-
-    if aggregation == "cumsum" and series:
-        with pytest.xfail(reason="https://github.com/dask/dask/issues/9313"):
```
wence-, line 103: This was only partially fixed, but enough for this test.

python/cudf/cudf/tests/test_udf_masked_ops.py
brandon-b-miller: Many long discussions of this in the past. Short answer is that it's big problem which is also hard to solve because libcudf will never special case single values like this. So we end up having to introspect data in python which has historically been a non-starter. Won't hijack this PR but happy to discuss further.

python/cudf/cudf/tests/test_array_ufunc.py
vyasr: Would it help to introduce something like the `expect_warning_if` contextmanager that I just created, except in this case for exception handling? https://github.com/rapidsai/cudf/blob/branch-23.02/python/cudf/cudf/testing/_utils.py#L393

vyasr: That may also help reduce or usage of the `nullcontext`.

python/cudf/cudf/tests/test_udf_masked_ops.py
wence-: I note there _is_ precedent for post-hoc inspection (with integer division by zero). Arguably checking for NA is easier because we can just inspect `null_count` which doesn't require a scan of the column every time.

python/dask_cudf/dask_cudf/tests/test_groupby.py
wence-: Full fix (dask/dask#9716) is now in.

docs/cudf/source/developer_guide/testing.md
```diff
@@ -140,6 +140,81 @@ def test_odds():
...
+`XPASS` which is considered a failure by the test suite since we use
+the pytest option
+[`xfail_strict=true`](https://docs.pytest.org/en/latest/how-to/skipping.html#strict-parameter).
+Another option is to use the programmatic `pytest.xfail` function to
+fail in the test body to `xfail` the relevant combination of
+parameters. **DO NOT USE THIS OPTION**, unlike the mark-based
+approach, `pytest.xfail` *does not* run the rest of the test body, so
+we will never know if the test starts to pass because the bug is
+fixed. Use of `pytest.xfail` is checked for, and forbidden, via
+a pre-commit hook.
```
vyasr, line 196: One note here, I believe that it should be possible to emulate the expected behavior of `pytest.xfail` by doing something like
```
@pytest.mark.parametrize(x, [1, 2, 3])
def test_x(x):
    if x > 1:
        requests.applymarker(pytest.mark.xfail)
```
I wouldn't encourage this usage, but maybe there are some cases where something like that is needed instead of the strictly static mark approach.

python/cudf/cudf/core/tools/datetimes.py
wence-: Done.

python/cudf/cudf/tests/test_array_ufunc.py
wence-: In the end, I went with the suggestion below of using `request.applymarker` since that feels a bit more ergonomic.

python/cudf/cudf/tests/test_array_ufunc.py
wence-: switched to `requests.applymarker(pytest.mark.xfail(condition=when, reason="foo"))`.

python/cudf/cudf/tests/test_array_ufunc.py
wence-: And gone again.

python/cudf/cudf/tests/test_array_ufunc.py
vyasr: Sounds good. That works much better for failures than warnings.

python/cudf/cudf/tests/pytest.ini
```diff
@@ -3,3 +3,4 @@
 [pytest]
 markers =
     spilling: mark benchmark a good candidate to run with `CUDF_SPILL=ON`
+xfail_strict = true
```
bdice, line 6: Can we do the same for other packages in the repo? I would expect that dask-cudf / custreamz / cudf-kafka / strings_udf won't have as many problems as cudf. This is fine to do in a follow-up PR, I just want testing conventions to be consistent in the repo.

python/cudf/cudf/tests/test_csv.py
```diff
@@ -1366,7 +1367,10 @@ def test_csv_reader_column_names(names):
         assert list(df) == list(names)
 
 
-@pytest.mark.xfail(reason="https://github.com/rapidsai/cudf/issues/10618")
+@pytest.mark.xfail(
+    condition=PANDAS_LT_140,
```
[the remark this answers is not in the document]
bdice: This is a good question. CI currently tests with the latest pandas version that satisfies our dependency pinning. These version-conditional statements are all throughout our tests but I strongly doubt that any non-latest version would actually pass our tests, since we never verify that. I would recommend we discuss this in our next Python meeting. @quasiben ran into this before and we had a conversation about it here: https://github.com/rapidsai/cudf/pull/12150

python/cudf/cudf/tests/test_groupby.py
```diff
@@ -1058,7 +1058,16 @@ def test_groupby_index_type():
     "interpolation", ["linear", "lower", "higher", "nearest", "midpoint"]
 )
 @pytest.mark.parametrize("q", [0.25, 0.4, 0.5, 0.7, 1])
-def test_groupby_quantile(interpolation, q):
+def test_groupby_quantile(request, interpolation, q):
+    request.applymarker(
+        pytest.mark.xfail(
+            condition=(q == 0.5 and interpolation == "nearest"),
+            reason=(
+                "Pandas NaN Rounding will fail nearest " "interpolation at 0.5"
```
bdice, line 1066:
```suggestion
                "Pandas NaN Rounding will fail nearest interpolation at 0.5"
```

bdice (approved):
I have a few small comments, otherwise LGTM.

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -225,19 +221,34 @@ def test_ufunc_series(ufunc, has_nulls, indexed):
...
+            condition=(
+                fname in {"greater", "greater_equal", "logical_and"}
+                and has_nulls
+            ),
+            reason=(
+                "cudf and pandas incompatible casting nans "
+                "to nulls in binops"
+            ),
         )
+    )
```
vyasr, line 251: I think the need to applymarker this frequently is an indication that these tests are probably overparametrized. I'm guilty of this since I wrote most of the problematic tests in this file, but it's also a problem that's pretty prevalent throughout the package. Once this PR is merged we may want to follow up with splitting these up.

python/cudf/cudf/tests/test_binops.py
```diff
@@ -940,7 +931,43 @@ def test_floordiv_zero_float64(series_dtype, divisor_dtype, scalar_divisor):
...
+        ),
+        pytest.param(
+            True,
+            marks=pytest.mark.xfail(
+                reason="https://github.com/rapidsai/cudf/issues/12162"
+            ),
+        ),
+    ],
+)
+def test_floordiv_zero_bool(scalar_divisor):
```
vyasr, line 954: Rather than marking each parameter, why not just mark the test so that you can parametrize with [False, True]?

python/cudf/cudf/tests/test_concat.py
```diff
@@ -1032,9 +1032,6 @@ def test_concat_join_no_overlapping_columns_many_and_empty(
                 ),
                 None,
             ],
-            marks=pytest.mark.xfail(
```
vyasr, line 1035: I guess this has just been xpassing for a while?

python/cudf/cudf/tests/test_csv.py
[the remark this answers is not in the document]
vyasr: Ah this looks like almost exactly the same comment that I had in my review. I agree, let's discuss it in the next meeting and see if we can remove these.

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -285,13 +286,15 @@ def test_append_index(a, b):
         pytest.param(
             {1: [], 2: [], 3: []},
             marks=pytest_xfail(
-                reason="https://github.com/rapidsai/cudf/issues/11080"
+                condition=not PANDAS_GE_150,
+                reason="https://github.com/rapidsai/cudf/issues/11080",
```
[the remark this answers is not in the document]
vyasr: Do you know what changed in pandas for them to pass?

python/cudf/cudf/tests/test_orc.py
```diff
@@ -1715,6 +1700,72 @@ def test_orc_columns_and_index_param(index_obj, index, columns):
...
+        pytest.param(
+            ["a", "b"],
+            None,
+            ["x", "y", "z"],
+            marks=pytest.mark.xfail(
+                reason="https://github.com/rapidsai/cudf/issues/12026"
+            ),
+        ),
+    ],
+)
```
vyasr, line 1764: Can we apply mark.xfail to the whole test to simplify the parametrization to not need `pytest.param`?

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -2169,16 +2169,30 @@ def test_parquet_nullable_boolean(tmpdir, engine):
...
+    pdf.to_parquet(pandas_buffer, index=index)
+    gdf.to_parquet(cudf_buffer, index=index)
+
+    expected = pd.read_parquet(cudf_buffer)
+    actual = cudf.read_parquet(pandas_buffer)
+
+    assert_eq(expected, actual, check_index_type=True)
+
+    expected = pd.read_parquet(pandas_buffer)
+    actual = cudf.read_parquet(cudf_buffer)
```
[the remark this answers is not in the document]
vyasr: Should we delete this to avoid a potential xfail? I assume there's a bug tracking this issue?

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -2207,23 +2221,30 @@ def test_parquet_nullable_boolean(tmpdir, engine):
...
+        ),
+        pytest.param(
+            True,
+            marks=pytest.mark.xfail(
+                reason="https://github.com/rapidsai/cudf/issues/12243",
+            ),
+        ),
+        False,
+    ],
+)
```
vyasr, line 2244: One more case of mark the test to avoid `pytest.param`?

python/cudf/cudf/tests/test_timedelta.py
```diff
@@ -908,7 +903,7 @@ def test_timedelta_index_ops_with_scalars(data, other_scalars, dtype, op):
         actual = other_scalars // gtdi
 
     if op == "floordiv":
-        # Hand-coding pytest.xfail behaviour for certain combinations
+        # Hand-coding xfail behaviour for certain combinations
```
vyasr, line 906: Let's not use xfail since it is easy to conflate pytest xfails
```suggestion
        # Expect exceptions for certain combinations
```

python/cudf/cudf/tests/test_timedelta.py
```diff
@@ -1002,7 +997,7 @@ def test_timedelta_index_ops_with_cudf_scalars(data, cpu_scalar, dtype, op):
         actual = gpu_scalar // gtdi
 
     if op == "floordiv":
-        # Hand-coding pytest.xfail behaviour for certain combinations
+        # Hand-coding xfail behaviour for certain combinations
```
vyasr, line 1000:
```suggestion
        # Expect exceptions for certain combinations
```

python/cudf/cudf/tests/test_udf_masked_ops.py
vyasr: Probably best to follow up on this issue separately from this PR.

vyasr (changes requested):
This is great to see! Just a few more things to do I think.

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -52,9 +52,18 @@ def _hide_ufunc_warnings(ufunc):
...
 @pytest.mark.parametrize("ufunc", _UFUNCS)
-def test_ufunc_index(ufunc):
+def test_ufunc_index(request, ufunc):
     # Note: This test assumes that all ufuncs are unary or binary.
     fname = ufunc.__name__
+    request.applymarker(
+        pytest.mark.xfail(
+            condition=(
+                fname in {"bitwise_and", "bitwise_or", "bitwise_xor"}
+                and not PANDAS_GE_150
```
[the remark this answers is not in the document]
galipremsagar: Historically we have been doing this version-based xfailing when we do an upgrade in cudf, but the question here seems to be how do we phase out the very very old versions we don't test or match against anymore. I'm in favor of stripping away very old version-based xfails, but we should probably discuss the pros and cons of doing so with internal teams too.
My pros:
1. Quality of life improvement and lesser ambiguity in tests.
2. Easy to review

Cons:
1. Someone who tests latest cudf with an old version of pandas(say 1.0) will not be able to get all xfails passing. But the tests that are not being xfailed too will anyway fail because our latest cudf is actually trying to match pandas-1.5 rather than pandas-1.0, so this isn't a huge issue. 

Added this to our weekly agenda to discuss what the team feels about on this issue. @quasiben ran into this recently too.

python/cudf/cudf/tests/pytest.ini
wence-: Opened an issue.

python/cudf/cudf/tests/test_binops.py
wence-: Good point; done.

python/cudf/cudf/tests/test_array_ufunc.py
wence-: Opened an issue.

python/cudf/cudf/tests/test_concat.py
wence-: Yep :)

python/cudf/cudf/tests/test_orc.py
wence-: Only some of these parameters fail, but I should split out the ones that don't...

python/cudf/cudf/tests/test_parquet.py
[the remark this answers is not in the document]
wence-: Yes: https://github.com/rapidsai/cudf/issues/12243; I think it is ok (if we fix the first bug, we'll get an immediate fail here which will prompt us to fix things)

python/cudf/cudf/tests/test_parquet.py
wence-: Note that the `False` case works fine, so I need to split things out... (done).

python/cudf/cudf/tests/test_timedelta.py
wence-: Fixed by going to the applymarker approach

python/cudf/cudf/tests/test_timedelta.py
wence-: Switched to applymarker approach

python/cudf/cudf/tests/test_binops.py
```diff
@@ -1706,22 +1733,25 @@ def test_datetime_dateoffset_binaryop(
...
+                reason="https://github.com/pandas-dev/pandas/issues/49897"
             ),
         ),
         pytest.param(
             {"nanoseconds": 12},
-            marks=pytest_xfail(
-                reason="Pandas gets the wrong answer for nanoseconds"
+            marks=pytest.mark.xfail(
+                condition=not PANDAS_GE_150,
+                reason="Pandas gets the wrong answer for nanoseconds",
```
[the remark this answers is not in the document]
brandon-b-miller: I think that issue might be superseded by https://github.com/pandas-dev/pandas/issues/46877

python/cudf/cudf/tests/test_dataframe.py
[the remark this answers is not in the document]
wence-: No clue sorry, this was "mindless" condition based on what I observe.

python/cudf/cudf/tests/test_timedelta.py
```diff
@@ -907,35 +904,34 @@ def test_timedelta_index_ops_with_scalars(data, other_scalars, dtype, op):
...
+            condition=(
+                op == "floordiv"
+                and (None not in ptdi)
+                and np.nan not in expected
+                and (
+                    expected.astype("float64").astype("int64")
+                    != expected.astype("int64")
+                ).any()
+            ),
+            reason="https://github.com/rapidsai/cudf/issues/12393",
```
wence-, line 931: I thought all the floor division issues were fixed, but it turns out there was special-casing in timedelta.__floordiv__; I'll handle the _fix_ separately.

python/cudf/cudf/tests/test_udf_masked_ops.py
```diff
@@ -738,214 +743,172 @@ def f(x):
...
-    def func(row):
-        return row["str_col"].isalpha()
+    def func(x):
+        return operator(x)
 
-    run_masked_udf_test(func, str_udf_data, check_dtype=False)
+    run_masked_udf_series(func, data, check_dtype=False)
 
 
-@string_udf_test
```
wence-, line 815: All of the test of this file is whitespace change, I grouped the string udf tests into a class to make skipping them easier.

python/dask_cudf/dask_cudf/tests/test_groupby.py
```diff
@@ -82,28 +84,29 @@ def test_groupby_cumulative(aggregation, pdf, series):
...
+    with context:
+        a = getattr(gdf_grouped, aggregation)()
+        b = getattr(ddf_grouped, aggregation)()
+
+        if aggregation == "cumsum" and series:
+            with pytest.raises(AssertionError):
+                # https://github.com/dask/dask/issues/9313
+                dd.assert_eq(a, b)
+        else:
+            dd.assert_eq(a, b)
```
wence-, line 109: I'll go round again and do this properly as part of fixing up the other packages.

wence-:
> This is great to see! Just a few more things to do I think.

Thanks, I think I hit them all

python/cudf/cudf/tests/test_orc.py
vyasr: Oh I think the GH GUI just wasn't quite readable enough for me to notice the ones that didn't fail or I wouldn't have asked. Thanks for splitting though!

python/cudf/cudf/tests/test_parquet.py
vyasr: Thanks again... the nested parens really make this easy to miss.

vyasr (approved):
Great!

vyasr:
@wence- I took care of merging the latest 23.02, which should unblock tests here.

wence-:
Final test fails were real, we fixed a bug in dask that was previously an xfail in dask-cudf.

python/cudf/cudf/tests/test_array_ufunc.py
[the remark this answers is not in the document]
brandon-b-miller: Are our tests considered user facing? IIRC this was invented to get us through some kind of complex CI situation where different test containers ended up pulling different versions of pandas that were allowed within the window. To me, support for a certain range of pandas versions means "your cuDF code will work and produce the same answer, no matter what version of pandas you use within this window" and not much about our unit testing. 

I think viewed from that lens, what this is pointing to is that even though it's very convenient to use pandas to generate "expected" values in most of our tests, it's not actually right to do that because we're aiming for a moving target. Until we stop doing that I suspect we might end up needing flags like this from time to time. I'd rather see us long term take those tests and really take some time to define together with pandas exactly what that API should do, and then write pandas out of the test completely.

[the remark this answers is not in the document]
vyasr: Agreed. #6435 is relevant here too.

vyasr:
@gpucibot merge

quasiben (approved):

Issue #3678, opened 2019-12-30
[FEA] Series and DataFrame duplicated


beckernick:
For API compatibility, we should implement `Series/DataFrame.duplicated`. Exploratory analysis libraries like `pandas-profiling` utilize this.

The `duplicated` method returns a Boolean series indicating whether the value (row, if DataFrame) is a duplicate of another value (row). Behavior around which values should be considered the "original" and which should be considered a "duplicate" is configurable, and pandas [docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.duplicated.html) have some instructive examples.

PR #12246, opened 2022-11-28
[REVIEW] Add `duplicated` support for `Series`, `DataFrame` and `Index`

galipremsagar:
## Description
Resolves: #3678

This PR adds `Series.duplicated`, `DataFrame.duplicated` and `Index.duplicated` by re-using the existing `drop_duplicates` functionality.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/indexed_frame.py
+++ b/python/cudf/cudf/core/indexed_frame.py
@@ -30,12 +30,12 @@ import pandas as pd
 
 import cudf
 import cudf._lib as libcudf
+from cudf._lib.types import size_type_dtype
 from cudf._typing import (
     ColumnLike,
     DataFrameOrSeries,
     Dtype,
     NotImplementedType,
-    size_type_dtype,
 )
 from cudf.api.types import (
     _is_non_decimal_numeric_dtype,
--- a/python/cudf/cudf/_lib/types.pyx
+++ b/python/cudf/cudf/_lib/types.pyx
@@ -17,6 +17,8 @@ from cudf._lib.types cimport (
 
 import cudf
 
+size_type_dtype = np.dtype("int32")
+
 
 class TypeId(IntEnum):
     EMPTY = <underlying_type_t_type_id> libcudf_types.type_id.EMPTY
--- a/python/cudf/cudf/_typing.py
+++ b/python/cudf/cudf/_typing.py
@@ -20,7 +20,6 @@ else:
 # Many of these are from
 # https://github.com/pandas-dev/pandas/blob/master/pandas/_typing.py
 
-size_type_dtype = np.dtype("int32")
 Dtype = Union["ExtensionDtype", str, np.dtype]
 DtypeObj = Union["ExtensionDtype", np.dtype]
 


python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -9815,3 +9815,32 @@ def test_complex_types_from_arrow():
...
+            "style": ["cup", "cup", "cup", "cup", "pack"],
+            "rating": [4, 4, 3.5, 4, 5],
+        },
+    ],
+)
+@pytest.mark.parametrize(
+    "subset", [None, ["brand"], ["rating"], ["style", "rating"]]
+)
+@pytest.mark.parametrize("keep", ["first", "last", False])
+def test_dataframe_duuplicated(data, subset, keep):
```
bdice, line 9839:
```suggestion
def test_dataframe_duplicated(data, subset, keep):
```

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1800,6 +1800,60 @@ def drop_duplicates(
...
+        diff = set(subset) - set(self._data)
+        if len(diff) != 0:
+            raise KeyError(f"columns {diff} do not exist")
+
+        if isinstance(self, cudf.Series):
+            df = self.to_frame(name="None")
+            subset = ["None"]
+        else:
+            df = self.copy(deep=False)
+        df._data["index"] = cudf.core.column.arange(
```
bdice, line 1840: Should this use RangeIndex somehow? I hesitate to see hardcoded `size_type`.

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1800,6 +1800,60 @@ def drop_duplicates(
...
+        keep : {'first', 'last', False}, default 'first'
+            Determines which duplicates (if any) to mark.
+
+            - ``first`` : Mark duplicates as ``True`` except for the first occurrence.
+            - ``last`` : Mark duplicates as ``True`` except for the last occurrence.
+            - False : Mark all duplicates as ``True``.
+        """  # noqa: E501
+        if subset is None:
+            subset = self._column_names
+        elif (
```
bdice, line 1824: We repeat this kind of logic a lot for subsets by name(s). Is there a helper? Should there be?

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1800,6 +1800,60 @@ def drop_duplicates(
...
+        subset : column label or sequence of labels, optional
+            Only consider certain columns for identifying duplicates, by
+            default use all of the columns.
+        keep : {'first', 'last', False}, default 'first'
+            Determines which duplicates (if any) to mark.
+
+            - ``first`` : Mark duplicates as ``True`` except for the first occurrence.
+            - ``last`` : Mark duplicates as ``True`` except for the last occurrence.
+            - False : Mark all duplicates as ``True``.
+        """  # noqa: E501
```
bdice, line 1821: This seems unnecessary.
```suggestion
        """
```

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1800,6 +1800,60 @@ def drop_duplicates(
...
+        ----------
+        subset : column label or sequence of labels, optional
+            Only consider certain columns for identifying duplicates, by
+            default use all of the columns.
+        keep : {'first', 'last', False}, default 'first'
+            Determines which duplicates (if any) to mark.
+
+            - ``first`` : Mark duplicates as ``True`` except for the first occurrence.
+            - ``last`` : Mark duplicates as ``True`` except for the last occurrence.
+            - False : Mark all duplicates as ``True``.
```
bdice, line 1820: Code font formatting is inconsistent in this area across the various docstrings. `False` vs False, 'first' vs `first`, etc.

python/cudf/cudf/core/index.py
```diff
@@ -343,6 +343,10 @@ def astype(self, dtype, copy: bool = True):
     def drop_duplicates(self, keep="first"):
         return self
 
+    @_cudf_nvtx_annotate
+    def duplicated(self, keep="first"):
```
bdice, line 347: Does this inherit a docstring? Or does it need something?

python/cudf/cudf/core/dataframe.py
```diff
@@ -3247,6 +3247,99 @@ def drop_duplicates(
...
+        To find duplicates on specific column(s), use ``subset``.
+
+        >>> df.duplicated(subset=['brand'])
+        0    False
+        1     True
+        2    False
+        3     True
+        4     True
+        dtype: bool
+        """  # noqa: E501
```
bdice, line 3340: I think this noqa can be eliminated with appropriate wrapping.

python/cudf/cudf/core/dataframe.py
```diff
@@ -3247,6 +3247,99 @@ def drop_duplicates(
...
+
+        >>> df.duplicated(keep='last')
+        0     True
+        1    False
+        2    False
+        3    False
+        4    False
+        dtype: bool
+
+        By setting ``keep`` on False, all duplicates are True.
```
bdice, line 3321: Let’s fix this from the pandas example.
```suggestion
        By setting ``keep`` to False, all duplicates are True.
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -3247,6 +3247,99 @@ def drop_duplicates(
...
+        >>> df.duplicated()
+        0    False
+        1     True
+        2    False
+        3    False
+        4    False
+        dtype: bool
+
+        By using 'last', the last occurrence of each set of duplicated values
+        is set on False and all others on True.
```
bdice, line 3311:
```suggestion
        is set to False and all others are True.
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -3247,6 +3247,99 @@ def drop_duplicates(
...
+        >>> df
+             brand style  rating
+        0  Yum Yum   cup     4.0
+        1  Yum Yum   cup     4.0
+        2  Indomie   cup     3.5
+        3  Indomie  pack    15.0
+        4  Indomie  pack     5.0
+
+        By default, for each set of duplicated values, the first occurrence
+        is set on False and all others on True.
```
bdice, line 3300:
```suggestion
        is set to False and all others are True.
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -3247,6 +3247,99 @@ def drop_duplicates(
...
+        See Also
+        --------
+        Index.duplicated : Equivalent method on index.
+        Series.duplicated : Equivalent method on Series.
+        Series.drop_duplicates : Remove duplicate values from Series.
+        DataFrame.drop_duplicates : Remove duplicate values from DataFrame.
+
+        Examples
+        --------
+        Consider dataset containing ramen rating.
```
bdice, line 3283:
```suggestion
        Consider a dataset containing ramen product ratings.
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -3247,6 +3247,99 @@ def drop_duplicates(
...
+            Determines which duplicates (if any) to mark.
+
+            - ``first`` : Mark duplicates as ``True`` except for the first occurrence.
+            - ``last`` : Mark duplicates as ``True`` except for the last occurrence.
+            - False : Mark all duplicates as ``True``.
+
+        Returns
+        -------
+        Series
+            Boolean series for each duplicated rows.
```
bdice, line 3272:
```suggestion
            Boolean series indicating duplicated rows.
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -3247,6 +3247,99 @@ def drop_duplicates(
...
+
+        Considering certain columns is optional.
+
+        Parameters
+        ----------
+        subset : column label or sequence of labels, optional
+            Only consider certain columns for identifying duplicates, by
+            default use all of the columns.
+        keep : {'first', 'last', False}, default 'first'
+            Determines which duplicates (if any) to mark.
```
bdice, line 3263: This sentence is written poorly (in 3 different ways!) in all the upstream docs for DataFrame, Series, and Index. I would prefer this line in all three docstrings:
```suggestion
            Determines the value or values in a set of duplicates to mark.
```

python/cudf/cudf/core/_base_index.py
```diff
@@ -1457,6 +1457,63 @@ def drop_duplicates(
...
+        >>> idx.duplicated(keep='first')
+        array([False, False,  True, False,  True])
+
+        By using 'last', the last occurrence of each set of duplicated values
+        is set on False and all others on True:
+
+        >>> idx.duplicated(keep='last')
+        array([ True, False,  True, False, False])
+
+        By setting keep on ``False``, all duplicates are True:
```
bdice, line 1510:
```suggestion
        By setting keep to ``False``, all duplicates are True:
```

python/cudf/cudf/core/_base_index.py
```diff
@@ -1457,6 +1457,63 @@ def drop_duplicates(
...
+        >>> idx.duplicated()
+        array([False, False,  True, False,  True])
+
+        which is equivalent to
+
+        >>> idx.duplicated(keep='first')
+        array([False, False,  True, False,  True])
+
+        By using 'last', the last occurrence of each set of duplicated values
+        is set on False and all others on True:
```
bdice, line 1505:
```suggestion
        is set to False and all others to True:
```

bdice (commented):
Initial feedback. Some minor docstring edits are `duplicated` between the methods. 😉

python/cudf/cudf/core/dataframe.py
```diff
@@ -3247,6 +3247,99 @@ def drop_duplicates(
...
+
+        >>> df.duplicated(subset=['brand'])
+        0    False
+        1     True
+        2    False
+        3     True
+        4     True
+        dtype: bool
+        """  # noqa: E501
+        return super().duplicated(subset=subset, keep=keep)
```
shwina, line 3341: Looks like this method exists only so we have a distinct docstring. Is there any way to write a single docstring that encapsulates all `Series`, `DataFrame` and `Index` use-cases? If that feels like it would be doing too much, I'm fine with this.

shwina (approved):

python/cudf/cudf/core/indexed_frame.py
galipremsagar: By `RangeIndex` did you intend to do `df._data["index"] = cudf.RangeIndex(0, x)._as_int_index()`? This by the way still returns an `int64` dtype rather than `int32`.

python/cudf/cudf/core/index.py
galipremsagar: This will inherit docstring `BaseIndex.duplicated`

python/cudf/cudf/core/dataframe.py
galipremsagar: The docstrings of `duplicated` should now be `de-duplicated`.

python/cudf/cudf/core/dataframe.py
galipremsagar: Done

python/cudf/cudf/core/dataframe.py
galipremsagar: Done

python/cudf/cudf/core/indexed_frame.py
galipremsagar: Fixed it

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

DaceT:
run benchmarks

python/cudf/cudf/core/dataframe.py
shwina: Sorry, I should have been more clear before. My question was more along the lines of, can we just delete this method altogether? All it does is call `super().method()`, and the only reason it exists is to override the docstring of the base class. Thus, can we write a single, generalized, base class docstring that all subclasses simply inherit? For example,

```python
class Frame:

    def duplicated():
        """
        Indicate duplicated rows of the input Series, Index, or DataFrame
        """
```

But, I realize that this is made difficult by the fact that `Series.duplicated` and `DataFrame.duplicated` take different positional arguments.

I'm not convinced the new approach using `_doc_duplicated_docstring` is better than what we were doing before my review comment (I'd rather live with the small amount of duplication than the added indirection and complexity).

galipremsagar: got it, could you see if the current changes that I force-pushed are satisfactory?

shwina: Yup - looks great to me!

python/cudf/cudf/core/_base_index.py
```diff
@@ -1457,6 +1457,63 @@ def drop_duplicates(
...
+        last occurrence of duplicates can be indicated.
+
+        Parameters
+        ----------
+        keep : {'first', 'last', False}, default 'first'
+            The value or values in a set of duplicates to mark as missing.
+
+            - 'first' : Mark duplicates as ``True`` except for the first
+              occurrence.
+            - 'last' : Mark duplicates as ``True`` except for the last
```
bdice, line 1475:
```suggestion
            - ``'first'`` : Mark duplicates as ``True`` except for the first
              occurrence.
            - ``'last'`` : Mark duplicates as ``True`` except for the last
```

python/cudf/cudf/core/indexed_frame.py
bdice: I would have expected there to be a better way to generate a sequence of indices. We hardcoded `"int32"` below, which I assume is only because of the definition of `using size_type = int32_t;` in C++. If possible we should use some other way to generate this sequence of indices that is already aware of the libcudf index type and doesn't require hardcoding `"int32"`. We're trying to prepare for a future where we could potentially replace `int32` indices with `int64` types with minimal code changes.

python/cudf/cudf/core/series.py
```diff
@@ -2591,6 +2591,86 @@ def transpose(self):
...
+        last occurrence of duplicates can be indicated.
+
+        Parameters
+        ----------
+        keep : {'first', 'last', False}, default 'first'
+            Method to handle dropping duplicates:
+
+            - 'first' : Mark duplicates as ``True`` except for the first
+              occurrence.
+            - 'last' : Mark duplicates as ``True`` except for the last
```
bdice, line 2610:
```suggestion
            - ``'first'`` : Mark duplicates as ``True`` except for the first
              occurrence.
            - ``'last'`` : Mark duplicates as ``True`` except for the last
```

python/cudf/cudf/core/series.py
```diff
@@ -2591,6 +2591,86 @@ def transpose(self):
...
+        Returns
+        -------
+        Series[bool]
+            Series indicating whether each value has occurred in the
+            preceding values.
+
+        See Also
+        --------
+        Index.duplicated : Equivalent method on pandas.Index.
+        DataFrame.duplicated : Equivalent method on pandas.DataFrame.
```
bdice, line 2623:
```suggestion
        Index.duplicated : Equivalent method on cudf.Index.
        DataFrame.duplicated : Equivalent method on cudf.DataFrame.
```

bdice (changes requested):

python/cudf/cudf/core/indexed_frame.py
galipremsagar: Should we have some kind of a mapping in cython to `size_type` <-> `np.dtype`?

bdice: We already have that partially done here: https://github.com/rapidsai/cudf/blob/8e0ee53fea2db1e5d8a7725049756237e7bf9438/python/cudf/cudf/_lib/cpp/types.pxd#L7

Maybe we need a `size_type_dtype` or similar.

bdice: Note: there's a similar definition using Numba types in `strings_udf`, too:
https://github.com/rapidsai/cudf/blob/8e0ee53fea2db1e5d8a7725049756237e7bf9438/python/strings_udf/strings_udf/_typing.py#L17-L18

galipremsagar: Added a `size_type_dtype` to `cudf._typing` to be in line with `strings_udf`. There are soo many places that I see a candidate fit to use this variable, but I'm refraining from polluting this PR with those changes, will open a separate PR.

shwina: I think this should be in `cudf._lib.types` instead. The `_typing` module is more for defining types for static analysis (mypy)

galipremsagar: > I think this should be in `cudf._lib.types` instead. The `_typing` module is more for defining types for static analysis (mypy)

Done 👍

bdice (approved):
Thanks @galipremsagar! Can you file a follow-up issue to use `size_type_dtype` across more of the Python code?

galipremsagar:
@gpucibot merge

PR #12247, opened 2022-11-28
Correct behaviour of sort in `concat` for singleton concatenations

### LINKED ISSUE #12132: [BUG] `cudf.concat` with `sort=True` does not match pandas behaviour when concatenating a single object

wence-:
## Description

The sort argument to `concat` requests a sort of the non-concatenated axis.
Previously this was handled correctly except when passing only a single object
to concatenate.

Closes #12132.

## Checklist

### DIFF
--- a/python/cudf/cudf/tests/test_concat.py
+++ b/python/cudf/cudf/tests/test_concat.py
@@ -1796,7 +1796,7 @@ def test_concat_categorical_ordering():
 
 
 @pytest.fixture(params=["rangeindex", "index"])
-def concat_index(request):
+def singleton_concat_index(request):
     if request.param == "rangeindex":
         return pd.RangeIndex(0, 4)
     else:
@@ -1804,8 +1804,8 @@ def concat_index(request):
 
 
 @pytest.fixture(params=["dataframe", "series"])
-def concat_obj(request, concat_index):
-    if request == "dataframe":
+def singleton_concat_obj(request, singleton_concat_index):
+    if request.param == "dataframe":
         return pd.DataFrame(
             {
                 "b": [1, 2, 3, 4],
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/tests/test_concat.py
```diff
@@ -1793,3 +1793,48 @@ def test_concat_categorical_ordering():
...
+            {
+                "b": [1, 2, 3, 4],
+                "d": [7, 8, 9, 10],
+                "a": [4, 5, 6, 7],
+                "c": [10, 11, 12, 13],
+            },
+            index=concat_index,
+        )
+    else:
+        return pd.Series([4, 5, 5, 6], index=concat_index)
```
shwina, line 1819: This seems like a reasonable use of fixtures to me, except `concat_index` and `concat_obj` seem like very general names for fixtures that are only ever used by a single test. If the intent is to use them more generally, should they appear near the top of this file?

python/cudf/cudf/tests/test_concat.py
```diff
@@ -1793,3 +1793,48 @@ def test_concat_categorical_ordering():
...
+def concat_index(request):
+    if request.param == "rangeindex":
+        return pd.RangeIndex(0, 4)
+    else:
+        return pd.Index(["a", "h", "g", "f"])
+
+
+@pytest.fixture(params=["dataframe", "series"])
+def concat_obj(request, concat_index):
+    if request == "dataframe":
```
shwina, line 1808:
```suggestion
    if request.param == "dataframe":
```

python/cudf/cudf/tests/test_concat.py
wence-: I'm not sure they're more generic (renamed to `singleton_XXX`)

shwina (approved):

shwina:
@gpucibot merge

PR #12096, opened 2022-11-08
Add wheel builds

vyasr:
## Description

This PR enables building wheels. It mostly leverages various build options that have already been added to the repository.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12250, opened 2022-11-29
Fix include line for IO Cython modules

vyasr:
## Description

This PR ensures that numpy include directories are visible to the cuIO Cython TUs. #12096 changed the way that numpy was being found but did not update this line. The old variable is automatically being created by scikit-build's Python code, so as long as skbuild correctly finds the libraries the old variable still works, but we do not want to rely on that since it is not true in all environments.

## Checklist

### DIFF
--- a/python/cudf/cudf/_lib/io/CMakeLists.txt
+++ b/python/cudf/cudf/_lib/io/CMakeLists.txt
@@ -22,5 +22,5 @@ rapids_cython_create_modules(
 
 set(targets_using_numpy io_datasource io_utils)
 foreach(target IN LISTS targets_using_numpy)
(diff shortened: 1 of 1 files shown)


bdice (approved):

vyasr:
> How does this relate to the comment below? Does something else need changed/updated?
> 
> https://github.com/rapidsai/cudf/blob/2a5411ad4624c771747e2cd45e368156b1f91fbe/python/cudf/cudf/_lib/CMakeLists.txt#L82-L84

That comment is referring to the changes I made to `_lib/CMakeLists.txt` in the wheels PR. This PR is updating `_lib/io/CMakeLists.txt`, which was relying on the variables that were previously populated by the old `find_package` call in `_lib/CMakeLists.txt`. I changed one call without changing the other.

PR #12252, opened 2022-11-29
Fix spill to device limit

madsbk:
Fixing bug in `SpillManager.spill_device_memory` where it would spill more than the limit. 

### Checklist

### DIFF
--- a/python/cudf/cudf/core/buffer/spill_manager.py
+++ b/python/cudf/cudf/core/buffer/spill_manager.py
@@ -359,7 +359,7 @@ class SpillManager:
         return spilled
 
     def spill_to_device_limit(self, device_limit: int = None) -> int:
-        """Spill until device limit
+        """Try to spill device memory until device limit
 
         Notice, by default this is a no-op.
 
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/buffer/spill_manager.py
```diff
@@ -380,18 +380,10 @@ def spill_to_device_limit(self, device_limit: int = None) -> int:
...
-                break
-            nbytes = self.spill_device_memory(nbytes=limit - unspilled)
-            if nbytes == 0:
-                break  # No more to spill
-            ret += nbytes
-        return ret
+        unspilled = sum(
+            buf.size for buf in self.buffers() if not buf.is_spilled
+        )
+        return self.spill_device_memory(nbytes=unspilled - limit)
```
vyasr, line 386: IIUC the old code was providing a negative number resulting in the while loop continuing until nbytes = 0. Is that right? If so, how did that not completely tank any benchmarks (or did we just not run benchmarks/measure statistics on the final version that we merged)?

shwina (approved):

galipremsagar (approved):

vyasr (approved):

vyasr:
@gpucibot merge

Issue #12216, opened 2022-11-21
[FEA] Allow creation of nullable Index objects via `to_pandas`


galipremsagar:
**Is your feature request related to a problem? Please describe.**
We currently support converting a `cudf.Series` to a pandas nullable Series by doing `series.to_pandas(nullable=True)`. But pandas recently added support for nullable types in `pd.Index`, so we should add support for `index.to_pandas(nullable=True)` to be possible.

bdice:
The test function `test_index_hasnans` should be updated from #12214 when this feature is added.

galipremsagar:
> The test function `test_index_hasnans` should be updated from #12214 when this feature is added.

Sounds good

mroeschke:
If possible, allowing the user to chose the nullable type e.g. `pd.ArrowDtype(pa.int64)` vs `pd.Int64Dtype` would be great to include

galipremsagar:
> If possible, allowing the user to chose the nullable type e.g. `pd.ArrowDtype(pa.int64)` vs `pd.Int64Dtype` would be great to include

@mroeschke is there a good documentation where we can learn how pandas uses arrow types? or is it configurable by pandas options or a variable?

mroeschke:
As of 1.5.2, user typically need to specify `dtype="<type>[pyarrow]"` or `dtype=pd.ArrowDtype(pa.<type>())` to opt into the pyarrow types, but otherwise a lot of common operations work similarly with these types. Actually, since this issue is specifically referencing converting `Index` objects, my suggestion may be a bit premature. `pandas.Index` can _store_ arrow types but the _indexing operations_ (engine) work on numpy arrays currently

The documentation isn't too great right now. https://pandas.pydata.org/pandas-docs/stable/reference/arrays.html#pyarrow is the only documentation.

PR #12268, opened 2022-11-30
[REVIEW] Add pandas nullable type support in `Index.to_pandas`

galipremsagar:
## Description

Resolves: #12216

This PR adds support for `nullable` parameter to `Index.to_pandas`. When `nullable=True`, the corresponding dtype of the returned index object will try to have a pandas nullable dtype if one exists.

## Checklist


python/cudf/cudf/core/_base_index.py
```diff
@@ -626,10 +626,18 @@ def notna(self):
...
         raise NotImplementedError
 
-    def to_pandas(self):
+    def to_pandas(self, nullable=False):
         """
         Convert to a Pandas Index.
 
+        Parameters
+        ----------
+        nullable : Boolean, Default False
```
shwina, line 635:
```suggestion
        nullable : bool, Default False
```

python/cudf/cudf/core/_base_index.py
```diff
@@ -626,10 +626,18 @@ def notna(self):
...
+            If ``nullable`` is ``True``, the resulting series will be
+            having a corresponding nullable Pandas dtype. If ``nullable``
+            is ``False``, the resulting series will either convert null
+            values to ``np.nan`` or ``None`` depending on the dtype.
```
shwina, line 639:
```suggestion
            If ``nullable`` is ``True``, the resulting index will be
            having a corresponding nullable Pandas dtype. If ``nullable``
            is ``False``, the resulting index will either convert null
            values to ``np.nan`` or ``None`` depending on the dtype.
```

galipremsagar: Yikes, apologies for overlooking the text here.

shwina (approved):

python/cudf/cudf/core/_base_index.py
```diff
@@ -626,10 +626,18 @@ def notna(self):
...
-    def to_pandas(self):
+    def to_pandas(self, nullable=False):
         """
         Convert to a Pandas Index.
 
+        Parameters
+        ----------
+        nullable : bool, Default False
+            If ``nullable`` is ``True``, the resulting index will be
+            having a corresponding nullable Pandas dtype. If ``nullable``
```
vyasr, line 637:
```suggestion
            If ``nullable`` is ``True``, the resulting index will have
            a corresponding nullable Pandas dtype. If ``nullable``
```

vyasr (commented):
Rather than silently producing non-nullable dtypes in cases where no pandas nullable equivalent exists, should we raise an error? If pandas ever introduces additional nullable dtypes the current approach would lead to silently creating indexes of the wrong dtype. A loud failure would probably be easier to debug.

galipremsagar:
> Rather than silently producing non-nullable dtypes in cases where no pandas nullable equivalent exists, should we raise an error if nullable is True? If pandas ever introduces additional nullable dtypes the current approach would lead to silently creating indexes of the wrong dtype. A loud failure would probably be easier to debug.

The tricky part is in case of a dataframe with a mix of dtypes that would end up being pandas nullable & non-nullable dtypes: https://github.com/rapidsai/cudf/blob/branch-23.02/python/cudf/cudf/core/dataframe.py#L5037 i.e., we'd be making the following throw an error:

```python

In [2]: s = cudf.Series([1, 2, 3, 4], dtype='timedelta64[ns]')


In [7]: df = cudf.DataFrame({
   ...:    'time': s,
   ...:     'style': ['cup', 'cup', 'cup', 'pack'],
   ...: })

In [8]: df
Out[8]: 
                       time style
0 0 days 00:00:00.000000001   cup
1 0 days 00:00:00.000000002   cup
2 0 days 00:00:00.000000003   cup
3 0 days 00:00:00.000000004  pack

In [9]: df.to_pandas(nullable=True)
Out[9]: 
                       time style
0 0 days 00:00:00.000000001   cup
1 0 days 00:00:00.000000002   cup
2 0 days 00:00:00.000000003   cup
3 0 days 00:00:00.000000004  pack
```

galipremsagar:
> I lean towards the latter since it seems pretty reasonable for users to use this op to convert a DataFrame where only a subset of columns convert cleanly

Yeah, I lean toward 2 too. The reason being with a warning, it might end up being too noisy.

galipremsagar:
> For all other Index types it's well-defined whether such conversion is valid.

By types, I mean't dtypes..`datetime` & `timedelta` dtypes still don't have corresponding nullable-dtypes. So `Series`/`Index`/`DataFrame`/`MultiIndex` will all have similar behavior when one of these dtypes are present.

vyasr:
Right, I just mean that except for DataFrame/MultiIndex there's no mixing. If a Series/Index has types that don't have a nullable pandas dtype, there's no partial conversion to perform. In any case, I'm still fine with the behavior, just wanted to clarify so that the docstring is accurate.

vyasr (approved):

galipremsagar:
> Right, I just mean that except for DataFrame/MultiIndex there's no mixing. If a Series/Index has types that don't have a nullable pandas dtype, there's no partial conversion to perform. In any case, I'm still fine with the behavior, just wanted to clarify so that the docstring is accurate.

Updated docstring, @vyasr could you take a look at the changes before we merge?

vyasr:
Yup looks good to me, thanks Prem!

vyasr:
@gpucibot merge

PR #12272, opened 2022-11-30
Null element for parsing error in numeric types in JSON, CSV reader

karthikeyann:
## Description
This PR changes behaviour of nested json reader to return null element on parsing error of numeric types.
Breaking behaviour in read_csv, and read_json:

- old behaviour: parsing error returns quiet NAN for that type as result for that row.
- new behaviour: parsing error returns null for that row.

Moved commit 0e45d13  https://github.com/rapidsai/cudf/pull/12022#issuecomment-1314569271 to 23.02
## Checklist

### DIFF
--- a/cpp/tests/io/json_test.cpp
+++ b/cpp/tests/io/json_test.cpp
@@ -914,8 +914,7 @@ TEST_P(JsonReaderParamTest, InvalidFloatingPoint)
 
   const auto col_data = cudf::test::to_host<float>(result.tbl->view().column(0));
   // col_data.first contains the column data
-  for (const auto& elem : col_data.first)
-    ASSERT_TRUE(std::isnan(elem));
+  // ignore all data because it is all nulls.
   // col_data.second contains the bitmasks
   ASSERT_EQ(0u, col_data.second[0]);
 }
@@ -1496,7 +1495,7 @@ TEST_P(JsonReaderParamTest, JsonDtypeParsing)
 
   auto int_col = int_wrapper{
     {0, 0, int_NA, 1, 1, int_NA, int_NA, int_NA, int_NA, 1, 0, int_NA, 1, 0, int_NA, int_NA},
-    cudf::test::iterators::nulls_at(std::vector<int>{8})};
+    make_validity(validity)};
   auto float_col = float_wrapper{{0.0,
                                   0.0,
                                   double_NA,
@@ -1535,7 +1534,7 @@ TEST_P(JsonReaderParamTest, JsonDtypeParsing)
                                 false,
                                 bool_NA,
                                 bool_NA},
-                               cudf::test::iterators::nulls_at(std::vector<int>{8})};
+                               make_validity(validity)};
 
   // Types to test
   const std::vector<data_type> dtypes = {
--- a/cpp/tests/io/json_test.cpp
+++ b/cpp/tests/io/json_test.cpp
@@ -1532,31 +1532,45 @@ TEST_P(JsonReaderParamTest, JsonDtypeParsing)
       0, [=](auto i) -> bool { return static_cast<bool>(validity[i]); });
   };
 
-  constexpr int int_NA       = 0;
-  constexpr double double_NA = std::numeric_limits<double>::quiet_NaN();
-  constexpr bool bool_NA     = false;
+  constexpr int int_ignore{};
+  constexpr double double_ignore{};
+  constexpr bool bool_ignore{};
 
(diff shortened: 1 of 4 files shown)


vyasr:
I got confused reviewing this thinking that I already did, then it turns out that I had! So we decided to remove this breaking change from #12022 so that we could give users more time to adapt by putting it in early in 23.02?

karthikeyann:
That's right. It also gives us more time to fix any downstream issues.

python/cudf/cudf/tests/test_csv.py
```diff
@@ -834,7 +834,7 @@ def test_csv_reader_bools_NA():
     expected = pd.DataFrame(
         {
             "text": ["true", "false", "foo", "bar", "qux"],
-            "int": [1, 0, 1, 0, 0],
+            "int": [1.0, 0.0, 1.0, 0.0, np.nan],
```
wence-, line 837: Should we instead use `[1, 0, 1, 0, None]` with dtype "Int64"? I presume that's truer in spirit to the column that libcudf now constructs?

python/cudf/cudf/tests/test_csv.py
```diff
@@ -834,7 +834,7 @@ def test_csv_reader_bools_NA():
...
         }
     )
     # breaking behaviour is np.nan for qux
```
wence-, line 840: Does this comment still apply?

wence- (commented):
Minor query about the test

cpp/tests/io/json_test.cpp
```diff
@@ -1496,7 +1495,7 @@ TEST_P(JsonReaderParamTest, JsonDtypeParsing)
 
   auto int_col = int_wrapper{
     {0, 0, int_NA, 1, 1, int_NA, int_NA, int_NA, int_NA, 1, 0, int_NA, 1, 0, int_NA, int_NA},
```
vuule, line 1497: int_NA, double_NA and bool_NA don't seem relevant any more, since all those elements will be null in the output table

cpp/tests/io/json_test.cpp
```diff
@@ -914,8 +914,7 @@ TEST_P(JsonReaderParamTest, InvalidFloatingPoint)
 
   const auto col_data = cudf::test::to_host<float>(result.tbl->view().column(0));
   // col_data.first contains the column data
-  for (const auto& elem : col_data.first)
-    ASSERT_TRUE(std::isnan(elem));
+  // ignore all data because it is all nulls.
   // col_data.second contains the bitmasks
   ASSERT_EQ(0u, col_data.second[0]);
```
vuule, line 919: do we need to_host here?
I'd assume that we just need to check `num_rows()` and `null_count()` of `result.tbl->view().column(0)`.

python/cudf/cudf/tests/test_csv.py
karthikeyann: That's a great question. 
Right now, when we encounter nulls in input integer column, we infer the column as float (WWPD). But with null support, we can keep it as "int64". But this will be another breaking behavior. I would prefer to make another PR for that breaking change.

cpp/tests/io/json_test.cpp
karthikeyann: That's right. I have to give with some data there. `int_NA` is just a placeholder data now.

cpp/tests/io/json_test.cpp
karthikeyann: updated.

python/cudf/cudf/tests/test_csv.py
karthikeyann: removed.

python/cudf/cudf/tests/test_csv.py
wence-: Thanks!

wence- (approved):
Python changes look good!

cpp/tests/io/json_test.cpp
karthikeyann: renamed to *_ignore.

vyasr (approved):

cpp/tests/io/json_test.cpp
vuule: I would also be fine with zeroes, but maybe I', missing something in the test.

vuule (approved):

karthikeyann:
@gpucibot merge

Issue #12266, opened 2022-11-30
[BUG] cudf.DataFrame.equals reports False for inputs with NaNs


csadorf:
**Describe the bug**

According to the [documentation](https://docs.rapids.ai/api/cudf/stable/api_docs/api/cudf.DataFrame.equals.html#cudf.DataFrame.equals), the result of `DataFrame.equals()` should be `True` when the two dataframes are equal, even when they contain NaNs:

> NaNs in the same location are considered equal.

However, this does not seem to be the case under certain circumstances.

**Steps/Code to reproduce bug**

The following snippet demonstrates the unexpected behavior:

```python
import numpy as np
import cudf

input_array = np.array([np.nan])

a = cudf.DataFrame(input_array)
b = cudf.DataFrame(input_array)

assert a.equals(b)
```
will fail with an `AssertionError`.

**Expected behavior**

The result of `DataFrame.equals()` should be True in case that values are the same and NaNs are in the same place as described by the method's documentation.

**Environment overview (please complete the following information)**
 - Environment location: Bare-metal
 - Method of cuDF install: conda

**Environment details**
<details><summary>Click here to see environment details</summary><pre>
     
     **git***
     Not inside a git repository
     
     ***OS Information***
     DGX_NAME="DGX Server"
     DGX_PRETTY_NAME="NVIDIA DGX Server"
     DGX_SWBUILD_DATE="2020-03-04"
     DGX_SWBUILD_VERSION="4.4.0"
     DGX_COMMIT_ID="ee09ebc"
     DGX_PLATFORM="DGX Server for DGX-1"
     DGX_SERIAL_NUMBER="QTFCOU6300082"
     DISTRIB_ID=Ubuntu
     DISTRIB_RELEASE=18.04
     DISTRIB_CODENAME=bionic
     DISTRIB_DESCRIPTION="Ubuntu 18.04.4 LTS"
     NAME="Ubuntu"
     VERSION="18.04.4 LTS (Bionic Beaver)"
     ID=ubuntu
     ID_LIKE=debian
     PRETTY_NAME="Ubuntu 18.04.4 LTS"
     VERSION_ID="18.04"
     HOME_URL="https://www.ubuntu.com/"
     SUPPORT_URL="https://help.ubuntu.com/"
     BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
     PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
[...]

PR #12275, opened 2022-11-30
Support equal NaNs in NullEquals binary operation

davidwendt:
## Description
Test change.
Reference #12266 

## Checklist

### DIFF
--- a/cpp/src/binaryop/compiled/operation.cuh
+++ b/cpp/src/binaryop/compiled/operation.cuh
@@ -390,9 +390,11 @@ struct NullEquals {
     TypeLhs x, TypeRhs y, bool lhs_valid, bool rhs_valid, bool& output_valid) -> decltype(x == y)
   {
     output_valid = true;
-    if (!lhs_valid && !rhs_valid) return true;
-    if (lhs_valid && rhs_valid) return x == y;
-    return false;
+    if (!lhs_valid || !rhs_valid) return rhs_valid == lhs_valid;
+    if constexpr (cudf::is_floating_point<TypeLhs>() && cudf::is_floating_point<TypeRhs>()) {
+      if (isnan(x) && isnan(y)) return true;
+    }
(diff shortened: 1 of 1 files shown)


cpp/src/binaryop/compiled/operation.cuh
```diff
@@ -390,9 +390,11 @@ struct NullEquals {
...
   {
     output_valid = true;
-    if (!lhs_valid && !rhs_valid) return true;
-    if (lhs_valid && rhs_valid) return x == y;
-    return false;
+    if (!lhs_valid || !rhs_valid) return rhs_valid == lhs_valid;
+    if constexpr (cudf::is_floating_point<TypeLhs>() && cudf::is_floating_point<TypeRhs>()) {
+      if (isnan(x) && isnan(y)) return true;
+    }
+    return x == y;
```
shwina, line 397: I think doing this unconditionally will break us. We usually want `NaN`s to compare not equal:

```python
In [11]: s = cudf.Series([1.0, 2.0, np.nan], nan_as_null=False)

In [12]: s == s
Out[12]: 
0     True
1     True
2    False
dtype: bool
```

Except when using `.equals()`, when we want NaNs to compare equal:

```
# s.equals(s.copy()) should return [True, True, True]
```

davidwendt: Ok. I wasn't sure if coding this only in `NullsEquals` would you help here.

revans2:
I think this is fine from the spark side. Everywhere I looked that we use NULL_EQUALS it is followed by special case handling to make NaNs equal. 

I would add that we have to do some special case processing for NaNs on all of the comparison operators, not just NULL_EQUALS. But I think those are more likely to cause issues if we don't do this carefully.

davidwendt:
Closing this based on comment from Ashwin that this would break cudf.

PR #12277, opened 2022-12-01
Clean up handling of max_page_size_bytes in Parquet writer

etseidl:
## Description
Reverts a change made in #12182 which exposed some potential issues with calculating page boundaries.  Instead, checks are added to ensure that page sizes will not exceed what can be represented by a signed 32-bit integer.  Also fixes some bounds checking that is no longer correct given changes made to the max page fragment size, and changes the column_index_truncation_length from `size_type` (which was used incorrectly) to `int32_t`.

## Checklist


vuule:
ok to test

cpp/src/io/parquet/page_enc.cu
```diff
@@ -200,6 +219,10 @@ __global__ void __launch_bounds__(block_size)
       len = block_reduce(reduce_storage).Sum(len);
       if (t == 0) { s->frag.fragment_data_size += len; }
       __syncthreads();
+      // page fragment size must fit in a 32-bit signed integer
+      if (s->frag.fragment_data_size > INT32_MAX) {
```
vuule, line 223: IMO `numeric_limits` is preferred.
```suggestion
      if (s->frag.fragment_data_size > std::numeric_limits<int32_t>::max()) {
```

vuule (commented):
Looks good, just one suggestion

cpp/src/io/parquet/page_enc.cu
etseidl: I'll hunt down the rest of them

vuule (approved):

ttnghia:
>  changes the column_index_truncation_length from size_type (which was used incorrectly) to int32_t.

Can you clarify why `size_type` is incorrect, please?

bdice (approved):

bdice:
> > changes the column_index_truncation_length from size_type (which was used incorrectly) to int32_t.
> 
> Can you clarify why `size_type` is incorrect, please?

I'm guessing it's because the use of `int32_t` here is defined by the Parquet format, and it is not dependent on cuDF's internal definition of `using size_type = int32_t;`. Is this correct?

etseidl:
> I'm guessing it's because the use of `int32_t` here is defined by the Parquet format, and it is not dependent on cuDF's internal definition of `using size_type = int32_t;`. Is this correct?

Yep, you got it.  As I understand it, `size_type` is for indexing purposes inside libcudf.  The parameter in question should be an `int32_t` regardless of the definition of `size_type` (which could change in the future for all I know).

vuule:
@gpucibot merge

PR #12279, opened 2022-12-01
Change reductions any/all to return valid values for empty input

### LINKED ISSUE #8757: [BUG] `all` aggregation does not return true when column is empty or contains all nulls
**Describe the bug**
The `all` aggregation should return `true` for an empty column or an entirely nulls column, but instead it returns `false`.  Semantically an `all` reduction indicates there are no `false` values anywhere in the column, complementing the `any` reduction which will only return true if there is at least one `true` value anywhere in the column.  For an empty column or entirely nulls column, there are no valid values to inspect and therefore no `false` value to be found.

Attached is a patch that adds reduction unit tests for these cases.

[all-reduction-empty-or-nulls-patch.txt](https://github.com/rapidsai/cudf/files/6831690/all-reduction-empty-or-nulls-patch.txt)

**Expected behavior**
The `all` aggregation is essentially a search for `false` and should return `true` unless there is at least one `false` value found in the column.  Nulls are skipped, so an empty column or entirely nulls column means there are no `false` values to find and therefore should result in `true`.

davidwendt:
## Description

Changes reduction aggregation-all to return true for empty or all-null column.
Changes reduction aggregation-any to return false for empty or all-null column.
This is consistent with the behavior of [`std::all_of` and `std::any_of`](https://en.cppreference.com/w/cpp/algorithm/all_any_none_of) for empty input.
Additional gtests added for this new behavior.
The overall empty/all-null logic input in `cudf::reduce` was simplified as well. 

Closes #8757 

## Checklist


cpp/src/reductions/reductions.cpp
```diff
@@ -158,25 +158,30 @@ std::unique_ptr<scalar> reduce(
     CUDF_FAIL(
       "Initial value is only supported for SUM, PRODUCT, MIN, MAX, ANY, and ALL aggregation types");
   }
-  // Returns default scalar if input column is non-valid. In terms of nested columns, we need to
-  // handcraft the default scalar with input column.
+
+  // Returns default scalar if input column is empty or all non-valid
```
bdice, line 162:
```suggestion
  // Returns default scalar if input column is empty or all null
```

bdice (changes requested):
I have thought about this before and I’m not sure what semantics are best. This breaks some symmetry. Perhaps we only want to return a boolean instead of a null if an initial value is supplied. Otherwise we are introducing odd behavior with respect to the monoidal structure of reductions. I can try to document this in more detail later or discuss on a call. I have notes somewhere.

bdice:
I wrote a comment about my views on this proposed change:
https://github.com/rapidsai/cudf/issues/8757#issuecomment-1344451607

I will remove my request for changes on this PR, but I would recommend having a bit more discussion on that issue about whether this is the right thing to do.

bdice (approved):
Code looks good. I had some reservations about the change in behavior but am okay with this in its current state after some consideration.

cpp/tests/reductions/reduction_tests.cpp
```diff
@@ -893,8 +893,26 @@ TEST_F(ReductionErrorTest, empty_column)
...
   std::vector<bool> valids(col_size, 0);
 
-  cudf::test::fixed_width_column_wrapper<T> col_empty = construct_null_column(col_data, valids);
-  CUDF_EXPECT_NO_THROW(statement(col_empty));
+  cudf::test::fixed_width_column_wrapper<T> col_nulls = construct_null_column(col_data, valids);
+  CUDF_EXPECT_NO_THROW(statement(col_nulls));
+
+  auto any_agg   = cudf::make_any_aggregation<cudf::reduce_aggregation>();
+  auto all_agg   = cudf::make_all_aggregation<cudf::reduce_aggregation>();
+  auto bool_type = cudf::data_type{cudf::type_id::BOOL8};
```
ttnghia, line 901: Nit: `const`?

cpp/tests/reductions/reduction_tests.cpp
```diff
@@ -893,8 +893,26 @@ TEST_F(ReductionErrorTest, empty_column)
   std::vector<T> col_data(col_size);
   std::vector<bool> valids(col_size, 0);
 
-  cudf::test::fixed_width_column_wrapper<T> col_empty = construct_null_column(col_data, valids);
-  CUDF_EXPECT_NO_THROW(statement(col_empty));
+  cudf::test::fixed_width_column_wrapper<T> col_nulls = construct_null_column(col_data, valids);
```
ttnghia, line 896: Nit: `const`?

ttnghia (approved):

davidwendt:
@gpucibot merge

PR #12284, opened 2022-12-01
Deprecate Frame.sum_of_squares

vyasr:
## Description

This reduction is not supported by pandas and can be trivially accomplished as `(s**2).sum()`.

## Checklist


python/cudf/cudf/core/frame.py
```diff
@@ -2028,6 +2028,11 @@ def sum(
         b    34
         dtype: int64
         """
+        if isinstance(self, cudf.BaseIndex):
+            warnings.warn(
+                "Support for Index.sum is deprecated and will be removed",
```
brandon-b-miller, line 2033: Would it be better to do 
```suggestion
                f"Support for {self.__class__}.sum is deprecated and will be removed",
```
here and throughout?

python/cudf/cudf/core/frame.py
```diff
@@ -2463,6 +2487,10 @@ def any(self, axis=0, skipna=True, level=None, **kwargs):
             **kwargs,
         )
 
+    # TODO: This is not a pandas reduction, it's something that cudf added. Do
+    # we want to continue supporting this? If so, for which types? It seems odd
```
brandon-b-miller, line 2491: this appears to be supported all the way down into libcudf. It could still be useful to have around for something that uses it in a formula, but I don't see a reason to keep it exposed to users especially if we have to invent/support rules for its edge cases and whatnot ourselves. What if we keep it around as a private method?

shwina: groupby-apply would obviate the need for this when it's available. I recall `sum_of_squares` being needed somewhere in the dask_cudf machinery in the past (cc: @rjzamora) but it looks like this is no longer the case. Even if it is, dask-cudf could move to using groupby-apply.

Let's deprecate, and remove in 23.04.

mroeschke:
Noting that I opened an issue in pandas to discuss whether pandas wants to support reduction operations on Index https://github.com/pandas-dev/pandas/issues/50021

As there has been some effort in making Series and Index  behave consistently, API consistency possibly included

vyasr:
Based on discussions with @mroeschke it looks like pandas is going to move towards supporting Index reductions, so I'm going to repurpose this PR as just a deprecation of the `sum_of_squares` reduction.

python/cudf/cudf/core/frame.py
vyasr: Deprecated.

vyasr:
@shwina @brandon-b-miller @mroeschke this PR is now pretty tiny, just a deprecation of `Frame.sum_of_squares` and a small simplification of the kurtosis->kurt alias.

mroeschke (approved):

brandon-b-miller (approved):

vyasr:
@gpucibot merge

PR #12287, opened 2022-12-02
Custom spilling handler 

madsbk:
This PR enables objects to register a spilling handler for a specific spillable buffer. For now, this is used by `RangeIndex` to delete rather than spill its cached data. 

### Motivation 
Normally, a `RangeIndex`, which consists of a `start`, `stop`, `step`, and a `dtype`, doesn’t use any device memory and spilling it should be a no-op. However, if the `RangeIndex` has been materialized, the spill manager might decide to spill the underlying buffer. This can degrade performance and increase memory usage significantly.  

### Checklist


python/cudf/cudf/core/index.py
```diff
@@ -249,9 +252,29 @@ def _num_rows(self):
     @_cudf_nvtx_annotate
     def _values(self):
         if len(self) > 0:
-            return column.arange(
+            ret = column.arange(
                 self._start, self._stop, self._step, dtype=self.dtype
             )
+            manager = get_global_manager()
```
shwina, line 258: I'm concerned about knowledge of the spilling manager leaking into other cuDF types like `RangeIndex`. Is there any way we can do this totally outside of `RangeIndex`?

madsbk:
It might be possible to do in a decorator that could replace
cached_property. Would that be sufficient?

shwina:
> It might be possible to do in a decorator that could replace cached_property. Would that be sufficient?

Yes, I think it would be a significant improvement

python/cudf/cudf/core/index.py
madsbk: Introducing a `cached_property_delete_column_when_spilled` instread

madsbk:
> > It might be possible to do in a decorator that could replace cached_property. Would that be sufficient?
> 
> Yes, I think it would be a significant improvement

I have created a decorator instead, do you think this is acceptable @shwina?

shwina:
Thanks! I do think this is an improvement. Some questions/thoughts:

* `cached_property_delete_column_when_spilled` needs the property to return a `Column`, but what about cached properties that possibly return buffers or arrays? Would each of those need a custom `cached_property` specialization that does things slightly differently with the return value?

* Can we aim to make the changes to `RangeIndex` (and potentially other classes) even more minimal? For example:
  - Instead of using `cached_property_delete_when_spilled`, can we universally use a custom cached property in cuDF that knows about spilling? That is, everywhere that we currently use `functools.cached_property`, can we use `cudf.utils.cached_property`? Or are there situations we don't want to delete cached device objects when spilled?
  - Instead of using a decorator, can we monkeypatch methods like `RangeIndex._values` when spilling is enabled? This allows `RangeIndex` to be truly untouched.

  Admittedly I don't think either of those are fantastic suggestions; curious what you think?

madsbk:
@shwina,  I have generalized and renamed the decorator. Now, it can be used everywhere with no ill effect.

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -134,3 +140,83 @@ def get_spill_lock() -> Union[SpillLock, None]:
...
+        cache_hit = self.attrname in instance.__dict__
+        ret = super().__get__(instance, owner)
+        if cache_hit or not isinstance(ret, cudf.core.column.ColumnBase):
+            return ret
+
+        manager = get_global_manager()
+        if manager is None:
+            return ret
+
+        buf = ret.base_data
```
shwina, line 210: What about the `mask` buffer of the `Column`, or - in the case of nested columns - its children?

madsbk: Good point.  

I have simplified the code a bit. Instead of supporting any type, we now check that the instance is a `RangeIndex`. I haven't found any other use of `cached_property` where the cache can be the sole owner of a buffer, so I think we should limit the scope to `RangeIndex` for now.

madsbk:
rerun tests

madsbk:
rerun tests

python/cudf/cudf/core/buffer/utils.py
vyasr: Seems related to the question I asked above. I'm OK with not overengineering to support anything else, although it would be good to ensure that if we do need to support other types later it only requires adding new code and not restructuring the existing code (hence my comment above).

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -134,3 +140,92 @@ def get_spill_lock() -> Union[SpillLock, None]:
...
+    # memory by clearing the cache. We know of three references:
+    # `instance.__dict__`, `cached`, and `sys.getrefcount`.
+    if sys.getrefcount(cached) > 3:
+        return None
+
+    instance.__dict__.pop(attrname, None)  # Clear cache atomically
+    return nbytes
+
+
+class cached_property(functools.cached_property):
```
vyasr, line 188: I would prefer to use a different name.

vyasr: Maybe `cached_property_no_spill` or something.

python/cudf/cudf/core/buffer/spill_manager.py
```diff
@@ -347,13 +351,22 @@ def spill_device_memory(self, nbytes: int) -> int:
...
                 try:
                     if not buf.is_spilled and buf.spillable:
+                        # Check if `buf` has a registered spill handler
+                        handler = self._spill_handlers.get(buf, None)
+                        if handler is not None:
+                            func, args, kwargs = handler
+                            s = func(*args, **kwargs)
+                            if s is not None:
+                                spilled += s
+                                self._spill_handlers.pop(buf, None)
```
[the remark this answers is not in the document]
vyasr: OK I see why it's removed, I guess it's being added every time the cached property is accessed.

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -134,3 +140,92 @@ def get_spill_lock() -> Union[SpillLock, None]:
...
+        cache_hit = self.attrname in instance.__dict__
+        ret = super().__get__(instance, owner)
+        # Make sure to only register a handler when the cache has changed.
+        # Also, for now, we only handle `RangeIndex` instances that returns
+        # a numerical column with no mask.
+        if (
+            cache_hit
+            or not isinstance(instance, cudf.RangeIndex)
+            or not isinstance(ret, cudf.core.column.NumericalColumn)
+            or ret.nullable
```
vyasr, line 211: Why is the nullable check needed?

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -134,3 +140,92 @@ def get_spill_lock() -> Union[SpillLock, None]:
...
+        ):
+            return ret
+
+        manager = get_global_manager()
+        if manager is None:
```
vyasr, line 216: Under what circumstances could this be None? Wouldn't the call to `ret = super().__get__(instance, owner)` trigger the creation of a `SpillableBuffer` (assuming this decorator is only applied to functions that return something that wraps a `SpillableBuffer`) which in turn would always lead to the creation of a global `SpillManager` if one isn't already initialized.

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -134,3 +140,92 @@ def get_spill_lock() -> Union[SpillLock, None]:
...
+        return 0
+
+    cached = instance.__dict__.get(attrname, None)
+    if cached is None:
+        return None  # The cached has been cleared
+
+    # If `cached` is known outside of the cache, we cannot free any
+    # memory by clearing the cache. We know of three references:
+    # `instance.__dict__`, `cached`, and `sys.getrefcount`.
+    if sys.getrefcount(cached) > 3:
```
vyasr, line 181: I find this line disconcerting, but I'm having trouble articulating exactly why so I'm just going to try to explain my thought process and see if we can resolve my confusion.

At the point where we call `buf.spill` (and now the spill handler) we already know that the buffer is spillable. A buffer is spillable according to our existing logic if it hasn't been handed out to an external consumer and if there are no spill locks around it. If we get to to this point and the number of references is greater than 3, doesn't that imply that we should just spill? It seems like at that point you're in the exact same scenario that you would be for any other SpillableBuffer where a future use could try to unspill it.

vyasr (changes requested):

python/cudf/cudf/core/buffer/utils.py
madsbk: > If we get to to this point and the number of references is greater than 3, doesn't that imply that we should just spill? It seems like at that point you're in the exact same scenario that you would be for any other SpillableBuffer where a future use could try to unspill it.

Correct, this is also what is done. By returning `None`, we are telling the spill manager to spill the buffer as usual. 
We could move this responsibility to the handler always. It would then be up to the handler to spill the buffer?

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -134,3 +140,92 @@ def get_spill_lock() -> Union[SpillLock, None]:
...
+        return None
+
+    instance.__dict__.pop(attrname, None)  # Clear cache atomically
```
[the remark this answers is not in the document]
madsbk: A local lock here would not help since changes to the class instance could come from anywhere. The fact that `dict` access is thread safe is relied on throughout Python including in `functools.cached_property`.

python/cudf/cudf/core/buffer/utils.py
madsbk: It is just to avoid having to handle a mask

python/cudf/cudf/core/buffer/utils.py
madsbk: This is to handle the case where spilling is disabled. Notice, this decorator is also used when spilling is disabled.

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -134,3 +140,92 @@ def get_spill_lock() -> Union[SpillLock, None]:
...
+
+    def __get__(self, instance: T, owner=None):
+        cache_hit = self.attrname in instance.__dict__
+        ret = super().__get__(instance, owner)
+        # Make sure to only register a handler when the cache has changed.
+        # Also, for now, we only handle `RangeIndex` instances that returns
+        # a numerical column with no mask.
+        if (
+            cache_hit
+            or not isinstance(instance, cudf.RangeIndex)
```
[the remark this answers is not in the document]
madsbk: This is an attempt to address the concerns @shwina points out here and to avoid monkey patching.  

In the original implementation, the decorator was specific to `RangeIndex`, see https://github.com/rapidsai/cudf/pull/12287/commits/8df1adaf57af7285632e4dac075f9d41d0da808f.

After working through this PR a couple of times now, I think having a new decorator specifically for cached Columns is the preferable approach. Would that be acceptable @shwina? 

Either way, I think we should avoid monkey patching.

shwina:
Retargeted to 23.04

vyasr:
@madsbk is this still something that we want? If so, at this point should it still wait until after you've done the final merging of COW/spilling to avoid conflicting code?

madsbk:
> @madsbk is this still something that we want? If so, at this point should it still wait until after you've done the final merging of COW/spilling to avoid conflicting code?

Yes and yes :)
The materialization of `RangeIndex` is something we want to avoid but let's wait until COW and spilling has been unified.

vyasr:
Note, the above unification is done now so this work is unblocked if desired.

vyasr:
While I think we may still want something like this feature eventually, as pylibcudf and cudf-polars mature and as we develop more spilling functionality in rapidsmpf, I think we are going to have to reenvision our spilling strategies in general and this approach is probably not going to be the right one since it is so specific to the spilling logic used in cudf. I think we need more general solutions in the long run.

Issue #10363, opened 2022-02-25
[FEA] Remove `FutureWarning`s from Python tests


bdice:
This is a meta-issue as part of #9999. Here, I catalog the current list of test files that raise a `FutureWarning`.

The use of a `FutureWarning` indicates a deprecated API, mostly from cudf or pandas. We should explicitly catch or filter out these warnings, depending on the circumstances. Eventually we will be able to run `pytest -W error::FutureWarning` to catch any deprecated APIs being called by our tests, which may conceal a regression in the user API.

- [x] `test_csv.py` #10362
- [x] `test_cuda_apply.py` #10402
- [ ] `test_dataframe.py`
- [x] `test_groupby.py`
- [x] `test_index.py`
- [x] `test_indexing.py` #12305
- [x] `test_joining.py` #12304
- [x] `test_multiindex.py` #12310
- [x] `test_numerical.py` #10402
- [x] `test_orc.py`
- [x] `test_parquet.py` (partial) #10416
- [x] `test_pickling.py` #10402
- [x] `test_rolling.py` #10405
- [x] `test_stats.py` #12293
- [x] `test_timedelta.py` #10418
- [x] `test_unaops.py` #10402
- [ ] Update tests to fail on `FutureWarning`: `pytest -W error::FutureWarning`

bdice:
I traced down some unexpected behavior and it turns out to be an error in the Python docs. This comment explains what I found, for posterity/reference. cc: @galipremsagar

I tried to specify a warning filter `pytest -W error::FutureWarning:cudf.*` on the command line. This follows the guidelines for [how to specify warning filters](https://docs.python.org/3/library/warnings.html#describing-warning-filters) as `action:message:category:module:line`.

I expected that this would raise an error for any `FutureWarning` raised by the `cudf` module or its subpackages, by interpreting the `cudf.*` module string as a regular expression. However, we can't do this with `-W`, because the input strings for `message` and `module` are escaped and not interpreted as regular expressions when defined with `-W` or `PYTHONWARNINGS`. The docs indicate that the module string is always treated as a regular expression, but that is only true for `warnings.filterwarnings` and not the command line option `-W` or environment variable `PYTHONWARNINGS`. This problem in the docs was reported in https://bugs.python.org/issue42272 and fixed in https://github.com/python/cpython/pull/23172.

The workaround for this is to define this in our pytest configuration, `setup.cfg`, with the following:

```ini
[tool:pytest]
[...]

Issue #9999, opened 2022-01-07
[Story] Reorganize and improve cudf classic tests


vyasr:
**Note for developers**
This is a meta-issue aiming to categorize a wide range of issues. Developers who want to tackle a specific item from the checklist below should create a new issue for just that item, self-assign that issue, and then link it to the checklist above.

**Is your feature request related to a problem? Please describe.**
There are currently a number of different problems that make it difficult to find, add, or run tests.
- Tests are partially and inconsistently organized by functionality, data types, and parametrizations, so it's not clear which file tests of a specific function might be in. The partial organization by dtype is particularly confusing because it means that in some files we test a single function across many dtypes, whereas in other files we test many functions for a single dtype.
- Test files are too large and contain too many tests.
- There are currently many tests that raise warnings as well as many xfailed tests that actually xpass.
- Tests are currently slow to run because we rely on excessive parametrization and we test many private APIs. Additionally, cuIO tests are especially slow because the corresponding libcudf APIs are difficult to test, so a greater burden is placed on the Python APIs to capture a wider range of issues.

**Describe the solution you'd like**
These are tasks that we propose to undertake to address the various issues discussed above:

- [x] See #7386. Change pytest to run fail on xpasses by specifying the `strict` parameter. [Here are the relevant pytest docs](https://docs.pytest.org/en/latest/how-to/skipping.html#strict-parameter). 
- [x] Make pytest fail when it encounters warnings. We can accomplish this by setting in the appropriate config file (probably setup.cfg). (Done in #12468)
  - [x] #10363
  - [x] Remove other warnings from Python tests (this was done across a large number of PRs)
- [ ] Reorganize test files to match the groupings in our API documentation. See also #4730
- [ ] Reorganize tests into classes that reflect the actual class hierarchies. For example, tests that should be performed for Series and DataFrame objects can go into a `test_indexed_frame.py` file.
- [ ] Reduce excessive parametrization. Many individual tests are actually run hundreds or even thousands of times because we construct a matrix of parameters. In many cases those parametrizations aren't actually testing anything useful, for instance when a test is parametrized by input size. In cases where we're worried about how the underlying libcudf algorithm behaves when block sizes are exceeded, for instance, we should push those tests down to C++. Python tests should favor fewer parameters and in some cases more specialized tests to handle specific edge cases.
[...]

PR #12293, opened 2022-12-02
Fix warnings in test_stats.py

vyasr:
## Description

## Checklist

### DIFF
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -5861,7 +5861,49 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
                     for col in source._data.names
                 ]
             except AttributeError:
-                raise TypeError(f"Not all column dtypes support op {op}")
+                numeric_ops = (
+                    "mean",
+                    "min",
+                    "max",
+                    "sum",
+                    "product",
+                    "prod",
+                    "std",
+                    "var",
+                    "kurtosis",
+                    "kurt",
+                    "skew",
+                )
+
+                if numeric_only is None and op in numeric_ops:
+                    warnings.warn(
+                        "The default value of numeric_only in DataFrame.kurt "
+                        "is deprecated. In a future version, it will default "
+                        "to False. In addition, specifying "
+                        "'numeric_only=None' is deprecated. Select only valid "
+                        "columns or specify the value of numeric_only to "
+                        "silence this warning.",
+                        FutureWarning,
+                    )
+                    numeric_cols = (
+                        name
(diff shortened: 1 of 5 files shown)


python/cudf/cudf/core/dataframe.py
```diff
@@ -5861,7 +5861,49 @@ def _reduce(
...
+                    "std",
+                    "var",
+                    "kurtosis",
+                    "kurt",
+                    "skew",
+                )
+
+                if numeric_only is None and op in numeric_ops:
+                    warnings.warn(
+                        "The default value of numeric_only in DataFrame.kurt "
```
shwina, line 5880: Is this warning just for `DataFrame.kurt`?

vyasr: Nope, thanks. Fixed.

python/cudf/cudf/tests/test_stats.py
```diff
@@ -442,7 +449,34 @@ def test_corr1d(data1, data2, method):
...
+        is_identical and not is_singular and method == "spearman"
+    )
+    if method == "spearman":
+        import scipy
+
+        expected_warning = scipy.stats._warnings_errors.ConstantInputWarning
+    elif method == "pearson":
+        expected_warning = RuntimeWarning
+
+    with expect_warning_if(cond, expected_warning):
```
shwina, line 478: I'm not so sure this is a major improvement over just:

```python
if cond:
    with pytest.warns(expected_warning):
```

My worry with helpers like this is they don't end up being used everywhere that they are meant to - because developers don't typically reach for a helper for something like this.

If an existing call to `warns()` needs to change to happen conditionally, a developer is _much_ more likely to just move it inside a new `if` and be done with it.

As Nick would say though, this is a strong opinion - weakly held. If you feel strongly otherwise, please feel free to keep this.

vyasr: The downside of not having this is that you have to duplicate the code executed within the context:
```
if cond:
    with pytest.warns():
        foo()
else:
    foo()
```
Ideally `foo` would always be a simple one-liner, but that is not necessarily the case either (although maybe we should require it to be, it's certainly better to scope exception/warning handling as tightly as possible). Would you prefer the code duplication?

shwina: Ah, that's a good point. I'm good with keeping it then - although yes, things like `raises()` and `warns()` should be tightly scoped anyway

shwina (approved):

vyasr:
rerun tests

vyasr:
@gpucibot merge

Issue #2230, opened 2019-07-12
[BUG] cudf.DataFrame.merge should support implicit type conversions on join columns


randerzander:
When merging on columns with different types in Pandas, it implicitly converts to matching types and runs the merge:
```
>>> df = pd.DataFrame()
>>> df['id'] = [0, 1, 2]
>>> df['val'] = [9, 9, 9]

>>> df_2 = pd.DataFrame()
>>> df_2['id'] = [0, 1, 2]

>>> df.dtypes
id     int64
val    int64
dtype: object

>>> df_2.dtypes
id    int64
dtype: object
>>> df_2['id'] = df_2['id'].astype('float64')

>>> df.merge(df_2, on=['id'])
   id  val
0   0    9
1   1    9
2   2    9
```

cudf doesn't, and fails with type mismatch:
```
>>> import cudf
>>> df = cudf.from_pandas(df)
>>> df_2 = cudf.from_pandas(df_2)
>>> df.dtypes
id     int64
val    int64
dtype: object
>>> df_2.dtypes
id    float64
dtype: object
>>> df.merge(df_2, on=['id']).to_pandas()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/conda/envs/rapids/lib/python3.7/site-packages/cudf-0.9.0a0+1094.gdcdf3596a.dirty-py3.7-linux-x86_64.egg/cudf/dataframe/dataframe.py", line 1934, in merge
    lhs._cols, rhs._cols, left_on, right_on, how, method
  File "cudf/bindings/join.pyx", line 26, in cudf.bindings.join.join
  File "cudf/bindings/join.pyx", line 124, in cudf.bindings.join.join
  File "cudf/bindings/cudf_cpp.pyx", line 487, in cudf.bindings.cudf_cpp.check_gdf_error
cudf.bindings.GDFError.GDFError: b'GDF_DTYPE_MISMATCH'
```

kkraus14:
@harrism is this something we can handle down in libcudf or would you want us to typecast in Python before passing to libcudf?

harrism:
@jrhemstad are you more familiar with join logic in order to answer?

One thing we would need from you @kkraus14 is the expected casting rules, so we can discuss differences from expected C++ casting and make sure we do the right thing...

jrhemstad:
> @harrism is this something we can handle down in libcudf or would you want us to typecast in Python before passing to libcudf?

[...]

PR #12304, opened 2022-12-05
Fix warnings in test_joining.py

vyasr:
## Description

## Checklist

### DIFF
--- a/python/cudf/cudf/core/join/join.py
+++ b/python/cudf/cudf/core/join/join.py
@@ -419,8 +419,8 @@ class Merge:
         ):
             warnings.warn(
                 "merging between different levels is deprecated and will be "
-                "removed in a future version. (2 levels on the left, 1 on the "
-                "right)",
+                f"removed in a future version. ({lhs.nlevels} levels on the "
+                f"left, {rhs.nlevels} on the right)",
                 FutureWarning,
             )
 
--- a/python/cudf/cudf/core/join/join.py
+++ b/python/cudf/cudf/core/join/join.py
@@ -415,12 +415,12 @@ class Merge:
         if (
             isinstance(lhs, cudf.DataFrame)
             and isinstance(rhs, cudf.DataFrame)
-            and lhs.columns.nlevels != rhs.columns.nlevels
+            and lhs._data.nlevels != rhs._data.nlevels
         ):
             warnings.warn(
                 "merging between different levels is deprecated and will be "
-                f"removed in a future version. ({lhs.nlevels} levels on the "
-                f"left, {rhs.nlevels} on the right)",
+                f"removed in a future version. ({lhs._data.nlevels} levels on "
+                f"the left, {rhs._data.nlevels} on the right)",
                 FutureWarning,
             )
 
--- a/python/cudf/cudf/core/join/join.py
+++ b/python/cudf/cudf/core/join/join.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2020-2022, NVIDIA CORPORATION.
 from __future__ import annotations
 
+import warnings
 from typing import Any, ClassVar, List, Optional
 
 import cudf
@@ -410,6 +411,18 @@ class Merge:
                         "there are overlapping columns but "
                         "lsuffix and rsuffix are not defined"
                     )
+
+        if (
+            isinstance(lhs, cudf.DataFrame)
+            and isinstance(rhs, cudf.DataFrame)
+            and lhs.columns.nlevels != rhs.columns.nlevels
+        ):
+            warnings.warn(
+                "merging between different levels is deprecated and will be "
+                "removed in a future version. (2 levels on the left, 1 on the "
+                "right)",
+                FutureWarning,
+            )
 
 
 class MergeSemi(Merge):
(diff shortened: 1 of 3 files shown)


python/cudf/cudf/core/join/join.py
```diff
@@ -411,6 +412,18 @@ def _validate_merge_params(
...
                     )
 
+        if (
+            isinstance(lhs, cudf.DataFrame)
+            and isinstance(rhs, cudf.DataFrame)
+            and lhs.columns.nlevels != rhs.columns.nlevels
+        ):
+            warnings.warn(
+                "merging between different levels is deprecated and will be "
+                "removed in a future version. (2 levels on the left, 1 on the "
```
brandon-b-miller, line 422:
```suggestion
                f"removed in a future version. ({lhs.nlevels} levels on the left, {rhs.nlevels} on the "
```

brandon-b-miller (approved):

python/cudf/cudf/tests/test_joining.py
```diff
@@ -1123,11 +1124,21 @@ def test_typecast_on_join_overflow_unsafe(dtypes):
...
+        (dtype_l.kind == "f" and dtype_r.kind in {"i", "u"})
+        or (dtype_l.kind in {"i", "u"} and dtype_r.kind == "f"),
         UserWarning,
-        match=(f"Can't safely cast column" f" from {dtype_r} to {dtype_l}"),
     ):
-        merged = lhs.merge(rhs, on="a", how="left")  # noqa: F841
+        expect = p_lhs.merge(p_rhs, on="a", how="left")
+    got = lhs.merge(rhs, on="a", how="left")
+
+    # The dtypes here won't match exactly because pandas does some unsafe
```
mroeschke, line 1138: In pandas, we raise a UserWarning too when a user tries to merge on an int and float column but still allow the operation. Maybe we should suggest the user that they cast their columns to avoid the warning

mroeschke (approved):

python/cudf/cudf/core/join/join.py
vyasr: Whoops thanks.

python/cudf/cudf/tests/test_joining.py
vyasr: I think the other issue I had with the pandas operation is the output dtype it chooses isn't safe (hence `check_dtypes=False` below). I've forgotten exactly what output type I got, it might have inferred an integer type?

vyasr:
rerun tests

python/cudf/cudf/tests/test_joining.py
mroeschke: Ah yes thats true

```
In [5]: lhs = pd.DataFrame({"a": range(5)}, dtype="int64")

In [6]: rhs = pd.DataFrame({"b": [1, 2, 3, 4, np.iinfo(np.int64).max + 1]}, dtype="float64")

In [9]: lhs.merge(rhs, on="a", how="left").dtypes
<ipython-input-9-267702b4848a>:1: UserWarning: You are merging on int and float columns where the float values are not equal to their int representation.
  lhs.merge(rhs, on="a", how="left").dtypes
Out[9]:
a    int64
dtype: object
```

I guess technically it's safer if the resulting type was float even thought the values are the "same" and `how="left"` has integer type

vyasr: I guess that it's reasonable to assume that if `how="left"` we should use the dtype of the left column. Joining on a floating point-typed column is pretty suspect in any case. @shwina what do you think we should do in this case? Maybe it does make sense to convert to int. @mroeschke does pandas do any introspection to determine whether the values in the column are equivalent to integers (e.g. via `x == x.round()`) or does it just cast and hope for the best?

mroeschke: > does pandas do any introspection to determine whether the values in the column are equivalent to integers (e.g. via x == x.round()) or does it just cast and hope for the best?

In this code path, it's a naive cast-left-to-right's-dtype and `==` to see if they are equivalent.

More generally we have a "can this scalar fit in this dtype" function https://github.com/pandas-dev/pandas/blob/dba2080af7647145ce82669c9d67cf2408b2a48a/pandas/core/dtypes/cast.py#L1642. I think we might have an array version of this but can't find it at the moment

vyasr: The "can this scalar fit in this dtype" approach is a non-starter for cudf due to the level of introspection it would require and the associated performance bottlenecks. The naive cast and check is a potentially acceptable though. @brandon-b-miller what do you think?

brandon-b-miller: Originally we wanted to [match what pandas did here](https://github.com/rapidsai/cudf/issues/2230#issuecomment-555562743), but there were some edge cases I ran into at the time it seemed like we shouldn't try and match. This led to us wanting to [have our own set of rules](https://github.com/rapidsai/cudf/issues/2230#issuecomment-555578556).

Some of what we came up with _did_ involve data introspection. For instance as you said, for a left join it seemed to make sense to prioritize the dtype of the left table and cast the other operand. IIRC it was while pondering this that I realized the danger that both rounding and overflow errors present in this scheme. For example if the left join column is int dtype and the right is float, we'd match a `1` on the left with both a `1.0` and a `1.5` on the right and end up with too many rows in the result. IIRC, this is what led us to `can_cast_safely`, which is supposed to tell us if every value in that column can be fit into a target dtype losslessly.

vyasr: Got it. It looks like a lot of thought already went into accepting this divergence from pandas, so I'm fine with leaving this as is. Happy to revisit in the future if anyone else feels strongly.

vyasr:
@gpucibot merge

PR #12309, opened 2022-12-05
CSV, JSON reader to infer integer column with nulls as int64 instead of float64

karthikeyann:
## Description
When there are nulls in an integer column, libcudf csv, json readers infer it as float column (similar to pandas).
Pandas uses `float nan` because default types do no have null representation. But cuDF has null representation on all types.
Integer column with nulls in CSV or JSON will be infered as nullable integer column, instead of current behaviour of inferring as float64 column.

## Checklist


davidwendt (approved):

galipremsagar (approved):

karthikeyann:
@gpucibot merge

PR #12310, opened 2022-12-05
Fix warnings in test_multiindex.py

vyasr:
## Description

## Checklist

### DIFF
--- a/python/cudf/cudf/core/column_accessor.py
+++ b/python/cudf/cudf/core/column_accessor.py
@@ -249,9 +249,7 @@ class ColumnAccessor(abc.MutableMapping):
             # prevents coercion of values to a different type
             # (e.g., ''->NaT)
             result = pd.MultiIndex.from_frame(
-                pd.DataFrame(
-                    self.names, columns=self.level_names, dtype="object"
-                ),
+                pd.DataFrame(self.names, columns=self.level_names)
             )
         else:
             result = pd.Index(self.names, name=self.name, tupleize_cols=False)
--- a/python/cudf/cudf/core/multiindex.py
+++ b/python/cudf/cudf/core/multiindex.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import itertools
 import numbers
 import pickle
+import warnings
 from collections import abc
 from functools import cached_property
 from numbers import Integral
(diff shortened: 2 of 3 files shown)


python/cudf/cudf/core/column_accessor.py
```diff
@@ -249,9 +249,7 @@ def to_pandas_index(self) -> pd.Index:
             # prevents coercion of values to a different type
             # (e.g., ''->NaT)
             result = pd.MultiIndex.from_frame(
-                pd.DataFrame(
-                    self.names, columns=self.level_names, dtype="object"
-                ),
```
shwina, line 254: Is the `dtype="object"` needed to prevent the kind of coercion alluded to in the comment above?

vyasr: I don't think so. AFAICT making a pandas `DataFrame` (with typed columns) and using `from_frame` is what's needed here. Specifying `dtype="object"` is problematic because pandas plans to stop inferring dtypes when a MultiIndex is constructed from a `DataFrame` where all the dtypes are `object`. Do you have an example of a case that we should test and verify?

shwina: I think the comment refers to cases like the one below:

```python
In [107]: ca = cudf.core.column_accessor.ColumnAccessor({'': [], pd.Timestamp('2001-01-01'): []})

In [108]: pd.MultiIndex.from_frame(pd.DataFrame(ca.names))  # '' becomes 'NaT'
<warning snipped>
Out[108]: 
MultiIndex([(       'NaT',),
            ('2001-01-01',)],
           names=[0])

In [110]: pd.MultiIndex.from_frame(pd.DataFrame(ca.names, dtype='object'))  # '' stays ''
Out[110]: 
MultiIndex([(                 '',),
            (2001-01-01 00:00:00,)],
           names=[0])
```

vyasr: Done. I also inserted an assertion that the pandas version is less than 2.0 to make sure that we revisit this when the time comes to upgrade.

shwina (approved):

vyasr:
@gpucibot merge

Issue #12299, opened 2022-12-03
[FEA] Support multiple compressed inputs in `read_json`


vuule:
JSON reader currently concatenates data from all input sources before (optional) decompression and parsing.
This is not correct for compressed input files, as each file is compressed separately.
https://github.com/rapidsai/cudf/pull/12285 disables multiple inputs for compressed data.
`read_json` should decompress each input file before concatenating the decompressed data.

karthikeyann:
Note that Only a single source is currently supported in `read_csv`.
This work could help expand the same feature to csv reader.

PR #12314, opened 2022-12-06
Use `device_read` in experimental `read_json`

vuule:
## Description

Enables use of GDS and other optimizations that apply to `device_read` and `device_read_async`.

Seeing ~10% performance gain in the benchmarks.

## Checklist


vuule:
@karthikeyann would love to get some early feedback from you

cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -38,29 +39,47 @@ size_t sources_size(host_span<std::unique_ptr<datasource>> const sources,
...
-      auto destination = buffer.data() + bytes_read;
-      bytes_read += source->host_read(range_offset, data_size, destination);
-    }
-  }
 
   if (compression == compression_type::NONE) {
-    return buffer;
+    auto d_buffer     = rmm::device_uvector<char>(total_source_size, stream);
+    size_t bytes_read = 0;
+    for (const auto& source : sources) {
```
karthikeyann, line 55: Concatenating entire sources makes sense for json_lines.

This logic of concatenating multiple chunks (with range_offset) from different sources does not make sense for json, even for uncompressed. IIRC, I replicated this logic from another cudf code. May be we should rethink this logic again.

karthikeyann (commented):
Logic on the concatenation from multiple sources needs review.
Rest of the code looks good.

cpp/src/io/json/experimental/read_json.cpp
vuule: I was wondering about some code in here :) For example, it's not clear to me what `byte_range` means with multiple sources.
What do you think about addressing this in a separate PR? This one should be focused on `host_read` -> `device_read` transition.
Is there an issue where we can discuss this further? I only opened https://github.com/rapidsai/cudf/issues/12299

karthikeyann: Let's fix it in another PR. This needs more discussion. byte range for different sources may be different, so we may request byte range for each source.
For now, it could throw error "Multiple-sources with byte range is not supported"  (non-zero range_size).

vuule:
@karthikeyann I'm seeing a slight drop in peak memory use in the benchmarks. Any idea where this could be coming from? The only thing I can think of is that I eliminated a copy somewhere. But I'm worried that it's a bug in the benchmark instead.

cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -72,27 +91,9 @@ size_type find_first_delimiter_in_chunk(host_span<std::unique_ptr<cudf::io::data
   auto const buffer = ingest_raw_input(sources,
```
karthikeyann, line 91: Note: `find_first_delimiter_in_chunk` is used only in unit test `json_chunked_reader.cpp`.

karthikeyann (approved):
LGTM 👍

cpp/src/io/json/nested_json_gpu.cu
```diff
@@ -1713,18 +1713,20 @@ std::pair<std::unique_ptr<column>, std::vector<column_name_info>> json_column_to
...
                                            rmm::mr::device_memory_resource* mr)
 {
   // Range of orchestrating/encapsulating function
   CUDF_FUNC_RANGE();
 
-  auto const new_line_delimited_json = options.is_enabled_lines();
+  std::vector<SymbolT> h_input;
+  h_input.resize(d_input.size());
+  CUDF_CUDA_TRY(cudaMemcpyAsync(
+    h_input.data(), d_input.data(), h_input.size(), cudaMemcpyDefault, stream.value()));
```
bdice, line 1727: For my education -- where it is safe to use `cudaMemcpyDefault` instead of `cudaMemcpyXToY` (where `X`, `Y` are `Host` or `Device`)?

The docs say:
> Direction of the transfer is inferred from the pointer values. Requires unified virtual addressing.
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g18fa99055ee694244a270e4d5101e95b

but I am not confident about whether this means "always safe" or "only sometimes safe." Is UVA always available? If so, should we adopt `cudaMemcpyDefault` everywhere? If there is a performance cost or downside of some other kind, should we never use `cudaMemcpyDefault` and always use a specific host/device mode?

cpp/tests/io/nested_json_test.cpp
```diff
@@ -469,9 +469,10 @@ TEST_P(JsonParserTest, ExtractColumn)
   cudf::io::json_reader_options default_options{};
 
   std::string const input = R"( [{"a":0.0, "b":1.0}, {"a":0.1, "b":1.1}, {"a":0.2, "b":1.2}] )";
+  auto const d_input      = cudf::detail::make_device_uvector_sync(
```
bdice, line 472: It should be safe to use async streams here (and below), right?

bdice (changes requested):

cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -38,29 +39,47 @@ size_t sources_size(host_span<std::unique_ptr<datasource>> const sources,
...
+        if (source->is_device_read_preferred(data_size)) {
+          bytes_read += source->device_read(range_offset, data_size, destination, stream);
+        } else {
+          auto const h_buffer = source->host_read(range_offset, data_size);
+          CUDF_CUDA_TRY(cudaMemcpyAsync(destination,
+                                        h_buffer->data(),
+                                        h_buffer->size(),
+                                        cudaMemcpyHostToDevice,
+                                        stream.value()));
+          stream.synchronize();
```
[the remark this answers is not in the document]
hyperbolic2346: I believe this synchronization is needed to prevent the destruction of the host buffer before the copy has completed. That said, I wonder about `cudaMemcpyAsync` here as well.

cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -38,29 +39,47 @@ size_t sources_size(host_span<std::unique_ptr<datasource>> const sources,
...
+          bytes_read += h_buffer->size();
+        }
+      }
+    }
+    return d_buffer;
+
   } else {
-    return decompress(compression, buffer);
+    auto buffer = std::vector<uint8_t>(total_source_size);
+    // only a single compressed source is supported
```
hyperbolic2346, line 77: Should there be an assert here that `sources.size() == 1`?

cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -38,29 +39,47 @@ size_t sources_size(host_span<std::unique_ptr<datasource>> const sources,
...
+    sources[0]->host_read(range_offset, total_source_size, buffer.data());
```
hyperbolic2346, line 78: Do we not want to do a device read here?

hyperbolic2346 (changes requested):
Looking good. Couple of questions.

cpp/src/io/json/experimental/read_json.cpp
vuule: Unfortunately not. Since the whole file is compressed as a single chunk, decompression is expected to be much faster on the CPU.
Once we add support for multiple compressed files, we can consider heuristics that choose where decompression is done based on the size and number of inputs. But I expect that this is a niche case.

cpp/src/io/json/experimental/read_json.cpp
vuule: We have this assert elsewhere:
https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/src/io/json/experimental/read_json.cpp#L164-L166
I chose not to repeat it, but I'm not adamant about that decision.

cpp/src/io/json/experimental/read_json.cpp
bdice: I’d consider adding that info (CPU is faster for single chunks) to a comment here. Otherwise I’d never realize that is the reasoning.

cpp/src/io/json/experimental/read_json.cpp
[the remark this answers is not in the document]
vuule: Yes, we sync because of the short life span of `h_buffer`. We could avoid syncing in each iteration if we keep the host buffer around for a bit longer, but I'm not sure it that would have a measurable impact. Let me know if you think this is a worthwhile change.
I default to `cudaMemcpyAsync` because it takes a stream, and we seem to be completely avoiding `cudaMemcpy` in libcudf (presumably because of the use of default stream?).

cpp/src/io/json/nested_json_gpu.cu
vuule: Great question :)
I asked around and it seems like `cudaMemcpyDefault` is preferable if your system has unified virtual addressing (and it seems like only 32bit Windows does not). There's no performance downside.

cpp/src/io/json/experimental/read_json.cpp
vuule: added a comment

cpp/src/io/json/experimental/read_json.cpp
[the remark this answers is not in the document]
bdice: Good answer. Thanks!

cpp/src/io/json/nested_json_gpu.cu
```diff
@@ -1725,18 +1725,20 @@ std::pair<std::unique_ptr<column>, std::vector<column_name_info>> json_column_to
...
```
bdice, line 1739: Should `h_input.size()` be `d_input.size() * sizeof(SymbolT)`? It looks like the factor of `sizeof(SymbolT)` is missing here.

If so, can this whole block be replaced with [`make_std_vector_async`](https://github.com/rapidsai/cudf/blob/8d1e2ccaccce86fcc9c9e71012137960c205bcf9/cpp/include/cudf/detail/utilities/vector_factories.hpp#L307)?

If that's _not_ the case and the size is correct, can we have a comment about why this is the correct size?
```suggestion
  auto h_input = cudf::detail::make_std_vector_async(d_input, stream);
```

bdice (changes requested):
Question about a raw `cudaMemcpyAsync`.

cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -38,29 +39,48 @@ size_t sources_size(host_span<std::unique_ptr<datasource>> const sources,
...
+        auto data_size   = (range_size != 0) ? range_size : source->size();
+        auto destination = reinterpret_cast<uint8_t*>(d_buffer.data()) + bytes_read;
+        if (source->is_device_read_preferred(data_size)) {
+          bytes_read += source->device_read(range_offset, data_size, destination, stream);
+        } else {
+          auto const h_buffer = source->host_read(range_offset, data_size);
+          CUDF_CUDA_TRY(cudaMemcpyAsync(destination,
+                                        h_buffer->data(),
+                                        h_buffer->size(),
+                                        cudaMemcpyHostToDevice,
```
bdice, line 66: Replaced across the codebase in #12466.
```suggestion
                                        cudaMemcpyDefault,
```

cpp/src/io/json/nested_json_gpu.cu
```diff
@@ -1725,18 +1725,17 @@ std::pair<std::unique_ptr<column>, std::vector<column_name_info>> json_column_to
...
+table_with_metadata host_parse_nested_json(device_span<SymbolT const> d_input,
                                            cudf::io::json_reader_options const& options,
                                            rmm::cuda_stream_view stream,
                                            rmm::mr::device_memory_resource* mr)
 {
   // Range of orchestrating/encapsulating function
   CUDF_FUNC_RANGE();
 
-  auto const new_line_delimited_json = options.is_enabled_lines();
+  auto h_input = cudf::detail::make_std_vector_async(d_input, stream);
```
bdice, line 1736: Looks like this can be `const`.
```suggestion
  auto const h_input = cudf::detail::make_std_vector_async(d_input, stream);
```

bdice (approved):
Two minor suggestions. Thanks for the iterations on this!

ttnghia (approved):

hyperbolic2346 (approved):

karthikeyann:
Great work @vuule ! 🏅

Issue #11890, opened 2022-10-10
[BUG] cuDF ORC reader incorrectly reads file written by pyarrow


vuule:
Elements after the first row group are empty.

Only reproes with a string column.
Does not repro when the file is created by cuDF.

Minimal repro code:
```
expected = cudf.DataFrame({"str": ["*"] * 10001}, dtype="string")

buffer = BytesIO()
expected.to_pandas().to_orc(buffer)

got = cudf.read_orc(buffer)
assert_eq(expected, got)
```

vuule:
Seems to be related to row group index; reading nonsensical offsets for the second row group (larger than the streams):
```
strm_offset0, strm_offset1: 272 10000
strm_len, strm_len2: 10001 80
```
Even if we switch the offsets around 272 does not make any sense. Trying to find whether the stream lengths are wrong or the offsets.

vuule:
Turns out that the row index streams are read in incorrect order. 
The values that we read from the pyarrow file are 
byte offset, RLE run
272,         0
10000,       72

The correct values are 
byte offset, RLE run
10000,       0
72,          272

With a hack to read the streams in inverse order, the file is read correctly.

vuule:
Now, as to why the row index streams are read in wrong order:
libcudf ORC reader reader these streams in the same order as the order of corresponding data streams in the stripe footer.
When reading the pyarrow file these data streams' descriptors are read from the stripe footer as [LENGTH, DATA], while we read the row index streams in the [DATA, LENGTH] order. This difference in order causes the difference shown in the comment above.

vuule:
So, why do we read these in inconsistent order? Either libcudf mixes up the order somehow, or the file stores this info in the incorrect order.
With some trace "logging" from the ORC reader, we can inspect the stripe footer and the row index, as they appear in the file (protobuf parsing always marches forward over the input, so the output order much match the input order).

_Reading the same data written by cuDF with dictionary encode disabled (RI stands for row index):_
StripeFooter begin
Stream begin
enum read: 6
Stream end
Stream begin
enum read: 6
Stream end
Stream begin
**enum read: 1** <- DATA stream type
Stream end
Stream begin
**enum read: 2** <- LENGTH stream type
Stream end
ColumnEncoding begin
enum read: 0
ColumnEncoding end
ColumnEncoding begin
enum read: 2
ColumnEncoding end
StripeFooter end
RI: 10 RI: 22 
[...]

PR #12325, opened 2022-12-06
Fail loudly to avoid data corruption with unsupported input in `read_orc`

vuule:
## Description
Issue https://github.com/rapidsai/cudf/issues/11890

Motivating issue: The ORC reader reads nulls in row groups after the first one when reading a string column encoded with Pandas, with direct encoding. The root cause is that cuDF reads offsets from the row group index as larger then the stream sizes.

This PR does not fix the issue, but ensures that the reader fails loudly when the row group index offsets are read as too large to be correct. This should prevent data corruption until the fix is implemented. 

This PR also sets up a mechanism to report decode errors from unsupported data. 

## Checklist

### DIFF
--- a/cpp/src/io/orc/orc_gpu.hpp
+++ b/cpp/src/io/orc/orc_gpu.hpp
@@ -287,6 +287,7 @@ void DecodeNullsAndStringDictionaries(ColumnDesc* chunks,
  * @param[in] num_rowgroups Number of row groups in row index data
  * @param[in] rowidx_stride Row index stride
  * @param[in] level Current nesting level being processed
+ * @param[out] error_count Number of errors during decode
  * @param[in] stream CUDA stream used for device memory operations and kernel launches
  */
 void DecodeOrcColumnData(ColumnDesc* chunks,
@@ -299,6 +300,7 @@ void DecodeOrcColumnData(ColumnDesc* chunks,
                          uint32_t num_rowgroups,
                          uint32_t rowidx_stride,
                          size_t level,
+                         unsigned long long int* error_count,
                          rmm::cuda_stream_view stream);
 
 /**
--- a/cpp/src/io/orc/orc_gpu.hpp
+++ b/cpp/src/io/orc/orc_gpu.hpp
@@ -300,7 +300,7 @@ void DecodeOrcColumnData(ColumnDesc* chunks,
                          uint32_t num_rowgroups,
                          uint32_t rowidx_stride,
                          size_t level,
-                         unsigned long long int* error_count,
+                         size_type* error_count,
                          rmm::cuda_stream_view stream);
 
 /**
--- a/cpp/src/io/orc/reader_impl.cu
+++ b/cpp/src/io/orc/reader_impl.cu
@@ -36,6 +36,7 @@
 #include <cudf/utilities/bit.hpp>
 #include <cudf/utilities/error.hpp>
 #include <cudf/utilities/traits.hpp>
+#include <rmm/device_scalar.hpp>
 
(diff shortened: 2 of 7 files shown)


galipremsagar (approved):

cpp/src/io/orc/orc_gpu.hpp
```diff
@@ -287,6 +287,7 @@ void DecodeNullsAndStringDictionaries(ColumnDesc* chunks,
  * @param[in] num_rowgroups Number of row groups in row index data
  * @param[in] rowidx_stride Row index stride
  * @param[in] level Current nesting level being processed
+ * @param[out] error_count Number of errors during decode
```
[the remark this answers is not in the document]
vuule: `DecodeOrcColumnData` is asynchronous. The fact that we copy `chunks` to host immediately after calling `DecodeOrcColumnData` should not impact how its implemented. If we return the error code we are enforcing this synchronization even though it might not be required otherwise.

wence- (commented):
Comment on the xfail in the test (sorry this was a bit delayed)

python/cudf/cudf/tests/test_orc.py
```diff
@@ -1834,3 +1834,22 @@ def test_reader_empty_stripe(datadir, fname):
...
+    # Reading this file should not lead to data corruption, even if it fails
+    try:
+        got = cudf.read_orc(buffer)
+    except RuntimeError:
+        pytest.mark.xfail(
+            reason="Unsupported file, "
+            "see https://github.com/rapidsai/cudf/issues/11890"
+        )
+    else:
+        assert_eq(expected, got)
```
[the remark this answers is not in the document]
vuule: done, thank you

hyperbolic2346 (approved):
I think IO is a special beast as far as the cudf mantra of not validating input. I also think in this case there isn't really any extra overhead to do it. I like this change.

cpp/src/io/orc/reader_impl.cu
```diff
@@ -641,6 +642,7 @@ void reader::impl::decode_stream_data(cudf::detail::hostdevice_2dvector<gpu::Col
     update_null_mask(chunks, out_buffers, stream, _mr);
   }
 
+  rmm::device_scalar<unsigned long long int> error_count(0, stream);
```
ttnghia, line 645: `unsigned long long int` seems very long and ambiguous. Please use the fixed width type, such as `uint64_t`.

ttnghia: In addition, can we just use `uint32_t` instead? I'm not sure if `uint64_t` is supported in `atomicAdd` below.

vuule: These are atomicAdd overloads:
```
int atomicAdd(int* address, int val);
unsigned int atomicAdd(unsigned int* address,
 unsigned int val);
unsigned long long int atomicAdd(unsigned long long int* address,
 unsigned long long int val);
 ```
 I can use a smaller type, but should not be using `uint32_t` or `uint64_t`, AFAIK.

vuule: Went over the uses of atomicAdd in libcudf, and, to my surprise, we often use it with `size_type`. I'll change this to `size_type`.

ttnghia: > I can use a smaller type, but should not be using uint32_t or uint64_t, AFAIK.

`unsigned long long int` should just be alias of `uint64_t` if I was not wrong.

ttnghia (approved):

vuule:
@gpucibot merge

Issue #5097, opened 2020-05-05
[QST] Would the new Arrow C Data Interface be useful for CuDF?


andygrove:
**What is your question?**

Apache Arrow has released a new C Data Interface [1] and it seems like it might potentially be of interest to the CuDF project, so I wanted to make sure that people are at least aware of this.

Here is an extract from the blog post that seems relevant.

> However, some projects may wish to import and export the Arrow columnar format without taking on a new library dependency, such as the Arrow C++ library. We have therefore designed an alternative which exchanges data at the C level, conforming to a simple data definition. The C Data Interface carries no dependencies except a shared C ABI between binaries which use it. C ABIs are platform-wide standards which are necessarily adhered to by all compilers which generate binaries and are extremely stable, ensuring portability of libraries and executable binaries. Two libraries that utilize the C structures defined by the C Data Interface can do zero-copy data transfers at runtime without any build-time or link-time dependency requirements.

[1] https://arrow.apache.org/blog/2020/05/03/introducing-arrow-c-data-interface/

kkraus14:
Documentation on the interface: https://arrow.apache.org/docs/format/CDataInterface.html

kkraus14:
Big concern from me is there's no way to differentiate between GPU and host memory via this interface currently. This means if we exposed this interface with GPU memory pointers, a downstream consumer could pick it up assuming it's host memory and blow up because of it.

While we can use built in CUDA features to check whether a given pointer is host or device memory, non-GPU libraries wouldn't perform these checks. We should work with the community to extend the interface to allow differentiating between host and device memory.

fsaintjacques:
This is meant to pass data around between libraries/runtimes without having to link with Arrow's C++ (and copying). The first consumer in the wild of this interface is to pass data from R to Python without copying in the same process.

I would say that implementing this feature scoped to host memory only (both in and out) is already a major milestone, essentially dispatching some computation to the GPU an getting results back.

kkraus14:
> I would say that implementing this feature scoped to host memory only (both in and out) is already a major milestone, essentially dispatching some computation to the GPU an getting results back.

So all of our data structures are ultimately backed by GPU memory where for consuming the interface it makes sense where we can copy it to the GPU and then immediately call the release callback once the copy is done.

In the other direction, we'd need to allocate host memory to copy the GPU memory into and then manage the lifetime of those memory buffers via the release callback.

fsaintjacques:
[...]

PR #12327, opened 2022-12-07
[REVIEW] Upgrade to `arrow-10.0.1`

galipremsagar:
## Description
This PR upgrades `arrow` to `10.0.1` in `cudf`

## Checklist


python/cudf/cudf/_lib/CMakeLists.txt
```diff
@@ -89,6 +89,19 @@ foreach(target IN LISTS targets_using_dlpack)
   target_include_directories(${target} PRIVATE "${DLPACK_INCLUDE_DIR}")
 endforeach()
 
+find_package(Python 3.8 REQUIRED COMPONENTS Interpreter)
```
galipremsagar, line 92: This change was necessary because of the change of placement of header files in arrow-10:
```
arrow-9:

/nvme/0/pgali/envs/cudfdev/include/arrow/python/platform.h




arrow-10:

/nvme/0/pgali/envs/cudfdev/lib/python3.9/site-packages/pyarrow/src/arrow/python/platform.h
```

conda/environments/all_cuda-115_arch-x86_64.yaml
```diff
@@ -9,7 +9,8 @@ channels:
 - nvidia
 dependencies:
 - aiobotocore>=2.2.0
-- arrow-cpp=9
+# - libarrow=10.0.1
```
galipremsagar, line 12:
```suggestion
```

ci/gpu/build.sh
```diff
@@ -120,6 +120,9 @@ function install_dask {
 
 install_dask
 
+gpuci_conda_retry remove --force pyarrow arrow-cpp
+gpuci_mamba_retry install -y "pyarrow=10.0.1" "arrow-cpp=10.0.1"
```
galipremsagar, line 124: # TODO: Need to remove these after https://github.com/rapidsai/integration/pull/577 is merged.

ci/cpu/build.sh
```diff
@@ -78,7 +78,8 @@ conda config --set ssl_verify False
 
 # TODO: Move boa install to gpuci/rapidsai
 gpuci_mamba_retry install boa
-
+gpuci_conda_retry remove --force pyarrow arrow-cpp
+gpuci_mamba_retry install -y "pyarrow=10.0.1" "arrow-cpp=10.0.1"
```
galipremsagar, line 82: TODO: Need to remove these after https://github.com/rapidsai/integration/pull/577 is merged.

ci/gpu/build.sh
bdice: We should use the new naming: libarrow instead of arrow-cpp.
https://github.com/conda-forge/arrow-cpp-feedstock/blob/21e42a6c6a9566acebd50cc9efd6d921635204de/recipe/meta.yaml#L186

dependencies.yaml
```diff
@@ -57,16 +57,16 @@ dependencies:
...
           - dlpack>=0.5,<0.6.0a0
           - ninja
-          - &pyarrow_ver pyarrow=9.0.0
+          - pyarrow=10.0.1
           - rmm=23.02.*
           - scikit-build>=0.13.1
       - output_types: conda
         packages:
-          - arrow-cpp=9
+          - arrow-cpp=10
```
bdice, line 65: Use libarrow as above (and elsewhere).

galipremsagar: Done 👍  Was planning to do it a separate PR. But pushed it to this PR now, @ajschmidt8 do we need to change this elsewhere outside cudf? `arrow-cpp` is being replaced by `libarrow`

bdice: Probably just this repo, based on search results. https://github.com/search?q=org%3Arapidsai%20arrow-cpp&type=code

ci/gpu/java.sh
```diff
@@ -88,6 +88,8 @@ conda activate rapids
 # https://docs.rapids.ai/maintainers/depmgmt/
 # gpuci_conda_retry remove --force rapids-build-env rapids-notebook-env
 # gpuci_mamba_retry install -y "your-pkg=1.0.0"
+gpuci_conda_retry remove --force pyarrow arrow-cpp
+gpuci_mamba_retry install -y "pyarrow=10.0.1" "arrow-cpp=10.0.1"
```
robertmaynard, line 92:
```suggestion
gpuci_mamba_retry install -y "pyarrow=10.0.1" "libarrow=10.0.1"
```

robertmaynard (changes requested):

python/cudf/cudf/_lib/CMakeLists.txt
robertmaynard: Was this an intended change of pyarrow or was this a accidental change / bug? We should open an issue if this wasnt intended

conda/environments/all_cuda-115_arch-x86_64.yaml
```diff
@@ -9,7 +9,7 @@ channels:
 - nvidia
 dependencies:
 - aiobotocore>=2.2.0
-- arrow-cpp=9
+- libarrow=10.0.1
```
kkraus14, line 12: We had previously made this `9` instead of pinning to a specific patch to allow floating forward as new patch versions are released. The Arrow community has strived to ensure patch releases are ABI compatible.

Any issue with making this `libarrow=10` or if the patch is needed then `libarrow>=10.0.1,<11`?

galipremsagar: Oh yes, now I remember. We can probably make this `libarrow=10`

cc: @robertmaynard for vis

galipremsagar, line 12:
```suggestion
- libarrow=10
```

conda/recipes/libcudf/conda_build_config.yaml
```diff
@@ -16,8 +16,8 @@ cmake_version:
 gtest_version:
   - "=1.10.0"
 
-arrow_cpp_version:
-  - "=9"
+libarrow_version:
+  - "=10.0.1"
```
galipremsagar, line 20:
```suggestion
  - "=10"
```

ci/cpu/build.sh
```diff
@@ -78,7 +78,8 @@ conda config --set ssl_verify False
 
 # TODO: Move boa install to gpuci/rapidsai
 gpuci_mamba_retry install boa
-
+gpuci_conda_retry remove --force pyarrow arrow-cpp
+gpuci_mamba_retry install -y "pyarrow=10.0.1" "libarrow=10.0.1"
```
galipremsagar, line 82:
```suggestion

```

ci/gpu/build.sh
```diff
@@ -120,6 +120,9 @@ function install_dask {
...
+
```
galipremsagar, line 125:
```suggestion
```

ci/gpu/java.sh
```diff
@@ -88,6 +88,8 @@ conda activate rapids
 # https://docs.rapids.ai/maintainers/depmgmt/
 # gpuci_conda_retry remove --force rapids-build-env rapids-notebook-env
 # gpuci_mamba_retry install -y "your-pkg=1.0.0"
+gpuci_conda_retry remove --force pyarrow arrow-cpp
+gpuci_mamba_retry install -y "pyarrow=10.0.1" "libarrow=10.0.1"
```
galipremsagar, line 92:
```suggestion
```

python/cudf/cudf/_lib/CMakeLists.txt
galipremsagar: This was an intended change, all the python headers are removed from `arrow-cpp` and it is now being shipped as `libarrow` and the python headers have moved to its own `pyarrow` package.

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -365,7 +365,7 @@ endfunction()
 
 if(NOT DEFINED CUDF_VERSION_Arrow)
   set(CUDF_VERSION_Arrow
-      9.0.0
+      10.0.1
```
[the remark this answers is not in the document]
galipremsagar: Yup

conda/environments/all_cuda-115_arch-x86_64.yaml
robertmaynard: Sounds good to me

python/cudf/cudf/_lib/CMakeLists.txt
robertmaynard: Right but why would the location of the headers change? They can still be part of `pyarrow` but exist in the old location

galipremsagar: These actually never existed in `pyarrow` before. Only existed in `arrow-cpp`, with `10.x` the files have been moved from `arrow-cpp` to `pyarrow`. Do you mean `/nvme/0/pgali/envs/cudfdev/include/arrow/python/platform.h` should still exist? I feel it is a path that corresponds to `arrow-cpp`(since it says `arrow` & not `pyarrow`).

kkraus14: @robertmaynard for wheel packaging (as opposed to conda) everything needs to be self contained in the `site-packages` directory, so it couldn't land in a place where CMake would naturally search.

In theory this could be handled differently for the conda packages, but it would likely be a non-trivial amount of work and maintenance.

cpp/cmake/thirdparty/get_arrow.cmake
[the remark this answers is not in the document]
kkraus14: What is the behavior if someone has 10.0.0 installed and we try to find it? Will it reject it because we have this 10.0.1 version?

What happens if there's subsequent 10.0.2 and 10.0.3 patch releases? Would we continue to upgrade this version and potentially break being able to find and build with previously patch versions?

python/cudf/cudf/_lib/CMakeLists.txt
robertmaynard: Thanks @galipremsagar and @kkraus14 for the explanations. If this is the layout that makes the most sense from a python package perspective we will have to live with it.

robertmaynard (approved):

cpp/cmake/thirdparty/get_arrow.cmake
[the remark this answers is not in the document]
vyasr: AFAIK that depends on Arrow's CMake code. Looking at the version config file in my local copy, I think the answer is yes; it seems like it will automatically fail on any version strictly less than the requested version. I guess this relates to my other question, are arrow libraries ABI compatible across patch versions (hopefully)? Minor versions? Is it safe to compiler against 10.0.0 then use in a system with libarrow 10.0.1 installed?

conda/environments/all_cuda-115_arch-x86_64.yaml
```diff
@@ -43,15 +43,14 @@ dependencies:
...
 - packaging
 - pandas>=1.0,<1.6.0dev0
 - pandoc<=2.0.0
 - pip
 - pre-commit
-- protobuf>=3.20.1,<3.21.0a0
+- protobuf=4.21
 - ptxcompiler
-- pyarrow=9.0.0
+- pyarrow=10
```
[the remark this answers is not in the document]
vyasr: Just realized that this is the generated file, will redirect future queries to dependencies.yaml.

python/cudf/cudf/_lib/CMakeLists.txt
vyasr: This is one of the fundamental problems for pip packaging wherein both library dependencies and CMake configs don't get installed into any globally visible places. Not much we can do here, given the limitations I think it makes sense for pyarrow not to diverge in its wheel and conda packaging. FWIW this approach has precedent from packages like numpy, and cudf already uses this exact logic to find arrow when we build wheels against the libarrow contained in the pyarrow wheel.

python/cudf/cudf/_lib/io/CMakeLists.txt
```diff
@@ -20,7 +20,16 @@ rapids_cython_create_modules(
   LINKED_LIBRARIES "${linked_libraries}" MODULE_PREFIX io_ ASSOCIATED_TARGETS cudf
 )
 
-set(targets_using_numpy io_datasource io_utils)
-foreach(target IN LISTS targets_using_numpy)
+find_package(Python 3.8 REQUIRED COMPONENTS Interpreter)
```
vyasr, line 23: @robertmaynard this is more of a CMake best practices question; should we duplicate this `find_package` here so that we don't rely on the global state of the find happening in a higher-level CMakeLists.txt file, or should we avoid duplicating and trust future devs (most like you or me when we upgrade to CMake 3.25) to make the necessary adjustments?

dependencies.yaml
```diff
@@ -57,16 +57,16 @@ dependencies:
           - cython>=0.29,<0.30
           - dlpack>=0.5,<0.6.0a0
           - ninja
-          - &pyarrow_ver pyarrow=9.0.0
+          - pyarrow=10
```
vyasr, line 60: Porting my question from the env file: is there a reason that this is only a major version now?

dependencies.yaml
```diff
@@ -57,16 +57,16 @@ dependencies:
...
           - scikit-build>=0.13.1
       - output_types: conda
         packages:
-          - arrow-cpp=9
+          - libarrow=10
           - c-compiler
           - cxx-compiler
           - librdkafka=1.7.0
-          - protobuf>=3.20.1,<3.21.0a0
+          - protobuf=4.21
```
vyasr, line 69: Porting my protobuf question for @ajschmidt8 @bdice: do we expect any issues with this hard pinning?

vyasr (changes requested):

conda/recipes/cudf/meta.yaml
```diff
@@ -45,19 +45,19 @@ requirements:
...
     - numba >=0.56.2
     - dlpack>=0.5,<0.6.0a0
-    - pyarrow =9
+    - pyarrow =10
     - libcudf ={{ version }}
     - rmm ={{ minor_version }}
     - cudatoolkit ={{ cuda_version }}
   run:
-    - protobuf>=3.20.1,<3.21.0a0
+    - protobuf=4.21
```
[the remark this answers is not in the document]
ajschmidt8: Nothing comes to mind, though Bradley might have input. I think the best bet to solve our woes at this point is to just adopt Arrow 10 ASAP.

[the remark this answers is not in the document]
bdice: +1 to what AJ said. I have no additional context to add but updating seems best.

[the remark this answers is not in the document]
vyasr: Sounds good, resolving. Let's move forward and we can deal with any issues if they arise.

cpp/cmake/thirdparty/get_arrow.cmake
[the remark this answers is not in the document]
kkraus14: Yes, Arrow libraries are ABI compatible across patch versions. Arrow does not do minor releases at this time, but if it does in the future it would be ABI compatible as well. Yes it is safe to compile against `10.0.0` and then use `10.0.1`.

dependencies.yaml
kkraus14: To allow floating forward on patch versions since they're ABI compatible.

cpp/cmake/thirdparty/get_arrow.cmake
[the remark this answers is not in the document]
vyasr: In that case I think it would be sufficient to list a lower bound here (10.0.0) (or if we want to protect against Arrow 11 I think something like `10.0.0...<11.0.0`).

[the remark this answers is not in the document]
kkraus14: Unfortunately this function isn't set up to handle that. This version feeds into the `CPMFindPackage` call in two places:
* To specify a version for the `find_package` call
  * https://github.com/rapidsai/cudf/pull/12327/files/0f3fb8bde7ffe17618aad04c3de02b04b454c8fb#diff-37169a2940bf7bc08ba411474ff544309f1baadbb3c0c3713efbb6d82c0b594fR151
* To define the git tag to use for when we want to build Arrow
  * https://github.com/rapidsai/cudf/pull/12327/files/0f3fb8bde7ffe17618aad04c3de02b04b454c8fb#diff-37169a2940bf7bc08ba411474ff544309f1baadbb3c0c3713efbb6d82c0b594fR156
  
I think we'd need to split things into two variables of something like a `find_version_spec` and `build_version`, which in our case would be `10.0.0...<11.0.0` and `10.0.1` respectively.

Alternatively, we could introduce logic to grab the list of available git tags and select the newest 10.0.x version, but I'm not sure we'd want to do that.

vyasr:
@revans2 with the update to arrow 10 we need to handle the statuses returned by various arrow functions. I've put in some basic exceptions to handle those in cases where they're returning a nonzero status. I'm not sure if you'd prefer more detailed error messages, I'll leave that for Java devs to decide.

conda/environments/all_cuda-115_arch-x86_64.yaml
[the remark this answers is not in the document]
galipremsagar: The reason we switched to not specifying minor and patch versions is due to the lower likelihood of ABI incompatibility between minor versions as discussed by @kkraus14 here: https://github.com/rapidsai/cudf/pull/12327#discussion_r1049794213

That's why relaxed `pyarrow` too and not just `libarrow`.

robertmaynard:
Arrow 10.0.1 doesn't build cleanly from source so please update this PR with patches / flag disables:

Errors:
```
FAILED: CMakeFiles/cudf.dir/src/interop/from_arrow.cu.o 
ccache /usr/local/cuda-11.5/bin/nvcc -forward-unknown-to-host-compiler -DJITIFY_PRINT_LOG=0 -DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_INFO -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CUDA -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_CPP -Dcudf_EXPORTS -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/dlpack-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/jitify-src -I/home/rmaynard/Work/cudf/cpp/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/include -I/home/rmaynard/Work/cudf/cpp/src -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/src -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/src/generated -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/libcudacxx-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/thrust-src -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/thrust-src/dependencies/cub -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/rmm-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/spdlog-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/cuco-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/kvikio-src/cpp/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/thirdparty/hadoop/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/thirdparty/flatbuffers/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-build/xsimd_ep/src/xsimd_ep-install/include -isystem=/usr/local/cuda-11.5/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/nvcomp_proprietary_binary-src/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/nvcomp_proprietary_binary-src/include/gdeflate -O3 -DNDEBUG --generate-code=arch=compute_60,code=[sm_60] --generate-code=arch=compute_70,code=[sm_70] --generate-code=arch=compute_75,code=[sm_75] --generate-code=arch=compute_80,code=[sm_80] --generate-code=arch=compute_86,code=[compute_86,sm_86] -Xcompiler=-fPIC --expt-extended-lambda --expt-relaxed-constexpr -Werror=all-warnings -Xcompiler=-Wall,-Werror,-Wno-error=deprecated-declarations -Xfatbin=-compress-all -Xcompiler -pthread -std=c++17 -MD -MT CMakeFiles/cudf.dir/src/interop/from_arrow.cu.o -MF CMakeFiles/cudf.dir/src/interop/from_arrow.cu.o.d -x cu -c /home/rmaynard/Work/cudf/cpp/src/interop/from_arrow.cu -o CMakeFiles/cudf.dir/src/interop/from_arrow.cu.o
/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/src/arrow/record_batch.h(297): error #2810-D: ignoring return value type with "nodiscard" attribute

/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/src/arrow/record_batch.h(300): error #2810-D: ignoring return value type with "nodiscard" attribute

2 errors detected in the compilation of "/home/rmaynard/Work/cudf/cpp/src/interop/from_arrow.cu".
[126/167] Building CUDA object CMakeFiles/cudf.dir/src/interop/to_arrow.cu.o
FAILED: CMakeFiles/cudf.dir/src/interop/to_arrow.cu.o 
ccache /usr/local/cuda-11.5/bin/nvcc -forward-unknown-to-host-compiler -DJITIFY_PRINT_LOG=0 -DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_INFO -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CUDA -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_CPP -Dcudf_EXPORTS -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/dlpack-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/jitify-src -I/home/rmaynard/Work/cudf/cpp/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/include -I/home/rmaynard/Work/cudf/cpp/src -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/src -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/src/generated -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/libcudacxx-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/thrust-src -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/thrust-src/dependencies/cub -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/rmm-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/spdlog-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/cuco-src/include -I/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/kvikio-src/cpp/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/thirdparty/hadoop/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/thirdparty/flatbuffers/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-build/xsimd_ep/src/xsimd_ep-install/include -isystem=/usr/local/cuda-11.5/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/nvcomp_proprietary_binary-src/include -isystem=/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/nvcomp_proprietary_binary-src/include/gdeflate -O3 -DNDEBUG --generate-code=arch=compute_60,code=[sm_60] --generate-code=arch=compute_70,code=[sm_70] --generate-code=arch=compute_75,code=[sm_75] --generate-code=arch=compute_80,code=[sm_80] --generate-code=arch=compute_86,code=[compute_86,sm_86] -Xcompiler=-fPIC --expt-extended-lambda --expt-relaxed-constexpr -Werror=all-warnings -Xcompiler=-Wall,-Werror,-Wno-error=deprecated-declarations -Xfatbin=-compress-all -Xcompiler -pthread -std=c++17 -MD -MT CMakeFiles/cudf.dir/src/interop/to_arrow.cu.o -MF CMakeFiles/cudf.dir/src/interop/to_arrow.cu.o.d -x cu -c /home/rmaynard/Work/cudf/cpp/src/interop/to_arrow.cu -o CMakeFiles/cudf.dir/src/interop/to_arrow.cu.o
/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/src/arrow/record_batch.h(297): error #2810-D: ignoring return value type with "nodiscard" attribute

/home/rmaynard/Work/cudf/cpp/build/local_cuda_11.5/_deps/arrow-src/cpp/src/arrow/record_batch.h(300): error #2810-D: ignoring return value type with "nodiscard" attribute
```

This is due to the `arrow::Result` type having a `nodiscard` attribute so the assignment operator in the form of `batch_ = Result`   is expected to produce a result which is being discarded ( or so the compiler thinks ). It would need to be `auto result = (batch_ = Result...)`

robertmaynard (changes requested):
Currently updating arrow to 10.0.1 breaks source builds due to nodiscard errors

python/cudf/cudf/_lib/io/CMakeLists.txt
robertmaynard: I think in this given situtation where we are building python bindings it is reasonable to expect that python has been found at the root most level.

java/src/main/native/src/TableJni.cpp
```diff
@@ -263,16 +263,24 @@ class native_arrow_ipc_writer_handle final {
       // empty table, so need to write an empty batch explicitly.
       // For more please see https://issues.apache.org/jira/browse/ARROW-17912.
       auto empty_batch = arrow::RecordBatch::MakeEmpty(arrow_tab->schema());
-      writer->WriteRecordBatch(*(*empty_batch));
+      if (!writer->WriteRecordBatch(*(*empty_batch)).ok()) {
+        throw std::runtime_error("writer failed to write batch");
```
jlowe, line 267: For here and elsewhere in this file, it would be good to add the `Status::ToString` result to the exception message thrown so that, if this does occur in practice, it's not a complete mystery why.

python/cudf/cudf/_lib/io/CMakeLists.txt
robertmaynard: Looking at this more. We should hoist up the `execute_process` to determine the pyarrow header location so we aren't doing that 2 times for each configuration.

java/src/main/native/src/TableJni.cpp
vyasr: Done.

jlowe (approved):
Java approval

python/cudf/cudf/_lib/io/CMakeLists.txt
galipremsagar: > We should hoist up the `execute_process` to determine the pyarrow header location so we aren't doing that 2 times for each configuration.

Not sure how trivial this is, should we take this up in a follow-up PR?

conda/environments/all_cuda-115_arch-x86_64.yaml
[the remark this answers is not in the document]
vyasr: Right, it sounds like we're OK on that front now. My understanding is that was not always the case historically and has necessitated hard pins, so it's good to know that we can rely on it now.

python/cudf/cudf/_lib/io/CMakeLists.txt
vyasr: It's straightforward, a deletion of the second instance is sufficient since they're already nested directories. I will make the change, test locally, and push.

cpp/cmake/thirdparty/get_arrow.cmake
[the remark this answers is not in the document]
robertmaynard: Splitting on two versions would be possible. The ranges isn't an issue as we would just say `10` and that implicitly means `10.0.0...<11.0.0` since `rapids_export` would state it is only compatible across the same version value

( See explicit version entry in https://docs.rapids.ai/api/rapids-cmake/stable/command/rapids_export.html )

revans2 (approved):
Thanks looks good.

cpp/cmake/thirdparty/get_arrow.cmake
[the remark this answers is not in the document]
vyasr: Yeah assuming that the range is only ever within a single major release I agree that the `rapids_export` behavior would be equivalent to the ranges.

python/cudf/cudf/_lib/io/CMakeLists.txt
vyasr: This is done.

cpp/cmake/thirdparty/get_arrow.cmake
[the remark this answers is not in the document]
kkraus14: My main concern with this is related to conda-forge, where libcudf 23.02 will get released with say Arrow 10.0.1, but then there's eventually an Arrow 10.0.2 and 10.0.3 and conda-forge will update its pinning and new versions of packages will end up with the dependency because currently pyarrow pins to exact builds of libarrow (this should probably be addressed as well).

Users could then end up being forced to use older versions of other packages (which they may have wanted newer versions of for new features), which is super non-ideal. Note this is only in the building a library / C++ application path and not in the runtime path as far as I can tell, so it is a bit niche, but it definitely impacts me directly :joy:.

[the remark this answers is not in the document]
robertmaynard: I am confused on the version numbers in `get_arrow` will effect conda pins / users. The `find_package` call will accept any patch of arrow 10.0 without issue.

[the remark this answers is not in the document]
kkraus14: If we pass in `10.0.1` here won't that get passed into the `find_package` call here: https://github.com/rapidsai/cudf/pull/12327/files/0f3fb8bde7ffe17618aad04c3de02b04b454c8fb#diff-37169a2940bf7bc08ba411474ff544309f1baadbb3c0c3713efbb6d82c0b594fR151

Then if I'm reading the Arrow cmake files correctly it treats that call as effectively `>=10.0.1,<11`.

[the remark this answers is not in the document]
kkraus14: I think I got things backwards in my head. As long as libcudf doesn't jump ahead of the conda-forge pinning I think we're fine.

kkraus14 (approved):

cpp/cmake/thirdparty/get_arrow.cmake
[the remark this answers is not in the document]
vyasr: Just to clarify, your real concern would be if libcudf required `>=10.0.2,<11` while other conda-forge packages were still pinned to `<=10.0.1`? I agree, that's something we'd need to avoid.

Implementing #5097 would also simplify these discussions substantially since then the only place any of this would be maintained is in the pyarrow pinnings in conda meta.yaml files and in setup.py for pip.

vyasr (approved):

vyasr:
Given the holidays, I'd recommend waiting to merge this until at least 2/3 of myself, Prem and Robert are available in case merging this breaks anything that needs fixing.

ajschmidt8 (approved):

robertmaynard (approved):

galipremsagar:
@gpucibot merge

PR #12328, opened 2022-12-07
RMM fragmentation statistics

madsbk:
Implements statistics of RMM fragmentation in cudf-spilling.

Enable by setting the stats level to 3 e.g.  `CUDF_SPILL_STATS=3`.

### Motivation
Memory fragmentation can be an issue when using an RMM memory pool. This PR tracks the lowest utilization of device memory, which is a measure of how good the device memory is being utilized. The difference between it and the size of the RMM pool is a measure of fragmentation.  

See `test_statistics_rmm_fragmentation` for a demo of  RMM fragmentation.

### Checklist

### DIFF
--- a/python/cudf/cudf/core/buffer/spill_manager.py
+++ b/python/cudf/cudf/core/buffer/spill_manager.py
@@ -91,13 +91,15 @@ class SpillStatistics:
         spilled_nbytes: int = 0
 
     spill_totals: Dict[Tuple[str, str], Tuple[int, float]]
+    exposes: Dict[str, SpillStatistics.Expose]
+    lowest_utilization: Optional[int]
 
     def __init__(self, level) -> None:
         self.lock = threading.Lock()
         self.level = level
         self.spill_totals = defaultdict(lambda: (0, 0))
-        # Maps each traceback to a Expose
-        self.exposes: Dict[str, SpillStatistics.Expose] = {}
+        self.exposes = {}
+        self.lowest_utilization = None
 
     def log_spill(self, src: str, dst: str, nbytes: int, time: float) -> None:
         """Log a (un-)spilling event
@@ -151,6 +153,21 @@ class SpillStatistics:
                 stat.total_nbytes += buf.nbytes
                 stat.spilled_nbytes += spilled_nbytes
 
+    def log_spill_on_demand(self, manager: SpillManager, nbytes: int):
+        if self.level < 3:
+            return
+
+        unspilled = sum(
+            buf.size for buf in manager.buffers() if not buf.is_spilled
+        )
+
+        if self.lowest_utilization is None:
+            self.lowest_utilization = unspilled + nbytes
+        else:
+            self.lowest_utilization = min(
+                self.lowest_utilization, unspilled + nbytes
+            )
+
     def __repr__(self) -> str:
         return f"<SpillStatistics level={self.level}>"
 
@@ -186,7 +203,15 @@ class SpillStatistics:
                     ),
                     prefix=" " * 4,
                 )
-            return ret[:-1]  # Remove last `\n`
+            ret += "  RMM fragmentation (level >= 3): "
+            if self.level < 3:
+                return ret + "disabled"
+            if self.lowest_utilization is None:
+                ret += "None"
+            else:
+                ret += "\n    lowest utilization: "
+                ret += format_bytes(self.lowest_utilization)
+            return ret
 
 
 class SpillManager:
@@ -271,6 +296,7 @@ class SpillManager:
         In order to avoid deadlock, this function should not lock
         already locked buffers.
         """
+        self.statistics.log_spill_on_demand(manager=self, nbytes=nbytes)
 
         # Let's try to spill device memory
         spilled = self.spill_device_memory(nbytes=nbytes)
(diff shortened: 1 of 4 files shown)


wence- (commented):
Mostly some minor comments about explanation/docs. The implementation itself looks fine.

python/cudf/cudf/tests/test_spilling.py
```diff
@@ -594,3 +594,43 @@ def test_statistics_expose(manager: SpillManager):
     assert stat.count == 10
     assert stat.total_nbytes == buffers[0].nbytes * 10
     assert stat.spilled_nbytes == buffers[0].nbytes * 10
+
+
+def _statistics_rmm_fragmentation(size: int):
```
vyasr, line 599: Normally I like separate functions, but in this case the call is now decoupled from the original value of size and nothing prevents it from being called with a larger value. The logic here is tightly coupled to the logic of the test below. I would just inline it.

python/cudf/cudf/tests/test_spilling.py
```diff
@@ -594,3 +594,43 @@ def test_statistics_expose(manager: SpillManager):
...
+    b4 = as_buffer(data=rmm.DeviceBuffer(size=size), exposed=False)
+    assert manager.statistics.lowest_utilization == size * 3
+
+    # By freeing the first and last buffer, we now have fragmentation:
+    # total free memory is `size * 2` but fragmented into two segments.
+    # The result is a "lowest utilization" of `size * 2`.
+    del b1, b4
+    as_buffer(data=rmm.DeviceBuffer(size=size * 2), exposed=False)
+    assert manager.statistics.lowest_utilization == size * 2
+    del b2, b3
```
vyasr, line 616: OK, I am confused by a lot of the logic of this test. I think I am seriously misunderstanding the ordering in which buffers are being spilled. Here is the sequence that I expect to happen.
- We have a pool of size 3N
- We allocate b1, b2, and b3, which fill up the pool
- We allocate b4, which triggers an OOM-spill event. Here is my first point of confusion. We implement LRU logic to determine what buffer to spill, so I would have expected b1 to be spilled at this point and for the memory to be ordered b4-b2-b3. You are then deleting b1 and b4 and claiming that frees up two fragments, but by my rationale that should only free up one because b1 should already have been evicted from device. What am I missing?

This leads to two branches. 

Branch 1:
Let us assume that you are correct and that `del b1, b4` actually deletes two buffers that are currently on device. That implies that b2 or b3 was spilled. Since buffers are ordered, I must conclude that b3 as the _most recently_ created buffer was what was spilled instead of the least recently created one, so we must now have our memory ordered b1-b2-b4 (not b4-b2-b3 as I originally surmised). If that is true, then the next `as_buffer` call should trigger an eviction of b2 and b4, because those are now the most recently created buffers. But that would not lead to any fragmentation, so this can't be right either.

Branch 2:
If my logic is correct and b1 was already evicted, then we should have b4-b2-b3 in memory and the next spill event should evict b2 and b3, memory which is then filled exactly by the new allocation and the assertion should fail. Since it's not, this also cannot be what is happening.

I'm a bit at a loss to understand exactly what _is_ happening. However, assuming all of the `assert` statements pass and everything works as expected, this code exposes another potential weak point of using the current measurement to determine fragmentation. Assuming that all the eviction ordering is happening in a sensible way, at the end we are allocating a buffer that triggers an OOM-spill event, but we aren't saving it to any local value, so the end result after `as_buffer` so it should be deallocated immediately, and as a result we're actually using 2N less memory than `lowest_utilization` claims.

vyasr (changes requested):
I'm uneasy about the level of implementation detail that leaks into the tests. Not only are we encoding implementation details about how we spill (which is perhaps unavoidable and we already do that for other spilling tests), we are also making assumptions about implementation details in RMM. RMM doesn't really make any guarantees about exactly what a "pool" means, and composing its pooled allocator with a different upstream allocator could also lead to different results.

I am also not confident that we're measuring what we intend with this PR. If we really wanted to measure fragmentation accurately, I think we would want to do something closer to the overlapping buffer calculations that were done in early iterations of the spilling code. In its current state I think there are multiple ways in which the result could be misinterpreted and not really relate to fragmentation.

How much value do we gain from these statistics?

python/cudf/cudf/core/buffer/spill_manager.py
```diff
@@ -151,6 +155,30 @@ def log_expose(self, buf: SpillableBuffer) -> None:
...
+
+        unspilled = sum(
+            buf.size for buf in manager.buffers() if not buf.is_spilled
+        )
+        if self.lowest_utilization is None:
+            self.lowest_utilization = unspilled + nbytes
+        else:
+            self.lowest_utilization = min(
+                self.lowest_utilization, unspilled + nbytes
+            )
```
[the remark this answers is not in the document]
madsbk: You are both absolutely right. I have to think some more about this.  

I used the current implementation to investigate strange spilling behavior. It worked because most allocations were similar sized and not too large (~100MB). In some extreme cases, spilling-on-demand for a 100MB allocation would trigger at a device utilization of 25GB (on a 32GB GPU). This let me to the idea of tracking the device utilization but as you both points out, this is not a good measurement.

shwina:
Retargeted to 23.04

vyasr:
Given the concerns expressed with this implementation of statistics, I'm going to close this PR. @madsbk if you decide to revisit feel free to reopen, or open a new one with a different implementation. Happy to discuss further if you go the latter route!

PR #12334, opened 2022-12-07
Fix warnings in test_monotonic.py

vyasr:
## Description

One note with these deprecations. pandas has a special value used for parameters that are defaulted, and as a result explicitly passing None will throw a warning. We don't do this, hence the difference in the warnings contexts I use for our calls vs pandas (`pytest.warns` vs `expect_warning_if`). I didn't feel like this was worth commenting on in every place, but I can if reviewers want.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/index.py
+++ b/python/cudf/cudf/core/index.py
@@ -523,6 +523,12 @@ class RangeIndex(BaseIndex, BinaryOperand):
         int
             Index of label.
         """
+        if kind is not None:
+            warnings.warn(
+                "'kind' argument in get_slice_bound is deprecated and will be "
+                "removed in a future version.  Do not pass it.",
+                FutureWarning,
+            )
         if side not in {"left", "right"}:
             raise ValueError(f"Unrecognized side parameter: {side}")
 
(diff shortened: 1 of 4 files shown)


python/cudf/cudf/core/index.py
```diff
@@ -523,6 +523,12 @@ def get_slice_bound(self, label, side, kind=None):
         int
             Index of label.
         """
+        if kind is not None:
+            warnings.warn(
+                "'kind' argument in get_slice_bound is deprecated and will be "
+                "removed in a future version.  Do not pass it.",
```
bdice, line 529: I think "Do not pass it" is implied. It sounds kind of heavy handed so maybe we can remove that.
```suggestion
                "removed in a future version.",
```

python/cudf/cudf/core/index.py
```diff
@@ -1377,6 +1383,12 @@ def notna(self):
 
     @_cudf_nvtx_annotate
     def get_slice_bound(self, label, side, kind=None):
+        if kind is not None:
+            warnings.warn(
+                "'kind' argument in get_slice_bound is deprecated and will be "
+                "removed in a future version.  Do not pass it.",
```
bdice, line 1389:
```suggestion
                "removed in a future version.",
```

bdice (approved):
Minor suggestion, otherwise LGTM.

python/cudf/cudf/core/index.py
vyasr: I just blindly copied the pandas warning, but I agree better without.

vyasr:
@gpucibot merge

PR #10944, opened 2022-05-24
Enable Zstandard decompression only when all nvcomp integrations are enabled

vuule:
Closes #9055
Limiting due to known nvcomp issues.

PR #12338, opened 2022-12-08
Enable max compression ratio small block optimization for ZSTD

vuule:
## Description

Updates the ZSTD max compression ratio to 2^15 (from 2^11). Up to 128KB can be compressed into 4 bytes with ZSTD, hence 2^17 / 2^2 = 2^15.
Removes the previous workaround for the incorrect max ratio.

## Checklist


vuule:
test pending

python/cudf/cudf/tests/test_orc.py
```diff
@@ -1848,3 +1848,17 @@ def test_reader_empty_stripe(datadir, fname):
...
+    expected = cudf.DataFrame({"s": [0.0] * 8 * 1024})
+
+    buff = BytesIO()
+    try:
+        expected.to_orc(buff, compression="ZSTD")
+    except RuntimeError:
+        pytest.mark.xfail(reason="Newer nvCOMP version is required")
+    else:
+        got = cudf.read_orc(buff)
+        assert_eq(expected, got)
```
karthikeyann, line 1864: https://github.com/rapidsai/cudf/pull/12017#discussion_r1023823604

https://github.com/wence-/cudf/blob/0a2306034bb18a4b88489cc6aebfd7d630caad2e/docs/cudf/source/developer_guide/testing.md#conditional-xfails

vuule: I'm not sure how this applies here. Both outcomes are valid, at least until the minimum nvCOMP version is less than 2.4.

karthikeyann (approved):

python/cudf/cudf/tests/test_orc.py
shwina: Simply catching `RuntimeError` is not sufficient here because a `RuntimeError` can be raised due to a legitimate bug or error in the test code.

I do agree with Karthikeyan and @wence-'s comment that in many situations like this, we should be doing `pytest.raises()` instead of simply `xfail`ing.

I think what we want to do here is check that pytest raises an error if the nvCOMP version is less than 2.4.  That being said, doing a `raises()` check conditionally is annoying. @vyasr wrote a helper called `expect_warning_if` that conditionally does `pytest.warns()` (https://github.com/rapidsai/cudf/blob/branch-23.02/python/cudf/cudf/testing/_utils.py#L392-L393). Perhaps we should also have an `expect_error_if` utility and use that here?

This is all assuming we have a reliable way of checking the nvCOMP version. If we don't have that, I would be happy just with the following change:

```suggestion
    try:
        expected.to_orc(buff, compression="ZSTD")
    except RuntimeError as e:
        if "expected error message" in str(e):
            pytest.mark.xfail(reason="Newer nvCOMP version is required")
        raise  # reraises the RuntimeError
    else:
        got = cudf.read_orc(buff)
        assert_eq(expected, got)
```

hyperbolic2346 (approved):
cpp looks good

python/cudf/cudf/tests/test_orc.py
vuule: I was told that we should not be checking error messages:
https://github.com/rapidsai/cudf/pull/10944#pullrequestreview-983953876
@bdice what is the current guideline on this?

bdice: Libcudf error messages are not considered to be part of its stable API. https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#libcudf-does-not-promise-specific-exception-messages

wence-: Note that this line:

```
            pytest.mark.xfail(reason="Newer nvCOMP version is required")
```

Does not do what you want. I've updated my advice in #12244 to the following:

```python
def test_orc_reader_zstd_max_ratio(request):
    request.applymarker(pytest.mark.xfail(condition=nvCOMP.__version__ < 2.4, reason="nvCOMP >= 2.4 is required"))
    ... # rest of test
```

That assumes that it is possible to determine the version of nvcomp programmatically.

vuule: I don't think it's possible to get the version. I think it would require an API to get it from libcudf.

vuule: I should just make this a C++ test. Including the nvcomp_adapter header would allow me to check the nvCOMP version in the test.

vyasr: - An `expect_error_if` would be useful. There are some tests do things like `expectation = pytest.raises(...) if cond else contextlib.ExitStack()` which is roughly equivalent. Would be good to centralize this as we refactor tests, but appears that is no longer in scope in this place if the test is moving to C++.
- Re error messages, Bradley already pointed to the section the dev guide. More context in https://github.com/rapidsai/cudf/pull/12076. I plan to make the same change in Python soon.
- Lawrence's xfail_strict PR is almost ready and should both prevent the undesired usage of pytest.xfail and provide up-to-date guidance on best practices there.

wence-: Was the conclusion here that this test would be moved into C++ (and so this one will go away?). If we want to keep this an can't inspect the nvcomp version in Python then please use:

```python
def test_orc_reader_zstd_max_ratio(request):
    # Encodes as 64KB of zeros, which compresses to 18 bytes with ZSTD
    expected = cudf.DataFrame({"s": [0.0] * 8 * 1024})

    buff = BytesIO()
    try:
        expected.to_orc(buff, compression="ZSTD")
    except RuntimeError:
        request.applymarker(
            reason="Newer nvCOMP version is required for ZSTD compression"
        )
        raise
    else:
        got = cudf.read_orc(buff)
        assert_eq(expected, got)
```

It's a bit ugly because it mask all runtime errors as an expected failure, whereas really only a specific runtime failure (old nvcomp version) should be an expected failure

vuule: The only reason I want to move the test to C++ is to be able to check the nvCOMP version.
It does mean that I would need to include a /src header in the test; I'm not sure what is the less of two evils.
I'm fine with applying the suggestion in https://github.com/rapidsai/cudf/pull/12338#discussion_r1050754041 if there are no objections.

vuule: Moved the test to C++ and made it such that it has to completely pass if the right nvCOMP version is used.

vuule:
C++ reviewers, please take a look at the new C++ test. The rest of the PR has been reviewed and approved.

bdice (approved):
C++ test looks good to me.

vuule:
@gpucibot merge

PR #12345, opened 2022-12-08
Add size check to make_offsets_child_column

davidwendt:
## Description
Adds the `cudf::sizes_to_offsets` utility to the `cudf::strings::detail::make_offsets_child_column` utility. Strings APIs use this when `make_strings_children` is not practical. Also lists APIs use this for building offsets too.

Reference: https://github.com/rapidsai/cudf/issues/12167

## Checklist

### DIFF
--- a/cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
+++ b/cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
@@ -242,17 +242,18 @@ static sizes_to_offsets_iterator<ScanIterator, LastType> make_sizes_to_offsets_i
  *               "Size of output exceeds column size limit");
  * @endcode
  *
- * @tparam SizesIterator Iterator type for input and output of the scan using addition operation
+ * @tparam SizesIterator Iterator type for input of the scan using addition operation
+ * @tparam OffsetsIterator Iterator type for the output of the scan
  *
  * @param begin Input iterator for scan
  * @param end End of the input iterator
  * @param result Output iterator for scan result
  * @return The last element of the scan
  */
-template <typename SizesIterator>
+template <typename SizesIterator, typename OffsetsIterator>
 auto sizes_to_offsets(SizesIterator begin,
                       SizesIterator end,
-                      SizesIterator result,
+                      OffsetsIterator result,
                       rmm::cuda_stream_view stream)
 {
   using SizeType = typename thrust::iterator_traits<SizesIterator>::value_type;
--- a/cpp/include/cudf/strings/detail/strings_children.cuh
+++ b/cpp/include/cudf/strings/detail/strings_children.cuh
@@ -17,6 +17,7 @@
 
 #include <cudf/column/column.hpp>
 #include <cudf/column/column_factories.hpp>
+#include <cudf/detail/iterator.cuh>
 #include <cudf/detail/sizes_to_offsets_iterator.cuh>
 #include <cudf/strings/detail/utilities.hpp>
 #include <cudf/utilities/default_stream.hpp>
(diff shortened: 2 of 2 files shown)


cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
```diff
@@ -269,5 +273,55 @@ auto sizes_to_offsets(SizesIterator begin,
   return last_element.value(stream);
 }
 
+/**
```
davidwendt, line 276: The `make_offsets_child_column` is moved from `cudf::strings::detail` in `strings_children.cuh` to `cudf::detail` in this file. This utility is used for strings and lists functions.

cpp/include/cudf/strings/detail/strings_children.cuh
```diff
@@ -32,40 +33,6 @@
 namespace cudf {
 namespace strings {
 namespace detail {
-/**
```
davidwendt, line 35: This is moved to `sizes_to_offsets_iterator.cuh`

cpp/include/cudf/strings/detail/strings_column_factories.cuh
```diff
@@ -103,42 +94,43 @@ std::unique_ptr<column> make_strings_column(IndexPairIterator begin,
   auto null_mask =
     (null_count > 0) ? std::move(new_nulls.first) : rmm::device_buffer{0, stream, mr};
 
-  auto const avg_bytes_per_row = bytes / std::max(strings_count - null_count, 1);
   // build chars column
-  std::unique_ptr<column> chars_column = [&] {
```
davidwendt, line 108: This is just reformatted after changing the capture block.

cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
```diff
@@ -269,5 +273,55 @@ auto sizes_to_offsets(SizesIterator begin,
...
+ * This function sets the offsets values by executing scan over the sizes in the provided
+ * Iterator.
+ *
+ * The return also includes the total number of elements -- the last element value from the
+ * scan.
+ *
+ * @throw cudf::logic_error if the total size of the scan (last element) greater than maximum value
+ * of `size_type`
+ *
+ * @tparam Iterator Used as input to scan to set the offset values
```
bdice, line 288:
```suggestion
 * @tparam InputIterator Used as input to scan to set the offset values
```

cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
```diff
@@ -269,5 +273,55 @@ auto sizes_to_offsets(SizesIterator begin,
...
+    return idx < count ? static_cast<size_type>(begin[idx]) : size_type{0};
+  };
+  auto input_itr = cudf::detail::make_counting_transform_iterator(0, map_fn);
+  // Use the sizes-to-offsets iterator to compute the total number of elements
+  auto const total_elements = sizes_to_offsets(input_itr, input_itr + count + 1, d_offsets, stream);
+  CUDF_EXPECTS(
+    total_elements <= static_cast<decltype(total_elements)>(std::numeric_limits<size_type>::max()),
+    "Size of output exceeds column size limit");
+
+  offsets_column->set_null_count(0);
```
bdice, line 322: Is `null_count == 0` implied by the use of `mask_state::UNALLOCATED`?

cpp/include/cudf/lists/detail/scatter.cuh
```diff
@@ -117,8 +117,8 @@ std::unique_ptr<column> scatter_impl(
 
   auto list_size_begin = thrust::make_transform_iterator(
     target_vector.begin(), [] __device__(unbound_list_view l) { return l.size(); });
-  auto offsets_column = cudf::strings::detail::make_offsets_child_column(
-    list_size_begin, list_size_begin + target.size(), stream, mr);
+  auto offsets_column = std::get<0>(cudf::detail::make_offsets_child_column(
```
bdice, line 120: Question: Sometimes I have seen we use a structured binding with a throwaway like:
```suggestion
  auto [offsets_column, _] = cudf::detail::make_offsets_child_column(
```
I think we use this structured binding style in quite a few places but am uncertain about whether there is a strong argument for/against this pattern. Are you aware of an argument for/against?

bdice (approved):
One minor suggestion and two questions. Approving because none of my comments are blocking.

cpp/include/cudf/lists/detail/scatter.cuh
davidwendt: I saw a review recently that suggested using `get<n>` instead of the `_` in places where the parameter can indeed be discarded. My worry has always been that I would get an `unused variable warning` for `_` even though I've not seen that occur yet.
Also, I agree that it seems cleaner to me as in this review comment: https://github.com/rapidsai/cudf/pull/12239#discussion_r1050099794

cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
davidwendt: Wonder why this did not fail the doxygen check.

cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
davidwendt: Sorry, I'm not sure what you are asking. 
Calling `set_null_count(0)` here fixes the null-count value invalidated by the call to `mutable_view()` on `offsets_column`.

cpp/include/cudf/lists/detail/scatter.cuh
bdice: Sounds good!

cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
bdice: The column is initially created with an unallocated mask. Does creating a mutable view invalidate the null count and allocate a mask? Or is it undefined behavior to write to a mutable view’s null mask if it is unallocated? (I'd have to look back at the implementation.)

davidwendt: Yes, I suspect this will all be fixed once we get rid of `UNKNOWN_NULL_COUNT`.
https://github.com/rapidsai/cudf/blob/8a1ff584e39084b588b82a4d8a5f909bd7451916/cpp/src/column/column.cu#L130

ttnghia (approved):

cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
```diff
@@ -269,5 +273,55 @@ auto sizes_to_offsets(SizesIterator begin,
...
+ *
+ * @tparam InputIterator Used as input to scan to set the offset values
+ * @param begin The beginning of the input sequence
+ * @param end The end of the input sequence
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ * @return Offsets column and total elements
+ */
+template <typename InputIterator>
+std::pair<std::unique_ptr<column>, size_type> make_offsets_child_column(
```
ttnghia, line 296: Just one question: Why this is not in `strings::` namespace? This is

davidwendt:
Will merge this after https://github.com/rapidsai/cudf/pull/12434 is merged.

davidwendt:
@gpucibot merge

PR #12355, opened 2022-12-09
Fix warnings in test modules up to test_dataframe.py

vyasr:
## Description

I realized that my previous warning reduction PRs were causing some circular work where I would add a new warning to cudf to match pandas, which would cause those new warnings to appear in modules that I had previously declared free of warnings. To prevent this, I've changed my approach to instead go through the test modules in alphabetical order and ensure that they are all error free up to that point. This PR removes warnings from all test modules up to test_dataframe.py.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/column/categorical.py
+++ b/python/cudf/cudf/core/column/categorical.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import warnings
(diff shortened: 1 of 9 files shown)


bdice (approved):

vyasr:
@gpucibot merge

PR #12360, opened 2022-12-12
Add code to detect possible malformed page data in parquet files.

nvdbaranec:
Fixes:

https://github.com/NVIDIA/spark-rapids/issues/7213
https://github.com/NVIDIA/spark-rapids/issues/7228

This adds code to detect a subset of possible malformed parquet page data.  Specifically:  where the input file contains N rows, but the page data for some (non-list) columns contains a number of values != N.   This is a very lightweight check.

There is an associated PR for the spark plugin that should be merged immediately after this one (otherwise builds will fail) so I'm adding the Do Not Merge tag.


nvdbaranec:
rerun tests

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1259,6 +1259,90 @@ struct start_offset_output_iterator {
   }
 };
 
+struct flat_column_num_rows {
+  gpu::PageInfo const* pages;
+  gpu::ColumnChunkDesc const* chunks;
+
+  __device__ size_type operator()(size_type pindex)
```
ttnghia, line 1266: `const`

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1259,6 +1259,90 @@ struct start_offset_output_iterator {
...
+  }
+};
+
+struct row_counts_nonzero {
+  __device__ bool operator()(size_type count) { return count > 0; }
+};
+
+struct row_counts_different {
+  size_type const expected;
+  __device__ bool operator()(size_type count) { return (count != 0) && (count != expected); }
```
ttnghia, line 1284: All `operator()` should be `const`.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1259,6 +1259,90 @@ struct start_offset_output_iterator {
...
+    }
+    return page.num_rows;
+  }
+};
+
+struct row_counts_nonzero {
+  __device__ bool operator()(size_type count) { return count > 0; }
+};
+
+struct row_counts_different {
```
ttnghia, line 1282: I think this one is small enough to just be an embedded lambda.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1259,6 +1259,90 @@ struct start_offset_output_iterator {
...
+  rmm::device_uvector<int> row_counts(pages.size(), stream);  // worst case:  num keys == num pages
+  auto size_iter = thrust::make_transform_iterator(
+    page_index.begin(), flat_column_num_rows{pages.device_ptr(), chunks.device_ptr()});
+  auto row_counts_begin                     = row_counts.begin();
+  [[maybe_unused]] auto [_, row_counts_end] = thrust::reduce_by_key(rmm::exec_policy(stream),
+                                                                    page_keys.begin(),
+                                                                    page_keys.end(),
+                                                                    size_iter,
+                                                                    thrust::make_discard_iterator(),
+                                                                    row_counts_begin);
```
ttnghia, line 1321: This can just be `auto const row_count_end = thrust::reduce_by_key(...).second;`.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1259,6 +1259,90 @@ struct start_offset_output_iterator {
...
+void detect_malformed_pages(hostdevice_vector<gpu::PageInfo>& pages,
+                            hostdevice_vector<gpu::ColumnChunkDesc> const& chunks,
+                            device_span<const int> page_keys,
+                            device_span<const int> page_index,
+                            std::optional<size_t> expected_row_count,
+                            rmm::cuda_stream_view stream)
+{
+  // sum row counts for all non-dictionary, non-list columns. other columns will indicated as 0
+  rmm::device_uvector<int> row_counts(pages.size(), stream);  // worst case:  num keys == num pages
+  auto size_iter = thrust::make_transform_iterator(
```
ttnghia, line 1313: `const`.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1259,6 +1259,90 @@ struct start_offset_output_iterator {
...
+  auto row_counts_begin                     = row_counts.begin();
+  [[maybe_unused]] auto [_, row_counts_end] = thrust::reduce_by_key(rmm::exec_policy(stream),
+                                                                    page_keys.begin(),
+                                                                    page_keys.end(),
+                                                                    size_iter,
+                                                                    thrust::make_discard_iterator(),
+                                                                    row_counts_begin);
+
+  auto nonzero = thrust::find_if(
+    rmm::exec_policy(stream), row_counts_begin, row_counts_end, row_counts_nonzero{});
```
ttnghia, line 1324: `find_if` is known to have performance regression. Use `thrust::count_if(...) > 0` instead.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1269,6 +1353,55 @@ void reader::impl::preprocess_pages(size_t skip_rows,
...
+  //   rows, but the total number of rows in the pages for column X is != N. while it
+  //   is possible to load this by just capping the number of rows read, we cannot tell
+  //   which rows are invalid so we may be returning bad data. in addition, this mismatch
+  //   confuses the chunked reader
+  detect_malformed_pages(pages,
+                         chunks,
+                         page_keys,
+                         page_index,
+                         uses_custom_row_bounds ? std::nullopt : std::make_optional(num_rows),
+                         _stream);
```
ttnghia, line 1403: This seems expensive to me instead of "lightweight". Do you know at what scale this computation will be, comparing to the number of rows?

ttnghia: Ah, is this just moved up from below?

nvdbaranec: Correct.  The sorting of pages by schema is needed for everything.

cpp/src/io/parquet/reader_impl_preprocess.cu
nvdbaranec: One option is to do a `copy_if` (of nonzero values) to a new buffer and use the result of that to check the condition.  I avoided that initially because I have a pathological fear of memory allocation.  But maybe with the pool allocator it's no big deal.

Also worth pointing out:  the size of the input here is at the very most in the thousands.  Typically it will be much less.  So the overhead of `find_if` may not be so bad.

nvdbaranec: Actually, what I can do is just double the size of the row_counts allocation and copy_if to the second half of the vector.  Although that may be a little too cowboy for you guys :)

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1364,49 +1497,14 @@ void reader::impl::preprocess_pages(size_t skip_rows,
...
+    _chunk_itm_data.page_keys  = std::move(page_keys);
+    _chunk_itm_data.page_index = std::move(page_index);
 
     // retrieve pages back
     pages.device_to_host(_stream, true);
 
-#if defined(PREPROCESS_DEBUG)
-    print_pages(pages, _stream);
-#endif
+    // print_pages(pages, _stream);
```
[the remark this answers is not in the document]
nvdbaranec: When you enable the preprocessor defines, you don't necessarily want everything on at once.  You want to selectively add things in depending on what you're debugging.  Not sure how this initially got checked in with the #if.

cpp/src/io/parquet/reader_impl_preprocess.cu
vyasr: My inclination based on the size is that it's not going to matter a whole lot to the overall performance of this function which option we select. @ttnghia do you have a better recollection of exactly how bad the slowdown from `find_if` could be, and whether it's relevant at these data sizes? Without a strong performance argument I'm pretty indifferent to which solution we pick. I suspect that any performance gains/losses in this check (whether from an extra allocation, more kernel launches, or a slow algorithm) will probably be washed out by the rest of the overall parquet reader logic given the small size of the inputs to this function.

ttnghia: `thrust::find_if` is the building block of logical operators `thrust::any_of`/`thrust::any_of` etc. 
The issue is also a long time issue: https://github.com/NVIDIA/thrust/issues/1016.

cpp/src/io/parquet/reader_impl_preprocess.cu
nvdbaranec: I normally would, but the compiler gives you the extended-lambda warning/error so I have to make them functors :(

ttnghia (approved):

cpp/src/io/parquet/reader_impl_preprocess.cu
vyasr: Yes, I'm familiar with that issue, but the data sizes there are massive (billions of elements) whereas we are talking about thousands. As a result I suspect it won't actually matter here.

cpp/src/io/parquet/reader_impl_preprocess.cu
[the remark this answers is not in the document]
vyasr: I'm not really sure why you would even check in the commented version. If it's really debug prints why not just add them in when you need them?

vyasr (approved):

cpp/src/io/parquet/reader_impl_preprocess.cu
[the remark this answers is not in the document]
nvdbaranec: They're common debug points.  That is to say:  when debugging, these are the usual places you want to start gathering information.  Speeds up the process so you don't have to dig through the code every time.

hyperbolic2346 (approved):
Small nit related to nodiscard that could be sprinkled about. I do wonder about the performance impact this may have.

nvdbaranec:
@gpucibot merge

PR #12363, opened 2022-12-12
Allow casting from `UDFString` back to `StringView` to call methods in `strings_udf`

brandon-b-miller:
This PR adds some code to cast a `UDFString` to a `StringView` which unblocks UDFs that end up calling further transformations on strings that have already been returned by other functions. It works by registering a set of attributes to `UDFString` instances that mirror the ones attached to `StringView`, and introducing lowering that allows a cast. The cast ultimately calls a shim function which wraps the `cudf::string_view` casting operator of `udf_string`.


python/strings_udf/strings_udf/_typing.py
```diff
@@ -120,6 +119,18 @@ def prepare_args(self, ty, val, **kwargs):
 str_view_arg_handler = StrViewArgHandler()
 
 
+# for use in testing only
+def sv_to_udf_str(sv):
```
vyasr, line 123: If this is only for testing, why not define it in the tests module?

vyasr: I assume it's because of the registration below. I think that probably belongs in the test too, though.

python/cudf/cudf/core/udf/strings_lowering.py
```diff
@@ -36,6 +36,7 @@
 
 
 @cuda_lower(len, MaskedType(string_view))
+@cuda_lower(len, MaskedType(udf_string))
```
vyasr, line 39: Was adding the len overload necessarily in scope for this PR?

python/strings_udf/strings_udf/lowering.py
```diff
@@ -160,18 +185,27 @@ def cast_string_view_to_udf_string(context, builder, fromty, toty, val):
...
+    _create_string_view_from_udf_string(udf_str, sv)
+
+
 # String function implementations
 def call_len_string_view(st):
     return _string_view_len(st)
 
 
 @cuda_lower(len, string_view)
+@cuda_lower(len, udf_string)
```
vyasr, line 208: Same question as before for why we're adding the len overload as part of this PR. I'm guessing that somewhere we need it to determine how many bytes to copy or something?

python/strings_udf/strings_udf/lowering.py
```diff
@@ -462,56 +511,63 @@ def id_func_impl(context, builder, sig, args):
...
 
 
-@create_unary_identifier_func("StringView.istitle")
+@create_unary_identifier_func("istitle")
 def istitle_impl(st, tbl):
     return _string_view_istitle(st, tbl)
+
+
+@cuda_lower(sv_to_udf_str, string_view)
+def sv_to_udf_str_testing_lowering(context, builder, sig, args):
```
vyasr, line 570: This function should also be defined in the test suite.

python/strings_udf/strings_udf/tests/utils.py
```diff
@@ -0,0 +1,26 @@
...
+from numba.core.typing.templates import AbstractTemplate
+from numba.cuda.cudadecl import registry as cuda_decl_registry
+from numba.cuda.cudaimpl import lower as cuda_lower
+
+from strings_udf._typing import StringView, string_view, udf_string
+from strings_udf.lowering import cast_string_view_to_udf_string
+
+
+def sv_to_udf_str(sv):
+    pass
```
vyasr, line 12: It looks like this function is already defined here so I'm not sure why it's also needed in the main codebase. Maybe that can just be removed?

vyasr (changes requested):
Couple of minor requests, but looks fine overall.

wence- (changes requested):
I think I am confused here about the need for this code, and particularly which (if any) parts need to be exposed to authors of UDFs. It seems like none, but then how did we previously get in a situation where string views needed to be converted to udf strings (or vice versa)?

python/strings_udf/strings_udf/tests/utils.py
brandon-b-miller: This is a great question. It happens as follows. All strings in `strings_udf` start life as a `string_view`, because the input column isn't being mutated. `column_to_string_view_array` handles this. It stays that way all the way until it hits the users UDF. Imagine they have something like this:

```python
def f(some_string: string_view):
    return some_string.startswith('a')
```

Following the typing in this case, `some_string` is known to be of type `string_view`. When `startswith` is called, numba searches for a declaration of `StringView.startswith(StringView)` and finds the one that returns a `bool`. The lowering to actually do the work is inserted there and everything is fine. 

Now, suppose the user has something like

```python
def f(some_string: string_view):
    new_string = some_string.upper()
    return new_string.startswith('a')
```

Again `some_string` starts life as an instance of `StringView`, but that's where the similarities end. In this case, numba needs to figure out what type `new_string` is. Since functions that return strings need to be built around a `udf_string`, that's what `upper()` does. Numba will find the overload of `StringView.upper() -> udf_string` and determine that the type of `new_string` is now `udf_string`. 

This is where we hit the problem because there _is no method_ `UDFString.startswith`. In the c++ layer there isn't even a `startswith` method associated with `udf_string` objects because they're considered to be a read-only operation, just like there's no `upper()` method of `cudf::string_view`. 

What this PR is trying to do is turn one into the other when necessary. When a `string_view` tries to call a `udf_string` method, it needs to be cast, and the same way the other way around.

python/strings_udf/strings_udf/_typing.py
brandon-b-miller: The problem is that we need to call this function both in `test_string_udfs.py` (inside the `strings_udf` package) _and_ inside `test_udf_masked_ops.py` in cuDF. If I defined this inside `test_string_udfs.py`, I would need to make that testing file a module that `cudf` could import from, or redefine it inside of `cudf`. I wasn't sure what the lesser of all the evils is here.

brandon-b-miller: I think maybe having a separate module as part of `strings_udf` named `_testing` or something similar which contains this code and may be imported but isn't user facing could be a good middle ground here. Let me set it up that way and see how it looks.

python/cudf/cudf/core/udf/strings_lowering.py
brandon-b-miller: The other thread has more context as to why this is necessary, but this PR actually is adding _all_ the overloads for all of the methods of `string_view` that aren't already attached to `udf_string`. This one just looks weird because it's not an attribute so it's defined off by itself. 

The function of this overload is the following (similar to the others). Here I am dropping the `Masked` view of things for simplicity even though technically we're discussing masked values.

1. Numba sees a call to `len(something)` where the argument being passed (`something`) is a `udf_string`. 
2. All registered typing of `len` are checked to see if one emits a signature when the first argument is a `udf_string`. 
3. The [typing for len](https://github.com/rapidsai/cudf/blob/branch-23.02/python/strings_udf/strings_udf/_typing.py#L129) returns the signature  `len(string_view) -> int32` (this requires the small update in this PR as it's currently returning `len(udf_string) -> int32`
4. Because the signature specifies `string_view` as the argument rather than `udf_string`, the cast from `udf_string` to `string_view` is inserted into the lowering before the rest of the `len` code executes as normal. Note that the fact that this is the effect of the emitted signature expecting the _desired_ type rather than the _input_ type. If we leave it as `len(udf_string) -> int32` it won't insert the cast.

shwina:
Given the experimental status of string UDFs, and that planned code freeze is a few hours away, let’s go ahead and target 23.04 for this PR.

python/strings_udf/strings_udf/tests/test_string_udfs.py
```diff
@@ -39,14 +40,27 @@ def get_kernel(func, dtype, size):
...
-    return kernel
+    @cuda.jit(
+        sig, link=[strings_udf.ptxpath], extensions=[str_view_arg_handler]
+    )
+    def udf_string_kernel(input_strings, output_col):
+        # test the string function with a udf_string as input
+        id = cuda.grid(1)
+        if id < size:
+            st = input_strings[id]
+            st = sv_to_udf_str(st)
```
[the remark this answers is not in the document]
brandon-b-miller: There should only be one of these now and should never have been two (likely a copy paste error). I'm not sure how numba resolves two functions that have the same name, I don't _think_ it's by the name, my guess is that since both copies of it didn't do anything except `pass` that their bytecode or whatever numba uses to do its lookup was similar enough that it worked. 

In any case, there's only one now and it's defined in `_testing`.

brandon-b-miller:
Sorry all for the noise from the branch change of base here!

python/strings_udf/strings_udf/_typing.py
vyasr: Would this problem go away when #12669 is merged? If so, maybe we wait to finish that up?

brandon-b-miller:
This has been updated with the changes from https://github.com/rapidsai/cudf/pull/12669 and should be ready to go again. The code is largely the same as when we last looked at it, the difference mainly being the shuffling around of files.

vyasr (changes requested):

python/cudf/cudf/tests/test_string_udfs.py
```diff
@@ -21,10 +21,10 @@
...
 )
 from cudf.core.udf.utils import _PTX_FILE, _get_extensionty_size
-from cudf.testing._utils import assert_eq
+from cudf.testing._utils import assert_eq, sv_to_udf_str
 
 
-def get_kernel(func, dtype, size):
+def get_kernels(func, dtype, size):
     """
     Create a kernel for testing a single scalar string function
```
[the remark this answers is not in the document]
brandon-b-miller: Fixed

brandon-b-miller:
sorry I forgot to get to this last week @vyasr , I think I addressed your comments. I added some docs to `sv_to_udf_str` in https://github.com/rapidsai/cudf/pull/12363/commits/0840cdc118688599661480fdde2106d81e66e3cb, curious what you think as this piece of things can be a little hard to explain conceptually.

wence- (approved):
/merge

PR #12364, opened 2022-12-12
Use correct memory resource in io::make_column

### LINKED ISSUE #5823: [BUG] Incorrect memory resource used in io::make_column
`io::make_column` creates a column from a `column_buffer` object. In the case of string columns, new buffers are allocated. In other cases, the `column_buffer`'s buffer is moved into the new column object. Because of this, the user memory resource should only be used to `column_buffers` that get moved (not strings). Currently the CSV reader has a workaround for this:
```
      const bool is_final_allocation = column_types[active_col].id() != type_id::STRING;
      out_buffers.emplace_back(column_types[active_col],
                               num_records,
                               true,
                               stream,
                               is_final_allocation ? mr_ : rmm::mr::get_default_resource());
```
However, other readers are still misusing the user memory resource.

Proposed solution:
Remove the `stream` and `mr` parameters from `make_column` and store them in the `column_buffer` instead. Currently we would anyway incorrectly handle having a different `stream`/`mr` because of the buffer move.
Make the `make_column` a member of `column_buffer`.
In the constructor of `column_buffer`, use the right memory resource based on the type, and the implementation of `make_column`. This way the logic is limited to `column_buffer` and applies to all readers.

https://github.com/rapidsai/cudf/issues/5823#tasklist-block-178ed760-02da-4799-a48c-558bc497f464

vyasr:
## Description

When creating a column from an `io::detail::column_buffer` object there is some subtlety in what memory resource should be used. Due to the way cudf's string APIs are designed, string columns in IO initially allocate temporary memory on construction of the `column_buffer` and only later allocate the final buffer to be returned to the user, while for all other types the initial allocation can be directly moved to the output. As a result, using the user-provided memory resource in the constructor of a `column_buffer` is sufficient to guarantee correct behavior for all column types except string columns, which also need a memory resource to be provided in `make_column`. This PR fixes the issue by storing the memory resource provided upon `column_buffer` construction, converting `make_column` into a method of the class, and removing the `mr` parameter to the method. With the new code, the saved `mr` is reused later ensuring that a consistent memory resource is used everywhere that needs to allocate memory returned to the user (as opposed to temporary memory that should be using `rmm::mr::get_current_device_resource()`).

Resolves #5823 

## Checklist

### DIFF
--- a/cpp/src/io/utilities/column_buffer.cpp
+++ b/cpp/src/io/utilities/column_buffer.cpp
@@ -36,6 +36,12 @@ void column_buffer::create(size_type _size,
 
   switch (type.id()) {
     case type_id::STRING:
+      // The contents of _strings will never be directly returned to the user.
+      // Due to the fact that make_strings_column copies the input data to
+      // produce its outputs, _strings is actually a temporary. As a result, we
+      // do not pass the provided mr to the call to
+      // make_zeroed_device_uvector_async here and instead let it use the
+      // default rmm memory resource.
       _strings = std::make_unique<rmm::device_uvector<string_index_pair>>(
         cudf::detail::make_zeroed_device_uvector_async<string_index_pair>(size, stream));
       break;
@@ -98,6 +104,10 @@ std::unique_ptr<column> make_column(column_buffer& buffer,
           schema_info->children.push_back(column_name_info{"chars"});
         }
 
+        // make_strings_column allocates new memory, it does not simply move
+        // from the inputs, so we need to pass it the memory resource given to
+        // the buffer on construction so that the memory is allocated using the
+        // resource that the calling code expected.
         return make_strings_column(*buffer._strings, stream, buffer.mr);
       } else {
         // convert to binary
--- a/cpp/src/io/utilities/column_buffer.cpp
+++ b/cpp/src/io/utilities/column_buffer.cpp
@@ -53,6 +53,7 @@ void column_buffer::create(size_type _size,
     _null_mask =
       cudf::detail::create_null_mask(size, mask_state::ALL_NULL, rmm::cuda_stream_view(stream), mr);
   }
+  this->mr = mr;
 }
 
 namespace {
@@ -82,14 +83,10 @@ column_buffer column_buffer::empty_like(column_buffer const& input)
   return new_buff;
 }
 
-/**
- * @copydoc cudf::io::detail::make_column
- */
 std::unique_ptr<column> make_column(column_buffer& buffer,
                                     column_name_info* schema_info,
                                     std::optional<reader_column_schema> const& schema,
-                                    rmm::cuda_stream_view stream,
-                                    rmm::mr::device_memory_resource* mr)
+                                    rmm::cuda_stream_view stream)
 {
   if (schema_info != nullptr) { schema_info->name = buffer.name; }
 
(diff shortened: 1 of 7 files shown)


divyegala (approved):

cpp/src/io/utilities/column_buffer.cpp
```diff
@@ -53,6 +53,7 @@ void column_buffer::create(size_type _size,
     _null_mask =
       cudf::detail::create_null_mask(size, mask_state::ALL_NULL, rmm::cuda_stream_view(stream), mr);
   }
+  this->mr = mr;
```
ttnghia, line 56: I think this still needs more work: The line 39 above creates a buffer without using `mr`.

cpp/src/io/utilities/column_buffer.hpp
```diff
@@ -140,8 +142,7 @@ struct column_buffer {
 std::unique_ptr<column> make_column(column_buffer& buffer,
                                     column_name_info* schema_info,
                                     std::optional<reader_column_schema> const& schema,
-                                    rmm::cuda_stream_view stream,
-                                    rmm::mr::device_memory_resource* mr);
+                                    rmm::cuda_stream_view stream);
```
PointKernel, line 145: Not really matter but the doc for `mr` can be removed.

cpp/src/io/utilities/column_buffer.cpp
PointKernel: IIUC `_strings` is not part of the "output" data so it shouldn't use `mr`.

ttnghia: `_strings` may be part of the output:
```
make_column(...) {
...
  return make_strings_column(*buffer._strings, stream, mr);
}
```

ttnghia: Humnn, it seems that I was wrong. `make_strings_column` copies from `_strings` to the output.

PointKernel: `string_index_pair` is a pair of char pointer and size which are information used to construct the output data but they are not output data themselves.

vyasr: Correct, the fact that `make_strings_column` makes a copy is precisely why this code needs to be specialized in this way for strings. Since it makes a copy, strings end up being the only branch where the first allocation is a temporary. This is why strings here don't use `mr` in `create` but then need to use it later, whereas all other types move from the buffers constructed in `create` so those buffers need to use `mr`.

cpp/src/io/utilities/column_buffer.hpp
vyasr: Done.

cpp/src/io/utilities/column_buffer.cpp
ttnghia: Thanks all.
@vyasr Can you add comments clarifying this into the code for creating `_strings`? So other devs will figure this out easier next time.

vyasr: Good call, yes I can do that.

ttnghia (approved):

PointKernel (approved):
LGTM

Thanks!

vyasr:
@gpucibot merge

PR #10326, opened 2022-02-17
Consolidate C++ `conda` recipes and add `libcudf_tests` package

ajschmidt8:
This PR includes the following changes:

- Adds a `libcudf_tests` package to the `libcudf` recipe
  - This is a prerequisite for removing "Project Flash" from our build/CI scripts
  - The `libcudf_tests` package was added as an additional output to the existing `libcudf` recipe (which was renamed to `libcudf-split`)
- Consolidates remaining C++ recipes into `libcudf-split` recipe
  - This gets rid of a lot of duplicate code between the recipes and reduces the number of times we have to call `conda build` in our CI scripts

PR #12368, opened 2022-12-13
Update conda recipes.

bdice:
## Description
This updates cudf's conda recipes to align with changes being made in other packages as part of the GitHub Actions migration.

## Checklist


conda/recipes/libcudf/meta.yaml
```diff
@@ -60,7 +59,7 @@ outputs:
       build:
         - cmake {{ cmake_version }}
       run:
-        - cudatoolkit {{ cuda_spec }}
+        - {{ pin_compatible('cudatoolkit', max_pin='x', min_pin='x') }}
```
bdice, line 62: @ajschmidt8 I see you changed this to use `cuda_spec` in #10326 but I don't fully understand why `cuda_spec` was preferred over using `pin_compatible`. Any advice on this?

bdice: Discussed offline with @ajschmidt8. The `cuda_spec` approach where we manually specify the bounds is needed in the C++ recipes like `libcudf/meta.yaml` because there is no `host:` section that `pin_compatible` can use to determine the runtime dependency bounds. That's because we have a split recipe with a top-level `requirements:` section that applies to all the recipe outputs (to build exactly once and avoid building the C++ twice for `libcudf` and `libcudf-tests`, for example).

There are several recipes in RAPIDS that are using `cuda_spec` or similar constructions where the `pin_compatible` _can_ (and should) be used instead, but this is not one of them.

conda/recipes/cudf/meta.yaml
```diff
@@ -2,9 +2,9 @@
 
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
 {% set minor_version = version.split('.')[0] + '.' + version.split('.')[1] %}
-{% set py_version=environ.get('CONDA_PY', 38) %}
-{% set cuda_version='.'.join(environ.get('RAPIDS_CUDA_VERSION', '11.5.2').split('.')[:2]) %}
-{% set cuda_major=cuda_version.split('.')[0] %}
+{% set py_version = environ.get('CONDA_PY', 38) %}
```
ajschmidt8, line 5: should we also change this to be required? seems pretty disastrous to accidentally publish a conda package with the wrong Python version in the build string.

```suggestion
{% set py_version = environ['CONDA_PY'] %}
```

ajschmidt8: (also relevant to other files in this PR)

bdice: I’m not sure exactly how this works. I think this variable is provided by conda and not the environment. https://docs.conda.io/projects/conda-build/user-guide/environment-variables.html

I’m open to the change you proposed, I just don’t know if there are any catches we are unaware of.

ajschmidt8 (approved):

bdice:
@gpucibot merge

ajschmidt8:
@gpucibot merge

ajschmidt8:
no queue :neutral_face:

PR #12371, opened 2022-12-13
Nested JSON depth benchmark

karthikeyann:
## Description
Adds benchmark with depth axis for nested json.
It generates json with list of lists, and struct of structs for given depth, with types - int, float, string, bool, null literal.
The json string generation is in host (so the benchmark data generation will slow for higher depth value.

Depends on #12314
## Checklist


vuule:
Looks good! Will do a full review once https://github.com/rapidsai/cudf/pull/12314 is merged.

cpp/benchmarks/io/json/nested_json.cpp
```diff
@@ -46,17 +126,17 @@ auto make_test_json_data(size_type string_size, rmm::cuda_stream_view stream)
...
   auto d_scalar         = cudf::strings::repeat_string(d_string_scalar, repeat_times);
-  auto& d_input         = static_cast<cudf::scalar_type_t<std::string>&>(*d_scalar);
 
-  auto generated_json    = std::string(d_input);
-  generated_json.front() = '[';
-  generated_json.back()  = ']';
-  return generated_json;
+  auto data = const_cast<char*>(d_scalar->data());
+  cudaMemsetAsync(data, '[', 1, stream.value());
+  cudaMemsetAsync(data + d_scalar->size() - 1, ']', 1, stream.value());
```
ttnghia, line 137: Wait, you can call `cuda*` APIs in a `cpp` file?

karthikeyann: Yes. [C++ runtime APIs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html) and driver APIs can be called in any C++ files. We use runtime APIs almost always.

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cuda-runtime
Include `cuda.h` and  `cuda_runtime.h`, Link with libcudart for runtime for non-nvcc compilers. 
https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#driver-api for driver APIs.

vuule: My mental model is that you just cannot include/declare/call any `__device__` functions. Since any `__device__` code behind `cudaMemsetAsync` (if any) is in a separate compilation unit, it's fine to call it here. Templated stuff (e.g. Thrust) is usually a no-go, since device code in usually in the headers.
I'm sure the truth is more nuanced, this is just my functional understanding.

cpp/benchmarks/io/json/nested_json.cpp
```diff
@@ -24,15 +24,95 @@
...
+{
+  std::string s = "{";
+  std::vector<std::string> elems{
+    R"(1)", R"(-2)", R"(3.4)", R"("5")", R"("abcdefghij")", R"(true)", R"(null)"};
+  for (int i = 0; i < num_columns; i++) {
+    s += R"("col)" + std::to_string(i) + R"(": )";
+    if (i % 2 == 0) {
+      s += generate_struct_of_structs(max_depth - 2, (max_struct_size), elems[i % elems.size()]);
+    } else {
+      s += generate_list_of_lists(max_depth - 2, (max_struct_size), elems[i % elems.size()]);
```
vuule, line 88: typo?
```suggestion
      s += generate_list_of_lists(max_depth - 2, max_list_size, elems[i % elems.size()]);
```

cpp/benchmarks/io/json/nested_json.cpp
```diff
@@ -24,15 +24,95 @@
...
-#include <cstdlib>
+#include <string>
+#include <vector>
 
-namespace cudf {
 namespace {
-auto make_test_json_data(size_type string_size, rmm::cuda_stream_view stream)
+
+// List of List nested.
+std::string generate_list_of_lists(int32_t max_depth, int32_t max_rows, std::string elem)
```
vuule, line 38: AFAIK, appending strings like this is very slow, since you keep making copies of the old data every (or at least often) time you append another string. I recommend using [stringstream](https://en.cppreference.com/w/cpp/io/basic_stringstream) instead.
All functions except `generate_json` should modify passed `stringstream` and then finally `generate_json` can return a string using `stringstream_object.str()`.

vuule (changes requested):

vyasr (approved):
Looks solid to me. Most of my comments ended up being sort of tangential. I'll trust @vuule to follow up regarding the perf question on data generation, I don't think it's a big deal here either way.

cpp/benchmarks/io/json/nested_json.cpp
karthikeyann: Tried passing `string&` after `reserve` and use `append`. It's slower than simply using `+=`. (2^30 string_size took 1m23s compared to 1m16s).
I profiled the code and found that `std::to_string`, `operator+`, string ctor took most time in decreasing order.
I will make some adjustments to reduce benchmark data generation time.

karthikeyann:
Memoized row strings and pre-generated `std::to_string(num)`.

cpp/benchmarks/io/json/nested_json.cpp
vyasr: Wait, you found that the code was slower after reserving than before? Something sounds fishy there. I wouldn't expect meaningful differences between `+=` and `append`, but maybe there are tiny error checking/exception handling overheads in one but not the other. Preallocating should definitely be faster though unless something is very wrong.

In any case, not a huge deal for this use case. It sounds as if you were able to reduce benchmark generation time, which is nice and I'm happy with that result however you did it.

vuule: Very interesting finds. Maybe not critical for this code, but good to know in general :)
I can't rationalize why preallocation would be slower.
All in all, no concerns with data gen time in this PR :)

cpp/benchmarks/io/json/nested_json.cpp
```diff
@@ -24,15 +24,115 @@
...
+
+std::string generate_row(
+  int num_columns, int max_depth, int max_list_size, int max_struct_size, size_t max_bytes)
+{
+  std::string s = "{";
+  const std::vector<std::string> elems{
+    R"(1)", R"(-2)", R"(3.4)", R"("5")", R"("abcdefghij")", R"(true)", R"(null)"};
+  for (int i = 0; i < num_columns; i++) {
+    s += R"("col)" + num_to_string(i) + R"(": )";
+    if (auto it = _row_cache.find({i % 2, max_depth - 2, max_struct_size, i % elems.size()});
```
vuule, line 86: memoization could be a tad simpler
```suggestion
    if (auto it = _row_cache.find({i % 2, max_depth, max_struct_size, i % elems.size()});
```

cpp/benchmarks/io/json/nested_json.cpp
```diff
@@ -24,15 +24,115 @@
...
+        it != _row_cache.end()) {
+      s += it->second;
+    } else {
+      if (i % 2 == 0) {
+        auto r =
+          generate_struct_of_structs(max_depth - 2, max_struct_size, elems[i % elems.size()]);
+        _row_cache[{i % 2, max_depth - 2, max_struct_size, i % elems.size()}] = r;
```
vuule, line 93:
```suggestion
        _row_cache[{0, max_depth, max_struct_size, i % elems.size()}] = r;
```

cpp/benchmarks/io/json/nested_json.cpp
```diff
@@ -24,15 +24,115 @@
...
+        s += r;
+      } else {
+        auto r = generate_list_of_lists(max_depth - 2, max_struct_size, elems[i % elems.size()]);
+        _row_cache[{i % 2, max_depth - 2, max_struct_size, i % elems.size()}] = r;
```
vuule, line 97:
```suggestion
        _row_cache[{1, max_depth, max_struct_size, i % elems.size()}] = r;
```

cpp/benchmarks/io/json/nested_json.cpp
```diff
@@ -24,15 +24,115 @@
...
+        s += r;
+      }
```
vuule, line 99: or something like this
```suggestion
      auto r = (i % 2 == 0) ?
          generate_struct_of_structs(max_depth - 2, max_struct_size, elems[i % elems.size()]) : 
          generate_list_of_lists(max_depth - 2, max_struct_size, elems[i % elems.size()]);
      _row_cache[{i % 2, max_depth, max_struct_size, i % elems.size()}] = r;
      s += r;
```

vuule (commented):
cool optimization!
minor suggestions, looks good otherwise

cpp/benchmarks/io/json/nested_json.cpp
karthikeyann: used this suggestion.

vuule (approved):

cpp/benchmarks/io/json/nested_json.cpp
karthikeyann: Top level string pre-allocation helps reduce a few seconds. But passing that string as `string&` to all functions and using `append`/`+=` was slower. I am not sure what might be the issue. (I didn't try this again after adding row cache).

karthikeyann:
I will disable `64` depth since, parsing doesn't complete for minutes with current code. I will disable `32` too since it takes ~30s for single iteration on 1 GB text.

cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -38,29 +39,48 @@ size_t sources_size(host_span<std::unique_ptr<datasource>> const sources,
...
+        if (source->is_device_read_preferred(data_size)) {
+          bytes_read += source->device_read(range_offset, data_size, destination, stream);
+        } else {
+          auto const h_buffer = source->host_read(range_offset, data_size);
+          CUDF_CUDA_TRY(cudaMemcpyAsync(destination,
+                                        h_buffer->data(),
+                                        h_buffer->size(),
+                                        cudaMemcpyHostToDevice,
+                                        stream.value()));
+          stream.synchronize();
```
ttnghia, line 68: Can we sync stream at the end of the `sources` loop?

ttnghia: I'm also not sure if we even need any sync at all.

vuule: we need the sync because copy needs to finish before h_buffer goes out of scope.

ttnghia: Can we append `h_buffer` into a vector (created before the loop) and only sync after the loop? By doing so we can just sync once.

vuule: Just realized this is code from #12314 . I'll make the change there :D

vuule: done

ttnghia (approved):

PR #12374, opened 2022-12-13
Fix reductions any/all return value for empty input

### LINKED ISSUE #12373: [BUG] any and all reductions results are backwards on empty columns
**Describe the bug**
#12279 changed the `any` and `all` reductions to always return valid results, but the results are incorrect for empty column cases (and possibly cases where the only values present are nulls).  As described in #8757, `any` is essentially a search for `true` and should only return true if there's at least one `true` value in the input.  Returning `true` for an empty input is incorrect, it should return false.  This also matches the behavior of Python on an empty input, e.g.:
```python
>>> any([])
False
```

Similarly `all` is essentially a search for `false` and should only return false if there's at least one `false` value in the input.
```python
>>> all([])
True
```

The current behavior also does not match the results from `std::any_of` and `std::all_of`, e.g.:
```c++
#include <algorithm>
#include <iostream>
#include <vector>

int main() {
  std::vector<int> x;
  std::cout << "all: " << std::all_of(x.begin(), x.end(), [](int i){return i != 0;}) << std::endl;
  std::cout << "any: " << std::any_of(x.begin(), x.end(), [](int i){return i != 0;}) << std::endl;
  return 0;
}

// === OUTPUT ===
// all: 1
// any: 0
```

Example using Java cudf bindings in a Scala shell:
```java
scala> ai.rapids.cudf.ColumnVector.fromBoxedBooleans().all()
res0: ai.rapids.cudf.Scalar = Scalar{type=BOOL8 value=false} (ID: 5 7f07c43c68e0)

scala> ai.rapids.cudf.ColumnVector.fromBoxedBooleans().any()
res1: ai.rapids.cudf.Scalar = Scalar{type=BOOL8 value=true} (ID: 9 7f0728002cd0)
```

**Expected behavior**
`any` should return true if there is at least one row in the input column that is true, otherwise false.
`all` should return false if there is at least one row in the input column that is false, otherwise true.

davidwendt:
## Description
Fixes error introduced in #12279 to return the correct boolean value for reductions on empty input for `any(false)` and `all(true)`.
Also fixes incorrectly coded gtest to verify the correct result.
Closes #12373 

## Checklist

### DIFF
--- a/cpp/src/reductions/reductions.cpp
+++ b/cpp/src/reductions/reductions.cpp
@@ -179,7 +179,7 @@ std::unique_ptr<scalar> reduce(
     if (agg.kind == aggregation::ANY || agg.kind == aggregation::ALL) {
       // empty input should return true for ANY and return false for ALL
       dynamic_cast<numeric_scalar<bool>*>(result.get())
-        ->set_value(agg.kind == aggregation::ANY, stream);
+        ->set_value(agg.kind == aggregation::ALL, stream);
     }
     return result;
   }
--- a/cpp/src/reductions/reductions.cpp
+++ b/cpp/src/reductions/reductions.cpp
@@ -177,7 +177,7 @@ std::unique_ptr<scalar> reduce(
 
     auto result = make_default_constructed_scalar(output_dtype, stream, mr);
     if (agg.kind == aggregation::ANY || agg.kind == aggregation::ALL) {
-      // empty input should return true for ANY and return false for ALL
+      // empty input should return false for ANY and return true for ALL
       dynamic_cast<numeric_scalar<bool>*>(result.get())
         ->set_value(agg.kind == aggregation::ALL, stream);
     }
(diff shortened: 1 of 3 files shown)


cpp/src/reductions/reductions.cpp
```diff
@@ -179,7 +179,7 @@ std::unique_ptr<scalar> reduce(
     if (agg.kind == aggregation::ANY || agg.kind == aggregation::ALL) {
       // empty input should return true for ANY and return false for ALL
```
jlowe, line 180:
```suggestion
      // empty input should return false for ANY and return true for ALL
```

jlowe (approved):
Minor comment fix but otherwise lgtm.  Verified RAPIDS Accelerator for Apache Spark tests that were failing before this fix now are passing with the fix applied.

bdice (approved):

ttnghia (approved):

davidwendt:
@gpucibot merge

PR #12377, opened 2022-12-13
Pin arrow 9 in testing dependencies to prevent conda solve issues

vyasr:
## Description

This pinning is necessary because other testing dependencies require arrow (currently, transformers has a transitive dependency on it) and don't include a pinning as strict as ours, resulting in a newer arrow package (arrow 10) being installed. When we attempt to install our cudf/libcudf packages in CI, it should be able to downgrade to arrow 9, but the downgrade fails because changes in arrow's package structure in arrow 10 include the creation of a new libarrow package (arrow-cpp was renamed to libarrow) and libarrow has no equivalent in arrow 9. As a result, conda does not see a way to downgrade it, and it also does not see a way to satisfy its dependencies if the rest of the arrow packages are downgraded to arrow 9 (because libarrow is tied to arrow 10 packages), so the conda solve fails.

The current changeset will fix Github Actions CI. We may need additional fixes in our Jenkins CI.

## Checklist

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -104,11 +104,11 @@ function install_dask {
     set -x
     if [[ "${INSTALL_DASK_MAIN}" == 1 ]]; then
         gpuci_logger "gpuci_mamba_retry install -c dask/label/dev 'dask/label/dev::dask' 'dask/label/dev::distributed'"
-        gpuci_mamba_retry install -c dask/label/dev "dask/label/dev::dask" "dask/label/dev::distributed"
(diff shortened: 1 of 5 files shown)


bdice (approved):
Thanks for the fix. Note that this should be resolved by #12327.

vyasr:
@ajschmidt8 As expected, we'll need to do something similar to fix this for gpuCI (or we can simply disable gpuCI at this point!).

ajschmidt8 (approved):
LGTM. I added a quick workaround that should hopefully work for Jenkins. But yes, we will soon disable Jenkins at which point those scripts will be removed entirely.

ajschmidt8:
@gpucibot merge

ajschmidt8:
@gpucibot merge

PR #12390, opened 2022-12-14
Expose the RMM pool size in JNI

revans2:
## Description
This exposes the previously set pool size.

## Checklist


jlowe (commented):
This seems OK yet incomplete.  Wouldn't we want to get the allocation mode as well?  Seems a bit odd to expose pool size yet not even be able to tell if pooling is enabled (we don't force the pool size to be zero or -1 if pooling is not requested).

revans2:
> This seems OK yet incomplete. Wouldn't we want to get the allocation mode as well? Seems a bit odd to expose pool size yet not even be able to tell if pooling is enabled (we don't force the pool size to be zero or -1 if pooling is not requested).

Even if pooling is disabled the pool size is used as the maximum allocation size. Should I rename the method? Add more documentation to it?

revans2:
@jlowe please take another look

jlowe (approved):

java/src/main/java/ai/rapids/cudf/Rmm.java
```diff
@@ -119,6 +121,27 @@ public static synchronized void initialize(int allocationMode, LogConf logConf,
...
+  /**
+   * Get the most recently set pool size or -1 if RMM has not been initialized or pooling is
+   * not enabled.
+   */
+  public static synchronized long getPoolSize() {
+    if (isPoolingEnabled()) {
+      return poolSize;
+    } else {
+      return -1;
+    }
```
ttnghia, line 137: `poolSize` has default value `-1` right? So should we just return `poolSize`?

java/src/main/java/ai/rapids/cudf/Rmm.java
```diff
@@ -119,6 +121,27 @@ public static synchronized void initialize(int allocationMode, LogConf logConf,
...
+  }
+
+  /**
+   * Return true if rmm is initialized and pooling has been enabled, else false.
+   */
+  public static synchronized boolean isPoolingEnabled() {
+    return initialized && poolingEnabled;
```
ttnghia, line 144:
```suggestion
    return poolingEnabled;
```
`poolingEnabled == true` only after initialization so we don't need to check for `initialized`.

java/src/main/java/ai/rapids/cudf/Rmm.java
revans2: Not really. If pooling is disabled we would return whatever the pool size is that the user passed in. But it is simple to just clean that up when we set it, so I will do that.

revans2:
@ttnghia I think I have addressed all of your review comments.

ttnghia (approved):

ttnghia:
build

ttnghia:
@gpucibot merge

PR #12396, opened 2022-12-15
When building Arrow shared look for a shared OpenSSL

robertmaynard:
## Description
When asked to build arrow from source, state that we want a shared library version of OpenSSL instead of a static. That ensures that we use the OpenSSL provided by conda ( which only provides shared ).

Plus in general building against a static OpenSSL is a bad idea as it makes deploying SSL CVE fixes very hard. 

## Checklist

### DIFF
--- a/cpp/cmake/thirdparty/get_arrow.cmake
+++ b/cpp/cmake/thirdparty/get_arrow.cmake
@@ -126,13 +126,13 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
     # Turn off CPM using `find_package` so we always download and make sure we get proper static
     # library.
     set(CPM_DOWNLOAD_Arrow TRUE)
-    # We don't want to explicitly force a static OpenSSL when `BUILD_STATIC` so we allow the
-    # caller to still request a shared OpenSSL, therefore we don't modify `ARROW_OPENSSL_USE_SHARED`
+    # We don't want to explicitly force a static OpenSSL when `BUILD_STATIC` so we allow the caller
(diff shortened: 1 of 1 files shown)


cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -126,9 +126,14 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
     # Turn off CPM using `find_package` so we always download and make sure we get proper static
     # library.
     set(CPM_DOWNLOAD_Arrow TRUE)
+    # We don't want to explicitly force a static OpenSSL when `BUILD_STATIC` so we allow the
```
bdice, line 129: This feels like it’s phrased a little backwards. Maybe explain the default behavior and how users can change it. The comment below this one follows that approach and makes more sense for a first-time reader in my opinion. For example:

By default ARROW will try to search for a static version of OpenSSL which is a bad idea given that shared linking is advised for critical components like SSL. If a static build is requested, we honor ARROW's default of static linking, but users may consider setting `ARROW_OPENSSL_USE_SHARED` even in static builds.

bdice (approved):
LGTM with a proposed comment edit

bdice:
Clarifying: did you mean “shared”? From the context I expected the other word.

> That ensures that we use the OpenSSL provided by conda ( which only provides **static** ).

robertmaynard:
> Clarifying: did you mean “shared”? From the context I expected the other word.
> 
> > That ensures that we use the OpenSSL provided by conda ( which only provides **static** ).

Whoops, you are correct. Fixed.

vyasr (approved):

vyasr:
@gpucibot merge

robertmaynard:
@gpucibot merge

PR #12399, opened 2022-12-16
Use `numpy.empty()` instead of `bytearray` to allocate host memory for spilling

madsbk:
A `bytearray` is zero-initialized using `calloc`, which we don't need.  Additionally, `numpy.empty` both skips the zero-initialization and uses hugepages when available <https://github.com/numpy/numpy/pull/14216>.


vyasr (approved):
I suspect that there are places in our code that use e.g. numpy.zeros that could also be improved in this way, but that's out of scope for this PR.

madsbk:
rerun tests

jakirkham (approved):
Looks great! Thanks Mads 😄

Is there anywhere else we should be doing this?

madsbk:
Thanks @jakirkham and @vyasr, I think it can be merged

PR #12425, opened 2022-12-20
JNI bindings to write CSV

mythrocks:
## Description
This change adds JNI bindings to write tables out as CSV, to either the filesystem or memory.

The Java `Table` class now has additional methods:
1. `Table.writeCSVToFile()`: Writes the current table out to the specified file on the filesystem.
2. `Table.writeCSVToBuffer()`: Writes the current table out to a `HostBufferConsumer`.

These calls are analogous to `cudf::io::write_csv()`.

### Current limitations:
1. The `cudf::io::csv_writer_options` interface binds the CSV options tightly to the `Table` being written. This makes it a little clumsy to write multiple `Table`s to the same `HostBufferConsumer`, because each could be written with different, contradictory options.
2. `cudf::io::write_csv(file_name)` overwrites the specified file, if it exists. There currently isn't a way to keep a file open, and write multiple tables to it; each write call overwrites the previous file.

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
+++ b/java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
@@ -1,0 +1,122 @@
+/*
+ *
+ *  Copyright (c) 2022, NVIDIA CORPORATION.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package ai.rapids.cudf;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class CSVWriterOptions {
+  
+  // private final Table table;
+  private String[] columnNames;
+  private Boolean includeHeader = false;
+  private String rowDelimiter = "\n";
+  private byte fieldDelimiter = ',';
+  private String nullValue = "\\N";
+
+  private CSVWriterOptions(Builder builder) {
+    // this.table = builder.table;
+    this.columnNames = builder.columnNames.toArray(new String[builder.columnNames.size()]);
+    this.nullValue = builder.nullValue;
+    this.includeHeader = builder.includeHeader;
+    this.fieldDelimiter = builder.fieldDelimiter;
+    this.rowDelimiter = builder.rowDelimiter;
+  }
+
+  // public Table getTable() {
+  //   return table;
+  // }
+
+  public String[] getColumnNames() {
+    return columnNames;
+  }
+
+  public Boolean getIncludeHeader() {
+    return includeHeader;
+  }
+
+  public String getRowDelimiter() {
+    return rowDelimiter;
+  }
+
+  public byte getFieldDelimiter() {
+    return fieldDelimiter;
+  }
+
+  public String getNullValue() {
+    return nullValue;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+
+    // private Table table;
+    private List<String> columnNames = Collections.emptyList();
+    private Boolean includeHeader = false;
+    private String rowDelimiter = "\n";
+    private byte fieldDelimiter = ',';
+    private String nullValue = "\\N";
+    
+    public CSVWriterOptions build() {
+      return new CSVWriterOptions(this);
+    }
+
+    // public Builder withTable(Table table) {
+    //   this.table = table;
+    //   return this;
+    // }
+
+    public Builder withColumnNames(List<String> columnNames) {
+      this.columnNames = columnNames;
+      return this;
+    }
+
+    public Builder withColumnNames(String... columnNames) {
+      List<String> columnNamesList = new ArrayList<>();
+      for (String columnName : columnNames) {
+        columnNamesList.add(columnName);
+      }
+      return withColumnNames(columnNamesList);
+    }
+
+    public Builder withIncludeHeader(Boolean includeHeader) {
+      this.includeHeader = includeHeader;
+      return this;
+    }
+
+    public Builder withRowDelimiter(String rowDelimiter) {
+      this.rowDelimiter = rowDelimiter;
+      return this;
+    }
+
+    public Builder withFieldDelimiter(byte fieldDelimiter) {
+      this.fieldDelimiter = fieldDelimiter;
+      return this;
+    }
+
+    public Builder withNullValue(String nullValue) {
+      this.nullValue = nullValue;
+      return this;
+    }
+  }
+}
--- a/java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
+++ b/java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
@@ -24,7 +24,6 @@ import java.util.List;
 
 public class CSVWriterOptions {
   
-  // private final Table table;
   private String[] columnNames;
   private Boolean includeHeader = false;
   private String rowDelimiter = "\n";
@@ -32,17 +31,12 @@ public class CSVWriterOptions {
   private String nullValue = "\\N";
 
   private CSVWriterOptions(Builder builder) {
-    // this.table = builder.table;
     this.columnNames = builder.columnNames.toArray(new String[builder.columnNames.size()]);
     this.nullValue = builder.nullValue;
     this.includeHeader = builder.includeHeader;
     this.fieldDelimiter = builder.fieldDelimiter;
     this.rowDelimiter = builder.rowDelimiter;
   }
-
-  // public Table getTable() {
-  //   return table;
-  // }
 
   public String[] getColumnNames() {
     return columnNames;
@@ -70,7 +64,6 @@ public class CSVWriterOptions {
 
   public static class Builder {
 
-    // private Table table;
     private List<String> columnNames = Collections.emptyList();
     private Boolean includeHeader = false;
     private String rowDelimiter = "\n";
@@ -80,11 +73,6 @@ public class CSVWriterOptions {
     public CSVWriterOptions build() {
       return new CSVWriterOptions(this);
     }
-
-    // public Builder withTable(Table table) {
-    //   this.table = table;
-    //   return this;
-    // }
 
     public Builder withColumnNames(List<String> columnNames) {
       this.columnNames = columnNames;
--- a/java/src/main/java/ai/rapids/cudf/Table.java
+++ b/java/src/main/java/ai/rapids/cudf/Table.java
@@ -874,7 +874,6 @@ public final class Table implements AutoCloseable {
 
   public void writeCSVToFile(CSVWriterOptions options, String outputPath)
   {
-    testBoolNative(options.getIncludeHeader());
     writeCSVToFile(nativeHandle, 
                    options.getColumnNames(), 
                    options.getIncludeHeader(), 
@@ -882,6 +881,26 @@ public final class Table implements AutoCloseable {
                    options.getFieldDelimiter(), 
                    options.getNullValue(), 
                    outputPath);
+  }
+
+  private static native void writeCSVToBuffer(long table,
+                                              String[] columnNames,
+                                              boolean includeHeader,
+                                              String rowDelimiter,
+                                              byte fieldDelimiter,
+                                              String nullValue,
+                                              HostBufferConsumer buffer) throws CudfException;
+
+  public void writeCSVToBuffer(CSVWriterOptions options, HostBufferConsumer bufferConsumer)
+  {
+    testBoolNative(options.getIncludeHeader());
+    writeCSVToBuffer(nativeHandle, 
+                    options.getColumnNames(), 
+                    options.getIncludeHeader(), 
+                    options.getRowDelimiter(), 
+                    options.getFieldDelimiter(), 
+                    options.getNullValue(), 
+                    bufferConsumer);
   }
 
   /**
--- a/java/src/main/java/ai/rapids/cudf/Table.java
+++ b/java/src/main/java/ai/rapids/cudf/Table.java
@@ -857,6 +857,33 @@ public final class Table implements AutoCloseable {
         opts.getFalseValues()));
   }
 
+  private static native void testBoolNative(boolean boo) throws CudfException;
+
+  public void testBool(boolean boo)
+  {
+    Table.testBoolNative(boo);
+  }
+
+  private static native void writeCSVToFile(long table,
+                                            String[] columnNames,
+                                            boolean includeHeader,
+                                            String rowDelimiter,
+                                            byte fieldDelimiter,
+                                            String nullValue,
+                                            String outputPath) throws CudfException;
+
+  public void writeCSVToFile(CSVWriterOptions options, String outputPath)
+  {
+    testBoolNative(options.getIncludeHeader());
+    writeCSVToFile(nativeHandle, 
+                   options.getColumnNames(), 
+                   options.getIncludeHeader(), 
+                   options.getRowDelimiter(), 
+                   options.getFieldDelimiter(), 
+                   options.getNullValue(), 
+                   outputPath);
+  }
+
   /**
    * Read a JSON file using the default JSONOptions.
    * @param schema the schema of the file.  You may use Schema.INFERRED to infer the schema.
--- a/java/src/main/java/ai/rapids/cudf/Table.java
+++ b/java/src/main/java/ai/rapids/cudf/Table.java
@@ -857,13 +857,6 @@ public final class Table implements AutoCloseable {
         opts.getFalseValues()));
   }
 
-  private static native void testBoolNative(boolean boo) throws CudfException;
-
-  public void testBool(boolean boo)
-  {
-    Table.testBoolNative(boo);
-  }
-
   private static native void writeCSVToFile(long table,
                                             String[] columnNames,
                                             boolean includeHeader,
--- a/java/src/main/java/ai/rapids/cudf/Table.java
+++ b/java/src/main/java/ai/rapids/cudf/Table.java
@@ -893,7 +893,6 @@ public final class Table implements AutoCloseable {
 
   public void writeCSVToBuffer(CSVWriterOptions options, HostBufferConsumer bufferConsumer)
   {
-    testBoolNative(options.getIncludeHeader());
     writeCSVToBuffer(nativeHandle, 
                     options.getColumnNames(), 
                     options.getIncludeHeader(), 
--- a/java/src/main/native/src/TableJni.cpp
+++ b/java/src/main/native/src/TableJni.cpp
@@ -1323,6 +1323,53 @@ JNIEXPORT jlongArray JNICALL Java_ai_rapids_cudf_Table_readCSV(
   CATCH_STD(env, NULL);
 }
 
+JNIEXPORT void JNICALL Java_ai_rapids_cudf_Table_testBoolNative(JNIEnv *env, jclass, jboolean boo)
+{
+  std::cout << "CALEB: Native: boo == " <<  static_cast<int>(boo) << std::endl; 
+  std::cout << "CALEB: Native: As bool: boo == " << std::boolalpha << static_cast<bool>(boo) << std::endl; 
+}
+
+JNIEXPORT void JNICALL Java_ai_rapids_cudf_Table_writeCSVToFile(JNIEnv *env,
+                                                                jclass,
+                                                                jlong j_table_handle,
+                                                                jobjectArray j_column_names,
+                                                                jboolean include_header,
+                                                                jstring j_row_delimiter,
+                                                                jbyte j_field_delimiter,
+                                                                jstring j_null_value,
+                                                                jstring j_output_path) 
+{
+  JNI_NULL_CHECK(env, j_table_handle, "table handle cannot be null.", );
+  JNI_NULL_CHECK(env, j_column_names, "column name array cannot be null", );
+  JNI_NULL_CHECK(env, j_row_delimiter, "row delimiter cannot be null", );
+  JNI_NULL_CHECK(env, j_field_delimiter, "field delimiter cannot be null", );
+  JNI_NULL_CHECK(env, j_null_value, "null representation string cannot be itself null", );
+  JNI_NULL_CHECK(env, j_output_path, "output path cannot be null", );
+
+  try {
+    cudf::jni::auto_set_device(env);
+
+    auto const native_output_path = cudf::jni::native_jstring{env, j_output_path};
+    auto const output_path = native_output_path.get();
+
+    auto const table = reinterpret_cast<cudf::table_view *>(j_table_handle);
+    auto const n_column_names = cudf::jni::native_jstringArray{env, j_column_names};
+    auto const column_names = n_column_names.as_cpp_vector();
+
+    auto const line_terminator = cudf::jni::native_jstring{env, j_row_delimiter};
+    auto const na_rep = cudf::jni::native_jstring{env, j_null_value};
+    auto options = cudf::io::csv_writer_options::builder(cudf::io::sink_info{output_path}, *table)
+                        .names(column_names)
+                        .include_header(static_cast<bool>(include_header))
+                        .line_terminator(line_terminator.get())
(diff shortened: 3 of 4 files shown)


mythrocks:
> This makes it a little clumsy to write multiple Tables to the same HostBufferConsumer

I'm working on a workaround for this, now. Should have it shortly.

revans2:
> > This makes it a little clumsy to write multiple Tables to the same HostBufferConsumer
> 
> I'm working on a workaround for this, now. Should have it shortly.

Be very careful when using the `HostBufferConsumer` because we don't want to write out the header multiple times. Also the others have checks in the C++ code when doing chunked writes that the column types/etc match between multiple invocations. IF we are going to create our own chunked writer API, we should add in our own corresponding checks too.

mythrocks:
> > > This makes it a little clumsy to write multiple Tables to the same HostBufferConsumer
> 
> > 
> 
> > I'm working on a workaround for this, now. Should have it shortly.
> 
> 
> 
> Be very careful when using the `HostBufferConsumer` because we don't want to write out the header multiple times. Also the others have checks in the C++ code when doing chunked writes that the column types/etc match between multiple invocations. IF we are going to create our own chunked writer API, we should add in our own corresponding checks too.

Yes, sir. That's what prompted me to write a chunked writer. It will reside in JNI for now. I can see about moving it to `cpp/` afterwards, if there's interest.

ajschmidt8:
The forked branch for this PR did not have the GH Actions CI files included because it was outdated, so I merged the latest changes into this branch.

mythrocks:
Sorry I missed your update, @ajschmidt8.

This new update might be out of date as well. I'll upmerge and push very shortly.

java/src/test/java/ai/rapids/cudf/TableTest.java
```diff
@@ -575,6 +574,150 @@ void testReadCSV() {
...
+      // Read back.
+      CSVOptions readOptions = CSVOptions.builder()
+                                         .includeColumn("i")
+                                         .includeColumn("f")
+                                         .includeColumn("b")
+                                         .includeColumn("str")
+                                         .hasHeader(false)
+                                         .withDelim('\u0001')
+                                         .build();
+      try (Table readTable = Table.readCSV(schema, readOptions, outputFile)) {
```
razajafri, line 610: Curious to know what your thoughts are about using a 3rd party reader to read the file and then compare the output as well

java/src/main/native/src/TableJni.cpp
```diff
@@ -1,4 +1,5 @@
 /*
+
```
razajafri, line 2: unnecessary

java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
```diff
@@ -0,0 +1,110 @@
...
+ *
+ */
+
+package ai.rapids.cudf;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class CSVWriterOptions {
```
razajafri, line 25: If this writer will be supporting nested types, which I assume it will at some point, does it not make sense to extend the `ColumnWriterOptions` just like we do for `ParquetWriterOptions` and `OrcWriterOptions`?

java/src/test/java/ai/rapids/cudf/TableTest.java
mythrocks: At the moment, I'm relying on Spark providing third party verification. :]

It might be good to have that extra verification, but it comes at the cost of an additional dependency.

What are the readers for the other formats (ORC/Parquet) using?

java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
mythrocks: I'll see about extending `ColumnWriterOptions`, but I don't currently see CSV supporting nested types.

mythrocks: Hmm. The more I look at this, the more it looks like this doesn't fit.

`CSVWriterOptions` are the options for writing the entire table to CSV, while `ColumnWriterOptions` seems to be per-column. `StructColumnWriterOptions` seems like a closer fit.
Then, again, extending `StructColumnWriterOptions` opens the door to users calling `withStructColumn`, `withListColumn`, etc. We would then be trading compile-time errors with runtime assertions.

I am inclined to revisit this when nested columns are attempted for CSV.

mythrocks:
Added @vuule for the change in `csv.hpp`, to reset the `table_view`.

java/src/main/native/src/jni_writer_data_sink.hpp
```diff
@@ -0,0 +1,177 @@
...
+      current_buffer_written += amount_to_copy;
+      total_written += amount_to_copy;
+      left_to_copy -= amount_to_copy;
+    }
+    stream.synchronize();
+  }
+
+  std::future<void> device_write_async(void const *gpu_data, size_t size,
+                                       rmm::cuda_stream_view stream) override {
+    // Call the sync version until figuring out how to write asynchronously.
```
hyperbolic2346, line 121: I would have expected that the sync version called the async version and then synchronized. Is there anything else needed here to make it async outside of calling `cudaMemcpyAsync`? If the call to write takes a long time to loop through the buffers, I could see spawning a thread to do the work. In that case I would think a common function that could be called and then sync'd or passed to the `std::async` function.

java/src/main/native/src/jni_writer_data_sink.hpp
```diff
@@ -0,0 +1,177 @@
...
+      current_buffer_written = 0;
+    }
+  }
+
+  size_t bytes_written() override { return total_written; }
+
+  void set_alloc_size(long size) { this->alloc_size = size; }
+
+private:
+  void rotate_buffer(JNIEnv *env) {
```
hyperbolic2346, line 145: Not a huge fan of this name. It implies a double-buffering scenario to me.

hyperbolic2346 (changes requested):
Some questions about some of this. I think this looks reasonable though.

java/src/main/native/src/jni_writer_data_sink.hpp
```diff
@@ -0,0 +1,177 @@
...
+      copy_from = copy_from + amount_to_copy;
+      current_buffer_written += amount_to_copy;
+      total_written += amount_to_copy;
+      left_to_copy -= amount_to_copy;
+    }
+  }
+
+  bool supports_device_write() const override { return true; }
+
+  void device_write(void const *gpu_data, size_t size, rmm::cuda_stream_view stream) override {
```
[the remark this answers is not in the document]
mythrocks: This class is unchanged. I've moved it to a common header, for access from elsewhere. 

That's a good point, though. I'll take that refactor in a separate issue.

java/src/main/native/src/jni_writer_data_sink.hpp
mythrocks: Agreed, there. It took me a while to grok the intention of that code.

java/src/main/native/src/csv_chunked_writer.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cassert>
+
+#include <cudf/io/csv.hpp>
+
+#include "jni_writer_data_sink.hpp"
```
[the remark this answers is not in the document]
mythrocks: Thank you pointing that out. Fixed.

java/src/main/native/src/csv_chunked_writer.hpp
```diff
@@ -0,0 +1,78 @@
...
+    CUDF_EXPECTS(sink_info.num_sinks() == 1, "csv_chunked_writer should have exactly one sink.");
+    CUDF_EXPECTS(sink_info.user_sinks()[0] == _sink.get(), "Sink mismatch.");
+  }
+
+  void write(cudf::table_view const &table) {
+    if (_first_write_completed) {
+      _options.enable_include_header(false); // Don't write header after the first write.
+    } else {
+      _first_write_completed = true;
+    }
```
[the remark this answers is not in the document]
mythrocks: Agreed. I was torn about setting it every time, when there are multiple writes.

But then, if the `write()` throws, then the state would have already been set. It wouldn't be a problem for `spark-rapids` (which would not commit the write), but it might be problematic for other callers.

Agreed. Fixed.

java/src/main/native/src/jni_writer_data_sink.hpp
```diff
@@ -0,0 +1,177 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cudf/io/data_sink.hpp>
+
+#include "cudf_jni_apis.hpp"
+#include "jni_utils.hpp"
```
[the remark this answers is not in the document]
mythrocks: Fixed.

java/src/main/native/src/jni_writer_data_sink.hpp
mythrocks: > Is there anything else needed here to make it async outside of calling cudaMemcpyAsync?

Based on the signature, I *think* the `async`ness is not cudaMemcpy async-ness, but `std::thread` async-ness.
It's probably valid for the `async` function to call the sync version via `std::async`. I'll add this to the follow-up issue.

java/src/main/native/src/jni_writer_data_sink.hpp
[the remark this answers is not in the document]
mythrocks: Filed #12456. I did not want a bug introduced in this (unrelated) refactor to cause this change to be reverted.

java/src/main/native/src/jni_writer_data_sink.hpp
mythrocks: #12456.

cpp/include/cudf/io/csv.hpp
```diff
@@ -1332,7 +1332,7 @@ class csv_writer_options {
   size_type _rows_per_chunk = std::numeric_limits<size_type>::max();
   // character to use for separating lines (default "\n")
   std::string _line_terminator = "\n";
-  // character to use for separating lines (default "\n")
+  // character to use for separating column values (default ",")
```
ttnghia, line 1335: Does `column values` mean `rows` from the input column?

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -857,6 +857,84 @@ public static Table readCSV(Schema schema, CSVOptions opts, HostMemoryBuffer buf
...
+                                            boolean includeHeader,
+                                            String rowDelimiter,
+                                            byte fieldDelimiter,
+                                            String nullValue,
+                                            String trueValue,
+                                            String falseValue,
+                                            String outputPath) throws CudfException;
+
+  public void writeCSVToFile(CSVWriterOptions options, String outputPath)
+  {
```
ttnghia, line 871: I assume that you didn't run formatting on this file.
```suggestion
  public void writeCSVToFile(CSVWriterOptions options, String outputPath) {
```

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -857,6 +857,84 @@ public static Table readCSV(Schema schema, CSVOptions opts, HostMemoryBuffer buf
...
+    @Override
+    public void write(Table table) {
+      if (writerHandle == 0) {
+        throw new IllegalStateException("Writer was already closed");
+      }
+      writeCSVChunkToBuffer(writerHandle, table.nativeHandle);
+    }
+
+    @Override
+    public void close() throws CudfException {
```
ttnghia, line 921: If this overrides `close`, should it extend `Autoclosable`?

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -857,6 +857,84 @@ public static Table readCSV(Schema schema, CSVOptions opts, HostMemoryBuffer buf
...
+      }
+      if (consumer != null) {
+        consumer.done();
+        consumer = null;
+      }
+    }
+  }
+
+  public static TableWriter getCSVBufferWriter(CSVWriterOptions options, HostBufferConsumer bufferConsumer)
+  {
```
ttnghia, line 934:
```suggestion
  public static TableWriter getCSVBufferWriter(CSVWriterOptions options, HostBufferConsumer bufferConsumer) {
```

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -857,6 +857,84 @@ public static Table readCSV(Schema schema, CSVOptions opts, HostMemoryBuffer buf
...
+                                                   String rowDelimiter,
+                                                   byte fieldDelimiter,
+                                                   String nullValue,
+                                                   String trueValue,
+                                                   String falseValue,
+                                                   HostBufferConsumer buffer) throws CudfException;
+
+  private static native void writeCSVChunkToBuffer(long writerHandle, long tableHandle);
+
+  private static native void writeCSVToBufferEnd(long writerHandle);
```
ttnghia, line 894: I think the names `write...ToBufferBegin` and `...ToBufferEnd` are incorrect. They say "writing to the beginning and the end positions of the buffer" instead, which is not what they are doing. So we should probably call `startWrite...` and `finishWrite...` or something similar.

java/src/main/native/src/TableJni.cpp
```diff
@@ -1323,6 +1175,118 @@ JNIEXPORT jlongArray JNICALL Java_ai_rapids_cudf_Table_readCSV(
...
+  CATCH_STD(env, );
+}
+
+JNIEXPORT void JNICALL Java_ai_rapids_cudf_Table_writeCSVToBufferEnd(JNIEnv *env, jclass,
+                                                                     jlong j_writer_handle) {
+  JNI_NULL_CHECK(env, j_writer_handle, "writer handle cannot be null.", );
+
+  using cudf::jni::io::csv_chunked_writer;
+  auto writer =
+      std::unique_ptr<csv_chunked_writer>{reinterpret_cast<csv_chunked_writer *>(j_writer_handle)};
```
ttnghia, line 1281: The `writer` in line 1280 can just be a pointer similar to line 1266.

java/src/main/native/src/jni_writer_data_sink.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+#include <cudf/io/data_sink.hpp>
+
+namespace cudf::jni {
+
+constexpr long MINIMUM_WRITE_BUFFER_SIZE = 10 * 1024 * 1024; // 10 MB
+
+class jni_writer_data_sink final : public cudf::io::data_sink {
```
ttnghia, line 27: The class name doesn't need to have `jni_` prefix.

ttnghia: ....because it is already in `cudf::jni::` namespace.

java/src/main/native/src/csv_chunked_writer.hpp
```diff
@@ -0,0 +1,76 @@
...
+ */
+#pragma once
+
+#include "jni_writer_data_sink.hpp"
+
+#include <cudf/io/csv.hpp>
+
+#include <cassert>
+
+namespace cudf::jni::io {
```
ttnghia, line 24: Please be consistent with the namespace. If we are going to use `...::io` then it should be added anywhere relevant.

cpp/include/cudf/io/csv.hpp
mythrocks: There is no input column here.

Each line in the input CSV file corresponds to a row. The commas (or delimiters) separate the fields, i.e. values of each column.

java/src/main/java/ai/rapids/cudf/Table.java
mythrocks: Oh. What formatter are you running on Java files?

java/src/main/java/ai/rapids/cudf/Table.java
mythrocks: Please clarify the question. `CSVTableWriter` extends `AutoCloseable` via `TableWriter`.

java/src/main/java/ai/rapids/cudf/Table.java
mythrocks: This naming follows the convention set with the Parquet/ORC writers. E.g. `writeORCBufferBegin`, `writeORCBufferEnd`.

But yes, the convention is clumsy. I'll rename these. `startWrite...` sounds much better.

java/src/main/native/src/TableJni.cpp
mythrocks: It can't, actually. The writer needs to be deleted, because it's allocated on heap.

java/src/main/native/src/jni_writer_data_sink.hpp
mythrocks: `jni_writer_data_sink` is a preexisting class. It's just been moved here. I'd rather we didn't change its name.
(Removing `jni_` would make this the `writer_data_sink`, which doesn't seem very specific. This is the `data_sink` that's used from the JNI layer.)

java/src/main/native/src/csv_chunked_writer.hpp
mythrocks: I would like for the IO classes to be moved to the `cudf::jni::io` namespace. There might be value in moving `jni_writer_data_sink` here as well.

mythrocks: I'll add this to the followup issue #12456 I'm working on.

mythrocks:
P.S. I appreciate the review, @ttnghia. I had removed you from this one, since we had 3 reviewers already.

java/src/main/native/src/jni_writer_data_sink.hpp
mythrocks: I'll see about renaming this class in #12456.

java/src/main/java/ai/rapids/cudf/Table.java
mythrocks: Renamed to `beginWriteCSVToBuffer` and `endWriteCSVToBuffer`.

java/src/main/java/ai/rapids/cudf/Table.java
mythrocks: TIL that I should be using `pre-commit run --all-files` instead of the `clang-format` script.

java/src/main/native/src/csv_chunked_writer.hpp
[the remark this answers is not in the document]
mythrocks: I'm afraid I've had to undo these changes. This fails the formatting used by `pre-commit`, which then fails the style check.

The formatter reorders the includes in the reverse order.

java/src/main/native/src/jni_writer_data_sink.hpp
hyperbolic2346: I was under the impression that this was new code. I'm perfectly fine with filing an issue even if the issue number is suspect.

hyperbolic2346 (approved):

java/src/main/native/src/jni_writer_data_sink.hpp
mythrocks: #12458 addresses the rename.

java/src/main/native/src/csv_chunked_writer.hpp
mythrocks: #12458 addresses the use of `cudf::jni::io`.

mythrocks:
Re: the CI test failures:
```
/opt/conda/bin/git -c protocol.version=2 fetch --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/*
  Error: fatal: unable to access 'https://github.com/rapidsai/cudf/': The requested URL returned error: 429
  The process '/opt/conda/bin/git' failed with exit code 128
```
Accessing the CUDF repo on GitHub hit rate limiting. :]

mythrocks:
Rerun tests

cpp/include/cudf/io/csv.hpp
vuule, line 1335: There's a copy-paste error in all _inter_column_delimiter functions in this file. Would you mind correcting them too, since you started here? :grin:

cpp/include/cudf/io/csv.hpp
```diff
@@ -1498,6 +1498,13 @@ class csv_writer_options {
    * @param val String to represent values == 0 in INT8 types
    */
   void set_false_value(std::string val) { _false_value = val; }
+
+  /**
+   * @brief (Re)sets the table being written.
+   *
+   * @param table Table to be written
+   */
+  void set_table(table_view const& table) { _table = table; }
```
vuule, line 1507: Looks good. As mentioned offline, we might want to look into separating the sink, input table and writer options to facilitate easier reuse of options. Not in scope for this PR.

vuule (approved):

ttnghia:
Gently remind that the changed files need to be updated copyright year ;)

mythrocks:
> Gently remind that the changed files need to be updated copyright year ;)

:)) Good catch. 
I thought I was safe because the commits are from last year, but we're doing squashed commits. I'll update.

cpp/include/cudf/io/csv.hpp
mythrocks: Updated.

ttnghia (approved):

mythrocks:
> Conflicting files
java/src/main/native/src/TableJni.cpp

I'm on it now.

razajafri (approved):

mythrocks:
Waiting on https://github.com/rapidsai/rapids-cmake/pull/337 to re-kick the build.

java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
```diff
@@ -0,0 +1,134 @@
...
+
+  private String[] columnNames;
+  private Boolean includeHeader = false;
+  private String rowDelimiter = "\n";
+  private byte fieldDelimiter = ',';
+  private String nullValue = "\\N";
```
revans2, line 31: Most CSV implementations have empty string be the default for nulls, even CUDF has it that way. If we want this for CSV we really should be trying to match CUDF and not hive text file format.

java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
```diff
@@ -0,0 +1,134 @@
...
+    return new Builder();
+  }
+
+  public static class Builder {
+
+    private List<String> columnNames = Collections.emptyList();
+    private Boolean includeHeader = false;
+    private String rowDelimiter = "\n";
+    private byte fieldDelimiter = ',';
+    private String nullValue = "\\N";
```
revans2, line 83: Same here. Null should default to empty string to match CUDF's C++ code.

java/src/main/native/src/csv_chunked_writer.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <cudf/io/csv.hpp>
+
+#include "jni_writer_data_sink.hpp"
+
+namespace cudf::jni::io {
+
+/**
+ * @brief Class to write multiple Tables into the jni_writer_data_sink.
+ *
+ * TODO: Consider moving to cpp/ in the future, if there is interest.
```
revans2, line 29: Do we need a follow on issue for this todo?  Should we just remove it instead?

revans2 (commented):
Most of the changes look fine to me. The one issue that I have is that for the hive text file format the cudf CSV write is not going to be compatible. By default hive does not do any special escapes for string values. CUDF does escaping.

https://github.com/rapidsai/cudf/blob/8d1e2ccaccce86fcc9c9e71012137960c205bcf9/cpp/src/io/csv/writer_impl.cu#L75

Should we spend some time to make this optional? Or am I missing what this is for?

mythrocks:
> CUDF does escaping.

Thank you for catching that. I will try tackle that in a follow-up issue, specifically to make the escaping optional.

java/src/main/native/src/csv_chunked_writer.hpp
mythrocks: Removed. No evidence of interest, currently.

java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
mythrocks: Changed.

java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
mythrocks: Changed.

revans2 (approved):

mythrocks:
I've merged this. Thank you to all who reviewed.
A couple of follow-up PRs will, well, follow.

PR #12426, opened 2022-12-20
Replace message parsing with throwing more specific exceptions

### LINKED ISSUE #10632: [FEA] Remove Parsing of C++ Exception String
Currently, in certain places in cuDF Python we parse the string messages returned from C++ [`std::exception::what`](https://en.cppreference.com/w/cpp/error/exception/what). This should be removed and replaced by returning a more appropriate exception that Cython can automatically handle. If there is no "appropriate" exception, then a custom exception should be created and the Cython machinery should be written to handle it.

An [example of this](https://github.com/rapidsai/cudf/blob/97ac349ec2e962a951a95a8e0d7f9fa0e80da88d/python/cudf/cudf/core/column/column.py#L586) is the following:

```py
        except RuntimeError as e:
            if "out of bounds" in str(e):
                raise IndexError(
                    f"index out of bounds for column of size {len(self)}"
                ) from e
            raise
```
where from the C++ code we could throw a [`std::out_of_range`](https://en.cppreference.com/w/cpp/error/out_of_range) exception which would automatically be promoted to an IndexError.

This is a first step of the initial work detailed in https://github.com/rapidsai/cudf/issues/10200.

vyasr:
## Description

This PR identifies places where cuDF Python is parsing exception messages from libcudf in order to throw an appropriate Python exception and modifies the associated libcudf error macros (`CUDF_EXPECTS` or `CUDF_FAIL`) to throw a more descriptive exception. In order to fully support this behavior with arbitrary libcudf exceptions, this PR also introduces a custom Cython exception handler that can be extended to catch and handle any new type of exception thrown by libcudf. To cases where issues with the dtype of an argument is treated as a TypeError rather than a ValueError in pandas, a new exception type `dtype_error : public logic_error` is defined and mapped to a TypeError in Python. This PR does not aim to exhaustively improve the choice of thrown exceptions throughout libcudf, only in places that are immediately relevant to removing message parsing in cudf Python.

Resolves #10632 

## Checklist


vyasr:
The longer term plan here is to update all of libcudf's exception handling to throw something more useful than a `cudf::logic_error`. I'd like to use this PR as a test case for that to sort through some questions. There are a few issues that need to be discussed to move this work forward:
- @jlowe @revans2 I'm not sure how changes like this impact Java. My understanding is that all exceptions get mapped to [CudfException](https://github.com/rapidsai/cudf/blob/branch-23.02/java/src/main/java/ai/rapids/cudf/CudfException.java) or [CudaException](https://github.com/rapidsai/cudf/blob/branch-23.02/java/src/main/java/ai/rapids/cudf/CudaException.java) or so according to the logic defined in [jni_utils.hpp](https://github.com/rapidsai/cudf/blob/branch-23.02/java/src/main/native/include/jni_utils.hpp). Based on what I see there, any exception that is not one of the "known types" in `CATCH_STD` (basically an rmm OOM or a CUDA error) gets mapped to a `CudfException`. If that is true, then AFAICT changing the exception types in C++ would have no effect at all on the JNI in its current state because even any `std::*` exception would just get mapped to a `CudfException`. Is that correct? 
- @jlowe @revans2 Aside from making sure that this change is non-breaking for the JNI, are there ways in which this change to libcudf could be more useful in the Java plugin?
- @shwina @jrhemstad currently this PR is incomplete and does not remove all cases where cuDF Python is parsing the exception message. Many of the remaining examples are in [lists.py](https://github.com/rapidsai/cudf/blob/branch-23.02/python/cudf/cudf/core/column/lists.py). The problem is that there are cases where the [Cython/C++ exception mapping](http://docs.cython.org/en/latest/src/userguide/wrapping_CPlusPlus.html#exceptions) is not just incomplete, but rather disagrees with what pandas (and therefore cudf) wants to throw. The most common example is that there are places where pandas throws a TypeError based on the dtype of an input. [Here is a case where libcudf is checking the input parameters for some equality condition](https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/src/lists/contains.cu#L312). I would expect this to throw a `std::invalid_argument`. However, in Python a `std::invalid_argument` is translated to a ValueError, whereas [what cudf actually wants to throw here is a TypeError](https://github.com/rapidsai/cudf/blob/branch-23.02/python/cudf/cudf/core/column/lists.py#L408). I am not sure what the best way to handle this kind of issue is. If there is only one way that a particular exception can be thrown by a libcudf API, then it's not a problem; we can just catch a ValueError and reraise it as a TypeError. The much more troublesome case is when there are multiple parameters, each of which could be invalid, so a `std::invalid_exception` from libcudf could be coming from any of them. We could try defining additional exception types in libcudf that are more specific in order to differentiate these cases, but I suspect that some of these case are not fundamentally different, they're just viewed as different by the pandas API. As a result, we would be leaking pandas details into the types defined in libcudf.

revans2:
For the Java side of things I see the general pattern to be that each time you add in a new exception type in C++, it would be good to add a corresponding subclass of CudfException, and set up a mapping between the two in CATCH_STD.  This would maintain backwards compatibility with all existing code, but allow us to add in much more specific exception types.

If you want to you can also keep the code as is for now and we can handle adding in the mappings we want to add in. The important part for us is to be able to distinguish between different types of errors so we can start to understand if there are better ways to recover from the error besides just crashing and letting Spark retry it.

shwina:
> As a result, we would be leaking pandas details into the types defined in libcudf. Any thoughts on the best way to handle this?

It sounds like inevitably, there's going to be cases where libcudf will throw the wrong kind of error. I'd say that it's a success if we're able to get rid of exception message parsing for the most part, while still having a few stragglers.

shwina:
Retargeted to 23.04

vyasr:
@shwina it turns out I was wrong. At least based on the current usage, there is a fairly specific class of cases where I was expecting to throw a `std::invalid_argument` but pandas wants to see a `TypeError`: cases where the validity issue is specifically due to the data type of the input. To account for this case, I've added a new `dtype_error` in libcudf that captures this more precisely, and I introduced a custom exception handler in our Cython layer to handle it.

vyasr:
@revans2 I introduced one new exception type in this PR and I added a suitable additional mapping for it in `CATCH_STD`. Let me know if there is anything additional that you would like to see.

vyasr:
For Python reviewers: once everyone has approved, I plan to run a batch sed update to change `except +` to `except +cudf_exception_handler` throughout our Cython code, and I will update our [developer documentation for exception handling](https://github.com/rapidsai/cudf/blob/branch-23.04/docs/cudf/source/developer_guide/contributing_guide.md#handling-libcudf-exceptions) accordingly.

revans2 (approved):

python/cudf/cudf/_lib/cpp/exception_handler.hpp
```diff
@@ -0,0 +1,80 @@
+/*
```
shwina, line 1: The `cudf/_lib/cpp` directory contains the Cython definitions ("declarations") for `libcudf` types and functions.

Perhaps we can place this (and the corresponding `.pxd` file) in `cudf/_lib/exception_handler.hpp` (and `cudf/_lib/exception_handler.pxd`) instead?

Alternately, I'd also be fine with just inlining this bit of C++ in `cudf/_lib_exception_handler.pxd`

vyasr: Yeah, I wasn't sure what the place was for this given the C++. Personally I prefer avoiding inline C++ in pxd files, but I can go that route if you like it. IMO it will be familiar to an even smaller subset of devs, and no editors are capable of syntactically or semantically parsing it as anything other than a comment. Moving it to `_lib` is also fine with me; it seemed somewhat appropriate to put in `cpp` since `exception_handler.pxd` is now also just a file of Cython "declarations" rather than new definitions, but on further consideration I think `_lib` is more appropriate too.

shwina: OK let's move them both to `_lib` (and keep them separate)

shwina (approved):

mroeschke (approved):

cpp/include/cudf/lists/combine.hpp
```diff
@@ -80,8 +80,8 @@ std::unique_ptr<column> concatenate_rows(
  * r is [ {1, 2, 3, 4, 5}, {6, 7, 8, 9} ]
  * @endcode
  *
- * @throws cudf::logic_error if the input column is not at least two-level depth lists column (i.e.,
- *         each row must be a list of list).
+ * @throws std::invalid_argument if the input column is not at least two-level depth lists column
+ * (i.e., each row must be a list of list).
```
ttnghia, line 84:
```suggestion
 *        (i.e., each row must be a list of list).
```

cpp/include/cudf/lists/contains.hpp
```diff
@@ -163,7 +163,7 @@ std::unique_ptr<column> index_of(
  * @return std::unique_ptr<column> INT32 column of `n` rows with the location of the `search_key`
  *
  * @throw cudf::logic_error If `search_keys` does not match `lists` in its number of rows
- * @throw cudf::logic_error If `search_keys` type does not match the element type in `lists`
+ * @throw cudf::dtype_error If `search_keys` type does not match the element type in `lists`
```
ttnghia, line 166: Please also remove the line below: it is no longer relevant.

ttnghia: In addition, there is one more overload of `index_of` in this file that needs to be updated.

cpp/include/cudf/lists/gather.hpp
```diff
@@ -43,7 +43,7 @@ namespace lists {
  * @endcode
  *
  * @throws cudf::logic_error if `gather_map_list` size is not same as `source_column` size.
- * @throws cudf::logic_error if gather_map contains null values.
+ * @throws std::invalid_argument if gather_map contains null values.
```
ttnghia, line 46: Should the line below be updated with `dtype_error` (and the corresponding code)?

java/src/main/native/include/jni_utils.hpp
```diff
@@ -861,6 +862,9 @@ inline void jni_cuda_check(JNIEnv *const env, cudaError_t cuda_status) {
   catch (const cudf::cuda_error &e) {                                                              \
     JNI_CHECK_CUDA_ERROR(env, cudf::jni::CUDA_ERROR_CLASS, e, ret_val);                            \
   }                                                                                                \
+  catch (const cudf::dtype_error &e) {                                                             \
+    JNI_CHECK_THROW_NEW(env, cudf::jni::CUDF_DTYPE_ERROR_CLASS, e.what(), ret_val);                \
```
ttnghia, line 866: Should be just `e` instead of `e.what()`? As the catch above did that.

vyasr:
@davidwendt I requested your review here since you had thoughts in the last meeting, let me know what you think of the changes in this PR. As you requested, I can rename `dtype_error` to `data_type_error` or so.

cpp/tests/strings/json_tests.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include <cudf_test/base_fixture.hpp>
 #include <cudf_test/column_wrapper.hpp>
+#include <stdexcept>
```
davidwendt, line 24:
```suggestion

#include <stdexcept>
```

cpp/tests/lists/combine/concatenate_list_elements_tests.cpp
```diff
@@ -20,6 +20,7 @@
 #include <cudf_test/type_lists.hpp>
 
 #include <cudf/lists/combine.hpp>
+#include <stdexcept>
```
davidwendt, line 23:
```suggestion

#include <stdexcept>
```

cpp/tests/copying/segmented_gather_list_tests.cpp
```diff
@@ -25,6 +25,7 @@
 #include <cudf/detail/null_mask.hpp>
 #include <cudf/lists/gather.hpp>
 #include <cudf/lists/lists_column_view.hpp>
+#include <stdexcept>
```
davidwendt, line 28:
```suggestion

#include <stdexcept>
```

cpp/tests/copying/concatenate_tests.cu
```diff
@@ -31,6 +31,7 @@
 
 #include <rmm/exec_policy.hpp>
 
+#include <stdexcept>
```
davidwendt, line 34: This include should move down with the other STL includes.

cpp/include/cudf/utilities/error.hpp
```diff
@@ -86,6 +86,29 @@ struct fatal_cuda_error : public cuda_error {
 };
 /** @} */
```
davidwendt, line 87: The `/** @} */` should go after the struct `};` below.

cpp/include/cudf/utilities/error.hpp
```diff
@@ -86,6 +86,29 @@ struct fatal_cuda_error : public cuda_error {
...
+struct dtype_error : public std::invalid_argument {
+  /**
+   * @brief Constructs a dtype_error with the error message.
+   *
+   * @param message Message to be associated with the exception
+   */
+  dtype_error(char const* const message) : std::invalid_argument(message) {}
+
+  /**
+   * @brief Construct a new logic error object with error message
```
davidwendt, line 105:
```suggestion
   * @brief Construct a new data_type_error object with error message
```

davidwendt (changes requested):
I think `data_type_error` instead of `dtype_error` would fit better with libcudf naming.

cpp/include/cudf/lists/gather.hpp
vyasr: No, because nullability is not a property of a data type; a column of any data type could contain nulls. The new exception is specifically designed for cases when an operation is attempted on a data type that doesn't make sense, e.g. trying to compute the sqrt of a string column.

cpp/include/cudf/lists/contains.hpp
vyasr: Is the line below no longer relevant because of changes in this PR, or because prior work converted this function to use the experimental comparators?

vyasr: I updated the other overload as you suggested, thanks for catching that.

python/cudf/cudf/_lib/cpp/exception_handler.hpp
vyasr: I've made this change, but one thing to note is that it does make for some slightly awkward circularity. We are now importing the handler from `_lib` into all the pxd files in `_lib/cpp` (because the latter is where we actually expose C++ APIs that need to be decorated with +cudf_exception_handler). Do you still like that outcome?

cpp/include/cudf/utilities/error.hpp
```diff
@@ -84,6 +84,29 @@ struct cuda_error : public std::runtime_error {
 struct fatal_cuda_error : public cuda_error {
   using cuda_error::cuda_error;  // Inherit constructors
 };
+
+/**
+ * @brief Exception thrown when an operation is attempted on an unsupported dtype.
+ *
+ * This exception should be thrown when an operation is attempted on an
+ * unsupported dtype. This exception should not be thrown directly and is
```
davidwendt, line 92:
```suggestion
 * unsupported data_type. This exception should not be thrown directly and is
```

davidwendt (approved):

python/cudf/cudf/_lib/cpp/exception_handler.hpp
shwina: I think that's fine

cpp/include/cudf/lists/contains.hpp
```diff
@@ -129,6 +129,7 @@ enum class duplicate_find_option : int32_t {
  * @throw cudf::logic_error If `search_key` type does not match the element type in `lists`
  * @throw cudf::logic_error If `search_key` is of a nested type, or `lists` contains nested
  * elements (LIST, STRUCT)
+ * @throw cudf::data_type_error If `search_keys` type does not match the element type in `lists`
```
ttnghia, line 132: Please remove from line 129 to line 131.
Lines 130-131 are left from previous changes (now all overloads of `index_of` support nested types).

cpp/include/cudf/lists/contains.hpp
```diff
@@ -163,8 +164,8 @@ std::unique_ptr<column> index_of(
  * @return std::unique_ptr<column> INT32 column of `n` rows with the location of the `search_key`
  *
  * @throw cudf::logic_error If `search_keys` does not match `lists` in its number of rows
- * @throw cudf::logic_error If `search_keys` type does not match the element type in `lists`
  * @throw cudf::logic_error If `lists` or `search_keys` contains nested elements (LIST, STRUCT)
+ * @throw cudf::data_type_error If `search_keys` type does not match the element type in `lists`
```
ttnghia, line 168: Please remove line 167. It was left from previous changes, no longer relevant (`index_of` now supports nested types).

cpp/include/cudf/lists/contains.hpp
vyasr: Done

cpp/include/cudf/lists/contains.hpp
vyasr: Done

ttnghia (approved):

PR #12427, opened 2022-12-20
Build wheels alongside conda CI

sevagh:
## Description

This PR adds pip wheel CI to the Conda CI, instead of having them work separately.

## Checklist


.github/workflows/pr.yaml
```diff
@@ -20,6 +20,10 @@ jobs:
       - conda-python-other-tests
       - conda-java-tests
       - conda-notebook-tests
+      - cudf-wheel-build
```
bdice, line 23: Wondering if we should align the first word of the job, somehow -- "conda" and "wheel". Maybe `wheel-build-cudf`, `wheel-tests-cudf`, `wheel-build-dask-cudf`, `wheel-tests-dask-cudf`?

ajschmidt8 (approved):

ajschmidt8 (approved):
(double approval because I forgot the :fire: in the first one)

:fire: :fire: :fire: :fire: :fire: :fire:

.github/workflows/build.yaml
```diff
@@ -55,3 +61,53 @@ jobs:
...
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      repo: rapidsai/cudf
+      branch: ${{ inputs.branch }}
+      sha: ${{ inputs.sha }}
+      date: ${{ inputs.date }}
+      package-name: cudf
+      package-dir: python/cudf
+      versioneer-override: ${{ inputs.wheel_versioneer_override }}
+      skbuild-configure-options: "-DCUDF_BUILD_WHEELS=ON -DDETECT_CONDA_ENV=OFF"
```
vyasr, line 76: We can leave this for now (something I will raise more broadly in the new year) but AFAIK this variable doesn't do anything right now. The question would be whether it should be set to do anything. That's a rapids-wide discussion though; I think only cuML is actively making use of this now and that's partly because it needs to forward along to deps like treelite.

vyasr: https://github.com/rapidsai/rapids-cmake/issues/334

.github/workflows/pr.yaml
```diff
@@ -78,3 +82,37 @@ jobs:
...
+      package-dir: python/cudf
+      skbuild-configure-options: "-DCUDF_BUILD_WHEELS=ON -DDETECT_CONDA_ENV=OFF"
+  wheel-tests-cudf:
+    needs: wheel-build-cudf
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-test.yml@upkeep/manylinux-improvements
+    with:
+      build_type: pull-request
+      package-name: cudf
+      test-before-arm64: "pip install tokenizers==0.10.2 cupy-cuda11x -f https://pip.cupy.dev/aarch64"
```
vyasr, line 100: Since it's already working we can leave it for now, but I'll note that in https://github.com/rapidsai/cudf/pull/12377 we actually removed tokenizers from arm tests and only test on x86 for conda packages.

ci/wheel_smoke_test_cudf.py
```diff
@@ -0,0 +1,20 @@
...
+    names = ["n_legs", "animals"]
+
+    print(animals)
+
+    foo = pa.table([n_legs, animals], names=names)
+    print(foo)
+
+    df = cudf.DataFrame.from_arrow(foo)
+    print(df)
+    print(df.loc[df["animals"] == "Centipede"]["n_legs"])
```
vyasr, line 18: I'd drop the prints, but maybe worth adding an extra assert or two (e.g. this final equality check could be an assertion.

vyasr (approved):
Couple of minor suggestions but looks awesome!

.github/workflows/build.yaml
```diff
@@ -52,3 +52,49 @@ jobs:
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
       sha: ${{ inputs.sha }}
+  wheel-build-cudf:
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@main
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      repo: rapidsai/cudf
```
ajschmidt8, line 60: we can remove this `repo` key from all of the `build.yaml` workflows (similarly to how I did for rmm [here](https://github.com/rapidsai/rmm/commit/7fb45d44c8ca4151581f5c57e413d1aea4b00f3f)).

This key was only ever used because we were using a `workflow_call` event to call this workflow from the https://github.com/rapidsai/actions/ repository for nightly builds.

Now that this file uses `workflow_dispatch` instead of `workflow_call`, this key isn't relevant anymore.

We should make this change for all jobs in this workflow and also all other wheel PRs.

ajschmidt8 (changes requested):

ajschmidt8 (approved):

PR #12432, opened 2022-12-22
Expose `nested_json.hpp` detail header

ttnghia:
This exposes the detail header `nested_json.hpp`, moving it from `src/io/json/` into `include/cudf/io/detail/` folder. It is necessary for accessing `cudf::io::json::detail::get_token_stream` in cudf Java-JNI.


ajschmidt8 (approved):

karthikeyann:
Can we move only the required declarations to detail header?
For example: `json_col_t`, `json_column, `device_json_column` are not used anywhere in function signatures.
Also, other types `tree_meta_t` , `reduce_to_column_tree` etc need not be exposed.

ttnghia:
> Can we move only the required declarations to detail header? For example: `json_col_t`, `json_column, `device_json_column`are not used anywhere in function signatures. Also, other types`tree_meta_t`,`reduce_to_column_tree` etc need not be exposed.

Sure. I only need `get_token_stream()`. Do you want to have two `nested_json.hpp` files (one file as the old one and one file as the new one)?

karthikeyann:
Yes. one in src, one in detail, is fine.

cpp/include/cudf/io/detail/nested_json.hpp
```diff
@@ -0,0 +1,146 @@
...
+ */
+
+#pragma once
+
+#include <cudf/io/json.hpp>
+#include <cudf/io/types.hpp>
+#include <cudf/types.hpp>
+#include <cudf/utilities/bit.hpp>
+#include <cudf/utilities/default_stream.hpp>
+#include <cudf/utilities/error.hpp>
```
karthikeyann, line 24: These includes could be removed.

vuule:
Thoughts on merging `detail/json.hpp` and `detail/nested_json.hpp`? Having `nested_json.hpp` kind of implies that stuff in `json.hpp` does not support nesting, which is not really the case. Alternatively, maybe `nested_json.hpp` should just be renamed.

ttnghia:
> Thoughts on merging `detail/json.hpp` and `detail/nested_json.hpp`? Having `nested_json.hpp` kind of implies that stuff in `json.hpp` does not support nesting, which is not really the case. Alternatively, maybe `nested_json.hpp` should just be renamed.

Yeah that sounds good too, but out of scope of this PR :smiley:

revans2:
> Yeah that sounds good too, but out of scope of this PR

But if we don't change the publicly exposed detail header now any code that depends on it will break when it is changed. I personally think that we should at least fix it for the header that we are exposing. Perhaps something like `tokenize_json.hpp`

ttnghia:
> I personally think that we should at least fix it for the header that we are exposing. Perhaps something like `tokenize_json.hpp`

This sounds good. I've renamed the new file to `tokenize_json.hpp`.

vuule (approved):

PR #12434, opened 2022-12-24
Rework remaining cudf::strings::from_xyz functions to use make_strings_children

vuule:
## Description
Issue https://github.com/rapidsai/cudf/issues/12167

Use `make_strings_children` in `cudf::strings::from_fixed_point`, `cudf::strings::from_floats` and `cudf::strings::from_integers` to simplify code and make the functions more robust.

## Checklist


cpp/src/strings/convert/convert_durations.cu
```diff
@@ -389,7 +350,29 @@ struct duration_to_string_fn : public duration_to_string_size_fn<T> {
     return ptr;
   }
 
-  __device__ void operator()(size_type idx)
+  __device__ size_type string_size(size_type idx)
+  {
+    if (d_durations.is_null(idx)) return 0;
```
davidwendt, line 355:
```suggestion
```
This is redundant since it is already handled by the caller. Since the `idx` parameter is only used to retrieve the `duration`, perhaps the calling function can resolve and just pass this instead.
This comment applies to other similar change patterns in this PR as well.

davidwendt (changes requested):

cpp/src/strings/convert/convert_floats.cu
```diff
@@ -351,31 +351,36 @@ struct ftos_converter {
 };
 
 template <typename FloatType>
-struct float_to_string_size_fn {
-  column_device_view d_column;
+struct from_floats_fn {
+  column_device_view d_floats;
+  int32_t* d_offsets;
```
davidwendt, line 356:
```suggestion
  size_type* d_offsets;
```
Use `size_type` or `offset_type` here instead. Same for `convert_integers.cu` too.

cpp/src/strings/convert/convert_durations.cu
vuule: Made the requested changes. Not sure I got the `idx` parameter suggestion right, though.

davidwendt (approved):

cpp/src/strings/convert/convert_durations.cu
```diff
@@ -389,16 +350,50 @@ struct duration_to_string_fn : public duration_to_string_size_fn<T> {
...
+      thrust::seq,
+      d_format_items,
+      d_format_items + items_count,
+      [this, &timeparts] __device__(format_item item) -> size_type {
+        if (item.item_type == format_char_type::literal)
+          return 1;
+        else if (item.length != -1)
+          return item.length;
+        else
+          return format_length(item.value, &timeparts);
```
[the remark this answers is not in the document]
vuule: oh no, my line count!
yeah, we usually do, will update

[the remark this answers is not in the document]
vuule: done, thanks

cpp/src/strings/convert/convert_durations.cu
```diff
@@ -186,11 +186,14 @@ __device__ void dissect_duration(T duration, duration_component* timeparts)
   }
 }
 
+namespace {
 template <typename T>
-struct duration_to_string_size_fn {
-  const column_device_view d_durations;
-  const format_item* d_format_items;
+struct from_durations_fn {
+  column_device_view d_durations;
```
ttnghia, line 192:
```suggestion
  column_device_view const d_durations;
```

cpp/src/strings/convert/convert_durations.cu
```diff
@@ -186,11 +186,14 @@ __device__ void dissect_duration(T duration, duration_component* timeparts)
...
+  format_item const* d_format_items;
   size_type items_count;
```
ttnghia, line 194: Why none of these variables `const`? They won't change at all.

vuule: They are data members, and marking data members as const impacts class' special members. IIRC, it disables the move ctor/assignment, because move operations would modify the "source" object. AFAIK we avoid const data members in libcudf for this reason.

ttnghia: I suspect that too. But from my view I don't see anywhere the functor can be moved. Unless you explicitly use `std::move()` the functor is most likely to be copied.

mythrocks: I'm glad for this discussion:
https://quuxplusone.github.io/blog/2022/01/23/dont-const-all-the-things/#data-members-never-const
On the face of it, it sounds like a bad idea.

cpp/src/strings/convert/convert_durations.cu
```diff
@@ -423,30 +414,16 @@ struct dispatch_from_durations_fn {
...
+      make_strings_children(from_durations_fn<T>{d_column, d_format_items, compiler.items_count()},
+                            strings_count,
+                            stream,
+                            mr);
 
     return make_strings_column(strings_count,
-                               std::move(offsets_column),
-                               std::move(chars_column),
+                               std::move(children.first),
+                               std::move(children.second),
```
mythrocks, line 426: Question: Would this benefit from using structured bindings?

mythrocks (approved):
LGTM. Some minor nitpicks.

cpp/src/strings/convert/convert_durations.cu
vuule: yup! updated

ttnghia (approved):

Issue #9497, opened 2021-10-22
[FEA] enhancement: not allow getData from ColumnVector of nested type


wjxiz1992:
**Is your feature request related to a problem? Please describe.**
When getData from a ColumnVector of nested type such as LIST, the address will be null, and that will trigger a NPE.

**Describe the solution you'd like**
When calling getData API from a ColumnVector, check the data type, if it's nested type, throw exception to inform that no data could be get from a nested type ColumnVector.

**Additional context**

```scala
// assume this columnVector is of nested type : LIST
val data = columnVector.getData  // null
val addr = childData.getAddress
```

GregoryKimball:
@ttnghia is this still needed?

ttnghia:
Yes this is still needed. Not a bug but having this can improve stability/reduce hidden bugs. Working on this should be very simple, just adding 1 LOC. I will take care of this.

jlowe:
Agree this is pretty minor.  The Javadoc clearly states that it returns null if there's no buffer.  We could change the behavior to throw on no buffer instead of returning null.  I don't know how many other places are already coded against the existing functionality, handling the null themselves, and would break with a thrown exception.  Not sure this change is worth the risk of breakage.

PR #12439, opened 2022-12-26
Unify native methods which are used in `ColumnVector.getData()`

ttnghia:
The method `getNativeDataAddress` is unified into `getNativeDataLength` which are always (and are only) called together in `ColumVector.getData()`. 

This PR is related to https://github.com/rapidsai/cudf/issues/9497, which asks for throwing exception if `ColumVector.getData()` is called on an input column of type LIST or STRUCT. This PR initially tried to fulfill it but discovered that the callers may get into trouble with complex handling for different input types. Thus, the issue https://github.com/rapidsai/cudf/issues/9497 can be addressed later, or be closed if it is no longer necessary.

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/ColumnView.java
+++ b/java/src/main/java/ai/rapids/cudf/ColumnView.java
@@ -3746,12 +3746,13 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
   /////////////////////////////////////////////////////////////////////////////
 
   static DeviceMemoryBufferView getDataBuffer(long viewHandle) {
-    long address = getNativeDataAddress(viewHandle);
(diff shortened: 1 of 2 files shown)


jlowe (commented):
I'm not a fan of this change, as it changes a JNI call that was not calling back into the JVM into something that is, allocating a temporary array object when one is not necessary.  It's also inconsistent with how the offsets and validity buffers are handled.

ttnghia:
This is an attempt to do minor cleanup.

> I'm not a fan of this change, as it changes a JNI call that was not calling back into the JVM into something that is, allocating a temporary array object when one is not necessary. It's also inconsistent with how the offsets and validity buffers are handled.

I see your point. Then I'll close this.

Issue #12440, opened 2022-12-26
[FEA] Implement `cudf::data_type::to_string()` for debugging purpose


ttnghia:
Many times, I got exception during cudf development like this:
```
ai.rapids.cudf.CudfException: CUDF failure at: ../src/column/column_factories.cpp:58: Invalid element type.
```

In order to debug the code, I typically add this line:
```
printf("type id: %d\n",(int) input->type().id());
```

The output value can be `type id: 23`. I need to further dig into the `cudf::type_id` enums to figure out what type the value `23` refers to.

It is much more convenient if we have `cudf::data_type::to_string()` so I just call:
```
printf("data type: %s\n",(int) input->type().to_string().c_str());
```
and can get a readable output:
```
data type: STRING
```

ttnghia:
Since this is for debugging/testing purpose, adding a test utility `cudf::test::to_string(data_type)` should be a better option.

vyasr:
Would the changes in https://github.com/rapidsai/cudf/pull/12158 solve your problem? It adds `type_to_name(data_type)`.

ttnghia:
Oh yeah, I didn't realize that. Close the issue.

PR #12442, opened 2022-12-27
Add `std::string cudf::test::to_string(cudf::type_id type)`

ttnghia:
This adds a test utility function to print out the column type as string, for debugging purpose. For example:
```
std::cout << "Type: " << cudf::test::to_string(input.type().id()) << "\n";
```

Output:
```
Type: STRING
```

Closes https://github.com/rapidsai/cudf/issues/12440.


vyasr (changes requested):
This may duplicate the work in https://github.com/rapidsai/cudf/pull/12158, can you confirm that the `type_to_name` implementation there will be sufficient @bdice?

ttnghia:
I didn't realize that there was duplicate work on this. Yes, `type_to_name` there should offer the same functionality to this. 

Close as no longer needed.

PR #12330, opened 2022-12-07
Improve JSON column creation performance (list offsets)

karthikeyann:
## Description
In JSON parser during JSON column creation stage, col_ids is sorted twice - once inside reduce_to_column_tree and before list offset computation.
This PR avoids repeated sort by col_ids for list offset computation. Reuses previously sorted col_ids and node_ids.

Improves parsing bandwidth from 2.170GB/s to 2.414GB/s on GV100 for 1GB json string. (The parsing bandwidth computation includes json string transfer time. Excluding data transfer time, this PR improves parsing bandwidth from 3.9GB/s to 4.5GB/s)

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12370, opened 2022-12-13
Purge non-empty nulls in `cudf::make_lists_column`

ttnghia:
During construction, the new lists column is also applied a null mask. Such null superposition allows the existence of non-empty nulls, which may undefined behavior later on for various cudf APIs.

This PR fixes it by removing non-empty nulls after lists column constructions. It also fixes unused variables `stream` and `mr`.

After lists creation is corrected in this PR, a lot of tests failed because either these tests were incorrect, or the corresponding APIs were wrongly handling nulls (i.e., they generate outputs with non-empty null lists). Below is the list of the tests that are also fixed in this PR:
 - [X] COPYING_TEST
 - [X] LISTS_TEST
 - [X] PARQUET_TEST
 - [X] STRUCTS_TEST
 - [X] UTILITIES_TEST

In addition, `JSON_TEST` failed due to an unrelated bug (https://github.com/rapidsai/cudf/issues/12418). The bug already existed there but is discovered only now due to changes in this PR. The corresponding failed test is just temporarily disabled and will be fixed separately.

Also closes:
 * https://github.com/rapidsai/cudf/issues/12405

PR #12447, opened 2022-12-28
Fix List offsets bug in mixed type list column in nested JSON reader

### LINKED ISSUE #12418: [BUG] `cudf::experimental::read_json` wrongly parse data when the input is invalid
From the current json test:
```
TEST_F(JsonReaderTest, JsonNestedDtypeSchema)
```

The test specifies an input:
```
std::string json_string = R"( [{"a":[123, {"0": 123}], "b":1.0}, {"b":1.1}, {"b":2.1}])";
```
Notice that the input for list `a` (lists of structs) is invalid: the first element is `123`, which is not a struct. As such, the test expects the result of parsing `a` to be `[null, 123.0]` which is reasonable.

However, if at the end of the test, we print out the result by:
```
cudf::test::print(result.tbl->get_column(0));
```
then we will get this:
```
List<Struct<float,>>:
Length : 3
Offsets : 0, 1, 1, 1
Null count: 2
001
   Struct<float,>:
   Length : 2:
   Null count: 1
   10
      10
      NULL, 123
```

The output column seems to have corrupted data. In particular, it is a lists column having only one non-null row (the first row), which in turn has only one element (offsets `0, 1]`). However, the child column has 2 elements.

karthikeyann:
## Description
Fixes the bug in list offsets in mixed types - string/value type with struct type.

In nested JSON reader, following json string
`[{"a":[123, {"0": 123}], "b":1.0}, {"b":1.1}, {"b":2.1}]`
should produce first column as

```
List<Struct<float,>>:
Length : 3
Offsets : 0, 2, 2, 2
Null count: 2
001
   Struct<float,>:
   Length : 2:
   Null count: 1
   10
      10
      NULL, 123
```

Depends on  #12330
closes #12418
## Checklist


ttnghia:
Can you also update the C++ unit test in https://github.com/rapidsai/cudf/issues/12418 too? I realize that python tests are updated in this PR but C++ tests are also need to be updated.

cpp/src/io/json/json_column.cu
```diff
@@ -195,6 +197,45 @@ reduce_to_column_tree(tree_meta_t& tree,
...
     });
 
+  // Mixed types in List children go to different columns,
+  // so all immediate children of list column should have same max_row_offsets.
+  //   create list's children max_row_offsets array. (initialize to zero)
+  //   atomicMax on  children max_row_offsets array.
+  //   gather the max_row_offsets from children row offset array.
+  {
+    rmm::device_uvector<NodeIndexT> list_parents_children_max_row_offsets(num_columns, stream);
+    thrust::fill(rmm::exec_policy(stream),
```
ttnghia, line 207:
```suggestion
    thrust::uninitialize_fill(rmm::exec_policy(stream),
```

cpp/src/io/json/json_column.cu
```diff
@@ -686,7 +744,11 @@ std::pair<std::unique_ptr<column>, std::vector<column_name_info>> device_json_co
       auto [child_column, names] =
         json_col.child_columns.empty()
           ? std::pair<std::unique_ptr<column>,
-                      std::vector<column_name_info>>{std::make_unique<column>(), {}}
+                      std::vector<column_name_info>>{std::make_unique<column>(
+                                                       data_type{type_id::INT8},
+                                                       0,
+                                                       rmm::device_buffer{0, stream, mr}),
```
ttnghia, line 750: Oh why? The lists column has empty child, right?

cpp/src/io/json/json_column.cu
karthikeyann: Difference between `thrust::fill` and [thrust::uninitialized_fill](https://thrust.github.io/doc/group__filling_ga7afcf37f1bd2799befe7e93aae7b43a1.html) is that `thrust::fill` uses assignment operator, and `thrust::uninitialized_fill` uses copy constructor. 
For primitive types, it does not make any difference.

cpp/src/io/json/json_column.cu
ttnghia: I don't see this change making any difference, except that the new child type is now `INT8` instead of `EMPTY`. This may avoid exception thrown by `type_dispatcher` in some code paths. But IMO I think keeping the return type of `EMPTY` column is more appropriate.

karthikeyann: Type dispatcher fails when the child column is empty. That's why it's defaulted to INT8.

karthikeyann:
> Can you also update the C++ unit test in #12418 too? I realize that python tests are updated in this PR but C++ tests are also need to be updated.
> 
> Note that I disabled that test before (in #12370). So we need to both update its content and enable it.

Enabled and updated.

ttnghia (approved):

python/cudf/cudf/tests/test_json.py
```diff
@@ -1006,3 +1006,95 @@ def test_json_round_trip_gzip():
         fo.seek(loc)
         written_df = cudf.read_json(fo, orient="records", lines=True)
     assert_eq(written_df, df)
+
+
+@pytest.mark.parametrize(
+    "jsonl_string",
+    [
+        # simple list with mixed types
+        """{"a":[123, {"0": 123}], "b":1.0}\n {"b":1.1}\n {"b":2.1}""",
```
bdice, line 1015: Why do these have newlines _and_ spaces separating the lines? Should the spaces be removed?
```suggestion
        """{"a":[123, {"0": 123}], "b":1.0}\n{"b":1.1}\n{"b":2.1}""",
```

python/cudf/cudf/tests/test_json.py
```diff
@@ -1006,3 +1006,95 @@ def test_json_round_trip_gzip():
...
+                    else x
+                )
+        return df
+
+    # both json lines and json string tested.
+    json_string = "[" + jsonl_string.replace("\n", ",") + "]"
+    pdf = pd.read_json(jsonl_string, orient="records", lines=True)
+    pdf2 = pd.read_json(json_string, orient="records", lines=False)
+    assert_eq(pdf, pdf2)
+    pdf = _replace_with_nulls(pdf, [123, "123", 12.3])
```
vuule, line 1079: can you please document why this modification is needed?

bdice (approved):
Approving Python test.

python/cudf/cudf/tests/test_json.py
karthikeyann: Spaces does not matter.

cpp/src/io/json/json_column.cu
davidwendt: The `EMPTY` type is likely to be removed: https://github.com/rapidsai/cudf/issues/12477

ttnghia: > Type dispatcher fails when the child column is empty. That's why it's defaulted to INT8.

Can you also add a comment quoting this please?

vuule:
@karthikeyann could you summarize the fix in the PR description? I'm having a hard time following the logic of the code changes, so it's probably a good idea to leave some additional info for future development.

python/cudf/cudf/tests/test_json.py
```diff
@@ -1006,3 +1006,95 @@ def test_json_round_trip_gzip():
         fo.seek(loc)
         written_df = cudf.read_json(fo, orient="records", lines=True)
     assert_eq(written_df, df)
+
+
+@pytest.mark.parametrize(
+    "jsonl_string",
+    [
```
vuule, line 1013: What happens when mixed types include both structs and lists? Example:
`{"a":[123, [{"0": 123}, {}], {"1": 321}], "b":1.0}`

cpp/src/io/json/json_column.cu
```diff
@@ -531,6 +532,63 @@ void make_device_json_column(device_span<SymbolT const> input,
...
+     num_list_children] __device__(size_type i) {
+      auto const node_id        = node_ids[i];
+      auto const parent_node_id = parent_node_ids[node_id];
+      if (i == 0 or parent_node_ids[node_ids[i - 1]] != parent_node_id) {
+        d_columns_data[parent_col_ids[i]].child_offsets[row_offsets[parent_node_id]] =
+          row_offsets[node_id];
+      }
+      if (i == num_list_children - 1 or parent_node_ids[node_ids[i + 1]] != parent_node_id) {
+        d_columns_data[parent_col_ids[i]].child_offsets[row_offsets[parent_node_id] + 1] =
+          row_offsets[node_id] + 1;
```
vuule, line 588: I think this part deserves a few comments, not sure why they were removed.

cpp/src/io/json/json_column.cu
```diff
@@ -531,6 +532,63 @@ void make_device_json_column(device_span<SymbolT const> input,
...
 
+  // 4. scatter List offset
+  // copy_if only node's whose parent is list, (node_id, parent_col_id)
+  // stable_sort by parent_col_id of {node_id}.
+  // For all unique parent_node_id of (i==0, i-1!=i), write start offset.
+  //                                  (i==last, i+1!=i), write end offset.
+  //    unique_copy_by_key {parent_node_id} {row_offset} to
+  //    col[parent_col_id].child_offsets[row_offset[parent_node_id]]
+
+  auto& parent_col_ids = sorted_col_ids;  // reuse sorted_col_ids
```
vuule, line 543: I'd prefer a copy here, any use of `sorted_col_ids` after alterations to `parent_col_ids` could lead to a bug. I'm more worried about such a bug than the cost of memcpy, but folks might disagree.
Maybe a future refactor (where each step is a function) could facilitate this reuse.

cpp/src/io/json/json_column.cu
```diff
@@ -531,6 +532,63 @@ void make_device_json_column(device_span<SymbolT const> input,
...
+                             node_ids.begin());
+  thrust::for_each_n(
+    rmm::exec_policy(stream),
+    thrust::make_counting_iterator<size_type>(0),
+    num_list_children,
+    [node_ids        = node_ids.begin(),
+     parent_node_ids = tree.parent_node_ids.begin(),
+     parent_col_ids  = parent_col_ids.begin(),
+     row_offsets     = row_offsets.begin(),
+     d_columns_data  = d_columns_data.begin(),
```
vuule, line 578:
```suggestion
    [node_ids        = node_ids.cbegin(),
     parent_node_ids = tree.parent_node_ids.cbegin(),
     parent_col_ids  = parent_col_ids.cbegin(),
     row_offsets     = row_offsets.cbegin(),
     d_columns_data  = d_columns_data.begin(),
```

python/cudf/cudf/tests/test_json.py
```diff
@@ -1006,3 +1006,95 @@ def test_json_round_trip_gzip():
...
+        """{"a":[{"0": 123}, "123"], "b":1.0}\n {"b":1.1}\n {"b":2.1}""",
+        """{"a":["123", {"0": 123}, "123"], "b":1.0}\n {"b":1.1}""",
+        """{"a":[123]}\n {"a":[{"0": 123}], "b":1.0}\n {"b":1.1}""",
+        """{"a":[{"0": 123}]}\n {"a":[123], "b":1.0}\n {"b":1.1}""",
+        """{"a":[{"0": 123}]}\n {"a": []}\n {"a":[123], "b":1.0}\n{"b":1.1}""",
+        """{"b":1.0, "a":[{"0": 123}]}\n {"a":[123]}\n {"b":1.1}\n{"a": []}""",
+        """{"a": []}\n {"a":[{"0": 123}]}\n {"a":[123], "b":1.0}\n{"b":1.1}""",
+        """{"a": []}\n {"a":[123], "b":1.0}\n {"a":[{"0": 123}]}\n{"b":1.1}""",
+        # nested list with mixed types
+        """{"a":[123, [{"0": 123}, {}]], "b":1.0}
```
vuule, line 1029: this might be a separate case:
`{"a":[123, {}], "b":1.0}`

vuule (commented):
Minor comments, still wrapping my head around the logic of the changes (hoping requested comments/description can help).
Tried to come up with more test cases anyway :)

cpp/src/io/json/json_column.cu
karthikeyann: `cbegin()` does not exist for device_span.

vuule: Have to admit, I didn't check types of all inputs here, suggestion is to use cbegin where applicable :)
Sorry for the extra churn this causes.

python/cudf/cudf/tests/test_json.py
karthikeyann: It's an error. It will throw an exception.

cpp/src/io/json/json_column.cu
karthikeyann: @ttnghia as discussed offline, gather throws error when children is type_id::EMPTY, during `purge_nonempty_nulls`.
Also, For expecting future gather operations on the resulting list column, we are choosing to return INT8 of size 0 here.


Also, `purge_nonempty_nulls` could be improved to skip the gather if it does not have non-empty nulls.

cpp/src/io/json/json_column.cu
karthikeyann: added again.

python/cudf/cudf/tests/test_json.py
```diff
@@ -1006,3 +1006,95 @@ def test_json_round_trip_gzip():
...
+        """{"a":[123, {"0": 123}, null], "b":1.0}\n {"b":1.1}\n {"b":2.1}""",
+        """{"a":["123", {"0": 123}], "b":1.0}\n {"b":1.1}\n {"b":2.1}""",
+        """{"a":[{"0": 123}, "123"], "b":1.0}\n {"b":1.1}\n {"b":2.1}""",
+        """{"a":["123", {"0": 123}, "123"], "b":1.0}\n {"b":1.1}""",
+        """{"a":[123]}\n {"a":[{"0": 123}], "b":1.0}\n {"b":1.1}""",
+        """{"a":[{"0": 123}]}\n {"a":[123], "b":1.0}\n {"b":1.1}""",
+        """{"a":[{"0": 123}]}\n {"a": []}\n {"a":[123], "b":1.0}\n{"b":1.1}""",
+        """{"b":1.0, "a":[{"0": 123}]}\n {"a":[123]}\n {"b":1.1}\n{"a": []}""",
+        """{"a": []}\n {"a":[{"0": 123}]}\n {"a":[123], "b":1.0}\n{"b":1.1}""",
+        """{"a": []}\n {"a":[123], "b":1.0}\n {"a":[{"0": 123}]}\n{"b":1.1}""",
```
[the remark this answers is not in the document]
karthikeyann: added

python/cudf/cudf/tests/test_json.py
karthikeyann: added

python/cudf/cudf/tests/test_json.py
karthikeyann: described at beginning of the test. I added another comment here too.

cpp/src/io/json/json_column.cu
karthikeyann: This is done to reduce memory usage.
Right now, for 1 GB JSON string, it takes 4.93 GB during`reduce_to_column_tree`. It's less compared to peak memory 8.9 GB.

vuule: I'm not quite sold on it, it's only an integer per column :)

PR #12451, opened 2022-12-29
Use rapids-cmake parallel testing feature

robertmaynard:
## Description
Converts libcudf over to use rapids-cmake new GPU aware parallel testing feature, which allows tests to run across all the GPUs on a machine without oversubscription.

This will allow developers to run `ctest -j<N>` and ctest will figure out given the current machine how many tests it can run in parallel given the current GPU set ( currently 2 tests per GPU ).

## Checklist


ajschmidt8:
(updating the branch because it will help me test something for GHAs)

robertmaynard:
Moving to 23.04

ci/test_cpp.sh
```diff
@@ -16,28 +16,25 @@ echo "STREAM_IDENTIFY_LIB=${STREAM_IDENTIFY_LIB_MODE_CUDF}"
...
+GTEST_FILTER="SpanTest.CanConstructFromDeviceContainers" ctest -R SPAN_TEST -VV
+LD_PRELOAD= ctest -R SPAN_TEST --output-on-failure
+
+SUITEERROR=$?
+
+if (( ${SUITEERROR} == 0 )); then
+    cd $CONDA_PREFIX/bin/gtests/libcudf_kafka/
+    ctest -j20 --output-on-failure
+    SUITEERROR=$?
+fi
```
vyasr, line 37: Why are we rerunning the entire suite if everything fails? If this is because we are executing `ctest` multiple times in order to support the span test, could we just combine error codes instead of actually rerunning tests?

ci/test_cpp.sh
```diff
@@ -16,28 +16,25 @@ echo "STREAM_IDENTIFY_LIB=${STREAM_IDENTIFY_LIB_MODE_CUDF}"
...
+export GTEST_CUDF_STREAM_MODE="new_cudf_default"
+export GTEST_OUTPUT=xml:${RAPIDS_TESTS_DIR}/
+export LD_PRELOAD=${STREAM_IDENTIFY_LIB_MODE_CUDF}
+
+ctest -E SPAN_TEST -j20 --output-on-failure
+
+# This one test is specifically designed to test using a thrust device vector,
+# so we expect and allow it to include default stream usage.
+GTEST_FILTER="SpanTest.CanConstructFromDeviceContainers" ctest -R SPAN_TEST -VV
+LD_PRELOAD= ctest -R SPAN_TEST --output-on-failure
```
vyasr, line 29: tests are failing because the current filtering is backwards. The positive filter needs to be run without the preload library, whereas the rest of the tests can be run with it
```suggestion
_allowlist_filter="SpanTest.CanConstructFromDeviceContainers"
GTEST_FILTER="-${_allowlist_filter}" ctest -R SPAN_TEST -VV
GTEST_FILTER="${_allowlist_filter}" LD_PRELOAD= ctest -R SPAN_TEST -VV
```

cpp/include/cudf_test/base_fixture.hpp
```diff
@@ -244,7 +244,11 @@ inline auto make_managed() { return std::make_shared<rmm::mr::managed_memory_res
 
 inline auto make_pool()
 {
-  return rmm::mr::make_owning_wrapper<rmm::mr::pool_memory_resource>(make_cuda());
+  auto const [free, total] = rmm::detail::available_device_memory();
+  auto min_alloc =
+    rmm::detail::align_down(std::min(free, total / 10), rmm::detail::CUDA_ALLOCATION_ALIGNMENT);
+  return rmm::mr::make_owning_wrapper<rmm::mr::pool_memory_resource>(
+    make_cuda(), thrust::optional<std::size_t>{min_alloc});
```
vyasr, line 251:
```suggestion
    make_cuda(), {min_alloc});
```

cpp/include/cudf_test/base_fixture.hpp
```diff
@@ -244,7 +244,11 @@ inline auto make_managed() { return std::make_shared<rmm::mr::managed_memory_res
 
 inline auto make_pool()
 {
-  return rmm::mr::make_owning_wrapper<rmm::mr::pool_memory_resource>(make_cuda());
+  auto const [free, total] = rmm::detail::available_device_memory();
+  auto min_alloc =
+    rmm::detail::align_down(std::min(free, total / 10), rmm::detail::CUDA_ALLOCATION_ALIGNMENT);
```
vyasr, line 249: Are we specifying this now because the default pool size rmm picks of half the available memory restricts the total parallelism available to the cudf test suite?

cpp/libcudf_kafka/tests/CMakeLists.txt
```diff
@@ -27,16 +32,18 @@ function(ConfigureTest test_name)
...
-  install(
-    TARGETS ${test_name}
-    COMPONENT testing
-    DESTINATION bin/gtests/libcudf_kafka
-    EXCLUDE_FROM_ALL
+  rapids_test_add(
+    NAME ${test_name}
+    COMMAND ${test_name}
+    GPUS 1
+    PERCENT 25
```
vyasr, line 39: Out of curiosity, how are you picking this number?

fetch_rapids.cmake
```diff
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2018-2022, NVIDIA CORPORATION.
+# Copyright (c) 2018-2023, NVIDIA CORPORATION.
```
vyasr, line 2: Seems like pre-commit got confused again. I'd like to improve the hook at some point. For now this can be manually undone.

vyasr (changes requested):

ci/test_cpp.sh
robertmaynard: This should be running the `libcudf_kafka` tests only if we have no errors

vyasr: Completely misread that, thanks.

cpp/libcudf_kafka/tests/CMakeLists.txt
robertmaynard: Entirely abitrary. We currently only have a single `libcudf_kafka` so the percent doesn't matter currently, so I selected a number that felt okay if we introduce more tests in the future.

cpp/include/cudf_test/base_fixture.hpp
robertmaynard: Exactly the reason.

robertmaynard: Due to the round-up on the 50% it generally causes failures even when running two tests at the same time.

vyasr: Cool makes sense.

ci/test_cpp.sh
```diff
@@ -16,28 +16,26 @@ echo "STREAM_IDENTIFY_LIB=${STREAM_IDENTIFY_LIB_MODE_CUDF}"
...
+export GTEST_OUTPUT=xml:${RAPIDS_TESTS_DIR}/
+export LD_PRELOAD=${STREAM_IDENTIFY_LIB_MODE_CUDF}
+
+ctest -E SPAN_TEST -j20 --output-on-failure
+
+# This one test is specifically designed to test using a thrust device vector,
+# so we expect and allow it to include default stream usage.
+_allowlist_filter="SpanTest.CanConstructFromDeviceContainers"
+GTEST_FILTER="-${_allowlist_filter}" ctest -R SPAN_TEST -VV
+GTEST_FILTER="${_allowlist_filter}" LD_PRELOAD= ctest -R SPAN_TEST -VV
```
vyasr, line 30:
```suggestion
export LD_PRELOAD=
GTEST_FILTER="${_allowlist_filter}" ctest -R SPAN_TEST -VV
```

vyasr: It seems like the preload library is getting used despite the setting. I never remember the subtleties of setting variables on the same line as a command, but my guess is that it's not playing nicely with the previous export? Going to test this.

ci/test_cpp.sh
```diff
@@ -16,28 +16,27 @@ echo "STREAM_IDENTIFY_LIB=${STREAM_IDENTIFY_LIB_MODE_CUDF}"
...
+export LD_PRELOAD=${STREAM_IDENTIFY_LIB_MODE_CUDF}
+
+ctest -E SPAN_TEST -j20 --output-on-failure
+
+# This one test is specifically designed to test using a thrust device vector,
+# so we expect and allow it to include default stream usage.
+_allowlist_filter="SpanTest.CanConstructFromDeviceContainers"
+GTEST_FILTER="-${_allowlist_filter}" ctest -R SPAN_TEST -VV
+export LD_PRELOAD=
+GTEST_FILTER="${_allowlist_filter}" ctest -R SPAN_TEST -VV
```
vyasr, line 31:
```suggestion
LD_PRELOAD= GTEST_CUDF_STREAM_MODE=default GTEST_FILTER="${_allowlist_filter}" ctest -R SPAN_TEST -VV
```

robertmaynard: Nice find!

ci/test_cpp.sh
```diff
@@ -16,28 +16,26 @@ echo "STREAM_IDENTIFY_LIB=${STREAM_IDENTIFY_LIB_MODE_CUDF}"
...
+export GTEST_OUTPUT=xml:${RAPIDS_TESTS_DIR}/
+export LD_PRELOAD=${STREAM_IDENTIFY_LIB_MODE_CUDF}
+
+ctest -E SPAN_TEST -j20 --output-on-failure
+
+# This one test is specifically designed to test using a thrust device vector,
+# so we expect and allow it to include default stream usage.
+_allowlist_filter="SpanTest.CanConstructFromDeviceContainers"
+GTEST_FILTER="-${_allowlist_filter}" ctest -R SPAN_TEST -VV
+LD_PRELOAD= GTEST_CUDF_STREAM_MODE=default GTEST_FILTER="${_allowlist_filter}" ctest -R SPAN_TEST -VV
```
vyasr, line 30: I would like to eventually set all of these using `set_tests_properties` so that
1. Running the executables directly doesn't do any stream modification
2. Running the executables with ctest will automatically use the correct stream mode
3. As public APIs are slowly modified to accept streams we can embed the appropriate testing approach into the CMake code.

I can do that work in a follow-up though. I am not sure what the implications of doing this would be for running installed tests (via the rapids-cmake install_relocatable) so we can hash that out later.

ajschmidt8 (approved):

PR #12455, opened 2022-12-30
Refactor `thrust_copy_if` into `cudf::detail::copy_if_safe`

ttnghia:
This extracts `thrust_copy_if` out of `json_tree.cu` and puts it into `cudf/detail/utilities/algorithm.cuh` with a new name `cudf::detail::copy_if_safe`. Such utility is useful not just for use in `json_tree.cu` but potentially in many other places. The next immediate use case will be to implement `from_json` in https://github.com/NVIDIA/spark-rapids-jni/pull/844.

This also changes the work in https://github.com/rapidsai/cudf/pull/12079 to adopt the new function `cudf::detail::copy_if_safe`.


cpp/include/cudf/detail/utilities/algorithm.cuh
```diff
@@ -25,4 +29,47 @@ __device__ __forceinline__ T accumulate(Iterator first, Iterator last, T init, B
...
   return init;
 }
+
+/**
+ * @brief Utility for calling thrust::copy_if.
+ *
+ * Workaround for thrust::copy_if bug (https://github.com/NVIDIA/thrust/issues/1302)
+ * where it cannot iterate over int-max values `distance(first,last) > int-max`
+ * This calls thrust::copy_if in 2B chunks instead.
+ */
```
PointKernel, line 39: Can you please update docs to include `@tparam`, `@param` and `@return`?

cpp/include/cudf/detail/utilities/algorithm.cuh
```diff
@@ -25,4 +29,47 @@ __device__ __forceinline__ T accumulate(Iterator first, Iterator last, T init, B
...
+
+  auto itr = first;
+  while (itr != last) {
+    auto const copy_end =
+      static_cast<std::size_t>(std::distance(itr, last)) <= copy_size ? last : itr + copy_size;
+    result = thrust::copy_if(policy, itr, copy_end, stencil, result, pred);
+    stencil += std::distance(itr, copy_end);
+    itr = copy_end;
+  }
+  return result;
```
PointKernel, line 62: Worth moving the implementation into the `cu/cpp` file thus we don't need the extra header inclusions in the current file. Also, this is a host API thus could be placed into `algorithm.hpp` header.

cpp/include/cudf/detail/utilities/algorithm.cuh
```diff
@@ -25,4 +29,47 @@ __device__ __forceinline__ T accumulate(Iterator first, Iterator last, T init, B
...
+ *
+ * Workaround for thrust::copy_if bug (https://github.com/NVIDIA/thrust/issues/1302)
+ * where it cannot iterate over int-max values `distance(first,last) > int-max`
+ * This calls thrust::copy_if in 2B chunks instead.
+ */
+template <typename InputIterator,
+          typename StencilIterator,
+          typename OutputIterator,
+          typename Predicate>
+OutputIterator copy_if(rmm::exec_policy policy,
```
PointKernel, line 44: I would prefer passing a stream view instead of an execution policy to align with other libcudf APIs.

cpp/include/cudf/detail/utilities/algorithm.cuh
ttnghia: This is a template function thus it can't be in `.cu/.cpp` file.

cpp/include/cudf/detail/utilities/algorithm.cuh
ttnghia: The caller will look like this:
```
cudf::detail::copy_if(rmm::execution_policy(stream),....);
```

This is similar to what we see if using `thrust::copy_if`:
```
thrust::copy_if(rmm::execution_policy(stream),....);
```

Thus, passing `execution_policy` is necessary here.

cpp/include/cudf/detail/utilities/algorithm.cuh
ttnghia: Done.

cpp/include/cudf/detail/utilities/algorithm.cuh
PointKernel: > passing execution_policy is necessary here

Passing a stream would work as well, right? e.g. 
```cpp
cudf::detail::copy_if(..., rmm::cuda_stream_view stream) {
  ...
  thrust::copy_if(rmm::execution_policy(stream.value()),....);
  ...
}
```
My point is that we should consistently expose streams as opposed to execution policies for libcudf APIs. The latter is thrust-centric and there is no such concept in libcudf yet.

ttnghia: Okay that makes sense.

ttnghia: Done.

PointKernel (approved):
LGTM

Thanks!

ttnghia:
I thought about that too but didn't have a very good candidate name. How about `safe_copy_if`?

davidwendt:
> I thought about that too but didn't have a very good candidate name. How about `safe_copy_if`?

We usually use `safe` as a suffix like the functions in here: https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/include/cudf/detail/utilities/integer_utils.hpp
Perhaps `copy_if_safe`? That sounds more like "copy, if safe" instead of "copy-if, safe". Naming is hard.
I'm ok with either `safe_copy_if` or `copy_if_safe`.
Other options: `blocked_copy_if` or `chunked_copy_if`

ttnghia:
> > I thought about that too but didn't have a very good candidate name. How about `safe_copy_if`?
> 
> We usually use `safe` as a suffix like the functions in here: https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/include/cudf/detail/utilities/integer_utils.hpp Perhaps `copy_if_safe`? That sounds more like "copy, if safe" instead of "copy-if, safe". Naming is hard. I'm ok with either `safe_copy_if` or `copy_if_safe`. Other options: `blocked_copy_if` or `chunked_copy_if`

Then let's use `copy_if_safe` for consistency with other APIs having the `_safe` suffix.

davidwendt (approved):

ttnghia:
@gpucibot merge

PR #12457, opened 2022-12-30
Build CUDA `11.8` and Python `3.10` Packages

ajschmidt8:
## Description

This PR updates `cudf` to build against branch [cuda-118](https://github.com/rapidsai/shared-action-workflows/compare/cuda-118) of the `shared-action-workflow` repository.

That branch contains updates for CUDA 11.8 and Python 3.10 packages.

It also includes some minor file renames.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rmm/blob/HEAD/CONTRIBUTING.md).


dependencies.yaml
```diff
@@ -199,6 +203,10 @@ dependencies:
               py: "3.9"
             packages:
               - python=3.9
+          - matrix:
+              py: "3.10"
+            packages:
+              - python=3.10
           - matrix:
             packages:
               - python>=3.8,<3.10
```
bdice, line 212:
```suggestion
              - python>=3.8,<3.11
```

bdice (commented):
We’ll need to double check some packaging for Python 3.10. See https://github.com/rapidsai/rmm/pull/1166 for examples of some of the additional edits we need to apply. I can take a closer look in a few days.

dependencies.yaml
```diff
@@ -108,6 +108,10 @@ dependencies:
               cuda: "11.5"
             packages:
```
robertmaynard, line 109: Do we need to drop the 11.5 enty in the matrix since the above matrix for `x86_64`,`aarch64` + cuda version doesn't have 11.5 in it anymore

robertmaynard (changes requested):

dependencies.yaml
ajschmidt8: For this `cudatoolkit:` section, we need to keep around any versions that we still test on.

So `11.5` should still stick around for now.

robertmaynard (approved):

brandon-b-miller (approved):
Approving python changes

galipremsagar (approved):

raydouglass (approved):

AyodeAwe (approved):

bdice:
Marking this as `DO NOT MERGE` until @ajschmidt8 gives the green light. We need to provision another CUDA 11.8 machine on GitHub Actions to ensure that we have enough resources for cuDF builds.

PR #12462, opened 2023-01-02
Update workflows for nightly tests

ajschmidt8:
Due to some limitations on the number of reusable workflows that can be used in a given GH Action workflow, I need to switch the nightly workflow files (`build.yaml` and `test.yaml`) to use `workflow_dispatch` instead of `workflow_call`.

Because of this switch, we can also remove the hardcoded `repo` field in `{build,test}.yaml`.


ajschmidt8:
admin merging since changes are benign and not tested in PRs.

PR #12464, opened 2023-01-03
Update cudf_assert docs with correct NDEBUG behavior

### LINKED ISSUE #12459: [BUG] `cudf_assert` does not have effect when `NDEBUG` is defined
The documentation of `cudf_assert` states that:
```
...
 * This is effectively the same as the standard `assert` macro, except it will
 * not be compiled out in Release builds, i.e., this macro will be present
 * regardless of the state of `NDEBUG`.
```

However, it is implemented as:
```
#if !defined(NDEBUG) && 
...
#else
#define cudf_assert(e) (static_cast<void>(0))
#endif
```

As such, when `NDEBUG` is defined, this macro is just a no-op and is useless.

robertmaynard:
## Description
Since 21.06 `cuff_assert` has been a no-op in release mode. This brings the docs inline with the implementation.

Fixes #12459

## Checklist

### DIFF
--- a/cpp/include/cudf/detail/utilities/assert.cuh
+++ b/cpp/include/cudf/detail/utilities/assert.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2022, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2023, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/cpp/include/cudf/detail/utilities/assert.cuh
+++ b/cpp/include/cudf/detail/utilities/assert.cuh
@@ -22,7 +22,7 @@
  * @brief `assert`-like macro for device code
  *
  * This is effectively the same as the standard `assert` macro, except it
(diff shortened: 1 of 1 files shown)


cpp/include/cudf/detail/utilities/assert.cuh
```diff
@@ -19,13 +19,11 @@
...
- * @brief `assert`-like macro for device code that persists in Release builds.
+ * @brief `assert`-like macro for device code
  *
- * This is effectively the same as the standard `assert` macro, except it will
- * not be compiled out in Release builds, i.e., this macro will be present
- * regardless of the state of `NDEBUG`.
- *
- * Relies on the `__PRETTY_FUNCTION__` macro which is specific to GCC and Clang.
+ * This is effectively the same as the standard `assert` macro, except it
+ * reiles on the `__PRETTY_FUNCTION__` macro which is specific to GCC and Clang
```
davidwendt, line 25:
```suggestion
 * relies on the `__PRETTY_FUNCTION__` macro which is specific to GCC and Clang
```

ttnghia (approved):
Approve, pending copyright year fix.

davidwendt (approved):

robertmaynard:
@gpucibot merge

PR #12465, opened 2023-01-03
Fix compile issue with arrow 10

ttnghia:
Arrow has been upgraded to version 10 in https://github.com/rapidsai/cudf/pull/12327. Unfortunately the new Arrow version trigger new warnings that are treated as compile error:
```
_deps/arrow-src/cpp/src/arrow/result.h(263): error #2810-D: ignoring return value type with "nodiscard" attribute
```

This PR fixes that by temporarily disabling such warnings.

### DIFF
--- a/cpp/include/cudf/io/datasource.hpp
+++ b/cpp/include/cudf/io/datasource.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2022, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2023, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/cpp/include/cudf/io/datasource.hpp
+++ b/cpp/include/cudf/io/datasource.hpp
@@ -31,10 +31,15 @@
 #include <arrow/filesystem/s3fs.h>
 #pragma nv_diag_default 611
 
+// We disable warning 2810 to workaround the compile issue (warning treated as error):
+// result.h(263): error #2810-D: ignoring return value type with "nodiscard" attribute
+#pragma nv_diag_suppress 2810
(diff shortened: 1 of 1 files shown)


davidwendt (approved):

robertmaynard (approved):

cpp/include/cudf/io/datasource.hpp
```diff
@@ -31,10 +31,15 @@
 #include <arrow/filesystem/s3fs.h>
 #pragma nv_diag_default 611
 
+// We disable warning 2810 to workaround the compile issue (warning treated as error):
+// result.h(263): error #2810-D: ignoring return value type with "nodiscard" attribute
+#pragma nv_diag_suppress 2810
+#include <arrow/result.h>
```
vuule, line 37: Can we file an Arrow issue for this?

robertmaynard: The underlying cause is an nvcc issue, which I believe @vyasr has submitted an issue for.

ttnghia: I think so. Also the issue that we did similarly in the past (warning 611: https://github.com/rapidsai/cudf/pull/12465/files/1ab9604856ff666d87847e569af8b2c4b49d5dbc#diff-e87e4906a9ed3d9d0c2c663d5e648ec99950d9042fb297a6a195a417276dc55fR29-R32).

ttnghia: We had similar workaround in the past in https://github.com/rapidsai/cudf/pull/8916#issuecomment-1299410665, not sure if all these issues have been reported.

ttnghia:
@gpucibot merge

cpp/include/cudf/io/datasource.hpp
vyasr: Yes, I have an open nvbug for this issue.

PR #12466, opened 2023-01-03
Use cudaMemcpyDefault.

bdice:
## Description
This uses `cudaMemcpyDefault` instead of `cudaMemcpy{Host,Device}To{Host,Device}`. After consultation with @vuule on another PR and asking the CUDA team, there is no additional cost to using the default value over specifying the host/device residency.

The only potential advantage of specifying the direction is that an error is raised if the direction does not match the source/destination parameters' location. @vuule and I are both +1 for this change but neither of us feel strongly if others feel differently. On the other hand, there is less to think about if we always use `cudaMemcpyDefault`.

## Checklist


davidwendt (approved):

ttnghia (approved):

bdice:
@gpucibot merge

PR #12468, opened 2023-01-03
Raise warnings as errors in the test suite

vyasr:
## Description

This PR fixes a final few warnings introduced in PRs merged after #12406 or with changes that weren't merged into #12406. More importantly, this PR updates the testing suite to always raise uncaught warnings as errors. This is a substantial change to the testing suite that will allow us to proactively react to deprecations in our dependencies, identify areas where pandas/pyarrow/some other dependency handles input data in different ways than us, etc. I've opted for the strictest possible setting, but am open to reviewer requests to be a little more lax with certain classes of warnings.

## Checklist

### DIFF
--- a/docs/cudf/source/developer_guide/testing.md
+++ b/docs/cudf/source/developer_guide/testing.md
@@ -222,14 +222,14 @@ point (and we will remember to update the test).
 
 Some code may be expected to throw warnings.
 A common example is when a cudf API is deprecated for future removal, but many other possibilities exist as well.
-The cudf testing suite [surfaces all warnings as errors](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#controlling-warnings).
+The cudf testing suite [surfaces all warnings as errors](https://docs.pytest.org/en/latest/how-to/capture-warnings.html#controlling-warnings).
 This includes warnings raised from non-cudf code, such as calls to pandas or pyarrow.
 This setting forces developers to proactively deal with deprecations from other libraries,
 as well as preventing the internal use of deprecated cudf APIs in other parts of the library.
 Just as importantly, it can help catch real errors like integer overflow or division by zero.
 
 When testing code that is expected to throw a warnings, developers should use the
-[`pytest.warns`](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#assertwarnings) context to catch the warning.
+[`pytest.warns`](https://docs.pytest.org/en/latest/how-to/capture-warnings.html#assertwarnings) context to catch the warning.
 For parametrized tests that raise warnings under specific conditions, use the `testing._utils.expect_warning_if` decorator instead of `pytest.warns`.
 
 ```{warning}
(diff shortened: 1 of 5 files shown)


docs/cudf/source/developer_guide/testing.md
```diff
@@ -218,6 +218,29 @@ This way, when the bug is fixed, the test suite will fail at this
 point (and we will remember to update the test).
 
 
+### Testing code that throws warnings
+
+Some code may be expected to throw warnings.
+A common example is when a cudf API is deprecated for future removal, but many other possibilities exist as well.
+The cudf testing suite [surfaces all warnings as errors](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#controlling-warnings).
```
bdice, line 225: Can we use “latest” or “stable” links instead of pytest 7.1.x links?

bdice (approved):

docs/cudf/source/developer_guide/testing.md
vyasr: Done.

python/cudf/cudf/tests/pytest.ini
```diff
@@ -4,3 +4,7 @@
 markers =
     spilling: mark benchmark a good candidate to run with `CUDF_SPILL=ON`
 xfail_strict = true
+filterwarnings =
+    error
+    ignore:::.*xdist.*
+    ignore:::.*pytest.*
```
mroeschke, line 10: I ran into this as well. Specifying a module regex isn't actually supported; you'll need to specify the fully qualified module path.

I had to do something like this https://github.com/pandas-dev/pandas/pull/50386/files#diff-fad2fa83c042cce646b2c1a0455e9b60c22dd5f105ed8859ce85dcf361c07fbdR34

bdice: Here's a set of related links about warnings and modules (I contributed a review on one of the CPython PRs below in hopes to get the docs to be aligned with the implementation):

https://stackoverflow.com/questions/71112211/why-does-pythons-filterwarning-module-regex-not-work-as-expected
https://github.com/python/cpython/pull/23172
https://github.com/python/cpython/pull/9358

vyasr: Specifying a module regex is only unsupported when passed from the command line. It works fine when used in Python code, or in this case when specified to pytest via an ini file. Quoting from [the relevant section of the warnings documentation](https://docs.python.org/3/library/warnings.html#the-warnings-filter):
> module is a string containing a regular expression that the start of the fully qualified module name must match, case-sensitively. In -W and PYTHONWARNINGS, module is a literal string that the fully qualified module name must be equal to (case-sensitively), ignoring any whitespace at the start or end of module. I believe there's a CPython bug somewhere discussing this inconsistency (and it wasn't always documented accurately either).

vyasr: Ah Bradley's comment didn't refresh onto my page until after I submitted. The cpython issues are the ones I was thinking of though!

mroeschke: Ah! Okay that makes sense why my attempts on the pandas side failed before. Thanks for the heads up!

mroeschke (approved):

PR #12478, opened 2023-01-05
`one_hot_encode` to use experimental row comparators

divyegala:
## Checklist

### DIFF
--- a/cpp/src/transform/one_hot_encode.cu
+++ b/cpp/src/transform/one_hot_encode.cu
@@ -55,10 +55,11 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
   auto all_encodings =
     make_numeric_column(data_type{type_id::BOOL8}, total_size, mask_state::UNALLOCATED, stream, mr);
 
-  auto t_lhs      = table_view{{input}};
-  auto t_rhs      = table_view{{categories}};
-  auto comparator = cudf::experimental::row::equality::two_table_comparator{t_lhs, t_rhs, stream};
-  auto device_comparator =
+  auto const t_lhs = table_view{{input}};
+  auto const t_rhs = table_view{{categories}};
+  auto const comparator =
+    cudf::experimental::row::equality::two_table_comparator{t_lhs, t_rhs, stream};
+  auto const d_equal =
     comparator.equal_to(nullate::DYNAMIC{has_nested_nulls(t_lhs) || has_nested_nulls(t_rhs)});
 
   thrust::transform(
@@ -66,18 +67,17 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
     thrust::make_counting_iterator(0),
     thrust::make_counting_iterator(total_size),
     all_encodings->mutable_view().begin<bool>(),
-    [input_size = input.size(), device_comparator] __device__(size_type i) {
+    [input_size = input.size(), d_equal] __device__(size_type i) {
       auto const element_index  = cudf::experimental::row::lhs_index_type{i % input_size};
       auto const category_index = cudf::experimental::row::rhs_index_type{i / input_size};
-      return device_comparator(element_index, category_index);
+      return d_equal(element_index, category_index);
     });
 
-  auto split_iter =
+  auto const split_iter =
     make_counting_transform_iterator(1, [width = input.size()](auto i) { return i * width; });
   std::vector<size_type> split_indices(split_iter, split_iter + categories.size() - 1);
 
-  auto views = split(all_encodings->view(), split_indices, stream);
-  table_view encodings_view{views};
+  auto const encodings_view = table_view{split(all_encodings->view(), split_indices, stream)};
 
   return std::pair(std::move(all_encodings), encodings_view);
 }
--- a/cpp/src/transform/one_hot_encode.cu
+++ b/cpp/src/transform/one_hot_encode.cu
@@ -61,22 +61,22 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
   auto device_comparator =
     comparator.equal_to(nullate::DYNAMIC{has_nested_nulls(t_lhs) || has_nested_nulls(t_rhs)});
 
-  thrust::transform(rmm::exec_policy(stream),
-                    thrust::make_counting_iterator(0),
-                    thrust::make_counting_iterator(total_size),
-                    all_encodings->mutable_view().begin<bool>(),
-                    [input_size = input.size(), device_comparator] __device__(size_type i) {
-                      auto element_index  = cudf::experimental::row::lhs_index_type{i % input_size};
-                      auto category_index = cudf::experimental::row::rhs_index_type{i / input_size};
-                      return device_comparator(element_index, category_index);
-                    });
+  thrust::transform(
+    rmm::exec_policy(stream),
+    thrust::make_counting_iterator(0),
+    thrust::make_counting_iterator(total_size),
+    all_encodings->mutable_view().begin<bool>(),
+    [input_size = input.size(), device_comparator] __device__(size_type i) {
+      auto const element_index  = cudf::experimental::row::lhs_index_type{i % input_size};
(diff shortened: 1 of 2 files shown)


cpp/src/transform/one_hot_encode.cu
```diff
@@ -117,7 +51,35 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
...
+  auto device_comparator =
+    comparator.equal_to(nullate::DYNAMIC{has_nested_nulls(t_lhs) || has_nested_nulls(t_rhs)});
+
+  thrust::transform(rmm::exec_policy(stream),
+                    thrust::make_counting_iterator(0),
+                    thrust::make_counting_iterator(total_size),
+                    all_encodings->mutable_view().begin<bool>(),
+                    [input_size = input.size(), device_comparator] __device__(size_type i) {
+                      auto element_index  = cudf::experimental::row::lhs_index_type{i % input_size};
+                      auto category_index = cudf::experimental::row::rhs_index_type{i / input_size};
```
PointKernel, line 70:
```suggestion
                      auto const element_index  = cudf::experimental::row::lhs_index_type{i % input_size};
                      auto const category_index = cudf::experimental::row::rhs_index_type{i / input_size};
```
nit: `const` when possible

cpp/src/transform/one_hot_encode.cu
```diff
@@ -117,7 +51,35 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
...
+                      return device_comparator(element_index, category_index);
+                    });
+
+  auto split_iter =
+    make_counting_transform_iterator(1, [width = input.size()](auto i) { return i * width; });
+  std::vector<size_type> split_indices(split_iter, split_iter + categories.size() - 1);
+
+  // TODO: use detail interface, gh9226
```
PointKernel, line 78: Should be able to use the detail interface since #9226 has been solved.

cpp/src/transform/one_hot_encode.cu
```diff
@@ -117,7 +51,35 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
...
-  return type_dispatcher(input.type(), one_hot_encode_launcher{}, input, categories, stream, mr);
+  auto const total_size = input.size() * categories.size();
+  auto all_encodings =
+    make_numeric_column(data_type{type_id::BOOL8}, total_size, mask_state::UNALLOCATED, stream, mr);
+
+  auto t_lhs      = table_view{{input}};
+  auto t_rhs      = table_view{{categories}};
+  auto comparator = cudf::experimental::row::equality::two_table_comparator{t_lhs, t_rhs, stream};
+  auto device_comparator =
+    comparator.equal_to(nullate::DYNAMIC{has_nested_nulls(t_lhs) || has_nested_nulls(t_rhs)});
```
ttnghia, line 62: All are `const`.

cpp/src/transform/one_hot_encode.cu
```diff
@@ -117,7 +51,35 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
...
 
-  return type_dispatcher(input.type(), one_hot_encode_launcher{}, input, categories, stream, mr);
+  auto const total_size = input.size() * categories.size();
+  auto all_encodings =
+    make_numeric_column(data_type{type_id::BOOL8}, total_size, mask_state::UNALLOCATED, stream, mr);
+
+  auto t_lhs      = table_view{{input}};
+  auto t_rhs      = table_view{{categories}};
+  auto comparator = cudf::experimental::row::equality::two_table_comparator{t_lhs, t_rhs, stream};
+  auto device_comparator =
```
ttnghia, line 61:
```suggestion
  auto const d_equal =
```

cpp/src/transform/one_hot_encode.cu
```diff
@@ -117,7 +51,35 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
...
+      auto const category_index = cudf::experimental::row::rhs_index_type{i / input_size};
+      return device_comparator(element_index, category_index);
+    });
+
+  auto split_iter =
+    make_counting_transform_iterator(1, [width = input.size()](auto i) { return i * width; });
+  std::vector<size_type> split_indices(split_iter, split_iter + categories.size() - 1);
+
+  auto views = split(all_encodings->view(), split_indices, stream);
+  table_view encodings_view{views};
```
ttnghia, line 80:
```suggestion
  auto const split_iter =
    make_counting_transform_iterator(1, [width = input.size()](auto i) { return i * width; });
  std::vector<size_type> const split_indices(split_iter, split_iter + categories.size() - 1);
  auto const encodings_view = table_view{split(all_encodings->view(), split_indices, stream)};
```

cpp/src/transform/one_hot_encode.cu
```diff
@@ -117,7 +51,35 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
...
+
+  return std::pair(std::move(all_encodings), encodings_view);
```
ttnghia, line 82:
```suggestion
  return std::pair(std::move(all_encodings), std::move(encodings_view));
```

PointKernel: We have to move `all_encodings` but not `encodings_view`. FYI https://stackoverflow.com/questions/4986673/c11-rvalues-and-move-semantics-confusion-return-statement

ttnghia: Yes we don't **have to** move it. But we can move it (for micro-optimization) right? Because it contain some internal variables, moving it can avoid some sort of mem  copy.

cpp/src/transform/one_hot_encode.cu
```diff
@@ -117,7 +51,35 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
...
+      auto const element_index  = cudf::experimental::row::lhs_index_type{i % input_size};
+      auto const category_index = cudf::experimental::row::rhs_index_type{i / input_size};
+      return d_equal(element_index, category_index);
+    });
+
+  auto const split_iter =
+    make_counting_transform_iterator(1, [width = input.size()](auto i) { return i * width; });
+  std::vector<size_type> split_indices(split_iter, split_iter + categories.size() - 1);
+
+  auto const encodings_view = table_view{split(all_encodings->view(), split_indices, stream)};
```
ttnghia, line 80: This cannot be `const` because we are going to `move` it. So everything else needs to be `const` except the one that we are going  to `move` or `return`.
Sorry for the confusion. 
```suggestion
  auto encodings_view = table_view{split(all_encodings->view(), split_indices, stream)};
```

cpp/src/transform/one_hot_encode.cu
divyegala: The linked SO example is different semantics, IMO. The `std::pair` still takes advantage of RVO. I agree with @ttnghia that there's microoptimization in creating an `std::pair` from moved objects, even if the moved object happens to be a view

cpp/src/transform/one_hot_encode.cu
divyegala: Yes sorry, I should have picked that up myself - my attention is a bit divided at the moment

ttnghia (approved):

cpp/src/transform/one_hot_encode.cu
PointKernel: We shouldn't move when returning a local variable since it would inhibit NRVO and force the compiler to use a move. `return std::pair(std::move(all_encodings), encodings_view);` already allows the use of the move constructor if one is available.

divyegala: Alright, I am okay with either way for this. @PointKernel @ttnghia I will let the two of you decide, this seems to be the last thing before we can merge this PR

divyegala: Done

PointKernel (approved):
LGTM

Issue #12376, opened 2022-12-13
[BUG] ParquetChunkedReaderTest.TestChunkedReadSimpleData device-side assert with libcudf compiled with Debug


davidwendt:
**Describe the bug**
Device side assert occurs when running `PARQUET_TEST`  `ParquetChunkedReaderTest.TestChunkedReadSimpleData` with libcudf compiled in Debug. The error occurs with `generate_input(true)` which creates columns with nulls. Otherwise the error does not occur. Here is sample of the error result:
```
/cudf/cpp/include/cudf/utilities/type_dispatcher.hpp:533: constexpr decltype(auto) cudf::type_dispatcher(cudf::data_type, Functor, Ts &&...) [with IdTypeMap = cudf::id_to_type_impl; Functor = cudf::io::detail::parquet::<unnamed>::row_size_functor; Ts = const int &; Ts = const __nv_bool &]: block: [0,0,0], thread: [128,0,0] Assertion `false && "Unreachable: " "Invalid type_id."` failed.
/cudf/cpp/include/cudf/utilities/type_dispatcher.hpp:533: constexpr decltype(auto) cudf::type_dispatcher(cudf::data_type, Functor, Ts &&...) [with IdTypeMap = cudf::id_to_type_impl; Functor = cudf::io::detail::parquet::<unnamed>::row_size_functor; Ts = const int &; Ts = const __nv_bool &]: block: [0,0,0], thread: [129,0,0] Assertion `false && "Unreachable: " "Invalid type_id."` failed.
...
```
The error is reported for all threads 256 threads in the single launched block.

<details>
<summary>A partial stack trace from `compute-sanitizer`</summary>

```
========= Warp assert
=========     at 0x1930 in __assertfail
=========     by thread (192,0,0) in block (0,0,0)
[...]

PR #12488, opened 2023-01-06
Fix issues with parquet chunked reader

nvdbaranec:
Fixes two issues:

- A very old issue where we were incorrectly setting the output nesting depth in the PageNestingInfo struct in some cases.  Previously, this was benign, but the chunked reader broke because of it.  I fixed this and did some variable renaming to make things more clear.

- Fixed an issue with an optimization that was added for the chunked reader:  We were incorrectly determining when we could early out for a given page during the preprocessing step.

There is an issue as far as generating a test for this PR.  The conditions under which the second bug occurs require that values from a given row span multiple pages.  I couldn't get the cudf writer to make this happen.  The alternative is to use a pre-created file and use the python tests, but the issue there is that we don't expose skip_rows/num_rows through that API.  So I have no good way of building a test.

This fix has been vetted against a known real-world failure case.

Fixes https://github.com/rapidsai/cudf/issues/12376

## Checklist


cpp/src/io/parquet/page_data.cu
```diff
@@ -92,6 +92,25 @@ struct page_state_s {
   int32_t row_index_lower_bound;              // lower bound of row indices we should process
 };
 
+/**
+ * @brief Returns whether or not a page spans either the beginning or the end of the
+ * specified row bounds
```
ttnghia, line 97: Does this intentionally cover these 2 cases?
 * A page is not fit in the given boundary
 * A page just sits at the row boundary but it is fit in the boundary

nvdbaranec: I think this logic handles all the cases correctly:

```
     |---|                                                         |---|
     A  B                                                          A   B
            |----------------------------------------------------|
            A                                                    B
                  |PPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPP|
                  X                 Page                    Y

              |---|                                          |---|
              A   B                                          A   B
              |-------|     |---|                        |------|
              A       B     A   B                        A       B
```
The top cases fail and the bottom cases pass.   We want to include the edge cases where B==X and A==Y because those represent the case where "Page" contains values from either the prev/next row.  Although maybe using "max_row" is a bit deceptive here, since `max_row` would really be `(min_row + num_rows)-1`

vuule: then maybe begin and end fit better than min and max?

ttnghia:
Thanks for working on this. In addition to the code, do we have any unit test to catch the bug?

cpp/src/io/parquet/page_data.cu
```diff
@@ -1667,35 +1685,25 @@ __global__ void __launch_bounds__(block_size)
...
-  }
-
-  // - if this page is completely trimmed, zero out sizes.
-  if (!is_base_pass && s->num_rows == 0) {
-    int d = 0;
-    while (d < s->page.num_nesting_levels) {
-      auto const i = d + t;
-      if (i < s->page.num_nesting_levels) { pp->nesting[i].batch_size = 0; }
+      if (i < s->page.num_output_nesting_levels) {
+        // if we're not a bounding page, we're either fully within the readable region or not
```
vuule, line 1695: What does "or not" mean here? I feel like the comment could be clearer.

nvdbaranec:
At the end of the day, it hasn't been possible to get cudf or Spark to generate a useful test file here.  Should I just check the "New or existing tests cover this change" box anyway? I'm not sure what else to do.

vuule:
> At the end of the day, it hasn't been possible to get cudf or Spark to generate a useful test file here. Should I just check the "New or existing tests cover this change" box anyway? I'm not sure what else to do.

I think it should be left unchecked. The reason for the missing test is documented in the description.

vuule (approved):
LGTM
As expected, this one's really in the weeds. It would be best if the other C++ review comes from @ttnghia or @PointKernel.

PointKernel (approved):
LGTM

PR #12499, opened 2023-01-07
Add JSON reader page in user guide

GregoryKimball:
## Description
Adding a walkthrough to the User Guide that extracts data from common JSON structures. 

## Checklist


GregoryKimball:
Hello @galipremsagar would you please take a look at this content? Thank you for your help.

docs/cudf/source/user_guide/read-json.md
```diff
@@ -0,0 +1,232 @@
...
+]'''
+df = cudf.read_json(j)
+
+# JSON Lines data
+j = '\n'.join([
+    '{"a": "v1", "b": 12}',
+    '{"a": "v2", "b": 7}',
+    '{"a": "v3", "b": 5}'
+])
+df = cudf.read_json(j, lines=True)
```
ttnghia, line 30: Sorry I'm not very familiar with this.

Why we have `lines=True` here but not in line 22 above? In `read_json` code, we have this assert:
```
CUDF_EXPECTS(reader_opts.is_enabled_lines(), "Only JSON Lines format is currently supported.\n");
```

ttnghia: In addition, I wonder if the input from line 25 is correct. Json objects should be wrapped around by brackets `'['` and `']'`.

GregoryKimball: Thanks @ttnghia for your comments. This user guide was written for the new JSON reader engine. I've omitted the parameter `engine='cudf_experimental'` in these examples because we plan to make the nested reader the default in 23.02. @galipremsagar and I were talking and we will likely add this parameter to the user guide here to make it consistent with the head of 23.02

ttnghia: If this is for the new engine (even for the old engine), it is wrong. Like my previous comment above:
```
>>> j = '\n'.join([
... '{"a": [{"k": 0}], "b": {"k": [0, 1], "m": 5}}',
... '{"a": [{"k": 1}, {"k": 2}], "b": {"k": [2, 3], "m": 6}}',
... ])
>>> j
'{"a": [{"k": 0}], "b": {"k": [0, 1], "m": 5}}\n{"a": [{"k": 1}, {"k": 2}], "b": {"k": [2, 3], "m": 6}}'
```

When read that json string, the json tokenizer returns `ErrorBegin` token.

Again:
> Json objects should be wrapped around by brackets '[' and ']'.

ttnghia: I was confused when `lines=True` and the input is not wrapped by a pair of list bracket `[' and `]`. Now I understand it, but can you clarify that a bit in the doc, please? It should state clearly that we should have the normal input wrapped in `[]` and don't need that if we specify `lines=True`.

galipremsagar (approved):

ttnghia (approved):

PR #12503, opened 2023-01-09
Fix SUM/MEAN aggregation type support.

### LINKED ISSUE #8399: [FEA] Cleanup on logic inside DeviceRolling::is_supported()
There is some logic in https://github.com/rapidsai/cudf/blob/0eeb0c9239eb62b08a2be81d339c03c1078659c3/cpp/src/rolling/rolling_detail.cuh#L82  that really should be covered by `is_valid_aggregation`, but `is_valid_aggregation` is currently missing some edge cases.

```
 // MIN/MAX supports all fixed width types
 (((O == aggregation::MIN || O == aggregation::MAX) && cudf::is_fixed_width<T>()) ||

  // SUM supports all fixed width types except timestamps
  ((O == aggregation::SUM) && (cudf::is_fixed_width<T>() && !cudf::is_timestamp<T>())) ||

  // MEAN supports numeric and duration
  ((O == aggregation::MEAN) && (cudf::is_numeric<T>() || cudf::is_duration<T>())));
```

Once these edge cases are cleared up, clear out this code.  Also, if `is_valid_aggregation ` can handle a wide enough set of cases, it might be possible to remove the various `is_supported()` functions entirely and just have a single generic global template (checking `is_valid_aggregation` and `has_corresponding_operator`) used in it's place.

bdice:
## Description
This PR closes #8399. We cleaned up the logic by fixing SUM/MEAN aggregation type support, which also eliminated `TODO` comments in the target type definitions.

We kept the restriction that rolling min/max requires fixed width types because min/max aggregations do support non-fixed width in other aggregation implementations (groupby does a argmin-and-gather approach on strings, for instance).

This PR is collaborative work with @karthikeyann.

## Checklist


cpp/tests/rolling/empty_input_test.cpp
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include "cudf/utilities/traits.hpp"
```
mythrocks, line 17: Should we not change this to use `<>` for headers?

karthikeyann: autofilled by IDE. will fix this.

karthikeyann:
```suggestion
```

cpp/tests/rolling/empty_input_test.cpp
```diff
@@ -24,6 +25,7 @@
 #include <cudf/detail/aggregation/aggregation.hpp>
 #include <cudf/rolling.hpp>
 #include <cudf/scalar/scalar.hpp>
+#include <gtest/gtest.h>
```
karthikeyann, line 28:
```suggestion
```

harrism (approved):

davidwendt (approved):

Issue #12470, opened 2023-01-04
[BUG]cudf fails to read the JSON string of an empty body


firestarman:
cudf can not read the JSON string of an empty body `{}`, but pandas can read it as an empty `DataFrame`.

```
>>> pandas.read_json("{}", lines=True)
Empty DataFrame
Columns: []
Index: [0]
>>> 
>>> cudf.read_json("{}", lines=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/xxxxxxx/miniconda3/envs/cudf/lib/python3.9/site-packages/cudf/io/json.py", line 94, in read_json
    df = libjson.read_json(
  File "json.pyx", line 29, in cudf._lib.json.read_json
  File "json.pyx", line 115, in cudf._lib.json.read_json
RuntimeError: CUDF failure at: /opt/conda/conda-bld/work/cpp/src/io/json/reader_impl.cu:645: Error determining column names.
```

Spark can also read it successfully.
```
$ cat no-body.json 
{}
```
```
scala> spark.read.json("/data/tmp/no-body.json").show
++
||
++
||
++


scala> spark.read.schema("a int").json("/data/tmp/no-body.json").show
+----+
|   a|
+----+
|null|
+----+
```

PR #12498, opened 2023-01-07
Support "values" orient (array of arrays) in Nested JSON reader

karthikeyann:
## Description
Legacy GPU JSON reader can read "values" orient data in JSON string (only JSON lines).
With this PR change, Nested JSON reader can also reader "values" orient data for both JSON lines and non-line JSON string.

Examples:
```
import cudf
json="[[1, 2, 3], [4, 5], [7, 8, 9]]"
cudf.read_json(json, engine="cudf_experimental")
   0  1     2
0  1  2     3
1  4  5  <NA>
2  7  8     9
json="[1, 2, 3]\n [4, 5, null]\n [7, 8, [9]]"
cudf.read_json(json, engine="cudf_experimental", lines=True)
   0  1     2
0  1  2  None
1  4  5  None
2  7  8   [9]
```

Note that pandas passes "values" data but with `orient="records"` argument, but it is parsed as "values". Similar support is added here too. Passing values with `orient="records"` will still work).

closes https://github.com/rapidsai/cudf/issues/12446

#### Changes in this PR:
- Added a lambda `is_non_list_parent` - added extra condition to skip array (level2 list) nodes.
- New algorithm to generate `list_indices` within each row.
- Use it for hashing.
- Use it for column names. (but generated again).

**Changes to tests:**
- added row orient experiemental reader testing param
- datetime, duration needs to be wrapped with double quotes if it has whitespace or colon. (limitation of nested json reader, also to be compliant with JSON specification).

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12509, opened 2023-01-09
[REVIEW] Switch `engine=cudf` to the new `JSON` reader

galipremsagar:
## Description

Fixes: https://github.com/rapidsai/cudf/issues/12470
This PR:

- [x] Switches the `cudf` engine in json reader to the map to the `newest` JSON reader. Introduces the `cudf_legacy` engine to map to the `old` JSON reader.
- [x] Fixes an issue with `_fsspec_data_transfer` & `compression` that is required for the switch, these failures are already caught by tests.

Note: When `engine='auto'`, and `line=False`, the `pandas` json reader will be used. To override the selection, we pass `engine='cudf'`.

Dependent on :

- [x] Adding support for an array of arrays in new json reader: https://github.com/rapidsai/cudf/pull/12498

## Checklist


python/cudf/cudf/tests/test_json.py
```diff
@@ -725,7 +723,7 @@ def test_json_types_data():
 def test_json_quoted_values_with_schema(col_type, json_str):
     experimental_df = cudf.read_json(
         StringIO(json_str),
-        engine="cudf_experimental",
+        engine="cudf",
```
karthikeyann, line 726: The Line 732 engine should be different.

similarly at some other places too.

galipremsagar: Done

docs/cudf/source/user_guide/io/read-json.md
```diff
@@ -19,14 +19,14 @@ each object corresponds to a row.
     {"a": "v2", "b": 7},
     {"a": "v3", "b": 5}
 ]'''
->>> df_records = cudf.read_json(j, engine='cudf_experimental')
+>>> df_records = cudf.read_json(j, engine='cudf')
```
vuule, line 22: Should We remove the engine parameter when the default value is passed?

python/cudf/cudf/io/json.py
```diff
@@ -37,16 +37,31 @@ def read_json(
...
+    if engine == "cudf_experimental":
+        raise ValueError(
+            "engine='cudf_experimental' support has been removed, "
+            "use `engine='cudf'`"
+        )
+
+    if engine == "cudf_legacy":
+        # TODO: Deprecated in 23.02, please
+        # give some time until `cudf_legacy`
+        # support can be removed completely.
```
bdice, line 49: > "give some time until `cudf_legacy` support can be removed completely

Will [our one-release deprecation policy](https://docs.rapids.ai/api/cudf/nightly/developer_guide/contributing_guide.html#deprecating-and-removing-code) be enough for this, or do you think a special exception is warranted? If more time is needed, let's try to indicate the release in which support should be removed in the error message or perhaps a code comment.

python/cudf/cudf/io/json.py
```diff
@@ -37,16 +37,31 @@ def read_json(
             f"or a bool, or None. Got {type(dtype)}"
         )
 
-    if engine == "cudf" and not lines:
+    if engine == "cudf_experimental":
+        raise ValueError(
+            "engine='cudf_experimental' support has been removed, "
+            "use `engine='cudf'`"
+        )
```
bdice, line 44: Should this be a hard error or a deprecation warning that replaces the value with `"cudf"`?

python/cudf/cudf/io/json.py
```diff
@@ -97,7 +112,7 @@ def read_json(
             lines,
             compression,
             byte_range,
-            engine == "cudf_experimental",
+            engine == "cudf_legacy",
```
bdice, line 115: This is switching from the new reader to the old reader. Should this be:
```suggestion
            engine == "cudf",
```

python/cudf/cudf/tests/test_json.py
```diff
@@ -580,17 +587,18 @@ def test_default_float_bitwidth(default_float_bitwidth):
...
         "c1": [{"f1": "sf11", "f2": "sf21"}, {"f1": "sf12", "f2": "sf22"}],
         "c2": [["l11", "l21"], ["l12", "l22"]],
     }
     pdf = pd.DataFrame(data)
-    pdf.to_json(fname, orient="records")
+    pdf.to_json(bytes_obj, orient="records")
 
-    df = cudf.read_json(fname, engine="cudf_experimental", orient="records")
-    pdf = pd.read_json(fname, orient="records")
+    df = cudf.read_json(bytes_obj, engine="cudf", orient="records")
```
vuule, line 599: same here, can we remove the parameter wherever the default value is set?
```suggestion
    df = cudf.read_json(bytes_obj, orient="records")
```

vuule (changes requested):
Looks good.
The only suggestion is to remove all `engine='cudf'` and `engine="cudf"` :)

python/dask_cudf/dask_cudf/io/tests/test_json.py
```diff
@@ -70,7 +70,7 @@ def test_read_json_lines(lines):
 
 
 def test_read_json_nested_experimental(tmp_path):
```
bdice, line 72: Can we rename the tests, too? The test names referencing `experimental` should be updated.
```suggestion
def test_read_json_nested(tmp_path):
```

python/cudf/cudf/io/json.py
galipremsagar: Since it was experimental, I feel we have the flexibility to make this an error, what do you think?

python/cudf/cudf/io/json.py
galipremsagar: The cython API changes too in this PR, that's why it has to be `engine=='cudf_legacy'`

python/cudf/cudf/io/json.py
galipremsagar: Yeah, we definitely want to give this a special exception. I don't think we have a specific release decided yet to remove it completely. I updated the comment here.

cc: @GregoryKimball @vuule too.

docs/cudf/source/user_guide/io/read-json.md
galipremsagar: Dropped in docs 👍

python/cudf/cudf/io/json.py
vuule: I think we'll pick the removal release based on 23.02 feedback. In general, we'll remove the old reader as soon as there are no user issues specific to the new one.

python/cudf/cudf/io/json.py
bdice: I think that's justifiable -- but we'll want to delete this error at some point later. Therefore, it's the same amount of work for us as developers to deprecate it as to force an error. "Add warning, delete warning later" vs. "add error, delete error later."

python/cudf/cudf/io/json.py
bdice: That sounds reasonable! I just wanted to make sure we discussed this topic before merging. 👍

python/cudf/cudf/tests/test_json.py
galipremsagar: Unfortunately, we can't. This is because the default for `lines` is `False` and our new json reader doesn't seem to support few `lines=False` cases for which when `engine='auto'` we toggle to `pandas` when `lines=False`:
```python
if engine == "auto":
        engine = "cudf" if lines else "pandas"
```

So to avoid switching to pandas reader in the above scenario, we will have explicitly pass `engine='cudf'` when `lines=False`. We still have support for `lines=False` in the new JSON reader but today we don't seem to have complete support.

python/cudf/cudf/io/json.py
bdice: Wow, my mistake. I misread this as a parameter assignment `engine="cudf_legacy"` and not an equality check.

docs/cudf/source/user_guide/io/read-json.md
```diff
@@ -49,7 +49,7 @@ reading nested JSON data.
     {"list": [0,1,2], "struct": {"k":"v1"}},
     {"list": [3,4,5], "struct": {"k":"v2"}}
 ]'''
->>> df = cudf.read_json(j, engine='cudf_experimental')
+>>> df = cudf.read_json(j)
```
galipremsagar, line 52:
```suggestion
>>> df = cudf.read_json(j, engine='cudf')
```

python/cudf/cudf/tests/test_json.py
bdice: Is there an issue that can be opened / tracked / tagged here?

galipremsagar: @karthikeyann would know

python/dask_cudf/dask_cudf/io/json.py
```diff
@@ -49,7 +49,7 @@ def read_json(url_path, engine="auto", **kwargs):
 
     Load nested JSON data
 
-    >>> read_json('myfile.json', engine='cudf_experimental')  # doctest: +SKIP
+    >>> read_json('myfile.json', engine='cudf')  # doctest: +SKIP
```
bdice, line 52:
```suggestion
    >>> read_json('myfile.json')  # doctest: +SKIP
```

bdice (approved):
Looks good to me, once conversations are resolved.

python/dask_cudf/dask_cudf/io/json.py
galipremsagar: I think we want this to be `cudf` instead of `auto`, for the same reason as: https://github.com/rapidsai/cudf/pull/12509/files#r1083132742

python/cudf/cudf/tests/test_json.py
vuule: Would be nice to remark on the default engine selection in the PR description, I completely missed that this PR impacts only JSON Lines.

galipremsagar: One of the case our reader can't support:
[lines_false.json.txt](https://github.com/rapidsai/cudf/files/10470954/lines_false.json.txt)

vuule (approved):
Looks good. Looking forward to future removal of `engine='cudf'`s :)

python/cudf/cudf/tests/test_json.py
vuule: Isn't this a different orient than what we support?
After 23.02 work is done, we need to figure out how the orient parameter fits into the new parser. For one, it would allow us to avoid falling back to pandas for supported orients, even with lines=false.

galipremsagar: > Would be nice to remark on the default engine selection in the PR description, I completely missed that this PR impacts only JSON Lines.

Done 👍 Added a note in docstring too. We emit a warning in our code but never had it mentioned clearly in our docs.

galipremsagar:
> Looks good. Looking forward to future removal of `engine='cudf'`s :)

Which also means the removal of `engine='pandas'`, actually the `engine` parameter entirely 😉

python/dask_cudf/dask_cudf/io/tests/test_json.py
```diff
@@ -82,6 +82,6 @@ def test_read_json_nested_experimental(tmp_path):
     kwargs = dict(orient="records", lines=True)
     with tmp_path / "data.json" as f:
         df.to_json(f, **kwargs)
-        actual = dask_cudf.read_json(f, engine="cudf_experimental", **kwargs)
+        actual = dask_cudf.read_json(f, engine="cudf", **kwargs)
```
rjzamora, line 85: If `engine="cudf"` now works, then I'd also like to be testing the `actual = dask_cudf.read_json(f, **kwargs)` case here as well.

python/dask_cudf/dask_cudf/io/json.py
galipremsagar: Yup, this actually worked. Checked locally and added a test for this.

python/dask_cudf/dask_cudf/io/tests/test_json.py
galipremsagar: Added this case too 👍

python/dask_cudf/dask_cudf/io/tests/test_json.py
```diff
@@ -82,6 +82,8 @@ def test_read_json_nested_experimental(tmp_path):
     kwargs = dict(orient="records", lines=True)
```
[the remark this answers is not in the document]
galipremsagar: Sounds good, Made the addition 👍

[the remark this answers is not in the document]
rjzamora: **Another Note**: If we feel it is simply "best practice" to specify `orient="records"` and `lines=True` explicitly, then I'm also okay with just adding those args to the doc-string example. However, it is probably best to for our tests to establish clear defaults.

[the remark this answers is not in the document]
galipremsagar: > **Another Note**: If we feel it is simply "best practice" to specify `orient="records"` and `lines=True` explicitly, then I'm also okay with just adding those args to the doc-string example. However, it is probably best for our tests to establish clear defaults.

Not necessary, as our json reader works for other values of those two parameters too.

rjzamora (approved):
Thanks @galipremsagar - Looks good on the dask side!

karthikeyann:
🚀

PR #12526, opened 2023-01-10
Update libcudf reduction docs for casting output types

### LINKED ISSUE #9988: [BUG] Reductions are not performed in the output precision.
**Describe the bug**
The `cudf::reduce` algorithm takes a parameter `cudf::data_type output_dtype`. The docs state that the reduction (computation) is performed in the output precision:
https://github.com/rapidsai/cudf/blob/7392f9f5f10dc6efb4d21cfcef18a14a0df421c3/cpp/include/cudf/reduction.hpp#L62

However, the computation precision does _not_ match the output precision. The `element_type_dispatcher`'s implementation of `reduce_numeric` casts floating-point values to `double` before performing the reduction. https://github.com/rapidsai/cudf/blob/7392f9f5f10dc6efb4d21cfcef18a14a0df421c3/cpp/src/reductions/simple.cuh#L364-L365

Similarly, `reduce_numeric` casts integral values to `int64_t` before performing the reduction. 
https://github.com/rapidsai/cudf/blob/7392f9f5f10dc6efb4d21cfcef18a14a0df421c3/cpp/src/reductions/simple.cuh#L387-L388

This code should be refactored to match the docs - we _should_ perform computations in the output precision, to avoid overflow and floating-point truncation issues. This invokes UB if the output type is `uint64_t` and the reduction exceeds `INT64_MAX`, for example. The type dispatch on `output_dtype` should be performed at the place where the reduction functor is called, rather than having a type dispatch to (down)cast the results at the end:
https://github.com/rapidsai/cudf/blob/7392f9f5f10dc6efb4d21cfcef18a14a0df421c3/cpp/src/reductions/simple.cuh#L391-L396

This also affects segmented reductions in #9621.

davidwendt:
## Description
Fixes documentation on how `cudf::reduce` and `cudf::segmented_reduce` function when the input column type does not match the specified output data type. When the input and output types match, the reduction is performed using that type. When the types do not match, the input is promoted to either `double` or `int64_t` as appropriate through a transform-input-iterator when computing the reduction. The output is then cast to the specified output data type. This works for unsigned  values as well. For UINT64 types, this works for all values where the result would be <= max(int64_t). For values where the result may be > max(int64_t) the `output_dtype == UINT64` and falls into the case where the types match and reduction is performed on that type.
This PR changes only the comments and doxygen for the reduce APIs.

Closes #9988 

## Checklist


bdice (approved):
Everything looks good. I noticed this was in draft after reviewing. If there are more changes, I will re-review, but I think this was the complete scope of what we discussed. Thank you very much!

davidwendt:
> Everything looks good. I noticed this was in draft after reviewing. If there are more changes, I will re-review, but I think this was the complete scope of what we discussed. Thank you very much!

I left it in draft in case there was a major concern. I just wanted you to review it first. I can put it in review mode now.

cpp/src/reductions/simple_segmented.cuh
```diff
@@ -465,11 +476,12 @@ struct column_type_dispatcher {
                                      rmm::cuda_stream_view stream,
                                      rmm::mr::device_memory_resource* mr)
   {
+    // If the output type matches the input type, then reduce using that type
     if (output_type.id() == cudf::type_to_id<ElementType>()) {
       return simple_segmented_reduction<ElementType, ElementType, Op>(
         col, offsets, null_handling, init, stream, mr);
     }
-    // reduce and map to output type
+    // otherwise, reduce and map to output type
```
ttnghia, line 484:
```suggestion
    // Otherwise, reduce and map to output type.
```

ttnghia (approved):

PR #12530, opened 2023-01-10
Fix compute sanitizer in GitHub Actions CI

bdice:
## Description
@davidwendt noticed that the [nightly builds](https://github.com/rapidsai/cudf/actions/workflows/test.yaml) were passing unexpectedly on GitHub Actions, after they had been regularly failing on a `compute-sanitizer --tool memcheck` step in gpuCI.

https://github.com/rapidsai/cudf/actions/runs/3880698576/jobs/6618987835#step:6:66543

After some investigation, I found the job was failing -- but for the wrong reason, and silently.

```
RAPIDS logger » [01/10/23 06:24:39]
┌──────────────────────────────────────────┐
|    Memcheck gtests with rmm_mode=cuda    |
└──────────────────────────────────────────┘

Running gtest ARROW_IO_SOURCE_TEST
ci/test_cpp.sh: line 95: compute-sanitizer: command not found
...
```

In this PR, I add the `cuda-sanitizer-api` package to our C++ test dependencies, which should fix the nightly tests (so they fail as expected). I also fixed a script bug that let the tests pass silently despite having failures.

## Checklist

### DIFF
--- a/ci/test_cpp.sh
+++ b/ci/test_cpp.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 
 set -euo pipefail
 
@@ -93,6 +93,11 @@ if [[ "${RAPIDS_BUILD_TYPE}" == "nightly" ]]; then
         fi
         echo "Running gtest $test_name"
         ${COMPUTE_SANITIZER_CMD} ${gt} | tee "${RAPIDS_TESTS_DIR}${test_name}.cs.log"
+        exitcode=$?
+        if (( ${exitcode} != 0 )); then
+            SUITEERROR=${exitcode}
+            echo "FAILED: GTest ${gt}"
+        fi
     done
     unset GTEST_CUDF_RMM_MODE
     # TODO: test-results/*.cs.log are processed in gpuci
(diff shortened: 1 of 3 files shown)


davidwendt (changes requested):
Could we make this only run on one configuration? Say just one of the 11.8 ones perhaps?
`conda-cpp-tests / tests (11.8.0, ubuntu20.04, arm64, 3.10, a100, 520)`

davidwendt:
According to the Jenkins logs for nightly runs, the compute-sanitizer step takes over 3 hours to run so it would be good to only run it for a single configuration and not all of them.

bdice:
Thanks for that info @davidwendt, I didn't realize it took so long. Good idea to split it up and limit the configurations.

When I get a chance (may not be until next week) I will split the `compute-sanitizer` portion of the C++ test script into a separate file, add it to a separate workflow job using a custom script path (rather than having it be part of the C++ matrix tests), and have this launch only on one runner configuration, which we label `gpu-latest-1`. We also run Java tests and notebooks tests on `gpu-latest-1` using a custom script, because they only need one configuration.

wence-:
Just to note that I think we also want to, if we can, run with `compute-sanitizer --tool initcheck` (I think this is even slower than `memcheck` so running only on a single configuration is also necessary here)

davidwendt:
> Just to note that I think we also want to, if we can, run with `compute-sanitizer --tool initcheck` (I think this is even slower than `memcheck` so running only on a single configuration is also necessary here)

I would not recommend adding `initcheck` since I think it cannot reliably verify usage of uninitialized memory.
https://github.com/rapidsai/cudf/issues/12667#issuecomment-1412646551

bdice:
Closing in favor of #12800 (already merged). Thanks @davidwendt for picking this up.

PR #12533, opened 2023-01-11
Use new nvCOMP API to optimize the compression temp memory size

vuule:
## Description
nvCOMP 2.6 introduces expanded API to determine the amount of temporary memory required to compress a set of chunks. New API also takes the total size of the uncompressed data.
This PR expands the nvCOMP adapter to use this API when available (nvCOMP version >= 2.6). With older versions the older API is still used and there is no behavior change.

## Checklist


jbrennan333 (commented):
Code changes look good to me.

jbrennan333 (approved):
This looks good to me.

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -37,9 +37,11 @@
 
 #define NVCOMP_HAS_DEFLATE(MAJOR, MINOR, PATCH) (MAJOR > 2 or (MAJOR == 2 and MINOR >= 5))
 
-#define NVCOMP_HAS_TEMPSIZE_EX(MAJOR, MINOR, PATCH) \
+#define NVCOMP_HAS_DECOMP_TEMPSIZE_EX(MAJOR, MINOR, PATCH) \
   (MAJOR > 2 or (MAJOR == 2 and MINOR > 3) or (MAJOR == 2 and MINOR == 3 and PATCH >= 1))
 
+#define NVCOMP_HAS_COMP_TEMPSIZE_EX(MAJOR, MINOR, PATCH) (MAJOR > 2 or (MAJOR == 2 and MINOR >= 6))
```
ttnghia, line 43: So the com and decomp are offered by different nvcomp versions and can be used differently?

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -216,6 +218,63 @@ auto batched_compress_temp_size(compression_type compression,
...
+                                  size_t max_uncomp_chunk_size,
+                                  size_t max_total_uncomp_size)
+{
+#if NVCOMP_HAS_COMP_TEMPSIZE_EX(NVCOMP_MAJOR_VERSION, NVCOMP_MINOR_VERSION, NVCOMP_PATCH_VERSION)
+  // Ignore errors in the expanded version; fall back to the old API in case of failure
+  try {
+    return batched_compress_get_temp_size_ex(
+      compression, num_chunks, max_uncomp_chunk_size, max_total_uncomp_size);
+  } catch (...) {
+  }
```
ttnghia, line 272: You catch but you don't throw so the code will continue even there was exception thrown. Is this expected? If so please add a comment to clarify that.

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -336,10 +395,15 @@ void batched_compress(compression_type compression,
 
   auto nvcomp_args = create_batched_nvcomp_args(inputs, outputs, stream);
 
-  auto const max_uncomp_chunk_size = skip_unsupported_inputs(
+  skip_unsupported_inputs(
     nvcomp_args.input_data_sizes, results, compress_max_allowed_chunk_size(compression), stream);
 
-  auto const temp_size = batched_compress_temp_size(compression, num_chunks, max_uncomp_chunk_size);
+  auto [max_uncomp_chunk_size, total_uncomp_size] =
```
ttnghia, line 401: `const`?

cpp/src/io/comp/nvcomp_adapter.cpp
vuule: There's a comment above the try block. Should it be placed elsewhere?

ttnghia: Oh sorry I really didn't see it. So I think the comment should be inside the `catch` block instead.

cpp/src/io/comp/nvcomp_adapter.cpp
vuule: yes, they were added in different releases (2.3.1 and 2.6).
They can be used independently, there's no interaction between the two API sets.

ttnghia (approved):

cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -178,10 +180,10 @@ void batched_decompress(compression_type compression,
   update_compression_results(nvcomp_statuses, actual_uncompressed_data_sizes, results, stream);
 }
 
-// Dispatcher for nvcompBatched<format>CompressGetTempSize
-auto batched_compress_temp_size(compression_type compression,
-                                size_t batch_size,
-                                size_t max_uncompressed_chunk_bytes)
+// Wrapper for nvcompBatched<format>CompressGetTempSize
```
mythrocks, line 183: :+1:

cpp/src/io/comp/nvcomp_adapter.cu
```diff
@@ -94,10 +94,10 @@ void update_compression_results(device_span<size_t const> actual_output_sizes,
     [] __device__(auto const& results) { return results.status != compression_status::SKIPPED; });
 }
 
-size_t skip_unsupported_inputs(device_span<size_t> input_sizes,
-                               device_span<compression_result> results,
-                               std::optional<size_t> max_valid_input_size,
-                               rmm::cuda_stream_view stream)
+void skip_unsupported_inputs(device_span<size_t> input_sizes,
```
mythrocks, line 97: Not sure why the previous version of this function was both skipping *and* returning the sum.
This version is better.

mythrocks (approved):
LGTM!

cpp/src/io/comp/nvcomp_adapter.cu
vuule: It's because I didn't want to add a function that would just be a wrapper around `thrust::reduce`.
The function now wraps TWO `thrust::reduce`s so it's okay 🙈

PR #12534, opened 2023-01-11
Dynamically read PTX version

brandon-b-miller:
Fixes an issue that came up during https://github.com/rapidsai/cudf/pull/12457 where the PTX version needed to be bumped. 

cc @bdice @gmarkall


python/strings_udf/strings_udf/__init__.py
```diff
@@ -14,8 +14,48 @@
...
+        raise ValueError("Could not read cuda version " "from ptx file")
+    version = ver_line.strip("\n").split(" ")[-1]
+    ptx_major, ptx_minor = version.split(".")
+
+    # from ptx_docs/release_notes above:
+    maj_ver_map = {"7": "11", "8": "12"}
+    cuda_major = maj_ver_map.get(ptx_major)
+    if cuda_major is None:
+        raise ValueError(
+            "Could not map major PTX version " f"{ptx_major} to a CUDA version"
```
bdice, line 55:
```suggestion
            f"Could not map major PTX version {ptx_major} to a CUDA version"
```

python/strings_udf/strings_udf/__init__.py
```diff
@@ -14,8 +14,48 @@
...
+        lines = f.readlines()
+    ver_line = None
+    for line in lines:
+        if ".version" in line:
+            ver_line = line
+            break
+    if ver_line is None:
+        raise ValueError("Could not read cuda version " "from ptx file")
+    version = ver_line.strip("\n").split(" ")[-1]
+    ptx_major, ptx_minor = version.split(".")
```
bdice, line 48: If a patch version is introduced, this will fail to unpack the values. Maybe avoid assuming the number of components?
```suggestion
    ptx_major = version.split(".")[0]
    ptx_minor = version.split(".")[1]
```

python/strings_udf/strings_udf/__init__.py
```diff
@@ -14,8 +14,48 @@
...
+
+    # from ptx_docs/release_notes above:
+    maj_ver_map = {"7": "11", "8": "12"}
+    cuda_major = maj_ver_map.get(ptx_major)
+    if cuda_major is None:
+        raise ValueError(
+            "Could not map major PTX version " f"{ptx_major} to a CUDA version"
+        )
+
+    return (int(cuda_major), int(ptx_minor))
```
bdice, line 58: These parens are extraneous.
```suggestion
    return int(cuda_major), int(ptx_minor)
```

python/strings_udf/strings_udf/__init__.py
```diff
@@ -14,8 +14,48 @@
 
 logger = get_logger()
 
-# tracks the version of CUDA used to build the c++ and PTX components
-STRINGS_UDF_PTX_VERSION = (11, 8)
+
+def get_cuda_version_from_ptx_file(path):
```
bdice, line 18: Should this be private?
```suggestion
def _get_cuda_version_from_ptx_file(path):
```

bdice (commented):
Good idea to fix this! Some suggestions attached.

python/strings_udf/strings_udf/__init__.py
```diff
@@ -26,11 +65,11 @@ def _get_appropriate_file(sms, cc):
         return None
 
 
-def maybe_patch_numba_linker(driver_version):
+def maybe_patch_numba_linker(driver_version, ptx_version):
```
gmarkall, line 68: Only a minor nit, but it's not really the PTX version you're passing in here, it's the toolkit version for the PTX - maybe `ptx_toolkit_version` is slightly closer to what the parameter is. Or, `min_driver_version` is what it represents.

python/strings_udf/strings_udf/__init__.py
```diff
@@ -14,8 +14,47 @@
...
+
+    """
+    with open(path) as ptx_file:
+        for line in ptx_file:
+            if line.startswith(".version"):
+                ver_line = line
+                break
+        else:
+            raise ValueError("Could not read CUDA version from ptx file.")
+    version = ver_line.strip("\n").split(" ")[-1]
```
bdice, line 45: No need to specify the strip characters. And I'd rather get element 1 than -1, in case there's additional garbage at the end of the line for some reason.
```suggestion
    version = ver_line.strip().split(" ")[1]
```

python/strings_udf/strings_udf/__init__.py
```diff
@@ -14,8 +14,47 @@
...
+
+    # from ptx_docs/release_notes above:
+    maj_ver_map = {"7": "11", "8": "12"}
+    cuda_major = maj_ver_map.get(ptx_major)
+    if cuda_major is None:
+        raise ValueError(
+            f"Could not map major PTX version {ptx_major} to a CUDA version"
+        )
+
+    return int(cuda_major), int(ptx_minor)
```
bdice, line 57: Are we guaranteed that the PTX minor version matches the CUDA minor version?

brandon-b-miller: In the docs this pattern holds for the cuda versions we support but doesn't seem to be guaranteed anywhere. While thinking about this I also realized we might be better off just mapping directly from the full version string in the file to a full cuda version to avoid confusing things if the pattern does not hold in the future. I made some changes in https://github.com/rapidsai/cudf/pull/12534/commits/49a210eba43b0c7e4414318d5a0490d4bfd07905, what do you think?

bdice: Seems fine to me! Fewer assumptions is better. :)

Do we need mappings for CUDA 11.0, 11.1? We don't support those in RAPIDS anymore. We also require CUDA 11.5 to build cudf, so if the PTX version is guaranteed to be at least the minimum version to build, we could drop 11.2-11.4 as well.

bdice (approved):
Thanks for the iterations!

python/strings_udf/strings_udf/__init__.py
brandon-b-miller: Dropped those ones!

brandon-b-miller:
> Thanks for the iterations!

Ty for review :)

PR #12538, opened 2023-01-11
Update List Lexicographical Comparator

### LINKED ISSUE #12298: Cudf vs Spark discrepancies when comparing nested list column with null entries
When we test `order by` on array of array (of short) with null entries, we find the following discrepancy.
For Spark,
`[[None], [-21827]] < [[None, None], [40, -21636], [None, None, -7359, 0], [13408], [-25000, 21133, -14473]]`
For Cudf, the above is false.

Due to this issue, we cannot support `order by` on nested list column (e.g. array of array type).

**Additional context**
Cudf supports nested list column comparison https://github.com/rapidsai/cudf/pull/11129.

### LINKED ISSUE #12493: [BUG] `stable_sorted_order/sorted_order(..., null_order::AFTER)` producing wrong results for lists
**Describe the bug**
`stable_sorted_order/sorted_order` always produce results with `null_order::BEFORE` even when explicitly passing `null_order::AFTER` with list column

```
lists_col col{{{}, {1}, {2, 2}, {2, 3}, {}, {} /*NULL*/, {2}, {} /*NULL*/}, nulls_at({5, 7})};
auto const input  = cudf::table_view{{col}};

auto const stable_order = cudf::stable_sorted_order(input, {cudf::order::ASCENDING}, {null_order::AFTER});
auto const stable_sorted_input = cudf::gather(input, *stable_order);
```

**Expected behavior**
NULLs to be pushed to the end, but they are being pushed to the front

cc @bdice

divyegala:
Closes #12493, Closes #12298 

## Description
The definitive order for sorting that is followed in this PR can be explained with the following:
```
Possible elements in List<List<int>> = NULL, [], [NULL], [[]], [[NULL]], [[int]]
```
`ASCENDING` sort with `null_order::BEFORE`:
```
NULL < [] < [NULL] < [[]] < [[NULL]] << [[int]]
```
`ASCENDING` sort with `null_order::AFTER`:
```
[] < [NULL] < [[]] < [[NULL]] << [[int]] < NULL
```
Only top-level `NULLs` are ordered by `null_order`, and at any arbitrary nesting level `[] < [NULL] < [int]`.

Note: `DESCENDING` would be the exact opposite order

Benchmarks present [below](https://github.com/rapidsai/cudf/pull/12538#issuecomment-1397264372)

## Checklist


ttnghia:
From the code change, I guess there may be some impact on performance when comparing lists. Can you run some benchmarks on the code, please? There already exists `nvbench_sort_lists` that is ready to run immediately.

divyegala:
@ttnghia yes, sure, I can run them. However, the previous implementation was wrong and as such, the benchmarks that currently exist should not form a baseline for performance. We should only compare when we have established a baseline that provides correct results.

Still, I will run them anyway.

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -442,43 +446,92 @@ class device_row_comparator {
...
+        auto const r_rep_level = r_rep_levels[r_dremel_index];
+
+        // only compare if left and right are at same nesting level
+        if (l_rep_level == r_rep_level) {
+          // check if left element or right is an empty list
+          auto const l_def_level         = l_def_levels[l_dremel_index];
+          auto const r_def_level         = r_def_levels[r_dremel_index];
+          bool const is_left_list_empty  = l_global_empties[l_dremel_index];
+          bool const is_right_list_empty = r_global_empties[r_dremel_index];
+          bool const is_left_list_null   = l_def_level < l_max_def_level && not is_left_list_empty;
```
ttnghia, line 462: Don't mix logical ops like that (`&& not`).

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -442,43 +446,92 @@ class device_row_comparator {
...
+            return _null_precedence == null_order::BEFORE
+                     ? cuda::std::pair(weak_ordering::LESS, _depth)
+                     : cuda::std::pair(weak_ordering::GREATER, _depth);
+          }
+          // compare right null to left leaf
+          else if (is_right_list_null && l_def_level == l_max_def_level) {
+            return _null_precedence == null_order::BEFORE
+                     ? cuda::std::pair(weak_ordering::GREATER, _depth)
+                     : cuda::std::pair(weak_ordering::LESS, _depth);
+          }
```
ttnghia, line 523: I wonder if all these if-else can be simplified somehow? They are too long, too many branches which is very bad in GPU processing.

divyegala: Yes, we might just scrap this complicated logic for now and go with what Spark expects in terms of ordering. That will be much simpler looking.

mythrocks:
Nitpick, or maybe I haven't understood the example.
```
List<List<int>> = [
[], [[], [1]], [NULL], [[]]
]
global_empties = 
[1, 1, 0, 0, 1]
```

Aren't there only 4 rows in this `List<List<int>>` column? Shouldn't there be only 4 entries in `global_empties?

divyegala:
@mythrocks we might end up scraping that extra bookkeeping because there is no real use-case yet for ordering NULLs and empties at depth > 0. But to answer for your curiosity:

Dremel encoding applies a value to every single element in all of the rows of the column, (`number of leafs + number of empties at level 1 + number of NULLs at level 1 + ...`). So, there are 5 values in `global_empties` but also in `rep_level` and `def_level`

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -442,43 +452,45 @@ class device_row_comparator {
...
-          state = (l_rep_levels[l_dremel_index] < r_rep_levels[r_dremel_index])
-                    ? weak_ordering::LESS
-                    : weak_ordering::GREATER;
-          return cuda::std::pair(state, _depth);
-        }
+        auto const l_rep_level = l_rep_levels[l_dremel_index];
+        auto const r_rep_level = r_rep_levels[r_dremel_index];
+
+        // only compare if left and right are at same nesting level
+        if (l_rep_level == r_rep_level) {
```
bdice, line 462: Can you write this as an "early exit" so that the `else` block isn't hidden far below? It should also help reduce an indentation level for the body of this `if`.

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -442,43 +452,45 @@ class device_row_comparator {
...
+            // most nested NULLs and leafs are present
+            // In this rare condition that we get to the most nested NULL, we increment
+            // element_index because either both rows have a deeply nested NULL at the
+            // same position, and we'll "continue" in our iteration, or we will early
+            // exit if only one of the rows has a deeply nested NULL
+            if (lcol.nullable() and l_def_levels[l_dremel_index] == l_max_def_level - 1) {
+              ++element_index;
+            }
+            if (l_def_level == r_def_level) { continue; }
+            // [] < [NULL] < [leaf]
```
bdice, line 478:
```suggestion
            // We require [] < [NULL] < [leaf] for nested nulls. The null_precedence only affects top level nulls.
```

cpp/include/cudf_test/print_utilities.cuh
```diff
@@ -25,7 +25,7 @@
 
 #include <type_traits>
 
-namespace cudf::test::print {
+namespace cudf::test {
```
bdice, line 28: These seem like unrelated changes. Are these necessary for this PR?

divyegala: Well, no, but it's a conflict right now if you include `cudf_test/column_utilities.hpp` and `cudf_test/print_utilities.hpp` because the former has a function called `cudf::test::print` and the latter has a namespace called `cudf::test::print`. Printing the array was a useful utility for me so I figured I should fix it

bdice: Ouch. Can you separate that into another PR? I can quickly approve that one, but I want to keep the comparators work relatively clean because of its high complexity.

cpp/include/cudf_test/print_utilities.cuh
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
PointKernel, line 2:
```suggestion
 * Copyright (c) 2022, NVIDIA CORPORATION.
```

Looks like a leftover.

bdice (approved):
This looks right to me. Thanks for extending the test coverage and running benchmarks.

ttnghia (approved):
The benchmark looks good 👍

Issue #7108, opened 2021-01-09
Feature request: add additional flags to to_csv() function to avoid csv writer adding additional quotation marks 


ogreen:
**Is your feature request related to a problem? Please describe.**
When using `to_csv` function for a string column, the csv writer automatically adds "quotation" signs to complete possibly missing signs. This can potentially add undesirable quotation marks.
For example the string
`"abcd`
will be written in the CSV as
`"""abcd"`

**Describe the solution you'd like**
Ideally, `to_csv` should duplicate the quotation functionality found in `read_csv`
This documentation can be found in the following [link](https://docs.rapids.ai/api/cudf/stable/api.html?highlight=read_csv#strings).

quoting: str or int, default 0
Controls quoting behavior. Set to one of 0 (csv.QUOTE_MINIMAL), 1 (csv.QUOTE_ALL), 2 (csv.QUOTE_NONNUMERIC) or 3 (csv.QUOTE_NONE). Quoting is enabled with all values except 3.

CCing @vuule .

PR #12539, opened 2023-01-12
Make string quoting optional on CSV write

mythrocks:
## Description
`cudf::io::write_csv()` currently adds quotes around all `STRING` rows that contain "special" characters, including:
1. `,`, or the field delimiter
2. `\n`, or the row/record delimiter
3. `"`, or double quotes

This is in congruence with Pandas conventions for CSV, and allows for strings with such special characters to be read correctly.

If `write_csv()` is used to implement writes for CSV-like formats (such as Hive delimited text files), one notices that quoting conventions would be at odds. Hive delimited text does not automatically quote problematic strings.

This commit adds the option of turning off quoting for `STRING` columns. The erstwhile behaviour is retained as the default; the user must opt into turning off quoting.

Additionally, this commit adds the following:
1. Tests for optionally unquoted reads, in C++ and Java.
2. JNI interfaces for CSV writes with quoting disabled.
3. Changes to Java `CSVOptions` to enable reads with quoting disabled.
4. Minor corrections to Java `TableTest.testWriteCSVToFile()` test combinations of `includeHeader`, `fieldDelim`, and `true`/`false` string representations.

## Checklist

### DIFF
--- a/cpp/include/cudf/io/csv.hpp
+++ b/cpp/include/cudf/io/csv.hpp
@@ -1340,6 +1340,8 @@ class csv_writer_options {
   std::string _false_value = std::string{"false"};
   // Names of all columns; if empty, writer will generate column names
   std::vector<std::string> _names;
+  // Quote string rows, if value contains delimiter/double-quote/newline.
+  bool _quote_tricky_strings = true;
 
   /**
    * @brief Constructor from sink and table.
@@ -1442,6 +1444,8 @@ class csv_writer_options {
    */
   [[nodiscard]] std::string get_false_value() const { return _false_value; }
 
+  [[nodiscard]] bool is_enabled_quote_tricky_strings() const { return _quote_tricky_strings; }
+
   // Setter
   /**
    * @brief Sets optional associated column names.
@@ -1505,6 +1509,8 @@ class csv_writer_options {
    * @param table Table to be written
    */
   void set_table(table_view const& table) { _table = table; }
+
+  void quote_tricky_strings(bool is_enabled_quoting) { _quote_tricky_strings = is_enabled_quoting; }
 };
 
 /**
@@ -1625,6 +1631,12 @@ class csv_writer_options_builder {
   csv_writer_options_builder& false_value(std::string val)
   {
     options._false_value = val;
+    return *this;
+  }
+
+  csv_writer_options_builder& quote_tricky_strings(bool is_enabled_quoting)
+  {
+    options._quote_tricky_strings = is_enabled_quoting;
     return *this;
   }
 
--- a/cpp/include/cudf/io/csv.hpp
+++ b/cpp/include/cudf/io/csv.hpp
@@ -1653,6 +1653,7 @@ class csv_writer_options_builder {
    * should be quoted
    *
    * @param is_enabled_quoting Set to `true` to enable quoting, `false` otherwise
+   * @return this for chaining
    */
   csv_writer_options_builder& enable_quote_strings(bool is_enabled_quoting)
   {
--- a/cpp/include/cudf/io/csv.hpp
+++ b/cpp/include/cudf/io/csv.hpp
@@ -1340,8 +1340,8 @@ class csv_writer_options {
   std::string _false_value = std::string{"false"};
   // Names of all columns; if empty, writer will generate column names
   std::vector<std::string> _names;
-  // Quote string rows, if value contains delimiter/double-quote/newline.
-  bool _quote_strings = true;
+  // Quote style. Currently only MINIMAL and NONE are supported.
+  quote_style _quoting = quote_style::MINIMAL;
 
   /**
    * @brief Constructor from sink and table.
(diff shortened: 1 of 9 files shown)


cpp/include/cudf/io/csv.hpp
```diff
@@ -1340,6 +1340,8 @@ class csv_writer_options {
   std::string _false_value = std::string{"false"};
   // Names of all columns; if empty, writer will generate column names
   std::vector<std::string> _names;
+  // Quote string rows, if value contains delimiter/double-quote/newline.
+  bool _quote_strings = true;
```
davidwendt, line 1344: Can this be an enum for the purposes of https://github.com/rapidsai/cudf/issues/7108?
It can be just two values for this PR. Unless you want to consider supporting all 4 quote-types in this PR?

mythrocks: I've gone back and forth about this. I considered using `quote_style` here instead of a `bool`. What should we do if a caller uses an unexpected/unsupported `quote_style`?

(Supporting all 4 quote styles is beyond scope. :] For Hive text support, all I'm looking to do is turn off quoting without breaking anyone else.)

mythrocks: I'll take another crack at using `quote_style`.

davidwendt: > I've gone back and forth about this. I considered using `quote_style` here instead of a `bool`. What should we do if a caller uses an unexpected/unsupported `quote_style`?

Just don't declare more styles than you actually support?
You are coding style `QUOTE_NONE` and the current implementation is style `QUOTE_NONNUMERIC` (I think).
Only provide those two styles and default to the current one.

mythrocks: Do you mean use a different enum entirely? Would that not be confusing?

davidwendt: I'm not sure what you are asking. There is no enum currently. You get to make it up here.

mythrocks: Oh, wait. Maybe I'm not being clear. 
We already have a `quote_style` enum in [`cudf/io/types.hpp`](https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/include/cudf/io/types.hpp#L85):
```c++
enum class quote_style {
  MINIMAL,     ///< Quote only fields which contain special characters
  ALL,         ///< Quote all fields
  NONNUMERIC,  ///< Quote all non-numeric fields
  NONE         ///< Never quote fields; disable quotation parsing
};
```
I would prefer not to introduce a competing enum.

davidwendt: Sorry, I did not realize that one already existed. I agree with not making up a new one.
I would then say use this one and thrown an exception on any that you are not adding support for.

mythrocks: Switched now to use `quote_style`.

davidwendt (approved):

java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
```diff
@@ -31,6 +34,9 @@ public class CSVWriterOptions {
   private String nullValue = "";
   private String falseValue = "false";
   private String trueValue = "true";
+  // Quote style used for CSV data.
+  // Currently, `true` corresponds to `MINIMAL`, `false` to `NONE`.
```
revans2, line 38: I don't think this comment is correct any more.

java/src/main/java/ai/rapids/cudf/QuoteStyle.java
```diff
@@ -0,0 +1,33 @@
...
+ *  limitations under the License.
+ *
+ */
+package ai.rapids.cudf;
+
+/**
+ * Quote style for CSV records, closely following cudf::io::quote_style.
+ */
+enum QuoteStyle {
+    MINIMAL(0),
```
revans2, line 24: nit: would be good to add in similar documentation here that exists for cudf::io::quote_style.

https://github.com/rapidsai/cudf/blob/6dda9d8a95003f932748d578c202857a7facede2/cpp/include/cudf/io/types.hpp#L82-L90

java/src/main/native/src/TableJni.cpp
```diff
@@ -1180,6 +1180,13 @@ JNIEXPORT jlongArray JNICALL Java_ai_rapids_cudf_Table_readCSV(
                                                       static_cast<std::size_t>(buffer_length)} :
                                 cudf::io::source_info{filename.get()};
 
+    auto const quote_style = static_cast<cudf::io::quote_style>(j_quote_style);
+    if (quote_style != cudf::io::quote_style::MINIMAL &&
```
revans2, line 1184: nit: why do we need to enforce this in JNI if cudf is also enforcing it?

java/src/main/native/src/TableJni.cpp
```diff
@@ -1227,6 +1235,12 @@ JNIEXPORT void JNICALL Java_ai_rapids_cudf_Table_writeCSVToFile(
     auto const na_rep = cudf::jni::native_jstring{env, j_null_value};
     auto const true_value = cudf::jni::native_jstring{env, j_true_value};
     auto const false_value = cudf::jni::native_jstring{env, j_false_value};
+    auto const quote_style = static_cast<cudf::io::quote_style>(j_quote_style);
+    if (quote_style != cudf::io::quote_style::MINIMAL &&
```
revans2, line 1239: Same here.

java/src/main/native/src/TableJni.cpp
```diff
@@ -1264,6 +1279,12 @@ JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_Table_startWriteCSVToBuffer(
...
```
revans2, line 1283: and here.

revans2 (approved):
Just a few nits and documentation requests.

java/src/main/java/ai/rapids/cudf/CSVWriterOptions.java
mythrocks: Thank you. I thought I'd gotten it already.

divyegala (approved):

mythrocks:
Thank you for the review, all. This change has been merged.

Issue #12542, opened 2023-01-12
Deprecate cudf::strings::repeat_strings_output_sizes in favor of cudf::repeat_strings throwing size limit error


davidwendt:
The `cudf::strings::repeat_strings_output_sizes` was created so the output sizes of the `cudf::strings::repeat_strings` API could be pre-computed and checked for possible overflow. The `cudf::strings::repeat_strings` can now throw an exception if the output size will exceed the column limit size so this API is no longer needed.

The following API can be removed: https://docs.rapids.ai/api/libcudf/stable/group__strings__copy.html#gad137eea9f5189b3e8c6384c6ab8def40
The optional parameter for this API can be removed: https://docs.rapids.ai/api/libcudf/stable/group__strings__copy.html#ga89b08a3e8f941f7760087cb9e5bf9758

@revans2

ttnghia:
I'll work on Spark plugin to adapt the changes addressed by this issue.

PR #12543, opened 2023-01-12
Rework repeat_strings to use sizes-to-offsets utility

davidwendt:
## Description
Rework `cudf::strings::repeat_strings` to use the internal `sizes_to_offsets` utility. This allows the operation to throw an error if it the output would exceed the size limit of a column. It will also allow deprecating and removing this `repeat_strings_output_sizes` function per issue #12542 

## Checklist


cpp/src/strings/repeat_strings.cu
```diff
@@ -235,6 +235,8 @@ struct compute_size_and_repeat_separately_fn {
  * This function is similar to `strings::detail::make_strings_children`, except that it accepts an
  * optional input `std::optional<column_view>` that can contain the precomputed sizes of the output
  * strings.
+ *
+ * @deprecated This will be removed with issue 12542
```
ttnghia, line 239:
```suggestion
 * @deprecated The parameter `output_strings_sizes` will be removed with issue 12542
```

davidwendt: This whole function can be removed as part of 12542

ttnghia (approved):

PR #12544, opened 2023-01-13
Default `cudf::io::read_json` to nested JSON parser

vuule:
## Description
Changes libcudf default JSON reader to the new nested reader. 
The previous default is now called "legacy", so the API is update to match.
C++ tests that used the default reader still do, except for "values" orient tests (will be addressed in https://github.com/rapidsai/cudf/pull/12498) and few cases where tested behavior is not desired in the new reader.
Modified Python and Java call sites to keep the current behavior. 

## Checklist


ttnghia (approved):

davidwendt (approved):

galipremsagar (approved):
🔥

PR #12545, opened 2023-01-13
Test that cuInit is not called when RAPIDS_NO_INITIALIZE is set

wence-:
## Description

An alternate approach to that tried in #12361, here we just script GDB and check if we hit a breakpoint in cuInit. When RAPIDS_NO_INITIALIZE is set in the environment, merely importing cudf should not call into the CUDA runtime/driver (i.e. no cuInit should be called). Conversely, to check that we are scripting GDB properly, when we create a cudf object, we definitely _should_ hit cuInit.

## Checklist


wence-:
This is an alternate approach to checking that we don't call `cuInit` that doesn't need to hook into dlsym (like #12361). I use GDB to find a breakpoint rather than running `nvprof` because `nvprof` is not supported on devices with compute capability 8 and greater. In those circumstances one is supposed to use `nsys`, but I can't make that work right, in that both with and without `RAPIDS_NO_INITIALIZE` in the environment, I always see `cuInit` in the profiling stack (even though under GDB I do not hit a breakpoint if `RAPIDS_NO_INITIALIZE` is set):

```
$ nsys profile --stats=true --trace=cuda python -c "import cudf;"
Generating '/tmp/nsys-report-fedd.qdstrm'
[1/6] [========================100%] report1.nsys-rep
[2/6] [========================100%] report1.sqlite
[3/6] Executing 'cudaapisum' stats report

 Time (%)  Total Time (ns)  Num Calls  Avg (ns)  Med (ns)  Min (ns)  Max (ns)  StdDev (ns)           Name         
 --------  ---------------  ---------  --------  --------  --------  --------  -----------  ----------------------
     97.9           158717        768     206.7     169.0       114      2233        126.8  cuGetProcAddress      
      1.9             3151          2    1575.5    1575.5      1528      1623         67.2  cuInit                
      0.2              327          2     163.5     163.5       145       182         26.2  cuModuleGetLoadingMode
```

versus:
```
$ RAPIDS_NO_INITIALIZE=1 nsys profile --stats=true --trace=cuda python -c "import cudf;"
Generating '/tmp/nsys-report-0f63.qdstrm'
[1/6] [========================100%] report2.nsys-rep
[2/6] [========================100%] report2.sqlite
[3/6] Executing 'cudaapisum' stats report

 Time (%)  Total Time (ns)  Num Calls  Avg (ns)  Med (ns)  Min (ns)  Max (ns)  StdDev (ns)           Name         
 --------  ---------------  ---------  --------  --------  --------  --------  -----------  ----------------------
     98.3            75237        384     195.9     157.0       107      1771        116.5  cuGetProcAddress      
      1.5             1138          1    1138.0    1138.0      1138      1138          0.0  cuInit                
      0.2              181          1     181.0     181.0       181       181          0.0  cuModuleGetLoadingMode
```

shwina (approved):
I'm +1 to this approach

davidwendt:
I like this approach too. 
I wonder if you could use `cuda-gdb` instead of `gdb` only because it may already be installed with the CUDA toolkit and therefore would not be required to be added to the yamls.

vyasr (approved):
This is much better! 

Minor nit: would it be simpler to write out the gdb command into a bash script (could even write to a tempfile within the Python script itself if we really wanted to be fancy, but I'm mostly just thinking of saving a bash script in the repo to be executed by the test)? It could accept a single argument, the Python command to execute, and then return an appropriate exit code. The subprocess lines aren't terribly readable.

wence-:
> I like this approach too. I wonder if you could use `cuda-gdb` instead of `gdb` only because it may already be installed with the CUDA toolkit and therefore would not be required to be added to the yamls.

Thanks, done!

wence-:
> This is much better!
> 
> Minor nit: would it be simpler to write out the gdb command into a bash script (could even write to a tempfile within the Python script itself if we really wanted to be fancy, but I'm mostly just thinking of saving a bash script in the repo to be executed by the test)? It could accept a single argument, the Python command to execute, and then return an appropriate exit code. The subprocess lines aren't terribly readable.

`gdb -x` accepts commands from stdin, so I've just encoded them as input to the subprocess call instead of creating a temporary file.

wence-:
Skipping test on arm64 for now, since cuda-gdb is non-functional.

ajschmidt8:
Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

wence-:
> Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

Originally I modified the dependencies.yaml but it turned out not to be necessary.

PR #12546, opened 2023-01-13
Remove `cudf::strings::repeat_strings_output_sizes` from Java and JNI

ttnghia:
As described in https://github.com/rapidsai/cudf/issues/12542, the function is no longer needed. Thus, this PR removes all its related callers.

Depends on (so that `strings::repeat_strings` will have the ability to detect output overflow):
 * https://github.com/rapidsai/cudf/pull/12543
 * https://github.com/NVIDIA/spark-rapids/pull/7513


revans2 (approved):
Ideally #12543 would go in first, then we would remove the usage of the API rom the spark plugin, and then finally we would remove the API here.

ttnghia:
Taking this into draft waiting for dependencies.

PR #12548, opened 2023-01-14
Add `regex_program` java APIs and unit tests

cindyyuanjiang:
## Description

Adds a set of java regex APIs that take in a `regex_program` as parameter and java unit tests. This is part of the solution for https://github.com/NVIDIA/spark-rapids/issues/7295.

## Checklist


java/src/main/java/ai/rapids/cudf/CaptureGroups.java
```diff
@@ -0,0 +1,39 @@
...
+ *  limitations under the License.
+ *
+ */
+
+package ai.rapids.cudf;
+
+/**
+ * This is the capture groups setting.
+ * For processing a regex pattern containing capture groups. 
+ * These can be used to optimize the generated regex instructions where the capture groups do not require extracting the groups.
```
mythrocks, line 24:
```suggestion
 * This is the capture groups setting for processing a regex pattern containing
 * capture groups. 
 * These can be used to optimize the generated regex instructions where the 
 * capture groups do not require extracting the groups.
```

I think the code formatter will fix this.

java/src/main/java/ai/rapids/cudf/CaptureGroups.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public enum CaptureGroups {
+  EXTRACT(0), /* capture groups processed normally for extract */
```
mythrocks, line 27: It was a good idea here to include the comments for each enum value.
I would change the comment style here:
```suggestion
  EXTRACT(0), // capture groups processed normally for extract
```

java/src/main/java/ai/rapids/cudf/CaptureGroups.java
mythrocks: I would also mention that this enum tracks closely with `cudf::strings::capture_groups` in `libcudf`.

mythrocks:
@cindyyuanjiang: I've added `Non-breaking` and `Feature` labels to this PR, so that CI might run.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -2539,6 +2539,31 @@ public final Table stringSplit(String pattern, int limit, boolean splitByRegex)
     return new Table(stringSplit(this.getNativeView(), pattern, limit, splitByRegex));
   }
 
+  /**
+   * Returns a list of columns by splitting each string using the specified regex program. The
```
mythrocks, line 2543: We should bump the date in the copyright header.

java/src/main/java/ai/rapids/cudf/CaptureGroups.java
```diff
@@ -0,0 +1,39 @@
...
+  NON_CAPTURE(1); /* convert all capture groups to non-capture groups */
+
+  private int captureGroups;
```
mythrocks, line 30: If we make this `final`, we don't need the explicit getter.
```java
final int nativeId;
```

java/src/main/java/ai/rapids/cudf/RegexProgram.java
```diff
@@ -0,0 +1,94 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ai.rapids.cudf;
+
+/**
+ * This is the regex program class.
+ */
+public class RegexProgram {
+  private String pattern; /* regex pattern */
```
mythrocks, line 22: In general, please consider switching to `//` for comments like these. (Doesn't apply to Javadocs, of course.)
Commenting code is easier without the `/* */` comments getting in the way.

java/src/main/java/ai/rapids/cudf/CaptureGroups.java
cindyyuanjiang: Updated. Thank you!

mythrocks (changes requested):
A couple of minor nitpicks to start.
I have yet to go over the tests.

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
```diff
@@ -4154,6 +4197,34 @@ void testMatchesRe() {
...
+    RegexProgram regexProg2 = new RegexProgram("[A-Za-z]+\\s@[A-Za-z]+", CaptureGroups.NON_CAPTURE);
+    RegexProgram regexProg3 = new RegexProgram(".*", CaptureGroups.NON_CAPTURE);
+    RegexProgram regexProg4 = new RegexProgram("", CaptureGroups.NON_CAPTURE);
+    try (ColumnVector testStrings = ColumnVector.fromStrings("", null, "abCD", "ovér the",
+          "lazy @dog", "1234", "00:0:00");
+         ColumnVector res1 = testStrings.matchesReRegexProg(regexProg1);
+         ColumnVector res2 = testStrings.matchesReRegexProg(regexProg2);
+         ColumnVector res3 = testStrings.matchesReRegexProg(regexProg3);
+         ColumnVector expected1 = ColumnVector.fromBoxedBooleans(false, null, false, false, false,
+           true, true);
```
mythrocks, line 4212: Should the last value be `false`? How come `"00:0:00"` matches `\\d+`?

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
```diff
@@ -4095,6 +4095,18 @@ void testExtractRe() {
       }
   }
 
+  @Test
+  void testExtractReRegexProg() {
+      try (ColumnVector input = ColumnVector.fromStrings("a1", "b2", "c3", null);
+            Table expected = new Table.TestBuilder()
```
mythrocks, line 4101: Is the indentation off by one here?

mythrocks (changes requested):
A couple more minor nitpicks.

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
mythrocks: Ah, `match`, not `fullmatch`.

cindyyuanjiang: Resolved through discussion.

java/src/main/java/ai/rapids/cudf/CaptureGroups.java
```diff
@@ -0,0 +1,35 @@
...
+ * For processing a regex pattern containing capture groups. These can be used
+ * to optimize the generated regex instructions where the capture groups do not
+ * require extracting the groups.
+ */
+public enum CaptureGroups {
+  EXTRACT(0),     // capture groups processed normally for extract
+  NON_CAPTURE(1); // convert all capture groups to non-capture groups
+
+  final int nativeId; // Native id, for use with libcudf.
+  CaptureGroups(int nativeId) {
```
mythrocks, line 32: Sorry for the late notice. The constructor can be private. Only the constant values are meant to be used.

java/src/main/java/ai/rapids/cudf/RegexFlags.java
```diff
@@ -0,0 +1,36 @@
...
+ * leave room for future flags and to match the Python flag values.
+ */
+public enum RegexFlags {
+  DEFAULT(0),   // default
+  MULTILINE(8), // the '^' and '$' honor new-line characters
+  DOTALL(16),   // the '.' matching includes new-line characters
+  ASCII(256);   // use only ASCII when matching built-in character classes
+
+  final int nativeId; // Native id, for use with libcudf.
+  RegexFlags(int nativeId) {
```
mythrocks, line 33: Private constructor.

java/src/main/java/ai/rapids/cudf/RegexFlags.java
```diff
@@ -0,0 +1,41 @@
...
+
+package ai.rapids.cudf;
+
+/**
+ * This is the regex flags setting.
+ * These types can be or'd to combine them.
+ * The values are chosen to leave room for future flags and to match the Python flag values.
+ */
+public enum RegexFlags {
+  DEFAULT(0), /* default */
```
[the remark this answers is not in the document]
mythrocks: D'oh. I meant `//`. :] Thanks for changing this.

mythrocks (approved):
Generally, LGTM. A couple of very minor nitpicks remain.

java/src/main/java/ai/rapids/cudf/CaptureGroups.java
cindyyuanjiang: Updated. Thank you!

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -2539,6 +2539,31 @@ public final Table stringSplit(String pattern, int limit, boolean splitByRegex)
...
+   * @param regexProg the regex program with UTF-8 encoded string identifying the split pattern
+   *                  for each input string.
+   * @param limit the maximum size of the list resulting from splitting each input string,
+   *              or -1 for all possible splits. Note that limit = 0 (all possible splits without
+   *              trailing empty strings) and limit = 1 (no split at all) are not supported.
+   * @param splitByRegex a boolean flag indicating whether the input strings will be split by a
+   *                     regular expression pattern or just by a string literal delimiter.
+   * @return list of strings columns as a table.
+   */
+  public final Table stringSplitRegexProg(RegexProgram regexProg, int limit, boolean splitByRegex) {
```
revans2, line 2557: There are some problems with this API that I wasn't happy with in the original API, but are even worse here.

Can we split this into two separate APIs instead of having a boolean flag to distinguish between them?

```
public final Table stringSplit(String delimiter, int limit) 
public final Table stringSplit(String delimiter)
```

Which will just do what they do today, splits strings on a fixed delimiter (no Regexp) using the given limit.  I am using limit because it matches the existing API, but it really would have been nice to use the same value that CUDF uses `maxSplit`, but that is separate.

The we add in the APIs

```
public final Table stringSplit(RegexProgram delimiter, int limit) 
public final Table stringSplit(RegexProgram delimiter)
```

These would split the input data on the regular expression.  Because there are new APIs it would be nice to use maxSplit, but for consistency I think we should keep the limit.

We then should deprecate all of the existing APIs that use the boolean `splitByRegexp`, and implement them in terms of the non-deprecated APIs. i.e.

```
  @Deprecated
  public final Table stringSplit(String pattern, boolean splitByRegex) {
    if (splitByRegex) {
      RegexpProgram program = new RegexpProgram(pattern, RegexpFlags.DEFAULT, CaptureGroups.NON_CAPTURE);
      return splitString(program);
    } else {
      return stringSplit(pattern);
    }
  }
```

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -2617,6 +2713,20 @@ public final ColumnVector stringSplitRecord(String pattern, boolean splitByRegex
...
+   * Returns a column that are lists of strings in which each list is made by splitting the
+   * corresponding input string using the specified regex program pattern.
+   *
+   * @param regexProg the regex program with UTF-8 encoded string identifying the split pattern
+   *                  for each input string.
+   * @param splitByRegex a boolean flag indicating whether the input strings will be split by a
+   *                     regular expression pattern or just by a string literal delimiter.
+   * @return a LIST column of string elements.
+   */
+  public final ColumnVector stringSplitRecordRegexProg(RegexProgram regexProg, boolean splitByRegex) {
```
revans2, line 2726: The same comment for `stringSplit` above goes for `stringSplitRecord`

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -2910,6 +3047,18 @@ public final ColumnVector replaceRegex(String pattern, Scalar repl) {
...
 
+  /**
+   * For each string, replaces any character sequence matching the given regex program pattern
+   * using the replacement string scalar.
+   *
+   * @param regexProg The regex program with pattern to search within each string.
+   * @param repl The string scalar to replace for each pattern match.
+   * @return A new column vector containing the string results.
+   */
+  public final ColumnVector replaceReRegexProg(RegexProgram regexProg, Scalar repl) {
```
revans2, line 3058: Instead of adding in a `RegexpProg` to the end of the methods, I think it would be better to remove the `Re` at the end and rely on the `RegexProgram` type to allow for method overloading.

```
public final ColumnVector replace(RegexProgram regexProg, Scalar repl) {
```

We then should deprecate the old APIs and implement them in terms of the new API. You need to be careful that we get the capture groups correct for the default because we want the default performance to be good. I don't think we have to worry about capture groups except for replace with backrefs and the extract APIs.

This goes for all of the APIs being added.  Internally the native APIs cannot have overloading so it is fine to have the longer name.

java/src/main/java/ai/rapids/cudf/RegexFlags.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package ai.rapids.cudf;
+
+/**
+ * Regex flags setting, closely following cudf::strings::regex_flags.
+ * These types can be or'd to combine them. The values are chosen to
+ * leave room for future flags and to match the Python flag values.
+ */
+public enum RegexFlags {
```
revans2, line 26: cudf::strings::regex_flags does not map well to an enum in java. The flags are actually bit sets that can be combined to produce a single result, i.e.

`regex_flags.MULTILINE | regexp_flags.ASCII` would make for a regular expression that only operated on ASCII but did multiline.

For us I think we want to rename this to be singular `RegexFlag` and then instead of passing in a `RegexFlag` tot he constructor of `RegexProgram` we pass in an `EnumSet<RegexFlag>`. This would provide the same semantics of the C++ code `EnumSet.of(RegexFlag.MULTILINE, RegexFlag.ASCII)`.  When we pass the results down to C++ we would want to combine them with a but OR and just pass down a single int.

```
int allFlags = 0;
flags.forEach(v => allFlags | v.nativeId);
```

java/src/main/java/ai/rapids/cudf/RegexProgram.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Regex program class, closely following cudf::strings::regex_program.
+ */
+public class RegexProgram {
+  private String pattern;        // regex pattern
+  private RegexFlags flags;      // regex flags for interpreting special characters in the pattern
+  private CaptureGroups capture; // controls how capture groups in the pattern are used
+
+  /**
+   * Constructor for RegexProgram
```
revans2, line 27: nit: Can we explicitly call out that the default is to extract a capture group? The reason I say this is because we want to match CUDF's defaults, but for many APIs this is not the ideal choice, and I want us to be careful with doing this.  As an alternative we could just provide a way for us to reset the capture group after it is created.

java/src/main/java/ai/rapids/cudf/RegexProgram.java
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Get the capture groups used to create this instance
+   *
+   * @param return Capture groups setting
+   */
+  public CaptureGroups capture() {
+    return capture;
+  }
+}
```
revans2, line 94: nit: can we add in native APIs to allow us to call the following APIs.

https://github.com/rapidsai/cudf/blob/22087b3c2a78b49140fdf96091b5abd325427b11/cpp/include/cudf/strings/regex/regex_program.hpp#L93-L113

I am fine if it is a follow on issue that we do later.

java/src/main/java/ai/rapids/cudf/RegexProgram.java
cindyyuanjiang: Thank you very much for your review and comments! I have updated all the changes.

java/src/main/java/ai/rapids/cudf/RegexProgram.java
cindyyuanjiang: I can definitely add them in a follow up issue. I find the current PR a bit large to track things easily. Thank you for the comment!

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -2531,12 +2531,37 @@ public final ColumnVector stringLocate(Scalar substring, int start, int end) {
...
+   *              trailing empty strings) and limit = 1 (no split at all) are not supported.
+   * @return list of strings columns as a table.
+   */
+  public final Table stringSplit(RegexProgram regexProg, int limit) {
     assert type.equals(DType.STRING) : "column type must be a String";
-    assert pattern != null : "pattern is null";
-    assert pattern.length() > 0 : "empty pattern is not supported";
+    assert regexProg != null : "regex program is null";
+    assert regexProg.pattern() != null : "pattern is null";
+    assert regexProg.pattern().length() > 0 : "empty pattern is not supported";
```
revans2, line 2561: nit: Do we ever want to allow a program to have an empty pattern or null pattern. Should these just be part of RegexProgram instead?

revans2 (approved):
Looks great

java/src/main/java/ai/rapids/cudf/ColumnView.java
cindyyuanjiang: Thank you! Do you think it's a good idea to conform with the C++ regex program implementation expectation?

revans2: If C++ is already doing checks for us, then we don't need to replicate them here. If not and we know that they are going to cause problems, then we should add the checks here.

cindyyuanjiang: I added pattern null check in RegexProgram constructor. I saw `cudf::strings::regex_program::create(empty_pattern)` in C++ tests so I think empty pattern should be allowed. Please let me know if I am wrong. Thanks!

revans2: Sorry I just found the reason for this.

https://github.com/rapidsai/cudf/blob/ed6daad3a7edca6b8873b1105a585e12757dcf1c/java/src/main/native/src/ColumnViewJni.cpp#L703-L707

It looks like because the behavior is different from java we are not allowing it.  The point of the CUDF API is not to match java behavior but to expose the CUDF API.  Can we remove the check here and in the native code too. We will keep it in the plugin because that is where we care about it being the same.

(And do it for stringSplitRecord too)

This is still a nit so I am happy to let this in as is, but we could just file a follow on issue to fix it if you want.

revans2 (approved):

java/src/main/java/ai/rapids/cudf/ColumnView.java
cindyyuanjiang: I see. Thank you very much for clarifying this! I can remove them now. To confirm, we want to remove the check from the ColumnView.java stringSplit APIs as well?

revans2 (approved):

revans2:
The test failures appear to be totally unrelated. It failed in a boolean segmented sort test. Not sure if you just need to upmerge or what.

revans2:
rerun tests

revans2:
hmm the same tests failed again. I wonder if others are seeing these same failures or not.

revans2:
@davidwendt it looks like `SegmentedSortInt.Bool` is failing in this PR only on v100 and this PR didn't touch any code that should be impacting this path. Is the patch just out of date or is there some other issue going on here?

ajschmidt8:
> rerun tests

@revans2, the `rerun tests` command does not work with the new GitHub Actions CI setup.

You can check out the docs below for information on how to rerun tests in GH Actions

- https://docs.rapids.ai/resources/github-actions/#rerunning-failed-github-actions

davidwendt:
> @davidwendt it looks like `SegmentedSortInt.Bool` is failing in this PR only on v100 and this PR didn't touch any code that should be impacting this path. Is the patch just out of date or is there some other issue going on here? Calling you out because I think you added this test after there was a bug in CUB with segmented sort.

Yes, this was a cmake error that is fixed by https://github.com/rapidsai/rapids-cmake/pull/353
You will need to do a `Rerun All Jobs` as referenced by AJ in his comment.

revans2:
@cindyyuanjiang your branch is still out of date. Please upmerge again.  Also because this is deprecating some APIs that the spark plugin uses we are going to need to put up a PR in the plugin that adds the warnings to an allow list so we don't break the build.

You will need to add an annotation similar to
```
@scala.annotation.nowarn("msg=method XXX in class YYY is deprecated")
```

You can put this on each off the methods that calls one of the deprecated methods.  Sorry I know this is a pain.

ajschmidt8:
> @cindyyuanjiang your branch is still out of date. Please upmerge again

@revans2, a fully up-to-date PR branch is not required for a PR to merge.

If a PR has all of the required reviews and all of the required GitHub checks are passing, then it can be merged by someone with `write` permissions or greater posting `/merge` as a comment on the PR.

The _Update Branch_ button is a GitHub feature that we enabled ([src](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-suggestions-to-update-pull-request-branches#managing-suggestions-to-update-a-pull-request-branch)) to compliment the new _Recently Updated_ check shown in the screenshot below.

You can read about the _Recently Updated_ check here: https://docs.rapids.ai/resources/recently-updated/

The _Update Branch_ button and the _Recently Updated_ check are separate but complimentary features.

image


At this time, the _Recently Updated_ check is not required, but that may change in the future.

revans2:
@ajschmidt8 you are right. I just got confused when I saw 

Screenshot from 2023-01-26 09-23-05

But it is not a blocker....

cindyyuanjiang:
> @cindyyuanjiang your branch is still out of date. Please upmerge again. Also because this is deprecating some APIs that the spark plugin uses we are going to need to put up a PR in the plugin that adds the warnings to an allow list so we don't break the build.

@revans2 Thank you very much! I will follow up with that.

revans2:
@cindyyuanjiang it's OK I have a patch for the warnings almost ready to go. I realized I hit merge a bit too soon on this so I figured I would take the hit and make the patch

PR #12554, opened 2023-01-16
`partition_by_hash()`: support index

madsbk:
AFAICT, `DataFrame.partition_by_hash()` doesn't handle index correctly. 

When `keep_index=True`, we now include the index in the call to `libcudf.hash.hash_partition()`

## Checklist


wence- (approved):
Looks like correct behaviour, thanks.

madsbk:
CI fails:

wence-:
> What is going on?

cudf tests have `xfail_strict = true` which means that any test marked with xfail that _passes_ turns into a failure in the test suite. The question is, why did code that fixed that bug without removing the xfail get into trunk?

Issue #11977, opened 2022-10-24
[FEA] CategoricalDtype as dtype in CSV reader


mattf:
**Is your feature request related to a problem? Please describe.**
rewriting pandas code with cudf, using `import cudf as pd`

**Describe the solution you'd like**
`cudf.read_csv(dtype=...)` to support category types.

```
$ python
Python 3.9.13 (main, May 18 2022, 00:00:00) 
[GCC 11.3.1 20220421 (Red Hat 11.3.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas, cudf, io

>>> pandas.__version__
'1.5.1'

>>> cudf.__version__
'22.10.00a+392.g1558403753'

>>> d = lambda: io.StringIO("a,b\n" + "1,one\n2,two\n"*2)

>>> pandas.read_csv(d(), dtype={'b': 'category'}).b
0     one
1     two
2     one
3     two
Name: b, dtype: category
Categories (2, object): ['one', 'two']

>>> cudf.read_csv(d(), dtype={'b': 'category'}).b
Traceback (most recent call last):
  File "csv.pyx", line 529, in cudf._lib.csv._get_cudf_data_type_from_dtype
NotImplementedError: CategoricalDtype as dtype is not yet supported in CSV reader
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../python3.9/site-packages/nvtx/nvtx.py", line 101, in inner
    result = func(*args, **kwargs)
  File ".../python3.9/site-packages/cudf/io/csv.py", line 87, in read_csv
    df = libcudf.csv.read_csv(
  File "csv.pyx", line 420, in cudf._lib.csv.read_csv
  File "csv.pyx", line 261, in cudf._lib.csv.make_csv_reader_options
SystemError: <class 'str'> returned a result with an error set
```

Issue #3960, opened 2020-01-28
[BUG] Category dtype gives unexpected hashed values for int32 when reading from CSV


taureandyernv:
**Describe the bug**
I unexpectedly get hashed values  of int32 instead of my original values and a categorical column when typing it a categorical column from CSV

**Steps/Code to reproduce bug**
```
import cudf
import pandas as pd

fn = 'test.csv'
lines = """id1,id2
1,45
2,3
3, 7
1, 25
"""
with open(fn, 'w') as fp:
    fp.write(lines)
pdf = pd.read_csv(fn, header=0, dtype={"id1":"category", "id2":"int32"})
cdf = cudf.read_csv(fn, header=0, dtype={"id1":"category", "id2":"int32"})
```
Output of `print(cdf)`:
```
          id1  id2
0  -677418915   45
1  -197494169    3
2  1957796822    7
3  -677418915   25
```
```
cdf['id1']

0    -677418915
1    -197494169
2    1957796822
3    -677418915
Name: id1, dtype: int32
```

**Expected behavior**
I expect the output to be that of `print(pdf)`
```
          id1  id2
0  -677418915   45
1  -197494169    3
2  1957796822    7
3  -677418915   25
```
```
pdf['id1']

0    1
1    2
2    3
3    1
Name: id1, dtype: category
Categories (3, object): [1, 2, 3]
```
Current work around is to first cast it as int32, then astype("category")

```
cdf2 = cudf.read_csv(fn, header=0, dtype={"id1":"int32", "id2":"int32"})
cdf2['id1'] = cdf2['id1'].astype("category")
cdf2['id1']
```
Outputs:
```
0    1
1    2
2    3
3    1
Name: id1, dtype: category
Categories (3, int64): [1, 2, 3]
```
**Environment overview (please complete the following information)**
 - Environment location: [Docker]
 - Method of cuDF install: [Docker]
- tested to affect cudf 0.11, 0.12

randerzander:
@taureandyernv as a workaround, you could read the CSV file with a string column and do `astype('category')` _after_ reading:

```
cdf = cudf.read_csv(fn, header=0)
cdf.id1.astype('category')
```
```
0    1
1    2
2    3
3    1
Name: id1, dtype: category
Categories (3, int64): [1, 2, 3]
```

I believe dictionary column types are currently being  worked on. @mjsamoht will cuIO's readers support using the `category` dtype at read time?

mjsamoht:
[...]

PR #12571, opened 2023-01-18
[REVIEW] Add support for `category` dtypes in CSV reader

galipremsagar:
## Description

Fixes: https://github.com/rapidsai/cudf/issues/11977, https://github.com/rapidsai/cudf/issues/3960

This PR enables support for `category` dtypes in `dtype` parameter. This PR contains a workaround that enables reading columns as categorical dtypes, we can remove this workaround once `libcudf` has native support for dictionary type mapping to categorical columns.

## Checklist


python/cudf/cudf/tests/test_csv.py
```diff
@@ -2046,19 +2033,27 @@ def test_csv_writer_category(df):
...
-        NotImplementedError,
-        match=re.escape(
-            "CategoricalDtype as dtype is not yet " "supported in CSV reader"
-        ),
-    ):
-        cudf.read_csv(StringIO(csv_buf), dtype="category")
+    actual = cudf.read_csv(StringIO(csv_buf), dtype=dtype)
+    expected = pd.read_csv(StringIO(csv_buf), dtype=dtype)
+
+    assert_eq(expected, actual)
```
vuule, line 2056: happy to see that direct comparison works here.

python/cudf/cudf/tests/test_csv.py
```diff
@@ -2046,19 +2033,27 @@ def test_csv_writer_category(df):
...
+        {"a": "category", "b": "str"},
+        {"b": "category"},
+        {"a": "category"},
+        {"a": pd.CategoricalDtype([1, 2])},
+        {"b": pd.CategoricalDtype([1, 2, 3])},
+        {"b": pd.CategoricalDtype(["b", "a"]), "a": "str"},
+        pd.CategoricalDtype(["a", "b"]),
+    ],
+)
+def test_csv_reader_category(dtype):
```
vuule, line 2049: do we need tests with null elements (if that's even supported)?

vuule (commented):
FWIW this seems fine to me

python/cudf/cudf/tests/test_csv.py
galipremsagar: Added nulls 👍 It's supported.

vuule (approved):

python/cudf/cudf/_lib/csv.pyx
```diff
@@ -429,6 +431,25 @@ def read_csv(
         column_names=meta_names
     ))
 
+    if dtype is not None:
+        if isinstance(dtype, abc.Mapping):
```
bdice, line 435: Is it possible for this to be a generic `Mapping`, or is it guaranteed to be a `dict`? If you only care about `dict`s and not `dict`-like objects, then avoid the abstract class check in favor of `isinstance(dtype, dict)`. It can be considerably faster.

python/cudf/cudf/_lib/csv.pyx
```diff
@@ -429,6 +431,25 @@ def read_csv(
...
+                    df._data[str(k)] = df._data[str(k)].astype(v)
+        elif (
+            cudf.api.types.is_scalar(dtype) or
+            isinstance(dtype, (
+                np.dtype, pd.core.dtypes.dtypes.ExtensionDtype, type
+            ))
+        ):
+            if cudf.api.types.is_categorical_dtype(dtype):
+                df = df.astype(dtype)
+        elif isinstance(dtype, abc.Collection):
```
bdice, line 447: Can dtypes be this broad? Is there a tighter constraint like "`list` or `tuple`" that excludes the cases handled above? Just trying to avoid the abstract type check as above.

bdice (approved):
Changes look good - just wondering if our type checks can be a little more tightly constrained.

python/cudf/cudf/_lib/csv.pyx
bdice: I suppose this might come from unsanitized user input with random dict-like classes as the `dtype`...? Maybe this is the best we can do (here and in the other thread below).

galipremsagar: Our check's aren't much tighter checking for `dict` because in pandas `dtype` is super overloaded param, where for one example they support `defaultdict` too: https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html

PR #12573, opened 2023-01-18
Add compound aggregations to cudf::segmented_reduce

davidwendt:
## Description
Adds mean, variance, and standard deviation aggregation support to `cudf::segmented_reduce`. These are compound (multi-step) aggregations and are modeled after the same aggregations supported but `cudf::reduce`. Once this approved and merged, the visitor pattern for this approach will be reworked for both `cudf::reduce` and `cudf::segmented_reduce` as per [#10432](https://github.com/rapidsai/cudf/issues/10432#issuecomment-1067560695).

The source tree for `src/reductions` has been adjusted to put all segmented-reduce source files into `src/reductions/segmented` and removing the `segmented_` prefix from those file names.
Also, the segmented-reduce functions have been moved from `cudf/detail/reduction_functions.hpp` into their own `cudf/detail/segmented_reduction_functions.hpp`. Likewise, the segmented-reduce CUB calls have been moved from `cudf/detail/reduction.cuh` to the new `cudf/detail/segmented_reduction.cuh` to help minimize including CUB headers.

Additionally, the sum-of-squares aggregation is also included since it was a simple reduction only requiring the appropriate aggregation class registration and source file.

Finally, gtests are added for these new types. The compound types only support floating-point outputs.

Follow on PRs will address the visitor pattern already mentioned above as well as additional data types. Discussion on additional aggregations will occur in the reference issue #10432.

## Checklist


davidwendt:
This will target 23.04 once it is ready.

cpp/include/cudf/detail/reduction.cuh
```diff
@@ -229,92 +228,6 @@ std::unique_ptr<scalar> reduce(InputIterator d_in,
   return std::unique_ptr<scalar>(result);
 }
 
-/**
```
davidwendt, line 232: These functions were moved to `segmented_reduction.cuh`

cpp/include/cudf/detail/reduction_functions.hpp
```diff
@@ -338,171 +338,5 @@ std::unique_ptr<scalar> merge_sets(
   rmm::cuda_stream_view stream,
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
-/**
```
davidwendt, line 341: These functions were moved to `segmented_reduction_functions.hpp`

cpp/include/cudf/detail/segmented_reduction_functions.hpp
```diff
@@ -0,0 +1,320 @@
...
+ * @param col Input column to compute product
+ * @param offsets Indices to identify segment boundaries
+ * @param output_dtype data type of return type and typecast elements of input column
+ * @param null_handling If `null_policy::INCLUDE`, all elements in a segment must be valid for the
+ * reduced value to be valid. If `null_policy::EXCLUDE`, the reduced value is valid if any element
+ * in the segment is valid.
+ * @param init Initial value of each product
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned scalar's device memory
+ * @return Product as scalar of type `output_dtype`
```
bdice, line 76:
```suggestion
 * @return Products of segments as type `output_dtype`
```

cpp/include/cudf/detail/segmented_reduction_functions.hpp
```diff
@@ -0,0 +1,320 @@
...
+ * @param col Input column to compute sum
+ * @param offsets Indices to identify segment boundaries
+ * @param output_dtype Data type of return type and typecast elements of input column
+ * @param null_handling If `null_policy::INCLUDE`, all elements in a segment must be valid for the
+ * reduced value to be valid. If `null_policy::EXCLUDE`, the reduced value is valid if any element
+ * in the segment is valid.
+ * @param init Initial value of each sum
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ * @return Sums of segments in type `output_dtype`
```
bdice, line 48:
```suggestion
 * @return Sums of segments as type `output_dtype`
```

cpp/include/cudf/detail/segmented_reduction_functions.hpp
```diff
@@ -0,0 +1,320 @@
...
+ * @param col Input column to compute minimum
+ * @param offsets Indices to identify segment boundaries
+ * @param output_dtype Data type of return type and typecast elements of input column
+ * @param null_handling If `null_policy::INCLUDE`, all elements in a segment must be valid for the
+ * reduced value to be valid. If `null_policy::EXCLUDE`, the reduced value is valid if any element
+ * in the segment is valid.
+ * @param init Initial value of each minimum
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned scalar's device memory
+ * @return Minimums of segments in type `output_dtype`
```
bdice, line 103:
```suggestion
 * @return Minimums of segments as type `output_dtype`
```

cpp/include/cudf/detail/segmented_reduction_functions.hpp
```diff
@@ -0,0 +1,320 @@
...
+ * @param col Input column to compute maximum
+ * @param offsets Indices to identify segment boundaries
+ * @param output_dtype Data type of return type and typecast elements of input column
+ * @param null_handling If `null_policy::INCLUDE`, all elements in a segment must be valid for the
+ * reduced value to be valid. If `null_policy::EXCLUDE`, the reduced value is valid if any element
+ * in the segment is valid.
+ * @param init Initial value of each maximum
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned scalar's device memory
+ * @return Maximums of segments in type `output_dtype`
```
bdice, line 130:
```suggestion
 * @return Maximums of segments as type `output_dtype`
```

cpp/include/cudf/detail/segmented_reduction_functions.hpp
```diff
@@ -0,0 +1,320 @@
...
+  std::optional<std::reference_wrapper<scalar const>> init,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Compute minimum of each segment in input column.
+ *
+ * If an input segment is empty, the segment result is null.
+ *
+ * @throw cudf::logic_error if input column type is convertible to `output_dtype`.
```
bdice, line 92:
```suggestion
 * @throw cudf::logic_error if input column type is not convertible to `output_dtype`.
```

cpp/include/cudf/detail/segmented_reduction_functions.hpp
```diff
@@ -0,0 +1,320 @@
...
+  std::optional<std::reference_wrapper<scalar const>> init,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Compute maximum of each segment in input column.
+ *
+ * If an input segment is empty, the segment result is null.
+ *
+ * @throw cudf::logic_error if input column type is convertible to `output_dtype`.
```
bdice, line 119:
```suggestion
 * @throw cudf::logic_error if input column type is not convertible to `output_dtype`.
```

cpp/include/cudf/detail/segmented_reduction_functions.hpp
```diff
@@ -0,0 +1,320 @@
...
+
+/**
+ * @brief Computes mean of elements of segments in the input column
+ *
+ * If input segment is empty, the segment result is null.
+ *
+ * If `null_handling==null_policy::INCLUDE`, all elements in a segment must be valid
+ * for the reduced value to be valid.
+ * If `null_handling==null_policy::EXCLUDE`, the reduced value is valid if any element
+ * in the segment is valid.
```
bdice, line 205: For the other reductions above, this is written next to `@param null_handling`. Where do you think it makes most sense? I would delete it here and duplicate the docs from other reductions in the parameter docstring.

cpp/include/cudf/detail/segmented_reduction_functions.hpp
```diff
@@ -0,0 +1,320 @@
...
+ *
+ * @throw cudf::logic_error if input column type is not arithmetic type
+ * @throw cudf::logic_error if `output_dtype` is not floating point type
+ *
+ * @param col Input column data
```
bdice, line 210: The wording of parameter descriptions for the new reductions differs from the wording of the other segmented reductions (e.g. `@param col Input column data` vs. `@param col Input column to compute all`). Maybe consider standardizing pieces of this?

cpp/include/cudf/detail/segmented_reduction_functions.hpp
```diff
@@ -0,0 +1,320 @@
...
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned scalar's device memory
+ * @return Column of `output_dtype` for the reduction results of the segments
+ */
+std::unique_ptr<column> segmented_standard_deviation(
+  column_view const& col,
+  device_span<size_type const> offsets,
+  data_type const output_dtype,
+  null_policy null_handling,
+  cudf::size_type ddof,
```
bdice, line 283: Should we use the namespace? (Inconsistent with other parameters.)
```suggestion
  size_type ddof,
```

cpp/src/reductions/segmented/compound.cuh
```diff
@@ -0,0 +1,193 @@
...
+};
+
+template <typename ElementType, typename Op>
+struct segmented_result_dispatcher {
+ private:
+  template <typename ResultType>
+  static constexpr bool is_supported_v()
+  {
+    // the operator `mean`, `var`, `std` only accepts
+    // floating points as output dtype
```
bdice, line 119: Can we write this explicitly, rather than relying on an assumption that all of our compound operators must return floating types? `is_supported_v` should also depend on `Op`.

cpp/src/reductions/segmented/mean.cu
```diff
@@ -0,0 +1,46 @@
...
+                                             rmm::mr::device_memory_resource* mr)
+{
+  using reducer = compound::detail::compound_segmented_dispatcher<cudf::reduction::op::mean>;
+  return cudf::type_dispatcher(col.type(),
+                               reducer{},
+                               col,
+                               offsets,
+                               output_dtype,
+                               null_handling,
+                               1,  // ddof is not used for mean
```
bdice, line 40: Do you think it would be a worthwhile refactor to make the `ddof` parameter into a `std::optional`?

cpp/tests/reductions/segmented_reduction_tests.cpp
```diff
@@ -770,6 +772,197 @@ TEST_F(SegmentedReductionTestUntyped, NonNullableInput)
   CUDF_TEST_EXPECT_COLUMNS_EQUAL(*res, null_init_expect);
 }
 
+TEST_F(SegmentedReductionTestUntyped, Mean)
```
bdice, line 775: Can we add tests for a couple more cases:
- Errors raised for invalid types (non-arithmetic inputs should raise, non-floating `data_type` should raise)
- A column with no nulls and `null_policy::EXCLUDE` to ensure that the `thrust::adjacent_difference` code path for computing valid counts is executed

cpp/include/cudf/detail/segmented_reduction_functions.hpp
davidwendt: I feel the description here is more about the details of the function than an attribute of the parameter and therefore belongs with the function's detailed description. Also, I feel adding extra functional description in the param section can unnecessarily inflate this section while also unnecessarily restricting the author from any creative elaboration.

cpp/src/reductions/segmented/compound.cuh
davidwendt: Seems adding explicit checks on the Op would be unnecessarily restrictive. Are you having an issue with the name of the functor? What if we changed the name to `compound_float_output_dispatcher` instead? This way it could be used by any future Op types without updating the `is_supported_v()` function.

cpp/include/cudf/detail/segmented_reduction.cuh
```diff
@@ -0,0 +1,200 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "reduction_operators.cuh"
```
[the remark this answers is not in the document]
bdice: I saw that and wondered if the last part was applicable. This is another header (.cuh), not a source (.cu) file.

cpp/src/reductions/segmented/mean.cu
davidwendt: It is a detail function which eventually must be resolved as an actual number when calling the compound-reduce kernel. That is, I would not want to add an if-statement to every device thread. So somewhere this has to happen. I have an issue with the comment itself and can reword it -- the hardcoded value `1` is correct for `mean` but a different hardcoded value may be required for some future compound reduction.

robertmaynard (approved):
CMake changes LGTM

cpp/include/cudf/detail/segmented_reduction.cuh
```diff
@@ -0,0 +1,200 @@
+/*
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
hyperbolic2346, line 2: With this being a new file, should it be 2023 only or is the copyright starting further back since this code was moved? If it is dated on code creation, should it go back to match the original source?

hyperbolic2346 (changes requested):

cpp/include/cudf/detail/segmented_reduction.cuh
davidwendt: I don't think it matters too much as long as it has the current year.

hyperbolic2346 (approved):

cpp/src/reductions/segmented/compound.cuh
```diff
@@ -0,0 +1,191 @@
...
+  auto [output_null_mask, output_null_count] =
+    cudf::detail::segmented_null_mask_reduction(bitmask,
+                                                first_bit_indices_begin,
+                                                first_bit_indices_end,
+                                                last_bit_indices_begin,
+                                                null_handling,
+                                                std::nullopt,
+                                                stream,
+                                                mr);
+  result->set_null_mask(std::move(output_null_mask), output_null_count);
```
davidwendt, line 108: Replace this with new `update_validity()`

bdice (commented):
Reviewed with @davidwendt. A few comments attached. Great work!

bdice (changes requested):
Holding off for a final review pass.

bdice (approved):
Great! Thanks @davidwendt.

Issue #12316, opened 2022-12-06
[BUG] read_parquet performance regression on V100


vuule:
Performance regression was caused by https://github.com/rapidsai/cudf/pull/11867
The difference in performance comes from kernel `gpuDecodePageData`.

PR #12577, opened 2023-01-19
Parquet reader optimization to address V100 regression.

nvdbaranec:
Addresses https://github.com/rapidsai/cudf/issues/12316

Some recent changes caused a performance regression in the parquet reader benchmarks for lists.  The culprit ended up being slightly different code generation happening for arch 70.  In several memory hotspots, the code was reading values from global, modifying them and then storing them.   Previously it had done a better job of loading and keeping them in registers and the L2 cache was helping keep things fast.  But the extra store was causing twice as many L2 access in these places and causing many long scoreboard stalls.

Ultimately the issue is that these values shouldn't be kept in global memory.  The initial implementation did it this way because the data was variable in size (based on depth of column nesting).  But in practice, we never see more than 2 or 3 levels of nesting.  So the solution is:

- Keep these values (in a struct called `PageNestingDecodeInfo`) that is kept in shared memory for up to N nesting levels.  N is currently 10.
- If the nesting information for the incoming column fits in the cache, use it.  Otherwise fall back to the arrays in global memory.  In practice, it is exceedingly rare to see columns nested >= 10 deep.

This addresses the performance regression and actually gives some performance increases.   Some comparisons for LIST benchmarks.
```
cudf 22.10 (prior to regression)
| data_type | cardinality | run_length | bytes_per_second | 
|-----------|-------------|------------|------------------|
|      LIST |           0 |          1 |     892901208    | 
|      LIST |        1000 |          1 |     952863876    |  
|      LIST |           0 |         32 |    1246033395    |  
|      LIST |        1000 |         32 |    1232884866    |  
```

```
cudf 22.12 (where the regression occurred)
| data_type | cardinality | run_length | bytes_per_second | 
|-----------|-------------|------------|------------------|
|      LIST |           0 |          1 |     747758436    | 
|      LIST |        1000 |          1 |     827763260    |  
|      LIST |           0 |         32 |    1026048576    |  
|      LIST |        1000 |         32 |    1022928119    |  
```

```
This PR
| data_type | cardinality | run_length | bytes_per_second | 
|-----------|-------------|------------|------------------|
|      LIST |           0 |          1 |     927347737    | 
|      LIST |        1000 |          1 |     1024566150   |  
|      LIST |           0 |         32 |    1315972881    |  
|      LIST |        1000 |         32 |    1303995168    |  
```


vuule (commented):
Really cool optimization. Just have a few (potential) suggestions.

cpp/src/io/parquet/page_data.cu
```diff
@@ -51,6 +51,12 @@ constexpr int non_zero_buffer_size = block_size * 2;
 
 constexpr int rolling_index(int index) { return index & (non_zero_buffer_size - 1); }
 
+// the PageNestingDecodeInfo struct is 48 bytes (as of 1/18/23) so
+// this cache costs 480 bytes per block. Columns nested more than 2 or 3 deep are exceedingly
+// rare.  If we encounter one that exceeds the below limit, the code falls back on the uncached
+// PageNestingDecodeInfo stored in global memory.
+constexpr int max_cacheable_nesting_decode_info = 10;
```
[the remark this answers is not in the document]
nvdbaranec: Somewhat arbitrary.  I feel like 480 bytes is a reasonable amount to reserve out of shared (1%) and 10 levels of nesting is far beyond typical.  I could go higher but since this kernel uses a small number of warps (4 max per block), by leaving more shared memory available it helps allow other stuff to be scheduled at the same time.

cpp/src/io/parquet/page_data.cu
```diff
@@ -908,21 +921,40 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
   int chunk_idx;
 
   // Fetch page info
-  if (t == 0) s->page = *p;
+  if (!t) s->page = *p;
```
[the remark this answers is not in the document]
nvdbaranec: Yes, and consistency with the rest of the code. `!t` is the idiom in this file.

nvdbaranec:
Tested against spark integration tests.

cpp/src/io/parquet/page_data.cu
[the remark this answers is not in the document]
vuule: maybe then it should be `512/sizeof(PageNestingDecodeInfo)` (assuming 512 is the actual limit you don't want to go over)

nvdbaranec:
rerun tests

vuule (approved):

cpp/src/io/parquet/page_data.cu
```diff
@@ -927,21 +939,40 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
...
   // Fetch column chunk info
   chunk_idx = s->page.chunk_idx;
-  if (t == 0) { s->col = chunks[chunk_idx]; }
+  if (!t) { s->col = chunks[chunk_idx]; }
 
-  // zero nested value and valid counts
+  // if we can use the decode cache, set it up now
+  auto const can_use_decode_cache = s->page.nesting_info_size <= max_cacheable_nesting_decode_info;
+  if (can_use_decode_cache) {
+    int d = 0;
```
bdice, line 953: Can we give `t` and `d` more meaningful names? I get why `t` was chosen but `tidx` or similar would be better. I have no idea what `d` means. `decode_index`?

cpp/src/io/parquet/page_data.cu
```diff
@@ -90,6 +95,13 @@ struct page_state_s {
   const uint8_t* lvl_start[NUM_LEVEL_TYPES];  // [def,rep]
   int32_t lvl_count[NUM_LEVEL_TYPES];         // how many of each of the streams we've decoded
   int32_t row_index_lower_bound;              // lower bound of row indices we should process
+
+  // a shared-memory cache of frequently used data when decoding. The source of this data is
+  // normally stored in global memory which can yield poor performance. So, when possible
+  // we copy that info here prior to decoding
+  PageNestingDecodeInfo nesting_decode_cache[max_cacheable_nesting_decode_info];
+  // points to either nesting_decode above when possible, or to the global source otherwise
+  PageNestingDecodeInfo* pndi;
```
bdice, line 104: Could this be named something like `decode_info` or `page_nesting_cache`? Just real words that indicate the purpose, since `pndi` is just a type abbreviation that loses its referent when used with `auto` or in untyped attribute accessors.

As a bonus: The comment `points to either...` is helpful but I would like it even more if we could encode that idea into the name while avoiding Hungarian notation? (Maybe not possible)

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -83,6 +83,35 @@ enum level_type {
...
+/**
+ * @brief Nesting information specifically needed by the decode and preprocessing
+ * kernels.
+ *
+ * This data is kept separate from PageNestingInfo to keep it as small as possible.
+ * It is used in a cached form in shared memory when possible.
+ */
+struct PageNestingDecodeInfo {
+  // set up prior to decoding
+  int32_t max_def_level;
```
bdice, line 95: Verifying our intended spellings: Should all these types by `int32_t` , or should some be `int` or `cudf::size_type`?

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -83,6 +83,35 @@ enum level_type {
...
+
+  // computed during decoding
+  int32_t null_count;
+
+  // used internally during decoding
+  int32_t valid_map_offset;
+  int32_t valid_count;
+  int32_t value_count;
+  uint8_t* data_out;
+  uint32_t* valid_map;
```
bdice, line 112: Should this be `cudf::bitmask_type`?

bdice (commented):
The core of the changes look good. I have minor questions about types and variable name suggestions. I can approve after that.

cpp/src/io/parquet/page_data.cu
```diff
@@ -927,21 +939,40 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
...
+    int d = 0;
+    while (d < s->page.nesting_info_size) {
+      if (d + t < s->page.nesting_info_size) {
+        // these values need to be copied over from global
+        s->nesting_decode_cache[d + t].max_def_level = s->page.nesting_decode[d + t].max_def_level;
+        s->nesting_decode_cache[d + t].page_start_value =
+          s->page.nesting_decode[d + t].page_start_value;
+        s->nesting_decode_cache[d + t].start_depth = s->page.nesting_decode[d + t].start_depth;
+        s->nesting_decode_cache[d + t].end_depth   = s->page.nesting_decode[d + t].end_depth;
+      }
```
hyperbolic2346, line 962: Neither `d` or `t` are const. Is it safe to assume the compiler will cache this result or should we add `auto const idx = d + t;` to help it?

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -94,30 +123,15 @@ enum level_type {
...
 struct PageNestingInfo {
-  // input repetition/definition levels are remapped with these values
-  // into the corresponding real output nesting depths.
-  int32_t start_depth;
-  int32_t end_depth;
-
-  // set at initialization
-  int32_t max_def_level;
-  int32_t max_rep_level;
+  // set at initialization (see start_offset_output_iterator in reader_impl_preprocess.cu)
```
hyperbolic2346, line 126: Is splitting this worth the memory savings? I get that we want PageNestingDecodeInfo to be as small as possible, but is it worth it now that it's been done? I will leave this simply as a question. I don't have a good idea of how much it gained us.

hyperbolic2346 (commented):
Couple of questions, but very nice find and a solid fix.

cpp/src/io/parquet/page_data.cu
nvdbaranec: How does this strike you:

```
    int depth = 0;
    while (depth < s->page.nesting_info_size) {
      int const thread_depth = depth + t;
      if (thread_depth < s->page.nesting_info_size) {
        // these values need to be copied over from global
        s->nesting_decode_cache[thread_depth].max_def_level = s->page.nesting_decode[thread_depth].max_def_level;
        s->nesting_decode_cache[thread_depth].page_start_value =
          s->page.nesting_decode[thread_depth].page_start_value;
        s->nesting_decode_cache[thread_depth].start_depth = s->page.nesting_decode[thread_depth].start_depth;
        s->nesting_decode_cache[thread_depth].end_depth   = s->page.nesting_decode[thread_depth].end_depth;
      }
      depth += blockDim.x;
    }
```

bdice: Wow. I hadn’t considered “d = depth.” It makes sense in retrospect. I appreciate the verbosity! This is helpful.

cpp/src/io/parquet/page_data.cu
nvdbaranec: The thing I don't like about `decode_info` is that this is more specific than that.  It's decode info related to the nesting of the columns (the `page_state_s` struct itself is really the `decode_info`).  I used `pndi` to make usage of it less crazy to read.  Maybe `nesting_info`?

cpp/src/io/parquet/page_data.cu
nvdbaranec: > Neither `d` or `t` are const. Is it safe to assume the compiler will cache this result or should we add `auto const idx = d + t;` to help it?

Sure.  See  https://github.com/rapidsai/cudf/pull/12577#discussion_r1085540818

cpp/src/io/parquet/page_data.cu
PointKernel: `nesting_info` +1

cpp/src/io/parquet/parquet_gpu.hpp
nvdbaranec: How far do we want to take this here?  Broadly, cuio tends to work with native types in the kernels and not cudf typedefs (mostly for historical reasons, not really philosophically).  I think my preference here would be to keep `int32_t` for now (although there's some errant `int` fields worth fixing up) and then maybe file an issue to do a more detailed audit cleanup.

cpp/src/io/parquet/page_data.cu
```diff
@@ -51,6 +51,11 @@ constexpr int non_zero_buffer_size = block_size * 2;
 
 constexpr int rolling_index(int index) { return index & (non_zero_buffer_size - 1); }
 
+// Use up to 512 bytes of shared memory as a cache for nesting information.
+// As of 1/20/23, this gives us a max nesting depth of 10 (after which it falls back to
+// global memory). This handles all but the most extreme cases.
+constexpr int max_cacheable_nesting_decode_info = (512) / sizeof(PageNestingDecodeInfo);
```
bdice, line 57: Do we have any tests in C++ or Python that cover the fallback (nesting depth > 10)?

cpp/src/io/parquet/page_data.cu
```diff
@@ -1808,6 +1852,8 @@ __global__ void __launch_bounds__(block_size) gpuDecodePageData(
       ((s->col.data_type & 7) == BOOLEAN || (s->col.data_type & 7) == BYTE_ARRAY) ? 64 : 32;
   }
 
+  PageNestingDecodeInfo* pni_base = s->nesting_info;
```
bdice, line 1855: Maybe rename this to align with the other renames (which I really like):
```suggestion
  PageNestingDecodeInfo* nesting_info_base = s->nesting_info;
```

cpp/src/io/parquet/page_data.cu
```diff
@@ -1931,6 +1976,17 @@ __global__ void __launch_bounds__(block_size) gpuDecodePageData(
     }
     __syncthreads();
   }
+
+  // if we are using the nesting decode cache, copy null count back
+  if (s->nesting_info == s->nesting_decode_cache) {
+    int d = 0;
```
bdice, line 1982: Looks like this block missed the renames performed in other places.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -601,6 +607,9 @@ void reader::impl::allocate_nesting_info()
           gpu::PageNestingInfo* pni =
             &page_nesting_info[nesting_info_index + (p_idx * per_page_nesting_info_size)];
 
+          gpu::PageNestingDecodeInfo* pndi =
```
bdice, line 610: Same here. Maybe needs renamed to `nesting_info`?

bdice (approved):
A few minor comments on naming. Otherwise looks good.

cpp/src/io/parquet/page_data.cu
nvdbaranec: Done.

PR #12579, opened 2023-01-19
[REVIEW] Remove a parquet writer xfail 

galipremsagar:
## Description
This PR fixes a pytest failure where the test is now passing on pandas-1.5.3 which was just released.

## Checklist

### DIFF
--- a/python/cudf/cudf/tests/test_parquet.py
+++ b/python/cudf/cudf/tests/test_parquet.py
@@ -2389,42 +2389,37 @@ def test_parquet_writer_list_statistics(tmpdir):
             ]
         },
         # Struct of Lists
-        pytest.param(
-            {
-                "Real estate records": [
-                    None,
-                    {
-                        "Status": "NRI",
-                        "Ownerships": {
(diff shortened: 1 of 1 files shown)


bdice (approved):

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -2389,42 +2389,37 @@ def test_parquet_writer_list_statistics(tmpdir):
...
+                {"Status": "RI", "Ownerships": None},
+                {
+                    "Status": None,
+                    "Ownerships": {
+                        "land_unit": [7, 8, 9],
+                        "flats": [[], [], []],
                     },
-                ]
-            },
-            marks=pytest.mark.xfail(
```
wence-, line 2424: I think we concluded in previous discussions that we should at least condition these on the last major version. So maybe add `condition=PANDAS_LT_153` rather than fully removing the xfail

galipremsagar: Done 👍

wence- (approved):

PR #11718, opened 2022-09-20
[REVIEW] Copy on write implementation

galipremsagar:
This PR introduces `copy-on-write`. As the name suggests when `copy-on-write` is enabled, when there is a shallow copy of a column made, both the columns share the same memory and only when there is a write operation being performed on either the parent or any of it's copies a true copy will be triggered. Copy-on-write(`c-o-w`) can be enabled in two ways:

1. Setting `CUDF_COPY_ON_WRITE` environment variable to `1` will enable `c-o-w`, unsetting will disable `c-o-w`.
2. Setting `copy_on_write` option in `cudf` options by doing `cudf.set_option("copy_on_write", True)` to enable it and `cudf.set_option("copy_on_write", False)` to disable it.

Note: Copy-on-write is not being enabled by default, it is being introduced as an opt-in.

A valid performance comparison can be done only with `copy_on_write=OFF` + `.copy(deep=True)` vs `copy_on_write=ON` + `.copy(deep=False)`:

```python
In [1]: import cudf

In [2]: s = cudf.Series(range(0, 100000000))

# branch-23.02 : 1209MiB
# This-PR : 1209MiB

In [3]: s_copy = s.copy(deep=True) #branch-23.02
In [3]: s_copy = s.copy(deep=False) #This-PR

# branch-23.02 : 1973MiB
# This-PR : 1209MiB

In [4]: s
Out[4]: 
0                  0
1                  1
2                  2
3                  3
4                  4
              ...   
99999995    99999995
99999996    99999996
99999997    99999997
99999998    99999998
99999999    99999999
Length: 100000000, dtype: int64

In [5]: s_copy
Out[5]: 
0                  0
1                  1
2                  2
3                  3
4                  4
              ...   
99999995    99999995
99999996    99999996
99999997    99999997
99999998    99999998
99999999    99999999
Length: 100000000, dtype: int64

In [6]: s[2] = 10001

# branch-23.02 : 3121MiB
# This-PR : 3121MiB

In [7]: s
Out[7]: 
0                  0
1                  1
2              10001
3                  3
4                  4
              ...   
99999995    99999995
99999996    99999996
99999997    99999997
99999998    99999998
99999999    99999999
Length: 100000000, dtype: int64

In [8]: s_copy
Out[8]: 
0                  0
1                  1
2                  2
3                  3
4                  4
              ...   
99999995    99999995
99999996    99999996
99999997    99999997
99999998    99999998
99999999    99999999
Length: 100000000, dtype: int64
```

[...]

PR #12556, opened 2023-01-17
[REVIEW] Update `copy-on-write` with `branch-23.02` changes

galipremsagar:
## Description
This PR updates the `copy-on-write` branch with `get_ptr` & `data_array_view` APIs.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12587, opened 2023-01-20
[REVIEW] Change ways to access `ptr` in `Buffer` 

galipremsagar:
## Description
This PR:
With the introduction of `copy-on-write`(https://github.com/rapidsai/cudf/pull/11718), and `spilling`, there are different ways each `Buffer` implementation expects the code to behave when a `.ptr` is accessed, i.e., in case of COW, a `ptr` can be accessed for read or write purposes, in case of spilling the ptr can be accessed with a spill lock and the buffer will be spillable after the spill lock goes out of scope during execution. For these reasons we introduced `ptr`, `mutable_ptr` and `data_array_view`, `_data_array_view`, `mask_array_view`, and `_mask_array_view`. With so many ways to access buffer ptr and array views, this has become quite difficult for devs to know when to use what unless you are fully familiar with the implementation details. It will also lead us to a lot of special case handling for `Buffer`, `SpillableBuffer`, and `CopyonWriteBuffer`. For this reason, we have decided to simplify fetching the pointer with a `get_ptr(mode="read"/"write")` API, fetching data & mask array views will also become methods that accept `mode` like `data_array_view(mode="read"/"write")` & `mask_array_view(mode="read"/"write")`. It is the expectation that when the caller passed "read", they don't tamper with the buffer or the memory it is pointing to. In the case of "write", they are good to mutate the memory it is pointing to.

Note that even with `mode="read"/"write"` the caller should still, if appropriate, acquire a spill lock for the duration of the access. If this is not done, and the buffer is a `SpillableBuffer`, it will permanently be marked as unspillable.

- [x] Introduces `get_ptr()` to replace `ptr` property. 
- [x] Replaces `data_array_view` & `mask_array_view` methods with `data_array_view(mode=r/w)` & `mask_array_view(mode=r/w)`

## Checklist


python/cudf/cudf/_lib/copying.pyx
```diff
@@ -765,7 +765,7 @@ cdef class _CPackedColumns:
         gpu_data = Buffer.deserialize(header["data"], frames)
 
         dbuf = DeviceBuffer(
-            ptr=gpu_data.ptr,
+            ptr=gpu_data.get_ptr(mode="write"),
```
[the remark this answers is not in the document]
wence-: Ah wait, no, in the `SpillableBuffer` case we don't want to mark these data as unspillable, no? Although I guess it doesn't matter because the `gpu_data` object doesn't persist past the end of this function (unless it's tracked somewhere globally?), so what we're doing here is transferring ownership of the pointer?

python/cudf/cudf/core/column/categorical.py
```diff
@@ -956,9 +956,8 @@ def clip(self, lo: ScalarLike, hi: ScalarLike) -> "column.ColumnBase":
             self.astype(self.categories.dtype).clip(lo, hi).astype(self.dtype)
         )
 
-    @property
-    def data_array_view(self) -> cuda.devicearray.DeviceNDArray:
-        return self.codes.data_array_view
+    def data_array_view(self, mode="read") -> cuda.devicearray.DeviceNDArray:
```
wence-, line 959: Can we default to safe but slow `"write"`? Or does `"write"` not differ from `read` for this case?

python/cudf/cudf/core/column/numerical.py
```diff
@@ -573,14 +573,14 @@ def _find_value(
...
-                self.data_array_view,
+                self.data_array_view(mode="read"),
                 value,
                 mask=self.mask,
             )
         if found == -1:
             if self.is_monotonic_increasing and closest:
                 found = find(
-                    self.data_array_view,
+                    self.data_array_view(mode="read"),
```
wence-, line 583: Do we need to `acquire_spill_lock` around this?

python/cudf/cudf/core/df_protocol.py
```diff
@@ -89,7 +89,7 @@ def ptr(self) -> int:
         """
         Pointer to start of the buffer as an integer.
         """
-        return self._buf.ptr
+        return self._buf.get_ptr(mode="write")
```
wence-, line 92: This is because the dataframe interchange protocol is a zero-copy one, so an external consumer is getting a view?

python/cudf/cudf/core/frame.py
```diff
@@ -1448,7 +1448,7 @@ def searchsorted(
 
         # Return result as cupy array if the values is non-scalar
         # If values is scalar, result is expected to be scalar.
-        result = cupy.asarray(outcol.data_array_view)
+        result = cupy.asarray(outcol.data_array_view(mode="read"))
```
wence-, line 1451: This should acquire a spill lock, I think, but it doesn't matter because `outcol` is new and doesn't escape this function.

python/cudf/cudf/utils/queryutils.py
```diff
@@ -220,7 +220,7 @@ def query_execute(df, expr, callenv):
             "or bool dtypes."
         )
 
-    colarrays = [col.data_array_view for col in colarrays]
+    colarrays = [col.data_array_view(mode="write") for col in colarrays]
```
wence-, line 223: These columns are only read in `make_aggregate_nullmask`, no? If yes, do we also need to acquire a spill lock?

wence- (changes requested):
Some small queries.

python/cudf/cudf/core/df_protocol.py
galipremsagar: I'm not quite familiar with df interchange protocol if it has to be a zero-copy one. I just made an equivalent switch of `.ptr` to `mode='write` assuming the buffer should be marked as exposed at this point since we are handing off to another library, my understanding here can be wrong too. @shwina to confirm.

python/cudf/cudf/core/frame.py
galipremsagar: That's correct

python/cudf/cudf/_lib/column.pyx
```diff
@@ -341,7 +332,9 @@ cdef class Column:
 
         cdef libcudf_types.bitmask_type* mask
         if self.nullable:
-            mask = <libcudf_types.bitmask_type*><uintptr_t>(self.base_mask_ptr)
+            mask = <libcudf_types.bitmask_type*><uintptr_t>(
+                self.base_mask.get_ptr(mode="read")
+            )
```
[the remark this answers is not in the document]
galipremsagar: Switched to `internal_write`

python/cudf/cudf/core/column/numerical.py
galipremsagar: Yup, added 👍

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,47 @@ def owner(self) -> Any:
     @property
     def __cuda_array_interface__(self) -> Mapping:
         """Implementation of the CUDA Array Interface."""
+        return self._get_cuda_array_interface(readonly=False)
+
+    def _get_cuda_array_interface(self, readonly=False):
         return {
-            "data": (self.ptr, False),
+            "data": (self.get_ptr(mode="read"), readonly),
```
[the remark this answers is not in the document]
galipremsagar: Yup, made the change 👍

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,47 @@ def owner(self) -> Any:
...
+            version=0,
+        )
+
+    def get_ptr(self, mode="write") -> int:
+        """Device pointer to the start of the buffer.
+
+        Parameters
+        ----------
+        mode : str, default 'write'
+            Supported values are {"read", "write"}
```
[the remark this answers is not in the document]
galipremsagar: Done 👍

python/cudf/cudf/utils/queryutils.py
galipremsagar: Yup, this can be a read-only with spill lock. Made the change 👍

galipremsagar:
> Some small queries.

Addressed 👍  This PR should be ready for another round of reviews.

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,66 @@ def owner(self) -> Any:
     @property
     def __cuda_array_interface__(self) -> Mapping:
         """Implementation of the CUDA Array Interface."""
+        return self._get_cuda_array_interface(readonly=False)
+
+    def _get_cuda_array_interface(self, readonly=False):
```
madsbk, line 217: short doc here, would be good

madsbk (changes requested):
Overall looks good but I don't think we need the `"internal_write"` mode.

python/cudf/cudf/core/buffer/spillable_buffer.py
```diff
@@ -287,7 +304,11 @@ def get_ptr(self) -> int:
         from cudf.core.buffer.utils import get_spill_lock
 
         spill_lock = get_spill_lock()
-        if spill_lock is None:
+
+        if mode == "internal_write":
+            if spill_lock is None:
+                spill_lock = SpillLock()
+        if spill_lock is None or mode == "write":
```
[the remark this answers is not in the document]
wence-: Is the idea that if no `spill_lock` exists that is saying "I want to expose this pointer for unlimited duration", whereas if there _is_ a `spill_lock` then I am only exposing things for the lifetime of the spill lock? If so, then I agree that `internal_write` is not needed, and it is up to the caller to acquire a spill lock if they want a scope lifetime exposure.

python/cudf/cudf/core/buffer/spillable_buffer.py
```diff
@@ -287,7 +304,11 @@ def get_ptr(self) -> int:
         from cudf.core.buffer.utils import get_spill_lock
 
         spill_lock = get_spill_lock()
-        if spill_lock is None:
+
+        if mode == "internal_write":
+            if spill_lock is None:
+                spill_lock = SpillLock()
```
wence-, line 310: Irrespective of the discussion below, I am not sure that this can be correct. Since the `spill_lock` created here doesn't escape this function, the lifetime of the lock is not long enough to guarantee that the returned pointer is safe for use on the device.

python/cudf/cudf/core/buffer/spillable_buffer.py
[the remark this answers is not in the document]
madsbk: Yes

python/cudf/cudf/core/buffer/spillable_buffer.py
galipremsagar: Removed this code

python/cudf/cudf/core/buffer/spillable_buffer.py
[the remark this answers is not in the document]
galipremsagar: Got it, this approach sounds good to me 👍

python/cudf/cudf/core/buffer/buffer.py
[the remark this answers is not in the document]
galipremsagar: Added docstring

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,77 @@ def owner(self) -> Any:
...
+        ----------
+        mode : str, default 'write'
+            Supported values are {"read", "write"}
+            If "write", the modification to the buffer
+            is allowed. Hence can pass it onto a
+            third-party library as well.
+            If "read", then modification to the buffer
+            isn't expected to be done by the user via this pointer.
+            Hence mostly intended for internal read purposes or
+            device to host copies.
```
wence-, line 270: How about this rewording?

> If "write", the data pointed to may be modified by the caller. If "read", the data pointed to must not be modified by the caller. Failure to fulfil this contract will cause incorrect behaviour.

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,77 @@ def owner(self) -> Any:
...
+
+        Notes
+        -----
+        In case of `Buffer` class, any value passed to `mode` is a no-op.
+        The significance of each value explained above is a
+        guidance for any implementation of `Buffer` class.
```
wence-, line 276: Link to the subclasses if that is easy?

python/cudf/cudf/core/column/categorical.py
wence-: Still `"read"` here AFAICT?

python/cudf/cudf/core/column/column.py
```diff
@@ -113,19 +113,63 @@ def as_frame(self) -> "cudf.core.frame.Frame":
...
+            with readonly flag set to True in CAI is returned.
+
+        Returns
+        -------
+        numba.cuda.cudadrv.devicearray.DeviceNDArray
         """
-        return cuda.as_cuda_array(self.data).view(self.dtype)
+        if self.data is not None:
+            if mode == "read":
+                obj = self.data._get_readonly_proxy_obj
```
wence-, line 137: If this is a spillable buffer the caller still has to hold a spill lock, right?

python/cudf/cudf/core/column/column.py
```diff
@@ -113,19 +113,63 @@ def as_frame(self) -> "cudf.core.frame.Frame":
...
+    ) -> "cuda.devicearray.DeviceNDArray":
         """
         View the data as a device array object
+
+        Parameters
+        ----------
+        mode : str, default 'write'
+            Supported values are {'read', 'write'}
+            If 'write' is passed, a device array object
+            with readonly flag set to False in CAI is returned.
```
wence-, line 127: Suggestion: make explicit what this means.

> If the caller wishes to modify the data returned through this view, they must pass mode="write".

python/cudf/cudf/core/column/timedelta.py
```diff
@@ -128,8 +128,12 @@ def values(self):
     def to_arrow(self) -> pa.Array:
```
wence-, line 128:
```suggestion
    @acquire_spill_lock
    def to_arrow(self) -> pa.Array:
```

python/cudf/cudf/core/window/rolling.py
```diff
@@ -488,7 +488,7 @@ def _window_to_window_sizes(self, window):
             return window
         else:
             return cudautils.window_sizes_from_offset(
-                self.obj.index._values.data_array_view, window
+                self.obj.index._values.data_array_view(mode="write"), window
```
wence-, line 491: Need to acquire a spill lock here (for the duration of the `window_sizes_from_offset` call)

python/cudf/cudf/core/window/rolling.py
```diff
@@ -527,13 +527,13 @@ def __init__(self, groupby, window, min_periods=None, center=False):
...
             return cudautils.grouped_window_sizes_from_offset(
-                column.arange(len(self.obj)).data_array_view,
+                column.arange(len(self.obj)).data_array_view(mode="read"),
                 self._group_starts,
                 window,
             )
         else:
             return cudautils.grouped_window_sizes_from_offset(
-                self.obj.index._values.data_array_view,
+                self.obj.index._values.data_array_view(mode="read"),
```
wence-, line 536: And a spill lock her.

python/cudf/cudf/utils/applyutils.py
```diff
@@ -143,17 +144,20 @@ def run(self, df, **launch_params):
         # Get input columns
```
wence-, line 144: Need to acquire a spill lock for this function.

python/strings_udf/strings_udf/_typing.py
```diff
@@ -112,7 +114,9 @@ def prepare_args(self, ty, val, **kwargs):
         if isinstance(ty, types.CPointer) and isinstance(
             ty.dtype, (StringView, UDFString)
         ):
-            return types.uint64, val.ptr
+            return types.uint64, val.ptr if isinstance(
+                val, rmm._lib.device_buffer.DeviceBuffer
+            ) else val.get_ptr(mode="read")
```
wence-, line 119: This will expose any buffer passed in permanently unless someone is holding a spill lock, but I don't know who should be doing so? Presumably it is the caller of the strings_udf kernel. Where is that @brandon-b-miller ?

wence- (commented):
Getting there, thanks! Some more minor things I think.

python/cudf/cudf/core/column/column.py
galipremsagar: Right

python/cudf/cudf/core/buffer/buffer.py
galipremsagar: Done 👍

python/cudf/cudf/core/buffer/buffer.py
galipremsagar: Linked 👍

python/cudf/cudf/core/column/categorical.py
galipremsagar: Sorry forgot to do so. Thanks for catching it. Should be fixed now.

python/cudf/cudf/core/column/column.py
galipremsagar: Done 👍

python/cudf/cudf/core/column/column.py
```diff
@@ -113,19 +113,63 @@ def as_frame(self) -> "cudf.core.frame.Frame":
...
         View the mask as a device array
+
+        Parameters
+        ----------
+        mode : str, default 'write'
+            Supported values are {'read', 'write'}
+            If 'write' is passed, a device array object
+            with readonly flag set to False in CAI is returned.
+            If 'read' is passed, a device array object
+            with readonly flag set to True in CAI is returned.
```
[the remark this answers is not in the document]
galipremsagar: Done

python/strings_udf/strings_udf/_typing.py
brandon-b-miller: two places:
https://github.com/rapidsai/cudf/blob/branch-23.02/python/strings_udf/strings_udf/tests/test_string_udfs.py#L71
https://github.com/rapidsai/cudf/blob/branch-23.02/python/cudf/cudf/core/indexed_frame.py#L2131

python/cudf/cudf/utils/applyutils.py
galipremsagar: Done

python/cudf/cudf/core/window/rolling.py
galipremsagar: Done 👍

python/cudf/cudf/core/series.py
```diff
@@ -4959,10 +4959,10 @@ def isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
         index = as_index(a.index)
 
     a_col = column.as_column(a)
-    a_array = cupy.asarray(a_col.data_array_view)
+    a_array = cupy.asarray(a_col.data_array_view(mode="read"))
 
     b_col = column.as_column(b)
-    b_array = cupy.asarray(b_col.data_array_view)
+    b_array = cupy.asarray(b_col.data_array_view(mode="read"))
```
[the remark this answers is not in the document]
galipremsagar: Added spill locks 👍

brandon-b-miller (approved):
Approving `strings_udf` change.

galipremsagar:
> Getting there, thanks! Some more minor things I think.

Thanks for catching the missing locks, @wence- ! Should be ready for review now.

wence- (approved):

wence-:
I think I am happy now. Thanks very much for your perseverance!

python/cudf/cudf/tests/test_spilling.py
```diff
@@ -119,7 +119,7 @@ def test_spillable_buffer(manager: SpillManager):
     buf = as_buffer(data=rmm.DeviceBuffer(size=10), exposed=False)
     assert isinstance(buf, SpillableBuffer)
     assert buf.spillable
-    buf.ptr  # Expose pointer
+    buf.get_ptr(mode="write")  # Expose pointer
```
madsbk, line 122:
```suggestion
    buf.mark_exposed()
```

python/cudf/cudf/tests/test_spilling.py
```diff
@@ -210,7 +209,7 @@ def test_spilling_buffer(manager: SpillManager):
     buf = as_buffer(rmm.DeviceBuffer(size=10), exposed=False)
     buf.spill(target="cpu")
     assert buf.is_spilled
-    buf.ptr  # Expose pointer and trigger unspill
+    buf.get_ptr(mode="write")  # Expose pointer and trigger unspill
```
madsbk, line 212:
```suggestion
    buf.mark_exposed()  # Expose pointer and trigger unspill
```

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,79 @@ def owner(self) -> Any:
...
+        return cuda_array_interface_wrapper(
+            ptr=self.get_ptr(mode="read"),
+            size=self.size,
+            owner=self,
+            readonly=True,
+            typestr="|u1",
+            version=0,
+        )
+
+    def get_ptr(self, mode="write") -> int:
```
madsbk, line 257: I think it is great that we are setting `mode` by keyword everywhere, let's make it mandatory?
```suggestion
    def get_ptr(self, *, mode="write") -> int:
```

python/cudf/cudf/core/buffer/spillable_buffer.py
```diff
@@ -269,7 +269,7 @@ def spill_lock(self, spill_lock: SpillLock) -> None:
             self.spill(target="gpu")
             self._spill_locks.add(spill_lock)
 
-    def get_ptr(self) -> int:
+    def get_ptr(self, mode="write") -> int:
```
madsbk, line 272:
```suggestion
    def get_ptr(self, *, mode="write") -> int:
```

python/cudf/cudf/core/column/column.py
```diff
@@ -363,7 +414,7 @@ def nullmask(self) -> Buffer:
         """The gpu buffer for the null-mask"""
         if not self.nullable:
             raise ValueError("Column has no null mask")
-        return self.mask_array_view
+        return self.mask_array_view(mode="write")
```
madsbk, line 417:
```suggestion
        return self.mask_array_view(mode="read")
```

python/cudf/cudf/core/column/column.py
```diff
@@ -363,7 +407,7 @@ def nullmask(self) -> Buffer:
...
```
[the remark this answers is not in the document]
madsbk: No, I think it can be `mode="read"`.
But notice, the access mode doesn't affect the spillability anymore. It is only whether `get_ptr(...)` is called within `acquire_spill_lock` or not that determines if the buffer is marked unspillable temporarily or permanently.

python/cudf/cudf/tests/test_spilling.py
```diff
@@ -589,7 +588,7 @@ def test_statistics_expose(manager: SpillManager):
     ]
 
     # Expose the first buffer
-    buffers[0].ptr
+    buffers[0].get_ptr(mode="write")
```
madsbk, line 591:
```suggestion
    buffers[0].mark_exposed()
```

python/cudf/cudf/tests/test_spilling.py
```diff
@@ -618,7 +617,7 @@ def test_statistics_expose(manager: SpillManager):
 
     # Expose the new buffers and check that they are counted as spilled
     for i in range(10):
-        buffers[i].ptr
+        buffers[i].get_ptr(mode="write")
```
madsbk, line 620:
```suggestion
        buffers[i].mark_exposed()
```

madsbk (commented):
Some minor suggestions

vyasr (commented):
Mostly looks good. I've got a few questions and suggestions, but not blocking merging. I'll defer to the other reviewers for a final approval since they've been much more active on this PR.

python/cudf/cudf/core/df_protocol.py
shwina: There's no promise that the interchange protocol is zero-copy. In-fact, in the case of cross-device (CPU<->GPU) interchange, it cannot be.

That being said, I think any public API via which we expose the raw pointer to a buffer should mark the buffer exposed.

galipremsagar: Thanks for clarifying @shwina! This will mark the buffer as exposed.

python/cudf/cudf/core/column/column.py
```diff
@@ -113,19 +113,69 @@ def as_frame(self) -> "cudf.core.frame.Frame":
...
+                obj = self.data
+            else:
+                raise ValueError(f"Unsupported mode: {mode}")
+        else:
+            obj = None
+        return cuda.as_cuda_array(obj).view(self.dtype)
 
-    @property
-    def mask_array_view(self) -> "cuda.devicearray.DeviceNDArray":
+    def mask_array_view(self, mode="read") -> "cuda.devicearray.DeviceNDArray":
```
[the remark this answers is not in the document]
galipremsagar: Switched to mandatory keyword-only arg like `get_ptr`

python/cudf/cudf/core/column/column.py
```diff
@@ -113,19 +113,69 @@ def as_frame(self) -> "cudf.core.frame.Frame":
             {None: self.copy(deep=False)}
         )
 
-    @property
-    def data_array_view(self) -> "cuda.devicearray.DeviceNDArray":
+    def data_array_view(
+        self, mode="write"
```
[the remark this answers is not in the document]
galipremsagar: Switched to mandatory keyword-only arg like `get_ptr`

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,79 @@ def owner(self) -> Any:
...
+                readonly,
+            ),
             "shape": (self.size,),
             "strides": None,
             "typestr": "|u1",
             "version": 0,
         }
 
+    @property
+    def _get_readonly_proxy_obj(self):
```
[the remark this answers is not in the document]
galipremsagar: Removed `get_` and renamed to `_readonly_proxy_cai_obj`

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,79 @@ def owner(self) -> Any:
...
+        return cuda_array_interface_wrapper(
+            ptr=self.get_ptr(mode="read"),
+            size=self.size,
+            owner=self,
+            readonly=True,
+            typestr="|u1",
+            version=0,
+        )
+
+    def get_ptr(self, *, mode="write") -> int:
```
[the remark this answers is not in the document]
galipremsagar: Maintaining a default was to be on the safe side when someone doesn't specify anything but yeah by making it keyword-only we have eliminated that problem. I don't have a strong preference here, happy to change it. @wence- WDYT? Do you have any preference here?

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,79 @@ def owner(self) -> Any:
...
+        """Device pointer to the start of the buffer.
+
+        Parameters
+        ----------
+        mode : str, default 'write'
+            Supported values are {"read", "write"}
+            If "write", the data pointed to may be modified
+            by the caller. If "read", the data pointed to
+            must not be modified by the caller.
+            Failure to fulfil this contract will cause
```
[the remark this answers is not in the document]
wence-: "fulfil" is a valid spelling (otherwise codespell would have complained!)

python/cudf/cudf/core/column/column.py
```diff
@@ -146,7 +146,9 @@ def data_array_view(
             obj = None
         return cuda.as_cuda_array(obj).view(self.dtype)
 
-    def mask_array_view(self, mode="read") -> "cuda.devicearray.DeviceNDArray":
+    def mask_array_view(
+        self, *, mode="read"
```
wence-, line 150:
```suggestion
        self, *, mode="write"
```

This one has still snuck through :)

galipremsagar: Yikes, Sorry!

python/cudf/cudf/core/buffer/buffer.py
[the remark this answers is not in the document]
galipremsagar: It is currently a no-op, but then COW is enabled. What do we want the default behavior to do? i.e., when someone does `.get_ptr()` call?

harrism:
@galipremsagar Can you add some detail on motivation and purpose in the PR description?  I came here because of the cuSpatial PR to adapt to this change, but can't find an explanation.

galipremsagar:
> @galipremsagar Can you add some detail on motivation and purpose in the PR description? I came here because of the cuSpatial PR to adapt to this change, but can't find an explanation.

Hi Mark, Thanks for asking for these details. I have updated the PR description now to also contain the motivation behind these changes.

python/cudf/cudf/core/buffer/buffer.py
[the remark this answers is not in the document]
wence-: @vyasr is suggesting this signature:

```python
def get_ptr(self, *, mode) -> int:
    ...
```

This a required keyword-only argument. So it would be invalid to call `buffer.get_ptr()`: the caller must _always_ provide a `mode` argument, but can _only_ do so as a keyword argument.

I think this would be a fine thing to do.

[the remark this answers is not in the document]
galipremsagar: Got it, thanks for clarifying. Dropped the default 👍

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,79 @@ def owner(self) -> Any:
...
+                readonly,
+            ),
             "shape": (self.size,),
             "strides": None,
             "typestr": "|u1",
             "version": 0,
         }
 
+    @property
+    def _readonly_proxy_cai_obj(self):
```
shwina, line 244: Honestly, I don't think we should add extra property here given that it's used only in one other function (do we expect more uses for it?)

If not, I think we can just use `cuda_array_interface_wrapper` in those places instead.

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -215,18 +212,79 @@ def owner(self) -> Any:
...
+            by the caller. If "read", the data pointed to
+            must not be modified by the caller.
+            Failure to fulfill this contract will cause
+            incorrect behavior.
+
+        Notes
+        -----
+        In case of `Buffer` class, any value passed to `mode` is a no-op.
+        The significance of each value explained above is a
+        guidance for any subclass of `Buffer`.
```
shwina, line 274:
```suggestion
```

I think it's better not to make any promises that `mode=` will always be ignored for `Buffer`.

python/cudf/cudf/core/buffer/buffer.py
galipremsagar: It is only being used in `data_array_view` & `mask_array_view`, and one more place for copy-on-write scenarios: https://github.com/rapidsai/cudf/pull/12556.

python/cudf/cudf/core/buffer/buffer.py
galipremsagar: okay by me 👍

python/strings_udf/strings_udf/tests/test_string_udfs.py
```diff
@@ -68,7 +69,8 @@ def run_udf_test(data, func, dtype):
     str_views = column_to_string_view_array(cudf_column)
 
     kernel = get_kernel(func, dtype, len(data))
-    kernel.forall(len(data))(str_views, output)
+    with acquire_spill_lock():
+        kernel.forall(len(data))(str_views, output)
```
shwina, line 73: I think we should avoid setting this precedent; tests really shouldn't have to worry about spilling.

galipremsagar, line 73:
```suggestion
    kernel.forall(len(data))(str_views, output)
```
Dropping adding of spill lock here for what @shwina brought up and consistency with cudf pytests too.

python/strings_udf/strings_udf/tests/test_string_udfs.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
galipremsagar, line 1:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

python/strings_udf/strings_udf/tests/test_string_udfs.py
```diff
@@ -10,6 +10,7 @@
 
 import cudf
 import rmm
+from cudf.core.buffer import acquire_spill_lock
```
galipremsagar, line 13:
```suggestion
```

python/cudf/cudf/core/buffer/buffer.py
shwina: OK -- let's leave as is now, and we can take stock in the merge of `copy-on-write`.

shwina (approved):

python/strings_udf/strings_udf/tests/test_string_udfs.py
galipremsagar: Discussed offline and removed this lock to keep things consistent where we are explicitly not testing for spilling ability.

PR #12589, opened 2023-01-20
Fix incorrect null counts for sliced columns in JCudfSerialization

jlowe:
## Description
This fixes a long-standing problem in JCudfSerialization where sliced columns do not have the correct null counts when serialized and then deserialized.  Previously the null count of the original column was used as a "good enough"  proxy, and this was fine as long as it was only interpreted as a boolean "do I need to worry about nulls" signal.  However it can lead to bad behavior for applications that use the null count directly, such as for statistics in a file footer.

This resolves the issue by explicitly treating sliced column null counts as unknown when it cannot be determined what the true null count is without scanning the sliced validity buffer.  By using the unknown null count sentinel value in the serialized data stream, the deserialized column will compute the null count on demand, producing the correct value.

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/JCudfSerialization.java
+++ b/java/src/main/java/ai/rapids/cudf/JCudfSerialization.java
@@ -1,6 +1,6 @@
 /*
  *
- *  Copyright (c) 2019-2022, NVIDIA CORPORATION.
+ *  Copyright (c) 2019-2023, NVIDIA CORPORATION.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


revans2 (approved):

java/src/main/java/ai/rapids/cudf/JCudfSerialization.java
```diff
@@ -1155,6 +1164,10 @@ private static long copySlicedAndPad(DataWriter out,
   // VALIDITY
   /////////////////////////////////////////////
 
+  private static boolean needsValidityBuffer(long nullCount) {
+    return nullCount > 0 || nullCount == ColumnView.UNKNOWN_NULL_COUNT;
+  }
```
ttnghia, line 1169: I am a bit nervous with this. Can the column be non-nullable if its null count is `UNKNOWN_NULL_COUNT`?

jlowe: No, because we already check if the sliced column came from a column with a null count of zero. In that case we know the null count of the sliced column as well (also zero). So it should only be unknown when we know there’s at least one null in the original column that was sliced, and therefore the validity buffer is present.

PR #12605, opened 2023-01-24
Changing `cudf::io::source_info` to use `host_span` of `std::byte`

### LINKED ISSUE #12576: Replace host_buffer with host_span<std::byte const>
cuIO exposes a `host_buffer` type that was implemented before `host_span` and has the same semantics.
The new device datasource uses `device_span<std::byte const>` for its input and we should do the same for the host datasource.

hyperbolic2346:
## Description

Closes #12576

This change converts `cudf::io::source_info` to take a `host_span` of `std::byte`.

## Checklist

### DIFF
--- a/cpp/benchmarks/io/cuio_common.cpp
+++ b/cpp/benchmarks/io/cuio_common.cpp
@@ -47,7 +47,9 @@ cudf::io::source_info cuio_source_sink_pair::make_source_info()
 {
   switch (type) {
     case io_type::FILEPATH: return cudf::io::source_info(file_name);
-    case io_type::HOST_BUFFER: return cudf::io::source_info(h_buffer.data(), h_buffer.size());
+    case io_type::HOST_BUFFER:
+      return cudf::io::source_info(cudf::host_span<std::byte const>(
+        reinterpret_cast<std::byte const*>(h_buffer.data()), h_buffer.size()));
     case io_type::DEVICE_BUFFER: {
       // TODO: make cuio_source_sink_pair stream-friendly and avoid implicit use of the default
       // stream
--- a/cpp/tests/io/json_test.cpp
+++ b/cpp/tests/io/json_test.cpp
@@ -253,7 +253,9 @@ struct JsonValidFixedPointReaderTest : public JsonFixedPointReaderTest<DecimalTy
                         return acc + (acc.empty() ? "" : "\n") + "{\"col0\":" + rhs + "}";
                       });
     cudf::io::json_reader_options const in_opts =
-      cudf::io::json_reader_options::builder(cudf::io::source_info{buffer.c_str(), buffer.size()})
+      cudf::io::json_reader_options::builder(
+        cudf::io::source_info{cudf::host_span<std::byte const>(
+          reinterpret_cast<std::byte const*>(buffer.data()), buffer.size())})
         .dtypes({data_type{type_to_id<DecimalType>(), scale}})
         .lines(true)
         .legacy(use_legacy_parser);
@@ -300,7 +302,9 @@ TEST_P(JsonReaderParamTest, BasicJsonLines)
   std::string data = is_row_orient_test(test_opt) ? row_orient : record_orient;
 
   cudf::io::json_reader_options in_options =
-    cudf::io::json_reader_options::builder(cudf::io::source_info{data.data(), data.size()})
+    cudf::io::json_reader_options::builder(
+      cudf::io::source_info{cudf::host_span<std::byte const>(
+        reinterpret_cast<std::byte const*>(data.data()), data.size())})
       .dtypes(std::vector<data_type>{dtype<int32_t>(), dtype<double>()})
       .lines(true)
       .legacy(is_legacy_test(test_opt));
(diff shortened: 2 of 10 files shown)


hyperbolic2346:
Just pushed a change to give back the original signature and allow creating a `source_info` from a pointer and size. This will allow us to almost remove breaking. I say almost, because we lost the `host_buffer` type, which was exposed and could have been used to construct a `source_info`.

It is possible to add `using host_buffer = cudf::host_span...` to mitigate this issue, but I haven't done that yet.

cpp/tests/io/json_test.cpp
```diff
@@ -1704,8 +1734,10 @@ TYPED_TEST(JsonFixedPointReaderTest, EmptyValues)
 TEST_F(JsonReaderTest, UnsupportedMultipleFileInputs)
 {
   std::string const data = "{\"col\":0}";
-  auto const buffer      = cudf::io::host_buffer{data.data(), data.size()};
-  auto const src         = cudf::io::source_info{{buffer, buffer}};
+  auto const buffer =
+    cudf::host_span<std::byte const>(reinterpret_cast<std::byte const*>(data.data()), data.size());
+  auto const src =
+    cudf::io::source_info{std::vector<cudf::host_span<std::byte const>>{buffer, buffer}};
```
vuule, line 1740: Very disappointed that we explicitly need to specify the type here. What is the error without the change in this line?

cpp/tests/io/json_test.cpp
```diff
@@ -1704,8 +1734,10 @@ TYPED_TEST(JsonFixedPointReaderTest, EmptyValues)
 TEST_F(JsonReaderTest, UnsupportedMultipleFileInputs)
 {
   std::string const data = "{\"col\":0}";
-  auto const buffer      = cudf::io::host_buffer{data.data(), data.size()};
-  auto const src         = cudf::io::source_info{{buffer, buffer}};
+  auto const buffer =
+    cudf::host_span<std::byte const>(reinterpret_cast<std::byte const*>(data.data()), data.size());
```
vuule, line 1738: could we remove the cast now?

cpp/tests/io/json_test.cpp
```diff
@@ -642,7 +648,9 @@ TEST_P(JsonReaderParamTest, JsonLinesDtypeInference)
   std::string data          = is_row_orient_test(test_opt) ? row_orient : record_orient;
 
   cudf::io::json_reader_options in_options =
-    cudf::io::json_reader_options::builder(cudf::io::source_info{data.data(), data.size()})
+    cudf::io::json_reader_options::builder(
+      cudf::io::source_info{cudf::host_span<std::byte const>(
+        reinterpret_cast<std::byte const*>(data.data()), data.size())})
```
vuule, line 653: would this work?
```suggestion
      cudf::io::source_info{cudf::host_span(data.data(), data.size())})
```

cpp/tests/io/json_test.cpp
```diff
@@ -1717,8 +1717,7 @@ TYPED_TEST(JsonFixedPointReaderTest, EmptyValues)
 
   cudf::io::json_reader_options const in_opts =
     cudf::io::json_reader_options::builder(
-      cudf::io::source_info{cudf::host_span<std::byte const>(
-        reinterpret_cast<std::byte const*>(buffer.data()), buffer.size())})
+      cudf::io::source_info{cudf::host_span<char const>(buffer)})
```
vuule, line 1720: wait, how does this work?

vuule (commented):
Like the API
Still figuring out how clean we can make the call sites.

cpp/benchmarks/io/cuio_common.cpp
```diff
@@ -47,7 +47,9 @@ cudf::io::source_info cuio_source_sink_pair::make_source_info()
 {
   switch (type) {
     case io_type::FILEPATH: return cudf::io::source_info(file_name);
-    case io_type::HOST_BUFFER: return cudf::io::source_info(h_buffer.data(), h_buffer.size());
+    case io_type::HOST_BUFFER:
+      return cudf::io::source_info(cudf::host_span<std::byte const>(
+        reinterpret_cast<std::byte const*>(h_buffer.data()), h_buffer.size()));
```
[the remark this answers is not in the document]
hyperbolic2346: No, it isn't able to with GCC 9.5. I didn't try other versions.

cpp/tests/io/json_test.cpp
hyperbolic2346: No, I tried this and variations multiple times and it was unable to determine what to do.

cpp/tests/io/json_test.cpp
hyperbolic2346: This is the new API I just added that is templated. I'm playing around with how we can do this without as much refactoring on the user's end.
``` 
template <typename T, CUDF_ENABLE_IF(is_byte_like_type<T>())>
explicit source_info(std::vector<cudf::host_span<T const>> const& host_buffers)
```

I also have 
```
template <typename T, CUDF_ENABLE_IF(is_byte_like_type<T>())>
explicit source_info(T const* host_data, size_t size)
```

Which should make all the changes go away. I don't know if we want to hide the `host_span` behind the API though.

cpp/tests/io/json_test.cpp
hyperbolic2346: I agree entirely. With the new API's I just added we may be able to get away with something better here. I will test.

```
error: no matching function for call to 'cudf::io::source_info::source_info(<brace-enclosed initializer list>)'
 1740 |     cudf::io::source_info{{buffer, buffer}};
```

I was able to compile and tests passed with
```
cudf::io::source_info{std::vector{buffer, buffer}}; 
```

PR #12607, opened 2023-01-25
Handle when spillable buffers own each other 

madsbk:
#12587 exposed a bug triggered when `as_buffer()` is given a pointer already owned by another spillable buffer. In this PR, we make this illegal and return a `SpillableBufferSlice` instead. 

cc. @galipremsagar 

#### Checklist

### DIFF
--- a/python/cudf/cudf/core/buffer/spillable_buffer.py
+++ b/python/cudf/cudf/core/buffer/spillable_buffer.py
@@ -27,6 +27,14 @@ if TYPE_CHECKING:
 T = TypeVar("T", bound="SpillableBuffer")
 
 
+def as_spillable_buffer(obj, exposed) -> SpillableBuffer:
+    if hasattr(obj, "__cuda_array_interface__"):
+        return SpillableBuffer._from_device_memory(obj, exposed=exposed)
+    if exposed:
+        raise ValueError("cannot created exposed host memory")
+    return SpillableBuffer._from_host_memory(obj)
+
+
 class SpillLock:
     pass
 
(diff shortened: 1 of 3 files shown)


galipremsagar (approved):
Thanks @madsbk ! Fixes the segfault I observed locally too.

python/cudf/cudf/core/buffer/spillable_buffer.py
```diff
@@ -27,6 +37,84 @@
...
+
+    If `data` isn't a buffer already, a new buffer that points to the memory of
+    `data` is created. If `data` represents host memory, it is copied to a new
+    `rmm.DeviceBuffer` device allocation. Otherwise, the memory of `data` is
+    **not** copied, instead the new buffer keeps a reference to `data` in order
+    to retain its lifetime.
+
+    If `data` is owned by a spillable buffer, a "slice" of the buffer is
+    returned. In this case, the spillable buffer must either be "exposed" or
+    spilled locked (called within an acquire_spill_lock context).
```
shwina, line 75: Can we add a note clarifying why this is necessary? Is it because otherwise, the owner could be spilled during the execution of this function?

madsbk: Added 
```
    If `data` is owned by a spillable buffer, a "slice" of the buffer is
    returned. In this case, the spillable buffer must either be "exposed" or
    spilled locked (called within an acquire_spill_lock context). This is to
    guarantee that the memory of `data` isn't spilled before this function gets
    to calculate the offset of the new slice.
```

shwina (approved):

PR #12609, opened 2023-01-25
Remove cudf::strings::repeat_strings_output_sizes and optional parameter from cudf::strings::repeat_strings

### LINKED ISSUE #12542: Deprecate cudf::strings::repeat_strings_output_sizes in favor of cudf::repeat_strings throwing size limit error

davidwendt:
## Description
Removes `cudf::strings::repeat_strings_output_sizes` and the optional sizes parameter from `cudf::strings::repeat_strings`.
This function (and corresponding optional parameter) is no longer needed now that the internal utilities will throw an error if the column output size exceeds the maximum.
Closes #12542

## Checklist


cpp/src/strings/repeat_strings.cu
```diff
@@ -189,146 +189,63 @@ struct compute_size_and_repeat_separately_fn {
   // If d_chars != nullptr: only repeat strings.
   char* d_chars{nullptr};
 
-  __device__ int64_t operator()(size_type const idx) const noexcept
+  __device__ void operator()(size_type const idx) const noexcept
```
karthikeyann, line 192: is this a `const` function still?

davidwendt: Yes, the member variables are not changed by this function.

karthikeyann: should writes to `d_offsets[idx]` and `output_ptr` be not considered for this?

karthikeyann: https://stackoverflow.com/questions/7390437/c-const-member-functions-are-modifying-member-variables

got it.

davidwendt: That's correct. The function does not change the actual `d_offsets` or `d_chars` variables. Changing data/memory they point to does not count.

karthikeyann (approved):

cpp/include/cudf/strings/repeat_strings.hpp
```diff
@@ -32,15 +32,15 @@ namespace strings {
  */
 
 /**
- * @brief Repeat the given string scalar by a given number of times.
+ * @brief Repeat the given string scalar by a given number of times
```
bdice, line 35: “Repeat by” sounds odd to me.
```suggestion
 * @brief Repeat the given string scalar a given number of times
```

cpp/include/cudf/strings/repeat_strings.hpp
```diff
@@ -50,33 +50,30 @@ namespace strings {
...
+ * @return New string scalar in which the input string is repeated
  */
 std::unique_ptr<string_scalar> repeat_string(
   string_scalar const& input,
   size_type repeat_times,
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
 /**
- * @brief Repeat each string in the given strings column by a given number of times.
+ * @brief Repeat each string in the given strings column by a given number of times
```
bdice, line 67:
```suggestion
 * @brief Repeat each string in the given strings column a given number of times
```

bdice (approved):
Thanks! Good cleanup.

PR #12574, opened 2023-01-18
Use kvikIO as the default IO backend

vuule:
## Description
Closes https://github.com/rapidsai/cudf/issues/10415
Switching the default backend to kvikIO. 
Benchmarks show the same or better throughput.

Removing the redundant libcudf tests that ran with kvikIO.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12616, opened 2023-01-26
Revert "Use kvikIO as the default IO backend"

vuule:
Reverts rapidsai/cudf#12574

Crash or segfault is observed when using kvikIO + CUDA 11.8 in process exit.


vuule:
issue is fixed in kvikIO side, closing this PR as it's not needed

PR #12619, opened 2023-01-26
[REVIEW] Merge `copy-on-write` feature branch into `branch-23.04`

galipremsagar:
## Description
This PR contains changes from https://github.com/rapidsai/cudf/pull/11718 primarily that will enable Copy on write feature in cudf.

This PR introduces `copy-on-write`. As the name suggests when `copy-on-write` is enabled, when there is a shallow copy of a column made, both the columns share the same memory and only when there is a write operation being performed on either the parent or any of it's copies a true copy will be triggered. Copy-on-write(`c-o-w`) can be enabled in two ways:

1. Setting `CUDF_COPY_ON_WRITE` environment variable to `1` will enable `c-o-w`, unsetting will disable `c-o-w`.
2. Setting `copy_on_write` option in `cudf` options by doing `cudf.set_option("copy_on_write", True)` to enable it and `cudf.set_option("copy_on_write", False)` to disable it.

Note: Copy-on-write is not being enabled by default, it is being introduced as an opt-in.

A valid performance comparison can be done only with `copy_on_write=OFF` + `.copy(deep=True)` vs `copy_on_write=ON` + `.copy(deep=False)`:

```python
In [1]: import cudf

In [2]: s = cudf.Series(range(0, 100000000))

# branch-23.02 : 1209MiB
# This-PR : 1209MiB

In [3]: s_copy = s.copy(deep=True) #branch-23.02
In [3]: s_copy = s.copy(deep=False) #This-PR

# branch-23.02 : 1973MiB
# This-PR : 1209MiB

In [4]: s
Out[4]: 
0                  0
1                  1
2                  2
3                  3
4                  4
              ...   
99999995    99999995
99999996    99999996
99999997    99999997
99999998    99999998
99999999    99999999
Length: 100000000, dtype: int64

In [5]: s_copy
Out[5]: 
0                  0
1                  1
2                  2
3                  3
4                  4
              ...   
99999995    99999995
99999996    99999996
99999997    99999997
99999998    99999998
99999999    99999999
Length: 100000000, dtype: int64

In [6]: s[2] = 10001

# branch-23.02 : 3121MiB
# This-PR : 3121MiB

In [7]: s
Out[7]: 
0                  0
1                  1
2              10001
3                  3
4                  4
              ...   
99999995    99999995
99999996    99999996
99999997    99999997
99999998    99999998
99999999    99999999
Length: 100000000, dtype: int64

In [8]: s_copy
Out[8]: 
0                  0
1                  1
2                  2
3                  3
4                  4
              ...   
99999995    99999995
99999996    99999996
99999997    99999997
99999998    99999998
99999999    99999999
Length: 100000000, dtype: int64
```

Stats around the performance and memory gains :

- [x] Memory usage of new copies will be 0 GPU memory additional overhead i.e., users will save 2x, 5x, 10x,...20x memory usage for making 2x, 5x, 10x,...20x deep copies respectively. **So, The more you copy the more you save 😉**(_as long as you don't write on all of them_) 
- [x] **copying times are now cut by 99%** for all dtypes when copy-on-write is enabled(`copy_on_write=OFF` + `.copy(deep=True)` vs `copy_on_write=ON` + `.copy(deep=False)`).

```python
In [1]: import cudf

In [2]: df = cudf.DataFrame({'a': range(0, 1000000)})

In [3]: df = cudf.DataFrame({'a': range(0, 100000000)})

In [4]: df['b'] = df.a.astype('str')

# GPU memory usage
# branch-23.02 : 2345MiB
# This-PR : 2345MiB

In [5]: df
Out[5]: 
                 a         b
0                0         0
1                1         1
2                2         2
3                3         3
4                4         4
...            ...       ...
99999995  99999995  99999995
99999996  99999996  99999996
99999997  99999997  99999997
99999998  99999998  99999998
99999999  99999999  99999999

[100000000 rows x 2 columns]

In [6]: def make_two_copies(df, deep):
   ...:     return df.copy(deep=deep), df.copy(deep=deep)
   ...: 

In [7]: x, y = make_two_copies(df, deep=True) # branch-23.02
In [7]: x, y = make_two_copies(df, deep=False) # This PR

# GPU memory usage
# branch-23.02 : 6147MiB
# This-PR : 2345MiB

In [8]: %timeit make_two_copies(df, deep=True) # branch-23.02
In [8]: %timeit make_two_copies(df, deep=False) # This PR

# Execution times
# branch-23.02 : 135 ms ± 4.49 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
# This-PR : 100 µs ± 879 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

- [x] Even when `copy-on-write` is disabled, `string`, `list` & `struct` columns **deep copies are now 99% faster**

```python

In [1]: import cudf

In [2]: s = cudf.Series(range(0, 100000000), dtype='str')

In [3]: %timeit s.copy(deep=True)


# branch-23.02 : 28.3 ms ± 5.32 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
# This PR : 19.9 µs ± 93.5 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)


In [9]: s = cudf.Series([[1, 2], [2, 3], [3, 4], [4, 5], [6, 7]]* 10000000)

In [10]: %timeit s.copy(deep=True)
# branch-23.02 : 25.7 ms ± 5.99 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
# This-PR : 44.2 µs ± 1.61 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)


In [4]: df = cudf.DataFrame({'a': range(0, 100000000), 'b': range(0, 100000000)[::-1]})

In [5]: s = df.to_struct()

In [6]: %timeit s.copy(deep=True)

# branch-23.02 : 42.5 ms ± 3.5 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
# This-PR : 89.7 µs ± 1.68 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

- [x] Add pytests
- [x] Docs page explaining copy on write and how to enable/disable it.


python/cudf/cudf/options.py
```diff
@@ -215,9 +241,28 @@ def _integer_and_none_validator(val):
...
+    textwrap.dedent(
+        """
+        Default behavior of performing shallow copies.
+        If set to `False`, each shallow copy will perform a true shallow copy.
+        If set to `True`, each shallow copy will perform a shallow copy
+        with underlying data actually referring to the actual column, in this
+        case a copy is only made when there is a write operation performed on
+        the column.
+        \tValid values are True or False. Default is False.
+    """
```
shwina, line 260: Should we simply point to the `copy-on-write` docs here rather than trying to explain what copy on write means in a sentence?

galipremsagar: Yup, that makes sense. Updated to point to docs.

randerzander:
benchmark bot, please test this PR

randerzander:
benchmark bot, please test this PR

randerzander:
benchmark bot, please test this PR

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
 Internally, these objects typically interact with cuDF objects at the Frame layer via composition.
 However, for performance reasons they frequently access internal attributes and methods of `Frame` and its subclasses.
+
+
+## Copy-on-write
+
+
+Copy-on-write (COW) is designed to reduce memory footprint on GPUs. With this feature, a copy (`.copy(deep=False)`) is only really made whenever
+there is a write operation on a column. It is first recommended to see
+the public usage [here](copy-on-write-user-doc) of this functionality before reading through the internals
```
vyasr, line 324: No need to rewrite the justifications since we're referring devs to the user guide anyway.
```suggestion
This section describes the internal implementation details of the copy-on-write feature.
It is recommended that developers familiarize themselves with [the user-facing documentation](copy-on-write-user-doc) of this functionality before reading through the internals
```

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+
+The core copy-on-write implementation relies on the `CopyOnWriteBuffer` class. This class stores the pointer to the device memory and size.
+With the help of `CopyOnWriteBuffer._ptr` we generate [weak references](https://docs.python.org/3/library/weakref.html) of `CopyOnWriteBuffer` and store it in `CopyOnWriteBuffer._instances`.
+This is a mapping from `ptr` keys to `WeakSet`s containing references to `CopyOnWriterBuffer` objects. This
+means all the new `CopyOnWriteBuffer`s that are created map to the same key in `CopyOnWriteBuffer._instances` if they have same `._ptr`
+i.e., if they are all pointing to the same device memory.
+
+When the cudf option `"copy_on_write"` is `True`, `as_buffer` will always return a `CopyOnWriteBuffer`. This class contains all the
+mechanisms to enable copy-on-write for all buffers. When a `CopyOnWriteBuffer` is created, its weakref is generated and added to the `WeakSet` which is in turn stored in `CopyOnWriterBuffer._instances`. This will later serve as an indication of whether or not to make a copy when a
+when write operation is performed on a `Column` (see below).
```
vyasr, line 335: This explanation is really good, just suggesting some reorganization to reduce duplication and help the flow.
```suggestion
The core copy-on-write implementation relies on the `CopyOnWriteBuffer` class.
When the cudf option `"copy_on_write"` is `True`, `as_buffer` will always return a `CopyOnWriteBuffer`.
This subclass of `cudf.Buffer` contains all the mechanisms to enable copy-on-write behavior.
The class stores [weak references](https://docs.python.org/3/library/weakref.html) to every existing `CopyOnWriteBuffer` in `CopyOnWriteBuffer._instances`, a mapping from `ptr` keys to `WeakSet`s containing references to `CopyOnWriterBuffer` objects.
This means that all `CopyOnWriteBuffer`s that point to the same device memory are contained in the same `WeakSet` (corresponding to the same `ptr` key) in `CopyOnWriteBuffer._instances`.
This data structure is then used to determine whether or not to make a copy when a
when write operation is performed on a `Column` (see below).
If multiple buffers point to the same underlying memory, then a copy must be made whenever a modification is attempted.
```

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+
+
+### Eager copies when exposing to third-party libraries
+
+If `Column`/`CopyOnWriteBuffer` is exposed to a third-party library via `__cuda_array_interface__`, we are no longer able to track whether or not modification of the buffer has occurred without introspection. Hence whenever
+someone accesses data through the `__cuda_array_interface__`, we eagerly trigger the copy by calling
+`_unlink_shared_buffers` which ensures a true copy of underlying device data is made and
+unlinks the buffer from any shared "weak" references. Any future shallow-copy requests must also trigger a true physical copy (since we cannot track the lifetime of the third-party object), to handle this we also mark the `Column`/`CopyOnWriteBuffer` as
```
vyasr, line 343:
```suggestion
unlinks the buffer from any shared "weak" references. Any future copy requests must also trigger a true physical copy (since we cannot track the lifetime of the third-party object). To handle this we also mark the `Column`/`CopyOnWriteBuffer` as
```

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+`obj._zero_copied=True` thus indicating any future shallow-copy requests will trigger a true physical copy
```
vyasr, line 344:
```suggestion
`obj._zero_copied=True` thus indicating that any future shallow-copy requests will trigger a true physical copy
```

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+rather than a copy-on-write shallow copy with weak references.
+
+### How to obtain read-only object?
```
vyasr, line 347:
```suggestion
### Obtaining a read-only object
```

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+
+A read-only object can be quite useful for operations that will not
+mutate the data. This can be achieved by calling `._readonly_proxy_cai_obj`
+API, this API will return a proxy object that has `__cuda_array_interface__`
```
vyasr, line 351:
```suggestion
mutate the data. This can be achieved by calling `._readonly_proxy_cai_obj`.
This function will return a proxy object that has `__cuda_array_interface__`
```

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+implemented and will not trigger a deep copy even if the `CopyOnWriteBuffer`
+has weak references. It is only recommended to use this API as long as
+the objects/arrays created with this proxy object gets cleaned up during
+the developer code execution. We currently use this API for device to host
```
vyasr, line 355:
```suggestion
has weak references. This API should only be used when the lifetime of the proxy object is restricted to cudf's internal code execution. Handing this out to external libraries or user-facing APIs will lead to untracked references and undefined copy-on-write behavior. We currently use this API for device to host
```

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+copies like in `ColumnBase.data_array_view(mode="read")` which is used for `Column.values_host`.
+
+Notes:
+1. Weak references are implemented only for fixed-width data types as these are only column
+types that can be mutated in place.
+2. Deep copies of variable width data types return shallow-copies of the Columns, because these
+types don't support real in-place mutations to the data. We just mimic in such a way that it looks
+like an in-place operation using `_mimic_inplace`.
```
vyasr, line 363: Let's just make this a separate section

```suggestion
### Variable width data types
Weak references are implemented only for fixed-width data types as these are only column
types that can be mutated in place.
Deep copies of variable width data types return shallow-copies of the Columns, because these
types don't support real in-place mutations to the data.
Internally, we mimic in-place mutations using `_mimic_inplace`, but the resulting data is always a deep copy of the underlying data.
```

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+```python
+>>> s2.data._ptr
+139796315899392
+>>> s3.data._ptr
+139796315897856
+>>> s1.data._ptr
+139796315879723
+```
+
+cudf Copy-on-write implementation is motivated by pandas Copy-on-write proposal here:
```
vyasr, line 491:
```suggestion
cudf's copy-on-write implementation is motivated by the pandas proposals documented here:
```

vyasr (changes requested):
Can we copy the PR description from #11718 since this is now the PR that will be merged and show up in the changelog? Of course, in case any updates are needed given the various changes we have made since that PR, please make those changes in this description.

This review focused exclusively on the docs (IMO very important to describe COW well in our documentation). Will follow up with a code review.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -538,7 +549,9 @@ cdef class Column:
                     owner=data_owner,
                     exposed=True,
                 )
-                if isinstance(data_owner, SpillableBuffer):
+                if isinstance(data_owner, CopyOnWriteBuffer):
+                    data_owner.ptr  # accessing the pointer marks it exposed.
```
vyasr, line 553: This looks wrong, `.ptr` is gone. Is this code path covered in tests, or am I completely missing something here?

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -194,6 +194,31 @@ def __getitem__(self, key: slice) -> Buffer:
...
+            If True, returns a deep copy of the underlying Buffer data.
+            If False, returns a shallow copy of the Buffer pointing to
+            the same underlying data.
+
+        Returns
+        -------
+        Buffer
+        """
+        if deep:
+            with cudf.core.buffer.acquire_spill_lock():
```
vyasr, line 213: This feels like a leaky abstraction. We shouldn't be acquiring spill locks in buffer, we should implement `SpillableBuffer.copy` as `with acquire_spill_lock(): return super().copy(...)`.

python/cudf/cudf/core/buffer/cow_buffer.py
```diff
@@ -0,0 +1,215 @@
...
+        ----------
+        deep : bool, default True
+            If True, returns a deep-copy of the underlying Buffer data.
+            If False, returns a shallow-copy of the Buffer pointing to
+            the same underlying data.
+
+        Returns
+        -------
+        Buffer
+        """
```
vyasr, line 149: We can inherit the docstring from the parent class.
```suggestion
```

python/cudf/cudf/core/series.py
```diff
@@ -539,7 +545,21 @@ def __init__(
                 data = {}
 
         if not isinstance(data, ColumnBase):
-            data = column.as_column(data, nan_as_null=nan_as_null, dtype=dtype)
+            has_cai = (
+                type(
```
vyasr, line 549: Please add a comment explaining why we need to use `gettatr_static`.

python/cudf/cudf/options.py
```diff
@@ -150,6 +150,33 @@ def _validator(val):
...
+        raise ValueError(
+            f"{val} is not a valid option. Must be one of {{False, True}}."
+        )
+
+
+def _spill_validator(val):
+    try:
+        if get_option("copy_on_write") and val:
+            raise ValueError(
+                "Spilling is not supported when Copy-on-write is enabled. "
```
vyasr, line 169:
```suggestion
                "Spilling is not supported when copy-on-write is enabled. "
```

python/cudf/cudf/options.py
```diff
@@ -215,9 +241,26 @@ def _integer_and_none_validator(val):
...
+
+_register_option(
+    "copy_on_write",
+    _env_get_bool("CUDF_COPY_ON_WRITE", False),
+    textwrap.dedent(
+        """
+        Default behavior of performing shallow copies.
+        If set to `False`, disables copy-on-write.
+        If set to `True`, enables copy-on-write.
+        Read more at: :ref: copy-on-write-user-doc
```
vyasr, line 256: Does this reference actually end up rendered somewhere?

python/cudf/cudf/options.py
```diff
@@ -215,9 +241,26 @@ def _integer_and_none_validator(val):
...
+    _spill_validator,
 )
 
+
+_register_option(
+    "copy_on_write",
+    _env_get_bool("CUDF_COPY_ON_WRITE", False),
+    textwrap.dedent(
+        """
+        Default behavior of performing shallow copies.
```
vyasr, line 253: I think this sentence is redundant, the default is explained later.
```suggestion
```

python/cudf/cudf/utils/applyutils.py
```diff
@@ -105,14 +105,15 @@ def apply_chunks(
...
+@acquire_spill_lock()
 def make_aggregate_nullmask(df, columns=None, op="__and__"):
 
     out_mask = None
     for k in columns or df._data:
         col = cudf.core.dataframe.extract_col(df, k)
         if not col.nullable:
             continue
-        nullmask = df[k].nullmask
+        nullmask = cudf.Series(df[k]._column.nullmask)
```
vyasr, line 116: Can this use `from_data`?

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,238 @@ def test_null_copy():
     col = Series(np.arange(2049))
     col[:] = None
     assert len(col) == 2049
+
+
+@pytest.mark.parametrize("copy_on_write", [True, False])
+def test_series_setitem_cow(copy_on_write):
```
vyasr, line 59: Since basically every assertion in this test is conditional, it may be better to just split it into two tests.

vyasr (changes requested):
OK done reviewing code as well. This is very close!

docs/cudf/source/user_guide/copy-on-write.md
```diff
@@ -0,0 +1,169 @@
...
+>>> import cudf
+>>> cudf.set_option("copy_on_write", True)
+```
+
+ii. Set the environment variable ``CUDF_COPY_ON_WRITE`` to ``1`` prior to the
+launch of the Python interpreter:
+
+```bash
+export CUDF_COPY_ON_WRITE="1" python -c "import cudf"
+```
```
wence-, line 27: Can I switch between CoW on and CoW off in the same run, or is this a one-time option I should set before I actually create and manipulate any cudf options? I suspect it is the latter. If so, we should call that out here.

wence- (commented):
Doc comments.

python/cudf/cudf/_lib/column.pyx
wence-: Yeah, I think this wants to be `data_owner.get_ptr(mode="write")` or (private API, bad) `data_owner._unlink_shared_buffers()`.

python/cudf/cudf/core/buffer/buffer.py
wence-: I agree, so concretely, here we should just do:
```
return self._from_device_memory(rmm.DeviceBuffer(ptr=self.get_ptr(mode="read"), size=self.size))
```

And `SpillableBuffer` should implement `copy`:

```
@acquire_spill_lock # or with ...
def copy(self, deep=True):
    return super().copy(deep=deep)
```

python/cudf/cudf/core/buffer/cow_buffer.py
```diff
@@ -0,0 +1,215 @@
...
+        if deep or self._zero_copied:
+            return self._from_device_memory(
+                rmm.DeviceBuffer(ptr=self._ptr, size=self.size)
+            )
+        else:
+            copied_buf = CopyOnWriteBuffer.__new__(CopyOnWriteBuffer)
```
wence-, line 155:
```suggestion
            cls = type(self)
            copied_buf = cls.__new__(cls)
```
?

In case someone inherits

python/cudf/cudf/core/buffer/spillable_buffer.py
```diff
@@ -552,6 +552,7 @@ def get_ptr(self, *, mode) -> int:
         A passthrough method to `SpillableBuffer.get_ptr`
         with factoring in the `offset`.
         """
+
```
wence-, line 555:
```suggestion
```

python/cudf/cudf/_lib/column.pyx
```diff
@@ -309,6 +313,7 @@ cdef class Column:
             return self._view(libcudf_types.UNKNOWN_NULL_COUNT).null_count()
 
     cdef mutable_column_view mutable_view(self) except *:
+
```
wence-, line 316:
```suggestion
```

python/cudf/cudf/_lib/column.pyx
```diff
@@ -294,6 +297,7 @@ cdef class Column:
         instead replaces the Buffers and other attributes underneath the column
         object with the Buffers and attributes from the other column.
         """
+
```
wence-, line 300:
```suggestion
```

python/cudf/cudf/_lib/column.pyx
```diff
@@ -194,6 +196,7 @@ cdef class Column:
             "The value for mask is smaller than expected, got {}  bytes, "
             "expected " + str(required_num_bytes) + " bytes."
         )
+
```
wence-, line 199:
```suggestion
```

python/cudf/cudf/core/column/column.py
```diff
@@ -418,24 +418,52 @@ def nullmask(self) -> Buffer:
             raise ValueError("Column has no null mask")
         return self.mask_array_view(mode="read")
 
+    def force_deep_copy(self: T) -> T:
+        """
+        A method to create deep copy irrespective of whether
+        `copy-on-write` is enable.
```
wence-, line 424:
```suggestion
        `copy-on-write` is enabled.
```

python/cudf/cudf/core/column/column.py
```diff
@@ -418,24 +418,52 @@ def nullmask(self) -> Buffer:
...
+
+        Parameters
+        ----------
+        deep : bool, default True
+            If True, a true physical copy of the column
+            is made.
+            If False and `copy_on_write` is False, the same
+            memory is shared between the buffers of the Column
+            and changes made to one Column will propagate to
+            it's copy and vice-versa.
```
wence-, line 441:
```suggestion
            its copy and vice-versa.
```

python/cudf/cudf/core/column/column.py
```diff
@@ -418,24 +418,52 @@ def nullmask(self) -> Buffer:
...
+                    if self.base_mask is None
+                    else self.base_mask.copy(deep=False),
                     size=self.size,
                     offset=self.offset,
-                    children=self.base_children,
+                    children=tuple(
+                        col.copy(deep=False) for col in self.base_children
+                    )
+                    if cudf.get_option("copy_on_write")
+                    else self.base_children,
```
wence-, line 466: Would be nice if this could just be
```suggestion
                    children=tuple(
                        col.copy(deep=False) for col in self.base_children
                    ),
```

Why does that not work?

python/cudf/cudf/core/column/column.py
```diff
@@ -1354,11 +1382,12 @@ def column_empty_like(
     ):
         column = cast("cudf.core.column.CategoricalColumn", column)
         codes = column_empty_like(column.codes, masked=masked, newsize=newsize)
+
```
wence-, line 1385:
```suggestion
```

python/cudf/cudf/core/column/datetime.py
```diff
@@ -279,6 +279,7 @@ def as_numerical(self) -> "cudf.core.column.NumericalColumn":
 
     @property
     def __cuda_array_interface__(self) -> Mapping[str, Any]:
+
```
wence-, line 282:
```suggestion
```

python/cudf/cudf/core/column/datetime.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2022, NVIDIA CORPORATION.
+# Copyright (c) 2019-2023, NVIDIA CORPORATION.
```
wence-, line 1:
```suggestion
# Copyright (c) 2019-2022, NVIDIA CORPORATION.
```

python/cudf/cudf/core/column/numerical.py
```diff
@@ -169,6 +168,7 @@ def __setitem__(self, key: Any, value: Any):
 
     @property
     def __cuda_array_interface__(self) -> Mapping[str, Any]:
+
```
wence-, line 171:
```suggestion
```

python/cudf/cudf/core/column/lists.py
```diff
@@ -198,6 +198,11 @@ def _with_type_metadata(
 
         return self
 
+    def copy(self, deep: bool = True):
+        # Since list columns are immutable, both deep and shallow copies share
+        # the underlying device data and mask.
+        return super().copy(deep=False)
```
wence-, line 204: Is there a potential footgun here in that if a mask column is taken from a non-list column in the construction, then this is a lie.

python/cudf/cudf/core/dataframe.py
```diff
@@ -6541,12 +6541,14 @@ def to_struct(self, name=None):
         field_names = [str(name) for name in self._data.names]
 
         col = cudf.core.column.build_struct_column(
-            names=field_names, children=self._data.columns, size=len(self)
+            names=field_names,
+            children=tuple(
+                [col.copy(deep=True) for col in self._data.columns]
```
wence-, line 6546:
```suggestion
                col.copy(deep=True) for col in self._data.columns
```

python/cudf/cudf/core/frame.py
```diff
@@ -1701,9 +1702,10 @@ def _colwise_binop(
                     # that nulls that are present in both left_column and
                     # right_column are not filled.
                     if left_column.nullable and right_column.nullable:
-                        lmask = as_column(left_column.nullmask)
-                        rmask = as_column(right_column.nullmask)
-                        output_mask = (lmask | rmask).data
+                        with acquire_spill_lock():
+                            lmask = as_column(left_column.nullmask)
+                            rmask = as_column(right_column.nullmask)
+                            output_mask = (lmask | rmask).data
```
wence-, line 1708: Did we miss this in the get_ptr refactor?

python/cudf/cudf/core/multiindex.py
```diff
@@ -156,7 +156,7 @@ def __init__(
 
         source_data = {}
         for i, (column_name, col) in enumerate(codes._data.items()):
-            if -1 in col.values:
+            if -1 in col:
```
wence-, line 159: optimisation, iirc

python/cudf/cudf/options.py
wence-: Would need, I think, at least ```:ref:`copy-on-write-user-doc` ```

python/cudf/cudf/testing/_utils.py
```diff
@@ -346,6 +346,11 @@ def get_ptr(x) -> int:
     assert len(lhs.base_children) == len(rhs.base_children)
     for lhs_child, rhs_child in zip(lhs.base_children, rhs.base_children):
         assert_column_memory_eq(lhs_child, rhs_child)
+    if isinstance(lhs, cudf.core.column.CategoricalColumn) and isinstance(
+        rhs, cudf.core.column.CategoricalColumn
+    ):
+        assert_column_memory_eq(lhs.categories, rhs.categories)
+        assert_column_memory_eq(lhs.codes, rhs.codes)
```
wence-, line 353: Decorate this function with `@acquire_spill_lock` while we're here.

wence- (commented):
Some minor code comments, looking good!

docs/cudf/source/user_guide/copy-on-write.md
```diff
@@ -0,0 +1,169 @@
...
+dtype: int64
+>>> s
+0    1
+1    2
+2    3
+3    4
+4    5
+dtype: int64
+```
+2. There are numerous other inconsistencies, which are solved by copy-on-write. Read more about them [here](https://phofl.github.io/cow-introduction.html).
```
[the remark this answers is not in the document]
galipremsagar: > Deep copies no longer copy the data unless the result is modified.

This is no longer the case, deep copies will result in actual copies. Shallow copies will result in a copy upon modification only.

python/cudf/cudf/core/frame.py
galipremsagar: Yup

python/cudf/cudf/core/multiindex.py
galipremsagar: Right

docs/cudf/source/user_guide/copy-on-write.md
galipremsagar: It's the former. Pandas allows that too:
```python
In [1]: import pandas as pd

In [2]: pd.options.mode.copy_on_write = True

In [3]: s = pd.Series([1, 2, 1, 2])

In [4]: pd.options.mode.copy_on_write = True

In [5]: pd.options.mode.copy_on_write = False

In [6]: s.head(2)
Out[6]: 
0    1
1    2
dtype: int64
```

Since all of our buffer creation calls go through, the `as_buffer` constructor we are able to support COW on and off in the same run.

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +314,180 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+
+A read-only object can be quite useful for operations that will not
+mutate the data. This can be achieved by calling `._readonly_proxy_cai_obj`
+API, this API will return a proxy object that has `__cuda_array_interface__`
+implemented and will not trigger a deep copy even if the `CopyOnWriteBuffer`
+has weak references. It is only recommended to use this API as long as
+the objects/arrays created with this proxy object gets cleaned up during
+the developer code execution. We currently use this API for device to host
+copies like in `ColumnBase.data_array_view(mode="read")` which is used for `Column.values_host`.
+
```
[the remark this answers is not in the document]
galipremsagar: Done 👍

python/cudf/cudf/testing/_utils.py
galipremsagar: Done

python/cudf/cudf/core/series.py
galipremsagar: Added a comment, explaining the same.

python/cudf/cudf/core/column/lists.py
galipremsagar: Umm, yes.. but that would be a bug in our construction no? We shouldn't be using a non-list column mask directly to create a list column.

python/cudf/cudf/core/dataframe.py

docs/cudf/source/user_guide/copy-on-write.md
```diff
@@ -0,0 +1,169 @@
...
+|---------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------|
+| `.copy(deep=True)`  | A true copy is made and changes don't propagate to the original object.                                                                                                                            | A true copy is made and changes don't propagate to the original object.                  |
+| `.copy(deep=False)` | Memory is shared between the two objects and but any write operation on one object will trigger a true physical copy before the write is performed. Hence changes will not propagate to the original object. | Memory is shared between the two objects and changes performed on one will propagate to the other object. |
```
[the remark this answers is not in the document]
galipremsagar: Moved it 👍

docs/cudf/source/user_guide/copy-on-write.md
```diff
@@ -0,0 +1,169 @@
...
+
+## How to enable it
+
+i. Use `cudf.set_option`:
```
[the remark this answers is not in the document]
galipremsagar: Yeah, it should be like @wence- mentioned above. Fixed it.

docs/cudf/source/user_guide/copy-on-write.md
```diff
@@ -0,0 +1,169 @@
...
+
+
+## How to disable it
+
+
+Copy-on-write can be disable by setting ``copy_on_write`` cudf option to ``False``:
+
+```python
+>>> cudf.set_option("copy_on_write", False)
+```
```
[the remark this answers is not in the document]
galipremsagar: Done

python/cudf/cudf/_lib/column.pyx
galipremsagar: Fixed it by adding a test for coverage.(test_dataframe_cow_slice_setitem)

python/cudf/cudf/core/buffer/buffer.py
galipremsagar: Done, added a `SpillableBuffer.copy` which holds the lock.

python/cudf/cudf/options.py
galipremsagar: Fixed it. We currently don't have a doc-rendered page for options.

python/cudf/cudf/options.py
galipremsagar: Done

python/cudf/cudf/utils/applyutils.py
galipremsagar: We can just use `as_column` itself, changed it.

python/cudf/cudf/tests/test_copying.py
galipremsagar: Done, split the tests.

python/cudf/cudf/core/buffer/cow_buffer.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class CopyOnWriteBuffer(Buffer):
+    """A Copy-on-write buffer that implements Buffer.
+
+    This buffer enables making copies of data only when there
+    is a write operation being performed.
+
+    See `Copy-on-write` section in `library_design.md` for
+    detailed information on `CopyOnWriteBuffer`.
```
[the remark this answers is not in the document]
galipremsagar: Done, added a link.

python/cudf/cudf/core/column/column.py
galipremsagar: Yup

python/cudf/cudf/core/column/lists.py
wence-: I guess I mean if somehow someone writes:
```
build_column(list_data, ...., mask=mask_from_a_non_list_column)
```
And then (much later), we copy the resulting list column, we've violated the contract on the mask column, I think.

Is there a way that this could, plausibly, happen?

wence-: Or are masks always copied?

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,309 @@ def test_null_copy():
     col = Series(np.arange(2049))
     col[:] = None
     assert len(col) == 2049
+
+
+def test_series_setitem_cow_on():
+    cudf.set_option("copy_on_write", True)
```
wence-, line 59: TODO: As a new issue, can `cudf.set_option` have context manager behaviour?

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,309 @@ def test_null_copy():
...
+        slice_df,
+        cudf.DataFrame(
+            {"a": [11, 1111, 13], "b": [30, 40, 50]}, index=[1, 2, 3]
+        ),
+    )
+    assert_eq(
+        df,
+        cudf.DataFrame({"a": [10, 11, 12, 13, 14], "b": [20, 30, 40, 50, 60]}),
+    )
+    cudf.set_option("copy_on_write", False)
```
wence-, line 361: All these restores should have done `cudf.get_option` to figure out what to restore to. (Otherwise imagine the pain when we flip the default).

python/cudf/cudf/tests/test_index.py
```diff
@@ -410,7 +411,18 @@ def test_index_copy_category(name, dtype, deep=True):
 def test_index_copy_deep(idx, deep):
     """Test if deep copy creates a new instance for device data."""
     idx_copy = idx.copy(deep=deep)
-    if not deep:
+
+    if (
+        isinstance(idx, cudf.StringIndex)
+        or not deep
+        or (cudf.get_option("copy_on_write") and not deep)
```
wence-, line 418: Is this tested?

python/cudf/cudf/tests/test_multiindex.py
```diff
@@ -846,6 +848,7 @@ def test_multiindex_copy_deep(data, deep):
         ]
 
         assert all((x == y) == same_ref for x, y in zip(lptrs, rptrs))
+    cudf.set_option("copy_on_write", False)
```
wence-, line 851: remember what to restore to.

python/cudf/cudf/tests/test_series.py
```diff
@@ -2080,3 +2080,25 @@ def test_series_duplicated(data, index, keep):
...
+    gsr = cudf.from_pandas(psr)
+
+    new_psr = pd.Series(psr, copy=copy)
+    new_gsr = cudf.Series(gsr, copy=copy)
+
+    new_psr.iloc[0] = 999
+    new_gsr.iloc[0] = 999
+
+    assert_eq(psr, gsr)
+    assert_eq(new_psr, new_gsr)
```
wence-, line 2104: What is this testing wrt copy-on-write?

wence- (commented):
Very minor comments (mostly docs wording), this is looking great! I like particularly how small the core implementation is, I'm much happier that we can follow what is going on.

I think there are just one or two more things to clean up from previous rounds and then this feels like it will be good to go.

python/cudf/cudf/core/buffer/cow_buffer.py
```diff
@@ -0,0 +1,215 @@
...
+        """Create a Buffer from an object exposing `__cuda_array_interface__`.
+
+        No data is being copied.
+
+        Parameters
+        ----------
+        data : device-buffer-like
+            An object implementing the CUDA Array Interface.
+        exposed : bool, optional
+            Mark the buffer as zero copied.
```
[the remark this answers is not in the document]
galipremsagar: > Sometimes we know we are construction from device memory that is not zero-copied, iiuc.

Yup, this is only used by `as_buffer`, the caller of `as_buffer` has to let us know if the `data` is being zero-copied or not.

python/cudf/cudf/tests/test_series.py
galipremsagar: Not `cow` related but have introduced `copy` parameter to `Series` constructor, this `pytest` is covering that behavior.

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +315,190 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+(copy-on-write-dev-doc)=
+
+## Copy-on-write
+
+This section describes the internal implementation details of the copy-on-write feature.
+It is recommended that developers familiarize themselves with [the user-facing documentation](copy-on-write-user-doc) of this functionality before reading through the internals
+below.
+
+The core copy-on-write implementation relies on the `CopyOnWriteBuffer` class.
```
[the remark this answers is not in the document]
galipremsagar: I think we did just above no? (`(copy-on-write-user-doc)`)

python/cudf/cudf/_lib/column.pyx
```diff
@@ -515,6 +517,12 @@ cdef class Column:
                     rmm.DeviceBuffer(ptr=data_ptr,
                                      size=(size+offset) * dtype_itemsize)
                 )
+            elif column_owner and isinstance(data_owner, CopyOnWriteBuffer):
+                # TODO: In future, see if we can just pass on the
+                # CopyOnWriteBuffer reference to another column
+                # and still create a weak reference.
+                # With the current design that's not possible.
```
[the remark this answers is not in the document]
galipremsagar: Done: https://github.com/rapidsai/cudf/issues/12734

python/cudf/cudf/core/column/lists.py
galipremsagar: Usually, masks are to be copied while being constructed from another type of column mask.

docs/cudf/source/user_guide/copy-on-write.md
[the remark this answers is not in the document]
galipremsagar: Removed the advantages side heading and merged into notes prose.

python/cudf/cudf/tests/test_copying.py
galipremsagar: Yup, done. Here is the issue: https://github.com/rapidsai/cudf/issues/12736

python/cudf/cudf/tests/test_index.py
galipremsagar: Nice catch, added coverage.

python/cudf/cudf/tests/test_multiindex.py
galipremsagar: done

python/cudf/cudf/tests/test_copying.py
galipremsagar: Yup, good call

python/cudf/cudf/tests/test_multiindex.py
```diff
@@ -781,13 +781,16 @@ def test_multiindex_copy_sem(data, levels, codes, names):
...
-def test_multiindex_copy_deep(data, deep):
+def test_multiindex_copy_deep(data, copy_on_write, deep):
     """Test memory identity for deep copy
     Case1: Constructed from GroupBy, StringColumns
     Case2: Constructed from MultiIndex, NumericColumns
     """
-    same_ref = not deep
+    original_cow_setting = cudf.get_option("copy_on_write")
+    cudf.set_option("copy_on_write", copy_on_write)
+    same_ref = (not deep) or (cudf.get_option("copy_on_write") and not deep)
```
vyasr, line 793: This var seems to no longer be used in the assertion below. Is that intentional/correct? If so, can this line be deleted?

python/cudf/cudf/testing/_utils.py
vyasr: Is this really necessary? The decorator is just an optimization (otherwise anything accessed internally just becomes permanently exposed) and it doesn't seem necessary in a testing utility. Is there a particular reason to do it here?

python/cudf/cudf/core/series.py
```diff
@@ -539,7 +545,28 @@ def __init__(
...
 
         if not isinstance(data, ColumnBase):
-            data = column.as_column(data, nan_as_null=nan_as_null, dtype=dtype)
+            # Using `getattr_static` to check if
+            # `data` is on device memory and perform
+            # a deep copy later. This is different
+            # from `hasattr` because, it doesn't
+            # invoke the property we are looking
+            # for and the later actually invokes
+            # the property.
```
vyasr, line 554:
```suggestion
            # for and the latter actually invokes
            # the property, which in this case could
            # be expensive or mark a buffer as
            # unspillable.
```

python/cudf/cudf/core/buffer/spillable_buffer.py
```diff
@@ -152,7 +152,7 @@ def __getitem__(self, i):
 
 
 class SpillableBuffer(Buffer):
-    """A spillable buffer that implements DeviceBufferLike.
+    """A spillable buffer that implements Buffer.
```
vyasr, line 155:
```suggestion
    """A Buffer that supports spilling memory off the GPU to avoid OOMs.
```

python/cudf/cudf/core/buffer/cow_buffer.py
```diff
@@ -0,0 +1,189 @@
...
+        return {
+            "data": (self._ptr, readonly),
+            "shape": (self.size,),
+            "strides": None,
+            "typestr": "|u1",
+            "version": 0,
+        }
+
+    @property
+    def _readonly_proxy_cai_obj(self) -> dict:
```
vyasr, line 160: Can't this be removed now that we have `_get_cuda_array_interface`?

python/cudf/cudf/core/buffer/cow_buffer.py
```diff
@@ -0,0 +1,189 @@
...
+            copied_buf._ptr = self._ptr
+            copied_buf._size = self._size
+            copied_buf._owner = self._owner
+            copied_buf._finalize_init()
+            return copied_buf
+
+    @property
+    def __cuda_array_interface__(self) -> dict:
+        # Unlink if there are any weak references.
+        self._unlink_shared_buffers()
```
vyasr, line 139: For clarity I would move this line below the comment just below. We wouldn't unlink if we weren't also setting to zero_copied, right? The two changes are coupled here for the same reason.

vyasr: Also, so that I understand, the implication is that accessing the CAI of a buffer that is currently sharing memory with other buffers will immediately trigger a copy, correct? i.e.
```
x = cudf.Series([1, 2, 3])
y = x[:2]
a = cp.asarray(x)  # This will force x to make a deep copy to unlink it from y
```

python/cudf/cudf/core/buffer/cow_buffer.py
```diff
@@ -0,0 +1,189 @@
...
+        Buffer.get_ptr
+        SpillableBuffer.get_ptr
+        """
+        if mode == "read":
+            return self._ptr
+        elif mode == "write":
+            self._unlink_shared_buffers()
+            return self._ptr
+        else:
+            raise ValueError(f"Incorrect mode passed : {mode}")
```
vyasr, line 122: FWIW you could simplify this with 
```suggestion
        if mode == "write":
            self._unlink_shared_buffers()
        elif mode != "read":
            raise ValueError(f"Incorrect mode passed : {mode}")
        return self._ptr
```
Just noting because it came to mind, not really necessary.

docs/cudf/source/user_guide/copy-on-write.md
```diff
@@ -0,0 +1,171 @@
+(copy-on-write-user-doc)=
+
+# Copy-on-write
+
+Copy-on-write is a memory management strategy that allows multiple cuDF objects containing the same data to refer to the same memory address as long as neither of them modify the underlying data.
+With this approach, any operation that generates an unmodified view of an object (such as copies, slices, or methods like `DataFrame.head`) returns a new object that points to the same memory as the original.
+However, when either the existing or new object is _modified_, a copy of the data is made prior to the modification, ensuring that the changes do not propagate between the two objects.
+This behavior is best understood by looking at the examples below.
+
+The default behaviour in cuDF is for copy-on-write to be disabled, so to use it, one must explicitly opt in by setting a cuDF option. This can be done operation by operation if this is required, but the most common use case is to enable copy-on-write for the duration of a script.
```
vyasr, line 10: I think this statement is somewhat misleading, unless I am misunderstanding how COW currently works. While it is true that something like this example is safe:
```
x = cudf.DataFrame(...)
# Do things with x
cudf.set_option(copy_on_write, True)
y = cudf.DataFrame(...)
# Do things with y
```
it has potentially surprising semantics. Operations on `x` won't have COW semantics since it predated the setting, right? In fact, I think that it might be operation dependent; if you trigger a deep copy of `x`, then the newly created buffer will be a CoWBuffer once the option has been set, whereas shallow operations will remain non-COW.

I don't think that we need to change any behavior here, but it is worth noting that behavior could be unexpected and that users should probably be sticking to a single mode for the duration of a script.

docs/cudf/source/user_guide/copy-on-write.md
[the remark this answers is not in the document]
vyasr: I'm confused, I don't see the change here. I still see a numbered list with two items.

docs/cudf/source/developer_guide/library_design.md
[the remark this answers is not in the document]
vyasr: I think @wence- wants a link to the class, something like :class:`cudf.core.buffer.CopyOnWriteBuffer`. AFAIK buffer classes don't actually show up in our rendered docs though since they aren't public APIs, so we may not benefit from this.

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +315,190 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+someone accesses data through the `__cuda_array_interface__`, we eagerly trigger the copy by calling
+`_unlink_shared_buffers` which ensures a true copy of underlying device data is made and
+unlinks the buffer from any shared "weak" references. Any future copy requests must also trigger a true physical copy (since we cannot track the lifetime of the third-party object). To handle this we also mark the `Column`/`CopyOnWriteBuffer` as
+`obj._zero_copied=True` thus indicating that any future shallow-copy requests will trigger a true physical copy
+rather than a copy-on-write shallow copy with weak references.
+
+### Obtaining a read-only object
+
+A read-only object can be quite useful for operations that will not
+mutate the data. This can be achieved by calling `._readonly_proxy_cai_obj`.
```
vyasr, line 349: I don't think this property should exist anymore, it's just the getter function with `readonly=True`.

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -314,3 +315,190 @@ The pandas API also includes a number of helper objects, such as `GroupBy`, `Rol
...
+Since it is unsafe to access the raw pointer associated with a buffer when
+copy-on-write is enabled, in addition to the readonly proxy object described above,
+access to the pointer is gated through `Buffer.get_ptr`. This method accepts a mode
+argument through which the caller indicates how they will access the data associated
+with the buffer. If only read-only access is required (`mode="read"`), this indicates
+that the caller has no intention of modifying the buffer through this pointer.
+In this case, any shallow copies are not unlinked. In contrast, if modification is
+required one may pass `mode="write"`, provoking unlinking of any shallow copies.
+The same mechanism is also used for `SpillableBuffer`, though in that case one must
+also consider acquiring a spill lock (see `Buffer` [design](Buffer-design) for more details).
```
vyasr, line 367: We can remove reference to the spill lock since that's implicit now.

vyasr: Did we update the relevant documentation for SpillableBuffer when we made that change?

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,324 @@ def test_null_copy():
...
     col[:] = None
     assert len(col) == 2049
+
+
+# TODO: Make use of set_option context manager
+# once https://github.com/rapidsai/cudf/issues/12736
+# is resolved.
+
+
+def test_series_setitem_cow_on():
```
vyasr, line 63: This function tests multiple different cases. Can it be split up so that they are tested independently?

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,324 @@ def test_null_copy():
...
+
+    new_slice = actual[2:]
+    assert new_slice._column.base_data._ptr == actual._column.base_data._ptr
+    new_slice[0:2] = 10
+    assert_eq(new_slice, cudf.Series([10, 10, 5], index=[2, 3, 4]))
+    assert_eq(actual, cudf.Series([1, 2, 3, 4, 5]))
+    cudf.set_option("copy_on_write", original_cow_setting)
+
+
+def test_series_setitem_cow_off():
```
vyasr, line 99: Same request as above, please split up.

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,324 @@ def test_null_copy():
...
+
+    new_slice = actual[2:]
+    assert new_slice._column.base_data._ptr == actual._column.base_data._ptr
+    new_slice[0:2] = 10
+    assert_eq(new_slice, cudf.Series([10, 10, 5], index=[2, 3, 4]))
+    assert_eq(actual, cudf.Series([1, 2, 10, 10, 5]))
+    cudf.set_option("copy_on_write", original_cow_setting)
+
+
+def test_multiple_series_cow():
```
vyasr, line 135: This test is... complicated. But I get why it's complicated, and I guess it is worthwhile to have at least one such integration-level test. Maybe an extra comment explaining how at each step of modification one more series is being uncoupled, and then at the end deletion needs to not affect still extant series?

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,324 @@ def test_null_copy():
...
+
+    assert_eq(s3, cudf.Series([10, 20, 30, 40, 50]))
+    assert_eq(s7, cudf.Series([10, 55, 55, 6000, 50]))
+
+    del s3
+    assert_eq(s7, cudf.Series([10, 55, 55, 6000, 50]))
+    cudf.set_option("copy_on_write", original_cow_setting)
+
+
+def test_series_zero_copy_cow_on():
```
vyasr, line 227: Same as above, can you add a comment explaining the purpose of this test.

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,324 @@ def test_null_copy():
...
+    s4 = cudf.Series([10, 20, 30, 40, 50])
+    s5 = cudf.Series(s4)
+    assert_eq(s5, cudf.Series([10, 20, 30, 40, 50]))
+    s5[0:2] = 1
+    assert_eq(s5, cudf.Series([1, 1, 30, 40, 50]))
+    assert_eq(s4, cudf.Series([1, 1, 30, 40, 50]))
+    cudf.set_option("copy_on_write", original_cow_setting)
+
+
+def test_series_zero_copy_cow_off():
```
vyasr, line 264: Comment.

vyasr (changes requested):
Almost there! I mostly have small requests at this point, then I think this is ready to merge.

python/cudf/cudf/testing/_utils.py
galipremsagar: > The decorator is just an optimization (otherwise anything accessed internally just becomes permanently exposed)

That's correct. 

I don't have any strong opinions here since this is a testing utility, let's see what @wence- feels about it.

python/cudf/cudf/tests/test_multiindex.py
galipremsagar: It is being used in the `elif` block below. Moved it into the block.

python/cudf/cudf/tests/test_copying.py
galipremsagar: Done

python/cudf/cudf/tests/test_copying.py
galipremsagar: Done

python/cudf/cudf/tests/test_copying.py
galipremsagar: Done

python/cudf/cudf/tests/test_copying.py
galipremsagar: Done

python/cudf/cudf/tests/test_copying.py
galipremsagar: Done, added detailed comments.

python/cudf/cudf/core/buffer/cow_buffer.py
galipremsagar: > For clarity I would move this line below the comment just below. We wouldn't unlink if we weren't also setting to zero_copied, right? The two changes are coupled here for the same reason.

Right, Done 👍 



> Also, so that I understand, the implication is that accessing the CAI of a buffer that is currently sharing memory with other buffers will immediately trigger a copy, correct? i.e.
> 
> ```
> x = cudf.Series([1, 2, 3])
> y = x[:2]
> a = cp.asarray(x)  # This will force x to make a deep copy to unlink it from y
> ```

Yup, your understanding is correct.

docs/cudf/source/user_guide/copy-on-write.md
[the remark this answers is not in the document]
galipremsagar: Discussed offline and dropped 2nd point as it is a duplication of 1st.

docs/cudf/source/user_guide/copy-on-write.md
galipremsagar: Updated to reflect this.

docs/cudf/source/developer_guide/library_design.md
galipremsagar: > We can remove reference to the spill lock since that's implicit now.

Done



> Did we update the relevant documentation for SpillableBuffer when we made that change?

Yup, we did.

docs/cudf/source/developer_guide/library_design.md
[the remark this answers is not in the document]
galipremsagar: Right, we don't generate docstrings for non-public APIs, so we will not really see a link here anyway.

docs/cudf/source/developer_guide/library_design.md
galipremsagar: Removed this property

python/cudf/cudf/core/buffer/cow_buffer.py
galipremsagar: Yup, removed this from Buffer since we need at two places only in `Column`, made code-changes there accordingly.

python/cudf/cudf/testing/_utils.py
wence-: I don't really mind one way or the other.

python/cudf/cudf/tests/test_series.py
wence-: Thanks!

wence- (approved):
I think this is really good and am happy to hit go. Can you just check that all the comments are resolved before merge? (I provided a few very minor ones here).

python/cudf/cudf/testing/_utils.py
galipremsagar: Thanks, dropped the spill lock here.

docs/cudf/source/user_guide/copy-on-write.md
```diff
@@ -0,0 +1,175 @@
...
+1    2
+2    3
+3    4
+dtype: int64
+```
+
+
+## Notes
+
+When copy-on-write is enabled, there is no concept of views. Modifying any view created inside cuDF will always trigger a copy and will not modify the original object.
```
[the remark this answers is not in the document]
galipremsagar: Do you suggest modifying the above to the following?:

```
When copy-on-write is enabled, modifying any view created inside cuDF will always trigger a copy and will not modify the original object.
```

docs/cudf/source/user_guide/copy-on-write.md
```diff
@@ -0,0 +1,175 @@
...
+0    1
+1    2
+2    3
+3    4
+4    5
+dtype: int64
+
+
+>>> import cudf
+>>> cudf.set_option("copy_on_write", True)
```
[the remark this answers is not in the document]
galipremsagar: Thanks @wence- !

docs/cudf/source/user_guide/copy-on-write.md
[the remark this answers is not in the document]
vyasr: Yeah I had this question too during an earlier round of review, but I wasn't sure how to improve the phrasing here. Technically it is correct: `s2 = s.iloc[:]` creates a view, so future modifications of `s2` are modifications of a view, whereas `s.iloc[:] = 1` does not create any views but modifies `s` directly. I don't know if that's worth spelling out, or if it is a distinction we should expect users to understand.

python/cudf/cudf/core/buffer/cow_buffer.py
vyasr: @galipremsagar unless my GH view isn't updated properly, it looks like the comments were merged but the unlink call is still happening above the comment. Not sure if that was intentional, I would still recommend moving that.

python/cudf/cudf/core/column/column.py
```diff
@@ -2561,3 +2600,14 @@ def concat_columns(objs: "MutableSequence[ColumnBase]") -> ColumnBase:
                 ) from e
             raise
     return col
+
+
+def _proxy_cai_obj(cai, owner):
```
vyasr, line 2605: Add a comment for when this proxy is used vs a proper CAI.

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,395 @@ def test_null_copy():
...
+    original_cow_setting = cudf.get_option("copy_on_write")
+    cudf.set_option("copy_on_write", False)
+    actual = cudf.Series([1, 2, 3, 4, 5])
+    new_copy = actual.copy(deep=False)
+
+    actual[1] = 100
+    assert_eq(actual, cudf.Series([1, 100, 3, 4, 5]))
+    assert_eq(new_copy, cudf.Series([1, 100, 3, 4, 5]))
+    cudf.set_option("copy_on_write", original_cow_setting)
+
```
vyasr, line 86: It might be worth writing a small context manager for use with these tests now, and then replacing them once #12736 is addressed for the more generic `option` solution. Up to you, though.
```suggestion
@contextlib.contextmanager
def with_copy_on_write(on):
    original_cow_setting = cudf.get_option("copy_on_write")
    cudf.set_option("copy_on_write", on)
    yield
    cudf.set_option("copy_on_write", original_cow_setting)

@with_copy_on_write(on=False)
def test_series_setitem_cow_off():
    actual = cudf.Series([1, 2, 3, 4, 5])
    new_copy = actual.copy(deep=False)

    actual[1] = 100
    assert_eq(actual, cudf.Series([1, 100, 3, 4, 5]))    
```

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,395 @@ def test_null_copy():
...
+    assert_eq(s4, cudf.Series([10, 20, 5000, 5000, 50]))
+    assert_eq(s2, cudf.Series([10, 4000, 4000, 4000, 50]))
+    assert_eq(s1, cudf.Series([10000, 10000, 10000, 40, 50]))
+    assert_eq(s6, cudf.Series([3000, 3000, 3000, 40, 50]))
+    for ser in [s3]:
+        assert_eq(ser, cudf.Series([10, 20, 30, 40, 50]))
+
+    # Deleting any of the following series objects
+    # shouldn't delete rest of the weekly referenced data
+    # else-where.
```
vyasr, line 239:
```suggestion
    # elsewhere.
```

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,395 @@ def test_null_copy():
...
+    # Ensure all original data & zero-copied
+    # data is same.
+    assert_eq(s, cudf.Series([1, 2, 3, 4, 5]))
+    assert_eq(s1, cudf.Series([1, 2, 3, 4, 5]))
+    assert_eq(cp_array, cp.array([1, 2, 3, 4, 5]))
+
+    cp_array[0:3] = 10
+    # Modifying a zero-copied array should only
+    # modify `s` and will leave rest of the copies
+    # un-touched.
```
vyasr, line 291:
```suggestion
    # untouched.
```

python/cudf/cudf/tests/test_copying.py
```diff
@@ -52,3 +53,395 @@ def test_null_copy():
...
+    assert_eq(cp_array, cp.array([10, 10, 10, 4, 5]))
+
+    s2 = cudf.Series(cp_array)
+    assert_eq(s2, cudf.Series([10, 10, 10, 4, 5]))
+
+    s3 = s2.copy(deep=False)
+    cp_array[0] = 20
+    # Modifying a zero-copied array should modify
+    # `s2` and `s` only. Because `cp_array`
+    # is zero-copied shared with `s` & `s2`.
```
vyasr, line 304:
```suggestion
    # is zero-copy shared with `s` & `s2`.
```

vyasr (approved):
Some very minor comments, otherwise LGTM!

python/cudf/cudf/core/buffer/cow_buffer.py
```diff
@@ -0,0 +1,170 @@
...
+        self._unlink_shared_buffers()
+        # Unlink if there are any weak references.
+        # Mark the Buffer as ``zero_copied=True``,
+        # which will prevent any copy-on-write
+        # mechanism post this operation.
+        # This is done because we don't have any
+        # control over knowing if a third-party library
+        # has modified the data this Buffer is
+        # pointing to.
+        self._zero_copied = True
```
vyasr, line 145:
```suggestion
        # Unlink if there are any weak references.
        # Mark the Buffer as ``zero_copied=True``,
        # which will prevent any copy-on-write
        # mechanism post this operation.
        # This is done because we don't have any
        # control over knowing if a third-party library
        # has modified the data this Buffer is
        # pointing to.
        self._unlink_shared_buffers()
        self._zero_copied = True
```

python/cudf/cudf/tests/test_copying.py
galipremsagar: Done 👍  Less lines of code too 😉

docs/cudf/source/user_guide/copy-on-write.md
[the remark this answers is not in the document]
wence-: I guess this is because `s.iloc[:] = 1` calls `s.iloc.__setitem__(:, 1)`, whereas
```
s2 = s.iloc[:]
```
Calls `s.iloc.__getitem__(:)` which creates a new object (distinct from `s.iloc`) and now `s2.iloc[:] = 1` doesn't call `s.iloc.__setitem__`.

As noted with all of the copy-on-write discussions in pandas (and here) and all the explanations in https://pandas.pydata.org/docs/user_guide/indexing.html#indexing-view-versus-copy, the indexing on dataframes is actually really hard to grok in general.

In this particular case it's additionally confusing because the same expression can mean different things depending on whether it is an lvalue or rvalue.

In fact, with copy-on-write on, slicing behaves _much more_ like slicing on builtin python objects like lists. So perhaps we should say that, or something along those lines. People might be more used to that:
```python
x = [1,2, 3]
y = x[:] # copy!
y[0] = 5
assert x[0] == 1
```

So how about:

> When copy-on-write is enabled, there is no longer a concept of views when slicing or indexing. In this sense indexing behaves as one would expect for builtin Python containers like lists, rather than indexing numpy arrays. Modifying a "view" created by cuDF will always trigger a copy and will not modify the original object.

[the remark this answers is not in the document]
vyasr: Correct, this is fundamentally about `_setitem__` vs `__getitem__`. The rephrasing looks good to me.

PR #12625, opened 2023-01-26
Fix leaks in ColumnVectorTest

jlowe:
## Description
#12548 introduced a number of GPU resource leaks in ColumnVectorTest.  This cleans them up by wrapping them in `try` blocks.

## Checklist

### DIFF
--- a/java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
+++ b/java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
@@ -4040,16 +4040,17 @@ public class ColumnVectorTest extends CudfTestBase {
 
   @Test
   void testExtractRe() {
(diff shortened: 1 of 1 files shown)


revans2 (approved):

jbrennan333 (approved):
+1 lgtm

ttnghia (approved):
I wonder if we can make the tests failed if there is such resource leak. Maybe set an environment variable to control to throw an exception when leaking is detected.

jlowe:
> I wonder if we can make the tests failed if there is such resource leak.

We already have checks that try to catch this, and that's how we found it.  The problem is that it relies on the behavior of the JVM garbage collector to detect the leak, and apparently that's not always guaranteed to catch it.

PR #12643, opened 2023-01-30
Fix AllocateLikeTest gtests reading uninitialized null-mask

davidwendt:
## Description
Fixes a bug found in the gtests `AllocateLikeTest.ColumnNumericTestSameSize` and `AllocateLikeTest.ColumnNumericTestSpecifiedSize` where uninitialized column null-masks are used to verify null-counts.
This error was found by running the following in the libcudf build directory:
```
gtests/COPYING_TEST --rmm_mode=cuda
```
The `AllocateLikeTest` tests `cudf::allocate_like()` which produces an uninitialized fixed-width column which is usually filled in by internal functions. The uninitialized column data includes its null-mask. The gtest uses `CUDF_TEST_EXPECT_COLUMN_PROPERTIES_EQUAL` to verify the column returned by `allocate_like()` which also includes the `null_count()`. The null-count property returned by `allocate_like()` sets the null-count to `UNKNOWN_NULL_COUNT` which causes a future call to `null_count()` to calculate the count by reading the bits in the null-mask. Since the null-mask is uninitialized (garbage), the null-count would be invalid and likely not match any predictable value.
The gtests are updated to check for appropriate column properties excluding the null-count.

While debugging this issue, the `allocate_like()` logic was found to also attempt to build child columns incorrectly assuming the children will contain the same size and null-mask. Since the `allocate_like()` only supports fixed-width types which do not have children, this logic was removed.

## Checklist


vuule (approved):

vyasr (approved):
Nice find! I see two other uses of `allocate_like` in `column_device_view_test.cu`, but in that case it looks like the columns are filled before any comparison so I think this isn't an issue, right?

davidwendt:
> Nice find! I see two other uses of `allocate_like` in `column_device_view_test.cu`, but in that case it looks like the columns are filled before any comparison so I think this isn't an issue, right?

Yes, it is not an issue there because the uninitialized data is not read and there is no null-mask.

PR #12647, opened 2023-01-30
[REVIEW] Fix a mask data corruption in UDF

galipremsagar:
## Description

This PR fixes a mask data corruption issue encountered while working on https://github.com/rapidsai/cudf/pull/12619.

`column_empty` actually returns un-initialized values, but we would want a fully initialized mask column here. Hence it is very inconsistent to test for a failure here, which is the reason it has been passing in CI and many of the local builds.

## Checklist


brandon-b-miller (approved):

shwina (approved):

wence-:
Nice debugging. Is there a (or some) test(s) we could run under compute-sanitizer that might have spotted this early?

PR #12659, opened 2023-01-31
Pin wheel dependencies to same RAPIDS release

sevagh:
Now that wheel CI is enabled, the nightly index will contain next release wheels, so we need to ensure those don't get installed while still building this release.


ajschmidt8 (approved):

vyasr (approved):

msadang (approved):
LGTM

PR #11452, opened 2022-08-03
Implement groupby apply with JIT

bwyogatama:
## Description

Experimental cuDF Groupby Apply JIT pipeline.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #12669, opened 2023-02-01
Move `strings_udf` code into cuDF

brandon-b-miller:
With the merge of https://github.com/rapidsai/cudf/pull/11452 we have the machinery to build and deploy PTX libraries of shim functions as part of cuDF's build process. With this there is no reason to keep the `strings_udf` code separate anymore. This PR removes the separate package and all of it's related CI plumbing as well as supports the strings feature by default, just like GroupBy.


python/cudf/cudf/core/udf/__init__.py
```diff
@@ -6,10 +6,16 @@
 from numba.cuda.cudaimpl import lower as cuda_lower
 
 from cudf.core.dtypes import dtype
-from cudf.core.udf import api, row_function, utils
-from cudf.utils.dtypes import STRING_TYPES
-
-from . import groupby_lowering, groupby_typing, masked_lowering, masked_typing
+from cudf.core.udf import api, utils
+
+from . import (
```
bdice, line 11: This is starting to make _lots_ more sense now that more Numba JIT features are being handled in the same way.

python/cudf/cudf/core/udf/utils.py
```diff
@@ -19,32 +21,52 @@
...
     TIMEDELTA_TYPES,
 )
 from cudf.utils.utils import _cudf_nvtx_annotate
 
+# Maximum size of a string column is 2 GiB
+_STRINGS_UDF_DEFAULT_HEAP_SIZE = os.environ.get(
+    "STRINGS_UDF_HEAP_SIZE", 2**31
+)
+heap_size = 0
+cudf_str_dtype = dtype(str)
```
bdice, line 51: Should these be public? The second one looks like it should be a module constant.
```suggestion
_heap_size = 0
_CUDF_STR_DTYPE = dtype(str)
```

python/cudf/cudf/core/udf/strings_typing.py
```diff
@@ -1,226 +1,273 @@
...
-    string_view,
-    udf_string,
-)
+import rmm
 
-from cudf.core.udf import masked_typing
-from cudf.core.udf._ops import comparison_ops
-from cudf.core.udf.masked_typing import MaskedType
+# libcudf size_type
+size_type = types.int32
```
bdice, line 15: Let's move this definition next to where we hardcode other libcudf information like `size_type_dtype` (or reuse that declaration, if possible). https://github.com/rapidsai/cudf/blob/6a67e8fe49661f7f2f6d8da119996f683a21803a/python/cudf/cudf/_lib/types.pyx#L20

bdice (commented):
This is so nice! I have a few minor comments, some of which I think were deferred from the original `strings_udf` PR.

I assumed that most of the code changes in this PR were pure moves, and I didn't look too closely at the things that seemed familiar.

bdice:
@brandon-b-miller Can we rename this PR? “Sunset” sounds like a deprecation to me. I’d propose “Merge strings_udf into cudf package” or similar.

I also labeled this as `breaking` because it’s changing what packages are built and deployed.

bdice:
@brandon-b-miller Also can we file (or update) a companion issue that tracks any needed changes to resources outside this repo like blogs, RAPIDS website docs, etc. that tell users to install strings_udf?

python/cudf/cudf/core/udf/masked_typing.py
```diff
@@ -423,3 +429,203 @@ def generic(self, args, kws):
 
 for unary_op in unary_ops:
     cuda_decl_registry.register_global(unary_op)(MaskedScalarUnaryOp)
+
+
+# Strings functions and utilities
+def _is_valid_string_arg(ty):
```
vyasr, line 435: I'm a bit confused. It looks like the contents of strings_udf/_typing was moved into core/udf/strings_typing and then the old contents of strings_typing were moved into this file. Could you comment on the rationale for the current separation? Is there a distinction between strings and nullable strings logic being made?

vyasr (changes requested):
Looks like there are still some references to strings_udf in this repo based on a git grep. Let me know if you need help tracking everything down.

Excited to see this happening!

python/cudf/cudf/core/udf/masked_typing.py
brandon-b-miller: Yes! This is a great question. Our `MaskedType` extension is the type we use to carry around a value and a validity. When using `DataFrame.apply` or `Series.apply`, the scalar inside your UDF are really `MaskedType`:



```python
df = cudf.DataFrame(
    {
        'a':[1, 2, 3], 
        'b':['a','b','c']  # one string column 'b'
}) 
def f(row):
    x = row['a'] # MaskedType(int64)
    y = row['b'] # MaskedType(string_view)
    z = y.upper() # MaskedType(udf_string), not used just for demo
    return x + len(y)
```

I always thought the cleanest way of putting this together was to have the `string_view` and `udf_string` types actually implement the string methods, like `upper` or `len`, whereas the `MaskedType` that carries it around holds the nullability logic. At that point calling something like `len(MaskedType(string_view))` is programmed in the lowering to translate roughly to:

```python
len(MaskedType(string_view, valid=True)) = MaskedType(len(string_view), valid=True)
```
meaning it ends up taking the validity of the source. That would make our strings just another type of scalar that could be makes and everything would be great, ideally.

The requirement of the external package made a lot of this way harder though since we needed a lot of this to be optional. Before this PR we had this happening:

- string scalars and their ops defined in `strings_udf` 
- Masked string operations defined in `cudf` (in `strings_typing.py` and `strings_lowering.py` deliberately in a separate place so that it could be optionally imported/registered
- `cudf` optionally importing `strings_typing` and `strings_lowering` which contained all the overloads for _masked_ strings

Now what we have is:
- string scalars and their ops defined in `strings_typing.py` and `strings_lowering.py`
- Masked string operations defined in `masked_typing.py` and `masked_lowering.py` 
- Nothing is optional

I think the second way is a lot cleaner.

python/cudf/cudf/core/udf/utils.py
```diff
@@ -420,3 +442,35 @@ def maybe_patch_numba_linker(
...
+        (ret,) = cudart.cudaDeviceSetLimit(
+            cudart.cudaLimit.cudaLimitMallocHeapSize, size
+        )
+        if ret.value != 0:
+            raise RuntimeError("Unable to set cudaMalloc heap size")
+
+        heap_size = size
+
+
+@lru_cache(maxsize=None)
```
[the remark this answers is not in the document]
brandon-b-miller: Yup exactly, it's a hack to make sure it's only run once. Will look into initfunc.

bdice:
> This is quite a large PR. Can we split it into smaller PRs somehow?

Most of this PR is a pure move, but the pieces are pretty intertwined and it would be hard to break up. The code has been previously reviewed by several folks, so it's all familiar and shouldn't be hard to review.

ci/test_python_other.sh
```diff
@@ -44,41 +44,4 @@ pytest \
...
-  --cov-config=../.coveragerc \
-  --cov=cudf \
-  --cov-report=xml:"${RAPIDS_COVERAGE_DIR}/cudf-strings-udf-coverage.xml" \
-  --cov-report=term \
-  tests/test_udf_masked_ops.py
-popd
-
-rapids-logger "Test script exiting with value: $EXITCODE"
-exit ${EXITCODE}
+exit ${SUITEERROR}
```
ajschmidt8, line 47:
```suggestion
exit ${EXITCODE}
```

ajschmidt8 (changes requested):
minor change needed for the CI script updates

python/cudf/cudf/core/udf/masked_typing.py
wence-: Idle thought, not for implementation in this PR, is there a way we can take advantage of what looks to be the functorial nature of the `MaskedType` object to simplify the code and reuse that from `strings_typing`/`lowering` more systematically?

python/cudf/cudf/core/udf/masked_typing.py
```diff
@@ -46,6 +57,9 @@
 )
 supported_type_str = "\n".join(sorted(list(SUPPORTED_NUMPY_TYPES) + ["bool"]))
 MASKED_INIT_MAP: Dict[Any, Any] = {}
+MASKED_INIT_MAP[types.pyobject] = string_view
```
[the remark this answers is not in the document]
brandon-b-miller: This currently happens during the process that creates the `row` type used by the kernel based on the DataFrame that's calling the UDF [here](https://github.com/rapidsai/cudf/blob/branch-23.04/python/cudf/cudf/core/udf/row_function.py#L61). When including a string column, one of the dtypes being input is `np.dtype('O')`. The numba utility function `from_dtype` maps this to `types.pyobject`, so this was intended to catch that case. But I suppose it's better thinking about it now to do this inside the function that produces the row type instead. 

I updated this in https://github.com/rapidsai/cudf/pull/12669/commits/f424b0a4acb7a0ddeff659d265d687e56ed648a1.

ci/test_python_other.sh
```diff
@@ -44,41 +44,4 @@ pytest \
...
-  --numprocesses=8 \
-  --dist=loadscope \
-  --cov-config=../.coveragerc \
-  --cov=cudf \
-  --cov-report=xml:"${RAPIDS_COVERAGE_DIR}/cudf-strings-udf-coverage.xml" \
-  --cov-report=term \
-  tests/test_udf_masked_ops.py
-popd
-
-rapids-logger "Test script exiting with value: $EXITCODE"
```
bdice, line 83: Don't delete this line.

conda/recipes/strings_udf/meta.yaml
```diff
@@ -1,78 +0,0 @@
...
-  run:
-    - python
-    - typing_extensions
-    - numba >=0.54
-    - numpy
-    - libcudf ={{ version }}
-    - cudf ={{ version }}
-    - {{ pin_compatible('cudatoolkit', max_pin='x', min_pin='x') }}
-    - cachetools
-    - ptxcompiler >=0.7.0  # CUDA enhanced compatibility. See https://github.com/rapidsai/ptxcompiler
```
bdice, line 65: We don't require `ptxcompiler >=0.7.0` in the `cudf` recipe. We should, right?

python/cudf/cudf/core/udf/utils.py
```diff
@@ -420,3 +441,30 @@ def maybe_patch_numba_linker(
...
+        _heap_size = size
+
+
+def column_to_string_view_array_init_heap(col):
+    # lazily allocate heap only when a string needs to be returned
+    return column_to_string_view_array(col)
+
+
+strings_ptx_file = _get_ptx_file(os.path.dirname(__file__), "shim_")
+ptx_files.append(strings_ptx_file)
```
bdice, line 470: Why don't we just initialize `ptx_files` with this value above? We don't need to initialize-and-modify (append).

bdice (changes requested):
Packaging questions. Changes are probably needed.

conda/recipes/strings_udf/meta.yaml
brandon-b-miller: Added this. Currently this is being inherited through `cubinlinker`.

python/cudf/cudf/core/udf/utils.py
brandon-b-miller: I refactored this a bit in https://github.com/rapidsai/cudf/pull/12669/commits/c304f5c75165ae1528066765b2b73d6e45a14da5 .

bdice (approved):
Looks good to me! Thanks for your iterations, @brandon-b-miller.

python/cudf/cudf/core/udf/masked_typing.py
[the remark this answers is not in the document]
vyasr: Yup I think the new approach is cleaner and better scoped.

python/cudf/cudf/core/udf/masked_typing.py
vyasr: Thanks for the great explanation! That makes sense, and the new way is certainly simpler without the old extra package.

python/cudf/udf_cpp/CMakeLists.txt
```diff
@@ -19,29 +19,29 @@ include(rapids-cpm)
 include(rapids-cuda)
 include(rapids-find)
 
-rapids_cuda_init_architectures(strings-udf-cpp)
+rapids_cuda_init_architectures(udf-cpp)
```
vyasr, line 22: You should now be able to get rid of
- Any rapids-cuda
- The `project`

vyasr (changes requested):
One small change then I think this is good to go.

python/cudf/udf_cpp/CMakeLists.txt
brandon-b-miller: made a few cmake updates in https://github.com/rapidsai/cudf/pull/12669/commits/9b175e90eaac3c581a98de179bd91a428ab7b961, there's also some updates I needed to make to how things are `include`ed in that commit, but I think they're unrelated.

ajschmidt8 (approved):

PR #12680, opened 2023-02-02
Add segmented reduction support for fixed-point types

davidwendt:
## Description
Depends on #12573 

Adds additional support for fixed-point types in `cudf::segmented_reduce` for simple aggregations: sum, product, and sum-of-squares.
Reference: #10432 

## Checklist


cpp/src/reductions/segmented/compound.cuh
```diff
@@ -64,21 +65,13 @@ std::unique_ptr<column> compound_segmented_reduction(column_view const& col,
...
-    thrust::adjacent_difference(
-      rmm::exec_policy(stream), offsets.begin() + 1, offsets.end(), valid_counts.begin());
-  }
+  rmm::device_uvector<size_type> valid_counts =
+    cudf::reduction::detail::segmented_valid_counts(*d_col,
+                                                    col.has_nulls(),
+                                                    offsets,
+                                                    null_handling,
+                                                    stream,
+                                                    rmm::mr::get_current_device_resource());
```
hyperbolic2346, line 74: Is there a reason to not use the memory resource passed to this function?

cpp/src/reductions/segmented/simple.cuh
```diff
@@ -214,23 +212,57 @@ std::unique_ptr<column> fixed_point_segmented_reduction(
...
+      // The output scale needs to be the maximum valid count of all segments multiplied by
+      // the input scale value.
+      auto d_col = column_device_view::create(col, stream);
+      rmm::device_uvector<size_type> valid_counts =
+        cudf::reduction::detail::segmented_valid_counts(*d_col,
+                                                        col.has_nulls(),
+                                                        offsets,
+                                                        null_handling,
+                                                        stream,
+                                                        rmm::mr::get_current_device_resource());
```
hyperbolic2346, line 229: same question here about memory resource

cpp/src/reductions/segmented/simple.cuh
```diff
@@ -214,23 +212,57 @@ std::unique_ptr<column> fixed_point_segmented_reduction(
...
-  CUDF_FAIL("Segmented reduction on fixed point column only supports min and max reduction.");
+    if constexpr (std::is_same_v<Op, cudf::reduction::op::sum_of_squares>) {
+      return numeric::scale_type{col.type().scale() * 2};
+    }
+
+    return numeric::scale_type{col.type().scale()};
+  }();
+
+  auto size       = result->size();        // get these before
+  auto null_count = result->null_count();  // release() is called
```
hyperbolic2346, line 256:
```suggestion
  auto const size       = result->size();        // get these before
  auto const null_count = result->null_count();  // release() is called
```

cpp/src/reductions/segmented/valid_counts.cu
```diff
@@ -0,0 +1,54 @@
...
+                                                      bool has_nulls,
+                                                      device_span<size_type const> offsets,
+                                                      null_policy null_handling,
+                                                      rmm::cuda_stream_view stream,
+                                                      rmm::mr::device_memory_resource* mr)
+{
+  auto const num_segments = offsets.size() - 1;
+
+  if (has_nulls && (null_handling == null_policy::EXCLUDE)) {
+    return cudf::detail::segmented_count_bits(d_col.null_mask(),
```
hyperbolic2346, line 37: If we only use `d_col` for the null mask here, why do we bother to build the device view?

cpp/tests/reductions/segmented_reduction_tests.cpp
```diff
@@ -1094,223 +1094,220 @@ struct SegmentedReductionFixedPointTest : public cudf::test::BaseFixture {
...
     auto const input =
       cudf::test::fixed_point_column_wrapper<RepType>({1, 2, 3, 1, XXX, 3, 1, XXX, XXX, XXX},
                                                       {1, 1, 1, 1, 0, 1, 1, 0, 0, 0},
                                                       numeric::scale_type{scale});
-    auto const offsets = std::vector<cudf::size_type>{0, 3, 6, 7, 8, 10, 10};
-    auto const d_offsets =
-      cudf::detail::make_device_uvector_async(offsets, cudf::get_default_stream());
-    auto out_type     = cudf::column_view(input).type();
-    auto const expect = cudf::test::fixed_point_column_wrapper<RepType>(
+    auto out_type = cudf::column_view(input).type();
```
hyperbolic2346, line 1111:
```suggestion
    auto const out_type = cudf::column_view(input).type();
```

cpp/tests/reductions/segmented_reduction_tests.cpp
```diff
@@ -1094,223 +1094,220 @@ struct SegmentedReductionFixedPointTest : public cudf::test::BaseFixture {
...
+
   for (auto scale : {-2, 0, 5}) {
     auto const input =
       cudf::test::fixed_point_column_wrapper<RepType>({1, 2, 3, 1, XXX, 3, 1, XXX, XXX, XXX},
                                                       {1, 1, 1, 1, 0, 1, 1, 0, 0, 0},
                                                       numeric::scale_type{scale});
-    auto const offsets = std::vector<cudf::size_type>{0, 3, 6, 7, 8, 10, 10};
-    auto const d_offsets =
-      cudf::detail::make_device_uvector_async(offsets, cudf::get_default_stream());
+    auto out_type = cudf::column_view(input).type();
```
hyperbolic2346, line 1139:
```suggestion
    auto const out_type = cudf::column_view(input).type();
```

cpp/tests/reductions/segmented_reduction_tests.cpp
```diff
@@ -1094,223 +1094,220 @@ struct SegmentedReductionFixedPointTest : public cudf::test::BaseFixture {
...
+      cudf::test::fixed_point_column_wrapper<RepType>({-10, 0, 33, 100, XXX, 53, 11, XXX, XXX, XXX},
                                                       {1, 1, 1, 1, 0, 1, 1, 0, 0, 0},
                                                       numeric::scale_type{scale});
-    auto const offsets = std::vector<cudf::size_type>{0, 3, 6, 7, 8, 10, 10};
-    auto const d_offsets =
-      cudf::detail::make_device_uvector_async(offsets, cudf::get_default_stream());
-    auto out_type     = cudf::column_view(input).type();
-    auto const expect = cudf::test::fixed_point_column_wrapper<RepType>(
-      {1, 1, 1, XXX, XXX, XXX}, {1, 1, 1, 0, 0, 0}, numeric::scale_type{scale});
+    auto out_type = cudf::column_view(input).type();
```
hyperbolic2346, line 1217:
```suggestion
    auto const out_type = cudf::column_view(input).type();
```

cpp/tests/reductions/segmented_reduction_tests.cpp
```diff
@@ -1094,223 +1094,220 @@ struct SegmentedReductionFixedPointTest : public cudf::test::BaseFixture {
...
-    auto const d_offsets =
-      cudf::detail::make_device_uvector_async(offsets, cudf::get_default_stream());
-    auto out_type     = cudf::column_view(input).type();
-    auto const expect = cudf::test::fixed_point_column_wrapper<RepType>(
-      {1, 1, XXX}, {1, 1, 0}, numeric::scale_type{scale});
+    auto input =
+      cudf::test::fixed_point_column_wrapper<RepType>({-10, 0, 33, 100, XXX, 53, 11, XXX, XXX, XXX},
+                                                      {1, 1, 1, 1, 0, 1, 1, 0, 0, 0},
+                                                      numeric::scale_type{scale});
+    auto out_type = cudf::column_view(input).type();
```
hyperbolic2346, line 1291:
```suggestion
    auto const out_type = cudf::column_view(input).type();
```

hyperbolic2346 (changes requested):
Some questions and a few nit pics about const, but otherwise looks good to me.

cpp/src/reductions/segmented/simple.cuh
davidwendt: From https://github.com/rapidsai/cudf/blob/branch-23.04/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#temporary-memory

> Always use the default resource obtained from `rmm::mr::get_current_device_resource()` for temporary memory allocations

cpp/src/reductions/segmented/valid_counts.cu
davidwendt: Good point. We only really need to pass the null-mask.

cpp/src/reductions/segmented/compound.cuh
hyperbolic2346: Ah, I didn't realize they were temporaries.

cpp/src/reductions/segmented/simple.cuh
```diff
@@ -214,23 +212,57 @@ std::unique_ptr<column> fixed_point_segmented_reduction(
...
+      auto d_col = column_device_view::create(col, stream);
+      rmm::device_uvector<size_type> valid_counts =
+        cudf::reduction::detail::segmented_valid_counts(*d_col,
+                                                        col.has_nulls(),
+                                                        offsets,
+                                                        null_handling,
+                                                        stream,
+                                                        rmm::mr::get_current_device_resource());
+
+      auto max_valid_count = thrust::reduce(rmm::exec_policy(stream),
```
divyegala, line 231: I think it's a bit more expressive to use `thrust::max_element` here directly

cpp/src/reductions/segmented/simple.cuh
```diff
@@ -214,23 +212,57 @@ std::unique_ptr<column> fixed_point_segmented_reduction(
...
+  auto null_count = result->null_count();  // release() is called
+  auto contents   = result->release();
+
+  result = std::make_unique<column>(data_type{type_to_id<InputType>(), scale},
+                                    size,
+                                    std::move(*(contents.data.release())),
+                                    std::move(*(contents.null_mask.release())),
+                                    null_count);
+
+  return result;
```
divyegala, line 265: Is it possible to shorten these lines like this?
```suggestion
return std::make_unique<column>(data_type{type_to_id<InputType>(), scale},
                                                            size,
                                                            std::move(*(result->data.release())),
                                                            std::move(*(result->null_mask.release())),
                                                            null_count);
```

cpp/src/reductions/segmented/simple.cuh
davidwendt: I disagree. The `max_element` does not return an answer but a pointer/iterator to the element. I prefer `reduce` do that step for me.

divyegala: That's fair, since it looks like the result is needed on host right afterwards

hyperbolic2346 (approved):

cpp/src/reductions/segmented/valid_counts.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * of elements in each segment.
+ *
+ * @param null_mask Null values over which the segment offsets apply
+ * @param has_nulls True if d_col contains any nulls
+ * @param offsets Indices to segment boundaries
+ * @param null_handling How null entries are processed within each segment
+ * @param stream Used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ */
+rmm::device_uvector<size_type> segmented_valid_counts(bitmask_type const* null_mask,
```
ttnghia, line 45: I'm a bit confused. So when `null_handling != null_policy::EXCLUDE`, this is just a size count, isn't it? If so then the name `_valid_counts` is a bit misleading.

ttnghia: Other comment: This can also be adopted in sort-based groupby for `COUNT_VALID` and `COUNT_ALL`.

davidwendt: It returns number of valid rows in each segment in both cases.

ttnghia: But if `has_null == false || null_handling != null_policy::EXCLUDE` I see it just does this:
```
 rmm::device_uvector<size_type> valid_counts(num_segments, stream, mr);
  thrust::adjacent_difference(
    rmm::exec_policy(stream), offsets.begin() + 1, offsets.end(), valid_counts.begin());
```

ttnghia: My confusion comes from when nulls are considered as `INCLUDE`. Even in such cases, nulls are still "invalid elements", right?

davidwendt: Ultimately, that is up to the caller how to treat those elements. This is just providing the appropriate count based on the parameters.

bdice: Valid counts should always count set bits in the null mask, and only use segment lengths via adjacent difference if the column has no nulls / is non-nullable. Counting valid elements does not depend on a “null policy” at all. This parameter should be removed.

bdice (changes requested):

cpp/src/reductions/segmented/valid_counts.hpp
bdice: Is this function’s special behavior with null handling used more than once? (I can’t count the number of invocations right now on mobile.) I would accept “segment counts” but am unsure if factoring out the null handling is highly reusable.

davidwendt: Yes, it is used for computing the values (e.g. denominator(s)) for multiple compound aggregations. It is also used for fixed-point product aggregation to help compute the scale size for the output data type. All usages are in this PR. I'm happy with `segmented_counts()` which is more consistent with other utilities here.

ttnghia: `segment_counts` sounds good to me.

cpp/src/reductions/segmented/simple.cuh
```diff
@@ -214,23 +212,57 @@ std::unique_ptr<column> fixed_point_segmented_reduction(
...
-    col, offsets, null_handling, init, stream, mr);
-}
+  auto result =
+    simple_segmented_reduction<RepType, RepType, Op>(col, offsets, null_handling, init, stream, mr);
+  auto const scale = [&] {
+    if constexpr (std::is_same_v<Op, cudf::reduction::op::product>) {
+      // The product aggregation requires updating the scale of the fixed-point output column.
+      // The output scale needs to be the maximum valid count of all segments multiplied by
+      // the input scale value.
+      rmm::device_uvector<size_type> valid_counts =
```
bdice, line 222: We should use actual valid counts and not “segment counts” here. I believe the result scale should only depend on the maximum number of “actually valid” elements in any segment, regardless of null policy.

cpp/src/reductions/segmented/simple.cuh
```diff
@@ -214,23 +212,57 @@ std::unique_ptr<column> fixed_point_segmented_reduction(
...
+        cudf::reduction::detail::segmented_valid_counts(col.null_mask(),
+                                                        col.has_nulls(),
+                                                        offsets,
+                                                        null_handling,
+                                                        stream,
+                                                        rmm::mr::get_current_device_resource());
+
+      auto max_valid_count = thrust::reduce(rmm::exec_policy(stream),
```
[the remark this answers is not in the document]
ttnghia: Should be all `const` here, as I don't see any variable needs to be modified, even `d_col` below.

cpp/src/reductions/segmented/simple.cuh
```diff
@@ -214,23 +212,57 @@ std::unique_ptr<column> fixed_point_segmented_reduction(
...
+  auto const null_count = result->null_count();  // release() is called
+  auto contents         = result->release();
+
+  result = std::make_unique<column>(data_type{type_to_id<InputType>(), scale},
+                                    size,
+                                    std::move(*(contents.data.release())),
+                                    std::move(*(contents.null_mask.release())),
+                                    null_count);
+
+  return result;
```
ttnghia, line 265: I wonder if we ever need the variable `result` at all.

davidwendt:
I believe I have everything fixed up now. 
I also modified the product-aggregation gtest to ensure the counts did not include null-entries as well.

ttnghia (approved):

cpp/src/reductions/segmented/compound.cuh
```diff
@@ -64,21 +65,13 @@ std::unique_ptr<column> compound_segmented_reduction(column_view const& col,
   auto out_itr = result->mutable_view().template begin<ResultType>();
 
   // Compute valid counts
```
bdice, line 67:
```suggestion
  // Compute segmented counts
```

cpp/src/reductions/segmented/compound.cuh
```diff
@@ -64,21 +65,13 @@ std::unique_ptr<column> compound_segmented_reduction(column_view const& col,
...
-                                              offsets.end(),
-                                              valid_counts.data(),
-                                              thrust::plus<size_type>{},
-                                              0,
-                                              stream);
-  } else {
-    thrust::adjacent_difference(
-      rmm::exec_policy(stream), offsets.begin() + 1, offsets.end(), valid_counts.begin());
-  }
+  rmm::device_uvector<size_type> valid_counts =
```
bdice, line 68: Please rename this variable to match the new function name (like the function, it doesn't always represent "valid" counts).
```suggestion
  rmm::device_uvector<size_type> segmented_counts =
```

cpp/src/reductions/segmented/counts.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+namespace reduction {
+namespace detail {
+
+/**
+ * @brief Compute the number of elements per segment
+ *
+ * If `null_handling == null_policy::EXCLUDE`, the count for each
+ * segment omits any null entries. Otherwise, this returns the number
+ * of elements in each segment.
```
bdice, line 36: This is an excellent description. I'm glad we made this change.

bdice (approved):
I have two requests for fixing names in comments/variables, then this can be merged.

PR #12681, opened 2023-02-02
[JNI] Remove child fom newCudaAsyncMemoryResource

abellina:
Fixes a bug introduced here https://github.com/rapidsai/cudf/pull/12632 where the C++ version of the jni call `Rmm.newCudaAsyncMemoryResource` was taking an extra argument (`child`), causing the `threshold` to be set to a random value since the calling code defined the native method has taking 2 longs, not 3.

The above can affect performance if the threshold is set to 0 for example. The async pool will reduce its footprint by releasing memory, which means we need to go back and reallocate. This is how I found it.

I am starting the build process for this, so it's going to take me a bit to test.


abellina:
@revans2 @jlowe this should be ok to go in now.

abellina:
build

jlowe (approved):

abellina:
Dask/python tests failing it looks like.

abellina:
Upmerged to see if that helps the CI pass, since I see some related python changes went in.

PR #12627, opened 2023-01-26
[WIP] Adaptive fragment sizes in Parquet writer

etseidl:
## Description
Trying to write Parquet files where rows are very wide can result in pages that are much too large due to the default fragment size of 5000.  This in turn can have an adverse effect on file size when using Zstandard compression.  This PR attempts to address this by modifying the fragment size to a value where each fragment will still fit in the desired page size.

Fixes #12613

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

Issue #12590, opened 2023-01-23
[FEA] Support reading FIXED_LEN_BYTE_ARRAY as binary in read_parquet


GregoryKimball:
**Is your feature request related to a problem? Please describe.**
Currently, the libcudf parquet reader can only interpret FIXED_LEN_BYTE_ARRAY physical types as decimal. Spark3.4 includes a ([feature](https://issues.apache.org/jira/browse/SPARK-41096)) to read parquet FIXED_LEN_BYTE_ARRAY as binary type (i.e. list of int8 in libcudf). I would like to support reading FIXED_LEN_BYTE_ARRAY as binary in read_parquet

**Describe the solution you'd like**
Add FIXED_LEN_BYTE_ARRAY to list of types that can be read binary (list of int8) in libcudf read_parquet.

**Describe alternatives you've considered**
n/a

**Additional context**
Spark3.4 includes a ([feature](https://issues.apache.org/jira/browse/SPARK-41096)) to read parquet FIXED_LEN_BYTE_ARRAY as binary type (i.e. list of int8 in libcudf). The Spark-RAPIDS team has also opened an [issue](https://github.com/NVIDIA/spark-rapids/issues/7449), to provide a little more context. It seems like [11160](https://github.com/rapidsai/cudf/pull/11160) added support for reading variable-length BYTE_ARRAY, so adding FIXED_LEN_BYTE_ARRAY may not be too complex.

GregoryKimball:
@revans2 Do you have any sample files that would be useful for developing this feature?

revans2:
Sure. Here is what I have for a simple file. I can generate more if you want/need me to.

[FIXED_BIN_TEST.parquet.zip](https://github.com/rapidsai/cudf/files/11019725/FIXED_BIN_TEST.parquet.zip)

```
scala> spark.read.parquet("/data/tmp/FIXED_BIN_TEST.parquet").show()
+--------+                                                                      
|bin_test|
+--------+
| [41 42]|
| [31 32]|
+--------+
```

Python/pandas

```
>>> pd.read_parquet("/data/tmp/FIXED_BIN_TEST.parquet")
  bin_test
0    b'AB'
1    b'12'
```

PR #12685, opened 2023-02-02
Variable fragment sizes for Parquet writer

### LINKED ISSUE #12613: [BUG] Compressing a table with larger strings with ZSTD fails to compress
**Describe the bug**
This is a follow up to https://github.com/rapidsai/cudf/issues/12249
While current 23.02 snapshot builds do work for the case described in https://github.com/rapidsai/cudf/issues/12249, the customer provided us with another file that fails to compress on the GPU.

The former case had average string sizes of about 2500 characters, while this new file had an average string size of over 4K for one column, with some strings larger than 20,000 characters.

I am attaching a parquet file with one column of semi-random strings that mimics the shape of the main column in the customer file.  These are the sizes of the original uncompressed file and the cpu/gpu compressed versions:
```
jimb@jimb-ldt:/opt/data/jimb$ du -sh test-var*
71M	test-variable-strings
1.1M	test-variable-strings-cpu-zstd
71M	test-variable-strings-gpu-zstd
```

**Expected behavior**
We should expect the GPU zstd compressed file to be comparable in size to the cpu zstd compressed file.

I tested this using spark shell on an unbuntu system with an RTX4000 GPU.  I used this spark rapids jar:
rapids-4-spark_2.12-23.02.0-20230122.152330-55-cuda11.jar

A tar archive with the uncompressed and cpu/gpu compressed versions of the test parquet file.
[col-with-large-variable-strings.tgz](https://github.com/rapidsai/cudf/files/10503385/col-with-large-variable-strings.tgz)

etseidl:
## Description
Fixes #12613

This PR adds the ability for columns to have different fragment sizes.  This allows a large fragment size for narrow columns, but allows for finer grained fragments for very wide columns.  This change should make wide columns fit (approximately) within page size constraints, and should help with compressors that rely on pages being under a certain threshold (i.e. Zstandard).

## Checklist


etseidl:
This PR retains the performance gains seen in  #12627 for LIST benchmarks.

23.04:
```
## parquet_write_encode

### [0] NVIDIA RTX A6000

| data_type | cardinality | run_length | Samples |  CPU Time  | Noise |  GPU Time  | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|-----------|-------------|------------|---------|------------|-------|------------|-------|------------------|-------------------|-------------------|
|  INTEGRAL |           0 |          1 |      5x | 111.447 ms | 0.07% | 111.442 ms | 0.07% |       4817500960 |         2.146 GiB |       498.123 MiB |
|  INTEGRAL |        1000 |          1 |    332x |  45.153 ms | 1.95% |  45.148 ms | 1.95% |      11891471139 |         2.770 GiB |       161.438 MiB |
|  INTEGRAL |           0 |         32 |    154x |  35.018 ms | 0.50% |  35.013 ms | 0.50% |      15333351555 |         2.770 GiB |        27.720 MiB |
|  INTEGRAL |        1000 |         32 |     17x |  30.016 ms | 0.12% |  30.011 ms | 0.12% |      17889329183 |         2.770 GiB |        14.403 MiB |
|    STRING |           0 |          1 |      5x | 130.701 ms | 0.19% | 130.696 ms | 0.19% |       4107797930 |         1.342 GiB |       597.486 MiB |
|    STRING |        1000 |          1 |    528x |  26.473 ms | 0.98% |  26.467 ms | 0.98% |      20284442095 |       677.964 MiB |        46.473 MiB |
|    STRING |           0 |         32 |      5x | 130.522 ms | 0.10% | 130.517 ms | 0.10% |       4113429866 |         1.342 GiB |       597.486 MiB |
|    STRING |        1000 |         32 |     35x |  14.425 ms | 0.42% |  14.420 ms | 0.42% |      37232061752 |       677.964 MiB |         8.504 MiB |
|      LIST |           0 |          1 |      5x | 534.714 ms | 0.03% | 534.707 ms | 0.03% |       1004046830 |         1.602 GiB |       498.003 MiB |
|      LIST |        1000 |          1 |      5x | 355.770 ms | 0.06% | 355.763 ms | 0.06% |       1509067048 |         2.752 GiB |       166.640 MiB |
|      LIST |           0 |         32 |      5x | 270.287 ms | 0.04% | 270.281 ms | 0.04% |       1986343787 |         2.752 GiB |        37.257 MiB |
|      LIST |        1000 |         32 |      5x | 274.208 ms | 0.07% | 274.202 ms | 0.07% |       1957941480 |         2.752 GiB |        24.422 MiB |
|    STRUCT |           0 |          1 |      5x | 135.583 ms | 0.21% | 135.578 ms | 0.21% |       3959881290 |         1.283 GiB |       569.525 MiB |
|    STRUCT |        1000 |          1 |     56x |  40.450 ms | 0.50% |  40.444 ms | 0.50% |      13274383477 |         1.324 GiB |        90.699 MiB |
|    STRUCT |           0 |         32 |      5x | 108.386 ms | 0.44% | 108.380 ms | 0.44% |       4953599367 |         1.473 GiB |       409.317 MiB |
|    STRUCT |        1000 |         32 |     19x |  27.252 ms | 0.48% |  27.247 ms | 0.48% |      19703957784 |         1.324 GiB |        15.400 MiB |
```
This PR:
```
| data_type | cardinality | run_length | Samples |  CPU Time  | Noise |  GPU Time  | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|-----------|-------------|------------|---------|------------|-------|------------|-------|------------------|-------------------|-------------------|
|  INTEGRAL |           0 |          1 |      5x | 111.031 ms | 0.09% | 111.027 ms | 0.09% |       4835508962 |         2.143 GiB |       498.123 MiB |
|  INTEGRAL |        1000 |          1 |     12x |  44.471 ms | 0.25% |  44.466 ms | 0.25% |      12073838869 |         2.768 GiB |       161.438 MiB |
|  INTEGRAL |           0 |         32 |     15x |  34.655 ms | 0.36% |  34.650 ms | 0.36% |      15494046559 |         2.768 GiB |        27.720 MiB |
|  INTEGRAL |        1000 |         32 |     17x |  29.759 ms | 0.23% |  29.754 ms | 0.22% |      18043576900 |         2.768 GiB |        14.403 MiB |
|    STRING |           0 |          1 |      5x | 129.735 ms | 0.15% | 129.731 ms | 0.15% |       4138348486 |         1.342 GiB |       597.486 MiB |
|    STRING |        1000 |          1 |    528x |  26.598 ms | 0.85% |  26.592 ms | 0.85% |      20188830643 |       677.452 MiB |        46.473 MiB |
|    STRING |           0 |         32 |      5x | 129.458 ms | 0.19% | 129.453 ms | 0.19% |       4147236475 |         1.342 GiB |       597.486 MiB |
|    STRING |        1000 |         32 |     34x |  14.780 ms | 0.37% |  14.775 ms | 0.37% |      36337166491 |       677.452 MiB |         8.505 MiB |
|      LIST |           0 |          1 |      5x | 262.710 ms | 0.20% | 262.704 ms | 0.20% |       2043631541 |         1.602 GiB |       498.620 MiB |
|      LIST |        1000 |          1 |      5x | 137.624 ms | 0.35% | 137.619 ms | 0.35% |       3901142053 |         2.752 GiB |       167.154 MiB |
|      LIST |           0 |         32 |      5x | 113.366 ms | 0.20% | 113.361 ms | 0.20% |       4735919435 |         2.752 GiB |        37.597 MiB |
|      LIST |        1000 |         32 |      5x | 109.960 ms | 0.21% | 109.953 ms | 0.21% |       4882723938 |         2.752 GiB |        24.811 MiB |
|    STRUCT |           0 |          1 |      5x | 135.770 ms | 0.25% | 135.765 ms | 0.25% |       3954415646 |         1.283 GiB |       569.525 MiB |
|    STRUCT |        1000 |          1 |     13x |  40.804 ms | 0.37% |  40.799 ms | 0.37% |      13158931041 |         1.323 GiB |        90.699 MiB |
|    STRUCT |           0 |         32 |      5x | 108.203 ms | 0.19% | 108.198 ms | 0.19% |       4961947254 |         1.473 GiB |       409.317 MiB |
|    STRUCT |        1000 |         32 |     34x |  27.772 ms | 0.50% |  27.767 ms | 0.50% |      19334866180 |         1.323 GiB |        15.399 MiB |
```

etseidl:
For parquet_write_io_compression
23.04:
```
|    io    | compression | cardinality | run_length | Samples | CPU Time | Noise | GPU Time | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|----------|-------------|-------------|------------|---------|----------|-------|----------|-------|------------------|-------------------|-------------------|
| FILEPATH |      SNAPPY |           0 |          1 |      5x |  3.468 s | 0.42% |  3.468 s | 0.42% |        154828664 |         1.556 GiB |       493.950 MiB |
| FILEPATH |      SNAPPY |        1000 |          1 |      5x |  2.000 s | 0.44% |  2.000 s | 0.44% |        268476241 |         2.536 GiB |       161.238 MiB |
| FILEPATH |      SNAPPY |           0 |         32 |      5x |  1.603 s | 0.21% |  1.603 s | 0.21% |        334929961 |         2.532 GiB |        49.703 MiB |
| FILEPATH |      SNAPPY |        1000 |         32 |      5x |  1.600 s | 0.20% |  1.600 s | 0.20% |        335544238 |         2.536 GiB |        23.415 MiB |
```
This PR:
```
|    io    | compression | cardinality | run_length | Samples | CPU Time | Noise | GPU Time | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|----------|-------------|-------------|------------|---------|----------|-------|----------|-------|------------------|-------------------|-------------------|
| FILEPATH |      SNAPPY |           0 |          1 |      5x |  3.460 s | 0.55% |  3.460 s | 0.55% |        155186150 |         1.556 GiB |       494.475 MiB |
| FILEPATH |      SNAPPY |        1000 |          1 |      5x |  2.023 s | 0.43% |  2.023 s | 0.43% |        265387308 |         2.536 GiB |       161.668 MiB |
| FILEPATH |      SNAPPY |           0 |         32 |      5x |  1.620 s | 0.13% |  1.620 s | 0.13% |        331384852 |         2.532 GiB |        50.055 MiB |
| FILEPATH |      SNAPPY |        1000 |         32 |      5x |  1.613 s | 0.13% |  1.613 s | 0.13% |        332883310 |         2.536 GiB |        23.734 MiB |
```

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -85,6 +87,31 @@ parquet::Compression to_parquet_compression(compression_type compression)
   }
 }
 
+size_type column_size(column_view const& column, rmm::cuda_stream_view stream)
+{
+  if (column.num_children() == 0) {
+    return is_fixed_width(column.type()) ? size_of(column.type()) * column.size() : 0;
+  }
+
```
etseidl, line 95: Since this is passed leaf columns, do I need to check for STRUCT at all?  And does this work for binary columns (list[uint_8] right?)?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -897,6 +924,15 @@ void writer::impl::init_page_fragments(cudf::detail::hostdevice_2dvector<gpu::Pa
   frag.device_to_host(stream, true);
 }
 
+void writer::impl::init_page_fragments_1d(hostdevice_vector<gpu::PageFragment>& frag,
+                                          host_span<size_type const> frag_sizes)
+{
+  auto d_frag_sz = cudf::detail::make_device_uvector_sync(frag_sizes, stream);
```
etseidl, line 930: can this be asynch?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1407,17 +1454,29 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+  // for each leaf column.  Skip if the fragment size is not the default.
+  std::vector<size_type> column_frag_size(single_streams_table.num_columns(),
+                                          max_page_fragment_size_);
+
+  auto iter = thrust::make_counting_iterator<size_type>(0);
+  if (table.num_rows() > 0 && max_page_fragment_size_ == cudf::io::default_max_page_fragment_size) {
+    std::for_each(iter, iter + single_streams_table.num_columns(), [&](size_type index) {
+      auto const avg_len = std::max<size_type>(
+        column_size(single_streams_table.column(index), stream) / table.num_rows(), 1);
+      auto const frag_size    = max_page_size_bytes / avg_len;
```
etseidl, line 1468: rather than max_page_size/avg, should this shoot for multiple fragments per page to catch at least some overshooting?

etseidl:
LIST write benchmark got even better with the 4 fragments/page change, but this is due to allowing more parallelism in page encoding and compression due to simply adding more pages of smaller size.  Kind of a cheat :sweat_smile: 
```
| data_type | cardinality | run_length | Samples |  CPU Time  | Noise |  GPU Time  | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|-----------|-------------|------------|---------|------------|-------|------------|-------|------------------|-------------------|-------------------|
|      LIST |           0 |          1 |      5x | 255.131 ms | 0.35% | 255.126 ms | 0.35% |       2104338618 |         1.602 GiB |       498.625 MiB |
|      LIST |        1000 |          1 |      5x | 124.339 ms | 0.15% | 124.334 ms | 0.15% |       4317982003 |         2.752 GiB |       167.056 MiB |
|      LIST |           0 |         32 |      5x | 101.464 ms | 0.44% | 101.459 ms | 0.44% |       5291510044 |         2.752 GiB |        37.520 MiB |
|      LIST |        1000 |         32 |    150x | 100.153 ms | 1.63% | 100.148 ms | 1.63% |       5360782799 |         2.752 GiB |        24.722 MiB |
```

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1407,17 +1453,60 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+      max_page_fragment_size_ = std::min<size_type>(rg_frag_size, max_page_fragment_size_);
+    }
+
+    // dividing page size by average row length will tend to overshoot the desired
+    // page size when there's high variability in the row lengths. instead, shoot
+    // for multiple fragments per page to smooth things out. using 2 was too
+    // unbalanced in final page sizes, so using 4 which seems to be a good
+    // compromise at smoothing things out without getting fragment sizes too small.
+    auto frag_size_fn = [&](auto const& col, size_type col_size) {
+      const int target_frags_per_page = is_fixed_width(col.type()) ? 1 : 4;
```
vuule, line 1485: how do structs fit in here? a struct containing only fixed-width fields is effectively a fixed-width type.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1407,17 +1453,60 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+    };
+
+    std::transform(single_streams_table.begin(),
+                   single_streams_table.end(),
+                   column_sizes.begin(),
+                   std::back_inserter(column_frag_size),
+                   frag_size_fn);
   }
 
+  bool const has_non_uniform_fragments =
```
vuule, line 1503: how inefficient would it be to assume that we always have non uniform fragments?

vuule (commented):
First pass, couple of comments so far.
Completely forgot that statistics are computed on the fragments, this makes the PR more complicated than I expected.

cpp/src/io/parquet/writer_impl.cu
etseidl: True...I'll add a function that tests for that.  Fixed length byte arrays too when they're supported.

cpp/src/io/parquet/writer_impl.cu
etseidl: Let me think on this.  Now that the frag stats are computed after the chunks, I think we can get away with a single version of `gather_fragment_stats`.  Then it's just the fragments to deal with. I'll see if there's a better way to deal with that.

etseidl: I reworked this to detect the need to resize the fragments in a different way.  Was able to get rid of one of the gather statistics paths as a bonus.  Thanks for the nudge @vuule !

nvdbaranec:
As an aside, I was curious if you had a reliable way to get the current version of cudf to produce page-spanning list rows.  I have found it very hard to do when trying to build test cases.

etseidl:
> As an aside, I was curious if you had a reliable way to get the current version of cudf to produce page-spanning list rows. I have found it very hard to do when trying to build test cases.

I honestly don't think it's possible for cudf to do this, since page fragments are calculated based on row numbers, not values.  Since the addition of column indexes to the parquet spec, I don't think page-spanning rows are allowed.  The arrow-rust team recently had to grapple with this (https://github.com/apache/arrow-rs/issues/1777#issuecomment-1147686956),  Not sure where arrow-cpp is at, they may still have this behavior, which I think would mean pyarrow would as well.

vuule:
> LIST write benchmark got even better with the 4 fragments/page change, but this is due to allowing more parallelism in page encoding and compression due to simply adding more pages of smaller size. Kind of a cheat 😅
> 
> ```
> | data_type | cardinality | run_length | Samples |  CPU Time  | Noise |  GPU Time  | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
> |-----------|-------------|------------|---------|------------|-------|------------|-------|------------------|-------------------|-------------------|
> |      LIST |           0 |          1 |      5x | 255.131 ms | 0.35% | 255.126 ms | 0.35% |       2104338618 |         1.602 GiB |       498.625 MiB |
> |      LIST |        1000 |          1 |      5x | 124.339 ms | 0.15% | 124.334 ms | 0.15% |       4317982003 |         2.752 GiB |       167.056 MiB |
> |      LIST |           0 |         32 |      5x | 101.464 ms | 0.44% | 101.459 ms | 0.44% |       5291510044 |         2.752 GiB |        37.520 MiB |
> |      LIST |        1000 |         32 |    150x | 100.153 ms | 1.63% | 100.148 ms | 1.63% |       5360782799 |         2.752 GiB |        24.722 MiB |
> ```

Do you think it's worthwhile to check the performance with short lists?
Modifying the upper bound for the list length at
https://github.com/rapidsai/cudf/blob/branch-23.04/cpp/benchmarks/common/generate_input.hpp#L226-L227
from 100 to ~5 should do it.

etseidl:
> Do you think it's worthwhile to check the performance with short lists? Modifying the upper bound for the list length at https://github.com/rapidsai/cudf/blob/branch-23.04/cpp/benchmarks/common/generate_input.hpp#L226-L227 from 100 to ~5 should do it.

probably worth doing, but it probably won't give a 2-3x improvement :cry:

etseidl:
Yep, with smaller lists the improvement goes away :(
```
23.04
| data_type | cardinality | run_length | Samples |  CPU Time  | Noise |  GPU Time  | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|-----------|-------------|------------|---------|------------|-------|------------|-------|------------------|-------------------|-------------------|
|      LIST |           0 |          1 |     30x | 506.779 ms | 2.03% | 506.773 ms | 2.03% |       1059390375 |         4.255 GiB |       605.319 MiB |
|      LIST |        1000 |          1 |     38x | 398.524 ms | 0.57% | 398.518 ms | 0.57% |       1347168342 |         5.406 GiB |       333.790 MiB |
|      LIST |           0 |         32 |      5x | 315.475 ms | 0.23% | 315.470 ms | 0.23% |       1701814594 |         5.406 GiB |       147.600 MiB |
|      LIST |        1000 |         32 |      5x | 320.947 ms | 0.46% | 320.941 ms | 0.46% |       1672801574 |         5.406 GiB |       137.422 MiB |

this PR:
| data_type | cardinality | run_length | Samples |  CPU Time  | Noise |  GPU Time  | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|-----------|-------------|------------|---------|------------|-------|------------|-------|------------------|-------------------|-------------------|
|      LIST |           0 |          1 |     30x | 507.881 ms | 2.16% | 507.875 ms | 2.16% |       1057093441 |         4.253 GiB |       605.319 MiB |
|      LIST |        1000 |          1 |     37x | 409.769 ms | 2.92% | 409.762 ms | 2.92% |       1310200449 |         5.404 GiB |       333.790 MiB |
|      LIST |           0 |         32 |      5x | 318.574 ms | 0.28% | 318.569 ms | 0.28% |       1685259863 |         5.404 GiB |       147.600 MiB |
|      LIST |        1000 |         32 |     47x | 323.541 ms | 2.04% | 323.535 ms | 2.04% |       1659391773 |         5.404 GiB |       137.423 MiB |
```

vuule:
> But then speed wasn't the point, it's getting the page sizes right

If you ask me, with 2-3x speedup, speed becomes A point, at least :D

It's fine not to get any performance improvement here, I suggested this only to check if the 4 fragments/page "cheat" can reduce performance in the case of tiny lists. Looks like the perf is minimally impacted 👍

etseidl:
Another random aside...the V2 page header also has an `is_compressed` field.  Would be nice to have this so a single huge page won't result in no compression for the entire column chunk.

ttnghia:
Can you run the comparison script to show better view, please:
```
python nvbench/scripts/nvbench_compare.py ./before.json ./after.json
```
The script is here: https://github.com/NVIDIA/nvbench/blob/main/scripts/nvbench_compare.py

etseidl:
```# parquet_write_encode

## [0] NVIDIA RTX A6000

|  data_type  |  cardinality  |  run_length  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |           Diff |   %Diff |  Status  |
|-------------|---------------|--------------|------------|-------------|------------|-------------|----------------|---------|----------|
|  INTEGRAL   |       0       |      1       | 111.209 ms |       0.48% | 111.363 ms |       0.36% |     153.916 us |   0.14% |   PASS   |
|  INTEGRAL   |     1000      |      1       |  44.408 ms |       0.23% |  44.543 ms |       0.20% |     135.160 us |   0.30% |   FAIL   |
|  INTEGRAL   |       0       |      32      |  34.464 ms |       0.16% |  34.671 ms |       0.18% |     206.748 us |   0.60% |   FAIL   |
|  INTEGRAL   |     1000      |      32      |  29.626 ms |       0.23% |  29.786 ms |       0.13% |     159.562 us |   0.54% |   FAIL   |
|    FLOAT    |       0       |      1       | 106.309 ms |       0.07% | 106.367 ms |       0.07% |      57.765 us |   0.05% |   PASS   |
|    FLOAT    |     1000      |      1       |  30.554 ms |       0.36% |  30.571 ms |       0.30% |      17.543 us |   0.06% |   PASS   |
|    FLOAT    |       0       |      32      |  23.818 ms |       0.15% |  23.845 ms |       0.24% |      26.909 us |   0.11% |   PASS   |
|    FLOAT    |     1000      |      32      |  19.749 ms |       0.29% |  19.755 ms |       0.21% |       6.234 us |   0.03% |   PASS   |
|   DECIMAL   |       0       |      1       |  51.548 ms |       0.06% |  51.622 ms |       0.21% |      74.380 us |   0.14% |   FAIL   |
|   DECIMAL   |     1000      |      1       |  18.146 ms |       0.22% |  18.247 ms |       0.32% |     100.741 us |   0.56% |   FAIL   |
|   DECIMAL   |       0       |      32      |  15.474 ms |       0.27% |  15.547 ms |       0.34% |      73.450 us |   0.47% |   FAIL   |
|   DECIMAL   |     1000      |      32      |  13.515 ms |       0.34% |  13.569 ms |       0.45% |      53.684 us |   0.40% |   FAIL   |
|  TIMESTAMP  |       0       |      1       | 130.041 ms |       0.34% | 129.876 ms |       0.23% |    -165.021 us |  -0.13% |   PASS   |
|  TIMESTAMP  |     1000      |      1       |  26.185 ms |       0.44% |  26.077 ms |       0.19% |    -108.157 us |  -0.41% |   FAIL   |
|  TIMESTAMP  |       0       |      32      |  26.938 ms |       0.50% |  26.927 ms |       0.47% |     -10.823 us |  -0.04% |   PASS   |
|  TIMESTAMP  |     1000      |      32      |  17.640 ms |       0.42% |  17.615 ms |       0.50% |     -24.796 us |  -0.14% |   PASS   |
|  DURATION   |       0       |      1       |  79.801 ms |       0.16% |  79.582 ms |       0.07% |    -218.862 us |  -0.27% |   FAIL   |
|  DURATION   |     1000      |      1       |  25.569 ms |       0.16% |  25.426 ms |       0.16% |    -143.905 us |  -0.56% |   FAIL   |
|  DURATION   |       0       |      32      |  21.139 ms |       0.27% |  21.036 ms |       0.27% |    -102.816 us |  -0.49% |   FAIL   |
|  DURATION   |     1000      |      32      |  16.670 ms |       0.24% |  16.574 ms |       0.28% |     -95.921 us |  -0.58% |   FAIL   |
|   STRING    |       0       |      1       | 129.449 ms |       0.11% | 129.199 ms |       0.10% |    -250.488 us |  -0.19% |   FAIL   |
|   STRING    |     1000      |      1       |  25.977 ms |       0.84% |  26.601 ms |       0.85% |     624.246 us |   2.40% |   FAIL   |
|   STRING    |       0       |      32      | 129.283 ms |       0.17% | 129.763 ms |       0.12% |     480.383 us |   0.37% |   FAIL   |
|   STRING    |     1000      |      32      |  14.212 ms |       0.26% |  14.815 ms |       0.32% |     602.568 us |   4.24% |   FAIL   |
|    LIST     |       0       |      1       | 528.298 ms |       0.08% | 252.878 ms |       0.27% | -275420.004 us | -52.13% |   FAIL   |
|    LIST     |     1000      |      1       | 350.126 ms |       0.03% | 150.339 ms |       0.26% | -199787.198 us | -57.06% |   FAIL   |
|    LIST     |       0       |      32      | 266.526 ms |       0.07% |  93.177 ms |       0.11% | -173348.794 us | -65.04% |   FAIL   |
|    LIST     |     1000      |      32      | 270.225 ms |       0.06% |  91.216 ms |       0.41% | -179009.469 us | -66.24% |   FAIL   |
|   STRUCT    |       0       |      1       | 134.896 ms |       0.23% | 135.970 ms |       0.24% |       1.074 ms |   0.80% |   FAIL   |
|   STRUCT    |     1000      |      1       |  40.384 ms |       0.31% |  40.962 ms |       0.49% |     578.393 us |   1.43% |   FAIL   |
|   STRUCT    |       0       |      32      | 107.284 ms |       0.13% | 108.453 ms |       0.26% |       1.169 ms |   1.09% |   FAIL   |
|   STRUCT    |     1000      |      32      |  26.737 ms |       0.24% |  27.732 ms |       0.19% |     995.065 us |   3.72% |   FAIL   |

# parquet_write_io_compression

## [0] NVIDIA RTX A6000

|     io      |  compression  |  cardinality  |  run_length  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |          Diff |   %Diff |  Status  |
|-------------|---------------|---------------|--------------|------------|-------------|------------|-------------|---------------|---------|----------|
|  FILEPATH   |    SNAPPY     |       0       |      1       |    3.474 s |       0.34% |    3.447 s |       0.30% | -26989.307 us |  -0.78% |   FAIL   |
|  FILEPATH   |    SNAPPY     |     1000      |      1       |    2.005 s |       0.21% |    2.015 s |       0.41% |     10.296 ms |   0.51% |   FAIL   |
|  FILEPATH   |    SNAPPY     |       0       |      32      |    1.597 s |       0.39% |    1.612 s |       0.39% |     14.564 ms |   0.91% |   FAIL   |
|  FILEPATH   |    SNAPPY     |     1000      |      32      |    1.603 s |       0.32% |    1.599 s |       0.33% |  -4459.399 us |  -0.28% |   PASS   |
|  FILEPATH   |     NONE      |       0       |      1       |    1.767 s |       0.80% |    1.778 s |       0.78% |     10.391 ms |   0.59% |   PASS   |
|  FILEPATH   |     NONE      |     1000      |      1       |    1.271 s |       0.21% |    1.284 s |       0.49% |     12.541 ms |   0.99% |   FAIL   |
|  FILEPATH   |     NONE      |       0       |      32      |    1.370 s |       0.24% |    1.381 s |       0.25% |     10.557 ms |   0.77% |   FAIL   |
|  FILEPATH   |     NONE      |     1000      |      32      |    1.321 s |       0.42% |    1.325 s |       0.27% |      4.258 ms |   0.32% |   FAIL   |
| HOST_BUFFER |    SNAPPY     |       0       |      1       |    3.214 s |       0.32% |    3.196 s |       0.37% | -17981.348 us |  -0.56% |   FAIL   |
| HOST_BUFFER |    SNAPPY     |     1000      |      1       |    1.719 s |       0.27% |    1.732 s |       0.42% |     13.293 ms |   0.77% |   FAIL   |
| HOST_BUFFER |    SNAPPY     |       0       |      32      |    1.322 s |       0.45% |    1.330 s |       0.24% |      8.060 ms |   0.61% |   FAIL   |
| HOST_BUFFER |    SNAPPY     |     1000      |      32      |    1.315 s |       0.37% |    1.313 s |       0.13% |  -2190.063 us |  -0.17% |   FAIL   |
| HOST_BUFFER |     NONE      |       0       |      1       |    1.516 s |       0.53% |    1.521 s |       0.90% |      4.908 ms |   0.32% |   PASS   |
| HOST_BUFFER |     NONE      |     1000      |      1       |    1.008 s |       0.89% |    1.010 s |       0.49% |      2.035 ms |   0.20% |   PASS   |
| HOST_BUFFER |     NONE      |       0       |      32      |    1.094 s |       0.28% |    1.100 s |       0.46% |      6.219 ms |   0.57% |   FAIL   |
| HOST_BUFFER |     NONE      |     1000      |      32      |    1.037 s |       0.31% |    1.041 s |       0.50% |      3.960 ms |   0.38% |   FAIL   |
|    VOID     |    SNAPPY     |       0       |      1       |    3.107 s |       0.13% |    3.049 s |       0.61% | -57513.770 us |  -1.85% |   FAIL   |
|    VOID     |    SNAPPY     |     1000      |      1       |    1.691 s |       0.29% |    1.684 s |       0.22% |  -6339.966 us |  -0.37% |   FAIL   |
|    VOID     |    SNAPPY     |       0       |      32      |    1.312 s |       0.31% |    1.317 s |       0.33% |      5.475 ms |   0.42% |   FAIL   |
|    VOID     |    SNAPPY     |     1000      |      32      |    1.320 s |       0.41% |    1.309 s |       0.17% | -11235.376 us |  -0.85% |   FAIL   |
|    VOID     |     NONE      |       0       |      1       |    1.390 s |       0.67% |    1.393 s |       0.72% |      3.116 ms |   0.22% |   PASS   |
|    VOID     |     NONE      |     1000      |      1       | 955.064 ms |       0.45% | 955.646 ms |       0.49% |    582.048 us |   0.06% |   PASS   |
|    VOID     |     NONE      |       0       |      32      |    1.079 s |       0.37% |    1.084 s |       0.26% |      4.965 ms |   0.46% |   FAIL   |
|    VOID     |     NONE      |     1000      |      32      |    1.036 s |       0.48% |    1.033 s |       0.36% |  -3008.004 us |  -0.29% |   PASS   |

# parquet_write_options

## [0] NVIDIA RTX A6000

|     statistics      |  compression  |      file_path      |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |           Diff |   %Diff |  Status  |
|---------------------|---------------|---------------------|------------|-------------|------------|-------------|----------------|---------|----------|
|   STATISTICS_NONE   |    SNAPPY     | unused_path.parquet |    6.368 s |        inf% |    6.131 s |        inf% | -237494.954 us |  -3.73% |   PASS   |
|   STATISTICS_NONE   |    SNAPPY     |                     |    6.368 s |        inf% |    6.118 s |        inf% | -249891.276 us |  -3.92% |   PASS   |
|   STATISTICS_NONE   |     NONE      | unused_path.parquet |    4.503 s |        inf% |    4.352 s |        inf% | -151162.598 us |  -3.36% |   PASS   |
|   STATISTICS_NONE   |     NONE      |                     |    4.504 s |        inf% |    4.348 s |        inf% | -155216.675 us |  -3.45% |   PASS   |
| STATISTICS_ROWGROUP |    SNAPPY     | unused_path.parquet |    6.445 s |        inf% |    6.206 s |        inf% | -238667.969 us |  -3.70% |   PASS   |
| STATISTICS_ROWGROUP |    SNAPPY     |                     |    6.448 s |        inf% |    6.197 s |        inf% | -251497.396 us |  -3.90% |   PASS   |
| STATISTICS_ROWGROUP |     NONE      | unused_path.parquet |    4.583 s |        inf% |    4.421 s |        inf% | -161834.473 us |  -3.53% |   PASS   |
| STATISTICS_ROWGROUP |     NONE      |                     |    4.579 s |        inf% |    4.432 s |        inf% | -147024.536 us |  -3.21% |   PASS   |
|  STATISTICS_COLUMN  |    SNAPPY     | unused_path.parquet |    6.451 s |        inf% |    6.201 s |        inf% | -249066.895 us |  -3.86% |   PASS   |
|  STATISTICS_COLUMN  |    SNAPPY     |                     |    6.446 s |        inf% |    6.212 s |        inf% | -233949.870 us |  -3.63% |   PASS   |
|  STATISTICS_COLUMN  |     NONE      | unused_path.parquet |    4.584 s |        inf% |    4.426 s |        inf% | -157416.260 us |  -3.43% |   PASS   |
|  STATISTICS_COLUMN  |     NONE      |                     |    4.583 s |        inf% |    4.423 s |        inf% | -160635.864 us |  -3.50% |   PASS   |
|   STATISTICS_PAGE   |    SNAPPY     | unused_path.parquet |    6.444 s |        inf% |    6.190 s |        inf% | -253174.316 us |  -3.93% |   PASS   |
|   STATISTICS_PAGE   |    SNAPPY     |                     |    6.453 s |        inf% |    6.198 s |        inf% | -254405.436 us |  -3.94% |   PASS   |
|   STATISTICS_PAGE   |     NONE      | unused_path.parquet |    4.579 s |        inf% |    4.432 s |        inf% | -147026.367 us |  -3.21% |   PASS   |
|   STATISTICS_PAGE   |     NONE      |                     |    4.580 s |        inf% |    4.424 s |        inf% | -155563.599 us |  -3.40% |   PASS   |

# parquet_write_num_cols

## [0] NVIDIA RTX A6000

|  num_cols  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |       Diff |   %Diff |  Status  |
|------------|------------|-------------|------------|-------------|------------|---------|----------|
|     8      |  59.836 ms |       0.13% |  59.789 ms |       0.16% | -46.833 us |  -0.08% |   PASS   |
|    1024    | 111.604 ms |       3.88% | 130.776 ms |       0.11% |  19.172 ms |  17.18% |   FAIL   |

# parquet_chunked_write

## [0] NVIDIA RTX A6000

|  num_cols  |  num_chunks  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |         Diff |   %Diff |  Status  |
|------------|--------------|------------|-------------|------------|-------------|--------------|---------|----------|
|     8      |      8       |  87.663 ms |       0.09% |  87.580 ms |       0.18% |   -83.047 us |  -0.09% |   FAIL   |
|    1024    |      8       | 166.850 ms |       0.11% | 174.213 ms |       0.48% |     7.363 ms |   4.41% |   FAIL   |
|     8      |      64      | 307.274 ms |       0.02% | 306.239 ms |       0.10% | -1035.132 us |  -0.34% |   FAIL   |
|    1024    |      64      | 607.793 ms |       1.78% | 615.928 ms |       0.24% |     8.135 ms |   1.34% |   FAIL   |

# Summary

- Total Matches: 78
  - Pass    (diff <= min_noise): 32
  - Unknown (infinite noise):    0
  - Failure (diff > min_noise):  46
```

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1551,6 +1623,14 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
         column_chunk_meta.codec          = UNCOMPRESSED;
         column_chunk_meta.num_values     = ck.num_values;
+
+        if (column_frag_size[c] < max_page_fragment_size_) {
+          auto const new_frags_in_chunk =
+            util::div_rounding_up_unsafe(row_group.num_rows, column_frag_size[c]);
+          frags_per_column[c] += new_frags_in_chunk;
+        } else {
+          frags_per_column[c] += fragments_in_chunk;
+        }
```
vuule, line 1633: For your consideration:
```suggestion
        frags_per_column[c] += util::div_rounding_up_unsafe(row_group.num_rows, std::min(column_frag_size[c], max_page_fragment_size_));
```

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1572,7 +1652,80 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+
+  if (frags_per_column.size() > 0) {
+    std::exclusive_scan(frags_per_column.data(),
+                        frags_per_column.data() + num_columns + 1,
+                        std::back_inserter(frag_offsets),
+                        0);
+    total_frags = frag_offsets[num_columns];
+  }
+
+  bool const is_resize_fragments = total_frags != num_columns * num_fragments;
```
vuule, line 1672: I would be okay with the assumption that this is always true. Seems like a low overhead to help avoid some of the complexity below.

vuule (changes requested):
Finally got to the point where I understand the PR well enough to review it.

Looks good overall, so it's time to pick some nits :D

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1407,17 +1451,54 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+    };
+
+    std::transform(single_streams_table.begin(),
+                   single_streams_table.end(),
+                   column_sizes.begin(),
+                   column_frag_size.begin(),
+                   frag_size_fn);
   }
 
+  // do first pass with max_page_fragment_size_ to calculate row group boundaries
```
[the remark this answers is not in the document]
etseidl: Because at this point it should be either the default 5000, a scaled down value if the rows are very wide, or a user supplied value (which is then always used).

cpp/src/io/parquet/page_enc.cu
```diff
@@ -126,6 +126,82 @@ constexpr size_t underflow_safe_subtract(size_t a, size_t b)
...
+  auto const leaf_type       = s->col.leaf_column->type().id();
+  auto const dtype_len       = physical_type_len(physical_type, leaf_type);
+  auto const nvals           = s->frag.num_leaf_values;
+  auto const start_value_idx = s->frag.start_value_idx;
+
+  for (uint32_t i = 0; i < nvals; i += block_size) {
+    auto const val_idx  = start_value_idx + i + t;
+    auto const is_valid = (i + t < nvals && val_idx < s->col.leaf_column->size())
+                            ? s->col.leaf_column->is_valid(val_idx)
+                            : false;
```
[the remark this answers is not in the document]
etseidl: Ah, leftover code.  I had seen that but forgot to change it.  Nice catch!

cpp/src/io/parquet/writer_impl.hpp
```diff
@@ -136,18 +136,24 @@ class writer::impl {
                            device_span<int const> part_frag_offset,
                            uint32_t fragment_size);
 
+  /**
+   * @brief Gather page fragments for columns with differing fragment sizes
+   * 
+   * @param frag Destination page fragments
+   * @param frag_sizes Array of fragment sizes for each column
+  */
+  void init_page_fragments_1d(hostdevice_vector<gpu::PageFragment>& frag,
```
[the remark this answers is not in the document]
etseidl: I'm never very imaginative when I come up with names.  I'll think on it.  Maybe recalculate_page_fragments or something like that?

cpp/src/io/parquet/writer_impl.cu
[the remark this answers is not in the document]
vuule: To be more precise with my question - why don't we use the per-column fragment sizes here?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -897,23 +942,22 @@ void writer::impl::init_page_fragments(cudf::detail::hostdevice_2dvector<gpu::Pa
   frag.device_to_host(stream, true);
 }
 
-void writer::impl::gather_fragment_statistics(
-  device_2dspan<statistics_chunk> frag_stats_chunk,
-  device_2dspan<gpu::PageFragment const> frag,
-  device_span<gpu::parquet_column_device_view const> col_desc,
-  uint32_t num_fragments)
+void writer::impl::init_page_fragments_1d(hostdevice_vector<gpu::PageFragment>& frag,
```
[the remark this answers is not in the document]
etseidl: Or should I move the `host_to_device()` call here and keep the signature?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -897,23 +942,22 @@ void writer::impl::init_page_fragments(cudf::detail::hostdevice_2dvector<gpu::Pa
...
-
-  gpu::InitFragmentStatistics(frag_stats_group_2dview, frag, col_desc, stream);
-  detail::calculate_group_statistics<detail::io_file_format::PARQUET>(frag_stats_chunk.data(),
-                                                                      frag_stats_group.data(),
-                                                                      num_fragments * num_columns,
-                                                                      stream,
-                                                                      int96_timestamps);
+  auto d_frag_sz = cudf::detail::make_device_uvector_async(frag_sizes, stream);
+  InitPageFragments1D(frag, d_frag_sz, stream);
+  frag.device_to_host(stream, true);
```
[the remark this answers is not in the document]
etseidl: Yes, this one is unnecessary.  Will smite it.

cpp/src/io/parquet/writer_impl.cu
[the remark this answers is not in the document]
vuule: I'm leaning towards my version, it makes `init_page_fragments_1d` more flexible/reusable. But, since that does not really matter in this case, I don't have a strong opinion.

cpp/include/cudf/io/parquet.hpp
```diff
@@ -1076,7 +1076,7 @@ class chunked_parquet_writer_options {
   // Maximum size of column chunk dictionary (in bytes)
   size_t _max_dictionary_size = default_max_dictionary_size;
   // Maximum number of rows in a page fragment
-  size_type _max_page_fragment_size = default_max_page_fragment_size;
+  size_type _max_page_fragment_size = -1;
```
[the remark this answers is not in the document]
etseidl: Sigh.  And I've been doing scala programming how long and still don't use optionals much?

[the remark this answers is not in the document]
vuule: I'm gonna guess it's because you've been programming in C/C++ even longer :D

cpp/src/io/parquet/writer_impl.cu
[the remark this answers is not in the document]
etseidl: Because each column chunk in the row group has to have the same number of rows, and ensuring this with variable fragment sizes seems very tricky.  It's much easier to use uniform fragment sizes to get the boundaries of each row group, and then readjust when necessary.

cpp/include/cudf/io/parquet.hpp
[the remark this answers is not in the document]
etseidl: Only since 1990 or so :(

cpp/src/io/parquet/writer_impl.cu
etseidl: The fragment code is one of the few things in the parquet writer that is insanely parallel.  I surrender :D

cpp/src/io/parquet/writer_impl.cu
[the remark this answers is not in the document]
etseidl: This might be a TL;DR, but last night I thought more on this question.  Let's assume a table with 3 columns, each wanting a different fragment size.  As this PR stands, pass 1 first produces rowgroup fragments to get the rowgroup boundaries, and then does a second pass to get the page fragments.
```
           curr PR pass 1                 pass 2
 
       C0      C1      C2           C0      C1      C2
     -----------------------       -----------------------
RG0 | F0    | F0    | F0    | RG0 | F0    | F0    | F0    |
    |       |       |       |     |       |       |-------|
    |       |       |       |     |       |       | F1    |
    |       |       |       |     |       |-------|-------|
    |       |       |       |     |       | F1    | F2    |
    |-------|-------|-------|     |-------|       |-------|
    | F1    | F1    | F1    |     | F1    |       | F3    |
    |       |       |       |     |       |-------|-------|
    |       |       |       |     |       | F2    | F4    |
    |       |       |       |     |       |       |-------|
    |       |       |       |     |       |       | F5    |
    |=======|=======|=======|     |=======|=======|=======|
RG1 | F0    | F0    | F0    | RG1 | F0    | F0    | F0    |
    |       |       |       |     |       |       |-------|
    |       |       |       |     |       |       | F1    |
    |       |       |       |     |       |-------|-------|
    |       |       |       |     |       | F1    | F2    |
    |-------|-------|-------|     |-------|       |-------|
    | F1    | F1    | F1    |     | F1    |       | F3    |
     ------- ------- -------       ------- ------- -------
```
One way to do a single pass would be to calculate the number of fragments for C1 and C2 in mini batches of frag_len(C0).  This keeps the RG boundaries sane, but leads to very different fragment sizes for C1 (we could make C0 a multiple of C2 to help with those two, but if there are many columns each with a different desired size it may not be worth it.
```
       C0      C1      C2
     -----------------------
RG0 | F0    | F0    | F0    | 21 fragments total
    |       |       |-------|
    |       |       | F1    |
    |       |-------|-------|
    |       | F1    | F2    |
    |-------|-------|-------|
    | F1    | F2    | F3    |
    |       |       |-------|
    |       |       | F4    |
    |       |-------|-------|
    |       | F3    | F5    |
    |=======|=======|=======|
RG1 | F0    | F0    | F0    |
    |       |       |-------|
    |       |       | F1    |
    |       |-------|-------|
    |       | F1    | F2    |
    |-------|-------|-------|
    | F1    | F2    | F3    |
     ------- ------- -------
```
Another approach could be to just use two fragment sizes; the largest and the smallest. This approach asymptotes on the single fragment size from #12627.
```
       C0      C1      C2
     -----------------------
RG0 | F0    | F0    | F0    | 24 fragments total
    |       |-------|-------|
    |       | F1    | F1    |
    |       |-------|-------|
    |       | F2    | F2    |
    |-------|-------|-------|
    | F1    | F3    | F3    |
    |       |-------|-------|
    |       | F4    | F4    |
    |       |-------|-------|
    |       | F5    | F5    |
    |=======|=======|=======|
RG1 | F0    | F0    | F0    |
    |       |-------|-------|
    |       | F1    | F1    |
    |       |-------|-------|
    |       | F2    | F2    |
    |-------|-------|-------|
    | F1    | F3    | F3    |
     ------- ------- -------
```
Profiling with a fairly large file (14G uncompressed, 3G snappy compressed, with 112 row groups), it takes about 8 seconds to do the write, `InitPageFragments` takes 11ms, and `RecalculatePageFragments` takes 10ms.  Given the low computational overhead, do you think the alternatives are worth exploring?  I'm not sure if they'll save on code complexity or not.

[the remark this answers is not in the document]
vuule: I'm happy with the current approach. I just didn't realize that uniform fragment sizes are required because of row group alignment.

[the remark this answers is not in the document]
etseidl: Is it worth adding that to the comment and/or renaming the `fragments` array to `rowgroup_fragments` to make this more explicit?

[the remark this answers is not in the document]
vuule: Please do, I like both suggestions :)

etseidl:
@vuule I found another issue while testing this.  While computing page boundaries, the `rows_in_page` check (page_enc.cu L427) should be checking rows_in_page + frag_g.num_rows.  This worked with `>=` when `max_page_size_rows` was a multiple of `fragment_size`, but now `max_page_size_rows` can be exceeded. I know my PRs tend to suffer from feature creep, so should I fix this here or do a separate PR after this is merged?

vuule:
> @vuule I found another issue while testing this. While computing page boundaries, the `rows_in_page` check (page_enc.cu L427) should be checking rows_in_page + frag_g.num_rows. This worked with `>=` when `max_page_size_rows` was a multiple of `fragment_size`, but now `max_page_size_rows` can be exceeded. I know my PRs tend to suffer from feature creep, so should I fix this here or do a separate PR after this is merged?

My vote is to fix in this PR, since it exposes the issue. I assume it's a small fix? 🤞

etseidl:
> 
> My vote is to fix in this PR, since it exposes the issue. I assume it's a small fix? 🤞 

One line fix, but I should probably add a test too.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -85,6 +87,49 @@ parquet::Compression to_parquet_compression(compression_type compression)
...
+  }
+
+  return 0;
+}
+
+// checks to see if the given column has a fixed size.  This doesn't
+// check every row, so assumes string and list columns are not fixed, even
+// if each row is the same width.
+// TODO: update this if FIXED_LEN_BYTE_ARRAY is ever supported for writes.
+bool is_col_fixed_width(column_view const& column)
```
vuule, line 119: Does this refer to Parquet columns in the output file?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -85,6 +87,49 @@ parquet::Compression to_parquet_compression(compression_type compression)
...
+  if (is_fixed_width(column.type())) {
+    return true;
+  } else if (column.type().id() == type_id::STRUCT) {
+    for (int i = 0; i < column.num_children(); i++) {
+      if (not is_fixed_width(column.child(i).type())) { return false; }
+    }
+    return true;
+  }
+
+  return false;
```
vuule, line 130: Should we recursively call `is_col_fixed_width? My (simpler?) version:
```suggestion
  if (column.type().id() == type_id::STRUCT) {
    return std::all_of(column.child_begin(), column.child_end(), is_col_fixed_width);
  }

  return is_fixed_width(column.type());
```

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1407,23 +1451,63 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+    std::transform(single_streams_table.begin(),
+                   single_streams_table.end(),
+                   column_sizes.begin(),
+                   column_frag_size.begin(),
+                   frag_size_fn);
   }
 
+  // Fragments are calculated in two passes.  In the first pass, a uniform number of fragments
+  // per column is used.  This is done to satisfy the requirement that each column chunk within
+  // a row group has the same number of rows.  After the row group (and thus column chunk)
```
vuule, line 1502:
```suggestion
  // Fragments are calculated in two passes. In the first pass, a uniform number of fragments
  // per column is used. This is done to satisfy the requirement that each column chunk within
  // a row group has the same number of rows. After the row group (and thus column chunk)
```

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -85,6 +87,49 @@ parquet::Compression to_parquet_compression(compression_type compression)
...
+    for (int i = 0; i < scol.num_children(); i++) {
+      ret += column_size(scol.get_sliced_child(i), stream);
+    }
+    return ret;
+  } else if (column.type().id() == type_id::LIST) {
+    auto const lcol = lists_column_view(column);
+    return column_size(lcol.get_sliced_child(stream), stream);
+  }
+
+  return 0;
```
vuule, line 112: should we throw here? looks like we would get here only if we hit a compound type that is not covered in the branches above.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -85,6 +87,49 @@ parquet::Compression to_parquet_compression(compression_type compression)
   }
 }
 
+size_type column_size(column_view const& column, rmm::cuda_stream_view stream)
+{
+  if (column.num_children() == 0) {
+    return is_fixed_width(column.type()) ? size_of(column.type()) * column.size() : 0;
+  }
```
vuule, line 94: here as well, should this throw if we hit a non-fixed width type with no children? If we throw at the bottom, this can become
```suggestion
  if (column.num_children() == 0 and is_fixed_width(column.type())) {
    return size_of(column.type()) * column.size();
  }
```

vuule (changes requested):
couple of small suggestions, mostly code simplifications.

cpp/src/io/parquet/writer_impl.cu
etseidl: Well, these are `parquet_column_views` of the input table.  I had thought that at this level (schema leaf nodes), that the data types would be down to either fixed types, or lists/strings.  I'm still not quite sure why structs make it this far.

cpp/src/io/parquet/writer_impl.cu
etseidl: Heh.  Learned to type in 7th grade and the rule was 2 spaces after a period.  It's just habit :D

cpp/src/io/parquet/writer_impl.cu
etseidl: Sounds good. Like above, still not sure why there's a struct as a leaf.

cpp/src/io/parquet/writer_impl.cu
etseidl: I'll have to check again on this one, but I think I did it this way because of some of the empty table tests.

cpp/src/io/parquet/writer_impl.cu
vuule: yeah, looking at the FIXED_LEN_BYTE_ARRAY comment, it seems like the logic applies to Parquet columns/types.

cpp/src/io/parquet/writer_impl.cu
vuule: looks good!

vuule (approved):
Looking good, thank you for implementing this!
It's complicated, but not as much as it's beneficial :)

cpp/src/io/parquet/writer_impl.cu
etseidl: The FIXED_LEN_BYTE_ARRAY comment is in reference to #12590.  I'm assuming that once cuDF can read Parquet FLBA's, we'll want to write them too. IIRC right now FLBA is only used to encode decimals, so the cuDF column type here will be fixed len.  But if Parquet FLBA's turn into cuDF list(byte) or some such, then we'll need some schema info here to figure out whether it's fixed len or variable.

etseidl:
Thanks for all the help!  Going to release it into the wild now :D

cpp/src/io/parquet/page_enc.cu
```diff
@@ -2057,33 +2092,35 @@ __global__ void __launch_bounds__(1)
...
 }
 
-void InitFragmentStatistics(device_2dspan<statistics_group> groups,
-                            device_2dspan<PageFragment const> fragments,
-                            device_span<parquet_column_device_view const> col_desc,
+void CalculatePageFragments(device_span<PageFragment> frag,
+                            device_span<size_type const> column_frag_sizes,
+                            rmm::cuda_stream_view stream)
+{
+  gpuCalculatePageFragments<512><<<frag.size(), 512, 0, stream>>>(frag, column_frag_sizes);
```
hyperbolic2346, line 2114: Is this legal now to call this without `stream.value()`?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -886,34 +926,33 @@ gpu::parquet_column_device_view parquet_column_view::get_device_view(
...
-    device_2dspan<statistics_group>(frag_stats_group.data(), num_columns, num_fragments);
-
-  gpu::InitFragmentStatistics(frag_stats_group_2dview, frag, col_desc, stream);
-  detail::calculate_group_statistics<detail::io_file_format::PARQUET>(frag_stats_chunk.data(),
-                                                                      frag_stats_group.data(),
-                                                                      num_fragments * num_columns,
-                                                                      stream,
-                                                                      int96_timestamps);
+  auto d_frag_sz = cudf::detail::make_device_uvector_async(frag_sizes, stream);
+  gpu::CalculatePageFragments(frag, d_frag_sz, stream);
```
hyperbolic2346, line 945: So this takes a span made from a hostdevice_vector, builds a device vector from it and runs calculations on it? Why are we building this device vector?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1572,7 +1650,72 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
     }
   }
 
-  // Build chunk dictionaries and count pages
+  // until now we've been using a uniform fragment size. if there are any columns with
```
hyperbolic2346, line 1653: I would be a fan of rewording this. In 2 years this comment will be odd to read. Maybe Before 23.04..."

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1572,7 +1650,72 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+  std::vector<size_type> frag_offsets;
+  size_type total_frags = 0;
+
+  if (frags_per_column.size() > 0) {
+    std::exclusive_scan(frags_per_column.data(),
+                        frags_per_column.data() + num_columns + 1,
+                        std::back_inserter(frag_offsets),
+                        0);
+    total_frags = frag_offsets[num_columns];
+  }
```
hyperbolic2346, line 1668:
```suggestion
  size_type const total_frags = [&]() {
    if (frags_per_column.size() > 0) {
      std::exclusive_scan(frags_per_column.data(),
                          frags_per_column.data() + num_columns + 1,
                          std::back_inserter(frag_offsets),
                          0);
      return frag_offsets[num_columns];
    } else {
      return 0;
    }
  };
```

hyperbolic2346 (changes requested):
I really like the use of `std::optional` in here over a magic default value. Some comments, but mostly nits.

cpp/src/io/parquet/page_enc.cu
etseidl: There's a `operator cudaStream_t()` on `cuda_stream_view` so technically not necessary to call `stream.value()`, but I'll change to be consistent.

cpp/src/io/parquet/writer_impl.cu
etseidl: frag_sizes is a `host_span` from a `std::vector`, not a `hostdevice_vector`.  Just need to push it to device for this one kernel.

cpp/src/io/parquet/writer_impl.cu
etseidl: will do...was actually meaning "up until this point in the function", but I can see the source of confusion

cpp/src/io/parquet/writer_impl.cu
etseidl: Nice.  Just had to add `()` at the end.

cpp/src/io/parquet/writer_impl.cu
hyperbolic2346: Right, I was looking at the first argument at the call site. My mistake.

hyperbolic2346 (approved):

PR #12688, opened 2023-02-03
Fix missing trailing comma in json writer

### LINKED ISSUE #12687: [BUG] Write JSON misses a comma every 8 records with `lines=False`
**Describe the bug**
When using the new JSON writer in cuDF-python with `lines=False`, the output skips the necessary comma periodically. This appears to happen every 8 records from my testing.

Here is a repro for the issue:
```
df = cudf.DataFrame({'a':range(18)})
df.to_json('j.json', engine='cudf', lines=False)
```
Here is the file:
`[{"a":0},{"a":1},{"a":2},{"a":3},{"a":4},{"a":5},{"a":6},{"a":7}{"a":8},{"a":9},{"a":10},{"a":11},{"a":12},{"a":13},{"a":14},{"a":15}{"a":16},{"a":17}]`
Note the missing comma between `{"a":7}` and `{"a":8}`.

and here is the error:
```
df = cudf.read_json('j.json', engine='cudf', lines=False)
RuntimeError: CUDF failure at:/nfs/repo/cudf-23.02/cpp/src/io/json/json_tree.cu:264: JSON Parser encountered an invalid format at location 64
```

**Expected behavior**
I expect to write valid JSON, where each record is separated by a comma.

Docker image from [rapidsai-dev-nightly](https://hub.docker.com/r/rapidsai/rapidsai-dev-nightly/tags?page=1&name=23.02) with id `0022659d9d65`. Built from source commit `3c39be5a9d7d69adaad47121359e0c084b76decf`

**Environment details**
skipped

**Additional context**
I have not seen a similar issue with `lines=True`

karthikeyann:
## Description
Fix missing trailing comma in json writer for non-Lines json format.
Updates default rows_per_chunk because `8` is too small and writer is slow.
closes #12687

## Checklist


bdice (approved):
Looks good. Do you have performance data for different chunk sizes?

karthikeyann:
image

bdice:
@karthikeyann Is there a reason we don't use a larger chunk size? It appears that chunk size 512k performance is better than chunk size 256k. Are there other constraints like memory usage that balance against runtime performance? (Are "chunks" needed at all?)

karthikeyann:
@bdice Yes. Memory usage is a major concern. chunk is not neeed. If we don't pass chunk size, it will convert entire table into string and will incur high memory usage (and may fail). Just to be safe, I used 256K.
If we want to use `rows_per_chunk` effectively, we should measure free memory and _estimate_ the chunk size based on free memory, number of columns (child columns and non-child). It is still an approximation.

For example:  In these perf measurements, I used a 2 column dataframe.
without chunksize, >128M rows fail, >2GB json fail.
For 256K chunksize, >512M  rows fail, >8GB json fail.

Using 512K would be good too. I am trying to keep the rows_per_chunk small so that if number of columns are high, it will not exceed memory limit. 

image

bdice:
Interesting! Let's make sure that there is guidance in the docs about this: `rows_per_chunk` should be decreased if writes fail due to running out of memory.

bdice:
See also this related request for CSV writing: https://github.com/rapidsai/cudf/issues/12690#issuecomment-1415805133

cpp/src/io/json/write_json.cu
```diff
@@ -581,14 +581,13 @@ std::unique_ptr<column> make_column_names_column(host_span<column_name_info cons
...
                    strings_column_view const& str_column_view,
-                   std::string const& line_terminator,
+                   string_scalar const& d_line_terminator,
                    json_writer_options const& options,
                    rmm::cuda_stream_view stream,
                    rmm::mr::device_memory_resource* mr)
 {
   CUDF_EXPECTS(str_column_view.size() > 0, "Unexpected empty strings column.");
 
-  string_scalar d_line_terminator{line_terminator};
```
ttnghia, line 591: So this was unused before?

python/cudf/cudf/_lib/json.pyx
```diff
@@ -156,7 +156,7 @@ def write_json(
     bool include_nulls=True,
     bool lines=False,
     bool index=False,
-    int rows_per_chunk=8,
+    int rows_per_chunk=1024*256,  # 256K rows
```
ttnghia, line 159: Why it was set to `8` before? `8` rows per chunk is too small.

karthikeyann: CSV writer uses `8`, so used the same here!
https://github.com/rapidsai/cudf/blob/9c9862f97b00fc90760c3371f30ef5769f4163f5/python/cudf/cudf/_lib/csv.pyx#L476

cpp/src/io/json/write_json.cu
karthikeyann: It was used before too. Instead of passing `std::string` as argument, now `string_scalar` is passed.

ttnghia (approved):

Issue #11460, opened 2022-08-04
[FEA] Move pandas-related notes from Notes to pandas-compat


vyasr:
**Is your feature request related to a problem? Please describe.**
#8643 introduced a new rST directive `pandas-compat` to our documentation that allows us to collect all documentation relating to differences between cuDF methods and pandas methods. That directive enables us to much more effectively indicate to users when our behavior differs from pandas. However, we have not since made much use of this directive. Many docstrings still contain `Notes` sections that are essentially just descriptions of how a class/function/method differs from pandas.

**Describe the solution you'd like**
We should convert all pandas-related `Notes` sections of docstrings into `pandas-compat` sections. Doing so should be quite straightforward by simply searching for the header `Notes` in all files containing public APIs. For examples on how to use the `pandas-compat` directive, see #8643.

**Describe alternatives you've considered**
None

**Additional context**
There are likely many other places where we differ from pandas that are either 1) undocumented or 2) documented in different ways. We should also aim to improve that documentation. However, that is a nebulous target since it's not immediately clear where all those places are, so this issue is just focusing on the more specific goal of replacing `Notes` sections, which can be easily searched for and replaced as needed.

Touutae-lab:
I would like to help this issues.

shwina:
Hi @Touutae-lab -- thanks for your interest in contributing! This would absolutely be a welcome addition to cuDF. Please feel free to comment here if you need any help as you work on this issue.

Touutae-lab:
### Hi, I have some question how should i handle this ?

![Screenshot from 2023-01-28 16-09-24](https://user-images.githubusercontent.com/58826535/215257559-5fbcbe4f-485f-405d-83b5-7105dda3247a.png)

### so I handle like this
![Screenshot from 2023-01-28 16-23-12](https://user-images.githubusercontent.com/58826535/215258392-b35e2ecb-4ca8-4f2c-9ea3-0418a622e5bd.png)

### after compilation it not showing in the last `PandasCompat.html`
```python
"""
        Examples
        --------
        >>> import cudf
        >>> df = cudf.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})
        >>> df
           A  B
        0  1  4
        1  2  5
        2  3  6

        Rename columns using a mapping:

        >>> df.rename(columns={"A": "a", "B": "c"})
           a  c
        0  1  4
        1  2  5
        2  3  6

        Rename index using a mapping:

[...]

PR #12641, opened 2023-01-30
Notes convert to Pandas-compat

Touutae-lab:
# Description
Closes #11460
 
image

**I already change and test documentation in my localhost but some of it can't be change due**
> ### BUG
> can't convert Note froms `DataFrame.mode`
> image
>
> convert to 
> image
>
> ```
> Extension error (numpydoc.numpydoc):
> Handler <function mangle_signature at 0x7fe25c23d630> for event 'autodoc-process-signature' threw an exception (exception: 
> Error parsing See Also entry '.. pandas-compat::' in the docstring of mode in /home/touutae/miniconda3/envs/cudf_dev
> /lib/python3.10/contextlib.py.)
> make: *** [Makefile:20: html] Error 2
> ```
>
> it only happened when edited `DataFrame.mode`
>
> _Originally posted by @Touutae-lab in https://github.com/rapidsai/cudf/issues/11460#issuecomment-1407660086_

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12691, opened 2023-02-03
Convert Notes (NEW)

Touutae-lab:
From [11460](https://github.com/rapidsai/cudf/issues/11460)


Touutae-lab:
I opened new Pull because #12641 is broken

wence-:
> I opened new Pull because #12641 is broken

Sorry, can you explain how? It would be good if you could reopen and we can try and unbreak things there (that way we don't lose the existing review comments).

Touutae-lab:
OK will try to fix that and reopen

PR #12592, opened 2023-01-23
Add docs build job

AyodeAwe:
The PR adds a `docs_build` process to the PR and Build workflows for this repository. The generated docs are synced to s3 for only the build workflows.

Issue #9593, opened 2021-11-03
[FEA] Reduce implicit conversions to minimize unnecessary column materialization


vyasr:
**Is your feature request related to a problem? Please describe.**
A number of deceptively harmless functions in cuDF Python actually involve insidious memory allocations that in aggregate have substantial deleterious impacts on performance. The two most prominent examples of this are `ColumnBase.astype` calls and all `RangeIndex` logic that involves materializing an `Int64Index`, although others may exist as well. While these conversions are sometimes unavoidable, in many cases they are not. `RangeIndex` conversions are particularly harmful because the class is designed precisely to avoid such overheads, so many internal code paths are likely creating unnecessary and unexpected memory pressure.

**Describe the solution you'd like**
Type conversions that are currently performed deep in the call stack should be moved to the highest possible level. This change would prevent APIs from performing the cast except where absolutely necessary, and at minimum would reduce the number of calls to validation functions like `is_categorical_dtype`. Internal functions should accept a relatively narrow set of parameter types, and only user-facing functions should perform casts to handle the near-arbitrary user data that pandas APIs require us to accept.

On the Index front, there are a number of tasks that should be tackled separately. Roughly, the following two tasks should be possible to address in parallel:
- Currently `RangeIndex` implicitly defers all methods that aren't implemented to `Int64Index` via the `__getattr__` override. We should remove this override and actually implement the relevant methods in `RangeIndex`. In cases where materializing an `Int64Index` is necessary, we should exploit the sort of monkey-patching that we do for defining unary and binary operations for `RangeIndex` to add those methods explicitly. In other words, we should use an allowlist rather than a blocklist to choose methods where the conversion happens. 
- `BaseIndex` should be reduced as closely as possible to an abstract class. While there are a subset of APIs that truly make sense for all types of index objects, in almost all cases the optimal implementation for `RangeIndex` (and `MultiIndex`, for that matter) is very different from the implementation for `GenericIndex`. In addition, this change reduces cognitive load for developers by simplifying the inheritance hierarchy.

Once those two tasks are accomplished to whatever degree makes sense, we should revisit removing two attributes that contribute to this problem:
[...]

PR #12693, opened 2023-02-03
Remove outdated doc references

AyodeAwe:
## Description
Removes documentation for deprecated references. This issue was discovered while building the docs in #12592.

See https://github.com/rapidsai/cudf/pull/12684

### DIFF
--- a/docs/cudf/source/api_docs/index_objects.rst
+++ b/docs/cudf/source/api_docs/index_objects.rst
@@ -21,19 +21,14 @@ Properties
 .. autosummary::
    :toctree: api/
 
-   Index.empty
    Index.has_duplicates
    Index.duplicated
    Index.hasnans
    Index.is_monotonic
    Index.is_monotonic_increasing
    Index.is_monotonic_decreasing
-   Index.is_unique
-   Index.name
    Index.names
-   Index.ndim
    Index.nlevels
-   Index.shape
    Index.size
    Index.values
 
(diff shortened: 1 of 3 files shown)


AyodeAwe:
Getting this error as well. I'm not sure how to fix:


```
Warning, treated as error:
[autosummary] failed to import cudf.Index.dtype.
Possible hints:
* ImportError: 
* ModuleNotFoundError: No module named 'cudf.Index'
* AttributeError: type object 'Index' has no attribute 'dtype'
```

bdice:
I see this PR is deleting docs for a lot of Index methods such as `Index.where` and `Index.argsort`, which do exist in pandas. It looks like these exist in cudf too (I only checked these two):

https://github.com/rapidsai/cudf/blob/e3803319d9578e976e90b5cf05c99dc3d7063551/python/cudf/cudf/core/index.py#L847

https://github.com/rapidsai/cudf/blob/e3803319d9578e976e90b5cf05c99dc3d7063551/python/cudf/cudf/core/index.py#L863

I would guess that there's an issue with how the docs are generated, or something funny in our import logic or class hierarchy. I believe I have discussed this with @vyasr and @galipremsagar before -- do either of you have input?

bdice (changes requested):
Some of these method should not be deleted -- we probably have to fix something else. See comment above; requesting changes for now to block merger.

vyasr:
These issues arise because of our type hierarchy separating between `BaseIndex`, the actual parent class for all of our indexes, and `Index`, the subclass of `BaseIndex` that essentially functions as a factory for creating indexes corresponding to data of a particular dtype. The correct solution for most of the errors of the sort that you're pointing out is to define the methods in `BaseIndex` with a docstring but have them raise a `NotImplementedError`. Implementation should be left to the subclasses. This is described further in #9593, but the key relevant point is that:
> BaseIndex should be reduced as closely as possible to an abstract class.

PR #12698, opened 2023-02-03
Handle parquet list data corner case

nvdbaranec:
Fixes an issue with a particular arrangement of page data related to lists.  Specifically, it is possible for page `N` to contain "0" rows because the values for the row it is a part of start on page `N-1` and end on page `N+1`.  This was defeating logic in the decode kernel that would erroneously cause these values to be skipped.

Similar to https://github.com/rapidsai/cudf/pull/12488 this is only reproducible with data out in the wild.  In this case, we have a file that we could in theory check in to create a test with, but it is 16 MB so it's fairly large.  Looking for feedback on whether this is too big.

## Checklist

### DIFF
--- a/cpp/src/io/parquet/page_data.cu
+++ b/cpp/src/io/parquet/page_data.cu
@@ -119,6 +119,16 @@ inline __device__ bool is_bounds_page(page_state_s* const s, size_t min_row, siz
   return ((page_begin <= begin && page_end >= begin) || (page_begin <= end && page_end >= end));
 }
 
+/**
+ * @brief Returns whether or not a page is completely contained within the specified
+ * row bounds
+ *
+ * @param s The page to be checked
+ * @param min_row The starting row index
+ * @param num_rows The number of rows
+ *
+ * @return True if the page is completely contained within the row bounds
+ */
 inline __device__ bool page_is_contained(page_state_s* const s, size_t min_row, size_t num_rows)
 {
   size_t const page_begin = s->col.start_row + s->page.chunk_row;
@@ -1849,8 +1859,15 @@ __global__ void __launch_bounds__(block_size) gpuDecodePageData(
   bool const has_repetition = s->col.max_level[level_type::REPETITION] > 0;
 
   // if we have no work to do (eg, in a skip_rows/num_rows case) in this page.
+  //
   // corner case: in the case of lists, we can have pages that contain "0" rows if the current row
-  // starts before this page and ends after this page
+  // starts before this page and ends after this page:
+  //       P0        P1        P2
+  //  |---------|---------|----------|
+  //        ^------------------^
+  //      row start           row end
+  // P1 will contain "0" rows
+  //
   if (s->num_rows == 0 && !(has_repetition && (is_bounds_page(s, min_row, num_rows) ||
                                                page_is_contained(s, min_row, num_rows)))) {
     return;
--- a/cpp/src/io/parquet/page_data.cu
+++ b/cpp/src/io/parquet/page_data.cu
@@ -1749,10 +1749,11 @@ __global__ void __launch_bounds__(block_size)
       auto const thread_depth = depth + t;
       if (thread_depth < s->page.num_output_nesting_levels) {
         // if we are not a bounding page (as checked above) then we are either
-        // returning 0 rows from the page (completely outside the bounds) or all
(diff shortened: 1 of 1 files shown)


nvdbaranec:
@etseidl

cpp/src/io/parquet/page_data.cu
```diff
@@ -115,9 +115,30 @@ inline __device__ bool is_bounds_page(page_state_s* const s, size_t min_row, siz
...
+
   return ((page_begin <= begin && page_end >= begin) || (page_begin <= end && page_end >= end));
 }
 
+/**
+ * @brief Returns whether or not a page is completely contained within the specified
+ * row bounds
+ *
+ * @param s The page to be checked
+ * @param min_row The starting row index
```
bdice, line 127: I think "start" is a more appropriate word than "minimum."
```suggestion
 * @param start_row The starting row index
```

cpp/src/io/parquet/page_data.cu
```diff
@@ -1838,7 +1860,19 @@ __global__ void __launch_bounds__(block_size) gpuDecodePageData(
   bool const has_repetition = s->col.max_level[level_type::REPETITION] > 0;
 
   // if we have no work to do (eg, in a skip_rows/num_rows case) in this page.
-  if (s->num_rows == 0 && !(has_repetition && is_bounds_page(s, min_row, num_rows))) { return; }
+  //
+  // corner case: in the case of lists, we can have pages that contain "0" rows if the current row
```
bdice, line 1864: Do the quotes mean something? If not, I'd remove.
```suggestion
  // corner case: in the case of lists, we can have pages that contain 0 rows if the current row
```

cpp/src/io/parquet/page_data.cu
```diff
@@ -1838,7 +1860,19 @@ __global__ void __launch_bounds__(block_size) gpuDecodePageData(
...
+  // starts before this page and ends after this page:
+  //       P0        P1        P2
+  //  |---------|---------|----------|
+  //        ^------------------^
+  //      row start           row end
+  // P1 will contain "0" rows
```
bdice, line 1870:
```suggestion
  // P1 will contain 0 rows
```

nvdbaranec: The idea with the quotes was to imply zero-but-not-really.

bdice: I'm not sure if I understand the "but not really" portion. That seems important to understand the scope of the bug. Can you elaborate? I may just be missing some knowledge about the Parquet spec.

vuule: My understanding is that no rows start (or end) in such page. So it does not (fully) contain any rows, but it does contain data that we need to parse (and apparently we didn't, before this PR).
That's at least how I understand the ASCII art in the comment above :)

bdice: Got it. That helps. Thanks! I’ll approve.

bdice: We don’t need to remove the air-quotes, but maybe the comment can be updated with the description @vuule wrote above.

bdice (approved):

cpp/src/io/parquet/page_data.cu
```diff
@@ -115,9 +115,30 @@ inline __device__ bool is_bounds_page(page_state_s* const s, size_t min_row, siz
...
+ * @brief Returns whether or not a page is completely contained within the specified
+ * row bounds
+ *
+ * @param s The page to be checked
+ * @param min_row The starting row index
+ * @param num_rows The number of rows
+ *
+ * @return True if the page is completely contained within the row bounds
+ */
+inline __device__ bool page_is_contained(page_state_s* const s, size_t min_row, size_t num_rows)
```
vuule, line 132: I think this pattern is more common
```suggestion
inline __device__ bool is_page_contained(page_state_s* const s, size_t min_row, size_t num_rows)
```

vuule (commented):
looks good, just a naming suggestion

vuule (approved):

nvdbaranec:
rerun tests

vuule:
@nvdbaranec "rerun tests" does not work any more :(
You can rerun failed jobs in the UI once the CI completes current jobs.

PR #12704, opened 2023-02-06
`partition_by_hash()`: use `_split()`

madsbk:
#### Description
Reduce overhead of `Frame.partition_by_hash()` by calling `Frame._split()`.

#### Benchmark
Small benchmark of this PR shows ~1.5x speedup:
https://gist.github.com/madsbk/308df2dd58309510610fd27e0529f862

#### Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -4614,10 +4614,12 @@ def partition_by_hash(self, columns, nparts, keep_index=True):
...
         )
-        # Slice into partition
-        ret = [outdf[s:e] for s, e in zip(offsets, offsets[1:] + [None])]
-        if not keep_index:
-            ret = [df.reset_index(drop=True) for df in ret]
+        # Slice into partitions. Notice, `hash_partition` returns the start
+        # offset of each partition thus we skip the first offset
+        ret = outdf._split(offsets[1:], keep_index=keep_index)
+
+        # `_split()`` ignores empty intervals so we add empty partitions here
```
bdice, line 4621: Is it expected / by design that `_split` ignores empty partitions? That sounds odd to me. I glanced at the libcudf side but didn't see where empty partitions were being dropped.

bdice (commented):
Nice catch! I have a question about the design of `_split`.

wence- (commented):
Same kind of question as @bdice.

python/cudf/cudf/core/dataframe.py
madsbk: @bdice @wence-, you are both right. `_split` doesn't ignore empty partitions. I was confused because calling `_split` on an empty dataframe returns an empty list always:
```python
>>> import cudf
>>> cudf.DataFrame({"a": [1,2,3]})
>>> df = cudf.DataFrame({"a": []})
>>> df._split([0,0])
[]
```

I have updated the docs

python/cudf/cudf/core/dataframe.py
```diff
@@ -4614,10 +4614,12 @@ def partition_by_hash(self, columns, nparts, keep_index=True):
             self._column_names,
             self._index_names if keep_index else None,
         )
-        # Slice into partition
-        ret = [outdf[s:e] for s, e in zip(offsets, offsets[1:] + [None])]
-        if not keep_index:
-            ret = [df.reset_index(drop=True) for df in ret]
+        # Slice into partitions. Notice, `hash_partition` returns the start
+        # offset of each partition thus we skip the first offset
+        ret = outdf._split(offsets[1:], keep_index=keep_index)
```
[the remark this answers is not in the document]
madsbk: Yes, or at least down to 10 partitions on `dgx14`. When `nparts<10`, the run time is around 5ms for both of them.

python/cudf/cudf/core/dataframe.py
wence-: Semantically, `df._split([a, b])` should be equivalent to `[df[:a], df[a:b], df[b:]]`, so I would argue that this should have returned a list of three empty dataframes (AKA, this is a bug in the API).

But let us fix it separately.

python/cudf/cudf/core/dataframe.py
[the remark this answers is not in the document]
wence-: Thanks!

python/cudf/cudf/core/dataframe.py
wence-: Opened https://github.com/rapidsai/cudf/issues/12717

wence- (approved):
Java failures are waiting on https://github.com/rapidsai/cudf/pull/12723 I think.

PR #12705, opened 2023-02-06
Produce useful guidance on overflow error in `to_csv`

### LINKED ISSUE #12690: [ENH]: mention chunksize argument when to_csv fails
**Describe the bug**
I am facing the following error when saving a CSV file after reading a DataFrame from parquet.

```python
import cudf

df = cudf.read_parquet("10_000_000a_rows.parquet")
df.to_csv("10_000_000a_rows.csv",index=False)
```

> ---------------------------------------------------------------------------
> RuntimeError                              Traceback (most recent call last)
> /tmp/ipykernel_1285/3527798299.py in <module>
> ----> 1 df.to_csv("10_000_000a_rows.csv",index=False)
> 
> /opt/conda/envs/rapids/lib/python3.9/site-packages/cudf/core/dataframe.py in to_csv(self, path_or_buf, sep, na_rep, columns, header, index, encoding, compression, line_terminator, chunksize, storage_options)
>    6332         from cudf.io import csv
>    6333 
> -> 6334         return csv.to_csv(
>    6335             self,
>    6336             path_or_buf=path_or_buf,
> 
> /opt/conda/envs/rapids/lib/python3.9/contextlib.py in inner(*args, **kwds)
>      77         def inner(*args, **kwds):
>      78             with self._recreate_cm():
> ---> 79                 return func(*args, **kwds)
>      80         return inner
>      81 
> 
> /opt/conda/envs/rapids/lib/python3.9/site-packages/cudf/io/csv.py in to_csv(df, path_or_buf, sep, na_rep, columns, header, index, encoding, compression, line_terminator, chunksize, storage_options)
>     239             )
>     240     else:
> --> 241         libcudf.csv.write_csv(
>     242             df,
>     243             path_or_buf=path_or_buf,
> 
> csv.pyx in cudf._lib.csv.write_csv()
> 
> csv.pyx in cudf._lib.csv.write_csv()
> 
> RuntimeError: cuDF failure at: /workspace/.conda-bld/work/cpp/src/column/column_factories.cpp:82: Column size cannot be negative.

The file is too big to share here. I will share the file with @beckernick internally.

**Expected behavior**
No error.

RAPIDS 22.12 using NGC container.

**Environment details**
DGX-1 Server

wence-:
## Description

Since writing to CSV files is implemented by converting all columns in
a dataframe to strings, and then concatenating those columns, when we
attempt to write a large dataframe to CSV without specifying a chunk
size, we can easily overflow the maximum column size.

Currently the error message is rather inscrutable: that the requested
size of a string column exceeds the column size limit. To help the
user, catch this error and provide a useful error message that points
them towards setting the `chunksize` argument.

So that we don't produce false positive advice, tighten the scope by
only catching `OverflowError`, to do this, make partial progress
towards resolving #10200 by throwing `std::overflow_error` when
checking for overflow of string column lengths.

Closes #12690.


cpp/include/cudf/detail/sizes_to_offsets_iterator.cuh
```diff
@@ -317,7 +318,8 @@ std::pair<std::unique_ptr<column>, size_type> make_offsets_child_column(
   auto const total_elements = sizes_to_offsets(input_itr, input_itr + count + 1, d_offsets, stream);
   CUDF_EXPECTS(
     total_elements <= static_cast<decltype(total_elements)>(std::numeric_limits<size_type>::max()),
```
wence-, line 320: I was worried that the other uses of this pattern always explicitly cast to `int64_t` event though the output of `sizes_to_offsets` can be either `uint64_t` or `int64_t`. However, after reading all the implicit promotion and binop conversion rules a number of times, and looking at cppinsights, I think both patterns are safe (since in the case that one operand is unsigned the signed operand will be promoted which is not lossy).

In fact, I _think_ that is is safe to just have [`total_elements <= std::numeric_limits<size_type>::max()`](https://cppinsights.io/lnk?code=I2luY2x1ZGUgPGxpbWl0cz4KI2luY2x1ZGUgPGNzdGRpbnQ+CgppbnQgbWFpbih2b2lkKQp7CiAgICB1aW50NjRfdCB1YSA9IDEwOwogICAgaW50NjRfdCBpYSA9IDEwOwogICAgYXV0byBpYiA9IHN0ZDo6bnVtZXJpY19saW1pdHM8aW50Pjo6bWF4KCk7CiAgICBhdXRvIHViID0gc3RkOjpudW1lcmljX2xpbWl0czxpbnQ+OjptYXgoKTsKICAgIGF1dG8gdXUgPSB1YSA8PSB1YjsKICAgIGF1dG8gdWkgPSB1YSA8PSBpYjsKICAgIGF1dG8gaXUgPSBpYSA8PSB1YjsKICAgIGF1dG8gaWkgPSBpYSA8PSBpYjsKICAgIHJldHVybiAwOwp9&insightsOptions=cpp17,all-implicit-casts&std=cpp17&rev=1.0) although it's probably nicer to keep the casts to help readers.

davidwendt (approved):
C++ LGTM

shwina (approved):
Python changes look good!

cpp/src/strings/regex/utilities.cuh
```diff
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <stdexcept>
```
bdice, line 19: ...and similarly. Move to the bottom in its own section, or with other STL includes.

bdice (dismissed):

ttnghia (approved):

python/cudf/cudf/_lib/csv.pyx
```diff
@@ -533,8 +533,14 @@ def write_csv(
...
-    with nogil:
-        cpp_write_csv(options)
+    try:
+        with nogil:
+            cpp_write_csv(options)
+    except OverflowError:
+        raise OverflowError(
+            f"Writing CSV file with chunksize={rows_per_chunk} failed. "
+            "Consider providing a smaller chunksize argument."
+        )
```
karthikeyann, line 543: JSON writer also could use similar code as above.

karthikeyann: can you update json.pyx as well?

wence-: Done.

cpp/src/strings/repeat_strings.cu
```diff
@@ -36,6 +36,8 @@
 #include <thrust/transform.h>
 #include <thrust/transform_reduce.h>
 
+#include <stdexcept>
+
```
karthikeyann, line 40:
```suggestion
```

karthikeyann (approved):
LGTM 👍

cpp/src/strings/repeat_strings.cu
wence-: Thanks.

davidwendt:
There is one more test to fix as found here:
https://github.com/rapidsai/cudf/actions/runs/4192959486/attempts/1#summary-11386077118

wence-:
> There is one more test to fix as found here:
> https://github.com/rapidsai/cudf/actions/runs/4192959486/attempts/1#summary-11386077118

Thanks @davidwendt, will do!

Issue #12686, opened 2023-02-03
[BUG] Some groupby.apply UDFs compute `mean` incorrectly


GregoryKimball:
**Describe the bug**
Some combinations of `if/else` conditions and `mean` aggregations result in incorrect results with `groupby.apply` UDFs.

**Steps/Code to reproduce bug**
Here is a snippet that reproduces the behavior:
```
import cudf
df = cudf.DataFrame({
    'a':[0,0,0,1,1,1],
    'b':[1,1,1,2,3,4],    
})

def udf(x):
    if x['a'].mean() > 0:
        return x['b'].mean()
    return 0

g = df.groupby('a').apply(udf, engine='jit')
```
which returns:
```
   a      None
0  0  0.000000
1  1  0.666667
```
This is some strange case because changing the conditional or return statements can cause this kind of UDF to return the correct answer. In this case, the computation appears to be dividing the first value by the length rather than the sum by the length.

**Expected behavior**
I expect this result:
```
   a      None
0  0  0.0
1  1  3.0
```

**Environment overview (please complete the following information)**
Docker image from [rapidsai-dev-nightly](https://hub.docker.com/r/rapidsai/rapidsai-dev-nightly/tags?page=1&name=23.02) with id `0022659d9d65`. Built from source commit `3c39be5a9d7d69adaad47121359e0c084b76decf`

**Environment details**
<details><summary>Click here to see environment details</summary><pre>
     
     **git***
     commit 3c39be5a9d7d69adaad47121359e0c084b76decf (HEAD -> branch-23.02, origin/branch-23.02)
     Author: AJ Schmidt <ajschmidt8@users.noreply.github.com>
     Date:   Wed Feb 1 15:44:09 2023 -0500
     
     Don't upload `libcudf-example` to Anaconda.org (#12671)
     
     When initially created, the GitHub Actions scripts that upload conda packages would upload all packages that were built indiscriminately.
     
     As a result, the `libcudf-example` package has been uploading to Anaconda.org. It was not previously uploaded to Anaconda.org prior to migrating to GitHub Actions.
     
     While this isn't a huge deal, I recently added some functionality ([here](https://github.com/rapidsai/shared-action-workflows/pull/33)) that allows us to specify package names that shouldn't be uploaded to Anaconda.org in our workflows.
     
     This PR adds the `libcudf-example` package to the list of packages that should not be uploaded to Anaconda.org
     
     Authors:
     - AJ Schmidt (https://github.com/ajschmidt8)
[...]

PR #12706, opened 2023-02-06
Fix faulty conditional logic in JIT `GroupBy.apply`

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/12686

### DIFF
--- a/python/cudf/udf_cpp/groupby/function.cu
+++ b/python/cudf/udf_cpp/groupby/function.cu
@@ -284,7 +284,7 @@ extern "C" {
   __device__ int name##_##cname(return_type* numba_return_value, type* const data, int64_t size) \
   {                                                                                              \
     return_type const res = name<type>(data, size);                                              \
-    if (threadIdx.x == 0) { *numba_return_value = res; }                                         \
+    *numba_return_value   = res;                                                                 \
     __syncthreads();                                                                             \
     return 0;                                                                                    \
   }
@@ -309,8 +309,8 @@ extern "C" {
   __device__ int name##_##cname(                                                 \
     int64_t* numba_return_value, type* const data, int64_t* index, int64_t size) \
   {                                                                              \
-    auto const res = name<type>(data, index, size);                              \
-    if (threadIdx.x == 0) { *numba_return_value = res; }                         \
+    auto const res      = name<type>(data, index, size);                         \
+    *numba_return_value = res;                                                   \
     __syncthreads();                                                             \
     return 0;                                                                    \
   }
--- a/python/cudf/udf_cpp/groupby/function.cu
+++ b/python/cudf/udf_cpp/groupby/function.cu
@@ -285,6 +285,7 @@ extern "C" {
   {                                                                                              \
     return_type const res = name<type>(data, size);                                              \
     *numba_return_value   = res;                                                                 \
+    __syncthreads();                                                                             \
     return 0;                                                                                    \
   }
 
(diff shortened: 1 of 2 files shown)


python/cudf/udf_cpp/groupby/function.cu
```diff
@@ -284,7 +284,7 @@ extern "C" {
   __device__ int name##_##cname(return_type* numba_return_value, type* const data, int64_t size) \
   {                                                                                              \
     return_type const res = name<type>(data, size);                                              \
-    if (threadIdx.x == 0) { *numba_return_value = res; }                                         \
+    *numba_return_value   = res;                                                                 \
     __syncthreads();                                                                             \
```
wence-, line 288: I think the `__syncthreads()` is now redundant, but I am not very confident, since I realise I don't see how these device functions fit together (in terms of how they are called from a numba `__global__` function).

python/cudf/udf_cpp/groupby/function.cu
```diff
@@ -309,8 +309,8 @@ extern "C" {
   __device__ int name##_##cname(                                                 \
     int64_t* numba_return_value, type* const data, int64_t* index, int64_t size) \
   {                                                                              \
-    auto const res = name<type>(data, index, size);                              \
-    if (threadIdx.x == 0) { *numba_return_value = res; }                         \
+    auto const res      = name<type>(data, index, size);                         \
+    *numba_return_value = res;                                                   \
     __syncthreads();                                                             \
```
wence-, line 314: Same here, I think `__syncthreads()` is not needed.

brandon-b-miller: Removed

python/cudf/udf_cpp/groupby/function.cu
brandon-b-miller: Removed

wence- (approved):
can you run the test with `compute-sanitizer --tool racecheck` just to belt-and-braces this? But I think this is right.

brandon-b-miller:
> can you run the test with `compute-sanitizer --tool racecheck` just to belt-and-braces this? But I think this is right.

I actually _do_ see a race here, but _only_ with the `__syncthreads()` removed. It appears to be happening in shared memory:

```
python/cudf/cudf/tests/test_groupby.py ========= Warning: (Warp Level Programming) Potential WAR hazard detected at __shared__ 0x0 in block (1,0,0) :
=========     Read Thread (31,0,0) at 0x1350 in BlockMean_int64
=========     Write Thread (0,0,0) at 0x150 in BlockMean_int64
=========     Current Value : 0, Incoming Value : 0
```
I also only observe this race when calling this single conditional function. I am not _exactly_ sure why we'd see it under these conditions but I am thinking it maybe sees the possibility that the `block_sum` variable could be overwritten by a thread that manages to encounter the second call to `mean` while some of the threads are still executing the first call.

Perhaps it's safer to as a policy for now keep the synchronization in and keep our blocks of threads logically working on the same task at the same time until all of them are done.

wence- (approved):
Thanks.

We should figure out that racecheck thing, but let us not hold up this fix for that.

python/cudf/udf_cpp/groupby/function.cu
wence-: And back in because racecheck...

python/cudf/udf_cpp/groupby/function.cu
wence-: And back in...

PR #12710, opened 2023-02-06
[REVIEW] Unpin `dask` and `distributed` for development

galipremsagar:
## Description
This PR unpins `dask` and `distributed` for `23.04` development.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12695, opened 2023-02-03
[REVIEW] Pin `dask` and `distributed` for release

galipremsagar:
## Description
This PR pins `dask` and `distributed` to `2023.1.1` for `23.02` release.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12712, opened 2023-02-06
Remove gpuCI scripts.

bdice:
Since migrating to GitHub Actions, the gpuCI scripts are no longer needed. This PR removes those outdated gpuCI scripts.


davidwendt:
I would recommend keeping this files until all build features have been migrated to GitHub Actions since the scripts here would be useful to when enabling them. This includes the Build Metrics Report and the compute-sanitizer checks.

bdice:
@davidwendt I'm tracking those tasks in a separate issue: https://github.com/rapidsai/ops/issues/2447#issuecomment-1343422135

If it's alright with you, I'd like to remove these files sooner than later (and refer back to previous branches as needed) because there is maintenance work like #12695, #12710 that is editing these unused files.

raydouglass (approved):

ajschmidt8 (approved):

PR #12714, opened 2023-02-07
Reenable stream identification library in CI

vyasr:
## Description
This PR reenables the preload library introduced for verifying stream usage in libcudf in #11875. This library was disabled during the GitHub Actions migration.

## Checklist


PointKernel (approved):

cpp/CMakeLists.txt
```diff
@@ -50,6 +50,11 @@ option(BUILD_SHARED_LIBS "Build cuDF shared libraries" ON)
 option(JITIFY_USE_CACHE "Use a file cache for JIT compiled kernels" ON)
 option(CUDF_BUILD_TESTUTIL "Whether to build the test utilities contained in libcudf" ON)
 mark_as_advanced(CUDF_BUILD_TESTUTIL)
+# TODO: Should this be on or off by default? Is it an advanced option?)
```
bdice, line 53: I am fine with it being on by default, as an advanced option. That matches the test utilities above.
```suggestion
```

cpp/CMakeLists.txt
```diff
@@ -754,10 +759,32 @@ if(CUDF_BUILD_TESTUTIL)
...
-
   add_library(cudf::cudftestutil ALIAS cudftestutil)
 
 endif()
+
+# * build cudf_identify_stream_usage --------------------------------------------------------------
+
+if(CUDF_BUILD_STREAMS_TEST_UTIL)
+  if(CUDA_STATIC_RUNTIME)
+    # TODO: Should this be a hard error, or just a skip? Depends on what the default is IMO.
```
bdice, line 770: If the default is ON, then we should skip gracefully instead of issuing a fatal error.

vyasr: I decided to have `CUDF_BUILD_STREAMS_TEST_UTIL` default to the value of `CUDA_STATIC_RUNTIME`. I think that should be sufficient.

ci/test_cpp.sh
```diff
@@ -34,13 +34,10 @@ nvidia-smi
...
 
-# TODO: Disabling stream identification for now.
-# Set up library for finding incorrect default stream usage.
-#pushd "cpp/tests/utilities/identify_stream_usage/"
-#mkdir build && cd build && cmake .. -GNinja && ninja && ninja test
-#STREAM_IDENTIFY_LIB="$(realpath build/libidentify_stream_usage.so)"
-#echo "STREAM_IDENTIFY_LIB=${STREAM_IDENTIFY_LIB}"
-#popd
+# Get library for finding incorrect default stream usage.
+STREAM_IDENTIFY_LIB=$(conda info | grep "active env location" | awk 'BEGIN { FS = " : " } ; {print $2}')/lib/libcudf_identify_stream_usage.so
```
ajschmidt8, line 38:
```suggestion
STREAM_IDENTIFY_LIB="${CONDA_PREFIX}/lib/libcudf_identify_stream_usage.so"
```

This does what you're looking for. 

We use it for gtests on line 47 in this file.

I also just confirmed locally in the `rapidsai/ci:latest` image.

ajschmidt8 (changes requested):

ci/test_cpp.sh
vyasr: Ah yes that's what I was looking for. Thanks!

ajschmidt8 (approved):

bdice (approved):
All good to me. I checked the CI logs and I think it looks correct.

davidwendt:
Just curious how long this takes to run?
Our cpp-test matrix is very large. I wonder if running this under "nightly" would be better since I don't the entire test matrix is necessary to verify stream usage.

vyasr:
> Just curious how long this takes to run? Our cpp-test matrix is very large. I wonder if running this under "nightly" would be better since I don't the entire test matrix is necessary to verify stream usage.

This doesn't really add measureable overhead because I'm not running any new tests, I'm just running all of the existing tests with stream validation on. For reference, the conda-cpp-tests jobs take 7-10 minutes from what I see on other PRs, and that's the same in this PR.

PR #12715, opened 2023-02-07
Allow setting the seed argument for hash partition

firestarman:
Signed-off-by: Liangcai Li <firestarmanllc@gmail.com>

## Description
This PR is exposing the `seed` parameter for the JNI hash partition APIs to support customizing the hash algorithm seed.

The existing tests should cover this change.

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/Table.java
+++ b/java/src/main/java/ai/rapids/cudf/Table.java
@@ -4265,8 +4265,7 @@ public final class Table implements AutoCloseable {
      *             besides IDENTITY and MURMUR3.
      * @param numberOfPartitions number of partitions to use
      * @param seed the seed value for hashing
-     * @return Table that exposes a limited functionality of the
-     * {@link Table} class
+     * @return Table that exposes a limited functionality of the {@link Table} class
      */
     public PartitionedTable hashPartition(HashType type, int numberOfPartitions, int seed) {
       int[] partitionOffsets = new int[numberOfPartitions];
--- a/java/src/main/java/ai/rapids/cudf/Table.java
+++ b/java/src/main/java/ai/rapids/cudf/Table.java
@@ -4265,7 +4265,7 @@ public final class Table implements AutoCloseable {
      *             besides IDENTITY and MURMUR3.
      * @param numberOfPartitions - number of partitions to use
      * @param seed - the seed of hash algorithm
-     * @return {@link PartitionedTable} - Table that exposes a limited functionality of the
+     * @return Table that exposes a limited functionality of the
      * {@link Table} class
      */
     public PartitionedTable hashPartition(HashType type, int numberOfPartitions, int seed) {
--- a/java/src/main/java/ai/rapids/cudf/Table.java
+++ b/java/src/main/java/ai/rapids/cudf/Table.java
@@ -4263,8 +4263,8 @@ public final class Table implements AutoCloseable {
      * @param type the type of hash to use. Depending on the type of hash different restrictions
      *             on the hash column(s) may exist. Not all hash functions are guaranteed to work
      *             besides IDENTITY and MURMUR3.
-     * @param numberOfPartitions - number of partitions to use
-     * @param seed - the seed of hash algorithm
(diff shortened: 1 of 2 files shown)


java/src/main/native/src/TableJni.cpp
```diff
@@ -2655,25 +2655,28 @@ JNIEXPORT jlongArray JNICALL Java_ai_rapids_cudf_Table_partition(JNIEnv *env, jc
...
     JNIEnv *env, jclass, jlong input_table, jintArray columns_to_hash, jint hash_function,
-    jint number_of_partitions, jintArray output_offsets) {
+    jint number_of_partitions, jint seed, jintArray output_offsets) {
 
   JNI_NULL_CHECK(env, input_table, "input table is null", NULL);
   JNI_NULL_CHECK(env, columns_to_hash, "columns_to_hash is null", NULL);
   JNI_NULL_CHECK(env, output_offsets, "output_offsets is null", NULL);
   JNI_ARG_CHECK(env, number_of_partitions > 0, "number_of_partitions is zero", NULL);
+  // For simplicity of converting a signed int to a unsigned int
+  JNI_ARG_CHECK(env, seed >= 0, "seed is negative", NULL);
```
jlowe, line 2665: Not sure why this check is here.  Java doesn't have unsigned ints, so this eliminates half of the potential seeds we could use.  It's also not documented that passing a negative seed is an error.

Personally, I would just static cast the signed int into an unsigned int and pass it on, since then we get access to the full range of available seeds in the underlying algorithm.

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -4253,12 +4254,28 @@ public PartitionedTable hashPartition(int numberOfPartitions) {
...
+      return hashPartition(type, numberOfPartitions, DEFAULT_HASH_SEED);
+    }
+
+    /**
+     * Hash partition a table into the specified number of partitions.
+     * @param type the type of hash to use. Depending on the type of hash different restrictions
+     *             on the hash column(s) may exist. Not all hash functions are guaranteed to work
+     *             besides IDENTITY and MURMUR3.
+     * @param numberOfPartitions - number of partitions to use
+     * @param seed - the seed of hash algorithm
```
ttnghia, line 4267:
```suggestion
     * @param numberOfPartitions number of partitions to use
     * @param seed the seed value for hashing
```

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -4253,12 +4254,28 @@ public PartitionedTable hashPartition(int numberOfPartitions) {
...
+     * @return {@link PartitionedTable} - Table that exposes a limited functionality of the
```
ttnghia, line 4268:
```suggestion
     * @return Table that exposes a limited functionality of the
```

java/src/main/native/src/TableJni.cpp
ttnghia, line 2665: Where does `seed` come from? The input is `jint_seed` isn't it?

firestarman: Nice suggestion, updated

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -4253,12 +4254,28 @@ public PartitionedTable hashPartition(int numberOfPartitions) {
...
+
+    /**
+     * Hash partition a table into the specified number of partitions.
+     * @param type the type of hash to use. Depending on the type of hash different restrictions
+     *             on the hash column(s) may exist. Not all hash functions are guaranteed to work
+     *             besides IDENTITY and MURMUR3.
+     * @param numberOfPartitions number of partitions to use
+     * @param seed the seed value for hashing
+     * @return Table that exposes a limited functionality of the
+     * {@link Table} class
```
ttnghia, line 4269:
```suggestion
     * @return Table that exposes a limited functionality of the {@link Table} class
```

ttnghia (approved):

jlowe (approved):

PR #12722, opened 2023-02-07
Fix memcheck read error in compound segmented reduce

davidwendt:
## Description
Fixes an out-of-bounds memory read in the compound segmented reduction logic. The number of segments was computed incorrectly causing an extra read passed the end of the valid-counts vector. This was found by running compute-sanitizer test on the reductions gtests as follows:
```
compute-sanitizer --tool memcheck --demangle full gtests/REDUCTIONS_TEST --rmm_mode=cuda
```
The number of segments is 1 less than the number of offsets.

## Checklist


bdice (approved):
Thanks for the fix!

vyasr (approved):

PR #12727, opened 2023-02-07
Add `always_nullable` flag to Dremel encoding

### LINKED ISSUE #12389: [BUG] `lower_bound` not working properly with list columns containing `null` entries
**Describe the bug**
`lower_bound` produces incorrect results for list columns containing `null` entries.

Here is a reproducible test in cudf:
```
{
  using T = TypeParam;
  using lcw = cudf::test::lists_column_wrapper<T, int32_t>;
  using cudf::test::iterators::nulls_at;
  lcw col1{
    lcw{{0}, nulls_at({0})},      // 0
    lcw{-80},                     // 1
    lcw{-17},                     // 2
  };

  lcw col2{
    lcw{27},                      // 0
    lcw{{0}, nulls_at({0})},      // 1
    lcw{},                        // 2
  };

  lcw val1{
    lcw{87},
  };

  lcw val2{
    lcw{},
  };

  cudf::table_view input{{col1, col2}};
  cudf::table_view values{{val1, val2}};
  std::vector<cudf::order> column_order{cudf::order::ASCENDING, cudf::order::DESCENDING};
  std::vector<cudf::null_order> null_order_flags{cudf::null_order::BEFORE, cudf::null_order::BEFORE};
  
  auto expect = cudf::test::fixed_width_column_wrapper<cudf::size_type>{3};
  auto result = cudf::lower_bound(input, values, column_order, null_order_flags);
  CUDF_TEST_EXPECT_COLUMNS_EQUAL(expect, *result);
}
```

The test fails with output:
`first difference: lhs[0] = 3, rhs[0] = 1`

In other words, `([87], [])` should be in Partition 3, but instead was in Partition 1.

Note that if we remove the `null` entries, the test would pass, as shown below.
```
{
  using T = TypeParam;
  using lcw = cudf::test::lists_column_wrapper<T, int32_t>;
  using cudf::test::iterators::nulls_at;
  lcw col1{
    lcw{},                        // 0
    lcw{-80},                     // 1
    lcw{-17},                     // 2
  };

  lcw col2{
    lcw{27},                      // 0
    lcw{},                        // 1
    lcw{},                        // 2
  };

  lcw val1{
    lcw{87},
  };

  lcw val2{
    lcw{},
  };

  cudf::table_view input{{col1, col2}};
  cudf::table_view values{{val1, val2}};
  std::vector<cudf::order> column_order{cudf::order::ASCENDING, cudf::order::DESCENDING};
  std::vector<cudf::null_order> null_order_flags{cudf::null_order::BEFORE, cudf::null_order::BEFORE};
  
  auto expect = cudf::test::fixed_width_column_wrapper<cudf::size_type>{3};
  auto result = cudf::lower_bound(input, values, column_order, null_order_flags);
  CUDF_TEST_EXPECT_COLUMNS_EQUAL(expect, *result);
}
```

divyegala:
Closes #12389 by fixing the bug describe here https://github.com/rapidsai/cudf/issues/12389#issuecomment-1419949751.

This flag, when `always_nullable=true`, generates `definition levels` in the Dremel encoding such that it considers every nested column and child to be `nullable`, even if they actually are not. In the context of `two_table_comparators`, this helps us with producing consistently mapped `definition levels` in case there are some nested columns or children that are not nullable in either one or both of the tables.

This PR now exposes two APIs:

1. `cudf::detail::get_dremel_data(...)` : This API is consistent with standard Dremel encoding
2. `cudf::detail::get_comparator_data(...)` : This API modifies the definition levels in Dremel encoding to produce the effect described above

## Checklist


cindyyuanjiang:
Tested this PR against integration tests in https://github.com/NVIDIA/spark-rapids/pull/7233 and the current tests passed.

cpp/include/cudf/lists/detail/dremel.hpp
```diff
@@ -195,4 +195,20 @@ dremel_data get_dremel_data(column_view h_col,
                             bool output_as_byte_array,
                             rmm::cuda_stream_view stream);
 
+/**
+ * @brief Get Dremel offsets, repetition levels, and modified definition levels to be used for
+ *        lexicographical comparators. The modified definition levels are produced by treating
+ *        each nested column in the input as nullable
+ *
+ * @param h_col Column of LIST type
```
ttnghia, line 203: What does `h_col` mean? We typically use a simple convention `input` var name if there is just one input column to a function.

cpp/include/cudf/lists/detail/dremel.hpp
```diff
@@ -195,4 +195,20 @@ dremel_data get_dremel_data(column_view h_col,
...
+ * @param nullability Pre-determined nullability at each list level. Empty means infer from
+ * `col`
```
ttnghia, line 205: "infer from `col`" while there is no "col" var :smile:

cpp/include/cudf/lists/detail/dremel.hpp
```diff
@@ -195,4 +195,20 @@ dremel_data get_dremel_data(column_view h_col,
...
+ * @param output_as_byte_array flag to indicate if output should be a byte array
```
ttnghia, line 206: "if output should be a byte array" but what otherwise?

cpp/include/cudf/lists/detail/dremel.hpp
divyegala: I suppose it means host column, I just maintained the current naming conventions as they were. I can change it to `col` everywhere

cpp/include/cudf/lists/detail/dremel.hpp
divyegala: Good question. I will have to review the code for what this flag does, unfortunately the current version does not document this flag either

cpp/include/cudf/lists/detail/dremel.hpp
```diff
@@ -195,4 +195,20 @@ dremel_data get_dremel_data(column_view h_col,
...
+ * @param stream CUDA stream used for device memory operations and kernel launches.
+ * @return A struct containing dremel data
+ */
+dremel_data get_comparator_data(column_view h_col,
```
ttnghia, line 210: This also returns a `dremel_data` object, similar to `get_dremel_data`. Thus I suggest to call it `get_comparator_dremel_data`.

cpp/include/cudf/lists/detail/dremel.hpp
ttnghia: Humnn, since the existing function calls it `h_col` so I'm fine to keep this name.

cpp/src/table/row_operators.cu
```diff
@@ -264,7 +264,7 @@ auto list_lex_preprocess(table_view table, rmm::cuda_stream_view stream)
   std::vector<detail::dremel_device_view> dremel_device_views;
   for (auto const& col : table) {
     if (col.type().id() == type_id::LIST) {
-      dremel_data.push_back(detail::get_dremel_data(col, {}, false, stream));
+      dremel_data.push_back(detail::get_comparator_data(col, {}, false, stream));
```
ttnghia, line 267:
```suggestion
      dremel_data.emplace_back(detail::get_comparator_data(col, {}, false, stream));
```

ttnghia: Also recommend to replace `push_back` by `emplace_back` in many other places, such as the immediate line below.

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
   CUDF_TEST_EXPECT_COLUMNS_EQUAL(expected, *result, verbosity);
 }
+
+struct ListLowerBound : public cudf::test::BaseFixture {
+};
+
+TEST_F(ListLowerBound, ListWithNulls)
+{
+  {
+    using lcw = cudf::test::lists_column_wrapper<double>;
```
ttnghia, line 357: For consistency, please add this line:
```
using doubles_col  = cudf::test::fixed_width_column_wrapper<double>;
```
Into here: https://github.com/rapidsai/cudf/blob/681039ae9e662dd160b91d5db083515357f0f583/cpp/tests/search/search_list_test.cpp#L31

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
+    using cudf::test::iterators::nulls_at;
```
ttnghia, line 358: This is already imported at https://github.com/rapidsai/cudf/blob/681039ae9e662dd160b91d5db083515357f0f583/cpp/tests/search/search_list_test.cpp#L28.
```suggestion
```

cpp/src/table/row_operators.cu
divyegala: I do not think `emplace_back` will work here. `detail::get_comparator_data` returns `detail::dremel_data&&`. `emplace_back` will instead require `Args&&.. args` which can be used to construct a `detail::dremel_data`

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
+    lcw col1{
+      lcw{-3.45967821e+12},  // 0
+      lcw{-3.6912186e-32},   // 1
+      lcw{9.721175},         // 2
+    };
+
+    lcw val1{
```
ttnghia, line 365: We don't have `col2` thus let's not use the number in var name.
```suggestion
    auto const haystack = lcw{
      lcw{-3.45967821e+12},  // 0
      lcw{-3.6912186e-32},   // 1
      lcw{9.721175},         // 2
    };

    auto const needles = lcw{
```

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
+      lcw{{0, 4.22671e+32}, nulls_at({0})},
+    };
+
+    cudf::table_view input{{col1}};
+    cudf::table_view values{{val1}};
+    std::vector<cudf::order> column_order{cudf::order::ASCENDING};
+    std::vector<cudf::null_order> null_order_flags{cudf::null_order::BEFORE};
+
+    auto expect = cudf::test::fixed_width_column_wrapper<cudf::size_type>{0};
+    auto result = cudf::lower_bound(input, values, column_order, null_order_flags);
```
ttnghia, line 375:
```suggestion
    auto const expect = int32s_col{0};
    auto const result = cudf::lower_bound(cudf::table_view{{haystack}}, cudf::table_view{{needles}}, {cudf::order::ASCENDING}, {cudf::null_order::BEFORE});
```

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
+    using lcw = cudf::test::lists_column_wrapper<double>;
+    using cudf::test::iterators::nulls_at;
+    lcw col1{
+      lcw{-3.45967821e+12},  // 0
+      lcw{-3.6912186e-32},   // 1
+      lcw{9.721175},         // 2
+    };
+
+    lcw val1{
+      lcw{{0, 4.22671e+32}, nulls_at({0})},
```
ttnghia, line 366:
```suggestion
      lcw{{0, 4.22671e+32}, null_at(0)},
```

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
+    std::vector<cudf::null_order> null_order_flags{cudf::null_order::BEFORE};
+
+    auto expect = cudf::test::fixed_width_column_wrapper<cudf::size_type>{0};
+    auto result = cudf::lower_bound(input, values, column_order, null_order_flags);
+    CUDF_TEST_EXPECT_COLUMNS_EQUAL(expect, *result);
+  }
+
+  {
+    using lcw = cudf::test::lists_column_wrapper<int32_t, int32_t>;
+    using cudf::test::iterators::nulls_at;
```
ttnghia, line 381:
```suggestion
```

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
+    lcw col1{
```
ttnghia, line 382: This and similar below:
```suggestion
    auto const col1 = lcw{
```

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
+      lcw{{0}, nulls_at({0})},  // 0
```
ttnghia, line 383: This and similar below:
```suggestion
      lcw{{0}, null_at(0)},  // 0
```

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
+      lcw{},
+    };
+
+    cudf::table_view input{{col1, col2}};
+    cudf::table_view values{{val1, val2}};
+    std::vector<cudf::order> column_order{cudf::order::ASCENDING, cudf::order::DESCENDING};
+    std::vector<cudf::null_order> null_order_flags{cudf::null_order::BEFORE,
+                                                   cudf::null_order::BEFORE};
+
+    auto expect = cudf::test::fixed_width_column_wrapper<cudf::size_type>{3};
```
ttnghia, line 408:
```suggestion
    auto const expect = int32s_col{3};
```

cpp/tests/search/search_list_test.cpp
```diff
@@ -347,3 +347,66 @@ TYPED_TEST(TypedListContainsTestColumnNeedles, ListsOfStructs)
...
+    auto result = cudf::lower_bound(input, values, column_order, null_order_flags);
```
ttnghia, line 409:
```suggestion
    auto const result = cudf::lower_bound(input, values, column_order, null_order_flags);
```

cpp/include/cudf/lists/detail/dremel.hpp
bdice: I vote to rename to `input`. I like that convention for libcudf.

cpp/include/cudf/lists/detail/dremel.hpp
bdice: @vyasr seemed to feel strongly that this is _not_ "dremel data" -- just wanted to solicit his input here.

cpp/tests/search/search_list_test.cpp
bdice: I think it's cleaner to use only specific names rather than the whole namespace, but I wouldn't invest a lot of time on this.

divyegala: I will update the existing line to only promote specific names instead of the whole namespace

cpp/src/table/row_operators.cu
ttnghia: You can call `vector<T>::emplace_back(T&&)` which does construct-in-place. Unless your `dremel_data` class is not movable but I don't think so since it is still copied via `push_back`.

divyegala: `push_back` does not create a copy for an r-value reference https://en.cppreference.com/w/cpp/container/vector/push_back

cpp/tests/search/search_list_test.cpp
```diff
@@ -366,22 +367,25 @@ TEST_F(ListLowerBound, ListWithNulls)
...
-      lcw{-17},                 // 2
+      lcw{-80},              // 1
+      lcw{-17},              // 2
     };
 
     lcw col2{
-      lcw{27},                  // 0
-      lcw{{0}, nulls_at({0})},  // 1
-      lcw{},                    // 2
+      lcw{27},               // 0
```
ttnghia, line 386:
```
auto const col2 = ...
```

PointKernel (approved):

cpp/src/lists/dremel.cu
```diff
@@ -46,6 +46,7 @@ struct def_level_fn {
   uint8_t const* d_nullability;
   uint8_t sub_level_start;
   uint8_t curr_def_level;
+  bool always_nullable;
```
bdice, line 49:
```suggestion
  bool const always_nullable;
```

bdice (approved):
Looks good!

Issue #123, opened 2018-03-26
DataFrame becomes empty when a column with empty array/series added.


wamsiv:
The DataFrame becomes empty when a column with no values being added. But the index of the dataframe remains same.

```>>> df = DataFrame()
>>> df['a'] = [1,2,3]
>>> df
     a
0    1
1    2
2    3
>>> df['b'] = []
>>> df
Empty DataFrame```

## Expected
`df` should return

```>>> df
     a    b
0    1    nan
1    2    nan
2    3    nan```

**Comment by kkraus14** [comment]
@wamsiv I believe this is the same issue as #110. Can you check the length of df['a'] after setting df['b']. I believe what's happening is when asking for df it only returns the index up to the smallest column, which in your case is a zero length df['b']. I believe your PR request can be more generalized away from just NaNs to setting constants in general.

**Comment by wamsiv** [comment]
@kkraus14 The above issue with index you are mentioning is fixed in the PR. Here's a snip:
```>>> df = DataFrame()
>>> df['a'] = [1,2,3]
>>> df['b'] = []
>>> df
     a    b
0    1  nan
1    2  nan
2    3  nan
>>> len(df)
3
```

Regarding #110. Just pushed a commit with the fix.

```>>> pdf = pd.DataFrame({"key": [1,1,1,2,2,2,3,3,3,4,4,4], "value": [1,2,3,4,5,6,7,8,9,10,11,12]})
>>> gdf = DataFrame.from_pandas(pdf)
>>> gdf['newcol'] = 1010
>>> gdf
    key value newcol
 0    1     1   1010
 1    1     2   1010
 2    1     3   1010
 3    2     4   1010
 4    2     5   1010
 5    2     6   1010
 6    3     7   1010
 7    3     8   1010
 8    3     9   1010
 9    4    10   1010
10    4    11   1010
11    4    12   1010
```

sklam:
It looks like this is not a supported operation in pandas

```python
In [1]: import pandas

In [2]: df = pandas.DataFrame()

In [3]: df['a'] = [1, 2, 3]

In [4]: df['b'] = []
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-c5ad7457a06d> in <module>()
----> 1 df['b'] = []

~/miniconda3/envs/dev-client-cudf-pygdf-py36/lib/python3.6/site-packages/pandas/core/frame.py in __setitem__(self, key, value)
   2517         else:
[...]

PR #12728, opened 2023-02-07
Add nvbench environment class for initializing RMM in benchmarks

davidwendt:
## Description
Adds an environment class to initialize the RMM pool memory resource manager before running benchmarks through nvbench. This removes the need to initialize RMM on every benchmark call which improves benchmark run performance and fixes some GPU metrics gathering when run under nsys.

This requires a patch to the nvbench source to enable this feature and is part of the following pull request:  https://github.com/NVIDIA/nvbench/pull/123
The patch can be removed once the PR is merged and source available to the libcudf build.

## Checklist


vyasr:
@davidwendt did you want to merge this PR and revert the patch once the nvbench PR is merged, or wait for the nvbench PR? I'm fine either way; I know that the dev cycle on nvbench PRs is slower than ours.

cpp/benchmarks/fixture/rmm_pool_raii.hpp
```diff
@@ -69,4 +69,15 @@ class rmm_pool_raii {
...
+ *
+ * Initializes the default memory resource to use the RMM pool device resource.
+ */
+struct nvbench_base_fixture {
+  rmm_pool_raii _mr;
+};
+
 }  // namespace cudf
+
+#define NVBENCH_ENVIRONMENT cudf::nvbench_base_fixture
```
vyasr, line 83: Could you use the `rmm_pool_raii` directly here? If the goal is to restrict the scope to more nvbench-specific things, the alternative question is, does `rmm_pool_raii` have any other purpose or could we just rename it to `nvbench_base_fixture`?

vyasr: As an aside, while we're making this change should we change the namespace? Our GTest fixture is in cudf::test, maybe this object should live in cudf::benchmark.

davidwendt: I wanted to just wrap in case we wanted add some other specific logic. I'd assume it had some other purpose. Regardless I felt it was clearer in its desired intent in the current implementation.
There is no `cudf::benchmark` namespace. I think creating the namespace would be out of scope for this PR.

davidwendt:
> @davidwendt did you want to merge this PR and revert the patch once the nvbench PR is merged, or wait for the nvbench PR? I'm fine either way; I know that the dev cycle on nvbench PRs is slower than ours.

I did not want to wait for the nvbench PR to be merged first.

ttnghia (approved):
So does this mean this PR depends on https://github.com/NVIDIA/nvbench/pull/123? Look like that #123 is ready to merge.

bdice (approved):
Approved. Let’s try to note a “fixed in” version if we know what we might expect.

cpp/cmake/thirdparty/patches/thrust_override.json
```diff
@@ -32,6 +32,11 @@
           "file" : "${current_json_dir}/cub_segmented_sort_with_bool_key.diff",
           "issue" : "Fix an error in CUB DeviceSegmentedSort when the keys are bool type [https://github.com/NVIDIA/cub/issues/594]",
           "fixed_in" : "2.1"
+        },
+        {
+          "file" : "${current_json_dir}/nvbench_global_setup.diff",
+          "issue" : "Fix add support for global setup to initialize RMM in nvbench [https://github.com/NVIDIA/nvbench/pull/123]",
+          "fixed_in" : ""
```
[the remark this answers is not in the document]
davidwendt: Not until it is merged and that does not look to be anytime soon.

cpp/cmake/thirdparty/patches/thrust_override.json
```diff
@@ -32,6 +32,11 @@
           "file" : "${current_json_dir}/cub_segmented_sort_with_bool_key.diff",
           "issue" : "Fix an error in CUB DeviceSegmentedSort when the keys are bool type [https://github.com/NVIDIA/cub/issues/594]",
           "fixed_in" : "2.1"
+        },
+        {
```
robertmaynard, line 36: This shouldn't be in the Thrust patches override file as it will be applied when Thrust is downloaded and not when nvbench is.

We will also want to make a second file ( nvbench_override ) and have just the new entry in it:

```
{
  "packages" : {
    "nvbench" : {
      "patches" : [        
        {
          "file" : "${current_json_dir}/nvbench_global_setup.diff",
          "issue" : "Fix add support for global setup to initialize RMM in nvbench [https://github.com/NVIDIA/nvbench/pull/123]",
          "fixed_in" : ""
        }
      ]
    }
  }
}
```

We will want to add a `get_nvbench.cmake` and have it do the following logic before call `rapids_cpm_nvbench`:
```cmake
  include(${rapids-cmake-dir}/cpm/nvbench.cmake)
  include(${rapids-cmake-dir}/cpm/package_override.cmake)

  set(cudf_patch_dir "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/patches")
  rapids_cpm_package_override("${cudf_patch_dir}/nvbench_override.json")
```

robertmaynard (changes requested):

cpp/benchmarks/fixture/rmm_pool_raii.hpp
```diff
@@ -69,4 +69,15 @@ class rmm_pool_raii {
...
 };
 
+/**
+ * Base fixture for cudf benchmarks using nvbench.
+ *
+ * Initializes the default memory resource to use the RMM pool device resource.
+ */
+struct nvbench_base_fixture {
+  rmm_pool_raii _mr;
+};
```
karthikeyann, line 79: Is it possible to inherit this from existing `NVBENCH_ENVIRONMENT` ?
This would allow global fixture to be layered.

cpp/CMakeLists.txt
```diff
@@ -788,8 +788,7 @@ if(CUDF_BUILD_BENCHMARKS)
   # do not support the latest version of fmt, which is automatically pulled into our conda
```
robertmaynard, line 788: Missed this the first time, can you move the following over to the function inside `get_nvbench`:
```cmake
  # do not support the latest version of fmt, which is automatically pulled into our conda
  # environments by mamba.
  set(CPM_DOWNLOAD_fmt TRUE)
```

cpp/cmake/thirdparty/patches/nvbench_override.json
```diff
@@ -0,0 +1,14 @@
+
+{
+  "packages" : {
+    "Thrust" : {
```
robertmaynard, line 4:
```suggestion
    "nvbench" : {
```

robertmaynard (changes requested):
CMake changes lgtm

cpp/benchmarks/fixture/rmm_pool_raii.hpp
davidwendt: This would be a problem since the `NVBENCH_ENVIRONMENT` needs to be set before including `nvbench.cuh` so we'd get a circular dependency. I'd rather nvbench support layering first and then use that.
Just curious but what is the use case for layering?

karthikeyann:
```suggestion
#ifndef NVBENCH_ENVIRONMENT
struct nvbench_base_fixture {
  rmm_pool_raii _mr;
};
#else
struct nvbench_base_fixture : NVBENCH_ENVIRONMENT {
  rmm_pool_raii _mr;
};
#undef NVBENCH_ENVIRONMENT
#endif
```

davidwendt: I do not agree with this change for the reasons given previously.

robertmaynard (approved):
CMake changes are now great :+1:

karthikeyann (approved):
LGTM 👍

PR #12729, opened 2023-02-08
Rework logic in cudf::strings::split_record to improve performance

### LINKED ISSUE #12694: [BUG] cudf::strings::split_record can be over 15x slower than a single thread on the CPU for some cases
**Describe the bug**
We have a customer that is complaining about a specific query that is taking a crazy long time to complete compared to the time that the CPU takes (orders of magnitude slower)

I did some benchmarks with synthetic data (happy to upload some or all of it, but the 1% of the data that really shows the problem is 150 MiB compressed). The data involves a lot of really long strings. We saw that for most operators, like reading parquet and even doing a regular expression replace on the long strings, the GPU was only 3x to 5x slower than a 12 thread 6 core CPU doing the same thing in java/scala.  But for `cudf::strings:split_record` it was a massively huge change. 15x slower than a single thread and over 238x slower than the 12 thread 6 core CPU.

I did some profiling with nsight systems and the slowness appears to come from the kernel that counts the number of matches per-input string and also the kernel that outputs the pointer and length of the matched parts.

The other thing that I noticed is that these kernels appear to take the entire GPU despite the size of the grid being so small that it should not (nsight says that it would in theory, but there are only a few thousand strings in the input data and the a6000 has a huge number of threads). On the larger full data set if I try to run with multiple threads it takes as long as running all of the data with a single thread, assuming the same number of batches. 

First kernel
```
_kernel_agent
Begins: 6.67374s
Ends: 136.101s (+129.427 s)
grid:  <<<5, 1, 1>>>
block: <<<256, 1, 1>>>
Launch Type: Regular
Static Shared Memory: 0 bytes
Dynamic Shared Memory: 0 bytes
Registers Per Thread: 38
Local Memory Per Thread: 0 bytes
Local Memory Total: 74,317,824 bytes
Shared Memory executed: 8,192 bytes
Shared Memory Bank Size: 4 B
Theoretical occupancy: 100 %
Launched from thread: 588148
Latency: ←12.216 μs
Correlation ID: 1229
Stream: Stream 13
```

Second Kernel
```
_kernel_agent
Begins: 136.101s
Ends: 348.364s (+212.263 s)
grid:  <<<5, 1, 1>>>
block: <<<256, 1, 1>>>
Launch Type: Regular
Static Shared Memory: 0 bytes
Dynamic Shared Memory: 0 bytes
Registers Per Thread: 39
Local Memory Per Thread: 0 bytes
Local Memory Total: 74,317,824 bytes
Shared Memory executed: 8,192 bytes
Shared Memory Bank Size: 4 B
Theoretical occupancy: 100 %
Launched from thread: 588148
Latency: ←11.004 μs
Correlation ID: 1289
Stream: Stream 13
```

For reference here are some kernels doing a replace_re on the same data
```
for_each_kernel
Begins: 31.0558s
Ends: 33.1574s (+2.102 s)
grid:  <<<10, 1, 1>>>
block: <<<256, 1, 1>>>
Launch Type: Regular
Static Shared Memory: 0 bytes
Dynamic Shared Memory: 192 bytes
Registers Per Thread: 79
Local Memory Per Thread: 0 bytes
Local Memory Total: 85,327,872 bytes
Shared Memory executed: 8,192 bytes
Shared Memory Bank Size: 4 B
Theoretical occupancy: 50 %
Launched from thread: 498744
Latency: ←20.579 μs
Correlation ID: 28726
Stream: Stream 21
```

```
for_each_kernel
Begins: 33.1575s
Ends: 35.384s (+2.227 s)
grid:  <<<10, 1, 1>>>
block: <<<256, 1, 1>>>
Launch Type: Regular
Static Shared Memory: 0 bytes
Dynamic Shared Memory: 192 bytes
Registers Per Thread: 79
Local Memory Per Thread: 0 bytes
Local Memory Total: 85,327,872 bytes
Shared Memory executed: 8,192 bytes
Shared Memory Bank Size: 4 B
Theoretical occupancy: 50 %
Launched from thread: 498744
Latency: ←21.218 μs
Correlation ID: 28775
Stream: Stream 21
```

4.329 seconds for regex_replace vs 341+ seconds for split_records. (79x slower)

```
  auto file_name = "/data/tmp/DUMP_2.parquet";
  auto source = cudf::io::source_info(file_name);
  auto builder = cudf::io::parquet_reader_options::builder(source);
  builder.columns({"value"});
  auto read = read_parquet(builder.build());
  std::cerr << "READ IN DATA" << std::endl;
  auto const strs_input = cudf::strings_column_view{read.tbl->get_column(0)};
  cudf::string_scalar sp("},{");
  std::cerr << "SETUP DONE... (RUNNING)" << std::endl;
  auto ret = cudf::strings::split_record(
          strs_input,
          sp,
          -1);
  std::cerr << "ALL DONE" << std::endl;
```

Like I said I am happy to upload DUMP_2.parquet for anyone who wants to work on it. (it is too large for github).

**Expected behavior**
I would love it if we could fix this so it was faster than the CPU by a significant amount, but I am happy if it just matches the other long string expressions where it is only 5x slower than the entire CPU (i.e. make it at least 50x faster than it is today)

davidwendt:
## Description
Updates the `cudf::strings::split_record` logic to match the more optimized code in `cudf::strings:split`.
The optimized code performs much better for longer strings (>64 bytes) by parallelizing over the character bytes to find delimiters before determining split tokens. 
This led to refactoring the code so it both APIs can share the optimized code.
Also fixes a bug found when using overlapped delimiters.
Additional tests were added for multi-byte delimiters which can overlap and span multiple adjacent strings.

Closes #12694

## Checklist


davidwendt:
Performance numbers for `cudf::strings::split_record`
```
/record/4096/32          0.186         0.262      0.71x
/record/4096/64          0.254         0.270      0.94x
/record/4096/128         0.408         0.281      1.45x
/record/4096/256         0.845         0.307      2.75x
/record/4096/512          2.25         0.280      8.04x
/record/4096/1024         7.89         0.307     25.70x
/record/4096/2048         29.3         0.376     77.93x
/record/4096/4096          115         0.498    230.92x
/record/4096/8192          464         0.752    617.02x
/record/32768/32         0.204         0.274      0.74x
/record/32768/64         0.279         0.297      0.94x
/record/32768/128        0.506         0.349      1.45x
/record/32768/256         1.05         0.444      2.36x
/record/32768/512         2.71         0.413      6.56x
/record/32768/1024        8.36         0.561     14.90x
/record/32768/2048        36.5         0.861     42.39x
/record/32768/4096         164          1.46    112.33x
/record/32768/8192         706          2.70    261.48x
/record/262144/32        0.328         0.419      0.78x
/record/262144/64        0.576         0.581      0.99x
/record/262144/128        1.47          1.16      1.27x
/record/262144/256        5.65          3.62      1.56x
/record/262144/512        12.3          1.51      8.15x
/record/262144/1024       48.3          2.71     17.82x
/record/262144/2048        216          5.59     38.64x
/record/262144/4096       1066          11.8     90.34x
/record/2097152/32        1.24          1.47      0.84x
/record/2097152/64        2.85          2.73      1.04x
/record/2097152/128       8.27          6.51      1.27x
/record/2097152/256       31.9          23.1      1.38x
/record/2097152/512       55.6          10.5      5.30x
/record/16777216/32       8.79          10.1      0.87x
/record/16777216/64       21.3          20.2      1.05x
```

cpp/src/strings/split/split_record.cu
```diff
@@ -23,123 +25,56 @@
...
+                                        rmm::cuda_stream_view stream,
+                                        rmm::mr::device_memory_resource* mr)
+{
+  if (input.is_empty()) { return make_empty_column(type_id::LIST); }
+  if (input.size() == input.null_count()) {
+    auto offsets = std::make_unique<column>(input.offsets(), stream, mr);
+    auto results = std::make_unique<column>(input.parent(), stream, mr);
+    return make_lists_column(input.size(),
+                             std::move(offsets),
+                             std::move(results),
```
ttnghia, line 55: Just to make sure I understand correctly? The output lists column will be lists of strings (all nulls)?

ttnghia: I ask this because if we want to have an output lists column of all null, we just need to have an empty child, instead of all-nulls child.

cpp/src/strings/split/split_record.cu
```diff
@@ -23,123 +25,56 @@
...
-        d_result[token_count - 1 - token_idx] = token;
-        end_pos                               = delimiter_pos;
-      }
-      token_idx++;
-    }
+  return make_lists_column(input.size(),
+                           std::move(offsets),
+                           std::move(strings_child),
+                           input.null_count(),
+                           copy_bitmask(input.parent(), stream, mr),
```
ttnghia, line 72: Again, making a lists column with bitmask copied from the input. I'm not sure if this will generate non-empty nulls.

cpp/tests/strings/split_tests.cpp
```diff
@@ -308,6 +308,31 @@ TEST_F(StringsSplitTest, SplitRecordWhitespaceWithMaxSplit)
...
+  // includes overlapping delimiters and delimiters that span across adjacent strings
+  auto input =
+    cudf::test::strings_column_wrapper({"u__", "w___x", "y____z", "{a=1}:{b=2}:", "{c=3}"});
+  auto view = cudf::strings_column_view(input);
+
+  auto result = cudf::strings::split_record(view, cudf::string_scalar("__"));
+  using LCW   = cudf::test::lists_column_wrapper<cudf::string_view>;
+  auto expected1 =
+    LCW({LCW{"u", ""}, LCW{"w", "_x"}, LCW{"y", "", "z"}, LCW{"{a=1}:{b=2}:"}, LCW{"{c=3}"}});
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(result->view(), expected1);
```
ttnghia, line 322: Recommend putting each test into a block `{}` so you won't have to enumerate variables like `expected1`, `expected2` etc.

davidwendt: I added additional tests for `split/rsplit` as well.

cpp/src/strings/split/split_record.cu
```diff
@@ -49,7 +49,7 @@ std::unique_ptr<column> split_record_fn(strings_column_view const& input,
   if (input.is_empty()) { return make_empty_column(type_id::LIST); }
   if (input.size() == input.null_count()) {
     auto offsets = std::make_unique<column>(input.offsets(), stream, mr);
-    auto results = std::make_unique<column>(input.parent(), stream, mr);
+    auto results = make_empty_column(type_id::STRING);
```
ttnghia, line 52: Sorry, after thinking about this more carefully I see a problem: In case the `input` column has non-empty nulls, this will cause data corruption as the `results` column is empty but `offsets` contains non-zero values copied from `input`. Thus, we should have either:
 * `results` is copied from `input.parent()` as your previous code, or
 * `offsets` is created from scratch with all `0` values.

davidwendt: Right. The offsets would already by all zeroes if every entry in the input was all nulls. 
But the child column can be empty because, by definition, the child data should never be accessed for any null rows.
So technically, the child column could be anything since so no code should ever read it but any data in the child column of an all-null parent column could be considered as unsanitized.

ttnghia (approved):

cpp/src/strings/split/split.cuh
```diff
@@ -0,0 +1,407 @@
...
+
+    // next, count the number of delimiters per string
+    auto d_string_indices = string_indices.data();  // identifies strings with delimiters only
+    thrust::for_each_n(rmm::exec_policy(stream),
+                       thrust::make_counting_iterator<size_type>(0),
+                       delimiter_count,
+                       [d_string_indices, d_delimiter_offsets] __device__(size_type idx) {
+                         auto const str_idx = d_string_indices[idx] - 1;
+                         atomicAdd(d_delimiter_offsets + str_idx, 1);
+                       });
```
nvdbaranec, line 362: Could this just be a `thrust::reduce_by_key`? If I'm reading this right, the inputs are a string index per delimiter, ie:

`0, 0, 1, 1, 1, 3, 3, 3, 5, 5, 6`

A `reduce_by_key` with the keys being the above and the value just a constant_iterator returning 1 should yield the same result without the atomics.

nvdbaranec: Although I guess you'd have to deal with the holes for strings that have no delimiters.

davidwendt: Right. Some of the strings have no delimiters and we need zero for those.

cpp/src/strings/split/split.cuh
```diff
@@ -0,0 +1,407 @@
...
+  auto d_tokens = tokens.data();
+  thrust::for_each_n(
+    rmm::exec_policy(stream),
+    thrust::make_counting_iterator<size_type>(0),
+    strings_count,
+    [tokenizer, d_tokens_offsets, d_positions, d_delimiter_offsets, d_tokens] __device__(
+      size_type idx) {
+      tokenizer.get_tokens(idx, d_tokens_offsets, d_positions, d_delimiter_offsets, d_tokens);
+    });
+
```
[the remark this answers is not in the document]
davidwendt: Actually, I tried that first but it seemed inefficient (GPU utilization) and was a bit slower (though I don't remember by how much now).

PointKernel (approved):
LGTM with some nits (not necessarily a change request though)

nvdbaranec (approved):

PR #12730, opened 2023-02-08
Changing `cudf::io::source_info` to use `cudf::host_span<std::byte>` in a non-breaking form

### LINKED ISSUE #12576: Replace host_buffer with host_span<std::byte const>

hyperbolic2346:
## Description

Closes #12576

This change converts `cudf::io::source_info` to take a `host_span<std::byte const>`. This version deprecates the original API, but leaves it intact to avoid breaking changes. After being deprecated for a few releases, they will be removed.

## Checklist


cpp/include/cudf/io/types.hpp
```diff
@@ -191,21 +208,70 @@ struct source_info {
...
+  {
+  }
+
+  /**
+   * @brief Construct a new source info object for multiple buffers in host memory
+   *
+   * @param host_buffers Input buffers in host memory
+   */
+  template <typename T, CUDF_ENABLE_IF(is_byte_like_type<T>())>
+  explicit source_info(std::vector<cudf::host_span<T const>> const& host_buffers)
```
vuule, line 248: we can use host_span here
```suggestion
  explicit source_info(cudf::host_span<cudf::host_span<T const>> host_buffers)
```

cpp/include/cudf/io/types.hpp
```diff
@@ -191,21 +208,70 @@ struct source_info {
...
+    : _type(io_type::HOST_BUFFER)
+  {
+    if constexpr (not std::is_same_v<std::remove_cv_t<T>, std::byte>) {
+      _host_buffers.resize(host_buffers.size());
```
vuule, line 252: I think this is a typo, since we append elements to _host_buffers after this.
```suggestion
      _host_buffers.reserve(host_buffers.size());
```

cpp/src/io/utilities/datasource.cpp
```diff
@@ -335,6 +335,13 @@ std::unique_ptr<datasource> datasource::create(host_buffer const& buffer)
     reinterpret_cast<const uint8_t*>(buffer.data), buffer.size));
 }
 
+std::unique_ptr<datasource> datasource::create(cudf::host_span<std::byte const> buffer)
```
vuule, line 338: probably not relevant, but `datasource::create(host_buffer const& buffer)` could call this one instead of creating the arrow source directly.

vuule (changes requested):
nice!
Few suggestions, and a potential bug fix.

cpp/include/cudf/io/types.hpp
```diff
@@ -166,6 +168,21 @@ struct host_buffer {
...
+ * @tparam T The representation type
+ * @return `true` if the type is considered a byte-like type
+ */
+template <typename T>
+constexpr inline auto is_byte_like_type()
+{
+  return cuda::std::is_same_v<T, int8_t> || cuda::std::is_same_v<T, char> ||
+         cuda::std::is_same_v<T, uint8_t> || cuda::std::is_same_v<T, unsigned char> ||
+         cuda::std::is_same_v<T, std::byte>;
+}
```
ttnghia, line 184: Can this be replaced completely with `sizeof(T) == 1`?

vuule: I've considered this as well. But, there's no guarantee that a one-byte object directly represents its data. If it's some sort of a pointer/index to actual data, we should not allow it to be treated as raw data. 
Maybe adding `is_integral` to `sizeof(T) == 1` would suffice?

ttnghia: I'm not sure `is_integrate(std::byte)==true`.

ttnghia: Anyway, if we don't have a perfect solution then keeping this check function is fine :smile:

hyperbolic2346: I'd love a perfect solution. The original plan was to accept `std::byte` only, but this produces problems with usage since it can't be implicitly cast. This template was added before we decided to deprecate the original API instead of remove it though. I could be convinced to only support `std::byte` on the new API, but it will require `reinterpret_cast` about everywhere it is used currently.

vuule: I've spent so much time looking for the right type trait ;(
The best I can come up with is `is_trivial` plus sizeof() == 1.

I think I need to experiment a bit with how passing a std::string gets resolved if the filepath overload actually take a std::filepath object. That could impact the decision to support host_span<char>.

cpp/include/cudf/io/types.hpp
hyperbolic2346: Good catch! Done!

ttnghia (approved):

cpp/include/cudf/io/types.hpp
ttnghia: Given that there may be uncertainty with the alternative solution, I'm fine with using the current one :smile:

cpp/include/cudf/io/types.hpp
```diff
@@ -214,7 +214,7 @@ struct source_info {
    */
   explicit source_info(std::vector<host_buffer> const& host_buffers) : _type(io_type::HOST_BUFFER)
   {
-    _host_buffers.resize(host_buffers.size());
+    _host_buffers.reserve(host_buffers.size());
```
ttnghia, line 217: Oh no, this is serious :scream:

hyperbolic2346: yeah, huge difference!

cpp/include/cudf/io/types.hpp
```diff
@@ -191,21 +208,70 @@ struct source_info {
...
+  {
+  }
+
+  /**
+   * @brief Construct a new source info object for multiple buffers in host memory
+   *
+   * @param host_buffers Input buffers in host memory
+   */
+  template <typename T, CUDF_ENABLE_IF(is_byte_like_type<T>())>
+  explicit source_info(cudf::host_span<cudf::host_span<T const>> const& host_buffers)
```
vuule, line 248: I suspect that we cannot take `cudf::host_span<cudf::host_span<char>>` (non const) with this signature.
Maybe something like this would work?
```suggestion
  template <typename T, CUDF_ENABLE_IF(is_byte_like_type<std::remove_cv_t<T>>())>
  explicit source_info(cudf::host_span<cudf::host_span<T>> host_buffers)
```

I suggest adding a test as well.

cpp/include/cudf/io/types.hpp
```diff
@@ -191,21 +208,70 @@ struct source_info {
...
+    : _type(io_type::HOST_BUFFER)
+  {
+    if constexpr (not std::is_same_v<std::remove_cv_t<T>, std::byte>) {
```
vuule, line 251: Is it really okay to remove const/volatile here?
I don't think we can copy cudf::host_span<cudf::host_span<std::byte>>` to `std::vector<cudf::host_span<std::byte const>>`.

Overall, we need a few tests to prove the new API works as intended.

cpp/include/cudf/io/types.hpp
```diff
@@ -191,21 +208,70 @@ struct source_info {
...
+    }
+  }
+
+  /**
+   * @brief Construct a new source info object for a single buffer
+   *
+   * @param host_data Input buffer in host memory
+   */
+  template <typename T, CUDF_ENABLE_IF(is_byte_like_type<T>())>
+  explicit source_info(cudf::host_span<T const> host_data)
```
vuule, line 271: I think we're good here, a non-const span should be implicitly convertible to const span. Still worth a test, IMO.

cpp/include/cudf/io/types.hpp
```diff
@@ -166,6 +168,21 @@ struct host_buffer {
...
+ *
+ * @tparam T The representation type
+ * @return `true` if the type is considered a byte-like type
+ */
+template <typename T>
+constexpr inline auto is_byte_like_type()
+{
+  return cuda::std::is_same_v<T, int8_t> || cuda::std::is_same_v<T, char> ||
+         cuda::std::is_same_v<T, uint8_t> || cuda::std::is_same_v<T, unsigned char> ||
+         cuda::std::is_same_v<T, std::byte>;
```
vuule, line 183: We can make `is_byte_like_type` more robust :)
```suggestion
{
  using non_cv_T = std::remove_cv_t<T>;
  return std::is_same_v<non_cv_T , int8_t> || std::is_same_v<non_cv_T , char> ||
         std::is_same_v<non_cv_T , uint8_t> || std::is_same_v<non_cv_T , unsigned char> ||
         std::is_same_v<non_cv_T , std::byte>;
```

Looks like std library includes `is_same_v` in C++17 onward, we don't need CUDA version.

vuule (changes requested):
few more suggestions. Mostly making sure the new API is as applicable as desired.

cpp/include/cudf/io/types.hpp
hyperbolic2346: Tests added for both new API endpoints. You were correct about span of chars and your suggested change worked well.

cpp/include/cudf/io/types.hpp
hyperbolic2346: My tests have non-const pointers and are working here. Things have shuffled a bit though, so please let me know what you think.

cpp/tests/io/parquet_test.cpp
```diff
@@ -357,6 +357,11 @@ struct ParquetWriterSchemaTest : public ParquetWriterTest {
   auto type() { return cudf::data_type{cudf::type_to_id<T>()}; }
 };
 
+template <typename T>
+struct ParquetReaderSourceTest : public ParquetReaderTest {
+  auto type() { return cudf::data_type{cudf::type_to_id<T>()}; }
```
vuule, line 362: I don't think this is needed, since the type does not refer to the data type in the test.
```suggestion
```

cpp/tests/io/parquet_test.cpp
```diff
@@ -5113,4 +5120,48 @@ TEST_P(ParquetSizedTest, DictionaryTest)
   EXPECT_EQ(nbits, GetParam());
 }
 
+TYPED_TEST(ParquetReaderSourceTest, BufferSourceTypes)
```
vuule, line 5123: should we cover const versions of the byte types here as well?

cpp/include/cudf/io/types.hpp
vuule: it is fine, `std::vector::assign` performs a per-element copy and `host_span<T const>` can be copy assigned/constructed from `host_span<T>`.

vuule (commented):
Thanks for adding the tests! Few more comments.

cpp/tests/io/parquet_test.cpp
hyperbolic2346: I was thinking about this as well this morning. I worry about adding a pile more tests, so I'll probably add it inside to test non-const and const of a type in a single function.

vuule (approved):
🔥 🔥

PR #12738, opened 2023-02-08
Reduce error handling verbosity in CI tests scripts

AjayThorve:
This PR adds a less verbose [trap method](https://github.com/rapidsai/cugraph/blob/f2b081075704aabc789603e14ce552eac3fbe692/ci/test.sh#L19), for error handling to help ensure that we capture all potential error codes in our test scripts, and works as follows:

- setting an environment variable, EXITCODE, with a default value of 0
- setting a trap statement triggered by ERR signals which will set EXITCODE=1 when any commands return a non-zero exit code

cc @ajschmidt8


ajschmidt8 (approved):

bdice (approved):
Great. This is the solution I had in mind.

PR #12740, opened 2023-02-08
Update default data source in cuio reader benchmarks

### LINKED ISSUE #12678: [FEA] Switch `orc_read_decode`, `parquet_read_decode` and `csv_read_input` to use DEVICE_BUFFER data source

The ORC reader benchmark `orc_read_decode` uses a [HOST_BUFFER](https://github.com/rapidsai/cudf/blob/b6dccb31c32eefbf29a7f731b4b5cd7d7f07fe0d/cpp/benchmarks/io/orc/orc_reader_input.cpp#L75) data source, but the Parquet reader benchmark `parquet_read_decode` uses a [FILEPATH](https://github.com/rapidsai/cudf/blob/b6dccb31c32eefbf29a7f731b4b5cd7d7f07fe0d/cpp/benchmarks/io/parquet/parquet_reader_input.cpp#L69) data source. The CSV `BM_csv_read_input` uses a FILEPATH data source.

Let's switch them all to DEVICE_BUFFER data sources.

Let's please also modify these benchmarks to accept an nvbench axis - so we can change the data source at runtime via the nvbench CLI.

I'd like to be consistent and use a DEVICE_BUFFER data source for decode and data type benchmarking.

We considered choosing FILEPATH or HOST_BUFFER for all of these benchmarks, but DEVICE_BUFFER is best for focusing on decoding.

**Additional context**
I wrote a detailed issue full of profiles before realizing the problem. 🤦

PointKernel:
Closes #12678

This PR sets the default data source in orc, parquet and csv reader benchmarks to `DEVICE_BUFFER`.

## Checklist


cpp/benchmarks/io/orc/orc_reader_input.cpp
```diff
@@ -72,7 +72,7 @@ void BM_orc_read_data(nvbench::state& state, nvbench::type_list<nvbench::enum_ty
                         data_profile_builder().cardinality(cardinality).avg_run_length(run_length));
   auto const view = tbl->view();
 
-  cuio_source_sink_pair source_sink(io_type::HOST_BUFFER);
+  cuio_source_sink_pair source_sink(io_type::FILEPATH);
```
vuule, line 75: ooh, I didn't fully comprehend the proposed change until now.
I would argue that HOST_BUFFER is a better choice here (device buffer is the best option IMO) because this set of benchmarks is meant to benchmark only the decode part of the reader (the name is "orc_read_decode"). I understand that this leaves us with a gap in benchmark coverage (different types + IO). maybe that should either be a separate benchmark or an option.

PointKernel: cc @GregoryKimball

ttnghia:
As the issue#12678 says:
> We considered choosing FILEPATH or HOST_BUFFER for all of these benchmarks, but DEVICE_BUFFER is best for focusing on decoding.

vuule (approved):

cpp/benchmarks/io/csv/csv_reader_input.cpp
```diff
@@ -64,13 +64,14 @@ void csv_read_common(DataType const& data_types,
   state.add_buffer_size(source_sink.size(), "encoded_file_size", "encoded_file_size");
 }
 
-template <data_type DataType>
-void BM_csv_read_input(nvbench::state& state, nvbench::type_list<nvbench::enum_type<DataType>>)
+template <data_type DataType, cudf::io::io_type IO>
```
ttnghia, line 67:
```suggestion
template <data_type DataType, cudf::io::io_type IOType>
```

ttnghia: Since we have `DataType` then let's have `IOType` ;)

ttnghia (approved):

PR #12741, opened 2023-02-09
Stop using versioneer to manage versions

vyasr:
## Description

This PR replaces usage of versioneer with hard-coded version numbers in setup.py and __init__.py. Since cudf needs to manage versions across a wide range of file types (CMake, C++, Sphinx and doxygen docs, etc), versioneer cannot be relied on as a single source of truth and therefore does not allow us to single-source our versioning to the Git repo as is intended. Additionally, since the primary means of installing cudf is via conda packages (or now, pip packages), information from the package manager tends to be far more informative than the version strings for troubleshooting and debugging purposes. Conversely, the nonstandard version strings that it produces tend to be problematic for other tools, which at best will ignore such versions but at worst will simply fail.

This PR also replaces usage of an environment variable to set the package name for wheels in setup.py, instead moving the renaming logic into the same sed script used to update package versions. This change makes setup.py essentially static, paving the way for migration to pyproject.toml.

## Checklist


vyasr:
Who doesn't love +53, -12954 :smiley:

python/cudf/cudf/__init__.py
```diff
@@ -113,8 +112,7 @@
 rmm.register_reinitialize_hook(clear_cache)
 
 
-__version__ = get_versions()["version"]
-del get_versions
+__version__ = "23.04.00"
```
wence-, line 115: Out of scope for this PR, but do you think it's possible that there can be _exactly one_ place where this version number lives? I see there are scripts for updating the versions in all possible places, it would be nice if somehow the `setup.py/__init__.py` and `CMakeLists.txt` could programmatically read from (say) `version.cfg` in the root directory.

wence-, line 115: Is this actually the right version for a pre-release of `23.04`?

I have:

```python
from packaging import version
version.parse("23.04.00") < version.parse("23.04")
# => False
version.parse("23.04.00") == version.parse("23.04")
# => True
```

I would have thought we want something like `23.04.a`

vyasr: The value encoded in __init__.py should not be the release version. The version will be updated as part of wheel builds to set the release. We'll probably want to institute something similar for conda builds, but things being out of sync in that respect is the status quo. There are other problems with using true alpha versions (our nightlies do not have such versions), and the version of a conda package as determined by the metadata on conda already doesn't match `__version__` in those packages, so I think all of these are future problems.

python/cudf/setup.py
```diff
@@ -21,9 +16,9 @@
     "typing_extensions",
     # Allow floating minor versions for Arrow.
     "pyarrow==10",
-    f"rmm{cuda_suffix}==23.4.*",
-    f"ptxcompiler{cuda_suffix}",
-    f"cubinlinker{cuda_suffix}",
+    "rmm==23.4.*",
+    "ptxcompiler",
+    "cubinlinker",
```
wence-, line 21: Does this have to go along with changes in those repos (to change the wheel names?)

AKA, what does changing the names of the dependencies have to do with removing versioneer?

python/dask_cudf/setup.py
```diff
@@ -1,19 +1,14 @@
...
-cuda_suffix = os.getenv("RAPIDS_PY_WHEEL_CUDA_SUFFIX", default="")
-
 install_requires = [
     "dask>=2023.1.1",
     "distributed>=2023.1.1",
     "fsspec>=0.6.0",
     "numpy",
     "pandas>=1.0,<1.6.0dev0",
-    f"cudf{cuda_suffix}==23.4.*",
+    "cudf==23.4.*",
```
wence-, line 11: Do we not still need to maintain the wheel suffix? Or has something else changed that means this won't be needed?

wence- (commented):
A little confused by the changing of pip package dependency names, but otherwise 👍

python/cudf/setup.py
vyasr: That's the second half of the PR, which was added after you last reviewed. I combined these changes because it was going to add significant complexity to the CI shared workflow to be able to make the version changes and the cuda suffix changes independently. In order to use the latest workflow, each RAPIDS repo needs to make both changes at once. I agree, it's suboptimal, but the tradeoff is added complexity in the shared workflow that we (me and ops folks) decided wasn't worth it.

python/dask_cudf/setup.py
vyasr: Something else has changed. Specifically, the suffix is now added in the same `apply_wheel_modification.sh` script that also sets the release version used during wheel builds.

python/cudf/setup.py
wence-: Thanks, makes sense.

ci/release/apply_wheel_modifications.sh
```diff
@@ -0,0 +1,35 @@
...
+sed -i "s/ptxcompiler/ptxcompiler${CUDA_SUFFIX}/g" python/cudf/setup.py
+sed -i "s/cubinlinker/cubinlinker${CUDA_SUFFIX}/g" python/cudf/setup.py
+
+# cudf pyproject.toml cuda suffixes
+sed -i "s/rmm/rmm${CUDA_SUFFIX}/g" python/cudf/pyproject.toml
+
+# dask_cudf setup.py cuda suffixes
+sed -i "s/name=\"dask-cudf\"/name=\"dask-cudf${CUDA_SUFFIX}\"/g" python/dask_cudf/setup.py
+# Need to provide the == to avoid modifying the URL
+sed -i "s/\"cudf==/\"cudf${CUDA_SUFFIX}==/g" python/dask_cudf/setup.py
```
wence-, line 35: So for wheel releases, these rename packages to `xxx${CUDA_SUFFIX}` effectively replacing the `os.getenv` overriding of suffix with a sed-based solution.

As the comment on line 34 indicates, this is somewhat fragile since there might be other strings `setup.py` that we _don't_ want to modify that could match. Effectively we're trying to treat setup.py as a templated file and making a template substitution at a certain point.

I wonder if it would be less fragile, although feel free to push back, if the interpolation was done with sed with a halfway house from what we used to have and what we have now.

setup.py
```python

# overridden during wheel releases
cuda_suffix = ""
...
setup(..., requires=[f"rmm{cuda_suffix}", ...])
```

And then sed just runs:
```
sed -i "/cuda_suffix = \"\"/cuda_suffix = \"${CUDA_SUFFIX}\"/" setup.py
```

Which has a smaller chance of hitting false positives. WDYT?

wence- (approved):
I am happy with this, though you may want to consider my suggestions for the apply_wheel_mods script (though the interplay with update-version may be more subtle than I spot).

ci/release/apply_wheel_modifications.sh
vyasr: You are absolutely right. The reason I don't think that this change is worthwhile is that the imminent next step is to move away from setup.py entirely by migrating all of this information into a pyproject.toml file, a la https://github.com/rapidsai/rmm/pull/1151 (and I plan to replicate this change across RAPIDS). This change is part of the longer term plan to move us to more modern builders, but regarding the changes in this PR the effect will be that we will also have a more well-defined config (TOML) file to modify in the future rather than an arbitrarily formatted setup.py file, which should allow us to make more structured modifications in the future.

wence-: Thanks, makes sense. In which case, good to go, although this lost the race with @brandon-b-miller so there are (hopefully trivial) conflicts to fix up :(.

PR #12742, opened 2023-02-09
Add JNI methods for detecting and purging non-empty nulls from LIST and STRUCT

razajafri:
## Description
This PR adds methods for detecting and purging non-empty nulls. 

## Checklist


ttnghia (changes requested):
What is the use case for this? We are working on the cudf side, trying to totally eliminate the non-empty null issue.

razajafri:
> What is the use case for this? We are working on the cudf side, trying to totally eliminate the non-empty null issue.

This is needed because we need to support HashAggregate[List] which cannot be done unless we support sorting with non-empty nulls, so as a work around cudf has these functions which we can use in the interim. For reference checkout https://github.com/NVIDIA/spark-rapids/pull/6679 and perhaps follow the trail from https://github.com/NVIDIA/spark-rapids/issues/6680. 

How far is the team in figuring out the solution for this and when is the fix in cudf expected to go in? As an alternative, and my preference, would be to let this go in for now and revert it when cudf is ready.

ttnghia:
I see the issue with `ColumnView`. Ideally, this (purge non-empty nulls) should only be called in cudf or JNI C++, not in Java. Unfortunately in Java we also have `ColumnView` constructor (and similar) that allow non-empty nulls to exist.

In addition, the PR description doesn't have any context/meaningful description. Please update it.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -4639,6 +4639,12 @@ static native long makeCudfColumnView(int type, int scale, long data, long dataS
 
   static native long applyBooleanMask(long arrayColumnView, long booleanMaskHandle) throws CudfException;
 
+  static native boolean hasNonEmptyNulls(long handle) throws CudfException;
+
+  static native boolean mayHaveNonEmptyNulls(long handle) throws CudfException;
```
ttnghia, line 4644: This API is redundant. From my experience, using it doesn't give any benefit but only performance degradation. Just remove it completely. I'll also do that on the cudf side.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -4997,4 +5003,57 @@ public HostColumnVector copyToHost() {
...
+   * - Worst case: `O(count_descendants(input)) * m`, where `m` is the number of rows in the largest
+   * descendant
+   *
+   * @return true, if either the column or its descendants MIGHT have null rows
+   *         false, if neither the column nor its descendants have null rows
+   */
+  public boolean mayHaveNonEmptyNulls() {
+    assert type.isNestedType() || type.hasOffsets() : "Columns should be a LIST, STRING or STRUCT";
+    return mayHaveNonEmptyNulls(viewHandle);
+  }
```
ttnghia, line 5033: Please remove it.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -4997,4 +5003,57 @@ public HostColumnVector copyToHost() {
...
+   * lists = data: [{{0,1}, {2,3}, {4,5}} validity: {true, false, true}]
+   * lists[1] is null, but the list's child column still stores `{2,3}`.
+   *
+   * After purging the contents of the list's null rows, the column's contents will be:
+   * lists = [data: {{0,1}, {4,5}} validity: {true, false, true}]
+   *
+   * @return A new column with equivalent contents to `input`, but with null rows purged
+   */
+  public ColumnVector purgeNonEmptyNulls() {
+    assert type.isNestedType() || type.hasOffsets() : "Columns should be a LIST, STRING or STRUCT";
```
ttnghia, line 5056: This assert is incorrect. We allow all types.

java/src/main/native/src/ColumnViewJni.cpp
```diff
@@ -2457,4 +2457,46 @@ JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_applyBooleanMask(
...
+JNIEXPORT jboolean JNICALL Java_ai_rapids_cudf_ColumnView_hasNonEmptyNulls(
+    JNIEnv *env, jclass, jlong column_view_handle) {
+  JNI_NULL_CHECK(env, column_view_handle, "column_view handle is null", 0);
+  try {
+    cudf::jni::auto_set_device(env);
+
+    cudf::column_view const *cv =
+        reinterpret_cast<cudf::column_view const *>(column_view_handle);
+
+    return cudf::has_nonempty_nulls(*cv);
```
ttnghia, line 2466:
```suggestion
    auto const cv = reinterpret_cast<cudf::column_view const *>(input_handle);
    return cudf::has_nonempty_nulls(*cv);
```

java/src/main/native/src/ColumnViewJni.cpp
```diff
@@ -2457,4 +2457,46 @@ JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_applyBooleanMask(
...
+  }
+  CATCH_STD(env, 0);
+}
+
+JNIEXPORT jboolean JNICALL Java_ai_rapids_cudf_ColumnView_mayHaveNonEmptyNulls(
+    JNIEnv *env, jclass, jlong column_view_handle) {
```
ttnghia, line 2472:
```suggestion
    JNIEnv *env, jclass, jlong input_handle) {
```

java/src/main/native/src/ColumnViewJni.cpp
```diff
@@ -2457,4 +2457,46 @@ JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_applyBooleanMask(
...
+  try {
+    cudf::jni::auto_set_device(env);
+
+    cudf::column_view const *cv =
+        reinterpret_cast<cudf::column_view const *>(column_view_handle);
+
+    return cudf::may_have_nonempty_nulls(*cv);
+  }
+  CATCH_STD(env, 0);
+}
```
ttnghia, line 2480: Remove this.

java/src/main/native/src/ColumnViewJni.cpp
```diff
@@ -2457,4 +2457,46 @@ JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_applyBooleanMask(
...
+
+JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_purgeNonEmptyNulls(
+    JNIEnv *env, jclass, jlong column_view_handle) {
```
ttnghia, line 2483:
```suggestion
    JNIEnv *env, jclass, jlong input_handle) {
```

java/src/main/native/src/ColumnViewJni.cpp
```diff
@@ -2457,4 +2457,46 @@ JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_applyBooleanMask(
...
+  JNI_NULL_CHECK(env, column_view_handle, "column_view handle is null", 0);
+  try {
+    cudf::jni::auto_set_device(env);
+
+    cudf::column_view const *cv =
+        reinterpret_cast<cudf::column_view const *>(column_view_handle);
+
+    return release_as_jlong(cudf::purge_nonempty_nulls(*cv));
```
ttnghia, line 2488:
```suggestion
    auto const cv = reinterpret_cast<cudf::column_view const *>(input_handle);
    return release_as_jlong(cudf::purge_nonempty_nulls(*cv));
```

java/src/main/java/ai/rapids/cudf/ColumnView.java
razajafri: I agree as I don't understand the rationale behind adding this API. @mythrocks I think you added this, can you comment here as to whether we should even keep this API?

razajafri: Upon further reading into [this](https://github.com/rapidsai/cudf/commit/84f88ceb18225850835a9912a18e4c82245d5620), I think it's a good fast check as it only checks if the hierarchy has any nulls in it, which means it will just go over the validity buffer. It's essentially a glorified `hasNulls()` check. I think, it should stay otherwise we will have to do the expensive check all the time.

ttnghia: Yes it is a fast check, but it also returns false positive result: 
```
if ((type == type_id::STRING || type == type_id::LIST) && input.has_nulls()) { return true; }
```

As such, `purge_nonempty_nulls` will be called (which is expensive) even if the input doesn't have non-empty nulls. Using it will make things slow down instead of faster. This was proved by a couple of recent cudf issues.

java/src/main/native/src/ColumnViewJni.cpp
```diff
@@ -2457,4 +2457,46 @@ JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_applyBooleanMask(
   CATCH_STD(env, 0);
 }
 
+JNIEXPORT jboolean JNICALL Java_ai_rapids_cudf_ColumnView_hasNonEmptyNulls(
+    JNIEnv *env, jclass, jlong column_view_handle) {
```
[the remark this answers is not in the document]
razajafri: IMO `column_view_handle` is a much more descriptive name and doesn't make sense to change it to a less descriptive name

java/src/main/java/ai/rapids/cudf/ColumnView.java
razajafri: Thank you for your feedback. 

Due to my lack of knowledge of this particular feature and the fact that I am leaning towards keeping this in, I would like to hear from others about their opinion. @mythrocks @hyperbolic2346

razajafri:
> I see the issue with `ColumnView`. Ideally, this (purge non-empty nulls) should only be called in cudf or JNI C++, not in Java. Unfortunately in Java we also have `ColumnView` constructor (and similar) that allow non-empty nulls to exist.
> 
> In addition, the PR description doesn't have any context/meaningful description. Please update it.

ColumnView in Java just calls into the cpp code, if cpp guarantees there will be no non-empty nulls then we won't need to call this from Java. As you said if we are working on a proper fix this wouldn't be needed hopefully.

java/src/main/java/ai/rapids/cudf/ColumnView.java
mythrocks: > This was proved by a couple of recent cudf issues.
@ttnghia: Could you point me to the issues, for reference?

I'm trying to determine if this function is safe to remove. I am yet to be convinced.

> As such, purge_nonempty_nulls will be called (which is expensive) even if the input doesn't have non-empty nulls. Using it will make things slow down instead of faster.

That only applies if `purge_nonempty_nulls()` is called at all. If `may_have_nonempty_nulls()` returns false, `purge_nonempty_nulls()` shouldn't be called at all.

mythrocks: In fact, I don't think `purge_nonempty_nulls()` should be called if `has_nonempty_nulls()` returns `false`. That's a more expensive check, but not as expensive as needlessly reseating a column.

I should peruse the implementation of `purge_nonempty_nulls()`.

ttnghia: Please check it out: https://nvidia.slack.com/archives/CCWT6NJ0K/p1673652348546619. The issue was due to using `purge_nonempty_nulls` with that false positive check, and was mitigated by switching to use `has_nonempty_nulls` check.

mythrocks: Ok. That's what I was talking about with `has_nonempty_nulls()`.

I think I see your point. If this is to decide whether or not to call `purge_nonempty_nulls()`, we should use `has_nonempty_nulls()` (which internally uses may_have_nonempty_nulls()`). Maybe there's little use for `may_have_nonempty_nulls()` in the JNI API.

java/src/main/native/src/ColumnViewJni.cpp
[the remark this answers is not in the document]
mythrocks: I think this is a minor quibble. My $0.02: it's not a pointer until it has been cast from the opaque `jlong`. I would keep it as `handle`. (Plus, it's consistent with the other methods.)

`input_handle` is descriptive, because it indicates that it's the input. `column_view` is more ambiguous and possibly redundant, because it's already a method in `ColumnView`.

(Again, very minor quibble.)

razajafri:
I am having difficulty running pre-commit hooks and thus having to push multiple updates due to formatting errors

java/src/main/native/src/ColumnViewJni.cpp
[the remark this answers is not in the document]
ttnghia: `column_view_handle` seems too long name. In addition, the existing functions already use `input_handle` like https://github.com/rapidsai/cudf/blob/d9bd29a6e7a382564790492b6cebe75e4d1cefc9/java/src/main/native/src/ColumnViewJni.cpp#L683. So please enforce consistency.

ttnghia:
> I am having difficulty running pre-commit hooks and thus having to push multiple updates due to formatting errors

You should run `clang-format` by yourself, not pre-commit hooks.

java/src/main/native/src/ColumnViewJni.cpp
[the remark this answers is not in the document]
revans2: This feels so minor to me. If the name is wrong or confusing, then yes lets change it. But most anything else is just a nit and really it should be up to @razajafri to decide if he uses other's advice or not.

ttnghia (approved):

razajafri:
re-run tests

ttnghia:
CI Java failed due to:
```
Error:  Tests run: 311, Failures: 0, Errors: 1, Skipped: 2, Time elapsed: 8.111 s <<< FAILURE! - in ai.rapids.cudf.ColumnVectorTest
Error:  ai.rapids.cudf.ColumnVectorTest  Time elapsed: 6.67 s  <<< ERROR!
ai.rapids.cudf.RmmException: Could not shut down RMM there appear to be outstanding allocations
```

razajafri:
> CI Java failed due to:
> 
> ```
> Error:  Tests run: 311, Failures: 0, Errors: 1, Skipped: 2, Time elapsed: 8.111 s <<< FAILURE! - in ai.rapids.cudf.ColumnVectorTest
> Error:  ai.rapids.cudf.ColumnVectorTest  Time elapsed: 6.67 s  <<< ERROR!
> ai.rapids.cudf.RmmException: Could not shut down RMM there appear to be outstanding allocations
> ```

Yes, I can see that thanks. The reason why I was confused about the CI failure was that the tests are passing locally for me, so I was wondering if the failure was because of something going haywire in the CI because the "update branch" was enabled.

razajafri:
The CI job is failing but when I try to run the exact CI jobs using @ajschmidt8 help and by following steps listed [here](https://docs.rapids.ai/resources/reproducing-ci/). The tests pass for me locally and inside rapidsai ci docker.

We are still looking into the cause of this

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
```diff
@@ -6691,4 +6692,56 @@ void testApplyBooleanMaskFromListOfStructure() {
       assertColumnsAreEqual(expectedCv, actualCv);
     }
   }
+
+  private ColumnView getColumnViewWithNonEmptyNulls() {
+    List<Integer> list0 = Arrays.asList(1, 2, 3);
+    List<Integer> list1 = Arrays.asList(4, 5, null);
+    List<Integer> list2 = Arrays.asList(7, 8, 9);
+    List<Integer> list3 = null;
+    ColumnVector input =  makeListsColumn(DType.INT32, list0, list1, list2, list3);
```
ttnghia, line 6701: CI failed because there is memory leak in Java tests. Sorry I overlooked the Java tests. Please fix them.

ttnghia: `ColumnVector` as well as `HostColumnVector` must be wrapped in `try` blocks.

razajafri: Hmm... makes me want to have this failure on my local so we can avoid this. It will still be good to get to the bottom of this discrepancy. 

In the meanwhile, I will push a fix. Thanks for pointing us to this.

ajschmidt8: Agreed, the discrepancy isn't ideal.

Could it be caused by the types of GPUs that are used in CI vs. on our local machines?

I'm wondering if that or the amount of memory on a given GPU would contribute to this.

ttnghia: Maybe there is a variable that controls the test behavior, and that var is set by the local environment. @revans2 Do you know about this?

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
```diff
@@ -6693,6 +6693,9 @@ void testApplyBooleanMaskFromListOfStructure() {
     }
   }
 
+  /**
+   * The caller needs to make sure to close the returned ColumnView
```
ttnghia, line 6697: Local variables in this function are still not closed.

ttnghia: Humnn, if this function is always called in a `try` block then we won't this that.

ttnghia, line 6697:
```suggestion
   * The caller needs to make sure to call this function in a `try` block as local variables need to be closed.
```

razajafri: ugh, I am rushing through this and context-switching isn't helping either. I think it will make sense for me to close all the local allocations and leave only the returned vector to be closed by the caller.

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
revans2: If a memory leak happens and java's GC is fast, then the leak is cleared up before RMM is shut down and there is no error.  If GC is slow, then the leak is not cleared and you get the error.  We have a flag to print out messages to help debug leaks, but it is off by default because it makes the tests take a lot longer.  Just set `-Dai.rapids.refcount.debug=true` on the maven commend line when running the tests. Even if they pass you can then look at the logs to see if any leaks were detected and use the stack traces to track down what is leaked and why.

razajafri: But in this case aren't the vectors off the heap? Did you mean the MemoryCleaner instead of GC?

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
```diff
@@ -6701,26 +6701,27 @@ private ColumnView getColumnViewWithNonEmptyNulls() {
...
+        assert (hostColumnVector.isNull(3));
+      }
+      try (ColumnVector expectedOffsetsBeforePurge = ColumnVector.fromInts(0, 3, 6, 9, 9)) {
+        ColumnView offsetsCvBeforePurge = input.getListOffsetsView();
+        assertColumnsAreEqual(expectedOffsetsBeforePurge, offsetsCvBeforePurge);
+      }
+      ColumnView colWithNonEmptyNulls = new ColumnView(input.type, input.rows, Optional.of(2L), dmb,
+          input.getDeviceBufferFor(BufferType.OFFSET), input.getChildColumnViews());
+      assertEquals(2, colWithNonEmptyNulls.nullCount);
+      return colWithNonEmptyNulls;
```
ttnghia, line 6723: I don't think this returned vector should be in this try block.

razajafri: That's not the problem, the problem is that the `input` column will be closed while the view won't. Unfortunately my build env is completely ruined and I will have to spend time to fix it before I can proceed. Thanks for your quick reviews

Issue #12748, opened 2023-02-09
[BUG] Slice scatter assignment uses incorrect num_keys calculation


thomcom:
**Describe the bug**
When assigning values to a `cudf.Series` using a slice with an odd starting position and a step size greater than 1, the computation in `_scatter_by_slice` for the number of keys is incorrect.
`num_keys = (stop - start) // step`
performs integer division, dropping the last value and throwing an error. It should be
`num_keys = ceil((stop - start) / step)`

**Steps/Code to reproduce bug**
```
import cudf
import cupy as cp
s = cudf.Series(cp.arange(5))
t = cudf.Series(cp.zeros(10))
t[1::2] = s
```
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[7], line 3
      1 s = cudf.Series(cp.arange(5))
      2 t = cudf.Series(cp.zeros(10))
----> 3 t[1::2] = s

File ~/compose/etc/conda/cuda_11.6/envs/notebooks/lib/python3.8/contextlib.py:75, in ContextDecorator.__call__.<locals>.inner(*args, **kwds)
     72 @wraps(func)
     73 def inner(*args, **kwds):
     74     with self._recreate_cm():
---> 75         return func(*args, **kwds)

File ~/cudf/python/cudf/cudf/core/series.py:1268, in Series.__setitem__(self, key, value)
   1265 @_cudf_nvtx_annotate
   1266 def __setitem__(self, key, value):
   1267     if isinstance(key, slice):
-> 1268         self.iloc[key] = value
   1269     else:
   1270         self.loc[key] = value

File ~/compose/etc/conda/cuda_11.6/envs/notebooks/lib/python3.8/contextlib.py:75, in ContextDecorator.__call__.<locals>.inner(*args, **kwds)
     72 @wraps(func)
     73 def inner(*args, **kwds):
     74     with self._recreate_cm():
---> 75         return func(*args, **kwds)

File ~/cudf/python/cudf/cudf/core/series.py:236, in _SeriesIlocIndexer.__setitem__(self, key, value)
    231     if to_dtype != self._frame._column.dtype:
    232         self._frame._column._mimic_inplace(
    233             self._frame._column.astype(to_dtype), inplace=True
    234         )
--> 236 self._frame._column[key] = value

File ~/cudf/python/cudf/cudf/core/column/numerical.py:160, in NumericalColumn.__setitem__(self, key, value)
[...]

PR #12749, opened 2023-02-09
Fix a bug with `num_keys` in `_scatter_by_slice`

thomcom:
## Description
This PR closes https://github.com/rapidsai/cudf/issues/12748 by changing the `num_keys` computation in `column._scatter_by_slice`.

## Checklist


python/cudf/cudf/core/column/column.py
```diff
@@ -570,7 +571,7 @@ def _scatter_by_slice(
         start, stop, step = key.indices(len(self))
         if start >= stop:
             return None
-        num_keys = (stop - start) // step
+        num_keys = ceil((stop - start) / step)
```
bdice, line 574: Should we worry about floating precision here? We can do this entirely as an integral computation.
```suggestion
        num_keys = (stop - start) // step)
        if (stop - start) % step:
            num_keys += 1
```

or even `divmod`, which I _think_ uses only integral math:
```suggestion
        num_keys, remainder = divmod(stop - start, step)
        if remainder:
            num_keys += 1
```

wence-: How about: `len(range(len(self))[key])`

thomcom: I'm not sure what the danger is w.r.t. floating point precision. Isn't the worst case going to be ((2^^32 - (2^^32-1)) / (2^^32-1)). `ceil` works fine on this value, anyway.

I'm not a fan of `len(range(len(self))[key]`. That has us building python range objects that are as much as 2^^32 in size, major performance crush.

bdice: @thomcom Indexing a `range` with a `slice` produces another `range`. It won't materialize that as a list of the indices.

bdice: @wence- has the best solution here afaict. Good call. I should've zoomed out and seen that in context. :)
```suggestion
        num_keys = len(range(len(self))[key])
```

wence-: Probably it is more obviously written as:
```
len(range(start, stop, step))
```

thomcom:
Did that

thomcom:
rerun tests

thomcom:
This is ready to merge I think, and a PR in cuSpatial depends on it (https://github.com/rapidsai/cuspatial/pull/926)

wence- (approved):
Sorry, I thought I had hit approve, but no. Looks great, thanks!

Issue #11844, opened 2022-10-03
[FEA] Story - Full support for nested types


GregoryKimball:
After the introduction of new row operators for nested types (#10186), it's time to build on that success with a new issue to focus on the outstanding items. The row operators for equality `=` (#10289) and hashing `#` (#10641) included support for arbitrarily-nested `List` and `Struct` types. List inequality `<` (#11129) added support for arbitrarily-nested `List<List>` and `Struct<List>`. Please note that the Spark-RAPIDS plugin has a companion issue (https://github.com/NVIDIA/spark-rapids/issues/8550) to this story issue.

### Part 1: Transition algorithms to new row operators

|Algorithm|Status|Operation|File|Notes|
|---|---|---|---|---|
|scan_rank| ✅ ️ ️|`=`️|[rank_scan.cu](https://github.com/rapidsai/cudf/blob/branch-22.12/cpp/src/reductions/scan/rank_scan.cu)|#10289|
|distinct| ✅ ️ |`#`|[distinct.cu](https://github.com/rapidsai/cudf/blob/branch-22.12/cpp/src/stream_compaction/distinct.cu)|#10641|
| distinct_count | ✅ #12776 | `#`, `=` | |
| unique_count | ✅ #12776 | `#`, `=` | |
|hash-groupby| ✅ ️|`#`, `=`|[groupby.cu](https://github.com/rapidsai/cudf/blob/branch-22.12/cpp/src/groupby/hash/groupby.cu)|#10770|
|sort|✅ ️|`<`|[sort_impl.cuh](https://github.com/rapidsai/cudf/blob/branch-22.12/cpp/src/sort/sort_impl.cuh)|#11129, using legacy `relational_compare` in `simple_comparator` as part of two-path solution in `sorted_order`|
|list contains|✅|`=`| [contains.cu](https://github.com/rapidsai/cudf/blob/branch-22.10/cpp/src/lists/contains.cu)| #10548, using legacy `equality_compare` as part of two-path solution in `search_list_non_nested_types_fn`, see also #11330 for the two-path solution for equality|
|struct binops|✅|`=`, `<` | |#11153 | 
|murmur hash |✅|`#`| | plus spark_murmur_hash |
[...]

Issue #12593, opened 2023-01-23
[FEA] Refactor `experimental/row_operators.cuh` and make it default


GregoryKimball:
**Is your feature request related to a problem? Please describe.**
libcudf contains two sets of row operators: [legacy row operators](https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/include/cudf/table/row_operators.cuh) for simple types and [experimental row operators](https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/include/cudf/table/experimental/row_operators.cuh) for complex types. When we have completed "Part 1" of #11844, then we can safely refactor the experimental row operators to be the default. 

**Describe the solution you'd like**
Ultimately we will deprecate the legacy row operators and move the experimental row operators out of the experimental namespace. Please note that portions of the legacy code are likely to live on as a "fast path" for simple types (see #11330 and #11129). Also we should consider suggestions such as #11012 when refactoring the experimental row comparators.

There are some open issues about how to reach this final state, such as:
* how long to keep the legacy comparators
* how to evolve the test suite 

**Describe alternatives you've considered**
n/a

**Additional context**
See #10186 and follow-on issue #11844 for more information about the nested type comparator project.

vyasr:
To be clear, I would expect the "fast paths" that you describe to largely be implemented as template specializations of the new comparators rather than leaving the old comparators in place. Much of the code should be reusable and that's the path we've been able to follow in other places e.g. #11129.

PR #12752, opened 2023-02-09
Update `is_sorted` to use `experimental::row::lexicographic`

divyegala:
This PR is a part of https://github.com/rapidsai/cudf/issues/11844.

## Checklist


cpp/tests/sort/is_sorted_tests.cpp
```diff
@@ -223,6 +224,65 @@ auto nulls_before<cudf::struct_view>()
...
+template <typename T>
+std::enable_if_t<std::is_same_v<T, cudf::list_view>, lcw> ascending()
+{
+  lcw col{lcw{lcw{lcw{0}}, lcw{lcw{0}}, lcw{lcw{0}}},
+          lcw{lcw{lcw{0}, lcw{0}, lcw{0}}},
+          lcw{lcw{lcw{0, 0}}, lcw{lcw{0, 0, 0, 0, 0, 0, 0, 0}}, lcw{lcw{0}}},
+          lcw{lcw{lcw{0, 0, 0}}},
+          lcw{lcw{lcw{0, 0, 0}}, lcw{lcw{0}}, lcw{lcw{0}}}};
+
+  return col;
```
ttnghia, line 249:
```suggestion
  return lcw{lcw{lcw{lcw{0}}, lcw{lcw{0}}, lcw{lcw{0}}},
          lcw{lcw{lcw{0}, lcw{0}, lcw{0}}},
          lcw{lcw{lcw{0, 0}}, lcw{lcw{0, 0, 0, 0, 0, 0, 0, 0}}, lcw{lcw{0}}},
          lcw{lcw{lcw{0, 0, 0}}},
          lcw{lcw{lcw{0, 0, 0}}, lcw{lcw{0}}, lcw{lcw{0}}}};
```

cpp/tests/sort/is_sorted_tests.cpp
```diff
@@ -223,6 +224,65 @@ auto nulls_before<cudf::struct_view>()
...
+std::enable_if_t<std::is_same_v<T, cudf::list_view>, lcw> descending()
+{
+  lcw col{lcw{lcw{lcw{0, 0, 0}}, lcw{lcw{0}}, lcw{lcw{0}}},
+          lcw{lcw{lcw{0, 0, 0}}},
+          lcw{lcw{lcw{0, 0}}, lcw{lcw{0, 0, 0, 0, 0, 0, 0, 0}}, lcw{lcw{0}}},
+
+          lcw{lcw{lcw{0}, lcw{0}, lcw{0}}},
+          lcw{lcw{lcw{0}}, lcw{lcw{0}}, lcw{lcw{0}}}};
+
+  return col;
```
ttnghia, line 262: Similar to above. Return directly.

cpp/src/sort/is_sorted.cu
```diff
@@ -36,31 +36,27 @@ namespace detail {
...
+  auto const comparator =
+    experimental::row::lexicographic::self_comparator{in, column_order, null_precedence, stream};
 
-  auto const d_input           = table_device_view::create(flattened, stream);
-  auto const d_column_order    = make_device_uvector_async(flattened.orders(), stream);
-  auto const d_null_precedence = has_nulls
-                                   ? make_device_uvector_async(flattened.null_orders(), stream)
-                                   : rmm::device_uvector<null_order>(0, stream);
+  if (cudf::detail::has_nested_columns(in)) {
+    auto const device_comparator = comparator.less<true>(has_nested_nulls(in));
```
PointKernel, line 46: Not sure if it's been discussed before, I feel it makes more sense to move this branching (whether has nested cols or not) into the comparator detail.

divyegala: This is the way it was implemented for `lexicographic::*_comparator`, so I followed the convention. That said, there are some algorithms that dispatch different code-paths for nested and non-nested types already, so you want the granular level of control where those algorithms can select a particular version of the comparator

ttnghia (approved):

cpp/src/sort/is_sorted.cu
PointKernel: Understood. It's definitely out of the scope of this PR. But if we always dispatch different code-paths, the fine control of those algorithms seems unnecessary.

PointKernel (approved):
LGTM

cpp/src/sort/is_sorted.cu
davidwendt: I agree. All of these if-statements scattered throughout the code seems like an anti-pattern we should try to factor out somehow. We should take a look once these are done. I'm also worried about compile time (surprise) and so that would be a reason to consider refactoring before these are complete perhaps.

divyegala: I would prefer any refactor for design patterns to be a follow up to this https://github.com/rapidsai/cudf/issues/12593

PR #12754, opened 2023-02-10
Throw an exception if an unsupported page encoding is detected in Parquet reader

etseidl:
## Description
If the Parquet reader comes across a page encoded with an unsupported encoding, the call to decode page data silently fails, leading to either an empty table or unrelated exceptions being thrown.  This PR adds code to validate the page encodings after the page headers are decoded.

## Checklist


etseidl:
Not sure why clang-format changed the lines at the top of parquet_test.cpp

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -329,6 +340,12 @@ void decode_page_headers(hostdevice_vector<gpu::ColumnChunkDesc>& chunks,
   chunks.host_to_device(stream);
   gpu::DecodePageHeaders(chunks.device_ptr(), chunks.size(), stream);
   pages.device_to_host(stream, true);
+
+  // validate page encodings
+  CUDF_EXPECTS(thrust::all_of(
```
mythrocks, line 345: `thrust::all_of()` continues to have performance challenges: https://github.com/NVIDIA/thrust/issues/1016

Could we consider switching this to `thrust::reduce()` instead?

cpp/tests/io/parquet_test.cpp
```diff
@@ -5113,4 +5113,58 @@ TEST_P(ParquetSizedTest, DictionaryTest)
...
+{
+  {
+    // TODO this will need to change when DELTA_BINARY_PACKED is supported
+    // cudf cannot produce a file with a non-supported encoding, so this is a small file
+    // produced by parquet-mr
+    const unsigned char delta_binary_parquet[] = {
+      0x50, 0x41, 0x52, 0x31, 0x15, 0x06, 0x15, 0x1e, 0x15, 0x22, 0x5c, 0x15, 0x06, 0x15, 0x00,
+      0x15, 0x02, 0x15, 0x0a, 0x15, 0x06, 0x15, 0x04, 0x00, 0x00, 0x03, 0x06, 0x03, 0x3f, 0x00,
+      0x0a, 0x24, 0x80, 0x01, 0x04, 0x03, 0x02, 0x02, 0x00, 0x00, 0x00, 0x00, 0x19, 0x11, 0x02,
+      0x19, 0x18, 0x08, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x19, 0x18, 0x08, 0x03,
```
mythrocks, line 5126: I don't know what the current policy is in cuIO, regarding writing out the contents of the file in the test, versus checking in an actual Parquet file. @vuule?

cpp/src/io/parquet/parquet_common.hpp
```diff
@@ -86,11 +86,12 @@ enum class Encoding : uint8_t {
   GROUP_VAR_INT           = 1,  // Deprecated, never used
   PLAIN_DICTIONARY        = 2,
   RLE                     = 3,
-  BIT_PACKED              = 4,
+  BIT_PACKED              = 4,  // Deprecated
```
mythrocks, line 89: Question: Could you please clarify how this change relates to the change in the reader?

mythrocks (changes requested):
LGTM, mostly. A couple of questions. We should, however, switch out of using `thrust::all_of()`.

cpp/src/io/parquet/reader_impl_preprocess.cu
etseidl: sure

cpp/src/io/parquet/parquet_common.hpp
etseidl: I'm just updating the enum to make it match the current parquet spec, which lists BIT_PACKED as a deprecated encoding.  Can remove if you'd prefer.

cpp/tests/io/parquet_test.cpp
vuule: If possible, a Python test that open a checked-in file is preferable.

etseidl: :tired_face:

etseidl: Many thanks to @nvdbaranec for python testing tips!

galipremsagar (approved):

cpp/src/io/parquet/reader_impl_preprocess.cu
ttnghia: Recommend using `thrust::count_if() > 0`.

cpp/src/io/parquet/parquet_common.hpp
ttnghia: Can you please add more comments about why/when/how etc. this enum is deprecated? As well as tracking issue (if any)?

mythrocks:
I've added a couple of labels (`Bug`, `Non-breaking`) to get the CI going.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -307,6 +307,17 @@ template <typename T = uint8_t>
   return total_pages;
 }
 
+constexpr bool is_supported_encoding(Encoding enc)
```
ttnghia, line 310: Similar to my comment above, can you add more clarification about the supported encoding? Like, why these encodings are supported while the others aren't, and where to look up?

cpp/src/io/parquet/parquet_common.hpp
etseidl: @ttnghia I did some digging through git blame on the parquet-format README.md.  I found the commit but couldn't find an associated JIRA issue. I added a little more context to the comment.

cpp/src/io/parquet/reader_impl_preprocess.cu
etseidl: @ttnghia should I also add a comment in `setupLocalPageInfo` to modify this function if additional encodings are added?

ttnghia: Yes, please do so. Ideally, the two functions should share code to always sync with each other but I don't know how to do that in the meantime. So please add a comment to remind other dev to sync them whenever they work on them.

ttnghia: BTW, if that code (page_data.cu:L1172) throws error when an unsupported encoding is detected, why do you still need this PR?

ttnghia: You mentioned that:
> the call to decode page data silently fails

So can we just modify that "silently fails" behavior to "loudly fails"? By doing so, we can avoid sync between different places.

etseidl: It actually doesn't throw an error. It just sets an error condition that is never checked.  I tried adding a CUDF_UNREACHABLE to the kernel, but that didn't halt execution as I thought it would.

etseidl: > You mentioned that:
> 
> > the call to decode page data silently fails
> 
> 
> 
> So can we just modify that "silently fails" behavior to "loudly fails"? By doing so, we can avoid sync between different places.

Yes, if I could actually get it to fail loud enough I'd prefer that approach.

ttnghia: Thanks for that. Then let's move ahead with this PR while we may be able to improve the error check in the kernels later.
Sound like there is room for improvement @nvdbaranec.

etseidl: Do I need to use CUDF_CUDA_TRY?

ttnghia: Throwing exception in device code is difficult. Typically, people store a global error state in global memory, which will be returned to the host after kernel launch. The caller will then investigate that error state and throw exception if needed.

etseidl: Ok, I added a comment to `setupLocalPageInfo`.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -329,6 +341,14 @@ void decode_page_headers(hostdevice_vector<gpu::ColumnChunkDesc>& chunks,
...
   gpu::DecodePageHeaders(chunks.device_ptr(), chunks.size(), stream);
   pages.device_to_host(stream, true);
+
+  // validate page encodings (avoiding use of thrust::any_of() NVIDIA/thrust #1016)
+  auto const num_valid_pages = static_cast<size_t>(thrust::count_if(
+    rmm::exec_policy(stream), pages.d_begin(), pages.d_end(), [] __device__(auto const& page) {
+      return is_supported_encoding(page.encoding);
+    }));
+
+  CUDF_EXPECTS(num_valid_pages == pages.size(), "Unsupported page encoding detected");
```
ttnghia, line 351: Wait, I just see that we already call `pages.device_to_host(stream, true);` in the line right above. That means we can call `is_supported_encoding` on host code:
```suggestion

  // validate page encodings (avoiding use of thrust::any_of() NVIDIA/thrust #1016)
  auto const num_valid_pages = static_cast<size_t>(thrust::count_if(
    rmm::exec_policy(stream), pages.begin(), pages.end(), [] auto const& page) {
      return is_supported_encoding(page.encoding);
    }));

  CUDF_EXPECTS(num_valid_pages == pages.size(), "Unsupported page encoding detected");
```

vuule: Can it be any_of/all_of now?

etseidl: I was wanting to use the thrust version to do the check in parallel since the number of pages can easily get into the thousands (or higher).  Do you thinks that's not worth worrying about?  Then, yes, I'd switch back to std::any_of/all_of.

ttnghia: The runtime of thousands of such trivial checks will be negligible. So you don't need to be worried about it.

etseidl: Ok, it's back on the host.  It's slower by 30us :rofl: (for a totally unscientific single pass of nsys profile).

ttnghia (approved):

cpp/src/io/parquet/reader_impl_preprocess.cu
vuule: +30us from 1us, or from 1ms? :D

etseidl: 30us to 60 us :D

etseidl:
ok to test?

vuule (approved):

etseidl:
looks like CI is fixed, can this be tested again please?

etseidl:
Any objections to merging this?

PR #12760, opened 2023-02-11
Automatic include sorting with clang-format

bdice:
## Description
This PR is inspired by [a comment from @wence-](https://github.com/rapidsai/cudf/pull/12705#discussion_r1099912044)

> I guess it would be nice to do this automatically with clang-format, but perhaps [`IncludeBlocks: Regroup`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includeblocks) plus [`IncludeCategories`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories) leads to too many bad automated decisions.

I looked into these clang-format options and I came up with some settings that I think are pretty good.

I haven't quite finished iterating on the rule set, so I'm opening this draft just to mark my progress. If you're interested in seeing the results or tweak the rules, check out this PR and run `pre-commit run clang-format --all-files`.

## Checklist

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -71,7 +71,110 @@ ForEachMacros:
   - foreach
   - Q_FOREACH
   - BOOST_FOREACH
-IncludeBlocks: Preserve
+IncludeBlocks: Regroup
+IncludeCategories:
+  # Local includes first
+  - Regex:           '".*"'
+    Priority:        0
+  # Internal includes from non-cudf source directories
+  - Regex:           '^<benchmarks/'
+    Priority:        1
+  - Regex:           '^<binaryop/'
+    Priority:        2
+  - Regex:           '^<groupby/'
+    Priority:        2
+  - Regex:           '^<hash/'
+    Priority:        3
+  - Regex:           '^<io/'
+    Priority:        3
+  - Regex:           '^<jit_preprocessed_files/'
+    Priority:        4
+  - Regex:           '^<jit/'
+    Priority:        5
+  - Regex:           '^<join/'
+    Priority:        3
+  - Regex:           '^<lists/'
+    Priority:        3
+  - Regex:           '^<quantiles/'
+    Priority:        3
+  - Regex:           '^<reductions/'
+    Priority:        3
+  - Regex:           '^<rolling/'
+    Priority:        3
+  - Regex:           '^<sort/'
+    Priority:        3
+  - Regex:           '^<src/'
+    Priority:        3
+  - Regex:           '^<strings/'
+    Priority:        3
+  - Regex:           '^<tests/'
+    Priority:        6
+  - Regex:           '^<text/'
+    Priority:        6
+  # cudf_test, cudf, nvtext are "proper" packages
+  - Regex:           '^<cudf_test/'
+    Priority:        7
+  - Regex:           '^<cudf/'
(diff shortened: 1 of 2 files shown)


davidwendt:
Adding specific directory names to clang-format seems problematic. It means we are not likely to be able to share a clang-format with other RAPIDS projects very easily. Also, it becomes a maintenance issue to try to remember to update this correctly.
Is this really important to solve this way? I don't think this is worth doing.

bdice:
@davidwendt I think that is valid criticism. I posed this PR for discussion because I am frequently commenting in PR reviews to keep includes grouped in accordance with our developer guide, and that seems like a job for tooling to handle on our behalf. I think maintaining this file with explicit rules _might_ be easier than the continuous burden of cleaning up includes that are automatically added by other developers’ IDEs in the wrong places or for developers who are not familiarized with the include order we prescribe in the developer guide. I would rather comment on PRs when things are organized incorrectly and add a new rule for clang-format. However, I don’t want to push this if it is viewed as an excessive maintenance burden relative to the benefit it might provide.

wence-:
I assume the reason for include sorting is style, rather than correctness? Modulo bugs in libstdc++ [like this one](https://travisdowns.github.io/blog/2019/11/19/toupper.html) you would hope that include sorting doesn't change the correctness of the code. Would it therefore be reasonable to just try and sort into buckets of "cuda", "libstdc++", and "other"? That would make things more portable across projects. FWIW, as an occasional (for now at least) contributed to the C++ code in cudf, I haven't fully internalised how "far away" everything is, and definitely clangd autocompletion providing IWYU by magic helps, but it obviously doesn't know about the rules so can't know where to put things.

bdice:
> If I contrast with the python side, I just import where-ever at the top of the file and then `isort` automagically fixes it for me, so I never have to think; the same criticism that David raises about the configuration being project-specific could be raised against isort (since one has to specify things), but that doesn't seem like a massive problem in usage AFAICT.

This agrees with my experience. In practice, it hasn't been painful to maintain for isort. The primary downsides are that the C++ code has a few more "categories" than Python, and it is harder to maintain a regex for some of the categories, such as what's in the C++ STL (isort maintains the list of Python standard libraries for us) and what's local to libcudf (e.g. `<io/...>`).

vyasr:
The difference I see between the way isort is configured in Python and how clang-format is configured for includes is that the includes from a given library may be in an arbitrary paths whereas Python package imports can be easily identified by either `import pkg` or `from pkg[*] import *`. Therefore this list is likely to see more churn than the isort list, although probably not all that much more.

My biggest takeaway from glancing through this PR is that our "distance metric" is too complicated. To @wence-'s point, I suspect that most infrequent contributors don't know the metric we use, and based on how often these comments are raised in reviews I would guess that many frequent contributors haven't quite internalized this either. Before going down the programmatic sorting route I think we should consider reducing the complexity of our groupings as well so that there is a smaller list to learn and maintain. I don't personally get much value out of groups beyond internal, stdlib, and thrust/cub/libcu++. Enforcing a simpler set would also reduce PR reviewer burden.

bdice:
Closing this as stale -- there are several difficult questions to answer about simplifying the rules we use to sort includes, and I don't think I can prioritize this work anytime soon.

Issue #12766, opened 2023-02-13
[FEA] Add alternate handling for all null list in  join_list_elements


karthikeyann:
**Is your feature request related to a problem? Please describe.**
All null list is considered as empty list as special handling in `join_list_elements` API. 
But this special handling may not be required by all the users.

**Describe the solution you'd like**
Provide an option to skip this special handling.

**Describe alternatives you've considered**
Replace all nulls in child columns with `"null"` before using `join_list_elements` API.  But it incurs extra memory and compute time.

**Additional context**

PR #12767, opened 2023-02-13
Fix bug in all-null list due to join_list_elements special handling

karthikeyann:
## Description
All-null list is considered as empty list in `join_list_elements`. So, nulls of children are replaced by `narep` before passing to `join_list_elements` API.
Related https://github.com/rapidsai/cudf/issues/12766

## Checklist

### DIFF
--- a/cpp/src/io/json/write_json.cu
+++ b/cpp/src/io/json/write_json.cu
@@ -438,7 +438,7 @@ struct column_to_strings_fn {
           ->view(),
         narep,
         stream_,
-        mr_);
+        rmm::mr::get_current_device_resource());
       auto const list_child_string =
         column_view(column.type(),
                     column.size(),
--- a/cpp/src/io/json/write_json.cu
+++ b/cpp/src/io/json/write_json.cu
@@ -419,26 +419,23 @@ struct column_to_strings_fn {
     auto list_string           = [&]() {
       // nulls are replaced due to special handling of all-null lists as empty lists
       // by join_list_elements
+      auto child_string_with_null = [&]() {
+        if (child_view.type().id() == type_id::STRUCT) {
+          return (*this).template operator()<cudf::struct_view>(
+            child_view,
+            children_names.size() > child_index ? children_names[child_index].children
+                                                          : std::vector<column_name_info>{});
+        } else if (child_view.type().id() == type_id::LIST) {
+          return (*this).template operator()<cudf::list_view>(
+            child_view,
+            children_names.size() > child_index ? children_names[child_index].children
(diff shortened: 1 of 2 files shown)


davidwendt (approved):

cpp/src/io/json/write_json.cu
```diff
@@ -417,21 +417,28 @@ struct column_to_strings_fn {
...
+              children_names.size() > child_index ? children_names[child_index].children
+                                                            : std::vector<column_name_info>{});
+          } else {
+            return cudf::type_dispatcher(child_view.type(), *this, child_view);
+          }
+        }()
+          ->view(),
+        narep,
+        stream_,
+        mr_);
```
ttnghia, line 441: This `child_string` is temporary variable thus we may use default `mr` for it.

karthikeyann:
```suggestion
        rmm::mr::get_current_device_resource());
```

ttnghia (approved):

cpp/src/io/json/write_json.cu
```diff
@@ -417,21 +417,28 @@ struct column_to_strings_fn {
...
-            child_view,
-            children_names.size() > child_index ? children_names[child_index].children
-                                                          : std::vector<column_name_info>{});
-        } else {
-          return cudf::type_dispatcher(child_view.type(), *this, child_view);
-        }
-      }();
+      // nulls are replaced due to special handling of all-null lists as empty lists
+      // by join_list_elements
+      auto child_string = cudf::strings::detail::replace_nulls(
```
bdice, line 422: Non-blocker: I think I'd prefer this to be two separate lines. First the IIFE, then call `cudf::strings::detail::replace_nulls`. It's just pretty long and complicated to read. After 20 lines of code, I forget that we're computing a function argument.

bdice (approved):

cpp/src/io/json/write_json.cu
karthikeyann: I used this to avoid the temporary. Updated.

cpp/src/io/json/write_json.cu
```diff
@@ -431,7 +433,9 @@ struct column_to_strings_fn {
         } else {
           return cudf::type_dispatcher(child_view.type(), *this, child_view);
         }
-      }();
+      };
```
ttnghia, line 436:
```suggestion
      }();
```

karthikeyann: It's written this way to avoid the temporary variable. It will help reduce peak memory.

PR #12769, opened 2023-02-13
Optimize set-like operations

ttnghia:
Set-like operations such as `intersect_distinct` and `difference_distinct` call `purge_nonempty_nulls` when the input is nullable. This PR optimizes these set APIs by checking the existence of non-empty nulls (using `has_nonempty_nulls`) before calling to `purge_nonempty_nulls`.


ttnghia:
Benchmark comparing before vs after this PR:
```
# intersect_distinct

## [0] Quadro RTX 6000

|  num_rows  |  depth  |  null_frequency  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |          Diff |   %Diff |  Status  |
|------------|---------|------------------|------------|-------------|------------|-------------|---------------|---------|----------|
|    2^10    |    1    |        0         | 236.523 us |      14.04% | 235.340 us |      11.01% |     -1.184 us |  -0.50% |   PASS   |
|    2^13    |    1    |        0         | 297.840 us |      14.22% | 294.913 us |       8.53% |     -2.927 us |  -0.98% |   PASS   |
|    2^16    |    1    |        0         | 927.063 us |       4.98% | 928.836 us |       4.78% |      1.774 us |   0.19% |   PASS   |
|    2^10    |    4    |        0         |  25.484 ms |       1.07% |  27.372 ms |       4.52% |      1.888 ms |   7.41% |   FAIL   |
|    2^13    |    4    |        0         | 136.627 ms |       0.81% | 146.370 ms |       0.49% |      9.743 ms |   7.13% |   FAIL   |
|    2^16    |    4    |        0         | 422.196 ms |       0.31% | 466.010 ms |       2.20% |     43.814 ms |  10.38% |   FAIL   |
|    2^10    |    1    |       0.2        | 365.748 us |       9.44% | 346.861 us |      60.68% |    -18.887 us |  -5.16% |   PASS   |
|    2^13    |    1    |       0.2        | 405.400 us |       9.24% | 396.616 us |      45.96% |     -8.784 us |  -2.17% |   PASS   |
|    2^16    |    1    |       0.2        | 833.861 us |       6.33% | 875.925 us |      31.60% |     42.064 us |   5.04% |   PASS   |
|    2^10    |    4    |       0.2        |  24.712 ms |       1.05% |  26.997 ms |       4.53% |      2.285 ms |   9.24% |   FAIL   |
|    2^13    |    4    |       0.2        |  13.221 ms |       1.32% |  11.813 ms |       8.87% |  -1407.219 us | -10.64% |   FAIL   |
|    2^16    |    4    |       0.2        |  95.765 ms |       0.50% |  80.123 ms |       4.85% | -15642.116 us | -16.33% |   FAIL   |
|    2^10    |    1    |       0.8        | 353.305 us |      17.76% | 331.605 us |      51.34% |    -21.700 us |  -6.14% |   PASS   |
|    2^13    |    1    |       0.8        | 368.859 us |       8.63% | 378.645 us |      51.42% |      9.786 us |   2.65% |   PASS   |
|    2^16    |    1    |       0.8        | 392.823 us |      10.70% | 390.992 us |      49.61% |     -1.831 us |  -0.47% |   PASS   |
|    2^10    |    4    |       0.8        |   1.621 ms |       4.38% |   1.504 ms |      27.91% |   -117.088 us |  -7.22% |   FAIL   |
|    2^13    |    4    |       0.8        |   1.748 ms |       3.64% |   1.521 ms |      21.72% |   -226.220 us | -12.94% |   FAIL   |
|    2^16    |    4    |       0.8        |   1.959 ms |       4.83% |   1.929 ms |      29.51% |    -30.942 us |  -1.58% |   PASS   **|**
```

PointKernel:
Are the performance regression in several test cases expected?

ttnghia:
> Are the performance regression in several test cases expected?

Yes. Previously, only a trivial check `null_count == 0` was used. Now `has_nonempty_nulls()` is called (if the input is nullable).

cpp/src/lists/set_operations.cu
```diff
@@ -181,8 +179,10 @@ std::unique_ptr<column> intersect_distinct(lists_column_view const& lhs,
                                   stream,
                                   mr);
 
-  return null_count == 0 ? std::move(output)
-                         : cudf::detail::purge_nonempty_nulls(output->view(), stream, mr);
+  if (auto const output_cv = output->view(); cudf::detail::has_nonempty_nulls(output_cv, stream)) {
```
vyasr, line 182: Does checking `null_count == 0` as an extra early exit before calling `has_nonempty_null` help at all here, or is the perf difference negligible?

cpp/tests/lists/set_operations/difference_distinct_tests.cpp
```diff
@@ -241,13 +241,19 @@ TEST_F(SetDifferenceTest, StringTestsWithNullsEqual)
...
+                                     null_at(1)}
+                         .release();
+      auto& child = str_lists->child(cudf::lists_column_view::child_column_index);
+      child.set_null_mask(cudf::create_null_mask(child.size(), cudf::mask_state::ALL_VALID), 0);
+      return str_lists;
+    }();
 
     auto const results_sorted = set_difference_sorted(lhs, rhs, NULL_EQUAL);
-    CUDF_TEST_EXPECT_COLUMNS_EQUAL(expected, *results_sorted);
+    CUDF_TEST_EXPECT_COLUMNS_EQUAL(*expected, *results_sorted);
```
vyasr, line 256: I realize this is a test that will terminate very soon, so it's not a big deal, but the changes here introduced a memory leak. Why call `release` above, why not just `std::move` the output if RVO isn't happening for you (although I think it should?)?

cpp/tests/lists/set_operations/intersect_distinct_tests.cpp
```diff
@@ -280,11 +281,18 @@ TEST_F(SetIntersectTest, StringTestsWithNullsUnequal)
        strings_lists{}, /* NULL */
        strings_lists{"aha", "this", "is another", "string???"}},
       null_at(1)};
-    auto const expected =
-      strings_lists{{strings_lists{}, strings_lists{} /*NULL*/, strings_lists{"this"}}, null_at(1)};
+    auto const expected = [] {
+      auto str_lists =
+        strings_lists{{strings_lists{}, strings_lists{} /*NULL*/, strings_lists{"this"}},
+                      null_at(1)}
+          .release();
```
vyasr, line 288: Another leak.

cpp/benchmarks/lists/set_operations.cpp
```diff
@@ -0,0 +1,93 @@
...
+      case SetOperations::HAVE_OVERLAP:
+        cudf::lists::have_overlap(cudf::lists_column_view{*lhs}, cudf::lists_column_view{*rhs});
+        break;
+      case SetOperations::INTERSECT_DISTINCT:
+        cudf::lists::intersect_distinct(cudf::lists_column_view{*lhs},
+                                        cudf::lists_column_view{*rhs});
+        break;
+
+      default: CUDF_FAIL("Invalid set op.");
+    }
```
vyasr, line 67: Minor nit, but instead of using an enum couldn't you pass a lambda so that you don't need to define the enum or be worried about invalid ops?

vyasr (changes requested):

cpp/src/lists/set_operations.cu
ttnghia: `has_nonempty_nulls` also checks for it so we don't need to do it here.

cpp/tests/lists/set_operations/difference_distinct_tests.cpp
ttnghia: `strings_lists` here is a `column_wrapper`, so `release()` will return a `unique_ptr<column>`. I guess you're misunderstanding it as `release()` is releasing the raw pointer.

We need it have `expected` as a `unique_ptr<column>` to access the `column& column::child()` function, which ultimately gives us the ability to call `set_null_mask`.

cpp/tests/lists/set_operations/intersect_distinct_tests.cpp
vyasr: This was an error on my part, see the other thread.

cpp/tests/lists/set_operations/difference_distinct_tests.cpp
vyasr: You're absolutely right, I thought it was a column and missed that it was a wrapper.

cpp/benchmarks/lists/set_operations.cpp
vyasr: This remains TODO

vyasr (approved):
I have one open request about getting rid of an unnecessary enum and just using a lambda, but since that got a :+1: I'm approving assuming that change will happen.

cpp/benchmarks/lists/set_operations.cpp
```diff
@@ -0,0 +1,86 @@
...
+  auto data_table =
+    create_random_table({cudf::type_id::LIST}, row_count{num_rows}, data_profile{builder});
+  return std::move(data_table->release().front());
+}
+
+template <typename BenchFuncPtr>
+void nvbench_set_op(nvbench::state& state, BenchFuncPtr bfunc)
+{
+  cudf::rmm_pool_raii pool_raii;
+
```
PointKernel, line 46:
```suggestion
```
This is no longer needed thanks to #12728

PointKernel (approved):
One non-blocking suggestion.

LGTM

vyasr (approved):

PR #12770, opened 2023-02-14
Deallocate encoded data in ORC writer immediately after compression

vuule:
## Description
Reduces total memory footprint over the execution time (but not peak memory usage, which happens during compression). This change helps overall memory usage when writing multiple files in parallel using the same GPU.

## Checklist


cpp/src/io/orc/writer_impl.cu
```diff
@@ -2246,6 +2246,11 @@ void writer::impl::write(table_view const& table)
                                   enc_data.streams,
                                   comp_results,
                                   stream);
+
+      // deallocate encoded data as it is not needed anymore
+      enc_data.data.resize(0, stream);
+      enc_data.data.shrink_to_fit(stream);
```
ttnghia, line 2252: Can this just this?
```
enc_data.data = rmm::device_uvector<uint8_t>{0, stream};
```

vuule: Yeah, I think so. I don't really like either option, so I can change it if you prefer the version with the ctor.

ttnghia: I don't really like either options too. The best option for me would be `.data = nullptr` :smile:
Nevertheless, using the version with ctor seems clearer to me, although they may produce the same result.

davidwendt: Another option is `(void)enc_data.data.release();` 
This will free the underlying storage and leave `.data` with an empty uvector.
The underlying deallocate is async and uses the stream on which the original `device_uvector` was created.

vuule: hm, looks like `release()` does not resize the vector. So it would not be safe to use after this. 
I'm leaning towards `enc_data.data = rmm::device_uvector<uint8_t>{0, stream};` at this point.

ttnghia (approved):

davidwendt (approved):

PR #12776, opened 2023-02-14
Update `distinct/unique_count` to `experimental::row` hasher/comparator

divyegala:
This PR is a part of #11844 

## Checklist

## Compilation Times
`distinct_count.cu`
This PR: 3m8.392s
main: 1m37.576s

`unique_count.cu`
This PR: 13m8.858s
main: 8m21.900s


cpp/src/stream_compaction/unique_count.cu
```diff
@@ -70,14 +71,26 @@ cudf::size_type unique_count(table_view const& keys,
...
+      thrust::counting_iterator<cudf::size_type>(keys.num_rows()),
+      [comp] __device__(cudf::size_type i) { return (i == 0 or not comp(i, i - 1)); });
+  } else {
+    auto const comp =
+      row_comp.equal_to<false>(nullate::DYNAMIC{has_nested_nulls(keys)}, nulls_equal);
+    return thrust::count_if(
+      rmm::exec_policy(stream),
+      thrust::counting_iterator<cudf::size_type>(0),
+      thrust::counting_iterator<cudf::size_type>(keys.num_rows()),
+      [comp] __device__(cudf::size_type i) { return (i == 0 or not comp(i, i - 1)); });
```
vyasr, line 92: Is it not worth doing the same `comparator_helper` trick here since the comparator is generated with a single simple line?

vyasr (changes requested):
Couple of minor questions.

cpp/src/stream_compaction/unique_count.cu
divyegala: Would have to create a functor for the device lambda then, since you can't have device lambdas inside host lambdas

cpp/src/stream_compaction/distinct_count.cu
```diff
@@ -125,33 +126,50 @@ cudf::size_type distinct_count(table_view const& keys,
...
-    return key_map.get_size() + static_cast<std::size_t>((null_count > 0) ? 1 : 0);
+  auto const comparator_helper = [&](auto const row_equal) {
+    // when nulls are equal, insert non-null rows only to improve efficiency
+    if (nulls_equal == null_equality::EQUAL and has_nulls) {
+      thrust::counting_iterator<size_type> stencil(0);
+      auto const [row_bitmask, null_count] = cudf::detail::bitmask_or(keys, stream);
+      row_validity pred{static_cast<bitmask_type const*>(row_bitmask.data())};
+
+      key_map.insert_if(iter, iter + num_rows, stencil, pred, hash_key, row_equal, stream.value());
+      return key_map.get_size() + static_cast<std::size_t>((null_count > 0) ? 1 : 0);
```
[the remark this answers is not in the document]
divyegala: `has_nulls` is acquired from `cudf::has_nested_nulls`. `null_count` is acquired from `bitmask_or` for top-level NULLs. I imagine "`has_nulls == true` but `null_count == 0`" scenario could arise for nested-types

cpp/src/stream_compaction/distinct_count.cu
```diff
@@ -125,33 +126,50 @@ cudf::size_type distinct_count(table_view const& keys,
...
 
-  // when nulls are equal, insert non-null rows only to improve efficiency
-  if (nulls_equal == null_equality::EQUAL and has_null) {
-    thrust::counting_iterator<size_type> stencil(0);
-    auto const [row_bitmask, null_count] = cudf::detail::bitmask_or(keys, stream);
-    row_validity pred{static_cast<bitmask_type const*>(row_bitmask.data())};
-
-    key_map.insert_if(iter, iter + num_rows, stencil, pred, hash_key, row_equal, stream.value());
-    return key_map.get_size() + static_cast<std::size_t>((null_count > 0) ? 1 : 0);
+  auto const comparator_helper = [&](auto const row_equal) {
```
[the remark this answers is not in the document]
divyegala: We can maybe capture the pattern to make the design look better such that each algorithm doesn't have to use these if-else blocks for nested types.. but separate code paths are inevitable since the comparator is two different types?

cpp/src/stream_compaction/distinct_count.cu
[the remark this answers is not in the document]
ttnghia: You can also use directly `static_cast<std::size_t>(null_count > 0)`. The boolean `null_count > 0` will be converted into `0/1`.

cpp/src/stream_compaction/unique_count.cu
ttnghia: Yeah this sucks. We can't call device lambda inside host lambda :angry:.

ttnghia (approved):

karthikeyann (commented):
How are benchmarks impacted by this PR?

cpp/src/stream_compaction/distinct_count.cu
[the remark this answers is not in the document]
vyasr: Yeah I think you're right, no way around that basic requirement.

cpp/src/stream_compaction/unique_count.cu
```diff
@@ -70,14 +71,24 @@ cudf::size_type unique_count(table_view const& keys,
...
-    [comp] __device__(cudf::size_type i) { return (i == 0 or not comp(i, i - 1)); });
+  auto const row_comp = cudf::experimental::row::equality::self_comparator(keys, stream);
+  if (cudf::detail::has_nested_columns(keys)) {
+    auto const comp =
+      row_comp.equal_to<true>(nullate::DYNAMIC{has_nested_nulls(keys)}, nulls_equal);
+    return thrust::count_if(
+      rmm::exec_policy(stream),
+      thrust::counting_iterator<cudf::size_type>(0),
+      thrust::counting_iterator<cudf::size_type>(keys.num_rows()),
+      [comp] __device__(cudf::size_type i) { return (i == 0 or not comp(i, i - 1)); });
```
vyasr, line 82: I assume that we can't capture the comparator by reference because it's a host object that needs to be copied to device for the lambda?

vyasr (approved):
One minor question, otherwise LGTM. Is the plan to keep moving forward with these for now and investigate compile times when we can?

cpp/src/stream_compaction/unique_count.cu
divyegala: Yes, that's correct

divyegala:
@vyasr and I followed up offline on his compile time question.

PR #12777, opened 2023-02-14
Update `tests/column_utilities` to use `experimental::equality` row comparator

divyegala:
This PR is a part of #11844 

## Checklist


cpp/tests/utilities/column_utilities.cu
```diff
@@ -371,31 +371,34 @@ struct column_property_comparator {
   }
 };
 
+template <typename DeviceComparator>
 class corresponding_rows_unequal {
  public:
-  corresponding_rows_unequal(table_device_view d_lhs,
-                             table_device_view d_rhs,
+  corresponding_rows_unequal(table_device_view /*d_lhs*/,
+                             table_device_view /*d_rhs*/,
```
ttnghia, line 378: The parameter `d_lhs` and `d_rhs` are used only here: https://github.com/rapidsai/cudf/pull/12777/files#diff-4292feecdf2f0106ee3b94779dd191033fd27c494bdffb944685c94fade6fc74R425.
This check can also be moved to the caller site. So we can remove these two parameters entirely.

cpp/tests/utilities/column_utilities.cu
```diff
@@ -410,7 +413,8 @@ class corresponding_rows_not_equivalent {
                                     table_device_view d_rhs,
                                     column_device_view lhs_row_indices_,
                                     column_device_view rhs_row_indices_,
-                                    size_type fp_ulps_)
+                                    size_type fp_ulps_,
+                                    DeviceComparator /*comp*/)
```
ttnghia, line 417: A comparator is still used here: https://github.com/rapidsai/cudf/pull/12777/files#diff-4292feecdf2f0106ee3b94779dd191033fd27c494bdffb944685c94fade6fc74R466. That `comp` also needs to be replaced by this `comp`.

cpp/tests/utilities/column_utilities.cu
```diff
@@ -536,28 +540,47 @@ struct column_comparator_impl {
...
+          *d_lhs, *d_rhs, *d_lhs_row_indices, *d_rhs_row_indices, fp_ulps, device_comparator));
+
+      differences.resize(thrust::distance(differences.begin(), diff_iter),
+                         cudf::get_default_stream());  // shrink back down
+    };
+
+    auto const comparator = cudf::experimental::row::equality::two_table_comparator{
+      lhs_tview, rhs_tview, cudf::get_default_stream()};
+    auto const has_nulls = cudf::has_nested_nulls(lhs_tview) or cudf::has_nested_nulls(rhs_tview);
+    if (cudf::detail::has_nested_columns(lhs_tview)) {
```
ttnghia, line 577: This struct is specialized to work with non-nested types. So this should not be any `if/else`.

cpp/tests/utilities/column_utilities.cu
divyegala: You are right, in fact the check is not even needed because the caller site only works with 1 column anyway

cpp/tests/utilities/column_utilities.cu
divyegala: You are right, in fact the check is not even needed because the caller site only works with 1 column anyway

cpp/tests/utilities/column_utilities.cu
```diff
@@ -459,23 +458,15 @@ class corresponding_rows_not_equivalent {
...
-      auto rhs_col = this->d_rhs.column(0);
-      return type_dispatcher(lhs_col.type(),
-                             typed_element_not_equivalent{},
-                             lhs_col,
-                             rhs_col,
-                             lhs_index,
-                             rhs_index,
-                             fp_ulps);
+      return type_dispatcher(
+        lhs.type(), typed_element_not_equivalent{}, lhs, rhs, lhs_index, rhs_index, fp_ulps);
```
ttnghia, line 469: This is no longer needed. `typed_element_not_equivalent` was needed to compare `NaN`. Now `comp` can do that.

ttnghia: I meant we also can remove `typed_element_not_equivalent` completely.

ttnghia: Humnn, probably we still need it to compare floating point numbers that are "very close".

cpp/tests/utilities/column_utilities.cu
```diff
@@ -536,28 +527,49 @@ struct column_comparator_impl {
...
+        lhs_tview, rhs_tview, cudf::get_default_stream()};
+      auto const has_nulls = cudf::has_nested_nulls(lhs_tview) or cudf::has_nested_nulls(rhs_tview);
+
+      auto const device_comparator = comparator.equal_to<false>(cudf::nullate::DYNAMIC{has_nulls});
+      comparator_helper(device_comparator);
+    } else {
+      // equivalence can be checked between column of different types,
+      // but the new comparator does not support that
+      comparator_helper(int{0});
+    }
```
ttnghia, line 572: > equivalence can be checked between column of different types

We don't check different types. `check_exact_equality` is used for checking column "equal" or "equivalent".

ttnghia:
```suggestion
      auto lhs_tview = table_view{{lhs}};
      auto rhs_tview = table_view{{rhs}};

      auto const comparator = cudf::experimental::row::equality::two_table_comparator{
        lhs_tview, rhs_tview, cudf::get_default_stream()};
      auto const has_nulls = cudf::has_nested_nulls(lhs_tview) or cudf::has_nested_nulls(rhs_tview);

      auto const device_comparator = comparator.equal_to<false>(cudf::nullate::DYNAMIC{has_nulls});
      comparator_helper(device_comparator);
```

ttnghia: We don't need to have such `comparator_helper`.

divyegala: Please look at the failures in CI https://github.com/rapidsai/cudf/actions/runs/4440315944/jobs/7794655639 in commit [d094524](https://github.com/rapidsai/cudf/pull/12777/commits/d094524d91abda6ee3eac65abd093a4d86f5fca7). `GetStructValueTest` fails because the comparator complains "cannot compare columns with different types"

ttnghia: That test is probably incorrect:
```
cudf::test::dictionary_column_wrapper<TypeParam, uint32_t> f3{42, 42, 24};

cudf::test::dictionary_column_wrapper<int32_t, TypeParam> ef3{24};
```

When `TypeParam==short`, we have `f3` and `ef3` are of different types. Let's try to fix it. The previous row comparator doesn't check input types so it didn't throw.

divyegala: Isn't that the point of these tests? It's called `TYPED_TEST(StructGetValueTestTyped, mixed_types_valid)
` so I think it is checking mixed types

ttnghia: Mixed types mean mixing these:
```
  cudf::test::fixed_width_column_wrapper<TypeParam> f1{1, 2, 3};
  cudf::test::strings_column_wrapper f2{"aa", "bbb", "c"};
  cudf::test::dictionary_column_wrapper<TypeParam, uint32_t> f3{42, 42, 24};
  LCW f4{LCW{8, 8, 8}, LCW{9, 9}, LCW{10}};
```

cpp/tests/utilities/column_utilities.cu
```diff
@@ -536,28 +531,43 @@ struct column_comparator_impl {
...
+      auto diff_iter = thrust::copy_if(
+        rmm::exec_policy(cudf::get_default_stream()),
+        input_iter,
+        input_iter + lhs_row_indices.size(),
+        differences.begin(),
+        ComparatorType(
+          *d_lhs_row_indices, *d_rhs_row_indices, fp_ulps, device_comparator, *d_lhs, *d_rhs));
+
+      differences.resize(thrust::distance(differences.begin(), diff_iter),
+                         cudf::get_default_stream());  // shrink back down
```
davidwendt, line 559: To improve build time, this `thrust::copy_if` could be a candidate for replacing with `thrust::transform` and `thrust::remove`
The existing `differences` can be used to hold the transform result.

```
thrust::transform( rmm::exec_policy(cudf::get_default_stream()),
                   input_iter,
                   input_iter + lhs_row_indices.size(),
                   differences.begin(),
                   ComparatorType(*d_lhs_row_indices, *d_rhs_row_indices, fp_ulps, device_comparator, *d_lhs, *d_rhs));

auto diff_iter = thrust::remove( 
             rmm::exec_policy(cudf::get_default_stream()),
             differences.begin(), differences.end(), 0); // remove the zero entries
```

divyegala:
@ttnghia all the mixed type test failures are now solved, these are the only failing tests. Any advice would be appreciated:
```
[ RUN      ] FixedPointTestAllReps/0.SimpleFixedPointColumnWrapper
unknown file: Failure
C++ exception with description "CUDF failure at: /opt/conda/conda-bld/work/cpp/src/table/row_operators.cu:337: Cannot compare tables with different column types" thrown in the test body.

[ RUN      ] StringsConvertTest.ToFixedPointLargeScale
unknown file: Failure
C++ exception with description "CUDF failure at: /opt/conda/conda-bld/work/cpp/src/table/row_operators.cu:337: Cannot compare tables with different column types" thrown in the test body.

[ RUN      ] JsonPathTests.GetJsonObjectEmptyQuery
unknown file: Failure
C++ exception with description "CUDF failure at: /opt/conda/conda-bld/work/cpp/src/strings/strings_column_view.cpp:31: strings column has no children" thrown in the test body.

[ RUN      ] StringsSplitTest.AllNullsCase
unknown file: Failure
C++ exception with description "CUDF failure at: /opt/conda/conda-bld/work/cpp/src/strings/strings_column_view.cpp:31: strings column has no children" thrown in the test body.
```

divyegala:
@davidwendt according to your baseline report, `column_utilities.cu` takes around ~14 minutes to build. In this PR, it takes around ~10 minutes https://downloads.rapids.ai/ci/cudf/pull-request/12777/121e45b/cpp_cuda11_x86_64.BuildMetricsReport.html

ttnghia:
The first two failed tests compare decimals with different scales. I'm not sure if we should continue to support it or not (for the new comparator). Please initialize a discussion on Slack to get more feedback for it.

The last two failed tests are due to comparing strings in a strings column having all-nulls (which doesn't have a child column). I believe this is a bug. So you need to file a bug issue for it, then either fix that bug before this PR can proceed, or xfail these two tests.

cpp/tests/utilities/column_utilities.cu
```diff
@@ -536,28 +531,43 @@ struct column_comparator_impl {
...
+
+    auto lhs_tview = table_view{{lhs}};
+    auto rhs_tview = table_view{{rhs}};
+
+    auto const comparator = cudf::experimental::row::equality::two_table_comparator{
+      lhs_tview, rhs_tview, cudf::get_default_stream()};
+    auto const has_nulls = cudf::has_nested_nulls(lhs_tview) or cudf::has_nested_nulls(rhs_tview);
+
+    auto const device_comparator = comparator.equal_to<false>(cudf::nullate::DYNAMIC{has_nulls});
+    comparator_helper(device_comparator);
```
ttnghia, line 570: Since this lambda is called just once, we don't have to have such lambda. Just use the code inside directly.

davidwendt:
> @davidwendt according to your baseline report, `column_utilities.cu` takes around ~14 minutes to build. In this PR, it takes around ~10 minutes https://downloads.rapids.ai/ci/cudf/pull-request/12777/121e45b/cpp_cuda11_x86_64.BuildMetricsReport.html

Thanks. I'd still like to consider this change: https://github.com/rapidsai/cudf/pull/12777#discussion_r1142681494

davidwendt:
For the fixed-point error, it appears that the intention was to be able to compare fixed-point values in different scales. The `FixedPointTestAllReps/SimpleFixedPointColumnWrapper` gtests specifically tests this.
I propose to relax the check here: https://github.com/rapidsai/cudf/blob/84ac1d5a87859bd96bd767b5a65b017e260024a1/cpp/src/utilities/type_checks.cpp#L66-L70
To only check the type id by changing the if-statement there to:
```
if (lhs.type().id() != rhs.type().id()) { return false; }
```

cpp/tests/utilities_tests/type_check_tests.cpp
```diff
@@ -147,6 +147,7 @@ TEST_F(ColumnTypeCheckTest, DifferentFixedWidth)
   fixed_point_column_wrapper<int32_t> rhs5({10000}, numeric::scale_type{0});
 
   EXPECT_FALSE(column_types_equal(lhs5, rhs5));
+  EXPECT_TRUE(column_types_equivalent(lhs5, rhs5));
```
davidwendt, line 150: Nice.

davidwendt (approved):

ttnghia (approved):

Issue #3958, opened 2020-01-28
[FEA] Make cudf::size_type 64-bit


harrism:
**Is your feature request related to a problem? Please describe.**
`cudf::size_type` is currently an `int32_t`, which limits column size to two billion elements (`MAX_INT`). Moreover, it limits child column size to the same. This causes problems, for example, for string columns, where there may be fewer than 2B strings, but the character data to represent them could easily exceed 2B characters.

A 32-bit size was originally chosen to ensure compatibility with Apache Arrow, which dictates that Arrow arrays have a 32-bit size, and that larger arrays are made by chunking into individual Arrays.

**Describe the solution you'd like**

 * Change `size_type` to be an `int64_t`.

 * Handle compatibility with Arrow by creating arrow chunked arrays in the libcudf `to_arrow` interface (not yet created), and combine arrow chunked arrays in the libcudf `from_arrow` interface. This can be dealt with when we create these APIs.

**Describe alternatives you've considered**

Chunked columns. This would be very challenging -- supporting chunked columns in every algorithm would result in complex distributed algorithms and implementations, where libcudf currently aims to be communication agnostic / ignorant. In other words, a higher level library handles distributed algorithms.

**Additional context**

A potential downside: @felipeblazing called us brave for considering supporting chunked columns. If we implement this feature request, perhaps he will not consider us quite so brave. :(

kkraus14:
> A potential downside: @felipeblazing called us brave for considering supporting chunked columns. If we implement this feature request, perhaps he will not consider us quite so brave. :(

🤣

kkraus14:
> Handle compatibility with Arrow by creating arrow chunked arrays in the libcudf to_arrow interface (not yet created), and combine arrow chunked arrays in the libcudf from_arrow interface. This can be dealt with when we create these APIs.

We have these APIs on the Python side already, so we'd ideally need these APIs sooner rather than later, though I do believe the PyArrow CPU side will handle splitting chunks for us.

felipeblazing:
This seems like a really reasonable way to ensure that we are compatible with larger columns almost for free. If we want to revisit later that might make sense but right now I can imagine that the discussions alone for going to chunked representation would be long.

revans2:
[...]

PR #12779, opened 2023-02-15
Improve performance for cudf::strings::count_characters for long strings

davidwendt:
## Description
Adds more efficient counting algorithm specifically for columns with long strings--greater than 64 bytes on average.
The internal detail method will be used to help improve performance in other strings functions.

## Checklist


davidwendt:
Benchmark results show most improvement for string widths greater 64 bytes on average.
```
|   rows   | width|   Ref Time |    Cmp Time |          Diff |   %Diff |
|----------|------|------------|-------------|---------------|---------|
|     4096 |   32 |  37.670 us |   37.532 us |     -0.138 us |  -0.37% |
|    32768 |   32 |  38.356 us |   38.225 us |     -0.131 us |  -0.34% |
|   262144 |   32 |  49.001 us |   48.659 us |     -0.342 us |  -0.70% |
|  2097152 |   32 | 332.821 us |  333.535 us |      0.714 us |   0.21% |
| 16777216 |   32 |   2.418 ms |    2.434 ms |     15.780 us |   0.65% |
|     4096 |   64 |  38.041 us |   38.462 us |      0.421 us |   1.11% |
|    32768 |   64 |  39.032 us |   39.507 us |      0.475 us |   1.22% |
|   262144 |   64 |  56.326 us |   56.501 us |      0.175 us |   0.31% |
|  2097152 |   64 | 381.756 us |  380.597 us |     -1.159 us |  -0.30% |
| 16777216 |   64 |   2.808 ms |    2.789 ms |    -19.355 us |  -0.69% |
|     4096 |  128 |  41.175 us |   41.297 us |      0.122 us |   0.30% |
|    32768 |  128 |  45.130 us |   45.286 us |      0.156 us |   0.35% |
|   262144 |  128 |  77.200 us |   77.078 us |     -0.122 us |  -0.16% |
|  2097152 |  128 | 497.075 us |  492.847 us |     -4.228 us |  -0.85% |
|     4096 |  256 |  49.684 us |   37.710 us |    -11.974 us | -24.10% |
|    32768 |  256 |  53.379 us |   49.197 us |     -4.182 us |  -7.83% |
|   262144 |  256 | 152.252 us |  139.544 us |    -12.709 us |  -8.35% |
|  2097152 |  256 | 963.613 us |    1.090 ms |    126.674 us |  13.15% |
|     4096 |  512 |  69.779 us |   41.161 us |    -28.618 us | -41.01% |
|    32768 |  512 |  75.668 us |   55.073 us |    -20.595 us | -27.22% |
|   262144 |  512 | 405.729 us |  166.213 us |   -239.516 us | -59.03% |
|  2097152 |  512 |   2.551 ms |    1.297 ms |  -1254.632 us | -49.18% |
|     4096 | 1024 | 106.060 us |   42.641 us |    -63.418 us | -59.79% |
|    32768 | 1024 | 127.396 us |   68.891 us |    -58.505 us | -45.92% |
|   262144 | 1024 |   1.223 ms |  237.241 us |   -986.090 us | -80.61% |
|     4096 | 2048 | 172.035 us |   46.194 us |   -125.841 us | -73.15% |
|    32768 | 2048 | 228.790 us |   94.118 us |   -134.671 us | -58.86% |
|   262144 | 2048 |   6.446 ms |  422.672 us |  -6023.192 us | -93.44% |
|     4096 | 4096 | 309.562 us |   52.891 us |   -256.671 us | -82.91% |
|    32768 | 4096 | 425.600 us |  138.327 us |   -287.273 us | -67.50% |
|   262144 | 4096 |  14.901 ms |  797.709 us | -14103.759 us | -94.65% |
```

ttnghia:
Will the performance with strings shorter than 64 bytes be impacted?

davidwendt:
> Will the performance with strings shorter than 64 bytes be impacted?

No. The optimized code did not perform as well as the original code under 64 bytes so the solution implemented here is to include both code paths.

cpp/src/strings/attributes.cu
```diff
@@ -85,21 +102,83 @@ std::unique_ptr<column> counts_fn(strings_column_view const& strings,
...
+  auto const d_lengths = results->mutable_view().data<size_type>();
+  auto const d_strings = cudf::column_device_view::create(input.parent(), stream);
+
+  // fill in the lengths
+  constexpr int block_size = 256;
+  cudf::detail::grid_1d grid{input.size() * cudf::detail::warp_size, block_size};
+  count_characters_parallel_fn<<<grid.num_blocks, grid.num_threads_per_block, 0, stream.value()>>>(
+    *d_strings, d_lengths);
+
+  results->set_null_count(input.null_count());  // reset null count
```
ttnghia, line 158: This is very easy to be overlooked. Without detailed comment, somebody may mistakenly consider it as redundant and delete it.
```suggestion
  // Reset null count, as after calling to `mutable_view()` the internal null count is set to `UNKNOWN_NULL_COUNT`.
  results->set_null_count(input.null_count());
```

ttnghia (approved):

cpp/src/strings/attributes.cu
```diff
@@ -85,21 +102,85 @@ std::unique_ptr<column> counts_fn(strings_column_view const& strings,
...
+  auto const d_lengths = results->mutable_view().data<size_type>();
+  auto const d_strings = cudf::column_device_view::create(input.parent(), stream);
+
+  // fill in the lengths
+  constexpr int block_size = 256;
+  cudf::detail::grid_1d grid{input.size() * cudf::detail::warp_size, block_size};
+  count_characters_parallel_fn<<<grid.num_blocks, grid.num_threads_per_block, 0, stream.value()>>>(
+    *d_strings, d_lengths);
+
+  // reset null count after call to  mutable_view()
```
bdice, line 158:
```suggestion
  // reset null count after call to mutable_view()
```

cpp/src/strings/attributes.cu
```diff
@@ -85,21 +102,85 @@ std::unique_ptr<column> counts_fn(strings_column_view const& strings,
...
 }  // namespace
 
-std::unique_ptr<column> count_characters(strings_column_view const& strings,
+std::unique_ptr<column> count_characters(strings_column_view const& input,
                                          rmm::cuda_stream_view stream,
                                          rmm::mr::device_memory_resource* mr)
 {
-  auto ufn = [] __device__(const string_view& d_str) { return d_str.length(); };
-  return counts_fn(strings, ufn, stream, mr);
+  if ((input.size() == input.null_count()) ||
```
bdice, line 170: Add a bit of commentary on why we have the branch here.
```suggestion
  // If the average string size is small, we use one thread per string.
  // For large strings, we use one warp per string.
  if ((input.size() == input.null_count()) ||
```

cpp/tests/strings/attrs_tests.cpp
```diff
@@ -93,3 +92,16 @@ TEST_F(StringsAttributesTest, StringsLengths)
...
+
+TEST_F(StringsAttributesTest, StringsLengthsLong)
+{
+  std::vector<std::string> h_strings(
+    40000, "something a bit longer than 32 bytes ééé ééé ééé ééé ééé ééé ééé");
+  cudf::test::strings_column_wrapper strings(h_strings.begin(), h_strings.end());
+  auto strings_view = cudf::strings_column_view(strings);
+
+  auto results = cudf::strings::count_characters(strings_view);
+  std::vector<int32_t> h_expected(h_strings.size(), 64);
```
bdice, line 104: I expected that these tests should all be using `cudf::size_type` for the counts, but I also see that `count_characters` is defined to return an `INT32` in its docstring. That seems undesirable, right? Shouldn't we be using `cudf::size_type` instead of hardcoding `int32_t` for the function return value and test comparisons?

bdice (changes requested):

cpp/src/strings/attributes.cu
davidwendt: There is already a complete comment on the declaration of `AVG_CHAR_BYTES_THRESHOLD` in this file.
I'd rather not repeat it here.

cpp/tests/strings/attrs_tests.cpp
davidwendt: Since there is no `SIZE_TYPE` column type, I think it is more correct to use the appropriate data type.

cpp/src/strings/attributes.cu
bdice: Sounds good.

cpp/tests/strings/attrs_tests.cpp
bdice: If changing the definition of size type would affect the correctness or consistency of this code, we should use size type.

davidwendt: I disagree but this seems out of scope for this PR.
There are at least 6 APIs that do this today.
Perhaps you can create a separate issue if you want to reopen the discussion.

We document what data-type is used for the output of the column of these APIs.
I'm not comfortable saying the column type is whatever size-type equates to at the moment.

bdice: Discussed a bit offline. The latest documented status is https://github.com/rapidsai/cudf/issues/3958#issuecomment-1166338595 but from conversation since that was posted, @GregoryKimball seemed interested in moving towards option 3. I won't hold this PR up on that account.

bdice (approved):

PR #12783, opened 2023-02-15
[REVIEW] Add `GroupBy.dtypes`

galipremsagar:
## Description
This PR adds `dtypes` property to `GroupBy`, this will also fix some upstream dask breaking changes introduced in: https://github.com/dask/dask/pull/9889

Issue was discovered in: https://github.com/rapidsai/cudf/pull/12768#issuecomment-1430650703

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12796, opened 2023-02-16
Tell cudf_kafka to use header-only fmt

vyasr:
## Description

The changes to spdlog/fmt packaging in rmm caused an undefined symbol issue in cudf_kafka. To unblock CI in https://github.com/rapidsai/cudf/pull/12783, I simply added fmt to the list of required libraries for the Python package (see https://github.com/rapidsai/cudf/pull/12783/commits/caf7adfb9a89e585bd4d3239c79643328449a242 and [the explanation](https://github.com/rapidsai/cudf/pull/12783#issuecomment-1431997720)). The underlying issue turns out to be that by default usage of fmt results in the dependent assuming that the headers are compiled in not header-only mode. When using CMake to manage the build, fmt relies on use of the appropriate target `fmt::fmt-header-only` to configure the build such that anything using fmt knows to use it in header-only mode, which sets the `FMT_HEADER_ONLY` preprocessor macro under the hood. Since the Python cudf_kafka package is not built using CMake, however, this information was not propagated to its build from its dependencies (libcudf/libcudf_kafka). As a result, it was compiled expecting an external definition of some fmt symbols rather than inlining them. This PR removes the undesirable library dependency on fmt introduced in https://github.com/rapidsai/cudf/pull/12783 in favor of properly telling fmt to expect inlined symbols.

## Checklist

### DIFF
--- a/python/cudf_kafka/setup.py
+++ b/python/cudf_kafka/setup.py
@@ -79,9 +79,9 @@ extensions = [
                 CUDF_KAFKA_ROOT,
             ]
         ),
-        libraries=["cudf", "cudf_kafka", "fmt"],
+        libraries=["cudf", "cudf_kafka"],
         language="c++",
-        extra_compile_args=["-std=c++17"],
+        extra_compile_args=["-std=c++17", "-DFMT_HEADER_ONLY=1"],
     )
 ]
 


galipremsagar (approved):

bdice (approved):
Thank you for the explanation in the PR description! This change looks good, and the rationale makes total sense to me. I am not aware of any other non-CMake consumers of libcudf. Should we document somehow that libcudf requires this to be defined for builds not consuming libcudf via CMake?

mroeschke (approved):

vyasr:
> Thank you for the explanation in the PR description! This change looks good, and the rationale makes total sense to me. I am not aware of any other non-CMake consumers of libcudf besides cudf_kafka.
> 
> Should we document somehow that libcudf requires this to be defined for builds not consuming libcudf via CMake?

This isn't even libcudf, it's coming from librmm. I had a chat with Rob about this and we have open questions about whether it should be possible to mix header-only fmt with library fmt. He's run into related issues in other places recently as well, so figuring out the best approach will be a follow-up task.

PR #12797, opened 2023-02-16
Update required spdlog and fmt version for our conda envs

robertmaynard:
## Description
This ensures that we don't have older versions of these packages installed causing rapids-cmake to download the newer versions.

## Checklist

### DIFF
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -26,6 +26,7 @@ dependencies:
 - dlpack>=0.5,<0.6.0a0
 - doxygen=1.8.20
 - fastavro>=0.22.9
+- fmt>=9.1.0,<10
 - fsspec>=0.6.0
 - gcc_linux-64=9.*
 - hypothesis
@@ -66,6 +67,7 @@ dependencies:
 - s3fs>=2022.3.0
 - scikit-build>=0.13.1
 - scipy
+- spdlog>=1.11.0,<1.12
 - sphinx
 - sphinx-autobuild
 - sphinx-copybutton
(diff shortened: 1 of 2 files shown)


vyasr:
@robertmaynard you'll want to run `rapids-dependency-file-generator` to update all the necessary files (e.g. conda env files). It can be pip installed and then run from the command line in the root of the repo.

robertmaynard:
> @robertmaynard you'll want to run `rapids-dependency-file-generator` to update all the necessary files (e.g. conda env files). It can be pip installed and then run from the command line in the root of the repo.

Thanks, updated with the output. I couldn't remember if we did that through some sort of merge time magic.

vyasr:
Not quite yet, but we're slowly moving in that direction. Should have pre-commit hooks in place for it very soon at least. pre-commit autofixes in CI are still a ways away though, and I don't recall from our debates about it whether people were generally in favor of that or not (due to the possible divergence between local and remote changes from a remote run of the hook). Probably a debate we'll have again once we're closer to being able to enable such a feature.

vyasr (approved):

ajschmidt8 (approved):

vyasr:
@robertmaynard out of curiosity, if cudf pulls the latest rmm won't we just inherit these requirements from rmm? Also, this will affect #12637 CC @vuule

ajschmidt8:
Do these packages need to be added to the conda recipes at all?

Right now that `build` dependencies list in `dependencies.yaml` is really only used for downstream projects like `rapids-compose` and `rapids-compose v2`

robertmaynard:
I added these to the dependencies.yaml as I was running into conda envs that had spdlog 7 in them.

Looking at the conda envs more closely it looks like that this is due to them bringing librmm 23.02 for some reason instead of librmm 23.04. 

I think we can close this PR as it isn't needed

vyasr:
OK good so cudf should indeed be pulling the requirement from rmm as I would have expected. I will close this for now, we can reopen if the issue reappears.

PR #12800, opened 2023-02-17
Add compute-sanitizer github workflow action to nightly tests

davidwendt:
## Description
Adds github workflow action to the nightly tests for running `compute-sanitizer` on the libcudf gtests.

Reference: #12530 

## Checklist


ci/test_cpp_memcheck.sh
```diff
@@ -0,0 +1,61 @@
...
+nvidia-smi
+
+EXITCODE=0
+trap "EXITCODE=1" ERR
+set +e
+
+# Run gtests with compute-sanitizer
+echo "RAPIDS_BUILD_TYPE=${RAPIDS_BUILD_TYPE}"
+
+if [[ "${RAPIDS_BUILD_TYPE}" == "nightly" ]]; then
```
ajschmidt8, line 41: The `test.yaml` workflow which runs this script is only ever executed for nightly tests.

Therefore this conditional block can be removed.

ci/test_cpp_memcheck.sh
```diff
@@ -0,0 +1,61 @@
...
+mkdir -p "${RAPIDS_TESTS_DIR}"
+
+rapids-print-env
+
+rapids-mamba-retry install \
+  --channel "${CPP_CHANNEL}" \
+  libcudf libcudf_kafka libcudf-tests
+
+rapids-logger "Check GPU usage"
+nvidia-smi
```
ajschmidt8, line 32: It looks like a lot of this code is the same as what's present in the `test_cpp.sh` script.

I think we can DRY this up a bit.

Take a look at the Python test scripts for `cudf` below, which make use of a `test_python_common.sh` script to remove duplication like this.

- https://github.com/rapidsai/cudf/blob/branch-23.04/ci/test_python_common.sh
- https://github.com/rapidsai/cudf/blob/branch-23.04/ci/test_python_cudf.sh
- https://github.com/rapidsai/cudf/blob/branch-23.04/ci/test_python_other.sh

ajschmidt8 (commented):
Few minor comments.

Looks great otherwise.

Nice job Dave.

dependencies.yaml
```diff
@@ -266,6 +268,14 @@ dependencies:
...
             packages:
               - cupy-cuda11x -f https://pip.cupy.dev/aarch64 # TODO: Verify that this works.
+  test_cpp:
+    specific:
+      - output_types: conda
+        matrices:
+          - matrix:
+              cuda: "11.8"
+            packages:
+              - cuda-sanitizer-api=11.8.86
```
ajschmidt8, line 278:
```suggestion
          - matrix:
              cuda: "11.8"
            packages:
              - cuda-sanitizer-api=11.8.86
          - matrix:
            packages:
```

By default, the `dependency-file-generator` tool requires that every matrix variation be covered in `specific` dependency lists like this one.

To override that, you can add an empty matrix option, which basically says "For all other matrix variations, use this `packages` list" (where this `packages` list happens to be empty in this case).

This is why CI is currently failing for this PR.

ci/test_cpp_memcheck.sh
davidwendt: Done. Let me know what you think.

ajschmidt8 (approved):

ci/test_cpp_memcheck.sh
```diff
@@ -0,0 +1,29 @@
...
+for gt in "$CONDA_PREFIX"/bin/gtests/{libcudf,libcudf_kafka}/* ; do
+    test_name=$(basename ${gt})
+    if [[ "$test_name" == "ERROR_TEST" ]]; then
+        continue
+    fi
+    if [[ "$test_name" == "STREAM_IDENTIFICATION_TEST" ]]; then
+        continue
+    fi
+    echo "Running gtest $test_name"
+    echo "${COMPUTE_SANITIZER_CMD} ${gt} --gtest_output=xml:${RAPIDS_TESTS_DIR}${test_name}.xml"
```
vyasr, line 23: Maybe store the arguments (`${gt}` onward) into a variable so that you don't have to copy-paste it below? I can totally see someone needing to make some minor modification in the future and forgetting to do it on both lines.

ci/test_cpp_memcheck.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+source "$(dirname "$0")/test_cpp_common.sh"
```
vyasr, line 4: Same question as before about the quotes, I don't think this does quite what we want. If, for instance, there are any spaces in the name `$0` the result here will be having them unquoted, which is usually the kind of thing you want to use quotes for.

dependencies.yaml
```diff
@@ -266,6 +268,16 @@ dependencies:
...
```
vyasr, line 278: This is a very specific pinning, any reason we can't do a >=?

vyasr (approved):
Couple of minor suggestions, but nothing blocking. Feel free to resolve as you see fit. Excited to have this running again!

ci/test_cpp.sh
```diff
@@ -1,35 +1,7 @@
...
-
-rapids-print-env
-
-rapids-mamba-retry install \
-  --channel "${CPP_CHANNEL}" \
-  libcudf libcudf_kafka libcudf-tests
-
-rapids-logger "Check GPU usage"
-nvidia-smi
+source "$(dirname "$0")/test_cpp_common.sh"
```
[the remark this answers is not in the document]
davidwendt: Copied from this line: https://github.com/rapidsai/cudf/blob/d077c9b2a3764818fb690995f2c2867a2c024811/ci/test_python_cudf.sh#L5
per instructions from AJ in this comment:
https://github.com/rapidsai/cudf/pull/12800#discussion_r1110043072

ci/test_cpp_memcheck.sh
davidwendt: I'll just remove the echo line

dependencies.yaml
davidwendt: This was stolen from here: https://github.com/rapidsai/cudf/pull/12530/files#diff-5475a6d76de4c506ee92cf6f941bba30a6a07b5881cfea531d42a6ec035095a6R278
cc @bdice

ci/test_cpp_memcheck.sh
davidwendt: @ajschmidt8 
https://github.com/rapidsai/cudf/blob/d077c9b2a3764818fb690995f2c2867a2c024811/ci/test_python_cudf.sh#L5

dependencies.yaml
bdice: @vyasr We don't want the compute sanitizer from CUDA 12 to be run with our CUDA 11.8 builds. We need to pin this.

bdice: We could pin like suggested below, but I wouldn't advise it. We have been pinning exact dependencies on CUDA packages in our conda build recipes like shown here: https://github.com/rapidsai/raft/blob/branch-23.04/conda/recipes/libraft/conda_build_config.yaml
```suggestion
              - cuda-sanitizer-api=11.8.*
```

bdice (approved):
This looks good to me, aside from @vyasr's suggestions.

ci/test_cpp_memcheck.sh
davidwendt: Looks like this is correct for command substitution. Here are a couple references: 
https://unix.stackexchange.com/questions/289574/nested-double-quotes-in-assignment-with-command-substitution
https://unix.stackexchange.com/questions/118433/quoting-within-command-substitution-in-bash

dependencies.yaml
ajschmidt8: I thought about this myself too and figured it was best left as-is.

The downside to this pin is that we'll need to update the version and corresponding `matrix` values in the `dependencies.yaml` file anytime the `rapidsai/ci:latest` image is updated (which is used [here](https://github.com/rapidsai/cudf/pull/12800/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692R35)).

Right now the `latest` tag points to an `11.8` image. When we update that image to CUDA 12 for instance, then `cuda-santizer-api` will no longer get installed and the job will fail.

bdice: That's correct -- but that's already true for every other part of our CI. We need to keep our conda recipes (and conda build configs) **and** dependencies.yaml matrix values up to date for every CUDA update we make. I don't think this is a new requirement, it's just another instance of things we must already do.

ci/test_cpp_memcheck.sh
vyasr: Ah yes pesky command substitution rules. I never remember any of this. Thanks for the refs.

dependencies.yaml
vyasr: I guess I was mostly looking to float forward minor versions while pinning the major version i.e. ` ==11.*` but if we're being consistent here that's fine with me.

PR #12802, opened 2023-02-17
Cache JIT `GroupBy.apply` functions

brandon-b-miller:
This PR sends incoming UDFs that go through the `engine='jit'` codepath through the main UDF cache. This should avoid recompiling if a user reuses the same UDF on different input data, so long as the types of that data are the same.


bdice:
Can we merge the big move PR #12669 before doing feature work on these areas of the code?

brandon-b-miller:
with the merge of https://github.com/rapidsai/cudf/pull/12669 this should be ready for review.

bdice (commented):
Is it possible to test the caching behavior?

brandon-b-miller:
> Is it possible to test the caching behavior?

What do you think about the approach [here](https://github.com/rapidsai/cudf/blob/branch-23.04/python/cudf/cudf/tests/test_udf_masked_ops.py#L714-L743) ?

bdice:
> > Is it possible to test the caching behavior?
> 
> What do you think about the approach [here](https://github.com/rapidsai/cudf/blob/branch-23.04/python/cudf/cudf/tests/test_udf_masked_ops.py#L714-L743) ?

Looks good to me, that would be a good test if you copied and altered it for the groupby case.

python/cudf/cudf/core/udf/groupby_utils.py
```diff
@@ -147,12 +149,18 @@ def jit_groupby_apply(offsets, grouped_values, function, *args):
     offsets = cp.asarray(offsets)
     ngroups = len(offsets) - 1
 
-    kernel, return_type = _get_groupby_apply_kernel(
-        grouped_values, function, args
-    )
-    return_type = numpy_support.as_dtype(return_type)
+    cache_key = _generate_cache_key(grouped_values, function)
```
bdice, line 152: Is there a reason we don't use `lru_cache` and instead manually track cache keys? I assume it has to do with types being supported in `lru_cache` keys?

python/cudf/cudf/core/udf/groupby_utils.py
```diff
@@ -147,12 +149,18 @@ def jit_groupby_apply(offsets, grouped_values, function, *args):
...
+    if precompiled.get(cache_key) is not None:
+        kernel, return_type = precompiled[cache_key]
+    else:
+        kernel, return_type = _get_groupby_apply_kernel(
+            grouped_values, function, args
+        )
+        precompiled[cache_key] = (kernel, return_type)
```
bdice, line 159: Fill the cache if it's empty, then read the cache. This reduces duplication.
```suggestion
    if cache_key not in precompiled:
        precompiled[cache_key] = _get_groupby_apply_kernel(
            grouped_values, function, args
        )
    kernel, return_type = precompiled[cache_key]
```

bdice (commented):
Test looks good but I think we can improve the cache itself.

shwina (approved):
Approving pending @bdice's improvement

python/cudf/cudf/core/udf/groupby_utils.py
brandon-b-miller: In this context, `precompiled` is a `cachetools.LRUCache`. Are you asking why we don't do the following from functools?
```python
@functools.lru_cache
def _get_groupby_apply_kernel(...)
```

If so the reason was that I wanted different UDF pipelines (`apply`, `groupby.apply` etc) to share the same cache.

bdice: Nevermind. 😄 I didn't look closely enough at `precompiled`. But to clarify, how do you distinguish the type of UDF? Could an apply function and a groupby apply function reuse the same exact kernel? If not, how are the cache keys distinguished (for functions of the same data types)?

brandon-b-miller: The cache key is based on the bytecode of the UDF, the particulars are found [here](https://github.com/rapidsai/cudf/blob/branch-23.04/python/cudf/cudf/utils/cudautils.py#L193). I suppose you could get a cache hit erroneously if you:
1. wrote a function `f` and executed it using `apply`
2. applied the exact same `f` on a `groupby` result whose columns were the exact same dtypes as the dataframe that it was first applied to

However I would expect the above to cause a crash in pandas case as well since each API enforces a different syntax for the kinds of UDFs it accepts, so using one kind of function with the other's `apply` API probably wouldn't work in most cases.

bdice: I'll defer to your judgment here -- but distinguishing keys clearly would be a plus, in my eyes. An erroneous cache hit would be bad.

bdice (approved):

python/cudf/cudf/core/udf/groupby_utils.py
brandon-b-miller: I added an extra tuple element to UDFs that go through `GroupBy.apply` that should break this degeneracy. 

https://github.com/rapidsai/cudf/pull/12802/commits/425a91282d8573710518208902cf7c80b23d740b

bdice: Nice. Thanks!

bdice: fwiw, I think I would have preferred an approach like `_generate_cache_key(grouped_values, function, suffix="__GROUPBY_APPLY_UDF")` where you provide a suffix to the function making the key. Not a dealbreaker but worth considering if we have more JIT code paths with separate JIT caches.

PR #12805, opened 2023-02-18
Make timezone table independent from ORC

vuule:
## Description
Modifies the creation of timezone transition tables to make it usable outside of the ORC context:

- Remove the ORC epoch offset from the transition table.
- Use `timestamp_s` and `duration_s` instead of integral values in the table.
- Return table as `cudf::table` instead of the special `timezone_table` struct.
- Move the implementation/header out of `cudf::io::orc`.
- Split the header into C++ and CUDA parts. C++ API has public and `detail` counterparts.

Other:
Adapt ORC reader to changes to the timezone transition table.

## Checklist


ttnghia:
I see that the timezone stuff is in the `cudf/orc/` folder? I believe that they are very independent, so please consider refactoring it out into somewhere else not in `cudf/orc/`, like `cudf/utilities/`. This would be useful for implementing other timezone-aware operations.

vuule:
> I see that the timezone stuff is in the `cudf/orc/` folder? I believe that it is very independent from ORC file format, so please consider refactoring it out into somewhere else not in `cudf/orc/`, like `cudf/utilities/`. This would be useful for implementing other timezone-aware operations.

This PR was a first draft just to unblock @shwina on the Python side. I've since split the headers to separate C++ and CUDA code and moved them to /include/detail. The namespace was also changed to cudf::detail.

vuule:
CC @shwina to confirm that new API and header location are suitable for Cython use.

shwina:
@vuule Thanks! while technically the new location is OK for Cython, can the function be moved out of the `detail` namespace?

vuule:
> @vuule Thanks! while technically the new location is OK for Cython, can the function be moved out of the `detail` namespace?

I placed it under `detail` because it's not really intended for end users. This is based on an offline discussion with @galipremsagar . I'm fine with moving it out of detail, the `detail` namespace is only meant to discourage direct use here.

shwina:
I think any code that is used outside of `libcudf` should be in its public API (even if that is just cuDF Python).

This avoids situations in the future where someone might decide to drop this code from `libcudf` because it's no longer being used internally, OR, a breaking API change is made but not considered breaking because it's not part of the public API. 

Perhaps it could be excluded from the doxygen-generated docs, so that libcudf consumers do not discover it?

ajschmidt8 (approved):

cpp/include/cudf/timezone.hpp
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param tzif_dir The directory where the TZif files are located
+ * @param timezone_name standard timezone name (for example, "US/Pacific")
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ *
+ * @return The transition table for the given timezone
+ */
+std::unique_ptr<table> make_timezone_transition_table(std::optional<std::string_view> tzif_dir,
+                                                      std::string_view timezone_name,
+                                                      rmm::cuda_stream_view stream);
```
shwina, line 47: Should this argument have a default value of `stream_default`?

shwina: Actually, I think this argument should just be eliminated from the public API

mythrocks: I've lost track of where we stand on eventually exposing `stream` parameters from non-`detail` public functions. Would removing the parameter mean having to add it back eventually?
Perhaps adding a default value is an acceptable compromise?

cpp/include/cudf/detail/timezone.cuh
```diff
@@ -0,0 +1,74 @@
...
+
+    auto const file_entry_end = ttimes.begin() + (ttimes.size() - cycle_entry_cnt);
+
+    if (ts <= *(file_entry_end - 1)) {
+      // Search the file entries if the timestamp is in range
+      return last_less_equal(ttimes.begin(), file_entry_end, ts);
+    } else {
+      auto project_to_cycle = [](timestamp_s ts) {
+        static constexpr duration_s cycle_s =
+          cuda::std::chrono::duration_cast<duration_s>(duration_D{365 * cycle_years + (100 - 3)});
```
mythrocks, line 63: ELI5: What's the calculation here? I'm afraid I don't follow.

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -49,7 +49,7 @@ constexpr int scratch_buffer_size = 512 * 4;
 // Workaround replaces zero-length patch lists by a dummy zero patch
 constexpr bool zero_pll_war = true;
 
-static __device__ __constant__ int64_t kORCTimeToUTC =
+static __device__ __constant__ int64_t orc_utc_epoch =
```
mythrocks, line 52: This constant looks similar to the `duration_s` in `stripe_data.cu`. I'm not sure I've understood why we have two constants for the same value.

cpp/include/cudf/timezone.hpp
shwina: cc: @vyasr

cpp/include/cudf/detail/timezone.cuh
vuule: It codifies the leap year rules:


In our Gregorian calendar three criteria must be taken into account to identify leap years:
The year must be evenly divisible by 4;
If the year can also be evenly divided by 100, it is not a leap year;
unless...
The year is also evenly divisible by 400. Then it is a leap year.

I believe this translates to 100 (every fourth year) - 4 (every hundredth year ) + 1 (every four hundredth year) leap days in 400 years.

cpp/include/cudf/timezone.hpp
vuule: yeah, now that I look at this as a regular libcudf API, it should not have a stream parameter. I'll add a `detail` counterpart that has the defaulted stream parameter.

cpp/src/io/orc/stripe_enc.cu
vuule: it is the same constant, just for the reader and the writer. I'll try to move it to a common header.

cpp/include/cudf/timezone.hpp
vyasr: Yeah I would remove the stream. As for where we stand, the answer is eventually but with no concrete timeline yet. We're making incremental progress towards it but for now I think the only public APIs exposing streams are ones that a privileged user (typically cuspatial) has explicitly requested as necessary.

cpp/include/cudf/detail/timezone.cuh
```diff
@@ -0,0 +1,74 @@
...
+ * for all times beyond the one covered by the TZif file
+ * @param ts ORC timestamp
+ *
+ * @return GMT offset
+ */
+inline __device__ duration_s get_gmt_offset(table_device_view tz_table, timestamp_s ts)
+{
+  if (tz_table.num_rows() == 0) { return duration_s{0}; }
+
+  cudf::device_span<timestamp_s const> ttimes(tz_table.column(0).head<timestamp_s>(),
```
bdice, line 43: Can we give this a longer name? Does `ttimes` mean `transition_times`? Let's try to increase clarity as much as possible in names.

cpp/include/cudf/detail/timezone.cuh
bdice: Is this a rule we can get (or reference) from some other source? If we must implement it, is this codified in a way that is specific to ORC conventions or general to all time zone libraries?

To the greatest extent possible, cuDF should avoid implementing or defining its own rules for time math. We should always use (or reference) `libcudacxx`, other timezone libraries, an RFC, or similar.

cpp/include/cudf/timezone.hpp
```diff
@@ -0,0 +1,49 @@
...
+// Two entries per year, over the length of the cycle
+static constexpr uint32_t cycle_entry_cnt = 2 * cycle_years;
+
+/**
+ * @brief Creates a transition table to convert ORC timestamps to UTC.
+ *
+ * Uses system's TZif files. Assumes little-endian platform when parsing these files.
+ *
+ * @param tzif_dir The directory where the TZif files are located
+ * @param timezone_name standard timezone name (for example, "US/Pacific")
```
bdice, line 40: Let's not use an obsolete time zone for our example. `US/Pacific` has been replaced by `America/Los_Angeles`.

http://www.timezoneconverter.com/cgi-bin/zoneinfo.tzc?s=default&tz=US/Pacific

bdice (changes requested):
Suggestions on time zone rules and naming. (Haven't dug much into the actual C++)

cpp/include/cudf/detail/timezone.cuh
vuule: This is the definition of the Gregorian calendar. It's unrelated to ORC or timezones.
Given how specific this is (only multiples of 400 years have a fixed number of days, otherwise you need to know the year range) I doubt any library can compute this without computing number of days for each year individually (1-400 in this case, I think).

I can add a link to Gregorian calendar and clarification comments.

bdice: I think I'm asking something different. I want a reference that states why we "search the 400-year cycle if outside of the file entries range" (in the comment below). We didn't come up with that, it's defined by some spec outside of libcudf.

vuule: Ah, that. I actually don't know where this comes from. I'll try to find it and incorporate in some way.

vuule:
Looks like I managed to straight-up murder this branch. Will open a new PR when I find the changes.

cpp/include/cudf/detail/timezone.cuh
vuule: So, I've been digging around the interwebs and could not find anything on the use of 400 year cycle to compute time zone offsets in the future. 
The only thing I found is the name for this cycle - [Solar cycle](https://en.wikipedia.org/wiki/Solar_cycle_(calendar)). And even that is a seemingly overloaded term.
I added this to the names for the relevant constants, will incorporate it here as well.

cpp/include/cudf/detail/timezone.cuh
```diff
@@ -0,0 +1,74 @@
...
+  auto const ts_ttime_it = [&]() {
+    auto last_less_equal = [](auto begin, auto end, auto value) {
+      auto const first_larger = thrust::upper_bound(thrust::seq, begin, end, value);
+      // Return start of the range if all elements are larger than the value
+      if (first_larger == begin) return begin;
+      // Element before the first larger element is the last one less or equal
+      return first_larger - 1;
+    };
+
+    auto const file_entry_end = ttimes.begin() + (ttimes.size() - cycle_entry_cnt);
```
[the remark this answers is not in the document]
vuule: Made the name more specific, but kept the `snake_case` style. I hope it's at least a bit easier to see that it's a constant, based on the name.

cpp/include/cudf/detail/timezone.cuh
```diff
@@ -0,0 +1,74 @@
...
+#include <cudf/types.hpp>
+#include <cudf/utilities/span.hpp>
+
+#include <thrust/binary_search.h>
+#include <thrust/execution_policy.h>
+
+namespace cudf::detail {
+
+/**
+ * @brief Returns the GMT offset for a given date and given timezone table.
```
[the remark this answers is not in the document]
vuule: My understanding is that UT, UTC and GMT are equivalent for `second` precision. I have changed the TZif related code to use UT, but since ORC references UTC, I kept it there as such.

cpp/include/cudf/detail/timezone.cuh
vuule: tried to make this clearer. Please let me know if it's still lacking

shwina (approved):
LGTM from Cython side!

cpp/include/cudf/timezone.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <optional>
+#include <string>
+
+namespace cudf {
+class table;
+
+// Cycle in which the time offsets repeat in Gregorian calendar
+static constexpr int32_t solar_cycle_years = 400;
+// Number of future entries in the timezone transition table:
+// Two entries per year, over the length of the Gregorian calendar's solar cycle
```
bdice, line 30: How do we know there are two entries per year? Is it guaranteed by the spec that timezones cannot have zero, one, three, or four entries?

bdice (commented):
Code looks good. One minor question about the far, far future.

cpp/include/cudf/timezone.hpp
vuule: There are two entries because of Daylight Saving Time. If the time zone does not have DST, all future entries will have the same time offset. We could optimize this case in the future.

bdice: It seems that we’re hardcoding an assumption that daylight savings time exists and results in exactly two changes per year. I don’t think either of those are safe assumptions.

vuule: It is a safe assumption, since that's how `make_timezone_transition_table` works (currently). It will always include two entires per year in cycle. The case when DST does not exist, it will create 2 X cycle_years of the same offset, which is still correct, just slower than necessary.
We could separate the output columns into tzif entires and future entries. With that, we can have any number of elements in the future column. I would like to keep optimizations like this to a future PR. Let me know what you think.

bdice: Document this ("it always includes two entries per year, assuming two transitions per year due to Daylight Savings Time") and any potential future work in a comment. That would be enough to resolve this and move forward.

vuule: Agree with the suggestion, this on not clear from the API.
Added more detailed description of the transition table.

cpp/include/cudf/timezone.hpp
```diff
@@ -0,0 +1,54 @@
...
+static constexpr uint32_t solar_cycle_entry_count = 2 * solar_cycle_years;
+
+/**
+ * @brief Creates a transition table to convert ORC timestamps to UTC.
+ *
+ * Uses system's TZif files. Assumes little-endian platform when parsing these files.
+ * The transition table starts with the entries from the TZif file. For timestamps after the file's
+ * last transition, the table includes entires that form a `solar_cycle_years`-year cycle (future
```
bdice, line 38:
```suggestion
 * last transition, the table includes entries that form a `solar_cycle_years`-year cycle (future
```

bdice (approved):
Typo fix. Otherwise LGTM!

PR #12808, opened 2023-02-21
Fixing parquet coalescing of reads

hyperbolic2346:
When digging through the new chunk writer for parquet pipelined reads, I noticed that the code changed and the read coalescing was no longer happening. Through some checking, I found that tests were issuing multiple smaller reads, specifically a read per rowgroup instead of a single read for multiple rowgroups. With this change, there is a single read per file.

I believe this is an unintentional change and a departure from the previous behavior.

## Checklist

### DIFF
--- a/cpp/src/io/parquet/reader_impl_preprocess.cu
+++ b/cpp/src/io/parquet/reader_impl_preprocess.cu
@@ -685,7 +685,6 @@ void reader::impl::load_and_decompress_data(std::vector<row_group_info> const& r
     auto const row_group_start  = rg.start_row;
     auto const row_group_source = rg.source_index;
     auto const row_group_rows   = std::min<int>(remaining_rows, row_group.num_rows);
-    auto const io_chunk_idx     = chunks.size();
 
     // generate ColumnChunkDesc objects for everything to be decoded (all input columns)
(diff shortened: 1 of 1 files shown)


cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -733,18 +732,19 @@ void reader::impl::load_and_decompress_data(std::vector<row_group_info> const& r
...
-                                                           _stream));
-
     remaining_rows -= row_group.num_rows;
   }
+
+  // Read compressed chunk data to device memory
+  read_rowgroup_tasks.push_back(read_column_chunks_async(_sources,
+                                                         raw_page_data,
+                                                         chunks,
+                                                         0,
```
ttnghia, line 742: So previously it read `io_chunk_idx` which is const set to `chunks.size()`. Now it is `0`?

hyperbolic2346: The `chunks.size()` was captured before adding to chunks for the row group, so the starting size of the chunks array. It would be 0, then however many chunks were there at the next iteration, etc.

The function parameters now say start at chunk 0 and read all the chunks.

ttnghia (approved):

PR #12809, opened 2023-02-21
Shuffling read into a sub function in parquet read

### LINKED ISSUE #12811: Refactor parquet read code to put chunk creation and reading into a function.
Parquet pipelining needs to be able to call read multiple times with smaller portions of the file. Refactoring is needed to better support this change.

hyperbolic2346:
## Description
This change is the first step toward the pipelined parquet reader and moves the chunk creation and file reads into another function. Right now, the operation is the same, but this change will allow for smaller groups to be read at a time for pipelining.

## Checklist

closes #12811


cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -130,6 +130,15 @@ class reader::impl {
                     bool uses_custom_row_bounds,
                     host_span<std::vector<size_type> const> row_group_indices);
 
+  /**
+   * @brief Create chunk information and start file reads
+   *
+   * @param row_groups_info vector of information about row groups to read
+   * @param num_rows  Maximum number of rows to read
+   * @return boolean indicating if compressed chunks were found
```
ttnghia, line 138: This is not correct. A pair is returned instead.

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -130,6 +130,15 @@ class reader::impl {
...
+   */
+  std::pair<bool, std::vector<std::future<void>>> create_and_read_column_chunks(
+    std::vector<row_group_info> const& row_groups_info, size_type num_rows);
```
ttnghia, line 141: I'm sure that somebody will request using `host_span` (not me) :smile:

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -651,16 +651,11 @@ void reader::impl::allocate_nesting_info()
   page_nesting_decode_info.host_to_device(_stream);
 }
 
-void reader::impl::load_and_decompress_data(std::vector<row_group_info> const& row_groups_info,
-                                            size_type num_rows)
+std::pair<bool, std::vector<std::future<void>>> reader::impl::create_and_read_column_chunks(
+  std::vector<row_group_info> const& row_groups_info, size_type num_rows)
 {
-  // This function should never be called if `num_rows == 0`.
-  CUDF_EXPECTS(num_rows > 0, "Number of reading rows must not be zero.");
```
ttnghia, line 658: So no more checking for `num_rows`? If `num_rows == 0` is acceptable now, should we short circuit this?

ttnghia:
Should this depends on https://github.com/rapidsai/cudf/pull/12808?

cpp/src/io/parquet/reader_impl_preprocess.cu
ttnghia: Nevermind, I just see this check below.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -733,39 +727,54 @@ void reader::impl::load_and_decompress_data(std::vector<row_group_info> const& r
...
+{
+  // This function should never be called if `num_rows == 0`.
+  CUDF_EXPECTS(num_rows > 0, "Number of reading rows must not be zero.");
+
+  auto& raw_page_data    = _file_itm_data.raw_page_data;
+  auto& decomp_page_data = _file_itm_data.decomp_page_data;
+  auto& chunks           = _file_itm_data.chunks;
+  auto& pages_info       = _file_itm_data.pages_info;
+
+  auto [has_compressed_data, read_rowgroup_tasks] =
```
ttnghia, line 759: `const`?

cpp/src/io/parquet/reader_impl.hpp
hyperbolic2346: Well, I agree so I went ahead and did the work for them.

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -130,10 +130,21 @@ class reader::impl {
                     bool uses_custom_row_bounds,
                     host_span<std::vector<size_type> const> row_group_indices);
 
+  /**
+   * @brief Create chunk information and start file reads
+   *
+   * @param row_groups_info vector of information about row groups to read
+   * @param num_rows  Maximum number of rows to read
+   * @return pair of boolean indicating if compressed chunks were found and a vector of futures for
+   * read completion
```
ttnghia, line 139:
```suggestion
   *         read completion
```

hyperbolic2346: I would like to push it out as well, but pre-commit did that format and changes it back to that.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -733,39 +727,54 @@ void reader::impl::load_and_decompress_data(std::vector<row_group_info> const& r
...
+  read_rowgroup_tasks.push_back(read_column_chunks_async(_sources,
+                                                         raw_page_data,
+                                                         chunks,
+                                                         0,
+                                                         chunks.size(),
+                                                         column_chunk_offsets,
+                                                         chunk_source_map,
+                                                         _stream));
+
+  CUDF_EXPECTS(remaining_rows <= 0, "All rows data must be read.");
```
ttnghia, line 743: Should this be this?
```suggestion
  CUDF_EXPECTS(remaining_rows == 0, "All rows data must be read.");
```

Because if `remaining_rows < 0` we probably "over-read" (i.e., wrongly read more rows than the file has)?

ttnghia (approved):

cpp/src/io/parquet/reader_impl_preprocess.cu
hyperbolic2346: Turns out this isn't going to work as-is. I agree that it should be a zero check here, but we're subtracting `row_group.num_rows` each iteration, but we may not have read that many rows.

```
auto const row_group_rows   = std::min<int>(remaining_rows, row_group.num_rows);
```

Still thinking about the right path here.

Issue #12806, opened 2023-02-18
[BUG] Series.str.character_ngrams(as_list=True) resets index when it shouldn't


daxiongshu:
**Describe the bug**
`Series.str.character_ngrams(as_list=True)` resets index when it shouldn't

**Steps/Code to reproduce bug**
Consider the following code:
```
import cudf
df = cudf.DataFrame(index=[3,4,5])
df['a'] = ['some text text','good good start','fast']
df['ngram'] = df['a'].str.character_ngrams(5,True).list.unique()
df['ngram']
```
output:
```
3    None
4    None
5    None
Name: ngram, dtype: list
```
This is because index was reset by `Series.str.character_ngrams(as_list=True) `
```
>>> df['a'].str.character_ngrams(5,True)
0    [some , ome t, me te, e tex,  text, text , ext...
1    [good , ood g, od go, d goo,  good, good , ood...
2                                                   []
Name: a, dtype: list
```

**Expected behavior**
do not reset index and output index should be the same as input `[3,4,5]` in this case
```
>>> df['a'].str.character_ngrams(5,True)
3    [some , ome t, me te, e tex,  text, text , ext...
4    [good , ood g, od go, d goo,  good, good , ood...
5                                                   []
Name: a, dtype: list
```

**Environment overview (please complete the following information)**
 - Environment location: Bare-metal
 - Method of cuDF install: conda

**Environment details**
```
>>> cudf.__version__
'23.02.00a+241.g66b846a01d'
```

shwina:
Thanks, @daxiongshu!

I also think it would make sense for the `as_list=False` case to return a result whose index values correspond to the string from which the ngram was computed. That is:

```
>>> df['a']
3     some text text
4    good good start
5               fast
Name: a, dtype: object

>>> df['a'].str.character_ngrams(3)
3    som
3    ome
3    me 
3    e t
3     te
3    tex
3    ext
3    xt 
3    t t
3     te
3    tex
3    ext
4    goo
4    ood
4    od 
4    d g
4     go
4    goo
4    ood
4    od 
4    d s
4     st
4    sta
4    tar
4    art
5    fas
5    ast
Name: a, dtype: object
```

Note that then, the result of `.character_ngrams(n, as_list=False)` is just the "exploded" version of `.character_ngrams(n, as_list=True)`, which seems like a good consistency.

PR #12814, opened 2023-02-21
Make string methods return a Series with a useful Index

shwina:
## Description
Closes https://github.com/rapidsai/cudf/issues/12806

Many string methods like `character_ngrams` currently return a `Series` with the default index (`RangeIndex`). This PR makes it so that the index of the result corresponds to the index of the input.

More specifically, this PR changes the index of the result of the following string methods:

- [x] `character_ngrams`
- [x] `tokenize`
- [x] `character_tokenize`

## Checklist


vyasr:
@shwina was this PR intentionally marked as draft?

shwina:
@vyasr - should be ready for review now

galipremsagar (approved):

python/cudf/cudf/core/column/string.py
```diff
@@ -4641,41 +4646,43 @@ def character_tokenize(self) -> SeriesOrIndex:
...
+        2    y
+        2    o
+        2    u
+        2    .
         dtype: object
         """
         result_col = libstrings.character_tokenize(self._column)
         if isinstance(self._parent, cudf.Series):
-            return cudf.Series(result_col, name=self._parent.name)
+            lengths = self.len().fillna(0)
```
vyasr, line 4683: Is the `fillna(0)` matching pandas behavior? It looks like the test is constructing the expected output manually to be the same, so it's not clear if it's going to match pandas or if the test is just constructing the same as what this outputs.

vyasr (approved):
One question about null behavior but I'm approving and will let you resolve accordingly.

PR #12817, opened 2023-02-21
[REVIEW] Deprecate `na_sentinel` in `factorize`

galipremsagar:
## Description

This PR:

- [x] Deprecates `na_sentinel` in `factorize`.
- [x] Introduces `use_na_sentinel` as an alternative.
- [x] Introduces `sort` in `factorize`.
- [x] Fixing up docs.

The above changes are required as part of enabling pandas 2.0 support.

## Checklist


python/cudf/cudf/core/algorithms.py
```diff
@@ -7,47 +7,111 @@
...
+        )
+    elif use_na_sentinel is None and na_sentinel is None:
+        use_na_sentinel = True
+        na_sentinel = -1
+
+    if use_na_sentinel is None:
+        use_na_sentinel = True
+    elif na_sentinel is None:
+        na_sentinel = -1
+    else:
```
shwina, line 111: The logic in this function is a bit difficult to follow (at least for me), with all the `if`s and `else`s.

If I'm reading correctly, the `else` is reached when:

```suggestion
    else:  # use_sentinel is either True or False, na_sentinel is None
```

Is that correct? if so, I think the comment helps clarify.

python/cudf/cudf/tests/test_series.py
```diff
@@ -486,12 +486,37 @@ def test_series_factorize(data, na_sentinel):
...
         expected_labels, expected_cats = psr.factorize(na_sentinel=na_sentinel)
-    actual_labels, actual_cats = gsr.factorize(na_sentinel=na_sentinel)
+    with pytest.warns(FutureWarning):
+        actual_labels, actual_cats = gsr.factorize(na_sentinel=na_sentinel)
 
     assert_eq(expected_labels, actual_labels.get())
     assert_eq(expected_cats.values, actual_cats.to_pandas().values)
 
 
+@pytest.mark.parametrize(
```
shwina, line 496: Should we also add a test for the new `sort=` argument?

python/cudf/cudf/core/algorithms.py
galipremsagar: Yea, it's a lot of code smell here. Let me give another try to simplify things here today.

python/cudf/cudf/core/column/column.py
```diff
@@ -1343,7 +1343,16 @@ def _return_sentinel_column():
...
+            dtype = min_scalar_type(
+                max(
+                    len(cats),
+                    -1
+                    if isinstance(na_sentinel, cudf.Scalar)
+                    and na_sentinel.value is cudf.NA
+                    else na_sentinel,
+                ),
+                8,
+            )
```
shwina, line 1355: I wonder if it's cleaner for internal methods like this to always accept a `cudf.Scalar` when a scalar is expected.

(in the same way that we moved towards internal methods always accepting a `ColumnBase`, rather than any column-like object).

@vyasr do you have any thoughts?

galipremsagar: I think that's a better idea too, switched to accept only `cudf.Scalar`'s here.

python/cudf/cudf/tests/test_series.py
galipremsagar: Done

python/cudf/cudf/core/algorithms.py
galipremsagar: Simplified the logic, let me know what you think about it now.

mroeschke: FWIW pandas had some mangled code when deprecating this in 1.5 https://github.com/pandas-dev/pandas/pull/47157/files#diff-c8f3ad29eaf121537b999e88e9117f3e3702d0b818a67516da25093fe2890ce8R794

python/cudf/cudf/core/algorithms.py
```diff
@@ -7,47 +7,109 @@
...
+            "Specifying the specific value to use for `na_sentinel` is "
+            "deprecated and will be removed in a future version of cudf. "
+            "Specify `use_na_sentinel=True` to use the sentinel value -1, "
+            "and `use_na_sentinel=False` to encode NA values.",
+            FutureWarning,
+        )
+
+    if use_na_sentinel is not None and na_sentinel is not None:
+        raise ValueError(
+            "Cannot specify both `na_sentinel` and `use_na_sentile`; "
```
mroeschke, line 99:
```suggestion
            "Cannot specify both `na_sentinel` and `use_na_sentinel`; "
```

python/cudf/cudf/core/column/column.py
vyasr: Yes, I would favor that approach.

python/cudf/cudf/core/column/column.py
```diff
@@ -1337,13 +1340,22 @@ def _label_encoding(
         """
         from cudf._lib.join import join as cpp_join
 
+        if na_sentinel is None:
```
shwina, line 1343: Is my understanding correct that if we replace this line with:

```suggestion
        if na_sentinel is None or na_sentinel.value is cudf.NA
```

shwina (approved):
Small change potentially requested, but approving otherwise.

python/cudf/cudf/core/column/column.py
galipremsagar: Yup, that's correct. Made the change 👍

PR #12818, opened 2023-02-21
[REVIEW] Deprecate `datetime_is_numeric` from `describe`

galipremsagar:
## Description

`datetime_is_numeric` is non-functional parameter in `cudf` and is removed in `pandas 2.0`, hence deprecating this param for removal.

## Checklist


python/cudf/cudf/core/series.py
```diff
@@ -3114,6 +3115,13 @@ def describe(
     ):
         """{docstring}"""
 
+        if datetime_is_numeric:
+            warnings.warn(
```
shwina, line 3119: I'm confused: if we are inside the `if`, does it not mean that `datetime_is_numeric` is already `True`?

galipremsagar: My bad, fixed it.

python/cudf/cudf/tests/test_series.py
```diff
@@ -462,7 +465,8 @@ def test_series_describe_other_types(ps):
     gs = cudf.from_pandas(ps)
 
     expected = ps.describe()
```
galipremsagar, line 467: There will be no warnings on pandas end for this because it was a directly removed parameter without any deprecations.

bdice (changes requested):
https://docs.rapids.ai/api/cudf/nightly/developer_guide/contributing_guide.html#deprecating-and-removing-code

Code looks good. Policy question that I want to block on: Do we require the `.. deprecated::` message? We haven't been consistent about that in the past, to my knowledge. We should declare in our developer guide whether deprecations only require a `FutureWarning` or should also be listed in the documentation.

bdice (approved):
Approving, policy question resolved.

wence-:
I think I am am +1 on deprecation warnings in docstrings (and hence the docs), since that way someone can determine if a function is deprecated before running their code (especially if they do so in a way that swallows output).

It's a little more burden when removing the deprecated code, but not too much.

PR #12825, opened 2023-02-22
[REVIEW] Deprecate `names` & `dtype` in `Index.copy`

galipremsagar:
## Description

This PR deprecates `dtype` in `Index & MultiIndex`.`copy`, and `names` in `Index.copy`.

## Checklist


python/cudf/cudf/core/multiindex.py
```diff
@@ -336,11 +336,29 @@ def copy(
...
+               The `levels` parameter is deprecated and will be removed in
+               a future version of cudf.
+
         codes : sequence of arrays, optional (default None)
             Integers for each level designating which label at each location.
             Original values used if None.
+
+            .. deprecated:: 23.02
+
+               The `levels` parameter is deprecated and will be removed in
```
galipremsagar, line 359:
```suggestion
               The `codes` parameter is deprecated and will be removed in
```

python/cudf/cudf/core/index.py
```diff
@@ -313,9 +313,20 @@ def copy(self, name=None, deep=False, dtype=None, names=None):
             Ignored for RangeIndex
         dtype : numpy dtype optional (default: None)
             Target dtype for underlying range data
+
+            .. deprecated:: 23.02
```
vyasr, line 317: We've never used this tag before, have we? OTOH, we do occasionally use the `Warns` section to show that we're going to raise a FutureWarning. I'm not opposed do doing this, jw if we should have a consistent approach (indicate warning, use the deprecated tag here, or both).

vyasr (commented):
Looks good, one question about how we want to broadcast these deprecations going forward.

python/cudf/cudf/core/index.py
galipremsagar: I think it's best to do both. I myself got to know about [admonitions](https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/admonitions.html) very recently. They get rendered nicely too. So felt it would be beneficial to have the deprecated admonition show up in docs too.

vyasr: Yep admonitions in Sphinx are nice. Happy to use both if you think both is best. Currently I don't see warnings documented, though; did you actually intend to include both, or only have the admonition?

galipremsagar: In this specific case `dtype` was already deprecated.

vyasr (approved):
LGTM

PR #12827, opened 2023-02-22
Make custom build backend work in legacy-editable mode

trxcllnt:
Not entirely sure how setuptools uses the custom build backend, but this PR makes editable installs work when doing:

```shell
SETUPTOOLS_ENABLE_FEATURES="legacy-editable" \
pip install --no-build-isolation --no-deps --editable .
```


vyasr:
@trxcllnt could you try again with the latest branch-23.04? https://github.com/rapidsai/cudf/pull/12741 should make this change unnecessary.

trxcllnt:
Yeah I'll give it a shot

PR #12595, opened 2023-01-23
[REVIEW] Add performance benchmarks to user facing docs

galipremsagar:
## Description

Resolves: #12295 

This PR introduces a notebook of benchmarks that users will be able to run if they download the notebook. The notebook also generates graphs which are going to show up in cudf python docs.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12849, opened 2023-02-24
Replace nbtest with nbval.

bdice:
## Description
This PR replaces the `nbtest` notebook testing shell scripts with a simpler package [nbval](https://github.com/computationalmodelling/nbval) that integrates with pytest to run notebooks. This allows us to get rid of a fair amount of code and also get more helpful CI outputs.

## Checklist


ci/test_notebooks.sh
```diff
@@ -29,33 +29,5 @@ rapids-mamba-retry install \
...
-    else
-        nvidia-smi
-        ${NBTEST} ${nbBasename}
-    fi
-done
-
-rapids-logger "Test script exiting with value: $EXITCODE"
-exit ${EXITCODE}
+# Run all notebooks with nbval
+python -m pytest -v --nbval --nbval-lax notebooks/
```
galipremsagar, line 33: Is it possible to exclude `notebooks` directory? Because the performance notebook being added in https://github.com/rapidsai/cudf/pull/12595 will require state-of-the-art hardware hence it is not fit to be run in CI.

bdice: I’ll look at how to do exclusions soon. We didn’t run dask notebooks with the old method, so I’ll investigate how to align with that.

For curiosity, what is the approximate runtime of the performance notebook @galipremsagar ?

galipremsagar: hours on H100, primarily because of CPU parts. GPU parts finish quick(Max 1-2 mins?).

vyasr:
@bdice do you still want to pursue this? The package [has been quiet for about a year now](https://github.com/computationalmodelling/nbval/commits/master/), but maybe it's widely used enough that we can still rely on the community keeping it afloat.

bdice:
Nope. I agree that nbval is not as active. I thought the pytest integration of nbval could be nice, but it's not worth changing. I'll close this.

PR #12850, opened 2023-02-24
Migrate as much as possible to pyproject.toml

vyasr:
## Description

I migrated as much build logic as possible to pyproject.toml for cudf, dask_cudf, and custreamz. A couple of notes:
- I didn't move cudf_kafka because it is currently still using a pure setuptools build with Cython files. We may as well do the pyproject.toml migration there after we have also done the scikit-build conversion since there's a lot of logic that can't be removed from setup.py until then. I'm also not entirely confident that everything will work as expected if we move data to pyproject.toml without switching to scikit-build based on issues like the ones mentioned in [this SO post about Cython and project data stored in pyproject.toml](https://stackoverflow.com/questions/73800736/pyproject-toml-and-cython-extension-module).
- I could get rid of setup.py entirely for custreamz at this point, but I expect that we'll run into a host of additional issues when we do eventually move to requiring `pip install X` instead of `python setup.py X` across RAPIDS and I'd rather deal with all of those at the same time with our main packages rather than using an esoteric package like custreamz as the test case.

## Checklist


raydouglass (approved):

python/cudf/pyproject.toml
```diff
@@ -16,6 +16,66 @@ requires = [
...
+    "cupy-cuda11x",
+]
+classifiers = [
+    "Intended Audience :: Developers",
+    "Topic :: Database",
+    "Topic :: Scientific/Engineering",
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.10",
```
galipremsagar, line 54: Should we have the python 3.9 classifier consistent across all `.toml`'s? Either we drop it everywhere or keep it incase python 3.9 packages are making a comeback.

python/dask_cudf/setup.py
```diff
@@ -2,44 +2,13 @@
...
-        "Programming Language :: Python",
-        "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: 3.9",
-        "Programming Language :: Python :: 3.10",
-    ],
+    include_package_data=True,
     packages=find_packages(exclude=["tests", "tests.*"]),
-    install_requires=install_requires,
-    extras_require=extras_require,
+    entry_points={
```
galipremsagar, line 8: This seems to be resulting in the following error message locally:
```
   ##########################################################################
    # configuration would be ignored/result in error due to `pyproject.toml` #
    ##########################################################################

    The following seems to be defined outside of `pyproject.toml`:

    `entry-points = {'dask.dataframe.backends': ['cudf = dask_cudf.backends:CudfBackendEntrypoint']}`

    According to the spec (see the link below), however, setuptools CANNOT
    consider this value unless 'entry-points' is listed as `dynamic`.
```

galipremsagar (approved):
LGTM, some non-blocking comments.

python/dask_cudf/setup.py
vyasr: This looks like a real error, thanks for the catch! I'll fix this and would appreciate if you could test it (or show me how, I'm not sure how to verify that dask entry points are properly configured by cudf).

python/cudf/pyproject.toml
vyasr: Good question. We removed it in rmm so I'm going to continue removing it everywhere else during this transition for now. We can readjust later if we want to.

python/dask_cudf/setup.py
vyasr: Pushed a fix.

python/dask_cudf/pyproject.toml
```diff
@@ -36,6 +36,7 @@ classifiers = [
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
 ]
+dynamic = ["entry_points"]
```
galipremsagar, line 39: Getting a new error locally:
```
writing list of installed files to 'record.txt'
/nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/setuptools/config/pyprojecttoml.py:108: _BetaConfiguration: Support for `[tool.setuptools]` in `pyproject.toml` is still *beta*.
  warnings.warn(msg, _BetaConfiguration)
configuration error: `project.dynamic[0]` must be one of ['version', 'description', 'readme', 'requires-python', 'license', 'authors', 'maintainers', 'keywords', 'classifiers', 'urls', 'scripts', 'gui-scripts', 'entry-points', 'dependencies', 'optional-dependencies']
GIVEN VALUE:
    "entry_points"

OFFENDING RULE: 'enum'

DEFINITION:
    {
        "enum": [
            "version",
            "description",
            "readme",
            "requires-python",
            "license",
            "authors",
            "maintainers",
            "keywords",
            "classifiers",
            "urls",
            "scripts",
            "gui-scripts",
            "entry-points",
            "dependencies",
            "optional-dependencies"
        ]
    }
  File "/nvme/0/pgali/cudf/python/dask_cudf/setup.py", line 5, in <module>
    setup(
  File "/nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/setuptools/__init__.py", line 108, in setup
    return distutils.core.setup(**attrs)
  File "/nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 159, in setup
    dist.parse_config_files()
  File "/nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/setuptools/dist.py", line 885, in parse_config_files
    pyprojecttoml.apply_configuration(self, filename, ignore_option_errors)
  File "/nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/setuptools/config/pyprojecttoml.py", line 62, in apply_configuration
    config = read_configuration(filepath, True, ignore_option_errors, dist)
  File "/nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/setuptools/config/pyprojecttoml.py", line 126, in read_configuration
    validate(subset, filepath)
  File "/nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/setuptools/config/pyprojecttoml.py", line 51, in validate
    raise ValueError(f"{error}\n{summary}") from None
ValueError: invalid pyproject.toml config: `project.dynamic[0]`.
configuration error: `project.dynamic[0]` must be one of ['version', 'description', 'readme', 'requires-python', 'license', 'authors', 'maintainers', 'keywords', 'classifiers', 'urls', 'scripts', 'gui-scripts', 'entry-points', 'dependencies', 'optional-dependencies']
(cudfdev) pgali@dt07:/nvme/0/pgali/cudf$ 
```

galipremsagar (changes requested):

python/dask_cudf/pyproject.toml
vyasr: Whoops that was a typo on my part. I had `entry_points` instead `entry-points`.

galipremsagar (approved):
Veirifed the backend dispatching works fine 👍

PR #12854, opened 2023-02-27
Remove tokenizers pre-install pinning.

vyasr:
## Description

We pinned tokenizers during early wheel development because at the time the latest version on PyPI was only a source distribution without wheels available for arm and we didn't want to have to compile the package. More recent versions appear to be providing aarch binaries. Conversely, the pinned version (0.10.2) predated the existence of Python 3.10 wheels, which is resulting in nightly CI failing because Python 3.10 runs no longer find a wheel and attempt to compile from source.

## Checklist


bdice (approved):
Thanks for this fix!

PR #12858, opened 2023-02-27
Improve performance for replace-multi for long strings

davidwendt:
## Description
Adds more efficient algorithm for multi-string version of `cudf::strings::replace` for longer strings (greater than 256 bytes on average in each row).

## Checklist


cpp/src/strings/replace/replace.cu
```diff
@@ -704,92 +704,6 @@ std::unique_ptr<column> replace_slice(strings_column_view const& strings,
                              cudf::detail::copy_bitmask(strings.parent(), stream, mr));
 }
 
-namespace {
```
davidwendt, line 707: This code moved to the new `multi.cu` source file.

davidwendt:
| Rows | Width | Baseline | Change | X |
|:----:|:-----:| --------:| ------:| ------:|
| 4096 | 32 | 0.134 | 0.114 | 1.18 |
| 4096 | 64 | 0.18 | 0.161 | 1.12 |
| 4096 | 128 | 0.291 | 0.269 | 1.08 |
| 4096 | 256 | 0.534 | 0.51 | 1.05 |
| 4096 | 512 | 1.01 | 0.396 | 2.55 |
| 4096 | 1024 | 2.03 | 0.496 | 4.09 |
| 4096 | 2048 | 4.05 | 0.717 | 5.65 |
| 4096 | 4096 | 8.4 | 1.13 | 7.43 |
| 4096 | 8192 | 16.7 | 2.49 | 6.71 |
| 32768 | 32 | 0.135 | 0.116 | 1.16 |
| 32768 | 64 | 0.189 | 0.172 | 1.10 |
| 32768 | 128 | 0.308 | 0.293 | 1.05 |
| 32768 | 256 | 0.565 | 0.549 | 1.03 |
| 32768 | 512 | 1.12 | 1.09 | 1.03 |
| 32768 | 1024 | 2.21 | 2.08 | 1.06 |
| 32768 | 2048 | 4.4 | 3.45 | 1.28 |
| 32768 | 4096 | 8.79 | 6.03 | 1.46 |
| 32768 | 8192 | 17.4 | 11.1 | 1.57 |
| 262144 | 32 | 0.219 | 0.204 | 1.07 |
| 262144 | 64 | 0.409 | 0.391 | 1.05 |
| 262144 | 128 | 0.855 | 0.834 | 1.03 |
| 262144 | 256 | 3.67 | 3.7 | 0.99 |
| 262144 | 512 | 11.9 | 6.18 | 1.93 |
| 262144 | 1024 | 27.8 | 11.8 | 2.36 |
| 262144 | 2048 | 68 | 23.4 | 2.91 |
| 262144 | 4096 | 136 | 49.1 | 2.77 |
| 2097152 | 32 | 0.94 | 0.91 | 1.03 |
| 2097152 | 64 | 2.23 | 2.2 | 1.01 |
| 2097152 | 128 | 5.21 | 5.12 | 1.02 |
| 2097152 | 256 | 15.9 | 15.7 | 1.01 |
| 2097152 | 512 | 52.6 | 46.8 | 1.12 |
| 16777216 | 32 | 6.88 | 6.81 | 1.01 |
| 16777216 | 64 | 17.1 | 16.9 | 1.01 |

Values are in `ms` and `X` column is improvement factor. The row-widths 32-256 use the existing code path (no change in performance). For row-widths `>256` the new optimized code path is used.

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,488 @@
...
+                                   size_type const* d_targets_offsets,
+                                   string_index_pair* d_all_strings) const
+  {
+    if (!is_valid(idx)) { return 0; }
+
+    auto const d_output       = d_all_strings + d_offsets[idx];
+    auto const d_output_count = d_offsets[idx + 1] - d_offsets[idx];
+
+    auto const d_str = get_string(idx);
+    if (d_output_count == 1) {
```
hyperbolic2346, line 197: This is pretty magical. Did I miss the explanation for it? In my past life I would suggest a `#define` for this magic number to give it a good name.

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,488 @@
...
+      d_output[0] = string_index_pair{d_str.data(), d_str.size_bytes()};
+      return d_str.size_bytes();
+    }
+
+    auto const d_str_end         = d_str.data() + d_str.size_bytes();
+    auto const base_ptr          = get_base_ptr();  //+ delim_size - 1;
```
hyperbolic2346, line 203:
```suggestion
    auto const base_ptr          = get_base_ptr();
```

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,488 @@
...
+      if (str_ptr <= tgt_ptr && tgt_ptr < d_str_end) {
+        auto const keep_size = static_cast<size_type>(thrust::distance(str_ptr, tgt_ptr));
+        if (keep_size > 0) { d_output[output_idx++] = string_index_pair{str_ptr, keep_size}; }
+        output_size += keep_size;
+
+        auto const d_repl = get_replacement_string(d_pair.second);
+        if (!d_repl.empty()) {
+          d_output[output_idx++] = string_index_pair{d_repl.data(), d_repl.size_bytes()};
+        }
+        output_size += d_repl.size_bytes();
```
hyperbolic2346, line 223: Is this still true even if `d_repl.empty()==true`? Is `d_repl.size_bytes() == 0` if it is empty? I guess my question really is if this should move inside the if block.

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,488 @@
...
+  column_device_view const d_strings;
+  column_device_view const d_targets;
+  column_device_view const d_repls;
+  int32_t* d_offsets{};
+  char* d_chars{};
+
+  __device__ void operator()(size_type idx)
+  {
+    if (d_strings.is_null(idx)) {
+      if (!d_chars) d_offsets[idx] = 0;
```
hyperbolic2346, line 286:
```suggestion
      if (!d_chars) { d_offsets[idx] = 0; }
```

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,488 @@
...
+
+}  // namespace
+
+std::unique_ptr<column> replace(strings_column_view const& input,
+                                strings_column_view const& targets,
+                                strings_column_view const& repls,
+                                rmm::cuda_stream_view stream,
+                                rmm::mr::device_memory_resource* mr)
+{
+  if (input.is_empty()) return make_empty_column(type_id::STRING);
```
hyperbolic2346, line 333:
```suggestion
  if (input.is_empty()) { return make_empty_column(type_id::STRING); }
```

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,488 @@
...
+  CUDF_EXPECTS(((targets.size() > 0) && (targets.null_count() == 0)),
+               "Parameters targets must not be empty and must not have nulls");
+  CUDF_EXPECTS(((repls.size() > 0) && (repls.null_count() == 0)),
+               "Parameters repls must not be empty and must not have nulls");
+  if (repls.size() > 1)
+    CUDF_EXPECTS(repls.size() == targets.size(), "Sizes for targets and repls must match");
```
hyperbolic2346, line 339:
```suggestion
  CUDF_EXPECTS(repls.size() == 1 || repls.size() == targets.size(), "Must provide a single replacement string or sizes for targets and replacements must match");
```

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,488 @@
...
+
+  auto d_strings = column_device_view::create(input.parent(), stream);
+
+  if (input.size() == input.null_count() ||
+      ((input.chars_size() / (input.size() - input.null_count())) < AVG_CHAR_BYTES_THRESHOLD)) {
+    auto d_targets      = column_device_view::create(targets.parent(), stream);
+    auto d_replacements = column_device_view::create(repls.parent(), stream);
+
+    auto children = cudf::strings::detail::make_strings_children(
+      replace_multi_fn{*d_strings, *d_targets, *d_replacements}, input.size(), stream, mr);
```
hyperbolic2346, line 349: I was really expecting to find `avg_bytes > AVG_CHAR_BYTES_THRESHOLD ? replace_multi_fn(...) : replace_multi_parallel_fn(...)`. Would it make sense to push some of this function into a subfunction for the parallel path? Not sure if my expectation indicates extra cognitive load here or not.

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,488 @@
...
+                        pos_itr,
+                        pos_itr + pos_count,
+                        string_indices.begin());
+
+    // compute offsets per string
+    auto targets_offsets   = rmm::device_uvector<size_type>(strings_count + 1, stream);
+    auto d_targets_offsets = targets_offsets.data();
+
+    // memset to zero-out the target counts for any null-entries or strings with no targets
+    CUDF_CUDA_TRY(cudaMemsetAsync(
```
hyperbolic2346, line 407: Developer guide shows use of `thrust::uninitialized_fill` to initialize a `device_uvector`. Should that be used here?

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,488 @@
...
+                    [fn, d_positions, d_targets_offsets] __device__(size_type idx) -> size_type {
+                      return fn.count_strings(idx, d_positions, d_targets_offsets);
+                    });
+
+  // create offsets from the counts
+  auto offsets =
+    std::get<0>(cudf::detail::make_offsets_child_column(counts.begin(), counts.end(), stream, mr));
+  auto const total_strings =
+    cudf::detail::get_value<size_type>(offsets->view(), strings_count, stream);
+  auto const d_strings_offsets = offsets->view().data<size_type>();
```
hyperbolic2346, line 443:
```suggestion
  auto const d_strings_offsets = offsets->view().data<offset_type>();
```

cpp/src/strings/replace/multi.cu
hyperbolic2346: I see this was copied code now. Unsure if you want to take this or not, but I'm leaving it so you can decide.

cpp/tests/strings/replace_tests.cpp
```diff
@@ -352,6 +345,56 @@ TEST_F(StringsReplaceTest, ReplaceMulti)
...
+TEST_F(StringsReplaceTest, ReplaceMultiLong)
+{
+  auto input = cudf::test::strings_column_wrapper(
+    {"This string needs to be very long to trigger the long-replace internal functions.",
+     "01234567890123456789012345678901234567890123456789012345678901234567890123456789",
+     "01234567890123456789012345678901234567890123456789012345678901234567890123456789",
+     "Test string for overlap check: bananaápple bananá ápplebananá banápple ápple bananá",
+     "",
+     ""},
+    {1, 1, 1, 1, 0, 1});
```
hyperbolic2346, line 357: should this calculate the size and assert it is > `AVG_CHAR_BYTES_THRESHOLD` to prevent losing coverage if `AVG_CHAR_BYTES_THRESHOLD` is ever changed?

hyperbolic2346 (changes requested):
Mostly nits here. I like this change and think it will be a great addition!

cpp/benchmarks/string/replace.cpp
```diff
@@ -69,7 +69,7 @@ static void generate_bench_args(benchmark::internal::Benchmark* b)
   int const row_mult   = 8;
   int const min_rowlen = 1 << 5;
   int const max_rowlen = 1 << 13;
-  int const len_mult   = 4;
+  int const len_mult   = 2;
```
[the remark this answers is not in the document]
davidwendt: I wanted to change the benchmark to have a wider range of row widths to help with determining the `AVG_CHAR_BYTES_THRESHOLD` value.

cpp/src/strings/replace/multi.cu
davidwendt: There is nothing magical. Unless the input string is null, there will always be at least one output string for any replace operation.

cpp/src/strings/replace/multi.cu
davidwendt: The `d_repl.size_bytes() == 0` will be true if `d_repl.empty()==true`
Just trying to minimize divergent logic. Seemed harmless to add 0 in lock step with the other threads in the warp.

cpp/src/strings/replace/multi.cu
davidwendt: Trying to move away from `offset_type` in favor of using `size_type` everywhere.

cpp/tests/strings/replace_tests.cpp
davidwendt: I think that is a good idea.

hyperbolic2346: Not that important. At most I would comment near the value that this test attempts to build a string large enough to exercise it. I wouldn't hold up this PR over it.

cpp/src/strings/replace/multi.cu
hyperbolic2346: I hadn't heard that yet. I guess that means we aren't considering 64-bit offsets any time soon.

bdice: @hyperbolic2346 It's actually a slightly different concern. Currently `offset_type` is defined to be the same as `size_type`. If the two were defined differently, however, it would cause a lot of issues. The library design and API return types pretty much require that `offset_type == size_type` regardless of whether it's a 32 or 64 bit value. We should be moving away from `offset_type` in favor of `size_type` everywhere. (Perhaps we should do a mass find-replace?)

hyperbolic2346: I know that there would be large issues if they two were ever not equal, but I thought the design intent was that they *could* be different sizes, and if we decided to do that we would require a good deal of cleanup to remove the cross-contamination.

I was under the, seemingly false, impression that the goal was to differentiate the two so we could switch the types.

cpp/src/strings/replace/multi.cu
hyperbolic2346: Why do we have this if though if the output size is 1? I don't understand why 1 output size of 1 means to grab the entire string.

hyperbolic2346 (approved):

cpp/src/strings/replace/multi.cu
```diff
@@ -0,0 +1,500 @@
...
+      auto const left_size = static_cast<size_type>(thrust::distance(str_ptr, d_str_end));
+      d_output[output_idx] = string_index_pair{str_ptr, left_size};
+      output_size += left_size;
+    }
+    return output_size;
+  }
+
+  replace_multi_parallel_fn(column_device_view const& d_strings,
+                            device_span<string_view const> const d_targets,
+                            device_span<string_view const> const d_replacements)
```
PointKernel, line 234:
```suggestion
                            device_span<string_view const> d_targets,
                            device_span<string_view const> d_replacements)
```
A const span doesn't make much sense since it has no ownership.

PointKernel (approved):
Looks great with one minor non-blocking nit.

Thanks!

cpp/src/strings/replace/multi.cu
davidwendt: Ah yes, good point.

bdice (approved):
Approving CMake changes.

PR #12864, opened 2023-02-28
Update to protobuf>=4.21.6,<4.22.

bdice:
## Description
This updates cudf's protobuf version to resolve #12830 and a downstream conflict with `ortools` (a dependency of cuopt), which requires `protobuf>=4.21.5`.

## Checklist


bdice:
See also: https://protobuf.dev/support/cross-version-runtime-guarantee/

cc: @rgsl888prabhu

bdice:
@galipremsagar and I discussed offline and decided to pin to `>=4.22,<4.23` but there is no conda package for 4.22 yet (released 2 weeks ago). I will revert 563909c. I filed this PR to get version 4.22.0 on conda-forge: https://github.com/conda-forge/protobuf-feedstock/pull/180

galipremsagar (approved):

beckernick:
Updating our protobuf will likely allow more easily mixing cuDF and Tensorflow pip packages. Tensorflow currently pins to 3.20.x, but the tensorflow 2.12 release candidate [relaxes the constraint and appears to be consistent with the security alert](https://github.com/tensorflow/tensorflow/blob/95caef72eb6f8558a7ec184e917194f1429bdc40/tensorflow/tools/pip_package/setup.py#L100). This could be a great UX improvement for our pip packages.

bdice:
@beckernick Conda packages aren't available yet for 4.22. To clarify, you're saying we should permit a more extensive range to ensure compatibility with TensorFlow? Should we pin to something like `>=4.21.6,<5`?

beckernick:
Today, installing cuDF and the Tensorflow release candidate pip packages is challenging. Tensorflow naturally wants to bring in 4.22 (though could bring in 4.21.6 if needed), as I believe we must bring in 4.21.0 because of the version matching clause of `protobuf==4.21`. A similar problem exists for the current Tensorflow release with earlier versions of protobuf.

If our pip packages (today) were using `protobuf>=4.21.6,<4.22`, we should be compatible with the Tensorflow release candidate, which is great. When the conda packges are ready, if we then update again to pin to `>=4.22,<4.23` it should be similarly compatible.

bdice:
@beckernick Fantastic explanation, thank you for the detailed response. I agree! For now we'll pin to a version that we can get with both pip and conda: `>=4.21.6,<4.22` and we'll update again at a later time.

jjacobelli (approved):

PR #12871, opened 2023-03-01
Use python -m pytest for nightly wheel tests

bdice:
## Description
Nightly wheel tests are failing with this error:
```
+ sh -c 'pytest -v -n 8 ./python/cudf/cudf/tests'
sh: 1: pytest: not found
```

However, pytest is being installed.

The wheel itself is being installed with `python -m pip`, like:
```
python -m pip install --verbose './dist/cudf_cu11-23.4.0.1677647948-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl[test]'
```

Calling `python -m pytest` may resolve this error. For consistency, we should use `python -m` for all executable modules (pip and pytest).

## Checklist

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -113,7 +113,7 @@ jobs:
       build_type: pull-request
       package-name: cudf
       # Install cupy-cuda11x for arm from a special index url
-      test-before-arm64: "pip install cupy-cuda11x -f https://pip.cupy.dev/aarch64"
+      test-before-arm64: "python -m pip install cupy-cuda11x -f https://pip.cupy.dev/aarch64"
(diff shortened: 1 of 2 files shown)


vyasr:
I confirmed that this change will fix the nightly test CI by making a modified branch of the shared workflows to run the entire matrix on a PR and then making the corresponding changes in https://github.com/rapidsai/cudf/pull/12872. The commit without `python -m` fails, whereas the commit with it succeeds. We can merge this one since it's already in a mostly working state (although we will need to resolve the docs-build issue, CC @ajschmidt8 who I think dealt with this recently).

jjacobelli (approved):

Issue #12786, opened 2023-02-16
[FEA] Verify that no nonempty nulls exist in testing macros


vyasr:
**Is your feature request related to a problem? Please describe.**
libcudf functions expect nulls to be sanitized (see "libcudf expects nested types to have sanitized null masks" at https://docs.rapids.ai/api/libcudf/stable/developer_guide). Providing unsanitized inputs to libcudf functions may result in incorrect results. We are willing to accept putting the onus on the user to ensure that any manually constructed inputs are properly sanitized in this way. However, currently this approach can lead to hidden bugs if a libcudf developer implements an API that _produces_ unsanitized columns, something that we do not test for. Adding tests for a particular API to ensure that nulls are sanitized seems helpful on its face, but in practice it requires identifying APIs that need this check _a priori_, severely reducing the effectiveness of this approach. Therefore, an alternative approach is needed that automatically verifies these invariants for all tests without any additional work on the part of developers.

**Describe the solution you'd like**
We should modify our `CUDF_TESTS_EXPECTS_*` assertion macros to verify that the input columns are properly sanitized prior to performing any comparison. If the inputs columns are unsanitized, a suitable error should be raised even if the intended assertion would otherwise have passed. This approach automatically injects sanitization validation into all tests and ensures that no libcudf APIs produce unsanitized outputs.

SurajAralihalli:
I can look into this, thanks!

vyasr:
@SurajAralihalli thanks! Let us know if you need any help with this.

GregoryKimball:
Amazing work @SurajAralihalli !!

PR #12873, opened 2023-03-01
Remove calling to `purge_nonempty_nulls` in `make_lists_column`

ttnghia:
This is a reversal of https://github.com/rapidsai/cudf/pull/12370, where added `purge_nonempty_nulls` inside `make_lists_column`. This is due to internal discussions that we should not rely on factory functions like this to enforce data correctness. Instead, the caller should be responsible to sanitize the input.


ttnghia:
Yes opening an issue to track this is good. In summary, to merge this, we need to make sure that all the APIs that output lists will have their result sanitized.

vyasr:
@ttnghia I updated this PR to 25.06. I think that after https://github.com/rapidsai/cudf/pull/14559 merged and closed https://github.com/rapidsai/cudf/issues/12786, our test suite is validating all our APIs for us so we should be safe. WDYT?

vyasr:
Yup, with running the tests we now get nice errors. It looks like there are basically two errors, one in list column wrappers and one in interop. The latter issue is tricky since it is around the 0-copy views of device data, so we will have to decide what we do there. Normally we don't want to introspect and check data, but in this case we could wind up with users generating data via some other library that doesn't follow the same rules as us, and in that case the user may also not be able to easily guarantee that the input data was not clean. We may need to introduce some extra options into conversion for specifically that case.

ttnghia:
Fixing the failed tests:
 - [X] INTEROP_TEST
 - [x] JSON_TEST
 - [x] COPYING_TEST
 - [x] UTILITIES_TEST
 - [x] FACTORIES_TEST
 - [x] STRUCTS_TEST
 - [x] LISTS_TEST

cpp/src/io/json/json_column.cu
```diff
@@ -504,7 +505,10 @@ std::pair<std::unique_ptr<column>, std::vector<column_name_info>> device_json_co
         stream,
         mr);
       // Since some rows in child column may need to be nullified due to mixed types, we can not
-      // skip the purge_nonempty_nulls call in make_lists_column factory
+      // skip the purge_nonempty_nulls call.
+      if (auto const output_cv = ret_col->view(); detail::has_nonempty_nulls(output_cv, stream)) {
```
vuule: @shrshi is there a better/faster way to do this in the reader? Not intending to block this PR, but maybe leave a TODO/issue.

shrshi: I think we can skip adding non-empty nulls while creating `device_json_column` by modifying the tree algorithm. Specifically, in `host_tree_algorithms.cu`,
https://github.com/rapidsai/cudf/blob/cd173ff120789c2206a0409ba5375533540a9b63/cpp/src/io/json/host_tree_algorithms.cu#L948

I think we would need to propagate the null mask to the list type children and then create the columns.

cpp/include/cudf_test/nanoarrow_utils.hpp
```diff
@@ -64,6 +53,18 @@ struct generated_test_data {
...
                    std::back_inserter(bool_data_validity),
                    [](auto val) { return static_cast<uint8_t>(val); });
+    list_validity = bool_data_validity;
+
+    constexpr cudf::size_type length_of_individual_list = 3;
+    list_offsets[0]                                     = 0;
+    for (cudf::size_type i = 0; i < length; ++i) {
+      list_offsets[i + 1] = list_offsets[i] + (list_validity[i] ? length_of_individual_list : 0);
+    }
+    list_int64_data.resize(list_offsets.back());
```
vyasr, line 63: Ah so this was where we had list data we didn't want.

cpp/src/io/json/json_column.cu
```diff
@@ -504,7 +505,11 @@ std::pair<std::unique_ptr<column>, std::vector<column_name_info>> device_json_co
         stream,
         mr);
       // Since some rows in child column may need to be nullified due to mixed types, we can not
```
vyasr: Maybe not worth another CI run, but 🤷 
```suggestion
      // Since some rows in child column may need to be nullified due to mixed types, we cannot
```

cpp/src/io/json/json_column.cu
```diff
@@ -504,7 +505,11 @@ std::pair<std::unique_ptr<column>, std::vector<column_name_info>> device_json_co
...
-      // skip the purge_nonempty_nulls call in make_lists_column factory
+      // skip the purge_nonempty_nulls call.
+      if (auto const output_cv = ret_col->view();
```
vyasr, line 509: I guess you declared this variable here because you're trying to minimize the scope of this variable?

vyasr (approved):
Thanks for this Nghia! Good to see that almost all the changes needed are just to tests where we're reusing inputs across tests that require purging in lists cases. That means our actual APIs are clean.

cpp/src/io/json/json_column.cu
vuule: Opened issue https://github.com/rapidsai/cudf/issues/19899 with Copilot's help.
@shrshi feel free to correct if the description is wrong.

cpp/src/io/json/json_column.cu
ttnghia: Yes, this is C++17 feature and we can do declaration just to use inside the `if` scope.

cpp/src/io/json/json_column.cu
vyasr: That would definitely save some cycles if we can change it.

Issue #12792, opened 2023-02-16
[FEA] Retry support for chunked orc writer


abellina:
The spark-rapids team has been working on a framework to retry GPU operations on failure, like Out Of Memory Error (OOM) https://github.com/NVIDIA/spark-rapids/issues/7252. We have seen OOM issues while writing ORC to begin with, but we also want to handle Parquet writes as well in a retriable fashion.

At a very high level, we are planning to handle failures like OOM exception from RMM, but we are planning to possibly use this generic framework to handle other types of failures in the future, and attempt to handle it in different ways. 

- Some retry attempts will be to throttle our tasks down. We would attempt the cuDF call with the same table we had previously, but we are reducing the number of threads that are allowed to call into cuDF concurrently. We believe this throttling work will help us serialize the memory usage, so we let one of the Spark tasks take as much memory as needed.

- In other cases we also have a "split and retry" approach, in the case where we think a smaller chunk may be beneficial (if we are handling an OOM, or we have hit a cuDF column limit for instance). In this case, we are going to split the data smartly and re-attempt the cuDF call with smaller chunks.

When calling into the cuDF writers, we need to make sure that they are idempotent. What we are looking from the cuDF team is to guarantee that this is possible today (and therefore the API needs to be documented), or discuss what may need to be changed in order to support this retry mechanism.
```[tasklist]
```

mythrocks:
Watching this also.

> When calling into the cuDF writers, we need to make sure that they are idempotent. 

I suspect we might need more detail. Specifically at what level the retries are expected.

The chunked writers are stateful, so I suspect there are currently no guarantees that the output buffers aren't modified at all when the nth chunk fails its write.

We might well be ok if the retries are not per-chunk, but per output-buffer. (i.e. The task retries all chunks/column-vectors for a given output-buffer.)

That said, I'm likely wildly off base. I'd be keen to see the IO experts' analysis.

abellina:
The approach we've been discussing is at the chunk level, the assumption being that the caller has passed say N-1 chunks already, but the Nth one is having trouble. The plan would be to first retry this chunk (since perhaps it was a temporary issue, like memory fragmentation, that caused the failure), and if the retries are unsuccessful, the chunk would be broken into smaller chunks, which are each in turn attempted and retried.

[...]

Issue #7015, opened 2020-12-15
`CompactProtocolWriter` and `ProtobufWriter` API provide no encapsulation of the output buffer


vuule:
Problem:
The protocol writer classes take a pointer to a vector and use it as the output buffer. Writes change the size of this vector. This vector is also modified (including size changes) outside of the writers. The ORC/Parquet writers have a `std::vector` data member that is reused for protocol writes and manually reset between uses. ORC writer also reuses the `ProtobufWriter` object. In addition, Parquet writer reuses the output buffer to output data unrelated to `CompactProtocolWriter`. All this makes the use error-prone.

Solution proposal:
Modify the protocol writer API to use an internal output buffer and only provide getters for it. Also, protocol writer objects should not be reused (and cannot, with the proposed API). There shouldn't be a buffer data member in `xyz::writer::impl`.
These changes would limit the scope of the state to functions instead of the lifetime of `xyz::writer::impl` objects.

vuule:
Issue still relevant.

PR #12877, opened 2023-03-02
Refactor `io::orc::ProtobufWriter`

ttnghia:
This refactors the class `cudf::io::orc::ProtobufWriter`, making it independent from the ORC writer buffer. From now, a new instance of `ProtobufWriter` will work on its own buffer. That avoids touching the ORC writer's internal states.

The PR is part of solution for https://github.com/rapidsai/cudf/issues/12792.


cpp/src/io/orc/orc.hpp
```diff
@@ -528,20 +532,24 @@ class ProtobufWriter {
...
+  size_t write(UserMetadataItem const&);
+  size_t write(StripeFooter const&);
+  size_t write(Stream const&);
+  size_t write(ColumnEncoding const&);
+  size_t write(StripeStatistics const&);
+  size_t write(Metadata const&);
 
  protected:
-  std::vector<uint8_t>* m_buf;
+  std::unique_ptr<std::vector<uint8_t>> m_buf;
```
vuule, line 552: why not just std::vector?

cpp/src/io/orc/orc.hpp
```diff
@@ -528,20 +532,24 @@ class ProtobufWriter {
                            TypeKind kind,
                            ColStatsBlob const* stats);
 
+  void resize(std::size_t bytes) { m_buf->resize(bytes); }
+
+  std::unique_ptr<std::vector<uint8_t>> release() { return std::move(m_buf); }
```
vuule, line 537: looks like the `ProtobufWriter` object is not safe to use after `release()`. If possible, we should limit the `release()` use to rvalues.

vuule: ideally, release should be called when `ProtobufWriter` is going out of scope. Does the current use allow this?

vuule (changes requested):
Is this based on my issue https://github.com/rapidsai/cudf/issues/7015 or a convergent idea? :)

Some comments about implementation details.

cpp/src/io/orc/orc.hpp
ttnghia: Because with `unique_ptr`, we can make sure any `put_` call after `release()` will throw an exception. Otherwise, we may accidentally put new data into the buffer.

cpp/src/io/orc/orc.hpp
ttnghia: Yes, ideally `release()` should be called when we no longer need the `ProtobufWriter` object. That's why I changed `m_buf` into a smart pointer: after `release()`, `m_buf` will be `nullptr` and will throw exception if you call `put_`.

randerzander:
benchmark bot, please test this PR

cpp/src/io/orc/orc.hpp
vuule: Exception is still a run time failure, I'd like it to fail earlier :)
ref-qualifying the member function would make the incorrect use difficult:
```suggestion
  std::vector<uint8_t> release() && { return std::move(m_buf); }
```
With this, the use should looks something like
```
auto const buff = [&](){
  ProtobufWriter pbw;

  // use protobuf writer
  ...

  return std::move(pbw).release();
}();
And with this I don't think there's need for the `unique_ptr` wrapper.

Comment (ttnghia) [reply cpp/src/io/orc/orc.hpp:537]: What if somebody does this?
```
ProtobufWriter pbw;
...
// use pbw
...
auto buff = std::move(pbw).release();
// use buff
...
pbw.put_bytes(....); // <=== use pbw again?

```

Comment (vuule) [reply cpp/src/io/orc/orc.hpp:537]: IMO it's an easy mistake to find, because an object is used after std::move. I don't think we can prevent such misuse in C++.

Comment (ttnghia) [reply cpp/src/io/orc/orc.hpp:537]: I created a test with that:
```
#include <iostream>
#include <vector>

struct Buff {
    std::vector<int> buff;

    void put(int x) { buff.push_back(x); }

    auto release() && { return std::move(buff); }
};

int main() {
    Buff b;
    b.put(10);

    auto buff = std::move(b).release();
    std::cout << "buff: " << buff[0] << std::endl;

    b.put(1);
    std::cout << "PUT!!!" << std::endl;

    return 0;
}
```
The output:
```
Program returned: 0
buff: 10
PUT!!!
```

So we can use `pbw` after `std::move(pbw).release()` without any problem.

Comment (vuule) [reply cpp/src/io/orc/orc.hpp:537]: This code invokes UB because of use after move. The move is explicit so IMO this is as much as we should do to prevent UB. 
I don't think we should design classes so that incorrect use causes exceptions instead of UB. Since we disagree, this is probably a good question for the rapids cpp channel.

Comment (ttnghia) [reply cpp/src/io/orc/orc.hpp:537]: Using the version of `std::unique_ptr<std::vector>>`, the program crashes as I expected:
```
Program returned: 139
buff: 10
```

Note that there is no exception being thrown. If we want to throw, we need to check for `nullptr` in all the `put_` functions.

Comment (ttnghia) [reply cpp/src/io/orc/orc.hpp:537]: > `std::move` is nothing more than a cast to an rvalue reference. In itself it doesn't actually move anything.

From my understanding, if you don't assign `std::move(...)` to anything then it will just be there.

File: cpp/src/io/orc/orc.hpp [hunk 3]
```diff
@@ -528,20 +531,28 @@ class ProtobufWriter {
                            TypeKind kind,
                            ColStatsBlob const* stats);
 
+  void resize(std::size_t bytes) { m_buf->resize(bytes); }
+
+  std::size_t size() const { return m_buf->size(); }
+  uint8_t const* data() { return m_buf->data(); }
```
vuule, line 537: could these be combined into a span?

cpp/src/io/orc/orc.hpp
vuule: now that release has been removed, can we remove unique_ptr?

cpp/src/io/orc/orc.hpp
ttnghia: Unfortunately no.
 * We need them to pass into `host_write`.
 * In addition, we also need to access `size()` separately for data processing.

cpp/src/io/orc/orc.hpp
ttnghia: Yes, the reason to have `unique_ptr` is not as compelling as before. I'll remove it.

vuule (approved):
:100:

DaceT:
benchmark bot, please test this PR

DaceT:
benchmark bot, please test this PR

DaceT:
benchmark bot, please test this PR

nvdbaranec (commented):
Looks ok.  One comment on a possible clarifying constructor.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -1443,21 +1442,22 @@ void writer::impl::write_index_stream(int32_t stripe_id,
     }
   }
 
-  buffer_.resize((compression_kind_ != NONE) ? 3 : 0);
+  ProtobufWriter pbw;
+  pbw.resize((compression_kind_ != NONE) ? 3 : 0);
```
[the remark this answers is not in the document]
ttnghia: Good point. I'll add that constructor.

nvdbaranec (approved):

PR #12878, opened 2023-03-03
Change return type of `cudf::structs::detail::flatten_nested_columns` to smart pointer

ttnghia:
This changes the return type of `cudf::structs::detail::flatten_nested_columns` from `flattened_table` to `std::unique_ptr<flattened_table>`. Using smart pointer, we can forward declare the output and reduce dependency on the header `cudf/detail/structs/utilities.hpp`.

This also does some cleanup to remove the unused header `cudf/detail/structs/utilities.hpp` from many places.


vuule:
hey, you're not allowed to work today! Get this PR out of my sight! :D

karthikeyann (commented):
Added include cleanup for cudf/detail/structs/utilities.hpp
Rest of the change looks good.

cpp/include/cudf/detail/structs/utilities.hpp
```diff
@@ -167,7 +167,7 @@ class flattened_table {
  * @return `flatten_result` with flattened table, flattened column order, flattened null precedence,
```
vuule, line 167: the docs mention `flattened_result` in a few places, but I don't see this type in the code. Would be nice to clean this up in this PR, since it touches the same files/functions.

cpp/tests/structs/utilities_tests.cpp
```diff
@@ -305,9 +305,9 @@ TYPED_TEST(TypedStructUtilitiesTest, StructOfStructWithNullsAtAllLevels)
                                     expected_nums_col_3,
                                     expected_strings_col}};
 
-  CUDF_TEST_EXPECT_TABLES_EQUAL(expected,
-                                cudf::structs::detail::flatten_nested_columns(
-                                  table, {}, {}, cudf::structs::detail::column_nullability::FORCE));
+  auto flattened_table = cudf::structs::detail::flatten_nested_columns(
+    table, {}, {}, cudf::structs::detail::column_nullability::FORCE);
+  CUDF_TEST_EXPECT_TABLES_EQUAL(expected, flattened_table->flattened_columns());
```
vuule, line 310: Another comment that's not about the changes in this PR :D
The lines where `flattened_columns` is used read a bit poorly. `flattened_columns` to me implies that a vector of columns is returned. IMO `view` would be a better name, since the function returns a `table_view` and a call site like `flattened_table->view()` reads exactly right.
This is just a side note, not requesting any changes.

vuule (commented):
Changes look good, just some comments about surrounding code.

cpp/tests/structs/utilities_tests.cpp
ttnghia: Oh thanks.
I would like to change the entire name of the class `flattened_table` instead, as it is not a table but a data structure holding a table_view and many more things. So this should probably be addressed in a future PR.

cpp/include/cudf/detail/join.hpp
```diff
@@ -37,6 +36,10 @@
 template <typename T>
 class default_allocator;
 
+namespace cudf::structs::detail {
```
bdice, line 39: What is the benefit of this forward declaration? Does it reduce compile time or yield some other tangible improvement?

cpp/src/join/hash_join.cu
```diff
@@ -300,7 +300,7 @@ hash_join<Hasher>::hash_join(cudf::table_view const& build,
   // need to store off the owning structures for some of the views in _build
   _flattened_build_table = structs::detail::flatten_nested_columns(
     build, {}, {}, structs::detail::column_nullability::FORCE);
-  _build = _flattened_build_table;
+  _build = _flattened_build_table->flattened_columns();
```
bdice, line 303: This looks like it changes the contents of the build table, and goes beyond the scope of changing a return type. Can you explain more about why this changed?

Is there an overload of `cudf::detail::build_join_hash_table` accepting a `cudf::structs::detail::flattened_table`? I'm a little unsure of the context here.

bdice (changes requested):
I'm not sure if I fully understand the motivation for this change. What benefit does this get us?

cpp/include/cudf/detail/join.hpp
ttnghia: Forward declaration allows you to avoid the header `cudf/detail/structs/utilities.hpp` in this file.

ttnghia: Not just this file, but mainly avoid including that header in `operator.cuh` which is already huge and is included in many other places.

cpp/src/join/hash_join.cu
ttnghia: `_build = _flattened_build_table;` means `_build = _flattened_build_table.operator table_view();`. This is because of implicit conversion. And `operator table_view()` is just alias of `flattened_columns()`.

ttnghia: What causes you confused also has been pointed out here: https://github.com/rapidsai/cudf/pull/12878#discussion_r1126932626.

It seems that we should rename that function (`flattened_columns`) and also remove the implicit conversion soon.

bdice: Yes. This is definitely unclear.

bdice: Should we consider some or all of those changes in the scope of this PR?

cpp/tests/structs/utilities_tests.cpp
vuule: I agree that this should be a separate PR

vuule:
> I'm not sure if I fully understand the motivation for this change. What benefit does this get us?

@bdice Are you dubious about the API change, or only the forward declaration part?

bdice:
> > I'm not sure if I fully understand the motivation for this change. What benefit does this get us?
> 
> @bdice Are you dubious about the API change, or only the forward declaration part?

Mostly the forward declaration.

davidwendt:
> > > I'm not sure if I fully understand the motivation for this change. What benefit does this get us?
> > 
> > 
> > @bdice Are you dubious about the API change, or only the forward declaration part?
> 
> Mostly the forward declaration.

That's too bad. I think we should use forward declaration when we can to minimize headers including other headers as much as possible.

bdice:
@davidwendt I asked above for numbers on whether this actually affects compile time. https://github.com/rapidsai/cudf/pull/12878#discussion_r1128766964

If it makes an improvement here, I'm happy to use a forward declaration. I am somewhat concerned that if we use forward declarations aggressively, the names (and namespaces) of the forward declarations could fall out of sync with the rest of the codebase and require extra work. Is this a realistic concern? Are there any other downsides to consider, like the verbosity of re-declaring names in many places? I'm happy to change my opinion here, if I'm not thinking about this correctly.

ttnghia:
With forward declaration, you can avoid including the header (here is `utilities.hpp`) in many other headers. Doing so can avoid recompiling many translation units when this header is changed.

Out-of-sync is an issue, but it will be detected right away because whenever that happens you won't be able to compile. The compiler will yell loudly that such (forward declared) class does not exist.

I'll test the compile time and will report later when I have the numbers.

ttnghia:
Okay, so I made some trivial changes somewhere in `detail/structs/utilities.hpp` (like `int x; (void)x;`) to force recompiling the TUs depending on it. I also use the branch in https://github.com/rapidsai/cudf/pull/12879 to test because that branch will adopt the changes in this PR in which it has more forward declaration of `cudf::structs::detail::flattened_table` in `row_operators.cuh`.

If not using forward declaration of `cudf::structs::detail::flattened_table`: 
 * Number of TUs recompiled: 67
 * Compile time: 6:46s
And with forward declaration: 
 * Number of TUs recompiled: 34
 * Compile time: 3:35s

cpp/include/cudf/detail/join.hpp
bdice: Thanks for that data in the other comment. Looks great. Resolving this.

bdice (approved):
Looks good. There is an open question about improving naming and conversion behavior. Let’s open an issue describing what needs done for the future.

ttnghia:
> Looks good. There is an open question about improving naming and conversion behavior. Let’s open an issue describing what needs done for the future.
> 
> [#12878 (comment)](https://github.com/rapidsai/cudf/pull/12878#discussion_r1128255260)

Yes sure. Issue filed: https://github.com/rapidsai/cudf/issues/12911

karthikeyann (approved):
LGTM 👍

Issue #11672, opened 2022-09-08
[FEA] Enable structs of lists in the new row lexicographic comparator


vyasr:
**Is your feature request related to a problem? Please describe.**
#11129 enables comparisons of list dtypes with the new comparator. In principle, this code should (in conjunction with prior work like #10164) be sufficient to enable structs of lists (but not lists of structs) via the struct decomposition approach used to preprocess structs. Currently, we forbid comparison of lists of structs in the comparator. During development of #11129 I attempted to remove this restriction, but I observed that this led to [unexpected test failures of [groupby_structs_test/lists_are_unsupported](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/cudf/job/prb/job/cudf-gpu-test/CUDA=11.0,GPU_LABEL=driver-450,LINUX_VER=centos7,PYTHON=3.8/10269/). These failures were not caused by a failure of `EXPECT_THROW` (which would be natural since I removed a restriction), but were actual failures in the computed outputs. I was unable to reproduce these failures locally, indicating that there is perhaps some unknown indeterminacy in the ordering of some step in the comparison.

**Describe the solution you'd like**
The struct of list restriction in [`check_lex_compatibility`](https://github.com/rapidsai/cudf/blob/branch-22.10/cpp/src/table/row_operators.cu) should be removed and [the failing test](https://github.com/rapidsai/cudf/blob/branch-22.10/cpp/tests/groupby/structs_tests.cpp#L318) should be fixed such that structs of lists are properly supported.

ttnghia:
For comparing lists, we need to generate dremel data. However:
 * Dremel data is generated only for lists columns that are not children of any other column: https://github.com/rapidsai/cudf/blob/9c9dd54f5ec6834ed87f26828c5e3878ced39b2d/cpp/src/table/row_operators.cu#L265-L270
 * On device, dremel data is considered for comparison only in the same situation: https://github.com/rapidsai/cudf/blob/9c9dd54f5ec6834ed87f26828c5e3878ced39b2d/cpp/include/cudf/table/experimental/row_operators.cuh#L559-L567
[...]

PR #12787, opened 2023-02-16
Update `join` to use experimental row hasher and comparator

divyegala:
## Description
Part of #11844. I will create a separate PR for `mixed_join`.

Compilation times:
`main` https://github.com/rapidsai/cudf/commit/94bbc82a117a96979d1c5d7949ba213ade88c3be : `16m47.513s`
This PR https://github.com/rapidsai/cudf/pull/12787/commits/5d75db866fd0b9239601286b3fc360e8c2a1cd38 : `16m47.520s`

Benchmarks: https://github.com/rapidsai/cudf/pull/12787#issuecomment-1434126344

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12879, opened 2023-03-03
 Fully support lexicographic comparison for nested types in table self-comparator

ttnghia:
This PR implements full support for lexicographic comparison for nested types, for table self-comparator (support for two-table comparator will come in a follow-up PR).

**Warning:** Supporting structs-of-lists is a temporary workaround, not an optimal long-term solution.

Algorithm:
 * If a column contains structs-of-lists: flatten it using `cudf::structs::detail::flatten_nested_columns`. While waiting for a long-term optimal solution, this workaround is fairly simple and efficient.
 * If a column contains lists-of-structs: follow the algorithm described in https://github.com/rapidsai/cudf/issues/11222.
 * If the input doesn't contain any lists-of-structs or structs-of-lists, there will not be any change/transformation applied to it, and the performance will not be impacted at all.

Depends on:
 * https://github.com/rapidsai/cudf/pull/12878
 * https://github.com/rapidsai/cudf/pull/12892

Closes:
 * https://github.com/rapidsai/cudf/issues/11222
 * https://github.com/rapidsai/cudf/issues/11672

---

Tested by spark-rapids integration tests. All tests pass. Data generator used to test:
```
struct_of_array_gen = [StructGen([['child0', ArrayGen(int_gen, max_length=10)]]),
                       StructGen([['child0', byte_gen],
                                  ['child1', ArrayGen(int_gen, max_length=10)],
                                  ['child2', int_gen]]),
                       StructGen([['child0', ArrayGen(int_gen, max_length=10)],
                                  ['child1', int_gen]]),
                       StructGen([['child0', ArrayGen(int_gen, max_length=10)],
                                  ['child1', ArrayGen(int_gen, max_length=10)],
                                  ['child2', ArrayGen(int_gen, max_length=10)]])]
nested_array_of_struct_gen = [ArrayGen(sub_gen, max_length=10) for sub_gen in struct_of_array_gen]
deeply_nested_array_gen = [ArrayGen(sub_gen, max_length=10) for sub_gen in nested_array_of_struct_gen] + \
                          [ArrayGen(sub_gen, max_length=10) for sub_gen in struct_of_array_gen]
deeply_nested_struct_gen = [StructGen([['child0', sub_gen] for ind, sub_gen in enumerate(deeply_nested_array_gen)])]

# This is used to test:
deeply_nested_gen = struct_of_array_gen + nested_array_of_struct_gen + deeply_nested_array_gen + \
                    deeply_nested_struct_gen
```


divyegala:
Just wanted to point out that changes in this PR will clash with changes in my open PR https://github.com/rapidsai/cudf/pull/12787. We should ensure that we only flatten struct columns if they are composed of list columns, since the new comparator takes care of primitive `Struct<...>` types.

Also, would it be possible to add to the scope of this PR to update `contains_table.cu` and `struct_minmax_util.cuh` to the new comparators? We can use 12787 for updating `hash_join.cu` as a follow-on

cpp/src/table/row_operators.cu
```diff
@@ -344,39 +345,227 @@ namespace row {
...
+      return {std::move(transformed_input), std::move(ranks)};
+    } else if (child.type().id() == type_id::LIST) {
+      auto [new_child, ranks_cols] = transform_lists_of_structs(child, stream);
+      if (ranks_cols) {
+        auto transformed_input = make_transformed_input(new_child);
+        return {transformed_input, std::move(ranks_cols)};
+      }
+    }
+  } else if (input.type().id() == type_id::STRUCT) {
+    CUDF_UNREACHABLE("Structs columns should be flattened before calling this function.");
```
divyegala, line 405: Can we not get struct columns at this point? After all, we are only flattening instances of `struct<list<...>>`

ttnghia: If there is any `struct<list<>>` column, the entire input table will be flattened. This helps simplifying other subsequent operations a lot. Otherwise, we have to:
 * Make correction of the column order and null order for both non-flattened and flattened columns, and
 * Maintains a vector of `unique_ptr<flattened_table>` to store different results from flattening columns.
 * Have to consider `list<struct<>>` while transforming lists of structs in this function, and return a list of columns (`vector<unique_ptr<column>>`) instead of just one column because we may have structs of multiple lists columns having structs child.

divyegala: Ah, I see the `std::any_of` in `has_nested_structs_of_lists`. Can we not move the `has_nested_structs_of_lists` check [here](https://github.com/rapidsai/cudf/blob/96e28b8629a129adab8f011b92eae87b93958673/cpp/src/structs/utilities.cpp#L171) instead? That way, we can always call `flatten_nested_structs` without any if-else. Any struct columns that have no lists in them will go in the `else` portion of the code I linked above

divyegala: That way we still flatten the entire table, but reduce the bookkeeping overhead by flattening only structs of lists. No correction to column/null order will be needed, and we'll be able to get away with a single `unique_ptr<flattened_table>`

ttnghia: Unfortunatley no. That utility function needs to **always** flatten structs columns, and is used in other places such as in join.

ttnghia: Even if we re-implement that utility function (so we can solve the first point I listed above), we still have to deal with 2 remaining points.

divyegala: Yes, for now. But if you think this strategy is viable then I can revisit this as a follow-on once all my comparator update PRs go in and I remove explicit flattenings from the whole code-base. Then, we'll only have implicit flattenings left through the new comparator.

I don't think we'll face point 2 we'll still flatten the whole table, but the actual struct flattenings will be limited to struct of lists. Other struct columns will be stored as is.

I will put some more thought for point 3. You are right that this strategy breaks for `struct<list<struct<>>>` because even after the conversion we are left with `struct<list<size_type>>`. I was hoping `struct<list<size_type>>` would work out-of-the-box with the new comparator.

But yes, for now we will keep the implementation as you have written it since the alternative is not viable yet.

divyegala: > adding a parameter allowing to choose if we want to flatten any structs columns, or flatten only structs-have-lists columns

Yes!

Here's what we can do for this PR:
1. Move `has_nested_structs_of_lists` to `structs/utilities.cpp`
2. Add a default parameter to `flatten_nested_columns`. Default option is to flatten only structs of lists, `join` and other affected algorithms can use non-default option to flatten all structs

Once I finish the comparator update, I remove the new parameter entirely and we only flatten struct of lists from that point onwards.

However, if you'd rather not worry about it in this PR then I'll take it up as a follow on after my comparator updates

ttnghia: Got it, thanks. Then let's do it in the follow-up PR instead, so we can minimize code change in this one.

bdice:
@ttnghia Could you run benchmarks on this change?

ttnghia:
> @ttnghia Could you run benchmarks on this change?

Before this, we can't compare structs-of-lists + lists-of-structs so we don't have any baseline. With this, the input will be passed through without any changes so comparing input without structs-of-lists + lists-of-structs in theory should not be affected. I planned to check this with a benchmark to make sure I'm correct, but also wanted to make sure if this is also what you want to see.

vyasr (changes requested):
Before we resort to flattening to support structs of lists, could we try to resolve https://github.com/rapidsai/cudf/issues/11672 properly? We originally expected structs of lists to "just work" ™️, and the only reason I didn't investigate further at the time was the urgency imposed on getting the comparator merged. I would like to have a sense for the feasibility of the long-term solution before we resort to a workaround.

ttnghia:
I ran a benchmark comparing sorting lists columns (nested lists-of-lists, or LoL) and sorting lists-of-structs columns (LoS). These LoS columns are generated by just putting the children of the LoL columns under a structs column. So under the hood, these LoL and LoS columns contain the same information to compare.

The benchmarks show very interesting results:
```
# sort_list

## [0] Quadro RTX 6000

|  size_bytes  |  depth  |  num_columns  |  null_frequency  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |             Diff |   %Diff |  Status  |
|--------------|---------|---------------|------------------|------------|-------------|------------|-------------|------------------|---------|----------|
|     2^10     |    1    |      10       |        0         |   4.075 us |     204.71% |   2.184 us |     291.37% |        -1.891 us | -46.40% |   PASS   |
|     2^18     |    1    |      10       |        0         |  32.663 ms |       3.13% |  38.994 ms |       2.32% |         6.331 ms |  19.38% |   FAIL   |
|     2^24     |    1    |      10       |        0         |   9.545 ms |       9.76% |  29.418 ms |       5.41% |        19.873 ms | 208.21% |   FAIL   |
|     2^10     |    2    |      10       |        0         |   6.744 us |      38.68% |   2.906 us |     207.01% |        -3.838 us | -56.91% |   FAIL   |
|     2^18     |    2    |      10       |        0         | 901.082 ms |       4.00% |  28.292 ms |       3.16% |   -872790.516 us | -96.86% |   FAIL   |
|     2^24     |    2    |      10       |        0         | 113.982 ms |       0.86% | 115.617 ms |       0.99% |         1.635 ms |   1.43% |   FAIL   |
|     2^10     |    3    |      10       |        0         |   9.589 us |      59.38% |   3.461 us |      85.14% |        -6.128 us | -63.91% |   FAIL   |
|     2^18     |    3    |      10       |        0         |   9.805 us |      42.68% |   3.259 us |      96.03% |        -6.546 us | -66.76% |   FAIL   |
|     2^24     |    3    |      10       |        0         |   55.376 s |        inf% | 154.392 ms |       2.17% | -55221986.520 us | -99.72% |   FAIL   |
|     2^10     |    4    |      10       |        0         |  12.853 us |      42.00% |   4.005 us |     154.35% |        -8.848 us | -68.84% |   FAIL   |
|     2^18     |    4    |      10       |        0         |  12.584 us |      39.01% |   3.753 us |      78.64% |        -8.831 us | -70.18% |   FAIL   |
|     2^24     |    4    |      10       |        0         |  13.187 us |      91.99% |   3.787 us |      85.74% |        -9.400 us | -71.28% |   PASS   |
```

In summary:
 * For shallow LoS, sorting them is much slower than sorting LoL. This is due to the overhead of computing ranks.
 * However, when the LoL columns have higher depth (just more than 1 depth level) then sorting them becomes slower than sorting LoS due to the iterative mechanism of the comparator when working with multiple levels of lists.

ttnghia:
> Before we resort to flattening to support structs of lists, could we try to resolve #11672 properly? We originally expected structs of lists to "just work" tm, and the only reason I didn't investigate further at the time was the urgency imposed on getting the comparator merged. I would like to have a sense for the feasibility of the long-term solution before we resort to a workaround.

@vyasr I think I found what's wrong with structs-of-lists: For comparing lists, we need to generate dremel data. However:
 * Dremel data is generated only for lists columns that are not children of any other column: https://github.com/rapidsai/cudf/blob/9c9dd54f5ec6834ed87f26828c5e3878ced39b2d/cpp/src/table/row_operators.cu#L265-L270
 * On device, dremel data is considered for comparison only in the same situation: https://github.com/rapidsai/cudf/blob/9c9dd54f5ec6834ed87f26828c5e3878ced39b2d/cpp/include/cudf/table/experimental/row_operators.cuh#L559-L567

So in order to fix this completely, we need both to generate and consider dremel data for nested lists columns. I'm not sure about the complexity of such work, so it should be done in a separate follow up PR.

bdice (commented):
Submitting this partial review. I will review the rest at a later time.

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -811,6 +816,14 @@ struct preprocessed_table {
...
   std::optional<std::vector<detail::dremel_data>> _dremel_data;
   std::optional<rmm::device_uvector<detail::dremel_device_view>> _dremel_device_views;
+
+  // Auxiliary data generated from `cudf::structs::detail::flatten_nested_columns`
+  // that needs to be kept alive.
+  std::unique_ptr<cudf::structs::detail::flattened_table> _flattened_input_aux_data;
+
+  // Auxiliary data generated from transforming lists-of-structs into lists-of-integer
+  // that needs to be kept alive.
+  std::vector<std::unique_ptr<column>> _transformed_structs_aux_data;
```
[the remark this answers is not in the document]
ttnghia: This is the to store the output of `cudf::rank` thus must be `unique_ptr`.

cpp/src/table/row_operators.cu
```diff
@@ -344,39 +345,229 @@ namespace row {
...
+ *
+ * For lists-of-structs column at any nested level, the child structs column will be replaced by an
+ * integer column of its ranks generated using `cudf::rank()`.
+ *
+ * If the input column is not lists-of-structs, or does not contain lists-of-structs at any nested
+ * level, the input will be passed through.
+ *
+ * @param input The input column to transform
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @return A pair of new column_view representing the transformed input and the generated rank
```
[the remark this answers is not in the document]
ttnghia: We actually don't need to call `make_lists_column` as that will copy the offsets column of the input lists. For the case of two-table comparator when we concate the child columns, we will split the output ranks into two ranks column and still apply the input null masks and offsets of the two input lists columns.

FYI: I've implemented the code for handling two-table comparator here: https://github.com/rapidsai/cudf/pull/12953/files#diff-c143ae2ecfe63796e604e5854f4a47cf6b831ebe5d68ce7c19a0e1533fceddc7R394-R413

cpp/src/table/row_operators.cu
```diff
@@ -344,39 +345,229 @@ namespace row {
...
+    // Should not use sliced child because we reuse the input's offset value and offsets child.
+    auto const child = input.child(lists_column_view::child_column_index);
+
+    if (child.type().id() == type_id::STRUCT) {
+      // Dense ranks can accurately reflect the order of structs: structs compared equal will have
+      // the same rank values.
+      // However, first ranks are computed faster and are good enough for ordering them. Structs
+      // compared equal always have consecutive rank values (in stable order) thus they are still
+      // sorted correctly by their ranks.
+      auto ranks             = cudf::detail::rank(child,
```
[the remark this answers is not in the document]
ttnghia: This is really tricky. The `rank` API calls `sort_order` which in turn uses the `less<>` comparator with default template argument `PhysicalElementComparator = sorting_physical_element_comparator`. We cannot control that element comparator in this preprocessing step, as we have no way to pass in such element comparator into `cudf::rank`.

So the only way I can think of now is to enforce using that default element comparator later and nothing else (of course, only if `cudf::rank` has been called during preprocessing).

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -721,30 +726,30 @@ struct preprocessed_table {
    * contain an empty `dremel_device_view`. As such, this uvector has as many elements as there are
    * columns in the table (unlike the `dremel_data` parameter, which is only as long as the number
    * of list columns).
+   * @param flattened_input_aux_data The data structure generated from
+   * `cudf::structs::detail::flatten_nested_columns` that contains additional information used for
+   * row comparisons.
+   * @param transformed_structs_columns The intermediate columns resulted from transforming child
```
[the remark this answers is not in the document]
ttnghia: I rename it into `structs_ranked_columns`.

ttnghia:
Close this. I will separate it into two separate PRs for structs-of-lists and lists-of-structs.

PR #12882, opened 2023-03-06
[ENH] Implement `groupby.sample`

wence-:
## Description
To do so, obtain the group offsets and values (and hence index). Sample within each group, and then pull out rows from the original object.

The fastest way to do this in Python is via the builtin random library, since neither numpy nor cupy offer a broadcasted/ufunc random.sample, and looping over the groups is very slow using either of them. Looping over the groups and using python random.sample is also slow, but less so.

## Checklist


wence-:
We needed groupby-sample for some work, this was the fastest implementation I could come up with that didn't involve writing specific C++. This is only a partial implementation of the full pandas API, and I am using the _Python_ random module (since both cupy and numpy are horribly slow when taking samples: there is no ufunc/vectorised version of `xxx.random.choice`).

Is this is a reasonable approach? Or should I come up with some way of doing the sample all-at-once on device?

Trivial benchmark:

```
import cudf
import cupy as cp
df = cudf.DataFrame({"a": cp.random.randint(0, 1000, size=10_000_000), "b": cp.arange(10_000_000)})
pdf = df.to_pandas()
%time pdf.groupby("a").sample(2) # => 1.2s
%time df.groupby("a").sample(2) # => 40ms
```

Bumping to 100_000 groups, pandas takes 4s, this approach takes 260ms.

shwina:
A while ago I prototyped a version of `group_sample` that perhaps you could salvage something from?

https://gist.github.com/shwina/ac4cdffe00ce341a65e5e27b78d0b2a0

wence-:
> A while ago I prototyped a version of `group_sample` that perhaps you could salvage something from?
> 
> https://gist.github.com/shwina/ac4cdffe00ce341a65e5e27b78d0b2a0
> 
> (feedback welcome -- I'm not sure if I might have overlooked something)

To inline that code for discussion

```python
def groupby_sample(df, by, n=2):
    df = df.sample(frac=1).reset_index(drop=True) # randomise order of dataframe
    return df.loc[df.groupby(by)[by].rank("first") <= 2, :]
```

With some futzing around `groupby.rank` not working for non-integer data which I've elided (which makes sense because some dtypes might not admit a total order).

This works because for `replace=False` with `weights is None` and `frac is None`, `df.sample` commutes with `df.groupby.rank`.

As soon as we introduce keyword arguments into the mix (and to be fair, I didn't get round to implementing those in this PR), the commutator is non-zero, and so we need an alternate approach (e.g. groupby.sample(replace=True) samples with replacement _within_ each group, whereas `df.sample(replace=True)` would sample with replacement from the whole dataframe).

shwina:
I agree -- perhaps we could use the `groupby.rank()` based approach for the simplest case (no kwargs), and iterate over the groups for the more complex cases?

wence-:
Benchmarks incoming.

@bdice suggested a way to squash one more slow path, but I worry that it might suffer from unfair sampling:

```
import numpy as np

group_sizes = np.array([5, 3, 4])
samples_from_each_group = np.array([2, 3, 0])

total_samples = np.sum(samples_from_each_group)

group_sample_modulus = np.repeat(group_sizes, samples_from_each_group)

# Technically we probably need a range [0, np.lcm.reduce(group_sizes)) to
# ensure fair sampling over the range?
rng = np.random.default_rng(12345)
rints = rng.integers(low=np.iinfo(np.int_).min, high=np.iinfo(np.int_).max, size=total_samples)

group_sample_ids = rints % group_sample_modulus
```

So I haven't implemented this one.

The general approach is ready to look at though.

wence-:
`.head/.tail` are pretty straightforward as well. One just concats indices from slices offset by the `offset` array that comes back from groupby._grouped()

bdice:
@wence- If you can do a segmented argsort, I may have a solution for sampling without replacement (SWOR). Based on https://timvieira.github.io/blog/post/2019/09/16/algorithms-for-sampling-without-replacement/

Use a modified `swor_exp` from the linked article, like this:
```python
import numpy as np

def swor_exp_segmented(total_samples, segment_indices):
    E = -np.log(np.random.uniform(0,1,size=total_samples))
    # You can implement segmented_argsort using CUB. There are probably primitives in cupy/cudf that can do this?
    return segmented_argsort(E, segment_indices)
```

This gives you a randomly sorted list of indices within each segment. Then you can use a gather to fetch the desired number of indices from each segment, and then use _that_ result as a gather map of indices to fetch from the original groupings.

wence-:
> @wence- If you can do a segmented argsort, I may have a solution for sampling without replacement (SWOR). Based on https://timvieira.github.io/blog/post/2019/09/16/algorithms-for-sampling-without-replacement/
> 
> **edit:** ooh, maybe not. The analytical probabilities of SWOR may be different here because the sampling should be _with respect to each group_. ~**I retract this idea.**~ A proper expression is probably derivable from similar reasoning to that presented in the article, though.
> 
> **edit 2:** On second thought, maybe this is still legitimate? The probabilities don't depend on group size, and doing this for each segment individually should be equivalent to doing all segments together, so long as you use the segmented argsort correctly.
> 
> Use a modified `swor_exp` from the linked article, like this:
> 
> ```python
> import numpy as np
> 
> def swor_exp_segmented(total_samples, segment_indices):
>     E = -np.log(np.random.uniform(0,1,size=total_samples))
>     # You can implement segmented_argsort using CUB. There are probably primitives in cupy/cudf that can do this?
>     return segmented_argsort(E, segment_indices)
> ```
> 
> This gives you a randomly sorted list of indices within each segment. Then you can use a gather to fetch the desired number of indices from each segment, and then use _that_ result as a gather map of indices to fetch from the original groupings.

Oh that's a cute trick!

bdice:
In fact, you don’t even need the “log” bit. Assuming all the samples have equal weight, _any_ random permutation of the group indices will be equally likely.

wence-:
> In fact, you don’t even need the “log” bit. Assuming all the sample elements have equal weight, _any_ random permutation of the group indices will be equally likely.

Yeah, so that post is fast ways of sampling from the categorical distribution (where you have specified weights for each element of the group, and all elements are unique). That will be useful (since groupby.sample with weights is something that is in the pandas API, though I am punting on it in this PR), though here we have a slightly simpler problem, we have a group of size $n$ which we can treat as $n$ categories, each with probably $\frac{1}{n}$. So then, if you want a sample of size $k$, you can just take a permutation of the group and select the first $k$ entries.

Indeed, this is exactly the implementation used in the PR, for the case of `groupby.sample(n=k, replace=False)`: permute the whole data frame, group and take the first `k` from each group. This could be extended to the `groupby.sample(frac=0.2, replace=False)` case if there's a good way to push the sample sizes into `groupby.rank(...)`. The downside to this implementation is that the memory footprint is 3x the input dataframe (rather than only 2x when you group and then select).

I'll noodle through a few more ideas...

bdice:
> The downside to this implementation is that the memory footprint is 3x the input dataframe (rather than only 2x when you group and then select).

It’s not so bad! You’re only dealing with integer columns of indices for gathering, not full dataframes (all columns) as far as I can tell. Usually I’d say that adding 2-3 columns of integers with the same number of rows as the dataframe is an acceptable memory cost. It is hefty for single-column dataframes but not a big deal for 10 column dataframes, as a fraction of the dataframe memory.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -699,6 +700,131 @@ def ngroup(self, ascending=True):
...
+
+        Returns
+        -------
+        New dataframe or series with samples of appropriate size drawn
+        from each group
+
+        """
+        if weights is not None:
+            raise NotImplementedError(
+                "Sorry, sampling with weights is not supported"
```
bdice, line 738: Don’t apologize! 😉 It’s a different tone than the rest of our errors.
```suggestion
                "Sampling with weights is not supported"
```

python/cudf/cudf/tests/test_groupby.py
```diff
@@ -2972,3 +2973,73 @@ def test_groupby_dtypes(groups):
...
+            # These are valid weights, but we don't implemented this yet.
+            df.groupby("a").sample(n=1, weights=[1 / len(df)] * len(df))
+
+    @pytest.mark.parametrize("frac", [0, 1 / 3, 1 / 2, 2 / 3, 1])
+    @pytest.mark.parametrize("replace", [False, True])
+    def test_fraction_rounding(self, df, by, frac, replace):
+        result = (
+            df.groupby(by).sample(frac=frac, replace=replace).sort_values("a")
+        )
+        assert_eq(self.expected(df, frac=frac), result.reset_index(drop=True))
```
wence-, line 3045: TODO, these tests don't check for statistical correctness that the sampling in each group is uniform.

python/cudf/cudf/tests/test_groupby.py
```diff
@@ -2972,3 +2973,73 @@ def test_groupby_dtypes(groups):
...
+        params=[
+            ["a", "a", "b", "b", "c", "c", "c", "d", "d", "d", "d", "d"],
+            [1, 1, 2, 2, 3, 3, 3, 4, 4, 4, 4, 4],
+        ],
+        ids=["str-group", "int-group"],
+    )
+    def df(self, request):
+        return cudf.DataFrame(
+            {"a": request.param, "b": request.param, "v": request.param}
+        )
```
wence-, line 2989: TODO: use non-range index as well.

wence- (commented):
Some notes.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -699,6 +700,131 @@ def ngroup(self, ascending=True):
...
+            else:
+                rng = cp.random.default_rng(seed=random_state)
+                offsets_col = as_column(group_offsets)
+                values = [index]
+                keys = [as_column(rng.random(size=len(index)))]
+                (indices,) = segmented_sort_by_key(
+                    values, keys, offsets_col, [], []
+                )
+                indices = cp.asarray(indices.data_array_view(mode="read"))
+            # For many groups this is the slowest part.
```
bdice: Is the slow part because this is numpy? The slow bit is generating “to_take” and not the actual taking, right? I have a solution I can write up in a bit if that’s accurate

python/cudf/cudf/core/groupby/groupby.py
wence-: Done.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -699,6 +700,125 @@ def ngroup(self, ascending=True):
...
+                keys = [as_column(rng.random(size=len(index)))]
+                (indices,) = segmented_sort_by_key(
+                    values, keys, offsets_col, [], []
+                )
+                indices = cp.asarray(indices.data_array_view(mode="read"))
+            # Which indices are we going to want?
+            mask = np.zeros(len(index), dtype=bool)
+            for offset, sample_size in zip(group_offsets, samples_per_group):
+                mask[offset : offset + sample_size] = True
+            indices = indices[mask]
```
bdice, line 815: I would like to test an alternative here, which might be more efficient with cupy (no raw loops, just array operators). I don't know how it will compare to the CPU NumPy implementation. Here's the sketch. You might benchmark it for various group sizes / sample sizes.
```python
import numpy as np

num_groups = 100_000
items_per_group = 10

group_offsets = np.arange(0, num_groups*items_per_group, items_per_group)
samples_per_group = np.repeat(items_per_group//3, num_groups)
indices = np.arange(0, num_groups*items_per_group)

def take_mask(indices, group_offsets, samples_per_group):
    # Which indices are we going to want?
    mask = np.zeros(len(indices), dtype=bool)
    for offset, sample_size in zip(group_offsets, samples_per_group):
        mask[offset : offset + sample_size] = True
    return indices[mask]

def take_offsets(indices, group_offsets, samples_per_group):
    to_take = np.arange(np.sum(samples_per_group))
    exclusive_scan = np.empty(len(samples_per_group), dtype=np.int_)
    exclusive_scan[0] = 0
    exclusive_scan[1:] = np.cumsum(samples_per_group[:-1])
    to_take += np.repeat(group_offsets - exclusive_scan, samples_per_group)
    return indices[to_take]

mask_result = take_mask(indices, group_offsets, samples_per_group))
offsets_result = take_offsets(indices, group_offsets, samples_per_group))
```

wence-: I'll have a go on monday. Also, I realise this in a broken state as soon as I added non-default indices to the dataframe (so I'll untangle that too).

python/cudf/cudf/tests/test_groupby.py
wence-: tl;dr: this works.

I did some statistical tests using this test dataframe:
```
test_df = cudf.DataFrame({"a": [1, 1, 1, 1, 1, 2, 2, 2, 2, 2], "b": [0, 1, 2, 3, 4, 0, 1, 2, 3, 4]}
```
So if I groupby "a" and `sample(n=4, replace=False)` I should draw length four subsequences from all length five permutations for both groups. For this setup there are 120 different results. So the expected probability of obtaining any given sample is 1/120.

Doing this 50_000 times, I observe the empirical distribution of the sample counts, and use `scipy.stats.normaltest` to check for normality. I can't reject at the 5% significance level (p=0.7) and so the null hypothesis that the data are from a normal distribution is reasonable.

I also use a nonparametric Kolmogorov-Smirnov test to compare to sample drawn using `np.random.choice([0, 1, 2, 3, 4], size=4, replace=False)` and again cannot reject the null hypothesis that the two datasets were drawn from the same distribution.

For sampling with replacement, I choose `n=3` (125 possible outcomes) and do the same thing again. With 50_000 samples I cannot reject the null hypothesis that the data are normal, nor can I reject the null hypothesis that the data are from the same underlying distribution as the "naive" numpy implementation.

So, I think this is good.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -699,6 +700,135 @@ def ngroup(self, ascending=True):
...
+            indices += np.repeat(group_offsets[:-1], samples_per_group)
+        else:
+            # Approach: do a segmented argsort of the index array and take
+            # the first samples_per_group entries from sorted array.
+            # We will shuffle the group indices and then pick them out
+            # from the grouped dataframe index.
+            nrows = len(group_values)
+            indices = cp.arange(nrows, dtype=np.int32)
+            if len(size_per_group) < 500:
+                # Empirically shuffling with cupy is faster at this scale
```
wence-, line 808: I suspect this is because cupy is really doing a shuffle here which is O(N) and is just data-movement, whereas the shuffle-by-argsort is NlogN, moves more data and requires comparisons.

wence-:
I did some basic benchmarking of this code.

```python
def create_df(n, nunique):
    return cudf.DataFrame(
        {"a": cp.random.randint(0, nunique, size=n), "b": cp.arange(n)}
    )


def sample_n(df, n, replace):
    start = time.time()
    _ = df.groupby("a").sample(n=n, replace=replace)
    end = time.time()
    return end - start


def sample_frac(df, frac, replace):
    start = time.time()
    _ = df.groupby("a").sample(frac=frac, replace=replace)
    end = time.time()
    return end - start


def gather_results():
    data = []
    for n in [100_000, 1_000_000, 10_000_000, 100_000_000]:
        print(f"DataFrame size {n=}")
        nuniques = [10, 100, 1000, 10_000, 100_000, 1_000_000, 10_000_000, 100_000_000]
        nuniques = nuniques[: nuniques.index(n) - 1]
        for nunique in nuniques:
            df = create_df(n, nunique)
            for pandas in [False, True]:
                if pandas:
                    idf = df.to_pandas()
                else:
                    idf = df
                for replace in [False, True]:
                    nsamp = 5
                    for nsamp in [1, 2, 10, 50]:
                        frac = nunique * nsamp / n
                        t = sample_n(idf, nsamp, replace)
                        data.append(
                            (
                                n,
                                nunique,
                                nsamp,
                                frac,
                                replace,
                                False,
                                "pandas" if pandas else "cudf",
                                t,
                            )
                        )
                        gc.collect()
                        t = sample_frac(idf, frac, replace)
                        data.append(
                            (
                                n,
                                nunique,
                                nsamp,
                                frac,
                                replace,
                                True,
                                "pandas" if pandas else "cudf",
                                t,
                            )
                        )
                        gc.collect()
    return pd.DataFrame(
        data,
        columns=[
            "df_size",
            "unique_vals",
            "sample_size",
            "sample_frac",
            "replace",
            "use_frac",
            "backend",
            "time",
        ],
    )
```

Run on an A6000 (cudf) and an intel xeon gold 6226R CPU (pandas).

Parquet data for further analysis attached ([groupby-sample-benchmarks.zip](https://github.com/rapidsai/cudf/files/10959380/groupby-sample-benchmarks.zip)), but I also upload some faceted-plots (log-scale on the timings axis).

tl;dr: For a dataframe with $10^8$ rows and $10^6$ groups (each around size 100), sampling _without replacement_ groups of size 50 takes ~0.9s, sampling _with replacement_ takes ~1.3s (pandas 50-55s for the same operation).

With replacement sampling is slower because there isn't a fast cupy way, AFAICT, to generate the random numbers which takes about a second of the run. Without replacement sampling could be faster if cupy had a batched/segmented shuffle.

In contrast, if one does the same thing but only selects a single value from each group, then the cudf implementation takes between 0.25 and 0.5s.

`replace=False, use_frac=False`
no-replace-no-use-frac

`replace=False, use_frac=True`
no-replace-use-frac

`replace=True, use_frac=False`
replace-no-use-frac

`replace=True, use_frac=True`
replace-use-frac

python/cudf/benchmarks/API/bench_dataframe.py
```diff
@@ -104,6 +104,33 @@ def bench_groupby_agg(benchmark, dataframe, agg, num_key_cols, as_index, sort):
...
+        minsize = grouper.size().min()
+        target_size = numpy.round(
+            target_sample_frac * minsize, decimals=0
+        ).astype(int)
+        kwargs = {"n": target_size, "replace": replace}
+
+    def _():
+        return grouper.sample(**kwargs)
+
+    benchmark(_)
```
bdice, line 131: Does this work?
```suggestion
    benchmark(grouper.sample, **kwargs)
```

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -870,6 +871,135 @@ def ngroup(self, ascending=True):
...
+            raise NotImplementedError(
+                "Only integer seeds are supported for random_state "
+                "in this case"
+            )
+        # Get the groups
+        # TODO: convince Cython to convert the std::vector offsets
+        # into a numpy array directly, rather than a list.
+        # TODO: this uses the sort-based groupby, could one use hash-based?
+        _, offsets, _, group_values = self._grouped()
+        group_offsets = np.asarray(offsets, dtype=np.int32)
```
bdice, line 944: Use the common definition of `size_type_dtype` if this is meant to map to libcudf's `size_type`. Applies in several places below.

https://github.com/rapidsai/cudf/blob/bf18cea1461b63ffef32a5404b7552d1e90240e9/python/cudf/cudf/_lib/types.pyx#L20

bdice (commented):
Nice job! A few comments.

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -143,6 +145,70 @@ def order_by(list columns_from_table, object ascending, str na_position):
...
+    cdef vector[null_order] c_null_precedence
+    cdef unique_ptr[table] result
+    ncol = len(values)
+    column_order = column_order or [True] * ncol,
+    null_precedence = null_precedence or ["first"] * ncol,
+    for asc, null in zip(column_order, null_precedence):
+        if asc:
+            c_column_order.push_back(order.ASCENDING)
+        else:
+            c_column_order.push_back(order.DESCENDING)
```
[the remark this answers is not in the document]
wence-: I guess I was just copying the previous `order_by` implementation, but done.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -870,6 +871,135 @@ def ngroup(self, ascending=True):
...
+        random_state: Union[np.random.RandomState, int, None] = None,
+    ):
+        """Return a random sample of items in each group.
+
+        Parameters
+        ----------
+        n
+            Number of items to return for each group, if sampling
+            without replacement must be at most the size of the
+            smallest group. Cannot be used with frac. Default is
```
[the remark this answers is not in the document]
wence-: I think it doesn't really matter

python/cudf/cudf/core/groupby/groupby.py
wence-: Done (also in the `head`/`tail` implementation where it slipped through).

wence- (commented):
Thanks, I think everything is addressed.

bdice (approved):

PR #12883, opened 2023-03-06
Expect cupy to now support bool arrays for dlpack.

vyasr:
## Description

The most recent versions of cupy now support bool arrays in dlpack (see [the release notes](https://github.com/cupy/cupy/releases/tag/v11.6.0) and [the PR adding the feature](https://github.com/cupy/cupy/pull/7376)). Our DataFrame protocol tests were expecting a failure.

## Checklist

### DIFF
--- a/python/cudf/cudf/tests/test_df_protocol.py
+++ b/python/cudf/cudf/tests/test_df_protocol.py
@@ -39,15 +39,12 @@ def assert_buffer_equal(buffer_and_dtype: Tuple[_CuDFBuffer, Any], cudfcol):
         cudfcol.apply_boolean_mask(non_null_idxs),
     )
 
-    if dtype[0] != _DtypeKind.BOOL:
-        array_from_dlpack = cp.from_dlpack(buf.__dlpack__()).get()
-        col_array = cp.asarray(cudfcol.data_array_view(mode="read")).get()
-        assert_eq(
-            array_from_dlpack[non_null_idxs.to_numpy()].flatten(),
-            col_array[non_null_idxs.to_numpy()].flatten(),
-        )
-    else:
-        pytest.raises(TypeError, buf.__dlpack__)
+    array_from_dlpack = cp.from_dlpack(buf.__dlpack__()).get()
+    col_array = cp.asarray(cudfcol.data_array_view(mode="read")).get()
+    assert_eq(
+        array_from_dlpack[non_null_idxs.to_numpy()].flatten(),
+        col_array[non_null_idxs.to_numpy()].flatten(),
+    )
 
 
 def assert_column_equal(col: _CuDFColumn, cudfcol):


python/cudf/cudf/tests/test_df_protocol.py
```diff
@@ -39,15 +39,12 @@ def assert_buffer_equal(buffer_and_dtype: Tuple[_CuDFBuffer, Any], cudfcol):
...
-            col_array[non_null_idxs.to_numpy()].flatten(),
-        )
-    else:
-        pytest.raises(TypeError, buf.__dlpack__)
+    array_from_dlpack = cp.from_dlpack(buf.__dlpack__()).get()
+    col_array = cp.asarray(cudfcol.data_array_view(mode="read")).get()
+    assert_eq(
+        array_from_dlpack[non_null_idxs.to_numpy()].flatten(),
+        col_array[non_null_idxs.to_numpy()].flatten(),
+    )
```
wence-, line 47: Does this pass need to be conditioned on a cupy version?

vyasr: Sure we can. We have not historically managed supported versions for testing consistently for any dependency aside from pandas AFAIK, but cupy is generally stable enough that maybe the few places where our tests are already conditional on cupy version are sufficient to capture all the cases we've cared about so far.

vyasr:
> Do we need a certain version in our dependencies/conda recipes? What is the minimum version needed?
> 
> https://github.com/rapidsai/cudf/blob/45a9e825f975ca55f57f5af9523ec0b88839ff56/dependencies.yaml#L252

I don't know how well cudf will work with cupy < 11. I haven't tried that in a long time. I can try installing 9.5 and see what happens.

python/cudf/cudf/tests/test_df_protocol.py
bdice: Are we able to just set a minimum cupy version in our dependencies that meets this criterion, rather than introduce conditional logic? https://github.com/rapidsai/cudf/pull/12883#issuecomment-1456397914

wence-: Just thinking that unless (as @bdice suggests) we bump the minimum version, any environment that was previously passing will now fail these tests.

bdice:
Great. Let’s get that bumped, we probably shouldn’t support 3 major versions of cupy, especially with cupy 12 just around the corner.

bdice:
I didn't realize CI was blocked by this. If that is the case, I'll approve and we can merge. However, I don't think we should continue extending support to such old cupy versions. I would bump the minimum version to one that is compatible with our tests, not just the package itself.

bdice (approved):
Approving this to unblock CI (changes look good) but we should revisit version support for cupy in a follow-up PR so we aren't supporting 3 major versions of cupy. Supported versions should pass all tests and we should avoid taking on debt in the form of version-conditional logic.

wence- (approved):

wence-:
Thanks for the digging, I agree with @bdice's suggestions.

vyasr:
Created #12884 to continue the discussion @bdice @wence-

PR #12887, opened 2023-03-06
[REVIEW] Update minimum `pandas` and `numpy` pinnings

### LINKED ISSUE #12785: Bump minimum pandas/numpy versions
To align with dask (https://github.com/dask/dask/pull/9950), we should bump cudf's minimum requirements to `pandas>=1.3.0` and `numpy>=1.21` and delete any outdated testing code for prior versions of pandas.

See also: https://github.com/rapidsai/cudf/issues/12398

galipremsagar:
## Description

This PR:

- [x] Increments the minimum pinning for `pandas` version from `1.0` to `1.3`.
- [x] Sets a minimum pinning for `numpy` as `>=1.21`
- [x] Fixes arm conda environment creation by removing `pandoc` version constraint.

Resolves #12785.

## Checklist


bdice:
Can we also bump numpy (currently it's unpinned) to align with dask and resolve this issue? https://github.com/rapidsai/cudf/issues/12785

galipremsagar:
> Can we also bump numpy (currently it's unpinned) to align with dask and resolve this issue? #12785

Done 👍

python/cudf/cudf/tests/test_rolling.py
```diff
@@ -310,10 +300,8 @@ def test_rolling_getitem():
 
 
 def test_rolling_getitem_window():
-    if PANDAS_GE_110:
-        kwargs = {"check_freq": False}
-    else:
-        kwargs = {}
+    kwargs = {"check_freq": False}
```
bdice, line 303: Let's explicitly specify these in the function instead of writing `**kwargs`. It's specific to cudf and isn't used in the pandas call. This applies in a couple places.

python/cudf/cudf/tests/test_reshape.py
```diff
@@ -89,12 +88,7 @@ def test_melt(nulls, num_id_vars, num_value_vars, num_rows, dtype):
     "dtype",
     list(NUMERIC_TYPES + DATETIME_TYPES)
```
bdice, line 89: This is an awkward way to write this. Try `[*NUMERIC_TYPES, *DATETIME_TYPES, "str"]`.

python/cudf/cudf/core/_compat.py
```diff
@@ -4,9 +4,6 @@
 from packaging import version
 
 PANDAS_VERSION = version.parse(pd.__version__)
-PANDAS_GE_110 = PANDAS_VERSION >= version.parse("1.1")
-PANDAS_GE_120 = PANDAS_VERSION >= version.parse("1.2")
-PANDAS_LE_122 = PANDAS_VERSION <= version.parse("1.2.2")
 PANDAS_GE_130 = PANDAS_VERSION >= version.parse("1.3.0")
```
bdice, line 7: `PANDAS_GE_130` is always true, right?

python/cudf/cudf/tests/test_reshape.py
wence-: Or `list(chain(NUMERIC_TYPES, DATETIME_TYPES, ["str"]))` if you don't like the splats :)

python/cudf/cudf/tests/test_rolling.py
wence-: Agreed, these were previously `kwargs` to avoid polluting the `assert_eq` with pandas version conditions, but now those are gone we can just inline things.

python/cudf/cudf/core/_compat.py
wence-: Yes (so there might be more conditions to cull!)

wence- (commented):
Minor issues.

python/cudf/cudf/core/_compat.py
galipremsagar: Yup, good catch. Cleaned it up

python/cudf/cudf/tests/test_reshape.py
galipremsagar: used the `chain` method, let me know if there is a strong preference if `chain` isn't a preference.

python/cudf/cudf/tests/test_rolling.py
galipremsagar: Done

conda/recipes/cudf/meta.yaml
```diff
@@ -60,10 +60,10 @@ requirements:
     - protobuf >=4.21.6,<4.22
     - python
     - typing_extensions
-    - pandas >=1.0,<1.6.0dev0
+    - pandas >=1.3,<1.6.0dev0
     - cupy >=9.5.0,<12.0.0a0
     - numba >=0.56.2
-    - numpy
+    - numpy >=1.21
```
bdice, line 66: Current versions of numba [require `numpy<1.24`](https://github.com/numba/numba/issues/8615#issuecomment-1386886044) (until numba 0.57 is released with [this fix](https://github.com/numba/numba/pull/8691)). We're getting this pinning [from the upstream feedstock](https://github.com/conda-forge/numba-feedstock/blob/9662fd9e630719122c598edde702d96fa328bf94/recipe/meta.yaml#L67) but this pinning doesn't work if the user installs numba another way.

@jayrodge ran into this locally when installing RAPIDS 23.02 on WSL, perhaps with a numba from some source other than `conda-forge`.

Should we actively prevent that problem by pinning like this until numba 0.57 is released?
```suggestion
    - numpy >=1.21,<1.24
```

galipremsagar: Even if we pin `numpy` in our feedstock, if the user install `numba` or as a matter of fact `numpy` in another we wouldn't be able to prevent these kinds of breakages no? I'm only hesitant on having an upper bound on `numpy` because we don't use it's internal APIs and this will have a maintenance overhead of tracking numpy releases and unpinning them manually. Whereas leaving it unpinned on the upper bound could help us detect breakages in our CI and have it fixed ASAP.

bdice: I understand the hesitation. I just wanted to pose the idea. Let’s leave it alone.

wence-:
I think this is just waiting to resolve the numpy upper bound pin for numba < 0.57, but is otherwise good to go?

bdice (approved):

wence- (approved):
Thanks!

PR #12888, opened 2023-03-06
Remove default parameters from detail headers in include

vyasr:
## Description

Contributes to #9854. None of these changes should affect users, nor do they impose a particularly onerous burden on libcudf developers (just some extra passing through `mr` or `cudf::get_default_stream()`.

## Checklist


cpp/src/copying/scatter.cu
```diff
@@ -393,7 +393,7 @@ std::unique_ptr<column> boolean_mask_scatter(column_view const& input,
 
   // The scatter map is actually a table with only one column, which is scatter map.
   auto scatter_map =
-    detail::apply_boolean_mask(table_view{{indices->view()}}, boolean_mask, stream);
+    detail::apply_boolean_mask(table_view{{indices->view()}}, boolean_mask, stream, mr);
```
davidwendt, line 396:
```suggestion
    detail::apply_boolean_mask(table_view{{indices->view()}}, boolean_mask, stream, rmm::mr::get_current_device_resource());
```
This appears to be a temporary and so should not use the passed in `mr`

cpp/src/sort/segmented_sort_impl.cuh
```diff
@@ -68,12 +68,13 @@ struct column_fast_sort_fn {
...
                  bool ascending,
-                 rmm::cuda_stream_view stream)
+                 rmm::cuda_stream_view stream,
+                 rmm::mr::device_memory_resource* mr)
   {
     // CUB's segmented sort functions cannot accept iterators.
     // We create a temporary column here for it to use.
     auto temp_col =
-      cudf::detail::allocate_like(input, input.size(), mask_allocation_policy::NEVER, stream);
+      cudf::detail::allocate_like(input, input.size(), mask_allocation_policy::NEVER, stream, mr);
```
davidwendt, line 77: This temporary should not be created with the passed in `mr`
https://github.com/rapidsai/cudf/blob/branch-23.04/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#temporary-memory

cpp/src/stream_compaction/distinct.cu
```diff
@@ -146,7 +146,7 @@ std::unique_ptr<table> distinct(table_view const& input,
   }
 
   auto const gather_map =
-    get_distinct_indices(input.select(keys), keep, nulls_equal, nans_equal, stream);
+    get_distinct_indices(input.select(keys), keep, nulls_equal, nans_equal, stream, mr);
```
davidwendt, line 149: The `gather_map` is not returned to the caller and so should use `rmm::mr::get_current_device_resource()` instead of `mr`

cpp/src/stream_compaction/stable_distinct.cu
```diff
@@ -39,7 +39,7 @@ std::unique_ptr<table> stable_distinct(table_view const& input,
   }
 
   auto const distinct_indices =
-    get_distinct_indices(input.select(keys), keep, nulls_equal, nans_equal, stream);
+    get_distinct_indices(input.select(keys), keep, nulls_equal, nans_equal, stream, mr);
```
davidwendt, line 42:
```suggestion
    get_distinct_indices(input.select(keys), keep, nulls_equal, nans_equal, stream, rmm::mr::get_current_device_resource());
```

davidwendt (changes requested):
Looks good. 
Some issues with using `mr` variable for temporary objects.

cpp/src/copying/scatter.cu
vyasr: Ah thanks I wasn't careful enough in checking on these. Will fix.

davidwendt (changes requested):

cpp/src/sort/segmented_sort_impl.cuh
```diff
@@ -68,12 +69,16 @@ struct column_fast_sort_fn {
                  column_view const& segment_offsets,
                  mutable_column_view& indices,
                  bool ascending,
-                 rmm::cuda_stream_view stream)
+                 rmm::cuda_stream_view stream,
+                 rmm::mr::device_memory_resource* mr)
```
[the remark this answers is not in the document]
vyasr: Ah good call I removed its usage for temporary allocations but didn't realize I'd gotten rid of all uses.

cpp/src/stream_compaction/distinct_count.cu
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include "cudf/utilities/default_stream.hpp"
```
davidwendt, line 17:
```suggestion
#include <cudf/utilities/default_stream.hpp>
```
This should also be moved below the other 2 includes.

cpp/src/stream_compaction/unique_count.cu
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include "cudf/utilities/default_stream.hpp"
```
davidwendt, line 17:
```suggestion
#include <cudf/utilities/default_stream.hpp>
```
Same here

cpp/src/sort/segmented_sort_impl.cuh
```diff
@@ -24,6 +24,7 @@
 #include <rmm/device_uvector.hpp>
 
 #include <cub/device/device_segmented_sort.cuh>
+#include <rmm/mr/device/per_device_resource.hpp>
```
davidwendt, line 27: Move this include up with the other `include <rmm` line

davidwendt (changes requested):
Looks good. Just some `#include` fixes.

cpp/src/stream_compaction/unique_count.cu
vyasr: Did we add a pre-commit hook or something that added these? I don't even recall manually adding them... @bdice does this sound familiar to you?

bdice: Not familiar. IDEs sometimes add headers automatically?

ttnghia (approved):

cpp/src/stream_compaction/distinct.cu
```diff
@@ -25,6 +25,7 @@
 #include <cudf/table/table_view.hpp>
 #include <cudf/types.hpp>
 
+#include <rmm/mr/device/per_device_resource.hpp>
```
davidwendt, line 28:
```suggestion
#include <rmm/mr/device/per_device_resource.hpp>

```
(adds an extra blank line between the rmm and thrust include sections)

cpp/tests/copying/detail_gather_tests.cu
```diff
@@ -57,8 +59,7 @@ TYPED_TEST(GatherTest, GatherDetailDeviceVectorTest)
 
   // test with device vector iterators
   {
-    std::unique_ptr<cudf::table> result =
-      cudf::detail::gather(source_table, gather_map.begin(), gather_map.end());
+    std::unique_ptr<cudf::table> result = cudf::gather(source_table, gather_map->view());
```
davidwendt, line 62: This is an interesting change. 
The intention of this source file seems to be wanting to test `cudf::detail::gather` functions -- specifically the iterator interfaces as per this comment from above:
https://github.com/rapidsai/cudf/pull/12888/files#diff-88fa573b86cc7871ca7ce24babaa1651cb41a2c76de8fdab1925c21dae450fc4R44-R45
Perhaps these tests are really not needed and this source file itself is not necessary at all.

vyasr: Hmm yeah that is a good point. This file had two invocations of `detail::gather` and it looks like the other one was properly passing the stream through. I think the minimal change for this PR would be to just pass the stream, then to consider removing in a separate PR, WDYT? I made the change that I did thinking that it would be best to move away from using detail APIs in test altogether, but I also want to limit scope creep.

davidwendt: We can evaluate removing these tests in a separate PR.
But then I think you should change this back to calling the `detail` API right?

vyasr: Yup was waiting for your approval on the plan before pushing the changes. I'll update it now.

cpp/benchmarks/common/generate_input.cu
```diff
@@ -32,6 +32,7 @@
 #include <rmm/device_buffer.hpp>
 #include <rmm/device_uvector.hpp>
 
+#include <rmm/mr/device/per_device_resource.hpp>
```
davidwendt, line 35:
```suggestion
#include <rmm/device_uvector.hpp>
#include <rmm/mr/device/per_device_resource.hpp>

```

cpp/src/rolling/detail/lead_lag_nested.cuh
```diff
@@ -27,6 +27,7 @@
 
 #include <rmm/exec_policy.hpp>
 
+#include <rmm/mr/device/per_device_resource.hpp>
```
davidwendt, line 30:
```suggestion
#include <rmm/exec_policy.hpp>
#include <rmm/mr/device/per_device_resource.hpp>

```

cpp/tests/copying/gather_str_tests.cpp
```diff
@@ -24,6 +24,7 @@
 #include <cudf/detail/gather.hpp>
 #include <cudf/table/table.hpp>
 #include <cudf/table/table_view.hpp>
+#include <rmm/mr/device/per_device_resource.hpp>
```
davidwendt, line 27:
```suggestion

#include <rmm/mr/device/per_device_resource.hpp>
```

davidwendt (approved):

Issue #12349, opened 2022-12-09
[FEA] Support `stream` and `mr` parameters for `structs::detail::flatten_nested_columns`


ttnghia:
In `structs::detail::flatten_nested_columns`, the default stream and default device resource are always used for computing the internal data. This should be corrected to explicitly use the current `stream` and `mr` values passed in by the caller function.

GregoryKimball:
I'm surprised that we didn't catch this in our stream sanitization tests. Do we have adequate unit tests for `structs::detail::flatten_nested_columns`?

ttnghia:
Yes we have a lot of tests for that API. But as you know, most unit tests never specify the `stream` or `mr` parameters.

PR #12007, opened 2022-10-26
Remove default parameters for nvtext::detail functions

davidwendt:
## Description
Removes default parameters from the `nvtext::detail` functions. Most of these were internal default parameters which were unnecessary. The nvtext detail functions are only used within nvtext APIs.

Reference #11967

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12006, opened 2022-10-26
Remove default parameters for cudf::dictionary::detail functions

davidwendt:
## Description
Removes default parameters from the `cudf::dictionary::detail` functions. Most of these were allowing for the default memory-resource which is unnecessary. One non-stream, non-mr parameter was defaulted but the default was never used.

Reference #11967

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12892, opened 2023-03-07
Add `stream` and `mr` parameters for `structs::detail::flatten_nested_columns`

ttnghia:
The internal API  `structs::detail::flatten_nested_columns` currently always use default stream and device resources. This PR adds `stream` and `mr` parameters for it, allowing to pass in the current `stream` and `mr` values.

Closes https://github.com/rapidsai/cudf/issues/12349.


cpp/include/cudf/detail/structs/utilities.hpp
```diff
@@ -164,14 +164,18 @@ class flattened_table {
...
  * alongside the supporting columns and device_buffers for the flattened table.
  */
 [[nodiscard]] flattened_table flatten_nested_columns(
   table_view const& input,
   std::vector<order> const& column_order,
   std::vector<null_order> const& null_precedence,
-  column_nullability nullability = column_nullability::MATCH_INCOMING);
+  column_nullability nullability,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
```
divyegala, line 178: Shouldn't we have no defaults in detail APIs, or is `mr` an exception?

ttnghia: I thought about that too. Currently there is no guideline for this. I see that in https://github.com/rapidsai/cudf/pull/12888 @vyasr removed all the default value for `mr` although the original issue of that PR was asking for just removing default value for `stream`. So I think removing default value for this is still optional.

PointKernel: Just FYI, @davidwendt created a series of PR removing default parameters for detail functions (e.g. #12003, #12006, #12007).

ttnghia: These sound good. So I'm going to remove the default parameter completely in this related file.

ttnghia: Oop, removing default `mr` value for the entire file would touch more places that are out of scope of this PR. I would rather do that in a separate PR.

vyasr: The relevant discussion on default parameters happened in #9584 and #11967. The tl;dr is that we want to remove default stream parameters in detail APIs, but there were concerns that this could lead to awkward parameter orderings (e.g. signatures like `..., stream, other_enum=enum_default, mr=rmm::mr::get_current_device_resource())`) if there are other default parameters. Since we don't know how many such APIs exist, the plan is to try to remove all default parameters and see how hard that is. In theory we could leave `mr` defaults since that parameter is always last, but the same arguments apply for `mr` as for `stream`: leaving that parameter defaulted leaves open room for developer error in forgetting to pass a user-provided `mr` from a public API to a detail one.

There is no official guideline yet because I am just now getting around to removing as many default parameters from detail APIs as is possible in PRs like #12888. If it turns out to be pretty painless then we should expect to update the dev docs with guidance to avoid default parameters altogether in detail APIs. As such, I'd go ahead and drop the mr default here. However, I agree with @ttnghia that the scope of this PR should be limited: don't add new default parameters, but also don't bother removing any that aren't directly related to your other changes.

vyasr (approved):
This LGTM pending the removal of the default mr from the `flatten_nested_columns` signature.

cpp/include/cudf/detail/structs/utilities.hpp
PointKernel: > don't add new default parameters, but also don't bother removing any that aren't directly related to your other changes.

:+1:

cpp/src/structs/utilities.cpp
```diff
@@ -98,11 +98,20 @@ struct table_flattener {
   std::vector<null_order> flat_null_precedence;
   column_nullability nullability;
```
bdice, line 100: No newline here. These go with the other struct elements.
```suggestion
```

cpp/src/structs/utilities.cpp
```diff
@@ -98,11 +98,20 @@ struct table_flattener {
...
-                  column_nullability nullability)
-    : column_order(column_order), null_precedence(null_precedence), nullability(nullability)
+                  column_nullability nullability,
+                  rmm::cuda_stream_view stream,
+                  rmm::mr::device_memory_resource* mr)
+    : column_order(column_order),
+      null_precedence(null_precedence),
+      nullability(nullability),
+      stream(stream),
+      mr(mr)
```
bdice, line 114: Since we are already touching these lines: prefer `{}` initializer syntax. It does not allow narrowing conversions and avoids ambiguity.
[ES.23: Prefer the {}-initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax
).
Also: [What are the advantages of list initialization (using curly braces)?](https://stackoverflow.com/questions/18222926/what-are-the-advantages-of-list-initialization-using-curly-braces)

```suggestion
    : column_order{column_order},
      null_precedence{null_precedence},
      nullability{nullability},
      stream{stream},
      mr{mr}
```

bdice (commented):
Two minor suggestions, otherwise LGTM!

cpp/src/structs/utilities.cpp
ttnghia: Actually the member variables are already grouped:
```
table_view input;
  // reference variables
  std::vector<order> const& column_order;
  std::vector<null_order> const& null_precedence;
  // output
  std::vector<std::unique_ptr<column>> validity_as_column;
  temporary_nullable_data nullable_data;
  std::vector<column_view> flat_columns;
  std::vector<order> flat_column_order;
  std::vector<null_order> flat_null_precedence;
  column_nullability nullability;
```

So if removing that empty line, `stream` and `mr` will be grouped with `// output` so I would prefer to keep it here.

bdice: Sorry, I didn’t read closely enough to realize those were outputs. What I should have said is that these should be grouped with the other constructor arguments, in initialization order. Let’s make that change: put stream and mr before the outputs and after the other constructor arguments.

ttnghia: That sound... a bit awkward to me, as `stream` and `mr` typically stay last in the parameter list :|

ttnghia: How about this?
```suggestion
// stream and memory resource
```

bdice: My point is that `flat_columns` and similar are not constructor arguments. Group stream and mr with the other constructor arguments here.

ttnghia: Or sorry I misunderstood that. Update pushed.

cpp/include/cudf/detail/unary.hpp
```diff
@@ -74,6 +74,16 @@ std::unique_ptr<cudf::column> unary_operation(
...
 
+/**
+ * @copydoc cudf::is_valid
+ *
+ * @param stream CUDA stream used for device memory operations and kernel launches.
+ */
+std::unique_ptr<cudf::column> is_valid(
+  cudf::column_view const& input,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
```
divyegala, line 85: Since this is a new API being introduced by this PR, I think we can drop the defaults going by @vyasr's comment above, right?

ttnghia: This is not new. It is just brought from the non-detail header. I want to keep consistency between it with all other detail APIs in the same file (which still have default value for `mr`).

hyperbolic2346 (approved):

divyegala (approved):

cpp/include/cudf/detail/unary.hpp
vyasr: That's fine I can do batch removals later.

PR #12900, opened 2023-03-07
Rework some code logic to reduce iterator and comparator inlining to improve compile time

davidwendt:
## Description
Disables inlining the device code logic for the row operators for nested column types did not work as hoped.
Some files took longer to compile and some functions ran 20% slower for large rows.

Reworking individual source files to break up the code logic into multiple kernels seems to work well for compile time while having a smaller effect on performance. The goal is to only rework the nested column code paths.
Here are some source files that have compile time issues and are improved in this PR.

| source file  | current | PR |
|:--- | ---:| ---:|
|   stream_compaction/unique_count.cu | 18 min |  13 min |
|   groupby/sort/group_nunique.cu   | 16 min |  2 min  |
|   stream_compaction/unique.cu  | 16 min | 5 min |
|   groupby/sort/sort_helper.cu | 10 min | 6.5 min |
|   search/contains_scalar.cu | 12 min | 4.7 min  |
|   sort/is_sorted.cu | 9 min | 7 min |
|   groupby/sort/group_std.cu | 7 min | 1.2 min |
|   groupby/sort/group_m2.cu | 6 min | 1.2 min  |

Available benchmarks showed minimal impact to performance.

## Checklist


cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -381,7 +381,7 @@ class device_row_comparator {
 
     template <typename Element,
               CUDF_ENABLE_IF(has_nested_columns and std::is_same_v<Element, cudf::struct_view>)>
-    __device__ cuda::std::pair<weak_ordering, int> operator()(
+    __attribute__((noinline)) __device__ cuda::std::pair<weak_ordering, int> operator()(
```
bdice, line 384: I think we typically write this as `__noinline__`. Internally, I believe CUDA expands it to the `__attribute__` form. https://docs.nvidia.com/cuda/cuda-c-programming-guide/#noinline-and-forceinline

PointKernel:
Is it fair to say that we should **avoid** the below functions to reduce build time?

- `cudf::detail::make_counting_transform_iterator`
- `thrust::make_transform_iterator`
- `thrust::unique_copy`
- `thrust::count_if`
- `thrust::is_sorted`

davidwendt:
> Is it fair to say that we should **avoid** the below functions to reduce build time?
> 
> * `cudf::detail::make_counting_transform_iterator`
> * `thrust::make_transform_iterator`
> * `thrust::unique_copy`
> * `thrust::count_if`
> * `thrust::is_sorted`

Not in general. The transform iterators are certainly not the issue. The `copy_if`, and `is_sorted` all end up calling some version of cub:reduce. If the iterators/comparators they are given are more than a few lines of logic, the code bloat can be extreme due to large amount of inlining of the inputs throughout the generated kernel. And `unique_copy` has similar issues with inlining the comparator.
I will try to work on a more formal document outlining what to avoid and how to get around this problem.

cpp/src/sort/is_sorted.cu
```diff
@@ -44,10 +46,16 @@ auto is_sorted(cudf::table_view const& in,
...
     auto const device_comparator = comparator.less<true>(has_nested_nulls(in));
 
-    return thrust::is_sorted(rmm::exec_policy(stream),
-                             thrust::make_counting_iterator(0),
-                             thrust::make_counting_iterator(in.num_rows()),
-                             device_comparator);
+    auto d_results = rmm::device_uvector<bool>(in.num_rows(), stream);
+    thrust::transform(rmm::exec_policy(stream),
+                      thrust::make_counting_iterator<size_type>(0),
+                      thrust::make_counting_iterator<size_type>(in.num_rows()),
```
PointKernel, line 52:
```suggestion
                      thrust::counting_iterator<size_type>(0),
                      thrust::counting_iterator<size_type>(in.num_rows()),
```

PointKernel (approved):
LGTM

Looking forward to seeing the build time guideline doc.

cpp/src/groupby/sort/group_m2.cu
```diff
@@ -62,15 +64,18 @@ void compute_m2_fn(column_device_view const& values,
                    ResultType* d_result,
                    rmm::cuda_stream_view stream)
 {
-  auto const var_iter = cudf::detail::make_counting_transform_iterator(
-    size_type{0},
-    m2_transform<ResultType, decltype(values_iter)>{
-      values, values_iter, d_means, group_labels.data()});
+  auto m2_fn = m2_transform<ResultType, decltype(values_iter)>{
+    values, values_iter, d_means, group_labels.data()};
+  auto const itr = thrust::make_counting_iterator<size_type>(0);
```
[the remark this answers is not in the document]
davidwendt: I thought the factory iterator was recommended. 
I do prefer specifying the type as you suggest here.
That is, I prefer specifying 0 as `size_type` rather than the compiler automatically identifying 0 as `int` like it would with the factory without the type.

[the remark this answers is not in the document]
PointKernel: >  I thought the factory iterator was recommended.

Yeah, I learned this from Jake's review a while ago.

> I prefer specifying 0 as size_type rather than the compiler automatically identifying 0 as int

Sounds good to me :+1:

cpp/src/stream_compaction/unique.cu
```diff
@@ -67,38 +67,46 @@ std::unique_ptr<table> unique(table_view const& input,
...
-  auto const comparator_helper = [&](auto const row_equal) {
-    // get indices of unique rows
-    auto result_end = unique_copy(thrust::counting_iterator<size_type>(0),
-                                  thrust::counting_iterator<size_type>(num_rows),
-                                  mutable_view->begin<size_type>(),
-                                  row_equal,
-                                  keep,
-                                  stream);
+  size_type const unique_size = [&] {
+    if (cudf::detail::has_nested_columns(keys_view)) {
```
bdice, line 71: If this compiles faster for nested types, does it also compile faster for non-nested types? If it's possible to unify these and have a single implementation of the algorithms, I would prefer that (rather than one `transform + copy_if` for nested types and one `unique_copy` for non-nested types).

If there are considerations like runtime, memory usage, etc. that warrant two separate implementations, then let's inform the reader with some comments explaining this decision.

cpp/src/stream_compaction/unique_count.cu
```diff
@@ -76,11 +77,16 @@ cudf::size_type unique_count(table_view const& keys,
...
     auto const comp =
       row_comp.equal_to<true>(nullate::DYNAMIC{has_nested_nulls(keys)}, nulls_equal);
-    return thrust::count_if(
-      rmm::exec_policy(stream),
-      thrust::counting_iterator<cudf::size_type>(0),
-      thrust::counting_iterator<cudf::size_type>(keys.num_rows()),
-      [comp] __device__(cudf::size_type i) { return (i == 0 or not comp(i, i - 1)); });
+
+    auto d_results = rmm::device_uvector<bool>(keys.num_rows(), stream);
+    thrust::transform(rmm::exec_policy(stream),
```
bdice, line 82: Comments explaining why we separate algorithms would be nice.
```suggestion
    // Use transform and count instead of count_if to improve compile time.
    thrust::transform(rmm::exec_policy(stream),
```

cpp/src/sort/is_sorted.cu
```diff
@@ -44,16 +46,22 @@ auto is_sorted(cudf::table_view const& in,
   if (cudf::detail::has_nested_columns(in)) {
     auto const device_comparator = comparator.less<true>(has_nested_nulls(in));
 
-    return thrust::is_sorted(rmm::exec_policy(stream),
-                             thrust::make_counting_iterator(0),
-                             thrust::make_counting_iterator(in.num_rows()),
-                             device_comparator);
+    auto d_results = rmm::device_uvector<bool>(in.num_rows(), stream);
+    thrust::transform(rmm::exec_policy(stream),
```
bdice, line 50:
```suggestion
    // Use transform and count instead of thrust's is_sorted to improve compile time.
    thrust::transform(rmm::exec_policy(stream),
```

bdice (commented):
This is fantastic. These optimizations look correct and seem sensible, when considering the kernel complexity that we are now able to split up. I'd like a few small comments like the ones I suggested here, to indicate that our choices of algorithms are informed by compile time. That will clear things up for the reader, and help prevent "clever" refactors down the line.

hyperbolic2346 (approved):

bdice (approved):
Love the comments. Thanks for that -- it helps a lot for future readers, and makes us more aware of the process of making compile time improvements.

ttnghia (approved):
I love this.

PR #12902, opened 2023-03-08
Add additional mock env vars for `moto`

ajschmidt8:
## Description

This PR adds some additional AWS related environment variable mocks to the `cudf` `pytest`s.

These environment variables are documented by `moto` here: https://docs.getmoto.org/en/latest/docs/getting_started.html#recommended-usage.

This is important because of some recent CI changes that we made which now populate the `AWS_SESSION_TOKEN` environment variable, which was not mocked prior to the changes in this PR.

## Checklist

### DIFF
--- a/python/cudf/cudf/tests/test_s3.py
+++ b/python/cudf/cudf/tests/test_s3.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
 
 import os
 import socket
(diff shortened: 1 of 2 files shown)


galipremsagar (approved):

python/cudf/cudf/tests/test_s3.py
```diff
@@ -62,6 +62,9 @@ def s3_base(endpoint_ip, endpoint_port):
         os.environ.setdefault("AWS_ACCESS_KEY_ID", "foobar_key")
```
bdice, line 62: Should these be “setdefault” or should they be overriding the existing environment variable if present?

bdice: The moto examples are overriding the environment explicitly. https://docs.getmoto.org/en/latest/docs/getting_started.html#recommended-usage

galipremsagar: Since this would be just for pytest session scope, I think it's _okay_ to just mock them here..

bdice: What I’m saying is that we shouldn’t ever send real credentials from the environment to the mocked S3 backend that moto is providing. Is that accurate?

ajschmidt8: I tend to agree with this, but I don't know the history of these tests. I wasn't sure if `setdefault` was used for a particular reason.

@bdice, should I push an update to match their docs?

bdice: Yes please. If tests pass, I think this is a better choice.

ajschmidt8: Pushed in f602511.

bdice (approved):

raydouglass (approved):

PR #12904, opened 2023-03-08
Update libcudf counting functions to specify cudf::size_type

davidwendt:
## Description
Adds section to developer guide about `cudf::size_type` and adds links to it from other relevant parts of the document.
The fundamental nature of this type seems important enough to mention in the developer guide since it is the basis for how much of the code is designed and implemented.
Also updates some doxygen for public APIs that are return `size_type` column values but had cited `INT32` specifically.

Reference: https://github.com/rapidsai/cudf/pull/12779#discussion_r1117315749

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12908, opened 2023-03-08
Remove return type from @return doxygen tags

davidwendt:
## Description
The doxygen `@return` does not require a type but only the description. However, the doxygen output looks cleaner without the type. There were only a few places in the public header files that needed to be corrected.
This technically a documentation-only change.
Found when working on #12904 

## Checklist


karthikeyann (approved):
Thanks for doing this. 
LGTM 👍

bdice (approved):
Thanks for this!

PR #12820, opened 2023-02-22
[REVIEW] Deprecate `Index.is_*` methods 

galipremsagar:
## Description

This PR deprecates the following Index APIs to match with pandas 2.0:

- [x] is_numeric
- [x] is_boolean
- [x] is_integer
- [x] is_floating
- [x] is_object
- [x] is_categorical
- [x] is_interval

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #12909, opened 2023-03-08
[REVIEW] Add information about `Index.is_*` method deprecation

galipremsagar:
## Description
This PR adds additional information for the following Index APIs to match with pandas 2.0:

- [x]  is_numeric
- [x]  is_boolean
- [x]  is_integer
- [x]  is_floating
- [x]  is_object
- [x]  is_categorical
- [x]  is_interval 

## Checklist


vyasr (changes requested):
Still need to clean out the implementations in index.py, right?

galipremsagar:
> Still need to clean out the implementations in index.py, right?

The implementations were converted as internal APIs to be used as fast paths for `is_numeric_dtype`, `is_bool_dtype`, etc: here: https://github.com/rapidsai/cudf/pull/12820/files#diff-ef94b3e28724ebfe959cd02a9f37427cefcdc118799490453fbe1c6299dc0324R63

vyasr:
Ah OK I see. Now that these APIs are not public I think it would be better to remove those helper functions and implement the `api.types` functions in terms of a simple hash lookup (i.e. store some module-level dicts in api/types.py mapping `{cudf.$INDEX_CLASS: bool}`). If these APIs are no longer public I don't think the organization of having helper functions like this is very helpful, it just makes the code less readable and harder to maintain (and incurs unnecessary performance costs too with Python function calls). If you'd like to refactor that in a separate PR I'd be OK with that though.

galipremsagar:
> Ah OK I see. Now that these APIs are not public I think it would be better to remove those helper functions and implement the `api.types` functions in terms of a simple hash lookup (i.e. store some module-level dicts in api/types.py mapping `{cudf.$INDEX_CLASS: bool}`). If these APIs are no longer public I don't think the organization of having helper functions like this is very helpful, it just makes the code less readable and harder to maintain (and incurs unnecessary performance costs too with Python function calls). If you'd like to refactor that in a separate PR I'd be OK with that though.

Got it, it makes sense to have this kind of a refactor done in this PR itself. I'll make those changes here. 👍

mroeschke:
FWIW these were only deprecated in pandas 2.0 not removed. These will be removed in pandas 3.0

galipremsagar:
> FWIW these were only deprecated in pandas 2.0 not removed. These will be removed in pandas 3.0

Ah, my bad. So let's probably keep them deprecated and not drop them now. Will add comments to this PR about the delayed removal of these APIs.

galipremsagar:
This should be now ready for review.

mroeschke (approved):

shwina (approved):

PR #12921, opened 2023-03-10
Mark dlpack tensor deleter as noexcept to match PyCapsule_Destructor signature.

bdice:
## Description
This is a fix for a Cython problem that appears when building with Cython 3.0.0 beta 1. There are a bunch of warnings I'd like to fix as well, but this is the only hard error. The fix is to use `noexcept` in the deleter definition.

```
Error compiling Cython file:
...
        )
    return pycapsule.PyCapsule_New(
        dlpack_tensor,
        'dltensor',
        dlmanaged_tensor_pycapsule_deleter
        ^
/home/bdice/code/cudf/python/cudf/cudf/_lib/interop.pyx:69:8: Cannot assign type 'void (object) except *' to 'PyCapsule_Destructor'
```

## Checklist


python/cudf/cudf/_lib/interop.pyx
```diff
@@ -66,7 +66,7 @@ def to_dlpack(list source_columns):
     return pycapsule.PyCapsule_New(
         dlpack_tensor,
         'dltensor',
-        dlmanaged_tensor_pycapsule_deleter
+        <pycapsule.PyCapsule_Destructor>dlmanaged_tensor_pycapsule_deleter
```
bdice, line 69: Note that we're following the advice in [this StackOverflow post](https://stackoverflow.com/questions/59708440/cython-declare-a-pycapsule-destructor-in-pyx-file) about how to define the deleter, but apparently Cython 3.0.0 is less lax (or has a bug?) about how it accepts this and does require an explicit cast.

shwina: Rather than an explicit cast, the fix is to define the `dlmanaged_tensor_pycapsule_deleter` function below with `noexcept`.

Function types must match up to their exception value. In Cython 3.0, non-extern functions (here, `dlmanaged_tensor_pycapsule_deleter) are implicitly declared with an exception value, and extern functions (here, `pycapsule.PyCapsule_Destructor`) are implicitly declared `noexcept`.

https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html?highlight=noexcept#error-return-values

bdice: @shwina Amazing, thank you. I’ll try that fix!

bdice: @shwina This seemed to work. I'll resolve this conversation and update the PR title/description.

python/cudf/cudf/_lib/interop.pyx
```diff
@@ -70,7 +70,7 @@ def to_dlpack(list source_columns):
     )
 
 
-cdef void dlmanaged_tensor_pycapsule_deleter(object pycap_obj):
+cdef void dlmanaged_tensor_pycapsule_deleter(object pycap_obj) noexcept:
```
wence-, line 73: I agree this fixes the issue, but is it a lie? On line 80, we call `pycapsule.PyCapsule_GetPointer(pycap_obj, "dltensor")`. This _feels_ like it could raise an exception, certainly I am guessing that the try/except around getting `used_dltensor` is because that attribute might not exist.

If I were writing this in C, I would probably do something like:
```c

dlpack_tensor = PyCapsule_GetPointer(obj, "used_dltensor");
if (dlpack_tensor) {
    // Already used, we're done
    return;
else {
    // name didn't exist, let's pretend nothing happened
    PyErr_Clear();
    dlpack_tensor = PyCapsule_GetPointer(obj, "dltensor");
    if (dlpack_tensor) {
        // we need to call the deleter
        dlpack_tensor.deleter(dlpack_tensor)
    else {
        // something failed, let's pretend nothing happened
        PyErr_Clear();
    }
}
```

So in cython, something like:

```cython
cdef void deleter(object obj) noexcept:
    cdef DLManagedTensor *tensor = <DLManagedTensor*>0 # aside this is an unsafe cast since that might not be the rep of null, should use NULL
    try:
        tensor = <DLManagedTensor*>pycapsule.PyCapsule_GetPointer(obj, 'used_dltensor')
    except Exception:
        try:
            tensor = <DLManagedTensor*>pycapsule.PyCapsule_GetPointer(obj, 'dltensor')
            tensor.deleter(tensor)
        except Exception:
            pass
```

shwina: You're right that the call to `PyCapsule_GetPointer` can _technically_ raise an exception. And because the destructor is [not expected to raise](https://github.com/cython/cython/blob/master/Cython/Includes/cpython/pycapsule.pxd#L16), we cannot propagate any exceptions from it. That is what the `noexcept` is doing for us: it is ensuring that Cython will not wrap calls to `deleter` with code that will check for and raise Python exceptions.

I think that adding the `try-except-pass` would do more harm than good. To see why, notice that neither of these Cython programs will raise a Python error, and both will print "hello":

```cython                                                                                                                                            
cdef void foo() noexcept:                                                                                                                                                         
    raise ZeroDivisionError()                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                  
foo() 
```

```cython
cdef void foo() noexcept:                                                                                                                                                             try:                                                                                                                                                                          
        raise ZeroDivisionError()                                                                                                                                                 
    except:                                                                                                                                                                               pass                                                                                                                                                                      
                                                                                                                                                                    
foo()                                                                                                                                                                             
print("hello")
```

But, the first will print a warning about an exception being ignored. The second will be totally silent about any exceptions.

shwina: This being said, should we expect the destructor to be able to propagate Python exceptions? I'm not sure when the destructor is expected to run in relation to the state of the Python runtime and if that would make sense.

wence-: OK, makes sense.

wence- (approved):

shwina (approved):

PR #12930, opened 2023-03-13
Preserve integer dtype of hive-partitioned column containing nulls

rjzamora:
## Description
This is a follow-up "fix" for https://github.com/rapidsai/cudf/pull/12866
While that PR enables the writing/reading of null hive partitions using `dask_cudf`, it does not preserve the type of integer partition columns containing nulls. This PR should address the remaining issue.

## Checklist


rjzamora:
cc @randerzander

python/cudf/cudf/io/parquet.py
```diff
@@ -625,14 +626,17 @@ def _parquet_to_frame(
...
+                if pd.isna(value):
+                    _value, _len = [value] * _len, None
                 if partition_meta is not None:
                     dfs[-1][name] = as_column(
-                        value,
-                        length=len(dfs[-1]),
+                        _value,
+                        length=_len,
                         dtype=partition_meta[name].dtype,
                     )
```
rjzamora, line 637: @galipremsagar - Do you know the most efficient way to construct a completely-null column in cudf?

galipremsagar: Yup, you can do `cudf.core.column.column_empty(row_count=..., dtype=..., masked=True)`

galipremsagar: or if you want to fill with a specific value, we have `cudf.core.column.full(size=..., fill_value=..., dtype=...)`

rjzamora: Thanks!

wence- (approved):
I think this looks fine (although I don't know a lot about this partitioning scheme).

I had a comment about the uniquifying of names but I see that this is a long piece of string to pull on.

python/cudf/cudf/io/parquet.py
```diff
@@ -609,11 +609,12 @@ def _parquet_to_frame(
...
         for (name, value) in part_key:
+            _len = len(dfs[-1])
             if partition_categories and name in partition_categories:
                 # Build the categorical column from `codes`
-                codes = as_column(
-                    partition_categories[name].index(value),
-                    length=len(dfs[-1]),
+                codes = full(
+                    size=_len,
+                    fill_value=partition_categories[name].index(value),
```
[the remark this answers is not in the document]
rjzamora: Yes, exactly. We are reading a file from a hive-partitioned dataset. This means we are filling every row in the hive-partitioned column with the same value (a value that was encoded in the name of the directory where the file was stored). For example, if we were reading from the following path `"root_dir/month=12/part.2.parquet"`, we would need to assign `12` (or the categorical equivalent) to every row of the `"month"` column, since the data would  have been left out of the actual `"part.2.parquet"` file (usually).

python/cudf/cudf/io/parquet.py
```diff
@@ -886,8 +896,9 @@ def _get_groups_and_offsets(
         grouped_df.reset_index(drop=True, inplace=True)
     grouped_df.drop(columns=partition_cols, inplace=True)
     # Copy the entire keys df in one operation rather than using iloc
-    part_names = part_keys.to_pandas().unique().to_frame(index=False)
-
+    part_names = (
+        part_keys.to_pandas(nullable=True).unique().to_frame(index=False)
+    )
```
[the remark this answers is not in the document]
rjzamora: This is a good point. It's a shame to be using pandas to perform `unique`, but we need to preserve the initial order (including nulls), and we cannot rely on `Column.unique`, because `part_keys` may be a `MultiIndex` here. The good news is that we *should* be able to do `part_keys.take(part_offsets[:-1])` instead of `unique` anyway.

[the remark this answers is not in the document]
wence-: Assuming the number of partition keys is small, I don' think pandas usage is really a performance footgun...

Issue #11922, opened 2022-10-14
[FEA] The final step of the parquet preprocess step can get very slow for highly complicated schemas.


nvdbaranec:
For parquet files that contain columns with lists in them, we do a preprocess step to compute column sizes (and other information).  As a very last step, we iterate over the input columns and march through them by depth here:

https://github.com/rapidsai/cudf/blob/e91d7d9ef1eb3128de99f78d0127050bb12110d4/cpp/src/io/parquet/page_data.cu#L1827

For very complicated schemas, this can result in a large number of calls to `thrust::reduce()` and `thrust::exclusive_scan_by_key()`. In some cases, cases, so much so that it can dominate the rest of the decompress/decode work.

We should spend some time figuring out how to coalesce this into fewer (perhaps 1 each of `reduce_by_key` and `exclusive_scan_by_key`) calls.

nvdbaranec:
@abellina

vyasr:
@SrikarVanavasam FYI [Github supports certain keywords to automate linking issues to PRs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). If you had commented `Closes #11922` instead of `Addresses #11922` in it would have automatically linked and closed this issue when the PR merged. For future reference :smile:

PR #12931, opened 2023-03-13
Compute column sizes in Parquet preprocess with single kernel 

SrikarVanavasam:
Addresses #11922 

Currently in Parquet preprocessing a `thrust::reduce()` and `thrust::exclusive_scan_by_key()` is performed to compute the column size and offsets for each nested column. For complicated schemas this results in a large number of kernel invocations. This PR calculates the sizes and offsets of all columns in single calls to `thrust::reduce_by_key()` and `thrust::exclusive_scan_by_key()`. 

This change results in around 1.3x speedup when reading a complicated schema.
Before:
image

After:
image


GregoryKimball:
Would it be accurate to say that column size computation is 4x faster for this schema? Would you please share a few details about the schema in your test case?

SrikarVanavasam:
> Would it be accurate to say that column size computation is 4x faster for this schema? Would you please share a few details about the schema in your test case?

The highlighted region in the timelines includes the memsets for allocating the columns as well so the just the columns size and offset computation is actually many times faster in this case since, but I think it would be fair to say column allocation overall is 4x faster for this schema. This schema has 1629 columns with up to 8 levels of nesting which is why there were so many `thrust::reduce()` and `thrust::exclusive_scan_by_key()` calls before.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1211,24 +1211,59 @@ struct get_page_schema {
...
 };
 
+struct input_col_info {
+  int schema_idx;
+  size_t nesting_depth;
+};
+
+/**
+ * @brief Converts a 1-dimensional index into page, depth and column indices used in
+ * allocate_columns to compute columns sizes. The input index will through pages, nesting depth and
```
divyegala, line 1221: I think there's a word missing after "index will"

divyegala (commented):
I don't have enough domain knowledge to review this PR. The code looks fine to me, but someone else will need to verify the logic

PointKernel (commented):
Looks good in general with some small questions/suggestions

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1211,24 +1211,59 @@ struct get_page_schema {
...
+
+/**
+ * @brief Converts a 1-dimensional index into page, depth and column indices used in
+ * allocate_columns to compute columns sizes. The input index will iterate through pages, nesting depth and
+ * column indices in that order.
+ */
+struct get_indices {
+  size_t page_idx;
+  size_t depth_idx;
+  size_t col_idx;
```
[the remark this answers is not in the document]
vuule: `col_idx` should be `size_type`, given that `table::num_columns()` return this type. Since we can't have deeper nesting than the number of columns, IMO it's safe to use `size_type` for `depth_idx` as well.
`page_idx` is computed as `index % num_pages`, where index is `int`. So either `index` should be `size_t`, or we should use `int` for `page_idx`. This one only depends on Parquet specs, so I'm not sure what the real limit is.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1616,52 +1657,70 @@ void reader::impl::allocate_columns(size_t skip_rows, size_t num_rows, bool uses
...
+
+    hostdevice_vector<input_col_info> input_cols{_input_columns.size(), _stream};
+    size_t max_depth = 0;
+    for (size_t i = 0; i < _input_columns.size(); i++) {
+      auto depth                  = _input_columns[i].nesting_depth();
+      max_depth                   = depth > max_depth ? depth : max_depth;
+      input_cols[i].nesting_depth = depth;
+      input_cols[i].schema_idx    = _input_columns[i].schema_idx;
+    }
+    input_cols.host_to_device(_stream);
```
vuule, line 1670: Couple of suggestions here: if the data is only copied once, we should not use a `hostdevice_vector`; the loop can be replaced with `std::transform` + `std::max_element`
Something like

```
    std::vector<input_col_info> h_cols_info;
    h_cols_info.reserve(_input_columns.size());
    std::transform(_input_columns.cbegin(), _input_columns.cend(), 
    std::back_inserter(h_cols_info), 
        [](auto & col){return {depth,input_columns[i].schema_idx}; });
    auto const max_depth = std::max_element(h_cols_info.cbegin(), h_cols_info.cend(), 
        [](auto& l, auto& r){return l.nesting_depth < r.nesting_depth;});
    auto const d_cols_info = make_device_uvector_async(h_cols_info, stream, rmm::mr::get_current_device_resource());
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1616,52 +1657,70 @@ void reader::impl::allocate_columns(size_t skip_rows, size_t num_rows, bool uses
...
-
         // if this buffer is part of a list hierarchy, we need to determine it's
         // final size and allocate it here.
         //
         // for struct columns, higher levels of the output columns are shared between input
         // columns. so don't compute any given level more than once.
         if ((out_buf.user_data & PARQUET_COLUMN_BUFFER_FLAG_HAS_LIST_PARENT) && out_buf.size == 0) {
-          int size =
-            thrust::reduce(rmm::exec_policy(_stream), size_input, size_input + pages.size());
+          int size = sizes[(idx * max_depth) + l_idx];
```
vuule, line 1716:
```suggestion
          auto size = sizes[(idx * max_depth) + l_idx];
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1211,24 +1211,59 @@ struct get_page_schema {
...
+  int schema_idx;
+  size_t nesting_depth;
+};
+
+/**
+ * @brief Converts a 1-dimensional index into page, depth and column indices used in
+ * allocate_columns to compute columns sizes. The input index will iterate through pages, nesting depth and
+ * column indices in that order.
+ */
+struct get_indices {
```
vuule, line 1224: I don't think this name should be a verb, since the class is not a functor.

vuule (changes requested):
Impressive speedups!
Some surface-level comments, did not go into the actual algorithm changes yet.
Echoing @PointKernel 's observations about mixed integral types.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1211,24 +1211,59 @@ struct get_page_schema {
...
 
+struct input_col_info {
+  int schema_idx;
+  size_t nesting_depth;
+};
+
+/**
+ * @brief Converts a 1-dimensional index into page, depth and column indices used in
+ * allocate_columns to compute columns sizes. The input index will iterate through pages, nesting depth and
+ * column indices in that order.
```
ttnghia, line 1222:
```suggestion
 * @brief Converts a 1-dimensional index into page, depth and column indices used in
 * allocate_columns to compute columns sizes.
 *
 * The input index will iterate through pages, nesting depth and column indices in that order.
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1211,24 +1211,59 @@ struct get_page_schema {
...
     }
-    return page.nesting[depth].batch_size;
+
+    return page.nesting[indices.depth_idx].batch_size;
   }
 };
 
+struct get_reduction_key {
+  size_t num_pages;
+  __device__ size_type operator()(int index) { return index / num_pages; }
```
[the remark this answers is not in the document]
ttnghia: Yeah, there will be implicit data conversion here. It is really a concern.

If you know what you're doing, let's add `static_cast`.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1290,11 +1327,15 @@ struct start_offset_output_iterator {
  private:
   __device__ reference dereference(int index)
   {
-    gpu::PageInfo const& p = pages[page_indices[index]];
-    if (p.src_col_schema != src_col_schema || p.flags & gpu::PAGEINFO_FLAGS_DICTIONARY) {
+    auto indices = get_indices{index, max_depth, num_pages};
```
ttnghia, line 1330: `const`?
I see that @PointKernel has pointed out a lot of places that need `const`. So please remember to add `const` wherever possible.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1616,52 +1657,70 @@ void reader::impl::allocate_columns(size_t skip_rows, size_t num_rows, bool uses
...
+    thrust::exclusive_scan_by_key(rmm::exec_policy(_stream),
+                                  reduction_keys,
+                                  reduction_keys + num_keys,
+                                  size_input,
+                                  start_offset_output_iterator{pages.device_ptr(),
+                                                               page_index.begin(),
+                                                               0,
+                                                               input_cols.device_ptr(),
+                                                               max_depth,
+                                                               pages.size()});
```
ttnghia, line 1701: I wonder if we can even do better by combining these 2 kernel calls? Since they are operating on `reduction_keys`. Maybe just one `reduce_by_key` with a custom device lambda/functor that can do both reduce and scan?

cpp/src/io/parquet/reader_impl_preprocess.cu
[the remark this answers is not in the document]
karthikeyann: https://github.com/rapidsai/cudf/blob/69abff73b81b05fe2a8d76e07a7ca784f6cf9a3d/cpp/src/io/parquet/reader_impl_helpers.cpp#L401-L403

AFAIK, this is where schema index originate as `size_type`. Originally it was `size_t` stored in `SchemaElement::children_idx`, it's casted to size_type above.

https://github.com/rapidsai/cudf/blob/0af0c5910fa231f6071853ceef3b73827323538a/cpp/src/io/parquet/parquet.hpp#L156

karthikeyann (commented):
Prefer `auto` and `const` everywhere. Same comments.

cpp/src/io/parquet/reader_impl_preprocess.cu
SrikarVanavasam: I'm not quite sure how to create a functor for `reduce_by_key` that would also perform the scan but it could be possible. Do you have an idea in mind?

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1215,24 +1215,60 @@ struct get_page_schema {
...
+/**
+ * @brief Converts a 1-dimensional index into page, depth and column indices used in
+ * allocate_columns to compute columns sizes.
+ *
+ * The input index will iterate through pages, nesting depth and column indices in that order.
+ */
+struct reduction_indices {
+  const size_t _page_idx;
+  const size_type _depth_idx;
+  const size_type _col_idx;
```
PointKernel, line 1232:
```suggestion
  size_t const _page_idx;
  size_type const _depth_idx;
  size_type const _col_idx;
```
"east const" is preferred in libcudf since there is only one rule to follow: what's left of `const` is constant

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1215,24 +1215,60 @@ struct get_page_schema {
...
+
 /**
  * @brief Returns the size field of a PageInfo struct for a given depth, keyed by schema.
  */
 struct get_page_nesting_size {
-  size_type const src_col_schema;
-  size_type const depth;
+  input_col_info const* input_cols;
+  const size_type max_depth;
+  const size_t num_pages;
```
PointKernel, line 1248:
```suggestion
  size_type const max_depth;
  size_t const num_pages;
```
east const

vuule:
@SrikarVanavasam there are failing Parquet tests with the current version. Looks like it's related to the review changes, since CI passed with https://github.com/rapidsai/cudf/pull/12931/commits/e8b7c24ebba3a7bd8a3dcc79fec96de4fe6b4b4d

vuule:
CI errors:
```
[ RUN      ] ParquetWriterTest.StringsAsBinary
/opt/conda/conda-bld/work/cpp/tests/utilities/column_utilities.cu:279: Failure
Expected equality of these values:
  lhs_size
    Which is: 51
  rhs_size
    Which is: 8
Google Test trace:
/opt/conda/conda-bld/work/cpp/tests/io/parquet_test.cpp:837:  <--  line of failure

/opt/conda/conda-bld/work/cpp/tests/utilities/column_utilities.cu:279: Failure
Expected equality of these values:
  lhs_size
    Which is: 53
  rhs_size
    Which is: 8
Google Test trace:
/opt/conda/conda-bld/work/cpp/tests/io/parquet_test.cpp:837:  <--  line of failure

[  FAILED  ] ParquetWriterTest.StringsAsBinary (7 ms)
[ RUN      ] ParquetWriterTest.SlicedTable
terminate called after throwing an instance of 'cudf::fatal_cuda_error'
  what():  Fatal CUDA error encountered at: /opt/conda/conda-bld/work/cpp/include/cudf/detail/utilities/pinned_allocator.hpp:170: 700 cudaErrorIllegalAddress an illegal memory access was encountered
CMake Error at run_gpu_test.cmake:34 (execute_process):
  execute_process abnormal exit: Subprocess aborted
  ```

**Review by nvdbaranec — COMMENTED** [review]
This looks good to me. Only minor thing I'd bring up is that in the reduction_indices struct I'd suggest using the underscores for the constructor parameters, and keeping the fields of the struct itself without them.

I like the abstraction of the reduction_indices itself - clarifies how we're organizing the keys.

Also, if there's a bug that got introduced during PR changes, be sure to also re-run the Spark integration tests after fixing.

File: cpp/src/io/parquet/reader_impl_preprocess.cu [hunk 12]
```diff
@@ -1215,24 +1215,60 @@ struct get_page_schema {
...
+ */
+struct reduction_indices {
+  const size_t _page_idx;
+  const size_type _depth_idx;
+  const size_type _col_idx;
+
+  __device__ reduction_indices(size_t index, size_type max_depth, size_t num_pages)
+    : _page_idx(index % num_pages),
+      _depth_idx((index / num_pages) % max_depth),
+      _col_idx(index / (max_depth * num_pages))
```
ttnghia, line 1237: Since the member variables are public, do not use underscore prefix for them. To differentiate between member vars and parameters which have similar name, add underscore suffix to the parameter. For example:
```suggestion
  __device__ reduction_indices(size_t index_, size_type max_depth_, size_t num_pages_)
    : page_idx(index_ % num_pages_),
      depth_idx((index_ / num_pages_) % max_depth_),
      col_idx(index_ / (max_depth_ * num_pages_))
```

ttnghia: In addition, this constructor can be removed entirely because the struct instance can be initialized by initializer list.

nvdbaranec: I don't think that's a good idea.  Doing that puts the burden of understanding how we are mapping things on the caller.  The purpose of this struct is to hide that.

ttnghia: Yeah that sounds reasonable!

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1215,24 +1215,60 @@ struct get_page_schema {
...
+ * allocate_columns to compute columns sizes.
+ *
+ * The input index will iterate through pages, nesting depth and column indices in that order.
+ */
+struct reduction_indices {
+  size_t const page_idx;
+  size_type const depth_idx;
+  size_type const col_idx;
+
+  __device__ reduction_indices(size_t index_, size_type max_depth_, size_t num_pages_)
```
ttnghia, line 1234: Add `const` qualifier to all the functions that don't modify member variables.
```suggestion
  __device__ reduction_indices(size_t index_, size_type max_depth_, size_t num_pages_) const
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1215,24 +1215,60 @@ struct get_page_schema {
...
+  }
+};
+
 /**
  * @brief Returns the size field of a PageInfo struct for a given depth, keyed by schema.
  */
 struct get_page_nesting_size {
-  size_type const src_col_schema;
-  size_type const depth;
+  input_col_info const* input_cols;
```
ttnghia, line 1246:
```suggestion
  input_col_info const* const input_cols;
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1215,24 +1215,60 @@ struct get_page_schema {
...
+  size_type const max_depth;
+  size_t const num_pages;
   gpu::PageInfo const* const pages;
+  int const* page_indices;
```
ttnghia, line 1250: Again, I'm not sure if there is a reason to mix `int` and `size_type`. @nvdbaranec do you know?

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1620,52 +1662,76 @@ void reader::impl::allocate_columns(size_t skip_rows, size_t num_rows, bool uses
...
+        ._nesting_depth;
+
+    auto const d_cols_info = cudf::detail::make_device_uvector_async(
+      h_cols_info, _stream, rmm::mr::get_current_device_resource());
+
+    // size iterator. indexes pages by sorted order
+    auto const size_input = cudf::detail::make_counting_transform_iterator(
+      0,
+      get_page_nesting_size{
+        d_cols_info.data(), max_depth, pages.size(), pages.device_ptr(), page_index.begin()});
```
ttnghia, line 1689: Transform iterator is known to cause inflated compiling time when used in `thrust` APIs like `reduce_by_key` and `scan_by_key`. 

In addition, `size_input` is used twice, and each time it is used the function `get_page_nesting_size::operator(index)` may be called multiple times for each input index. That is inefficient.

Please materialize it instead:
```
auto size_input = rmm::device_uvector<size_type>(pages.size(), _stream);
thrust::transform(thrust::make_counting_iterator(0), thrust::make_counting_iterator(pages.size()) , size_input.begin(), get_page_nesting_size{
        d_cols_info.data(), max_depth, pages.size(), pages.device_ptr(), page_index.begin()});

...
// use it: size_input.begin()
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1215,24 +1215,60 @@ struct get_page_schema {
   __device__ size_type operator()(gpu::PageInfo const& page) { return page.src_col_schema; }
 };
 
+struct input_col_info {
+  int const _schema_idx;
+  size_type const _nesting_depth;
```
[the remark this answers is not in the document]
SrikarVanavasam: schema_idx is defined as an `int` in the input_column_info struct in parqeut_gpu.hpp so I left it like that.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1620,52 +1662,76 @@ void reader::impl::allocate_columns(size_t skip_rows, size_t num_rows, bool uses
...
+      h_cols_info, _stream, rmm::mr::get_current_device_resource());
+
+    // size iterator. indexes pages by sorted order
+    auto const size_input = cudf::detail::make_counting_transform_iterator(
+      0,
+      get_page_nesting_size{
+        d_cols_info.data(), max_depth, pages.size(), pages.device_ptr(), page_index.begin()});
+
+    auto const reduction_keys =
+      cudf::detail::make_counting_transform_iterator(0, get_reduction_key{pages.size()});
```
ttnghia, line 1692: This looks similar to `size_input`. However, since `get_reduction_key::operator()` is too simple, we can keep this as-is.

cpp/src/io/parquet/reader_impl_preprocess.cu
SrikarVanavasam: page_indices is `int` since page_index in chunk_intermediate_data is `int`

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1261,9 +1297,10 @@ struct chunk_row_output_iter {
 struct start_offset_output_iterator {
   gpu::PageInfo* pages;
   int const* page_indices;
-  int cur_index;
-  int src_col_schema;
-  int nesting_depth;
+  size_t cur_index;
```
ttnghia, line 1300: Why `cur_index` type is changed? Do we really need `size_t`?

SrikarVanavasam: `cur_index` is `size_t` since the index in `reduction_indices` is `size_t`

cpp/src/io/parquet/reader_impl_preprocess.cu
PointKernel: Hopefully, this change won't bring any runtime surprises.

PointKernel (approved):
I'm happy with the current state of this PR. Thanks for your effort in addressing the review comments!

cpp/src/io/parquet/reader_impl_preprocess.cu
vuule: @SrikarVanavasam Loos like this is the last comment to address before we can merge.

ttnghia (approved):

vuule (approved):

PR #12933, opened 2023-03-13
Make Parquet writer `nullable` option application to single table writes

vuule:
## Description

When writing multiple tables into a single Parquet file, users can run into issues if a column does not have nulls in the first table, but have some in other tables. The `nullable` member of `column_in_metadata` was originally added to address this and allow users to enforce nullability of columns from multiple tables. Because of this, the `nullable` option is only applied to chunked writes.

Recently, a different use for the option has been identified, where tables are stored into individual Parquet files, which are later read and the read tables are concatenated. Without the option to enforce nullability, Parquet files can end up with different nullabilities, i.e. different schemas, causing concatenation to fail.

This PR allows the nullable option to apply to single writes as well. The write call throws if user tried to write a column with nulls as non-nullable.

## Checklist


robertmaynard (approved):

bdice (approved):
This behavior change makes sense. Is it "breaking"? Code looks good.

ttnghia (approved):

vuule:
> This behavior change makes sense. Is it "breaking"? Code looks good.

The Parquet writer used to throw when called with the `nullable` option, and with this PR it will work. I wouldn't call that breaking, as no user will have to adjust their code because of this change.

PR #12938, opened 2023-03-13
Fix calculation of null counts for Parquet statistics

etseidl:
## Description
The current Parquet writer sometimes generates wrong values for `null_count` in the column chunk statistics and page indexes. This occurs for nested schemas when nulls occur at a level above the leaf values. This PR fixes the calculation by adding a `non_leaf_nulls` field to the `statistics_group` struct. This field is added to the chunk `null_count` calculated over leaf values in `gpu_calculate_group_statistics()`.

## Checklist


cpp/tests/io/parquet_test.cpp
```diff
@@ -4465,6 +4471,132 @@ TEST_F(ParquetWriterTest, CheckColumnOffsetIndexStruct)
   }
 }
 
+TEST_F(ParquetWriterTest, CheckColumnIndexListWithNulls)
+{
+  auto valids  = cudf::detail::make_counting_transform_iterator(0, [](auto i) { return i % 2; });
+  auto valids2 = cudf::detail::make_counting_transform_iterator(0, [](auto i) { return i != 3; });
```
vuule, line 4477: You should be able to use our special test iterator here
```suggestion
  auto null_at_3 = cudf::test::iterators::null_at(3);
```

cpp/src/io/parquet/page_enc.cu
```diff
@@ -282,6 +282,7 @@ __global__ void __launch_bounds__(128)
       g.col            = ck_g->col_desc;
       g.start_row      = fragments[frag_id].start_value_idx;
       g.num_rows       = fragments[frag_id].num_leaf_values;
+      g.non_leaf_nulls = fragments[frag_id].num_values - g.num_rows;
```
vuule, line 285: This will account for all nulls above the leaf level, even if we have nested lists?

cpp/tests/io/parquet_test.cpp
```diff
@@ -4465,6 +4471,132 @@ TEST_F(ParquetWriterTest, CheckColumnOffsetIndexStruct)
   }
 }
 
+TEST_F(ParquetWriterTest, CheckColumnIndexListWithNulls)
+{
+  auto valids  = cudf::detail::make_counting_transform_iterator(0, [](auto i) { return i % 2; });
```
vuule, line 4476: Maybe a descriptive name for the masks, to make it easier to read the test cases below.
I think my suggested name is not great, feel free to pick a different name.
```suggestion
  auto null_at_odd  = cudf::detail::make_counting_transform_iterator(0, [](auto i) { return i % 2; });
```

vuule (commented):
Looks good, just a few nitpicks in tests

cpp/src/io/parquet/page_enc.cu
etseidl: Yes, it should.  For instance, here's a dump of the first test case:
```
value 1: R:0 D:2 V:<null>
value 2: R:1 D:3 V:2
value 3: R:1 D:2 V:<null>
value 4: R:0 D:1 V:<null>
value 5: R:0 D:3 V:4
value 6: R:1 D:3 V:5
value 7: R:0 D:0 V:<null>
```
The R:0:D1 value is an empty list, while R0:D0 is a null list.  The D2 nulls are leaves.  Likewise for list of list:
```
value 1:  R:0 D:5 V:1
value 2:  R:2 D:5 V:2
value 3:  R:2 D:5 V:3
value 4:  R:1 D:3 V:<null>
value 5:  R:1 D:5 V:4
value 6:  R:2 D:5 V:5
value 7:  R:1 D:2 V:<null>
value 8:  R:1 D:4 V:<null>
value 9:  R:2 D:5 V:6
value 10: R:2 D:4 V:<null>
value 11: R:0 D:5 V:7
value 12: R:2 D:5 V:8
value 13: R:0 D:1 V:<null>
value 14: R:0 D:0 V:<null>
```

cpp/tests/io/parquet_test.cpp
etseidl: But...I just copied another test :cry: :rofl:  More seriously, yes, valid1/valid2 makes it hard to tell what's expected.  Will fix.

vuule (approved):

cpp/tests/io/parquet_test.cpp
```diff
@@ -4279,6 +4279,9 @@ TEST_F(ParquetWriterTest, CheckColumnOffsetIndexNulls)
       auto const ci    = read_column_index(source, chunk);
       auto const stats = parse_statistics(chunk);
 
+      // should be half nulls, except no nulls in column 0
+      EXPECT_EQ(stats.null_count, c > 0 ? num_rows / 2 : 0);
```
bdice, line 4283: I weakly prefer this spelling:
```suggestion
      EXPECT_EQ(stats.null_count, c == 0 ? 0 : num_rows / 2);
```

cpp/tests/io/parquet_test.cpp
```diff
@@ -4465,6 +4471,138 @@ TEST_F(ParquetWriterTest, CheckColumnOffsetIndexStruct)
   }
 }
 
+TEST_F(ParquetWriterTest, CheckColumnIndexListWithNulls)
+{
+  auto null_at_even_idx =
+    cudf::detail::make_counting_transform_iterator(0, [](auto i) { return i % 2; });
+  auto null_at_idx_3 = cudf::test::iterators::null_at(3);
```
bdice, line 4478: I think we typically write `using cudf::test::iterators::null_at;` and then write `null_at(3)` in the appropriate locations.

cpp/tests/io/parquet_test.cpp
```diff
@@ -4465,6 +4471,138 @@ TEST_F(ParquetWriterTest, CheckColumnOffsetIndexStruct)
   }
 }
 
+TEST_F(ParquetWriterTest, CheckColumnIndexListWithNulls)
+{
+  auto null_at_even_idx =
```
bdice, line 4476: It looks like the places where this is used are only nulling out one or two values. Could we just write `using cudf::test::iterators::nulls_at;` and then `nulls_at({0, 2})` or similar in those locations? If there was a longer list, then I would support using an iterator, but not for 1-2 values where being explicit is still concise. I don't feel strongly here, so we could leave this as-is if you prefer.

cpp/tests/io/parquet_test.cpp
```diff
@@ -4465,6 +4471,138 @@ TEST_F(ParquetWriterTest, CheckColumnOffsetIndexStruct)
...
+  lcw col5{{{{{1, 2, 3}, {}, {4, 5}, {}, {{0, 6, 0}, null_at_even_idx}}, null_at_idx_3},
+            {{7, 8}},
+            lcw{},
+            lcw{lcw{}}},
+           null_at_idx_3};
+
+  // 4 nulls
+  using strlcw = cudf::test::lists_column_wrapper<cudf::string_view>;
+  cudf::test::lists_column_wrapper<cudf::string_view> col6{
+    {{"Monday", "Monday", "Friday"}, {}, {"Monday", "Friday"}, {}, {"Sunday", "Funday"}},
```
bdice, line 4541: I like Funday! 😆

bdice (approved):
A few small suggestions, mostly for tests. Nothing is blocking here, so I'll approve.

cpp/tests/io/parquet_test.cpp
etseidl: I wish I could take credit for it. :rofl:

cpp/tests/io/parquet_test.cpp
etseidl: should I get rid of the declaration here and just use `null_at(3)` inline everywhere? (as with `nulls_at({0, 2})` below?)

bdice: Yup, that would be fine.

hyperbolic2346 (approved):

PR #12939, opened 2023-03-14
Implement `groupby.head` and `groupby.tail`

### LINKED ISSUE #2592: [QST] top N rows by group
I would like to select top N (2 in below example) rows by group. Below is pandas code used for that.
```py
import pandas as pd
d = pd.DataFrame.from_dict(dict([('id2', [1, 2, 1, 2, 1, 2]), ('id4', [1, 1, 1, 1, 1, 1]), ('v3', [1, 3, 2, 3, 3, 3])]))
d.sort_values('v3', ascending=False).groupby(['id2','id4']).head(2)
```
What is the cudf API to achieve it?

### LINKED ISSUE #12245: [FEA] groupby.head and tail
I wish I could use cuDF to do groupby.head and tail

I'm expecting this behavior like pandas

```
In [2]: 
   ...: df = pd.DataFrame(
   ...:     [
   ...:         ["A", 1],
   ...:         ["A", 2],
   ...:         ["A", 3],
   ...:         ["B", 4],
   ...:         ["B", 5],
   ...:         ["B", 6],
   ...:         ["B", 7],
   ...:         ["C", 8],
   ...: 
   ...:     ], columns=["x", "y"],)
   ...: 

In [3]: df
Out[3]: 
   x  y
0  A  1
1  A  2
2  A  3
3  B  4
4  B  5
5  B  6
6  B  7
7  C  8

In [4]: df.groupby("x").head(2)
Out[4]: 
   x  y
0  A  1
1  A  2
3  B  4
4  B  5
7  C  8

```

wence-:
## Description

These methods can be implemented by grouping the dataframe and
then selecting appropriate slices from each group. This is less
memory-efficient than it could be (since the entire grouping must
be constructed before discarding most of it).

- Closes #2592
- Closes #12245

## Checklist

### DIFF
--- a/python/cudf/cudf/core/groupby/groupby.py
+++ b/python/cudf/cudf/core/groupby/groupby.py
@@ -608,6 +608,98 @@ class GroupBy(Serializable, Reducible, Scannable):
 
     aggregate = agg
 
+    def _head_tail(self, n, *, take_head=True):
+        """Return the head or tail of each group
+
+        Parameters
+        ----------
+        n
+           Number of entries to include (if negative, number of
+           entries to exclude)
+        take_head
+           Do we want the head or the tail of the group
+
+        Returns
+        -------
+        New DataFrame or Series
+
+        Notes
+        -----
+        Unlike pandas, this returns an object in group order, not
+        original order
+        """
+        # A more memory-efficient implementation would merge the take
+        # into the grouping, but that probably requires a new
+        # aggregation scheme in libcudf. This is probably "fast
+        # enough" for most reasonable input sizes.
+        _, offsets, _, group_values = self._grouped()
+        group_offsets = np.asarray(offsets, dtype=np.int32)
+        size_per_group = np.diff(group_offsets)
+        # "Out of bounds" n for the group size either means no entries
+        # (negative) or all the entries (positive)
+        if n < 0:
+            size_per_group = np.maximum(
+                size_per_group + n, 0, out=size_per_group
+            )
+        else:
+            size_per_group = np.minimum(size_per_group, n, out=size_per_group)
+        if take_head:
+            group_offsets = group_offsets[:-1]
+        else:
+            group_offsets = group_offsets[1:] - size_per_group
+        to_take = np.arange(size_per_group.sum(), dtype=np.int32)
+        fixup = np.empty_like(size_per_group)
+        fixup[0] = 0
+        np.cumsum(size_per_group[:-1], out=fixup[1:])
+        to_take += np.repeat(group_offsets - fixup, size_per_group)
+        return group_values.iloc[to_take]
+
+    def head(self, n: int = 5):
+        """Return first n rows of each group
+
+        Parameters
+        ----------
+        n
+            If positive: number of entries to include from start of group
+            If negative: number of entries to exclude from end of group
+
+        Returns
+        -------
+        Series or DataFrame
+            Subset of the original grouped object as determined by n
+
+        .. pandas-compat::
+
+           Note that the returned object will be ordered group-wise
+           (with the same ordering as ``.apply(lambda x: x.head(n))``
+           though the original index is preserved, rather than in the
+           original input order.
+        """
+        return self._head_tail(n, take_head=True)
+
+    def tail(self, n: int = 5):
+        """Return last n rows of each group
+
+        Parameters
+        ----------
+        n
+            If positive: number of entries to include from end of group
+            If negative: number of entries to exclude from start of group
+
+        Returns
+        -------
+        Series or DataFrame
+            Subset of the original grouped object as determined by n
+
+        .. pandas-compat::
+
+           Note that the returned object will be ordered group-wise
+           (with the same ordering as ``.apply(lambda x: x.tail(n))``
+           though the original index is preserved, rather than in the
+           original input order.
+        """
+        return self._head_tail(n, take_head=False)
+
     def nth(self, n):
         """
         Return the nth row from each group.
--- a/python/cudf/cudf/tests/test_groupby.py
+++ b/python/cudf/cudf/tests/test_groupby.py
@@ -2,6 +2,7 @@
 
 import datetime
 import itertools
+import operator
 import textwrap
 from decimal import Decimal
 
@@ -1474,7 +1475,6 @@ def test_grouping(grouper):
 @pytest.mark.parametrize("agg", [lambda x: x.count(), "count"])
 @pytest.mark.parametrize("by", ["a", ["a", "b"], ["a", "c"]])
 def test_groupby_count(agg, by):
-
     pdf = pd.DataFrame(
         {"a": [1, 1, 1, 2, 3], "b": [1, 2, 2, 2, 1], "c": [1, 2, None, 4, 5]}
     )
@@ -1540,7 +1540,6 @@ def test_groupby_nth(n, by):
     reason="https://github.com/pandas-dev/pandas/issues/43209",
 )
 def test_raise_data_error():
-
     pdf = pd.DataFrame({"a": [1, 2, 3, 4], "b": ["a", "b", "c", "d"]})
     gdf = cudf.from_pandas(pdf)
 
@@ -1551,7 +1550,6 @@ def test_raise_data_error():
 
 
 def test_drop_unsupported_multi_agg():
-
     gdf = cudf.DataFrame(
         {"a": [1, 1, 2, 2], "b": [1, 2, 3, 4], "c": ["a", "b", "c", "d"]}
     )
@@ -2567,7 +2565,6 @@ def test_groupby_apply_series():
     ],
 )
 def test_groupby_apply_series_args(func, args):
-
     got = make_frame(DataFrame, 100).groupby("x").y.apply(func, *args)
     expect = (
         make_frame(pd.DataFrame, 100)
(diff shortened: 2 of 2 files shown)


python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -608,6 +608,98 @@ def _scan(self, op: str, *args, **kwargs):
...
+        else:
+            group_offsets = group_offsets[1:] - size_per_group
+        to_take = np.arange(size_per_group.sum(), dtype=np.int32)
+        fixup = np.empty_like(size_per_group)
+        fixup[0] = 0
+        np.cumsum(size_per_group[:-1], out=fixup[1:])
+        to_take += np.repeat(group_offsets - fixup, size_per_group)
+        return group_values.iloc[to_take]
+
+    def head(self, n: int = 5):
```
wence-, line 657: I note that groupby methods are inconsistently annotated with the `_cudf_nvtx_annotate` decorator. Shall I go through and do all of the sui generis methods (`.agg` is annotated)?

shwina: I think that'd be fine to tack on to this PR, yes

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -608,6 +608,98 @@ def _scan(self, op: str, *args, **kwargs):
 
     aggregate = agg
 
+    def _head_tail(self, n, *, take_head=True):
```
bdice, line 611: An alternative API would be to accept a slice and use that to generate the indices directly.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -608,6 +608,98 @@ def _scan(self, op: str, *args, **kwargs):
...
+        group_offsets = np.asarray(offsets, dtype=np.int32)
+        size_per_group = np.diff(group_offsets)
+        # "Out of bounds" n for the group size either means no entries
+        # (negative) or all the entries (positive)
+        if n < 0:
+            size_per_group = np.maximum(
+                size_per_group + n, 0, out=size_per_group
+            )
+        else:
+            size_per_group = np.minimum(size_per_group, n, out=size_per_group)
```
bdice, line 645: No need to assign AND specify the out parameter, afaik. Same for maximum above.

wence-: true, though the return value exists when out is specified so I just do this for ease of reading.

python/cudf/cudf/core/groupby/groupby.py
wence-: I did it separately as https://github.com/rapidsai/cudf/pull/12941 so that we can argue about exactly what should be annotated.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -608,6 +608,98 @@ def _scan(self, op: str, *args, **kwargs):
...
+        ----------
+        n
+            If positive: number of entries to include from start of group
+            If negative: number of entries to exclude from end of group
+
+        Returns
+        -------
+        Series or DataFrame
+            Subset of the original grouped object as determined by n
+
```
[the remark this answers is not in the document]
wence-: Done.

python/cudf/cudf/tests/test_groupby.py
```diff
@@ -2963,3 +2960,58 @@ def test_groupby_dtypes(groups):
...
+                )
+            expect_a, expect_b = zip(
+                *itertools.chain.from_iterable(
+                    slicefunc(list(group))
+                    for _, group in itertools.groupby(
+                        sorted(df.values_host.tolist(), key=keyfunc),
+                        key=keyfunc,
+                    )
+                )
+            )
```
wence-, line 3007: Too much itertools?

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -608,6 +608,151 @@ def _scan(self, op: str, *args, **kwargs):
...
+        Notes
+        -----
+        Unlike pandas, this returns an object in group order, not
+        original order
+        """
+        # A more memory-efficient implementation would merge the take
+        # into the grouping, but that probably requires a new
+        # aggregation scheme in libcudf. This is probably "fast
+        # enough" for most reasonable input sizes.
+        _, offsets, _, group_values = self._grouped()
```
wence-, line 635: Note that this OOMs for setups where I _really_ think it shouldn't. I tried (on an A6000, 48GiB device memory) creating a dataframe with two int-columns with 10^9 rows (and 10^6 groups) [8GiB data]. `df.groupby("a")._grouped()` consistently OOMs for me. AIUI this is a sort-based implementation of reordering which I would have thought runs in-place on a copy of the table, so I was expecting to peak at around 16GiB of data (even if it is out-of-place I would expect not much more than 24GiB of peak footprint). That said, I don't know enough about the libcudf implementation of groupbys to understand if my expectation is reasonable.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -608,6 +608,151 @@ def _scan(self, op: str, *args, **kwargs):
...
+           Do we want the head or the tail of the group
+
+        Returns
+        -------
+        New DataFrame or Series
+
+        Notes
+        -----
+        Unlike pandas, this returns an object in group order, not
+        original order
```
beckernick, line 629: Is it possible to optionally preserve order without introducing a large performance hit? For scenarios in which behavioral alignment with pandas is desired.

wence-: I think so, let me check...

wence-: I added a preserve_order keyword argument. It costs a bit.
```
df = create_df(
    100_000_000, # num rows
    1_000_000, # num groups
)

In [7]: %timeit df.groupby("a").head(2, preserve_order=False)
207 ms ± 1.18 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: %timeit df.groupby("a").head(2, preserve_order=True)
269 ms ± 742 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

shwina: I'm inclined to say that this is a case where users must opt _in_ to performance rather than opt _out_. That is, I'm proposing that `preserve_order=True` be the default.

I'd be more comfortable with opt out if `preserve_order=` was a kwarg that Pandas supported and we used a different default. But here we are introducing a new kwarg and making the default behaviour different from Pandas. This makes it impossible for someone to write their code to work identically with Pandas and cuDF.

Thoughts?

shwina: Yes, but the `sort=` keyword is one that Pandas supports for merge/groupby. So one could write their code to work identically with cuDF and Pandas by explicitly passing `sort=True` everywhere. That is not possible here. 

I agree that when we do support a single config option to control sorting behaviour across all APIs, we can change the default behaviour here to be `preserve_order=False`.

wence-: I think this is a difficult circle to square. My feeling is that the result order for many pandas operations being consistent with the input is merely an implementation detail and not part of the API contract. I am not sure how many people are relying on this: one would hope not many since any perturbation of their input data could (potentially) change output results.

shwina: In the context of aggregations, `sort=True` will sort the result by the group keys. Since there is only one row per group key in the result of an aggregation, cuDF and Pandas should return the same result when `sort=True`, regardless of whether a hash or sort-based  groupby was used internally.

shwina: Unfortunately, it appears that it is a guarantee. From https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.groupby.html:

> Groupby preserves the order of rows within each group.

python/cudf/cudf/tests/test_groupby.py
```diff
@@ -2963,3 +2960,78 @@ def test_groupby_dtypes(groups):
...
+            return df._empty_like(keep_index=True)
+        else:
+            if preserve_order:
+                # Should match pandas here
+                g = df.to_pandas().groupby("a")
+                if take_head:
+                    return g.head(n=n)
+                else:
+                    return g.tail(n=n)
+            else:
```
shwina, line 3002: Instead of constructing the result by hand, is it equivalent to take the result as-is, and sorting it (somehow)? to match Pandas order, then compare with the Pandas result?

python/cudf/cudf/core/groupby/groupby.py
wence-: > Groupby preserves the order of rows within each group.

This we are doing! 

The difference in ordering for the `head`/`tail` implementation is the pandas implementation does:

"Figure out which rows in the input dataframe correspond to the first `n` rows of each group, select those rows"

Whereas my implementation does "get the grouped dataframe, select the first `n` rows of each group". So it's a global ordering difference.

So if your dataframe is:
```
  a b
0 1 0
1 0 1
2 1 2
3 3 3
4 0 4
5 3 5
```
Then pandas `groupby("a").head(2)` produces
```
  a b
0 1 0
1 0 1
2 1 2
3 3 3
4 0 4
5 3 5
```

Whereas my implementation (without `preserve_order=True`) produces
```
   a b
1 0 1
4 0 4
0 1 0
2 1 2
3 3 3
5 3 5
```

So the order in each group is preserved

python/cudf/cudf/tests/test_groupby.py
```diff
@@ -2963,3 +2960,78 @@ def test_groupby_dtypes(groups):
...
+                # We groupby "a" which is the first column. This
+                # possibly relies on an implementation detail that for
+                # integer group keys, cudf produces groups in sorted
+                # (ascending) order.
```
wence-, line 3006: This is a "detail" of using the sort-based `groupby` code path in the implementation. So I think it is safe.

python/cudf/cudf/tests/test_groupby.py
wence-: One could `sort_index` which would do the trick, but I would prefer to check against a ground truth. I'm only checking against pandas in the other case because that's explicitly meant to match.

python/cudf/cudf/core/groupby/groupby.py
shwina: I guess I'm a bit confused now. What ordering (if any) are we unsure about being an implementation detail of Pandas and versus an API guarantee?

wence-: Sorry, pandas [groupby.head](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.GroupBy.head.html) says:

> Return first n rows of each group.

> Similar to .apply(lambda x: x.head(n)), but it returns a subset of rows from the original DataFrame with **original index and order preserved** (as_index flag is ignored).

(Emphasis mine).

So I guess that this does say "`head` really acts like subsampling and doesn't sort".

My implementation needs to undo a sort to get back to the original order of the dataframe. The question is, is whether two dataframes with the same index (just sorted differently) are considered equivalent enough.

shwina: Gotcha - of course the answer to that depends on how people write their code. But generally speaking, I would consider them not equivalent, because when using the default (RangeIndex), one may expect to index with `iloc[]` rather than `loc[]`.

mroeschke: > From what I remember seeing in a few old threads it was originally an implementation detail but something that too many users came to rely on for them to remove safely (@mroeschke could correct me on that).

Yeah the default sorting with groupby of pandas just came out of the original implementation. I don't recall seeing threads of users directly worried of this changing, but it's highly theorized by the core team that since it's been the default for a while there's some hesitation in changing it: https://github.com/pandas-dev/pandas/pull/48809

wence-: OK, to resolve this discussion in this PR, I'll flip the default to match pandas, but leave the note about performance in the docstring.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -608,6 +608,187 @@ def _scan(self, op: str, *args, **kwargs):
...
+        Returns
+        -------
+        Series or DataFrame
+            Subset of the original grouped object as determined by n
+
+        See Also
+        --------
+        .tail
+
+        Examples
```
shwina, line 699: As we now match Pandas' behavior, these examples would need to change.

I also think we don't need a `pandas-compat` note anymore (as the default behaviour is to match Pandas).

shwina (approved):
This looks good to me (pending docstring changes). Thanks, @wence-!

python/cudf/cudf/core/groupby/groupby.py
wence-: Thanks, done.

PR #10374, opened 2022-03-01
[REVIEW] Add `nvtx` annotations for `Series` and `Index`

galipremsagar:
This PR adds some missing `nvtx` annotations for `Series` and `Index`.

PR #12941, opened 2023-03-14
Add nvtx annotatations to groupby methods

wence-:
## Description

Not sure if not annotating these was an oversight during implementation...

## Checklist


bdice (commented):
@wence- This would be a good opportunity to add a section to our Developer Guide about what methods should be annotated. (Searching "nvtx" returns nothing right now.) Should (all/some) private methods be annotated, or only public methods? Let's look at past NVTX PRs for guidance here, like these:
- https://github.com/rapidsai/cudf/pull/10396
- https://github.com/rapidsai/cudf/pull/10374
- https://github.com/rapidsai/cudf/pull/10256
- https://github.com/rapidsai/cudf/pulls?q=is%3Apr+nvtx+is%3Aclosed

shwina:
> Should (all/some) private methods be annotated, or only public methods?

We have been super inconsistent about this so far, but I think all (and only) public methods should be annotated.

galipremsagar:
> Should (all/some) private methods be annotated, or only public methods?

I'd vote for all public methods to be annotated and select private methods on a case-by-case basis. Historically we've annotated private methods for the sake of easier nvtx profiling while working on performance issues.

bdice (approved):
@wence- I updated this PR for 23.06. I think this is in a good state. Comments above mentioned that only "select" private methods should be annotated. I think the private methods that are annotated here are good candidates for annotation, so I would recommend merging this in the current state.

If we want to add documentation on when to add annotations, I think the discussion above is a good starting point but I would write that in a separate PR so this can merge without further delay.

PR #12942, opened 2023-03-14
[REVIEW] Fix set index error for Series rolling window operations

galipremsagar:
## Description
This PR fixes an issue with Series rolling window operations where there is a failure because `Series.set_index` has been dropped in the previous release. 

## Checklist

### DIFF
--- a/python/cudf/cudf/core/window/rolling.py
+++ b/python/cudf/cudf/core/window/rolling.py
@@ -557,5 +557,6 @@ class RollingGroupby(Rolling):
             )
         )
 
-        result = super()._apply_agg(agg_name).set_index(index)
+        result = super()._apply_agg(agg_name)
+        result.index = index
         return result
--- a/python/cudf/cudf/tests/test_rolling.py
+++ b/python/cudf/cudf/tests/test_rolling.py
@@ -547,3 +547,13 @@ def test_rolling_indexer_support(indexer):
     actual = gdf.rolling(window=indexer, min_periods=2).sum()
 
     assert_eq(expected, actual)
+
+
+def test_rolling_series():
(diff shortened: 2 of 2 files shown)


python/cudf/cudf/core/window/rolling.py
```diff
@@ -557,5 +557,6 @@ def _apply_agg(self, agg_name):
             )
         )
 
-        result = super()._apply_agg(agg_name).set_index(index)
+        result = super()._apply_agg(agg_name)
+        result.index = index
```
wence-, line 561: Is it the case that `super()._apply_agg(agg_name)` will always produce a `Series` here? Because if it can produce a `DataFrame`, this change seems bad (`DataFrame.set_index` does significantly more than just assigning the `.index` attribute)

galipremsagar: For objects of `Index`/`MultiIndex` type, either way leads to the same results:
```python
In [9]: df = cudf.DataFrame({'a':[10, 11, 12, 13]}, index=['a', 'b', 'c', 'd'])

In [10]: df
Out[10]: 
    a
a  10
b  11
c  12
d  13

In [11]: df.set_index(cudf.Index([100, 200, 300, 400]))
Out[11]: 
      a
100  10
200  11
300  12
400  13

In [12]: df = cudf.DataFrame({'a':[10, 11, 12, 13]}, index=['a', 'b', 'c', 'd'])

In [13]: df.index = cudf.Index([100, 200, 300, 400])

In [14]: df
Out[14]: 
      a
100  10
200  11
300  12
400  13
```

wence- (approved):

PR #12945, opened 2023-03-14
[REVIEW] Avoid building cython twice

galipremsagar:
## Description
This PR moves the `build_ext` step to only happen when `INSTALL_TARGET==''`, thus avoiding the cython build to occur twice when `build.sh` is invoked.

## Checklist

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -331,9 +331,10 @@ fi
 if buildAll || hasArg cudf; then
 
     cd ${REPODIR}/python/cudf
-    python setup.py build_ext --inplace -- -DCMAKE_PREFIX_PATH=${INSTALL_PREFIX} -DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR} -DCMAKE_CUDA_ARCHITECTURES=${CUDF_CMAKE_CUDA_ARCHITECTURES} ${EXTRA_CMAKE_ARGS} -- -j${PARALLEL_LEVEL:-1}
     if [[ ${INSTALL_TARGET} != "" ]]; then
-        python setup.py install --single-version-externally-managed --record=record.txt  -- -DCMAKE_PREFIX_PATH=${INSTALL_PREFIX} -DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR} ${EXTRA_CMAKE_ARGS} -- -j${PARALLEL_LEVEL:-1}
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -331,9 +331,10 @@ fi
 if buildAll || hasArg cudf; then
 
     cd ${REPODIR}/python/cudf
-    python setup.py build_ext --inplace -- -DCMAKE_PREFIX_PATH=${INSTALL_PREFIX} -DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR} -DCMAKE_CUDA_ARCHITECTURES=${CUDF_CMAKE_CUDA_ARCHITECTURES} ${EXTRA_CMAKE_ARGS} -- -j${PARALLEL_LEVEL:-1}
     if [[ ${INSTALL_TARGET} != "" ]]; then
-        python setup.py install --single-version-externally-managed --record=record.txt  -- -DCMAKE_PREFIX_PATH=${INSTALL_PREFIX} -DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR} ${EXTRA_CMAKE_ARGS} -- -j${PARALLEL_LEVEL:-1}
+        python setup.py install --single-version-externally-managed --record=record.txt  -- -DCMAKE_PREFIX_PATH=${INSTALL_PREFIX} -DCMAKE_CUDA_ARCHITECTURES=${CUDF_CMAKE_CUDA_ARCHITECTURES} -DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR} ${EXTRA_CMAKE_ARGS} -- -j${PARALLEL_LEVEL:-1}
```
bdice, line 335: Remove extra space.
```suggestion
        python setup.py install --single-version-externally-managed --record=record.txt -- -DCMAKE_PREFIX_PATH=${INSTALL_PREFIX} -DCMAKE_CUDA_ARCHITECTURES=${CUDF_CMAKE_CUDA_ARCHITECTURES} -DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR} ${EXTRA_CMAKE_ARGS} -- -j${PARALLEL_LEVEL:-1}
```

bdice (approved):
This looks good to me -- but I feel like there was some context about why this works as it does. @vyasr Was there some issue where `scikit-build` should have known to "just install" in the second command? If not, then this is probably fine to merge as-is.

ttnghia (approved):
The issue annoyed me for a long time. Thanks for this.

vyasr:
> This looks good to me -- but I feel like there was some context about why this works as it does. @vyasr Was there some issue where `scikit-build` should have known to "just install" in the second command? If not, then this is probably fine to merge as-is.

The pair of commands predate the scikit-build transition. At the time I left it as is because I assumed that users of build.sh may originally have been relying on being able to import the package from inside the directory. The main difference between the behavior of build_ext followed by install vs just install (before the pyproject.toml transition) is that when only doing install you'll get import errors when you try to e.g. `import cudf` from inside the `python/cudf` directory because Python will prefer the local directory even though it's not a package. Running both commands makes that work as expected. However this is going to become the norm anyway when we move build.sh to use `pip install` commands, so I'm OK starting to impose that limitation now. Moving metadata to pyproject.toml has other subtle implications that we may as well start figuring out now as well.

PR #12948, opened 2023-03-14
[WIP] Add support for DELTA_BINARY_PACKED and DELTA_BYTE_ARRAY encodings to Parquet reader

etseidl:
## Description
Some Parquet writers will fall back to DELTA_BINARY_PACKED or DELTA_BYTE_ARRAY encoding when dictionary encoding cannot be used. This PR is a first attempt at adding support for these encodings to the Parquet reader. A description of these encodings can be found starting [here](https://github.com/apache/parquet-format/blob/master/Encodings.md#delta-encoding-delta_binary_packed--5).

I'm mostly looking for feedback on my approach right now. In particular, the final decoding of strings in DELTA_BYTE_ARRAY. Each string is encoded as a prefix length from the preceding string, a suffix length, and then the suffix bytes. To reconstruct `string_i`, you need `prefix_length(i)` bytes from `string_(i-1)`, which at first blush seems to be a serial operation. I've used a few cheats to try to be a bit more parallel, but am open to suggestions to make it even more so. The logic for this is in the `StringScan` function starting at line 2105 of page_data.cu.

I'm also wondering if it makes more sense to use all 128 threads to do decoding, rather than the current approach of using one warp for rep/def level decoding, one or two warps for delta decoding, and one warp outputting values (which mirrors how the current decoder works).

## Checklist


cpp/src/io/parquet/page_data.cu
```diff
@@ -1845,8 +1940,636 @@ __global__ void __launch_bounds__(block_size)
...
+        if (prefix_lens[l] == 0 && lane_out != nullptr) {
+          memcpy(lane_out, offsets[l], prefix_len);
+          prefix_lens[lane_id] = prefix_len = 0;
+        }
+      }
+      __syncwarp();
+
+      // check for finished
+      if (__all_sync(0xffff'ffff, prefix_len == 0)) { return; }
+    }
```
vuule, line 2184: This is great.
I was thinking of an approach where for each string we know the previous string (index) with a shorter prefix. It would eliminate the while loop here, but AFAICT it would cost just as much to initialize in the first place, so 🤷

cpp/src/io/parquet/page_data.cu
```diff
@@ -1845,8 +1940,647 @@ __global__ void __launch_bounds__(block_size)
...
+      __syncwarp();
+
+      if (lane_id > i) {
+        // copy data from the nearest prefix_len==0 neighbor to the left (if any),
+        // but do nothing if we hit a neighbor with a prefix_len < ours.
+        // this may be a bit of premature optimization, intended for cases like:
+        //  0  1  2  3  4
+        //  X  X  X  X  X
+        //  X     X     X
+        //  X     X
```
vuule, line 2176: Does the illustration represent prefix lengths [0, 2, 0, 2, 1]?

cpp/src/io/parquet/page_data.cu
```diff
@@ -1845,8 +1940,647 @@ __global__ void __launch_bounds__(block_size)
...
+        // without the following loop, we'll fill in C1 when i==1, continue when i==2,
+        // and then fill in 3 & 4 when the __all_sync() above finally returns true. with this
+        // optimization, we'll fill in C1, C3, and C4 when i==0.
+        int l = lane_id - 1;
+        while (l > i && prefix_lens[l] != 0 && prefix_len <= prefix_lens[l]) {
```
vuule, line 2181: Do we need to search in every iteration? I think we can find the string from which we will eventually copy the prefix in the first iteration. It will be the first element on the left, with a shorter prefix. This "predecessor" will not change, as all elements between it and the current index will the set at the same time or after the current element.
If we set this up, we only need to check this element in every iteration, and copy its data once it's been resolved (from its own predecessor).
It would effectively construct a DAG that determines the order in which the elements are set/decoded.

etseidl: Ooh, that's good. I'll be happy to get rid of that inner loop :)

cpp/src/io/parquet/page_data.cu
etseidl: Yes

etseidl:
For similar data, I would up with the following read times (as measured by nsys):
```
                 Dict   Delta
zstd            218ms    74ms
buildStringDict  81ms     -   
computeStringsz   -      10ms
decodePageData   47ms    85ms
```
The delta decode step probably needs some work, but at least the increase in time there is offset by not needing to decompress dictionary pages.

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -430,6 +434,21 @@ void BuildStringDictionaryIndex(ColumnChunkDesc* chunks,
...
+/**
+ * @brief Compute the amount of string data in DELTA_BYTE_ARRAY encoded pages.
+ *
+ * @param[in,out] pages All pages to compute lengths for
+ * @param[in] chunks Chunks that contain the pages
+ * @param[in] num_rows Total number of rows to read
+ * @param[in] min_row crop all rows below min_row
+ * @param[in] stream CUDA stream to use, default 0
+ */
+void ComputePageStringSizes(hostdevice_vector<PageInfo>& pages,
```
nvdbaranec, line 446: Might be a good idea to throw a "Delta" somewhere in the name here to distinguish it a bit.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -121,7 +122,52 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
...
+                     pages.begin(),
+                     [&delta_binary_page_data](auto& page, auto offset) {
+                       if (page.page_strings_size != 0) {
+                         page.page_string_data =
+                           static_cast<uint8_t*>(delta_binary_page_data.data()) + offset;
+                       }
+                       return page;
+                     });
+
+      pages.host_to_device(_stream);
```
nvdbaranec, line 162: I think this could all be done on the device with thrust (except for the creation of the device_buffer of course), saving us a round-trip of the pages.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -121,7 +122,52 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
...
+    }
+  }
+
   gpu::DecodePageData(pages, chunks, num_rows, skip_rows, _stream);
+  if (has_delta_binary) { gpu::DecodeDeltaBinary(pages, chunks, num_rows, skip_rows, _stream); }
+  if (has_delta_byte_array) {
+    gpu::DecodeDeltaByteArray(pages, chunks, num_rows, skip_rows, _stream);
+  }
```
nvdbaranec, line 170: Note to future programmers:  if we get multi-stream cuIO working, launching all 3 of these simultaneously on 3 seperate streams would be awesome :)

nvdbaranec (commented):
Quick round of comments.  Still have to work through page_data.cu

cpp/src/io/parquet/reader_impl.cpp
etseidl: I tried this already :)  It was!

etseidl: Screenshot from 2023-03-28 13-39-53

nvdbaranec (commented):
I think it might be worth splitting this file (`page_data.cu`) up into multiple pieces at this point.  It's getting pretty huge.  Maybe 1 file each for 1.) preprocessing,  2.) regular decoding,  3.) delta decoding.

I'd be ok with doing that as a followup PR.

etseidl:
> I think it might be worth splitting this file (`page_data.cu`) up into multiple pieces at this point. It's getting pretty huge. Maybe 1 file each for 1.) preprocessing, 2.) regular decoding, 3.) delta decoding.
> 
> I'd be ok with doing that as a followup PR.

I don't think this is going to merge any time soon, so I'm happy to do the splitting.  I can refactor all the thrift decoders at the same time. It's just going to take some time...right now I'm experimenting with running the delta decoders across 4 warps, rather than the current staggered approach of 2-3 warps producing and one warp consuming. It's a more natural fit since the vast majority of the time there will be 128 values per block. The only hitch there is `gpuDecodeLevels`, but hopefully that won't hurt too bad. I got the delta binary working already, and that saw a 10% speed bump even with level decoding running on a single warp.

cpp/src/io/parquet/page_data.cu
```diff
@@ -1845,8 +1940,663 @@ __global__ void __launch_bounds__(block_size)
...
+      CalcMiniBlockValues(db, lane_id);
+
+      uint32_t const idx = db->current_value_idx + i + lane_id;
+      u64 const my_len   = idx < db->value_count ? db->value[rolling_index(idx)] : 0;
+
+      // get sum for warp
+      using WarpReduce = cub::WarpReduce<u64>;
+      __shared__ typename WarpReduce::TempStorage temp_storage;
+      // note: sum_len will only be valid on thread 0.
+      u64 const sum_len = WarpReduce(temp_storage).Sum(my_len);
```
nvdbaranec, line 2293: I think this is a bug.  If I'm reading this right, 2 warps will be in this loop at the same time.  But this shared memory is allocated a the block level (ie, it doesn't respect the scoping here) so these warps will be stomping on each other.  You would need to declare this as an array of size 2 indexed by the warp id.

Might be more clear if this were declared at the top of the function.

cpp/src/io/parquet/page_data.cu
```diff
@@ -1845,8 +1940,663 @@ __global__ void __launch_bounds__(block_size)
...
+
+      if (lane_id == 0) { atomicAdd(&total_bytes, sum_len); }
+      __syncthreads();
```
nvdbaranec, line 2296:
```suggestion
      __syncwarp();
```
I think this can just be a syncwarp.  The `atomicAdd` takes care of you at the block level.

Alternately, you could just store the per-warp total_bytes and do a single call to `cudf::detail::single_lane_block_sum_reduce` at the very end to sum them together, with absolute thread value t==0 doing the final store.  Would dodge the atomics entirely.

nvdbaranec (commented):
Still chipping away.

cpp/src/io/parquet/reader_impl.cpp
vuule: Could I interest you in `rmm::cuda_stream_pool`? :)

cpp/src/io/parquet/page_data.cu
```diff
@@ -99,6 +102,59 @@ struct page_state_s {
...
+  uint8_t const* cur_mb_start;   // pointer to the start of the current mini-block data
+  uint8_t const* cur_bitwidths;  // pointer to the bitwidth array in the block
+
+  uint64_t value[non_zero_buffer_size];  // circular buffer of delta values
+};
+
+struct delta_byte_array_state_s {
+  delta_binary_state_s prefixes;          // state of decoder for prefix lengths
+  delta_binary_state_s suffixes;          // state of decoder for suffix lengths
+  uint64_t offset[non_zero_buffer_size];  // circular buffer for string output offsets
```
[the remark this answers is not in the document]
etseidl: `value` is unbounded.  I'm still wondering what happens where you have adjacent original values of LONG_MAX and LONG_MIN, where the delta will be huge. If `min_delta` is then negative, there would be overflow.  I have to read the spec closer to see how that would be handled.

`offset` is an offset into a per-page buffer used to store the fully formed intermediate strings. This data will then be copied into the column data later on when `make_strings_column` is called. So I guess this could be `size_type`, if that's where your question was leading.

cpp/src/io/parquet/reader_impl.cpp
etseidl: Your ideas are intriguing to me, and I wish to subscribe to your newsletter. :)

cpp/src/io/parquet/page_data.cu
```diff
@@ -1845,8 +1940,663 @@ __global__ void __launch_bounds__(block_size)
...
+        db->current_value_idx++;
+        db->value[0] = db->last_value;
+      }
+    }
+    __syncwarp();
+
+    db->value[rolling_index(db->current_value_idx + warp_size * i + lane_id)] = delta;
+
+    // save delta from last lane in warp
+    if (lane_id == 31) { db->last_value = delta; }
```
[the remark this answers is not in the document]
etseidl: the `__syncwarp` is actually guarding the update of `current_value_idx`. I can tuck it into the `if` block.

cpp/src/io/parquet/page_data.cu
```diff
@@ -1845,8 +1940,663 @@ __global__ void __launch_bounds__(block_size)
   }
 }
 
+// DELTA_XXX encoding support
+//
+// returns the number of values encoded in the block data. when is_decode is true,
+// account for the first value in the header. otherwise just count the values encoded
+// in the mini-block data.
```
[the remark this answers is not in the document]
etseidl: Comment added...I'll split the delta stuff into a separate file tomorrow.

cpp/src/io/parquet/page_data.cu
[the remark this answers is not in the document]
etseidl: I wound up moving the increment of `current_value_idx` to outside the loop, so now I have a single `__syncwarp` after updating `last_value`

cpp/src/io/parquet/page_data.cu
```diff
@@ -1845,8 +1940,691 @@ __global__ void __launch_bounds__(block_size)
...
+        delta &= (1 << mb_bits) - 1;
+      }
+    }
+
+    // add min delta to get true delta
+    delta += db->cur_min_delta;
+
+    // do inclusive scan to get value - first_value at each position
+    __shared__ cub::WarpScan<int64_t>::TempStorage temp_storage;
+    cub::WarpScan<int64_t>(temp_storage).InclusiveSum(delta, delta);
```
nvdbaranec, line 2115: Same issue as below.  Multiple warps contending for the same shared memory.

cpp/src/io/parquet/page_data.cu
```diff
@@ -1845,8 +1940,691 @@ __global__ void __launch_bounds__(block_size)
...
+
+  for (int idx = start_idx; idx < end_idx; idx += warp_size) {
+    uint32_t const ln_idx = idx + lane_id;
+
+    // calculate offsets into suffix data
+    uint32_t const src_idx    = rolling_index(ln_idx);
+    uint64_t const suffix_len = ln_idx < suffix_db->value_count ? suffix_db->value[src_idx] : 0;
+    uint64_t suffix_off       = 0;
+    __shared__ cub::WarpScan<uint64_t>::TempStorage temp_storage;
+    cub::WarpScan<uint64_t>(temp_storage).ExclusiveSum(suffix_len, suffix_off);
```
nvdbaranec, line 2219: It looks like the only ways into this function involve one warp, so this will work. But it does seem vulnerable to someone rearranging things later on where multiple warps are used.

etseidl: There's a comment at the top that the function is only supported on a single warp. Not sure what else can be done to guard it.  No more vulnerable than `gpuDecodeLevels`...I guess I can make the warning stand out more in the docstring (well, once there is a docstring).

etseidl:
I *think* I've addressed most of the comments, so now I'll start breaking out the delta stuff into a separate file. There are still issues that came up during review.

The first is data typing. The spec actually specifies 128-bit varints, but I'm currently using a mix of 32 and 64. Are there performance issues using `__int128_t` when doing the thrift decoding, and then casting to smaller bit widths when possible? It's likely overkill, but as soon as I say "we're unlikely to ever see..." someone will prove me wrong.

The second is block/mini-block parameters. It seems like the parquet world has settled on block sizes of 128, and 4 mini-blocks of 32 elements each as the norm. This fits naturally with the current `block_size` of 128 used in the page decoder. The only counter I've found so far is in arrow-rs, where they'll use a block size of 256 and mini-block count of 64 when encoding 64-bit integers. But the spec merely says block size is a multiple of 128, and mini-block size is a multiple of 32. I can imagine an optimization for the following example. Say you have a column of 20000 zeros. Using 128/32 encoding, you'd have a first value of '0', and then 19999 '0' deltas. Since the deltas are all '0', you need zero bits to encode them.  Each block then is a varint encoded 0 for min delta, 4 bytes of 0 for the bitwidths, and nothing else.  Now you've got round_up(20000/128) * 5 bytes = 785 bytes to encode that column (plus like 7 bytes for the header).  But what if you set block size to 20096, and use a single mini-block of 20096. Now you have a single block with a 2 byte header (one byte for min delta and one byte of 0 bitwidth), plus the delta header and you have under 10bytes.  Now this can compare to dict/RLE which will also be very small.  But a single block of 20000 will blow up the current implementation. I'm thinking I'll need different implementations (maybe a templated version for the 128/32 and 256/64 cases and a more general one to handle everything else) and some kind of dispatcher to call the correct one. Does that seem reasonable?

etseidl:
Another TODO: `gpuDecodeTotalPageStringSize` needs to be made aware of the delta encoding. Right now strings sizes is turned off in `gpuComputePageSizes` when the page encoding is`DELTA_BYTE_ARRAY`.

etseidl:
Closing for now.  Will resubmit as part of #13501

PR #12949, opened 2023-03-14
Refactor orc chunked writer

ttnghia:
The current ORC chunked writer performs compressing/encoding and writing data into the output data sink without any safeguard. This PR modifies the internal `writer::impl::write()` function, separating it into multiple pieces:
 * A free function that performs compressing/encoding the input table into intermediate results. These intermediate results are totally independent of the writer. As such, the writer can be isolated from failures of this free function, allowing to retry upon failure.
 * After having the intermediate results in the previous step, these results will be actually applied to the output data sink to start the actual data writing.

Some cleanup is also performed on the existing code. That includes moving some member functions into free functions, which helps reducing potential dependencies between translation units.

There is no new implementation added in this work. Only the existing code is moved around.

Partially contributes to https://github.com/rapidsai/cudf/issues/12792.


cpp/src/io/orc/writer_impl.cu
```diff
@@ -2174,152 +2274,307 @@ void writer::impl::write(table_view const& table)
...
+                                           compression_blocksize_,
+                                           stats_freq_,
+                                           single_write_mode,
+                                           *out_sink_,
+                                           stream);
+
+  // Compression/encoding were all successful. Now write the intermediate results.
+  auto const num_rows = input.num_rows();
+  if (num_rows > 0) {
+    write_data_internal(streams,
```
vuule, line 2474: `write_orc_data_to_sink`?

cpp/src/io/orc/writer_impl.cu
```diff
@@ -2174,152 +2274,307 @@ void writer::impl::write(table_view const& table)
...
+                        segmentation,
+                        stripes,
+                        orc_table,
+                        compressed_data,
+                        intermediate_stats,
+                        stream_output.get());
+  }
+
+  // Update data into the footer. This needs to be called even when num_rows==0.
+  update_table_to_footer(orc_table, stripes, num_rows);
```
vuule, line 2487: `add_table_to_footer_data`?

cpp/src/io/orc/writer_impl.cu
```diff
@@ -2174,152 +2274,307 @@ void writer::impl::write(table_view const& table)
...
+{
+  // Write file header
+  out_sink_->host_write(MAGIC, std::strlen(MAGIC));
+}
+
+void writer::impl::write(table_view const& input)
+{
+  CUDF_EXPECTS(not closed, "Data has already been flushed to out and closed");
+
+  if (not table_meta) { table_meta = make_table_meta(input); }
```
vuule, line 2443: can `table_meta` be moved to `process_for_write`? I can't find where it's used in later steps.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -2123,9 +2174,51 @@ void writer::impl::write(table_view const& table)
...
+           hostdevice_vector<compression_result>,
+           hostdevice_2dvector<gpu::StripeStream>,
+           encoded_data,
+           file_segmentation,
+           std::vector<StripeInformation>,
+           orc_table_view,
+           rmm::device_buffer,
+           intermediate_statistics,
+           pinned_buffer<uint8_t>>
+process_for_write(table_view const& input,
```
vuule, line 2207: `convert_table_to_orc_data`?

cpp/src/io/orc/writer_impl.cu
```diff
@@ -2174,152 +2274,307 @@ void writer::impl::write(table_view const& table)
...
+                                           *table_meta,
+                                           max_stripe_size,
+                                           row_index_stride,
+                                           enable_dictionary_,
+                                           compression_kind_,
+                                           compression_blocksize_,
+                                           stats_freq_,
+                                           single_write_mode,
+                                           *out_sink_,
+                                           stream);
```
vuule, line 2469: Maybe we can catch exceptions coming from this call and log that no data has been written to the sink.

vuule (changes requested):
Looks good, just minor suggestions.
It feels like this initial split opens up opportunities for further clean up, now that the dependencies between steps are obvious.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -2174,152 +2274,307 @@ void writer::impl::write(table_view const& table)
...
+
+  // Compression/encoding were all successful. Now write the intermediate results.
+  auto const num_rows = input.num_rows();
+  if (num_rows > 0) {
```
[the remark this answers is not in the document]
ttnghia: Yes, this is obvious. I also thought about that, but didn't do it because we have to pass in `num_rows` just to move this `if()` line inside `write_data_internal`.
Do you have a suggestion to retrieve `num_rows` from other parameters?

cpp/src/io/orc/writer_impl.cu
ttnghia: `table_meta` is created once for all `write()` calls. So if we create it in `process_for_write`, we still need to assign it later. From the current code:
```
if (not table_meta) { table_meta = make_table_meta(input); }

auto [....] = process_for_write(...);
```
We will have this, which is ugly IMO:
```
auto [....table_meta_ ] = process_for_write(...);
table_meta = std::move(table_meta_);
```

cpp/src/io/orc/writer_impl.cu
ttnghia: How to log? Like this?
```
try {
....
} catch (exception& e) {
  CUDF_LOG_WARN("ORC writer encounters exception during write....");
  throw e;
}
```

vuule: Something like that, yes.
I was thinking that we should emphasize that the sink was not affected, as this is probably not expected (knowing previous behavior). We probably don't need to log the error, since it will be returned to the caller anyway.
`CUDF_LOG_ERROR("ORC writer encountered exception during processing. No data has been written to the sink.");`

cpp/src/io/orc/writer_impl.cu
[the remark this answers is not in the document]
vuule: You should be able to use `orc_table.num_rows()`.

vyasr (approved):
This PR was confusing for a minute until I realized that orc streams were not CUDA streams...

Nice refactor. I have some questions about the code, but AFAIK none of the questions are related to changes in this PR. The moves all look fine.

PR #12952, opened 2023-03-15
[REVIEW] Add `nullability` parameter to parquet writer

galipremsagar:
## Description
Requires: https://github.com/rapidsai/cudf/pull/12933

This PR adds `nullability` parameter to parquet writer. When it is `True`, all columns are written as `null` in the schema. When `False`, all columns are written as `not null` in the schema, however, if a column contains null values, this parameter is ignored.

## Checklist


python/cudf/cudf/_lib/parquet.pyx
```diff
@@ -675,15 +687,26 @@ cdef cudf_io_types.compression_type _get_comp_type(object compression):
         raise ValueError("Unsupported `compression` type")
 
 
-cdef _set_col_metadata(Column col, column_in_metadata& col_meta):
+cdef _set_col_metadata(
+    Column col,
+    column_in_metadata& col_meta,
+    object nullability
+):
+    if col.nullable and nullability is False:
```
vuule, line 695: Had to check that `None` is not equal to `False. That's neat, thank you Python :)

vuule (approved):
looks good, thank you for the quick PR!

python/cudf/cudf/_lib/parquet.pyx
```diff
@@ -467,6 +469,11 @@ cdef class ParquetWriter:
     max_page_size_rows: int, default 20000
         Maximum number of rows of each page of the output.
         By default, 20000 will be used.
+    nullability : bool, default None.
```
bdice, line 472: The default is None but the behavior of the default value is not explained.

python/cudf/cudf/_lib/parquet.pyx
```diff
@@ -675,15 +687,26 @@ cdef cudf_io_types.compression_type _get_comp_type(object compression):
...
 
 
-cdef _set_col_metadata(Column col, column_in_metadata& col_meta):
+cdef _set_col_metadata(
+    Column col,
+    column_in_metadata& col_meta,
+    object nullability
+):
+    if col.nullable and nullability is False:
+        warnings.warn(f"column {col_meta.get_name().decode()} contains null "
```
bdice, line 696: How do we decide if this is worth issuing a warning? It seems like there are valid use cases for nullability=False that are intentional and not deserving of a warning.

python/cudf/cudf/io/parquet.py
```diff
@@ -179,14 +182,18 @@ def write_to_dataset(
...
         If None, 20000 will be used.
-
     storage_options : dict, optional, default None
         Extra options that make sense for a particular storage connection,
         e.g. host, port, username, password, etc. For HTTP(S) URLs the
         key-value pairs are forwarded to ``urllib.request.Request`` as
         header options. For other URLs (e.g. starting with "s3://", and
         "gcs://") the key-value pairs are forwarded to ``fsspec.open``.
         Please see ``fsspec`` and ``urllib`` for more details.
+    nullability : bool, default None.
```
bdice, line 192: Same here, None needs docs.

python/cudf/cudf/_lib/parquet.pyx
vuule: > It seems like there are valid use cases for nullability=False

I'm not sure about this. Columns without nulls would have been written as non-nullable in Parquet(?) even without `nullability=False`, and the option will be ignored for columns with nulls. So unless we are passing this to the chunked writer, I don't think `nullability=False` does anything.

bdice: Let's align with @vuule on the C++ side -- I am fine with these names but would prefer an `enum class` instead of `bool` on the C++ side for alignment (if it's currently a `bool` -- can't recall).

galipremsagar: It's currently a bool on c++ side.

bdice: That sounds like an improvement to me. Let's do that. `force_nullable=True` or `force_nullable=False`, default is `False`.

python/cudf/cudf/_lib/parquet.pyx
```diff
@@ -675,15 +687,26 @@ cdef cudf_io_types.compression_type _get_comp_type(object compression):
...
+                      f"values, nullability={nullability} is being ignored")
+        col_meta.set_nullability(True)
+    elif nullability is not None:
```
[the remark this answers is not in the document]
bdice: This question does not need to be resolved if we go the route of `force_nullable`, since the option set will be constrained to `True, False`.

python/cudf/cudf/utils/ioutils.py
```diff
@@ -290,6 +290,11 @@
     include the file path metadata (relative to `root_path`).
     To request metadata binary blob when using with ``partition_cols``, Pass
     ``return_metadata=True`` instead of specifying ``metadata_file_path``
+nullability : bool, default None.
```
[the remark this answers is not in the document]
galipremsagar: Now stale with new changes.

python/cudf/cudf/io/parquet.py
galipremsagar: Now not relevant with new changes.

python/cudf/cudf/_lib/parquet.pyx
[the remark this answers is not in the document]
galipremsagar: Done 👍

python/cudf/cudf/_lib/parquet.pyx
galipremsagar: I went ahead with `force_nullable_schema` option and updated the docs.

python/cudf/cudf/_lib/parquet.pyx
galipremsagar: Now not relevant with new changes.

python/cudf/cudf/_lib/parquet.pyx
```diff
@@ -467,6 +470,11 @@ cdef class ParquetWriter:
     max_page_size_rows: int, default 20000
         Maximum number of rows of each page of the output.
         By default, 20000 will be used.
+    force_nullable_schema : bool, default False.
+        If True, writes all columns as `null` in schema.
+        If False, writes all columns as `not null` in schema,
+        however if a column contains null values, this parameter
+        is ignored.
```
vuule, line 477: I don't like the `If False, writes all columns as `not null` in schema` part.
I'd rather phrase this as "in <this case>, <this> happens, in the <other case> the <other thing> happens"

bdice: I'm not sure if I understand @vuule's proposal, but this is what I would write:
```suggestion
        If False, columns are written as `null` if they contain null values,
        otherwise as `not null`.
```

python/cudf/cudf/io/parquet.py
```diff
@@ -179,14 +182,18 @@ def write_to_dataset(
...
     storage_options : dict, optional, default None
         Extra options that make sense for a particular storage connection,
         e.g. host, port, username, password, etc. For HTTP(S) URLs the
         key-value pairs are forwarded to ``urllib.request.Request`` as
         header options. For other URLs (e.g. starting with "s3://", and
         "gcs://") the key-value pairs are forwarded to ``fsspec.open``.
         Please see ``fsspec`` and ``urllib`` for more details.
+    force_nullable_schema : bool, default False.
+        If True, writes all columns as `null` in schema.
+        If False, writes all columns as `not null` in schema,
```
bdice, line 194: Update this according to the suggestion above.

python/cudf/cudf/utils/ioutils.py
```diff
@@ -290,6 +290,11 @@
     include the file path metadata (relative to `root_path`).
     To request metadata binary blob when using with ``partition_cols``, Pass
     ``return_metadata=True`` instead of specifying ``metadata_file_path``
+force_nullable_schema : bool, default False.
+    If True, writes all columns as `null` in schema.
+    If False, writes all columns as `not null` in schema,
```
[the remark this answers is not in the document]
galipremsagar: Done

python/cudf/cudf/io/parquet.py
galipremsagar: Done

python/cudf/cudf/_lib/parquet.pyx
```diff
@@ -675,15 +688,33 @@ cdef cudf_io_types.compression_type _get_comp_type(object compression):
         raise ValueError("Unsupported `compression` type")
 
 
-cdef _set_col_metadata(Column col, column_in_metadata& col_meta):
+cdef _set_col_metadata(
+    Column col,
+    column_in_metadata& col_meta,
+    object force_nullable_schema
```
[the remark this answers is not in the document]
galipremsagar: Yup, fixed it.

python/cudf/cudf/_lib/parquet.pyx
galipremsagar: Sounds good to me 👍

python/cudf/cudf/_lib/parquet.pyx
```diff
@@ -467,6 +470,10 @@ cdef class ParquetWriter:
     max_page_size_rows: int, default 20000
         Maximum number of rows of each page of the output.
         By default, 20000 will be used.
+    force_nullable_schema : bool, default True.
```
bdice, line 473: We should be consistent in our defaults. Why is `write_parquet` using `False` but `ParquetWriter` using `True`?

This appears to be the only exception -- all other places are using `False` but this class uses `True`.

bdice (commented):
Fix the defaults, then I can approve.

python/cudf/cudf/_lib/parquet.pyx
galipremsagar: This was intentional. `ParquetWriter` is the chunked Parquet writer, which basically used to differ from the normal Parquet writer(`write_parquet`) in the following way : Chunked Parquet writer always writes every column irrespective of if they really have nulls as `null` in schema. This is done because, we would later want to read all the parquet file chunks and readers expect identical schema. This isn't the case with normal parquet writer, there `not null` and `null` were written to schema based on presence of null values in the column. 

Coming back to the defaults, having `True` in Chunked Parquet writer and `False` in normal Parquet writer as defaults preserve the behavior that has been present until now.

bdice: Okay. Should we include a comment explaining this in the docs, so readers are aware of this intended difference in behavior?

I'll approve and let you decide how to handle that.

bdice (approved):
Approving; please consider how to document the difference in intended defaults for chunked/non-chunked nullable schemas.

python/cudf/cudf/_lib/parquet.pyx
galipremsagar: Done, documented.

galipremsagar:
> Approving; please consider how to document the difference in intended defaults for chunked/non-chunked nullable schemas.

Done 👍

vuule (approved):

Issue #12932, opened 2023-03-13
[FEA] More optimization for lexicographic comparison of nested types


ttnghia:
During working in https://github.com/rapidsai/cudf/pull/12879, I did some benchmarks and found an interesting situation: For deeply nested lists-of-lists, if we convert the child column (which is also a lists column) into an integer column (through calling to `cudf::rank`) then the comparison performance can be boosted significantly. For example:
```
## [0] Quadro RTX 6000

|  size_bytes  |  depth  |  num_columns  |  null_frequency  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |             Diff |   %Diff |  Status  |
|--------------|---------|---------------|------------------|------------|-------------|------------|-------------|------------------|---------|----------|
|     2^10     |    1    |      10       |        0         |   4.075 us |     204.71% |   2.184 us |     291.37% |        -1.891 us | -46.40% |   PASS   |
|     2^18     |    1    |      10       |        0         |  32.663 ms |       3.13% |  38.994 ms |       2.32% |         6.331 ms |  19.38% |   FAIL   |
|     2^24     |    1    |      10       |        0         |   9.545 ms |       9.76% |  29.418 ms |       5.41% |        19.873 ms | 208.21% |   FAIL   |
|     2^10     |    2    |      10       |        0         |   6.744 us |      38.68% |   2.906 us |     207.01% |        -3.838 us | -56.91% |   FAIL   |
|     2^18     |    2    |      10       |        0         | 901.082 ms |       4.00% |  28.292 ms |       3.16% |   -872790.516 us | -96.86% |   FAIL   |
|     2^24     |    2    |      10       |        0         | 113.982 ms |       0.86% | 115.617 ms |       0.99% |         1.635 ms |   1.43% |   FAIL   |
|     2^10     |    3    |      10       |        0         |   9.589 us |      59.38% |   3.461 us |      85.14% |        -6.128 us | -63.91% |   FAIL   |
|     2^18     |    3    |      10       |        0         |   9.805 us |      42.68% |   3.259 us |      96.03% |        -6.546 us | -66.76% |   FAIL   |
|     2^24     |    3    |      10       |        0         |   55.376 s |        inf% | 154.392 ms |       2.17% | -55221986.520 us | -99.72% |   FAIL   |
|     2^10     |    4    |      10       |        0         |  12.853 us |      42.00% |   4.005 us |     154.35% |        -8.848 us | -68.84% |   FAIL   |
[...]

PR #12953, opened 2023-03-15
Support lists of structs in row lexicographic comparator

ttnghia:
This implements support for lexicographic comparison for lists-of-structs, following the proposed idea in https://github.com/rapidsai/cudf/issues/11222:
 * The child column of the lists-of-structs column is replaced by an integer column of its rank values. 
 * In the cases of comparing two tables, such child columns from both tables are concatenated, ranked, then split back into new child columns to replace the original child columns for each table.

Depends on:
 * https://github.com/rapidsai/cudf/pull/13005

Closes https://github.com/rapidsai/cudf/issues/11222.


cpp/src/table/row_operators.cu
```diff
@@ -155,17 +165,15 @@ table_view remove_struct_child_offsets(table_view table)
  * The list parents are still needed to define the range of elements in the leaf that belong to the
  * same row.
  *
- * In the case of structs column with a lists column as its first child such as
- * `Struct<List<int>, float>`, after decomposition we get three columns `Struct<>`,
- * `List<int>`, and `float`.
- *
  * @param table The table whose struct columns to decompose.
+ * @param decompose_lists Whether to decompose lists columns or output them unchanged
```
divyegala, line 169: Can you explain why this was needed and when it should be `true`?

cpp/src/table/row_operators.cu
```diff
@@ -321,6 +329,8 @@ void check_lex_compatibility(table_view const& input)
       check_column(list_col.child());
     } else if (c.type().id() == type_id::STRUCT) {
       for (auto child = c.child_begin(); child < c.child_end(); ++child) {
+        CUDF_EXPECTS(child->type().id() != type_id::LIST,
+                     "Cannot lexicographic compare a table with a STRUCT of LIST column");
```
divyegala, line 333:
```suggestion
                     "Cannot lexicographically compare a table with a STRUCT of LIST column");
```

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +383,467 @@ namespace row {
...
+    }
+    // else == child is not STRUCT or LIST: just go to the end of this function, no transformation.
+  }
+  // Any structs-of-lists should be decomposed into empty struct type `Struct<>` before being
+  // processed by this function.
+  else if (lhs.type().id() == type_id::STRUCT) {
+    CUDF_EXPECTS(std::all_of(lhs.child_begin(),
+                             lhs.child_end(),
+                             [](auto const& child) { return child.type().id() != type_id::LIST; }),
+                 "Structs columns should be decomposed before reaching this function.");
```
divyegala, line 552:
```suggestion
                 "Structs columns with list column descendants should be decomposed before reaching this function.");
```

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +383,467 @@ namespace row {
...
+
+  return std::any_of(input.begin(), input.end(), [&](auto const& col) {
+    // Any structs-of-lists should be decomposed into empty struct type `Struct<>` before being
+    // processed by this function.
+    if (col.type().id() == type_id::STRUCT) {
+      CUDF_EXPECTS(
+        std::all_of(col.child_begin(),
+                    col.child_end(),
+                    [](auto const& child) { return child.type().id() != type_id::LIST; }),
+        "Structs columns should be decomposed before reaching this function.");
```
divyegala, line 642:
```suggestion
        "Structs columns with list column descendants should be decomposed before reaching this function.");
```

cpp/benchmarks/sort/sort_lists.cpp
```diff
@@ -20,18 +20,82 @@
 
 #include <nvbench/nvbench.cuh>
 
-void nvbench_sort_lists(nvbench::state& state)
+namespace {
+auto constexpr min_val = 0;
+auto constexpr max_val = 100;
```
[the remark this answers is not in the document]
divyegala: Understood, thanks @bdice. I still stand by this suggestion because it's clearer to read

cpp/src/table/row_operators.cu
ttnghia: The message is correct: `decompose_struct` should be called before reaching this function, regardless of having lists column descendants or not.

ttnghia: The order can't be checked here, we can only detect wrong order when there is structs-of-lists column exists.

divyegala: > regardless of having lists column descendants or not.
But isn't the check just performing if a struct column has a list column as a direct child? This error won't be thrown if it's just a struct column

cpp/src/table/row_operators.cu
```diff
@@ -317,10 +326,12 @@ void check_lex_compatibility(table_view const& input)
     if (c.type().id() == type_id::LIST) {
       auto const& list_col = lists_column_view(c);
       CUDF_EXPECTS(list_col.child().type().id() != type_id::STRUCT,
-                   "Cannot lexicographic compare a table with a LIST of STRUCT column");
+                   "Cannot lexicographically compare a table with a LIST of STRUCT column");
```
divyegala, line 329: We can now compare list of struct column, right?

cpp/src/table/row_operators.cu
```diff
@@ -317,10 +326,12 @@ void check_lex_compatibility(table_view const& input)
...
       check_column(list_col.child());
     } else if (c.type().id() == type_id::STRUCT) {
       for (auto child = c.child_begin(); child < c.child_end(); ++child) {
+        CUDF_EXPECTS(child->type().id() != type_id::LIST,
+                     "Cannot lexicographically compare a table with a STRUCT of LIST column");
```
divyegala, line 334: And also able to compare struct of list?

cpp/src/table/row_operators.cu
divyegala: I see your point that after decomposition of `Struct<List, float>` we will have `Struct<>, List, float`. I think we need to phrase the comment better to explain why we only check for `Struct<List>`. What if there's `Struct<Struct<List>>` and it wasn't decomposed before calling this function?

cpp/src/table/row_operators.cu
```diff
@@ -133,10 +140,13 @@ table_view remove_struct_child_offsets(table_view table)
...
  *  i   f
  *
- * When list columns are present, the decomposition is performed similarly to pure structs but list
- * parent columns are NOT pruned
+ * In the case of structs column with a lists column as its first child such as
+ * `Struct<List<int>, float>`, after decomposition we get three columns `Struct<>`,
+ * `List<int>`, and `float`.
  *
- * For example, if the original table has a column `List<Struct<int, float>>`,
+ * When list columns are present, depending on the input flag `decompose_lists`, the decomposition
```
divyegala, line 147: Can you also explain how list columns are decomposed when it's `true` or `false`? `true` is used for equality comparison as you mention below, but what does the decomposition process entail in that case?

cpp/src/table/row_operators.cu
ttnghia: That's what I said above:
> The order can't be checked here, we can only detect wrong order when there is structs-of-lists column exists.

ttnghia: > `Struct<Struct<List>>`

That will be decomposed into a  table of `[Struct<Struct<>>, Struct<>, List<>]` and it will pass the check.

cpp/src/table/row_operators.cu
ttnghia: This check is to verify the intermediate result, which still can't contain lists-of-structs or structs-of-lists.

The input can be anything and will be supported all. But it will be preprocessed to not contain lists-of-structs or structs-of-lists.

bdice (changes requested):
Two structural change requests: don't use the random ID or track the use of floating point columns. (Luckily, these both make the code simpler.)

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+  rmm::cuda_stream_view stream)
+{
+  auto [decomposed_input, new_column_order, new_null_precedence, verticalized_col_depths] =
+    decompose_structs(input, false /*no decompose lists*/, column_order, null_precedence);
+
+  // Unused variables are generated for rhs table which is not available here.
+  [[maybe_unused]] auto [transformed_input, unused_0, transformed_columns, unused_1] =
+    transform_lists_of_structs(decomposed_input, std::nullopt, new_null_precedence, stream);
+
+  auto const ranked_children = transformed_columns.size() > 0;
```
bdice, line 673: I prefer the name `has_ranked_children` to indicate it's a boolean and not something that manages columns.
```suggestion
  auto const has_ranked_children = !transformed_columns.empty();
```

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+    decompose_structs(rhs, false /*no decompose lists*/, column_order, null_precedence);
+
+  // Transform any (nested) lists-of-structs column into lists-of-integers column.
+  auto [transformed_lhs, transformed_rhs_opt, transformed_columns_lhs, transformed_columns_rhs] =
+    transform_lists_of_structs(decomposed_lhs, decomposed_rhs, new_null_precedence_lhs, stream);
+
+  // This should be the same for both lhs and rhs but not all the time, such as when one table
+  // has 0 rows while the other has >0 rows. So we check separately for each of them.
+  auto const ranked_children_lhs = transformed_columns_lhs.size() > 0;
+  auto const ranked_children_rhs = transformed_columns_rhs.size() > 0;
```
bdice, line 710:
```suggestion
  auto const has_ranked_children_lhs = !transformed_columns_lhs.empty();
  auto const has_ranked_children_rhs = !transformed_columns_rhs.empty();
```

cpp/src/table/row_operators.cu
bdice, line 1: I'd like to know what happens if we _always_ enable `decompose_lists` when preprocessing structs.

What is the impact on run time? The rank kernel would always be called, so it could slow things down, but the device comparator could be faster.

Would always decomposing lists then allow us to simplify the logic of the device comparator? If so, what is _that_ impact on compile time / run time?

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @return A tuple consisting of new column_view representing the transformed input, along with
+ *         their ranks column(s) (of `size_type` type) and possibly new list offsets generated
+ *         during the transformation process
+ */
+std::tuple<column_view,
+           std::optional<column_view>,
+           std::vector<std::unique_ptr<column>>,
+           std::vector<std::unique_ptr<column>>>
+transform_lists_of_structs(column_view const& lhs,
```
bdice, line 411: I think we should consider adding an overload for single tables that skips returning the unused pieces. There is a lot of branching in this function right now for one vs. two tables. We might be able to improve the awkwardness at the call site by having two overloads for one and two tables (which would be non-optional), and pushing the implementation into a function that is structured like the current state here. Just wrapping this function in a clearer one/two table overload would be helpful, I think.

ttnghia: I thought about that too. But then didn't implement it because we may have potential optimization later (https://github.com/rapidsai/cudf/issues/12932). I plan to work on it after this PR. With that implemented, we will get rid of this branch entirely: https://github.com/rapidsai/cudf/pull/12953/files/3e4b7b2e22a332cef7b4473a69c65333b192f210#diff-c143ae2ecfe63796e604e5854f4a47cf6b831ebe5d68ce7c19a0e1533fceddc7R508-R509

cpp/src/table/row_operators.cu
bdice: I was misunderstanding what it means to decompose lists. What are the conditions in which we _should_ decompose lists?

cpp/src/table/row_operators.cu
bdice: Okay. I am fine with deferring further work on this for now.

cpp/src/table/row_operators.cu
ttnghia: Here is a better example: `input = [ [{"a", 1}, {"a", 2}], [{"a", 3} ]`.

The correct comparison is `"a"` vs `"a"`, then `1` vs `3`, then `"a"` vs `"a"`, then `2` vs `3`. However, we can shortcircuit at `1` vs `3` and return `row(0) < row(1)`.

If the lists column decomposed, we will have 2 lists columns `list0 = [ ["a", "a"], ["a"] ]` and `list1 = [ [1, 2], [3] ]`. The comparison now is comparing `["a", "a"]` vs `["a"]`, then `[1, 2]` vs `[3]`. However, we will stop after comparing `["a", "a"]` vs `["a"]` and return `row(0) > row(1)`.

ttnghia: > I was misunderstanding what it means to decompose lists. What are the conditions in which we _should_ decompose lists?

I've added a comment for that: https://github.com/rapidsai/cudf/pull/12953/files/3e4b7b2e22a332cef7b4473a69c65333b192f210#diff-c143ae2ecfe63796e604e5854f4a47cf6b831ebe5d68ce7c19a0e1533fceddc7R165-R168

> * Note that the `decompose_lists` flag should be specified as follow:
 *  - `true` when preprocessing a table for equality comparison.
 *  - `false` when preprocessing a table for lexicographic comparison, since we need to keep all
 *    lists columns intact to input into the next preprocessing step.

bdice: Why do we use `true` for equality comparisons? Would the behavior of `false` also be valid? The comment is helpful but I am not sure if I can see the reasoning for `true` in equality.

vyasr (commented):
Took a first pass through today, going to try and review this piecemeal to stop stalling progress so focused on getting a handle on the core algorithms in this review. I have a few concerns about specific aspects, but overall nothing too bad. I feel pretty confident that a couple more rounds of review will get us there.

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+    out_cols.emplace_back(
+      cudf::lists::detail::get_normalized_offsets(lists_column_view{input}, stream, default_mr));
+    return make_output(out_cols.back()->view(), new_child);
+  };
+
+  // Dense ranks should be used instead of first rank.
+  // Consider this example: `input = [ [{0, "a"}, {3, "c"}], [{0, "a"}, {2, "b"}] ]`.
+  // If first rank is used, `transformed_input = [ [0, 3], [1, 2] ]`. Comparing them will lead
+  // to the result row(0) < row(1) which is incorrect.
+  // With dense rank, `transformed_input = [ [0, 2], [0, 1] ]`, producing correct comparison.
```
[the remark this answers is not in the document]
bdice: Yes.

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -679,27 +679,89 @@ struct preprocessed_table {
...
    *
    * Sets up the table for use with lexicographical comparison. The resulting preprocessed table can
-   * be passed to the constructor of `lexicographic::self_comparator` to avoid preprocessing again.
+   * be passed to the constructor of `lexicographic::self_comparator` or
+   * `lexicographic::two_table_comparator` to avoid preprocessing again.
+   *
+   * Note that the output of this factory function should not be used in `two_table_comparator` if
+   * the input table contains lists-of-structs. In such cases, please use the overload
+   * `preprocessed_table::create(table_view const&, table_view const&,...)` to preprocess both input
+   * tables at the same time.
```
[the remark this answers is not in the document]
bdice: Agreed. I don't want to encourage independent preprocessing for two-table comparators (equality or lexicographic). Enforcing joint preprocessing at the call site gives us far more control over future changes, and we can always fall back to independent preprocessing internally.

cpp/src/table/row_operators.cu
divyegala: And the need to decompose lists for equality comparison is so that you don't get stuck in recursive dispatches, right? A struct is only compared with a struct, and so on

cpp/src/table/row_operators.cu
ttnghia: > This error is basically for the case where calling code fails to do the necessary preprocessing before calling this function, right?

Exactly.

cpp/src/table/row_operators.cu
ttnghia: Exactly. We can't do recursive dispatch in device code so we can't have lists of structs after preprocessing.

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+  // (`order::ASCENDING`), we have `transformed_input = [ [0, 2], [0, 1] ]`. Sorting of
+  // `transformed_input` will produce the same result as sorting `input` regardless of sorting
+  // order (ASC or DESC).
+  auto const compute_ranks = [&](auto const& input) {
+    return cudf::detail::rank(input,
+                              rank_method::DENSE,
+                              order::ASCENDING,
+                              null_policy::EXCLUDE,
+                              column_null_order,
+                              false /*percentage*/,
```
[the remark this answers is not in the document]
ttnghia: Maybe that boolean parameter was overlooked when reviewing the `rank` PR. Yeah of course we should have enum here instead.

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+  std::vector<std::unique_ptr<column>> out_cols_lhs;
+  std::vector<std::unique_ptr<column>> out_cols_rhs;
+
+  if (lhs.type().id() == type_id::LIST) {
+    auto const child_lhs = cudf::lists_column_view{lhs}.get_sliced_child(stream);
+
+    // Found a lists-of-structs column.
+    if (child_lhs.type().id() == type_id::STRUCT) {
+      if (rhs_opt) {  // rhs table is available
+        auto const child_rhs = cudf::lists_column_view{rhs_opt.value()}.get_sliced_child(stream);
```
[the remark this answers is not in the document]
ttnghia: That should be done before reaching to this function, by this: https://github.com/rapidsai/cudf/pull/12953/files/3e4b7b2e22a332cef7b4473a69c65333b192f210#diff-c143ae2ecfe63796e604e5854f4a47cf6b831ebe5d68ce7c19a0e1533fceddc7R690

> ```check_shape_compatibility(lhs, rhs);```

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+                         0,
+                         {offsets_cv, child_cv}};
+    };
+
+    if (input.offset() == 0) {
+      return make_output(input.child(lists_column_view::offsets_column_index), new_child);
+    }
+
+    out_cols.emplace_back(
+      cudf::lists::detail::get_normalized_offsets(lists_column_view{input}, stream, default_mr));
```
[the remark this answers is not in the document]
ttnghia: Exactly. The normalized `offsets` column here needs to be kept alive so its view can access it.

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+                std::move(out_cols_lhs),
+                std::move(out_cols_rhs)};
+      } else {  // rhs table is not available
+        out_cols_lhs.emplace_back(compute_ranks(child_lhs));
+        auto transformed_lhs = replace_child(lhs, out_cols_lhs.back()->view(), out_cols_lhs);
+
+        return {std::move(transformed_lhs),
+                std::nullopt,
+                std::move(out_cols_lhs),
+                std::move(out_cols_rhs)};
```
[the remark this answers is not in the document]
ttnghia: Bradley has smilar concern here: https://github.com/rapidsai/cudf/pull/12953/files/3e4b7b2e22a332cef7b4473a69c65333b192f210#r1176766350. I would say that I will try to simplify that again in a follow up PR.

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+  };
+
+  std::vector<std::unique_ptr<column>> out_cols_lhs;
+  std::vector<std::unique_ptr<column>> out_cols_rhs;
+
+  if (lhs.type().id() == type_id::LIST) {
+    auto const child_lhs = cudf::lists_column_view{lhs}.get_sliced_child(stream);
+
+    // Found a lists-of-structs column.
+    if (child_lhs.type().id() == type_id::STRUCT) {
```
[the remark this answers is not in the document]
ttnghia: When we call `rank()` on the child column, the input to `rank()` will be `Struct<List<Struct<>>>` as you mentioned. In the `rank()`, `sorted_order` will be called which in turn generates a preprocess table on that input. So `Struct<List<Struct<>>>` will be passed in `decompse_structs` first, producing a table `[Struct<>, List<Struct<>>]`, then columns of that table will be passed into this function: all these columns will be pass-through without any ranking.

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +385,391 @@ namespace row {
...
+    else if (child_lhs.type().id() == type_id::LIST) {
+      auto const child_rhs_opt =
+        rhs_opt
+          ? std::optional<column_view>{cudf::lists_column_view{rhs_opt.value()}.get_sliced_child(
+              stream)}
+          : std::nullopt;
+
+      // Recursively call transformation on the child column.
+      auto [new_child_lhs, new_child_rhs_opt, out_cols_child_lhs, out_cols_child_rhs] =
+        transform_lists_of_structs(child_lhs, child_rhs_opt, column_null_order, stream);
```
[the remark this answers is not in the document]
ttnghia: The original input should be passed in `decompse_struct` before flowing into this function. So basically this function cannot see any structs-of-lists column. It can only see structs-of-structs-of..... of basic types, or lists-of-lists, or lists-of-structs.

cpp/include/cudf/table/experimental/row_operators.cuh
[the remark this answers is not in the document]
ttnghia: I've just add a comment to encourage using the overload `create(lhs, rhs)`: https://github.com/rapidsai/cudf/pull/12953/commits/3a12c2d1a2cfdad3b5913971f243da61e8a4e337.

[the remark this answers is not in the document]
bdice: We should make sure that the two-table overload is being used everywhere in libcudf before merging this. I think that's in-scope for this PR -- or at least an immediate follow-up.

cpp/src/table/row_operators.cu
```diff
@@ -373,51 +391,391 @@ namespace row {
...
 two_table_comparator::two_table_comparator(table_view const& left,
                                            table_view const& right,
                                            host_span<order const> column_order,
                                            host_span<null_order const> null_precedence,
                                            rmm::cuda_stream_view stream)
-  : d_left_table{preprocessed_table::create(left, column_order, null_precedence, stream)},
-    d_right_table{preprocessed_table::create(right, column_order, null_precedence, stream)}
 {
-  check_shape_compatibility(left, right);
+  std::tie(d_left_table, d_right_table) =
```
bdice, line 777: Can you use a structured binding here instead of `std::tie`?

bdice (approved):
We've discussed a lot of follow-up work for this PR. I don't have a lot more to say about the current state, since we've deferred so much for now, so I am approving. Please make sure all the follow-up steps are organized and documented well.

cpp/src/table/row_operators.cu
ttnghia: Unfortunately, structure binding can't be used to bind the output to existing variables.

cpp/src/table/row_operators.cu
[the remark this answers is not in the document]
vyasr: Got it, thanks.

cpp/src/table/row_operators.cu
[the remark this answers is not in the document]
vyasr: Yeah I'm OK doing more simplification in a follow-up.

cpp/src/table/row_operators.cu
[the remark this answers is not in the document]
vyasr: Ah right and `decompose_structs` will find structs at any level of nesting, even inside lists, right? That's the key here is ensuring that even when there's mixed nesting every level gets processed correctly (or at least, whatever constitutes the "leaf" nodes according to that function).

cpp/tests/search/search_list_test.cpp
```diff
@@ -360,29 +371,30 @@ TEST_F(ListLowerBound, ListWithNulls)
...
     };
 
-    auto const expect = int32s_col{0};
-    auto const result = cudf::lower_bound(cudf::table_view{{haystack}},
-                                          cudf::table_view{{needles}},
-                                          {cudf::order::ASCENDING},
-                                          {cudf::null_order::BEFORE});
-    CUDF_TEST_EXPECT_COLUMNS_EQUAL(expect, *result);
+    auto const expect  = int32s_col{0};
+    auto const results = search_bounds(cudf::table_view{{haystack}},
```
vyasr, line 378: Structured binding?

vyasr: Also everywhere else using `search_bounds` in this file.

cpp/tests/search/search_list_test.cpp
vyasr, line 1: Can we add at least one test with >2 levels of nesting? I know this will be really ugly to write, but those are the cases I find hardest to reason when reviewing and those require following all the different preprocessing code paths very carefully. I would feel much more confident seeing a test that includes something stupid like a `List<List<Struct<Struct<List<Struct<int>>>>>>` (I'm sure that real-world data in Spark has this even though I would never want to write that in libcudf).

cpp/tests/sort/sort_nested_types_tests.cpp
```diff
@@ -242,3 +242,205 @@ TEST_F(NestedStructTest, StructsOfStructsHaveListsWithNulls)
...
+    auto const expected_order = int32s_col{4, 6, 1, 0, 7, 2, 5, 3};
+    auto const order =
+      cudf::stable_sorted_order(cudf::table_view{{*input}}, {cudf::order::DESCENDING});
+    CUDF_TEST_EXPECT_COLUMNS_EQUAL(expected_order, order->view());
+  }
+}
+
+TEST_F(NestedListTest, SlicedListsOfStructsNoNulls)
+{
+  auto constexpr dont_care{0};
```
vyasr, line 310: I get what you're trying to do by showing that these elements are sliced off, but tbh I was more confused by this than I would have been if you'd just put zeros in the array. I would have figured out to ignore those from the slice indices. I think this just adds confusion.

cpp/tests/table/experimental_row_operator_tests.cu
```diff
@@ -229,6 +260,148 @@ TYPED_TEST(TypedTableViewTest, TestLexicographicalComparatorSameTable)
...
+                           cudf::experimental::row::lexicographic::physical_element_comparator{},
+                           expected_rhs),
+                 cudf::logic_error);
+  };
+
+  // Generate preprocessed data for both lhs and lhs at the same time.
+  // Switching order of lhs and rhs tables then sorting them using their preprocessed data should
+  // produce exactly the same result.
+  {
+    auto const [preprocessed_lhs, preprocessed_rhs /*empty*/] =
```
vyasr, line 392: Just rename `[preprocess_]rhs` to `[preprocessed_]empty_rhs` rather than carrying around this comment everywhere.

cpp/benchmarks/sort/sort_lists.cpp
```diff
@@ -20,18 +20,82 @@
 
 #include <nvbench/nvbench.cuh>
 
-void nvbench_sort_lists(nvbench::state& state)
+namespace {
+cudf::size_type constexpr min_val = 0;
+cudf::size_type constexpr max_val = 100;
```
vyasr, line 25:
```suggestion
constexpr cudf::size_type min_val = 0;
constexpr cudf::size_type max_val = 100;
```

vyasr (changes requested):
I've stared at this enough to largely convince myself of the correctness of the algorithm at this point. I have some requests left, but they're mostly around testing or easy code quality improvements. In the interest of unblocking downstream development I am OK with merging this with more of the cleanup left as downstream work as long as it can be done in parallel with some of the feature development happening in other PRs.

cpp/include/cudf/detail/sorting.hpp
```diff
@@ -61,6 +62,20 @@ std::unique_ptr<table> sort_by_key(table_view const& values,
...
+ * @copydoc cudf::rank
+ *
+ * @param[in] stream CUDA stream used for device memory operations and kernel launches.
+ */
+std::unique_ptr<column> rank(column_view const& input,
+                             rank_method method,
+                             order column_order,
+                             null_policy null_handling,
+                             null_order null_precedence,
+                             bool percentage,
```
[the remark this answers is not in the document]
ttnghia: This is just exposing the `rank` API in `detail::` namespace. Changing this would be breaking so let's do it in some follow up PR.

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -679,27 +679,89 @@ struct preprocessed_table {
...
+   *        are sorted in ascending order.
+   * @param null_precedence Optional, an array having the same length as the number of columns in
+   *        the input tables that indicates how null values compare to all other. If it is empty,
+   *        the order `null_order::BEFORE` will be used for all columns.
+   * @param has_ranked_children Flag indicating if the input table was preprocessed to transform
+   *        any nested child column into an integer column using `cudf::rank`
+   * @param stream The stream to launch kernels and h->d copies on while preprocessing
+   * @return A shared pointer to a preprocessed table
+   */
+  static std::shared_ptr<preprocessed_table> create_preprocessed_table(
```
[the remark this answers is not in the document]
ttnghia: `create` sounds good to me. Originally I wanted to differentiate this (private) function from other public `create` functions so I named it differently.

cpp/include/cudf/table/experimental/row_operators.cuh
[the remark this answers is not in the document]
ttnghia: Unfortunately, that change is somewhat very large, requiring the `hash_join` class to be reimplemented which is too much diverged from this PR. I plan to have **two more** separate PRs for doing so instead.

cpp/src/table/row_operators.cu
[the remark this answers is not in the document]
ttnghia: > `decompose_structs` will find structs at any level of nesting, even inside lists, right?

Yes. All possible mixed nesting at any level will be processed to make sure this function will not miss any cases.

cpp/tests/sort/sort_nested_types_tests.cpp
ttnghia: Okay then I'll remove it.

cpp/tests/search/search_list_test.cpp
ttnghia: That's going to kill me :stuck_out_tongue:

cpp/tests/search/search_list_test.cpp
```diff
@@ -399,8 +409,262 @@ TEST_F(ListLowerBound, ListWithNulls)
...
+  // structs column of two input tables are concatenated, ranked, then split.
+  auto const [result_lower_bound, result_upper_bound] = search_bounds(
+    cudf::table_view{{*haystack}}, cudf::table_view{{*needles}}, {cudf::order::ASCENDING});
+  auto const expected_lower_bound = int32s_col{0, 1, 4, 9, 4, 9, 5, 4, 8};
+  auto const expected_upper_bound = int32s_col{1, 4, 4, 9, 5, 9, 8, 4, 9};
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(expected_lower_bound, *result_lower_bound, verbosity);
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(expected_upper_bound, *result_upper_bound, verbosity);
+}
+
+TEST_F(ListBinarySearch, CrazyListTest)
```
vyasr, line 555: Love the name. Thank you for writing this test! It gives me a lot more confidence that we're handling the arbitrary depth correctly.

ttnghia:
Thanks all for reviewing. I'm merging this now, so can unblock the follow-up work ASAP. The next step will be addressing
 * https://github.com/rapidsai/cudf/issues/13228
 * https://github.com/rapidsai/cudf/issues/12932

cpp/include/cudf/table/experimental/row_operators.cuh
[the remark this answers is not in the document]
ttnghia: Sorry I was wrong. When I was searching for `two_table_comparator`, I found a lot of instances and mistakenly considered all of them, including the instances of `equality::two_table_comparator`.

In fact, libcudf currently has only one instance of `lexicographic::two_table_comparator` and it doesn't need to be updated (already constructed properly). So this is no longer needed.

PR #12955, opened 2023-03-16
Correctly handle scalar indices in Index.__getitem__

### LINKED ISSUE #12954: [BUG] Indexing a non-RangeIndex with a numpy integer returns a slice, not an entry
**Describe the bug**

```python
import cudf
import numpy as np

index = cudf.Index([1, 2], dtype="int64")
value = index[np.int32(0)]
print(type(value)) # => cudf.core.index.Int64Index
value = index[0]
print(type(value)) # => np.int64
```

Contrast, pandas, where indexing with a single value gives us a scalar in both cases.

**Expected behavior**

This should work, I think.

wence-:
## Description
It is not sufficient to check for isinstance(i, int) since the index may be a numpy type for which this check is False. Instead, invert the condition and check if the return value from _get_elements_from_column is a Column, in which case we should get an Index back.

Closes #12954.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/index.py
+++ b/python/cudf/cudf/core/index.py
@@ -1403,7 +1403,7 @@ class GenericIndex(SingleColumnFrame, BaseIndex):
     @_cudf_nvtx_annotate
     def __getitem__(self, index):
         res = self._get_elements_from_column(index)
-        if not isinstance(index, int):
+        if isinstance(res, ColumnBase):
             res = as_index(res)
             res.name = self.name
         return res
--- a/python/cudf/cudf/tests/test_index.py
+++ b/python/cudf/cudf/tests/test_index.py
@@ -2886,3 +2886,20 @@ def test_index_to_pandas_nullable(data, expected_dtype):
     expected = pd.Index(data, dtype=expected_dtype)
 
     assert_eq(pi, expected)
+
+
+class TestIndexScalarGetItem:
+    @pytest.fixture(
+        params=[range(1, 10, 2), [1, 2, 3], ["a", "b", "c"], [1.5, 2.5, 3.5]]
+    )
+    def index_values(self, request):
+        return request.param
+
+    @pytest.fixture(params=[int, np.int8, np.int32, np.int64])
+    def i(self, request):
+        return request.param(1)
+
+    def test_scalar_getitem(self, index_values, i):
+        index = cudf.Index(index_values)
+
+        assert not isinstance(index[i], cudf.Index)


python/cudf/cudf/core/cut.py
```diff
@@ -279,12 +279,8 @@ def cut(
...
         if labels is not ordered and len(set(labels)) != len(labels):
             # when we have duplicate labels and ordered is False, we
-            # should allow duplicate categories. The categories are
-            # returned in order
-            new_data = [interval_labels[i][0] for i in index_labels.values]
-            return cudf.CategoricalIndex(
-                new_data, categories=sorted(set(labels)), ordered=False
-            )
+            # should allow duplicate categories.
+            return interval_labels[index_labels]
```
wence-, line 283: categories are now as the input category. This change is required because the previous code was working around the bug in `Index.__getitem__`.

python/cudf/cudf/core/index.py
```diff
@@ -1403,7 +1403,7 @@ def __repr__(self):
     @_cudf_nvtx_annotate
     def __getitem__(self, index):
         res = self._get_elements_from_column(index)
-        if not isinstance(index, int):
+        if isinstance(res, ColumnBase):
```
wence-, line 1406: Chasing the set of things that are `int`-like is always a bad idea, so instead dispatch on the complement (`_get_elements_from_column :: Scalar | Array -> Scalar | Column`)

python/cudf/cudf/tests/test_index.py
```diff
@@ -2886,3 +2886,20 @@ def test_index_to_pandas_nullable(data, expected_dtype):
...
+        return request.param
+
+    @pytest.fixture(params=[int, np.int8, np.int32, np.int64])
+    def i(self, request):
+        return request.param(1)
+
+    def test_scalar_getitem(self, index_values, i):
+        index = cudf.Index(index_values)
+
+        assert not isinstance(index[i], cudf.Index)
```
galipremsagar, line 2905:
```suggestion
        g_index = cudf.Index(index_values)
        p_index = g_index.to_pandas()

        assert not isinstance(g_index[i], cudf.Index)
        assert_eq(p_index[i], g_index[i])
```
Can we also assert the returned values?

galipremsagar (approved):
LGTM, one minor pytest comment.

python/cudf/cudf/tests/test_index.py
wence-: Done.

PR #12956, opened 2023-03-16
Include gtest in build environment.

vyasr:
## Description

Google test is currently being pulled via CPM when our packages are being built instead of using the conda package. This results in gtest (and all its CMake files) being bundled into the cudf package, which interferes with package discovery.

## Checklist

### DIFF
--- a/conda/recipes/libcudf/meta.yaml
+++ b/conda/recipes/libcudf/meta.yaml
@@ -51,6 +51,7 @@ requirements:
     - librdkafka {{ librdkafka_version }}
     - fmt {{ fmt_version }}
     - spdlog {{ spdlog_version }}
+    - gtest==1.10.0.*
 
 outputs:
   - name: libcudf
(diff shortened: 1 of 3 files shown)


conda/recipes/libcudf/meta.yaml
```diff
@@ -51,6 +51,7 @@ requirements:
     - librdkafka {{ librdkafka_version }}
     - fmt {{ fmt_version }}
     - spdlog {{ spdlog_version }}
+    - gtest==1.10.0.*
```
bdice, line 54: Use the existing variable.
https://github.com/rapidsai/cudf/blob/9ceecb1346bbdf79a3ba104a3e456cc07ce92762/conda/recipes/libcudf/conda_build_config.yaml#L16

vyasr: Ah I forgot that we had that. Thanks. Actually I think we can move both gtest and gmock out of the run requirements for libcudf-tests and into the run requirements for libcudf itself. libcudftestutil is part of libcudf, so even if someone isn't running our test suite they still need gtest if they want to use the test utils that are part of libcudf.

msadang (approved):
Version changes look good.

bdice (commented):
Looks good (reviewing after merge, requested by @vyasr).

PR #12957, opened 2023-03-16
cudftestutil supports static gtest dependencies

robertmaynard:
## Description
cudftestutil needs to be statically so that it can support gtest. So to allow this and the preload hooks we have to roll `tests/utilities/default_stream.cpp` into a separate lib.

Will correct downstream issues including: https://github.com/NVIDIA/spark-rapids-jni/issues/1006

## Checklist

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -766,6 +766,8 @@ if(CUDF_BUILD_TESTUTIL)
 
   add_library(cudf::cudftest_default_stream ALIAS cudftest_default_stream)
 
+  # Needs to be static so that we support usage of static builds of gtest which doesn't compile with
+  # fPIC enabled and therefore can't be embedded into shared libraries.
   add_library(
     cudftestutil STATIC
     tests/io/metadata_utilities.cpp
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -741,10 +741,9 @@ add_library(cudf::cudf ALIAS cudf)
 if(CUDF_BUILD_TESTUTIL)
   add_library(
     cudftest_default_stream
-    # This library when compiled as a dynamic library allows us to
-    # LD_PRELOAD injection of symbols. We currently leverage this for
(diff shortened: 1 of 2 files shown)


cpp/CMakeLists.txt
```diff
@@ -740,15 +740,38 @@ add_library(cudf::cudf ALIAS cudf)
 
 if(CUDF_BUILD_TESTUTIL)
   add_library(
```
bdice, line 742: Should we only be building this separate library if `CUDF_BUILD_STREAMS_TEST_UTIL`?

robertmaynard: We need to always build this library as the functions it defines are used by methods in `cudftestutil`

bdice: Great. @vyasr clarified a related question here, I didn't realize cudftestutil was part of libcudf (I thought it was only part of our tests). https://github.com/rapidsai/cudf/pull/12956#discussion_r1139215640

vyasr: Indeed, this part is somewhat confusing. The test util is shipped as part of libcudf so that other packages can use our test utilities without needing to load our test binaries.

robertmaynard (commented):
Updated to explain why we force static.

cpp/CMakeLists.txt
```diff
@@ -740,15 +740,37 @@ add_library(cudf::cudf ALIAS cudf)
...
+  target_link_libraries(
+    cudftest_default_stream
+    PUBLIC cudf
+    PRIVATE $<TARGET_NAME_IF_EXISTS:conda_env>
+  )
+
+  add_library(cudf::cudftest_default_stream ALIAS cudftest_default_stream)
+
+  add_library(
+    cudftestutil STATIC
```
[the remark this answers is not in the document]
robertmaynard: Done.

vyasr (approved):

raydouglass (approved):

PR #12960, opened 2023-03-16
Remove MANIFEST.in use auto-generated one for sdists and package_data for wheels

vyasr:
## Description

Using MANIFEST.in currently runs into a pretty nasty scikit-build bug (https://github.com/scikit-build/scikit-build/issues/886) that results in any file included by the manifest being copied from the install tree back into the source tree whenever an in place build occurs after an install, overwriting any local changes. We need an alternative approach to ensure that all necessary files are included in built packages. There are two types:
- sdists: scikit-build automatically generates a manifest during sdist generation if we don't provide one, and that manifest is reliably complete. It contains all files needed for a source build up to the cudf C++ code (which has always been true and is something we can come back to improving later if desired).
- wheels: The autogenerated manifest is not used during wheel generation because the manifest generation hook is not invoked during wheel builds, so to include data in the wheels we must provide the `package_data` argument to `setup`. In this case we do not need to include CMake or pyx files because the result does not need to be possible to build from, it just needs pxd files for other packages to cimport if desired.

I also reverted #12945, which was a stopgap solution to avoid this underlying problem. That change would have caused import issues inside the python/cudf directory when installing (the lack of an inplace build would have made the source tree unimportable) so this fix removes that minor limitation introduced in that PR.

## Checklist


python/cudf/setup.py
```diff
@@ -3,8 +3,9 @@
 from setuptools import find_packages
 from skbuild import setup
 
+packages = find_packages(include=["cudf*", "udf_cpp*"])
 setup(
-    include_package_data=True,
-    packages=find_packages(include=["cudf", "cudf.*"]),
+    packages=packages,
+    package_data={key: ["*.pxd", ".hpp", ".cuh"] for key in packages},
```
bdice, line 9: Why does `pxd` have a `*` but the other extensions do not?

bdice: According to docs, I think you want to change this. https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data
```suggestion
    package_data={key: ["*.pxd", "*.hpp", "*.cuh"] for key in packages},
```

vyasr: Because typos. Not intentional. Also half the reason I changed this to draft while I fixed haha.

python/cudf/setup.py
```diff
@@ -3,8 +3,9 @@
 from setuptools import find_packages
 from skbuild import setup
 
+packages = find_packages(include=["cudf*", "udf_cpp*"])
 setup(
-    include_package_data=True,
-    packages=find_packages(include=["cudf", "cudf.*"]),
+    packages=packages,
+    package_data={key: ["*.pxd", "*.hpp", "*.cuh"] for key in packages},
```
bdice, line 9: It looks like you can specify `""` as a catch-all package (though the setup.cfg and pyproject.toml forms use `"*"` as the package name... wonder if that's correct or a typo.)
```suggestion
    package_data={"": ["*.pxd", "*.hpp", "*.cuh"]},
```

vyasr: You can do this, but it doesn't work correctly due to bugs in scikit-build.

build.sh
```diff
@@ -331,10 +331,9 @@ fi
 if buildAll || hasArg cudf; then
 
     cd ${REPODIR}/python/cudf
+    python setup.py build_ext --inplace -- -DCMAKE_PREFIX_PATH=${INSTALL_PREFIX} -DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR} -DCMAKE_CUDA_ARCHITECTURES=${CUDF_CMAKE_CUDA_ARCHITECTURES} ${EXTRA_CMAKE_ARGS} -- -j${PARALLEL_LEVEL:-1}
```
bdice, line 334: Did this undo #12945?

bdice (approved):
Please verify that this doesn't break #12945 -- then feel free to merge.

build.sh
vyasr: Yes, that was intentional. The rebuild was happening because the scikit-build bug made Cython think that every file had been modified every time. I reverted the changes since the old approach is (very slightly) preferable until we move away from direct setup.py invocations entirely since it also avoids import errors from inside the source directory.

PR #12961, opened 2023-03-16
Add nvtext::minhash function

### LINKED ISSUE #12950: [QST] Efficient minhashing with cuDF
**What is your question?**
[Minhashing](https://en.wikipedia.org/wiki/MinHash) or [locality sensitive hashing](https://en.wikipedia.org/wiki/Locality_sensitive_hashing) is a popular technique used to group similar documents based on the similarity b/w words/[shingles](https://en.wikipedia.org/wiki/W-shingling) in those documents.

Right now I have an approach that computes the minhashes for documents as follows:
Series of documents ->  n-grams(shingles)-> hash_values -> groupby+min

---

Documents
```
0            doc 1
1    this is doc 2
2            doc 3
Name: doc_text, dtype: object
```
exploded n-grams
```
0     doc 
0     oc 1
1     this
1     his 
1     is i
1     s is
1     is 
1     is d
1     s do
1     doc
1     doc 
1     oc 2
3     doc 
3     oc 3
Name: doc_text, dtype: object
```
hash_values -> groupby+min
```
2    2005254014
0     216665549
1    1081029400
dtype: uint32
```

The need to explode each document into a bunch of tokens increase the memory consumption per document and adds the need to groupby document_id at the end

I was wondering if some sort of UDF or custom kernel might be more efficient for this task which can parallelize the sliding window hashing + storing minhash that would avoid this extra memory usage and the need for a groupby at the end.

cc: @davidwendt  @wence-

davidwendt:
## Description
Adds the `nvtext::minhash()` function to compute the MinHash of strings in a string column without needing extra memory to hold the intermediate substrings and hash values.
This also uses the `MurmurHash3_32` algorithm directly so the hash results have better parity with external algorithms. The API also accepts a hash seed value.

```
std::unique_ptr<cudf::column> minhash(
  cudf::strings_column_view const& input,
  cudf::size_type width               = 4,
  cudf::hash_value_type seed          = cudf::DEFAULT_HASH_SEED,
```

Closes #12950 

## Checklist


cpp/src/text/minhash.cu
```diff
@@ -0,0 +1,104 @@
...
+  {
+    if (d_strings.is_null(idx)) return 0;
+    auto const d_str = d_strings.element<cudf::string_view>(idx);
+
+    auto mh = cudf::hash_value_type{0};
+    for (cudf::size_type pos = 0; pos < d_str.length() - (width - 1); ++pos) {
+      auto const ss     = d_str.substr(pos, width);
+      auto const hasher = cudf::detail::MurmurHash3_32<cudf::string_view>{seed};
+      auto const hvalue = hasher(ss);
+      // cudf::detail::hash_combine(seed, hasher(ss)); matches cudf::hash() result
```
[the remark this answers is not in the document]
davidwendt: No. It is also considered an adequate result.

cpp/include/nvtext/minhash.hpp
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Hash values are computed from substrings of each string and the
+ * minimum hash value is returned for each string.
+ *
+ * All null row entries are ignored and the output contains all valid rows.
+ *
+ * @param input Strings column to compute minhash
+ * @param width The character width used for apply substrings;
+ *              Any string smaller than this width will not be hashed.
+ *              Default is 4 characters.
```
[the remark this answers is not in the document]
davidwendt: This is a draft PR. I'm not really expecting reviews right now but welcome your comments and questions.
Overall, I'm waiting for feedback from upstream usage since there are number of assumptions like this that I want input on from that team.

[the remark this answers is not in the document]
wence-: Apologies, I was a bit eager (am also working with this team).

ayushdg:
Thanks for the initial PR @davidwendt. I got a chance to test this out, and it's b/w 2-5x faster that the workaround described in #12950 .

What I didn't mention in the original issue (and I've only added this https://github.com/rapidsai/cudf/issues/12950#issuecomment-1481676168) is that it's common to compute multiple minhashes usually ranging from 100 to 20k. Calling `minhash` multiple times where each call is a pass over the string becomes slower in practice. 
Would it be possible to modify the kernel to instead accept `seeds` a cuDF Series, or host array. and with a single pass compute all `k` minhashes using these different `seed` values?

cpp/include/nvtext/minhash.hpp
```diff
@@ -0,0 +1,92 @@
...
+ * @param h_id  Hash algorithm to use;
+ *              Only HASH_MURMUR3 is currently supported.
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ * @return Minhash values for each string in input
+ */
+std::unique_ptr<cudf::column> minhash(
+  cudf::strings_column_view const& input,
+  cudf::numeric_scalar<cudf::hash_value_type> seed = cudf::numeric_scalar(cudf::DEFAULT_HASH_SEED),
+  cudf::size_type width                            = 4,
+  cudf::hash_id h_id                               = cudf::hash_id::HASH_MURMUR3,
```
bdice, line 56: This is called `hash_function` elsewhere in cudf. I don't have strong feelings but aligning might be nice.

cpp/src/text/minhash.cu
```diff
@@ -0,0 +1,187 @@
...
+    __syncwarp();
+
+    auto const begin = d_str.begin() + lane_idx;
+    auto const end   = [d_str, width = width] {
+      auto const length = d_str.length();
+      if (length > width) { return (d_str.end() - (width - 1)); }
+      return d_str.begin() + static_cast<cudf::size_type>(length > 0);
+    }();
+
+    // each lane hashes substrings of parts of the string
```
bdice, line 84: "Substrings of parts of the string" sounds redundant. I think this is clearer:
```suggestion
    // each lane hashes substrings of the given width
```

cpp/src/text/minhash.cu
```diff
@@ -0,0 +1,187 @@
...
+    for (auto itr = begin; itr < end; itr += cudf::detail::warp_size) {
+      auto const offset = itr.byte_offset();
+      auto const hash_str =
+        cudf::string_view(d_str.data() + offset, (itr + width).byte_offset() - offset);
+
+      // hashing each seed on the same section of string is 10x faster than
+      // re-substringing (my new word) for each seed
```
bdice, line 91: 😉 I think this is clear enough. You could alternatively write "Using substrings in the outer loop and hash seeds in the inner loop is 10x faster than the other way around."
```suggestion
      // computing the substrings for each seed
```

cpp/src/text/minhash.cu
```diff
@@ -0,0 +1,187 @@
...
+std::unique_ptr<cudf::column> minhash(cudf::strings_column_view const& input,
+                                      cudf::device_span<cudf::hash_value_type const> seeds,
+                                      cudf::size_type width,
+                                      cudf::hash_id h_id,
+                                      rmm::cuda_stream_view stream,
+                                      rmm::mr::device_memory_resource* mr)
+{
+  CUDF_EXPECTS(!seeds.empty(), "Parameter seeds cannot be empty", std::invalid_argument);
+  CUDF_EXPECTS(
+    width > 1, "Parameter width should be an integer value of 2 or greater", std::invalid_argument);
```
bdice, line 112: I like when the logical check matches the wording of the error. I'd suggest:
```suggestion
    width >= 2, "Parameter width should be an integer value of 2 or greater", std::invalid_argument);
```

bdice (commented):
Reviewed C++ (not Python yet). Great work! This is a cute algorithm/application.

cpp/src/text/minhash.cu
davidwendt: Some of these comments are just for you Bradley.

bdice: I appreciate it! It made me smile. 😊

wence- (changes requested):
Minor quibbles with the cudf API only.

python/cudf/cudf/core/column/string.py
```diff
@@ -5226,6 +5226,48 @@ def edit_distance_matrix(self) -> SeriesOrIndex:
             libstrings.edit_distance_matrix(self._column)
         )
 
+    def minhash(
+        self, seeds=None, n: int = 4, method: str = "murmur3"
```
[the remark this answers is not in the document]
davidwendt: Sorry, I don't know what that is or how to use it.
I don't see any usage of `Optional[Series]` parameter types in the current code base.
Is it important to be used here? Maybe someone with more Python cudf knowledge can weigh in.

python/cudf/cudf/core/column/string.py
```diff
@@ -5226,6 +5226,48 @@ def edit_distance_matrix(self) -> SeriesOrIndex:
...
+        >>> seeds = cudf.Series([0, 1, 2], dtype=np.uint32)
+        >>> str_series.str.minhash(seeds)
+        0    [21141582, 403093213, 1258052021]
+        1    [962346254, 677440381, 122618762]
+        dtype: list
+        """
+        if seeds is None:
+            seeds = column.as_column(0, dtype=np.uint32, length=1)
+        else:
+            seeds = seeds._column
```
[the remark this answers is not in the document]
davidwendt: Happy to add the check. Is there an example I could follow that does a check like this?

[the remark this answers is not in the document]
wence-: Perhaps this:
```python
        if seeds is None:
            seeds_column = column.as_column(0, dtype=np.uint32, length=1)
        elif isinstance(seeds, cudf.Series) and seeds.dtype == np.uint32:
            seeds_column = seeds._column
        else:
            raise ValueError(
                f"Expecting a Series with dtype uint32, got {type(seeds)}"
            )
        return self._return_or_inplace(
            libstrings.minhash(self._column, seeds_column, n, method)
        )
```

The renaming of seeds_column goes along with the proposal to add a type annotation to the `seeds` argument above (the type checkers complain if you try and assign something of the wrong type to a typed variable).

python/cudf/cudf/core/column/string.py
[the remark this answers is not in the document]
wence-: Ah apologies, like this:
```suggestion
        self,
        seeds: Optional["cudf.Series"] = None,
        n: int = 4,
        method: str = "murmur3",
```

This provides a type annotation for the `seeds` argument that matches up with what the function accepts (`Optional[X]` for some type `X` means "Either `None` or an `X`").

Given that you've added types to the other arguments, we may as well for the first one.

cpp/src/text/minhash.cu
```diff
@@ -0,0 +1,188 @@
...
+  auto hashes   = cudf::make_numeric_column(output_type,
+                                          input.size() * static_cast<cudf::size_type>(seeds.size()),
+                                          cudf::mask_state::UNALLOCATED,
+                                          stream,
+                                          mr);
+  auto d_hashes = hashes->mutable_view().data<cudf::hash_value_type>();
+
+  thrust::for_each_n(rmm::exec_policy(stream),
+                     thrust::counting_iterator<cudf::size_type>(0),
+                     input.size() * cudf::detail::warp_size,
```
wence-, line 132: Similar query about signed overflow here.

cpp/src/text/minhash.cu
```diff
@@ -0,0 +1,188 @@
...
+               "Only murmur3 hash algorithm supported",
+               std::invalid_argument);
+
+  auto output_type = cudf::data_type{cudf::type_to_id<cudf::hash_value_type>()};
+  if (input.is_empty()) { return cudf::make_empty_column(output_type); }
+
+  auto const d_strings = cudf::column_device_view::create(input.parent(), stream);
+
+  auto hashes   = cudf::make_numeric_column(output_type,
+                                          input.size() * static_cast<cudf::size_type>(seeds.size()),
```
wence-, line 124: Could this multiplication overflow, and do we care (either here, or in the python wrapper)?

python/cudf/cudf/core/column/string.py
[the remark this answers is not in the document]
davidwendt: I gave this a try and it seems to work.

cpp/src/text/minhash.cu
davidwendt: Yes. You are correct. This could technically overflow. 
This would mean we are given a large number of very small strings which is not very likely here but certainly possible. 
I will add a check for this.

davidwendt: Yes. Added a check for this and a gtest.

python/cudf/cudf/core/column/string.py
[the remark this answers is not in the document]
davidwendt: Thanks. Changed the code to match this.

cpp/src/text/minhash.cu
davidwendt: This one is different. This could definitely happen for a large number of smallish strings which is not expected for this API but certainly possible. Changed this to use `std::size_t` for the indices instead.

wence- (approved):
Thanks! All looks good to me.

python/cudf/cudf/core/column/string.py
```diff
@@ -5226,6 +5226,55 @@ def edit_distance_matrix(self) -> SeriesOrIndex:
...
+            Only 'murmur3' (MurmurHash3_32) is supported.
+            Default is 'murmur3'.
+
+        Examples
+        --------
+        >>> import cudf
+        >>> str_series = cudf.Series(['this is my', 'favorite book'])
+        >>> seeds = cudf.Series([0], dtype=np.uint32)
+        >>> str_series.str.minhash(seeds)
+        0     21141582
```
bdice, line 5257: I must be missing something -- is the list unwrapped into a numerical column if there is only one seed? Is that being done in C++ or Python? Is that documented? I glanced over the implementation again and didn't see it.

cpp/include/doxygen_groups.h
```diff
@@ -164,6 +164,7 @@
  *   @defgroup nvtext_edit_distance Edit Distance
  *   @defgroup nvtext_tokenize Tokenizing
  *   @defgroup nvtext_replace Replacing
+ *   @defgroup nvtext_minhash MinHashing
```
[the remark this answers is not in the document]
davidwendt: No, these are not intended to be alphabetized.

cpp/include/nvtext/minhash.hpp
```diff
@@ -0,0 +1,92 @@
...
+ * Any null row entries result in corresponding null output rows.
+ *
+ * @throw std::invalid_argument if the width < 2
+ * @throw std::invalid_argument if h_id is not HASH_MURMUR3
+ * @throw std::invalid_argument if seeds is empty
+ *
+ * @param input Strings column to compute minhash
+ * @param seeds Seed values used for the MurmurHash3_32 algorithm
+ * @param width The character width used for apply substrings;
+ *              Any string smaller than this width will not be hashed.
```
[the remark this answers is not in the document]
davidwendt: Oops. That is from a previous iteration and should be removed.

python/cudf/cudf/core/column/string.py
davidwendt: This happens in the libcudf API. I will update the doxygen.

bdice (approved):

cpp/include/nvtext/minhash.hpp
```diff
@@ -0,0 +1,91 @@
...
+ *              Default is 4 characters.
+ * @param hash_function Hash algorithm to use;
+ *                      Only HASH_MURMUR3 is currently supported.
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ * @return List column of minhash values for each string per seed
+ *         or a hash_value_type column if only a single seed is specified
+ */
+std::unique_ptr<cudf::column> minhash(
+  cudf::strings_column_view const& input,
+  cudf::device_span<cudf::hash_value_type const> seeds,
```
karthikeyann, line 85: In Cython, `seeds` is column_view, here it is device_span. 
Is this intentional? how is this working?

cpp/benchmarks/CMakeLists.txt
```diff
@@ -290,7 +292,6 @@ ConfigureBench(
   string/factory.cu
   string/filter.cpp
   string/find.cpp
-  string/like.cpp
```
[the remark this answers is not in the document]
davidwendt: Nope. It should not have been there. The `string/like.cpp` is part of the `STRINGS_NVBENCH` make.

cpp/include/nvtext/minhash.hpp
davidwendt: Yes, this is intentional. A `column_view` can be implicitly be cast to a `device_span<T const>`.

karthikeyann: got it. because of `operator device_span<T const>()` in column_view.

karthikeyann (approved):

ayushdg (approved):

PR #12963, opened 2023-03-17
Reduce peak memory use when writing compressed ORC files.

vuule:
## Description
This PR changes how the buffer for encoded data is allocated in the ORC writer. Instead of a single buffer for the whole table, each stream of each stripe is allocated separately. 
Since size of the encoded data is not known in advance, buffers are over sized in most cases (decimal types and dictionary encoded data being the exceptions). Resizing these buffers to the exact encoded data size before compression reduces peak memory usage.
The resizing of the encoded buffers is done in the step where row groups are gathered to make contiguous encoded stripe in memory. This way we don't incur additional copies (compared to previous approach to `gather_stripes`).

Other changes:
Removed `compute_offsets` because it is not needed with separate buffers for each stripe/stream.
Refactored parts of `encode_columns` to initialize data buffers + stream descriptors one stripe at a time, allowing future separation into per-stripe processing (for e.g. pipelining).

Impact: internal benchmarks show average reduction of peak memory use of 14% when SNAPPY compression is enabled, with minimal impact on performance.
cub::DeviceMemcpy::Batched can now be used in the ORC writer.

## Checklist


vuule:
This does increase memory usage when compression is not used.
Overview:
With SNAPPY, average peak memory use goes from 1.66GB to 1.43GB. :+1: 
Without compression, average peak memory use goes from 867MB to 1GB. :-1:

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1088,45 +1088,31 @@ __global__ void __launch_bounds__(1024)
...
       }
+
+      __syncthreads();
+      if (t == 0) { streams[ss.column_id][group].data_ptrs[cid] = dst_ptr; }
+      dst_ptr += len;
     }
-    dst_ptr += len;
-    __syncthreads();
   }
-  if (!t) { strm_desc[stripe_id][stream_id].stream_size = dst_ptr - strm0.data_ptrs[cid]; }
```
vuule, line 1129: no need to set the stream size, its been computed on the host

cpp/src/io/orc/writer_impl.cu
```diff
@@ -1104,75 +1060,81 @@ encoded_data encode_columns(orc_table_view const& orc_table,
...
-              strm.data_ptrs[strm_type] =
-                (rg_idx == 0) ? (encoded_data.data() + stream_offsets.non_rle_data_size +
-                                 stream_offsets.offsets[strm_id])
-                              : (col_streams[rg_idx - 1].data_ptrs[strm_type] +
-                                 col_streams[rg_idx - 1].lengths[strm_type]);
             }
-          } else {
-            strm.lengths[strm_type]   = 0;
+            // Allow extra space for alignment
+            stripe_size += strm.lengths[strm_type] + uncomp_block_align - 1;
```
vuule, line 1113: This is a potential bug fix for extreme corner cases where alignment can push the writing of encoded data into the next stream.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -1201,11 +1163,12 @@ encoded_data encode_columns(orc_table_view const& orc_table,
   }
   dictionaries.data.clear();
   dictionaries.index.clear();
-  stream.synchronize();
+  chunk_streams.device_to_host(stream, true);
 
   return {std::move(encoded_data), std::move(chunk_streams)};
 }
 
+// TODO: remove StripeInformation from this function and return strm_desc instead
```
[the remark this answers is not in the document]
vuule: The change is not directly related to this PR so I didn't want to include it. It's meant to be a note for future work.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -1104,75 +1060,81 @@ encoded_data encode_columns(orc_table_view const& orc_table,
...
-  for (size_t col_idx = 0; col_idx < num_columns; col_idx++) {
-    auto const& column = orc_table.column(col_idx);
-    auto col_streams   = chunk_streams[col_idx];
-    for (auto const& stripe : segmentation.stripes) {
-      for (auto rg_idx_it = stripe.cbegin(); rg_idx_it < stripe.cend(); ++rg_idx_it) {
-        auto const rg_idx = *rg_idx_it;
-        auto const& ck    = chunks[col_idx][rg_idx];
-        auto& strm        = col_streams[rg_idx];
+  // per-stripe, per-stream owning buffers
+  std::vector<std::vector<rmm::device_uvector<uint8_t>>> encoded_data(segmentation.num_stripes());
```
PointKernel, line 1064: Not in this PR, but this reminds me that we could extract the owning buffer strong type here (https://github.com/rapidsai/cudf/blob/e37bddb84be7a3a6915511702739f5b7e87386d7/cpp/include/cudf/io/datasource.hpp#L352) and apply it across cuIO code. Thus the above code would be a 2D vector of `owning_buffer`. This also helps unify data type by using `std::byte` consistently.

Also, owning buffer could be enforced via unique ptr.

PointKernel (approved):
The new logic is much simpler. Only some non-blocking suggestions.

Thanks for the work!

cpp/src/io/orc/orc_gpu.hpp
```diff
@@ -159,6 +159,7 @@ struct encoder_chunk_streams {
  * @brief Struct to describe a column stream within a stripe
  */
 struct StripeStream {
+  uint8_t* data_ptr;        // encoded and gathered output
```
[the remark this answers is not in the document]
ttnghia: I think refactor to `std::byte` should be done more extensively in a separate PR, as `uint8_t` is used in many other places in ORC writer code.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -1201,41 +1163,67 @@ encoded_data encode_columns(orc_table_view const& orc_table,
   }
   dictionaries.data.clear();
   dictionaries.index.clear();
-  stream.synchronize();
+  chunk_streams.device_to_host(stream, true);
```
ttnghia, line 1166: Nit: Whenever passing literal value, please always try to add comment saying what it is. Otherwise others may not know what is it.
```suggestion
  chunk_streams.device_to_host(stream, true /* device sync */);
```

cpp/src/io/orc/writer_impl.cu
```diff
@@ -1201,41 +1163,67 @@ encoded_data encode_columns(orc_table_view const& orc_table,
...
-std::vector<StripeInformation> gather_stripes(
-  size_t num_index_streams,
-  file_segmentation const& segmentation,
-  hostdevice_2dvector<gpu::encoder_chunk_streams>* enc_streams,
-  hostdevice_2dvector<gpu::StripeStream>* strm_desc,
-  rmm::cuda_stream_view stream)
+std::vector<StripeInformation> gather_stripes(size_t num_index_streams,
+                                              file_segmentation const& segmentation,
+                                              encoded_data* enc_data,
+                                              hostdevice_2dvector<gpu::StripeStream>* strm_desc,
```
ttnghia, line 1186: Nit: For non-iterator parameter, try to avoid pointer.
```suggestion
                                              hostdevice_2dvector<gpu::StripeStream>& strm_desc,
```

ttnghia (approved):

cpp/src/io/orc/writer_impl.cu
vuule: We used a pointer here to make it obvious at the call site that the parameter will the modified. I personally prefer references for non-optional parameters.

cpp/src/io/orc/writer_impl.cu
vuule: I don't think I get this one. `rmm::device_uvector` already implies an owning buffer. We know that this type deallocates the memory in the destructor.
In the datasource::buffer we have the distinction between owning and non-owning buffers because we use the abstract buffer type, which does not imply ownership, only the access API to the contained data.

cpp/src/io/orc/orc_gpu.hpp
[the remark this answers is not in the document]
vuule: I agree with @ttnghia here, I'd like to do it in a separate PR so it can be done more systematically. The new implementation here is really just a replacement which has to interact with the rest of the writer (that consistently uses uint8_t).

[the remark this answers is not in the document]
PointKernel: Works for me

cpp/src/io/orc/writer_impl.cu
PointKernel: > rmm::device_uvector already implies an owning buffer

My intention was to use a strong type like:
```cpp
struct owning_buffer{
  ...
  std::unique_ptr<rmm::device_uvector<uint8_t>> data;
};
```
to explicitly represent owning buffers in cuIO.

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1088,45 +1088,31 @@ __global__ void __launch_bounds__(1024)
...
-      for (uint32_t i = 0; i < len; i += 1024) {
-        uint8_t v = (i + t < len) ? src_ptr[i + t] : 0;
-        __syncthreads();
-        if (i + t < len) { dst_ptr[i + t] = v; }
+  auto dst_ptr   = ss.data_ptr;
+  for (auto group = ss.first_chunk_id; group < ss.first_chunk_id + ss.num_chunks; ++group) {
+    auto const len = streams[ss.column_id][group].lengths[cid];
+    if (len > 0) {
+      auto const src_ptr = streams[ss.column_id][group].data_ptrs[cid];
+      for (uint32_t i = t; i < len; i += 1024) {
```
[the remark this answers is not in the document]
vuule: Done, plus replaced the block size magic number with a constant. There's many more left in the file, though 😬

vuule:
Ran unit tests with cuda-memcheck on this branch, no errors found 👍

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1082,51 +1083,37 @@ __global__ void __launch_bounds__(block_size)
  * @param[in,out] strm_desc StripeStream device array [stripe][stream]
  * @param[in,out] streams List of encoder chunk streams [column][rowgroup]
  */
-// blockDim {1024,1,1}
-__global__ void __launch_bounds__(1024)
+// blockDim {compact_streams_block_size,1,1}
+__global__ void __launch_bounds__(compact_streams_block_size)
```
hyperbolic2346, line 1087: 🔥

cpp/src/io/orc/writer_impl.cu
hyperbolic2346: Potential bug as in we haven't seen this before in practice? Do we have a test for it? Should we?

hyperbolic2346 (approved):
Looks good to me. Have a question about the bug fix, but if we want to do something there I would suggest a new issue.

cpp/src/io/orc/writer_impl.cu
vuule: It would be nice to have this test, but it's not trivial to come up with the failing input.
Maybe a decimal column, since we use the exact size (and it doesn't have to be a multiple of 4, unlike floats)? I'll look into this, just not for this PR :)

PR #12972, opened 2023-03-17
Add nunique aggregation support for cudf::segmented_reduce

davidwendt:
## Description
Adds support for `NUNIQUE` aggregation type for `cudf::segmented_reduce`. This computes the number of unique elements within each segment specified. Due to the overhead of sorting, the segments must be sorted before calling this function otherwise the results are undefined. Also, only non-nested column types are supported as well.

Reference #10432 

## Checklist


cpp/include/cudf/reduction/detail/segmented_reduction_functions.hpp
```diff
@@ -325,6 +325,34 @@ std::unique_ptr<column> segmented_variance(column_view const& col,
...
+ * @brief Counts the number of unique values within each segment of a column
+ *
+ * Unique entries are counted by comparing adjacent values so the column segments
+ * are expected to be sorted before calling this function otherwise the results
+ * are undefined.
+ *
+ * If input segment is empty, the segment result is null.
+ *
+ * If `null_handling==null_policy::INCLUDE`, all elements in a segment must be valid
+ * for the reduced value to be valid.
```
karthikeyann, line 338: is this expected behaviour for `null_handling==null_policy::INCLUDE`
Or Should it count null as unique value too?

cpp/src/reductions/segmented/nunique.cu
```diff
@@ -0,0 +1,110 @@
...
+                                                null_policy null_handling,
+                                                rmm::cuda_stream_view stream,
+                                                rmm::mr::device_memory_resource* mr)
+{
+  // only support non-nested types
+  CUDF_EXPECTS(!cudf::is_nested(col.type()),
+               "segmented reduce nunique only supports non-nested column types");
+
+  // compute the unique identifiers within each segment
+  auto identifiers = [&] {
```
karthikeyann, line 64:
```suggestion
  auto const identifiers = [&] {
```

cpp/src/reductions/segmented/nunique.cu
```diff
@@ -0,0 +1,110 @@
...
+    auto const d_col = column_device_view::create(col, stream);
+    auto const comparator =
+      cudf::experimental::row::equality::self_comparator{table_view({col}), stream};
+    auto const row_equal =
+      comparator.equal_to<false>(cudf::nullate::DYNAMIC{false}, null_equality::EQUAL);
```
karthikeyann, line 69: clarification: is `cudf::nullate::DYNAMIC{false}` right usage here?

karthikeyann (changes requested):

cpp/src/reductions/segmented/nunique.cu
davidwendt: You are right. Good catch.

cpp/include/cudf/reduction/detail/segmented_reduction_functions.hpp
divyegala: I believe `unique/unique_count` counts all `NULLs` as a single unique value in this case

cpp/src/reductions/segmented/nunique.cu
```diff
@@ -0,0 +1,110 @@
...
+                                                rmm::cuda_stream_view stream,
+                                                rmm::mr::device_memory_resource* mr)
+{
+  // only support non-nested types
+  CUDF_EXPECTS(!cudf::is_nested(col.type()),
+               "segmented reduce nunique only supports non-nested column types");
+
+  // compute the unique identifiers within each segment
+  auto identifiers = [&] {
+    auto const d_col = column_device_view::create(col, stream);
```
divyegala, line 65: Is this needed anymore?

davidwendt: The `d_col` is needed only to build the `identifiers`. It is just a view so no actual data is created/destroyed. Everything we need is in the returned device-uvector.

cpp/include/cudf/reduction/detail/segmented_reduction_functions.hpp
davidwendt: Updated so that `null_policy::INCLUDE` will count a null entry as a unique value in the segment.

divyegala (approved):

bdice (approved):
Approving CMake.

vyasr (approved):
CMake approval

PR #12981, opened 2023-03-21
Adding `hostdevice_span` that is a span createable from `hostdevice_vector`

hyperbolic2346:
## Description
I ran into a need for a span-like view into a `hostdevice_vector`. I was chopping it up into pieces to pass into a function to process portions at a time, but it still wanted to do things like host to device on the spans. This class is a result of that need.

## Checklist


cpp/src/io/utilities/hostdevice_span.hpp
```diff
@@ -0,0 +1,145 @@
...
+  [[nodiscard]] size_t size() const noexcept { return _size; }
+
+  /**
+   * @brief Returns true if `size()` returns zero, or false otherwise
+   *
+   * @return True if `size()` returns zero, or false otherwise
+   */
+  [[nodiscard]] bool is_empty() const noexcept { return size() == 0; }
+
+  [[nodiscard]] size_t memory_size() const noexcept { return sizeof(T) * size(); }
```
ttnghia, line 122: Recommend to use the convention name `size_bytes`.

hyperbolic2346: I am matching the `hostdevice_vector` interface. I would be happy to change this, but I would like to change `hostdevice_vector` as well. Maybe this is best in a follow up issue?

ttnghia (approved):

cpp/src/io/utilities/hostdevice_vector.hpp
```diff
@@ -150,6 +151,27 @@ class hostdevice_vector {
...
+  {
+    return hostdevice_span<T>{host_data, d_data.data(), size()};
+  }
+
+  /**
+   * @brief Converts a part of a hostdevice_vector into a hostdevice_span.
+   *
+   * @return A typed hostdevice_span of the hostdevice_vector's data.
+   */
+  [[nodiscard]] hostdevice_span<T> slice(size_t starting_idx, size_t ending_idx)
```
vuule, line 169: we should verify inputs here
starting_idx <= ending_idx, ending_idx < size()

cpp/src/io/utilities/hostdevice_span.hpp
vuule: We can justify matching the `host_span`/`device_span` interface instead, if that helps :)

vuule (changes requested):
Few small suggestions.
I see the checkmark in the description, but no tests :)
Do we punt on tests for now?

hyperbolic2346:
> I see the checkmark in the description, but no tests :) Do we punt on tests for now?

I was working on those and just didn't have them up yet....promise. 😆

vuule (commented):
few small suggestions

ttnghia (approved):

vuule (approved):
🔥

PR #12988, opened 2023-03-21
Fix `sort_values` when column is all empty strings

eriknw:
## Description

See test for simple MRE.

This fixes https://github.com/rapidsai/cugraph/issues/3058

## Checklist

### DIFF
--- a/python/dask_cudf/dask_cudf/sorting.py
+++ b/python/dask_cudf/dask_cudf/sorting.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
 
 from collections.abc import Iterator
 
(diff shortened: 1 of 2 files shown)


python/dask_cudf/dask_cudf/sorting.py
```diff
@@ -218,9 +218,10 @@ def quantile_divisions(df, by, npartitions):
                 divisions[col].iloc[-1] += 1
                 divisions[col] = divisions[col].astype(dtype)
             else:
-                divisions[col].iloc[-1] = chr(
-                    ord(divisions[col].iloc[-1][0]) + 1
-                )
+                if last := divisions[col].iloc[-1]:
+                    divisions[col].iloc[-1] = chr(ord(last[0]) + 1)
+                else:
+                    divisions[col].iloc[-1] = chr(1)  # b/c "" < chr(1)
```
eriknw, line 224: Interestingly, using `chr(0)` here doesn't work, because `drop_duplicates` below drops it by treating `""` and `chr(0)` the same, even though `"" < chr(0)`. Do we care that `chr(1)` is not printable?

wence-: #12990.

eriknw: > it suffices to provide any non-empty string

Yup. How about `"wence was here"`?

wence-: I like David's suggestion (or we can wait for #12991 and use `chr(0)`)

wence- (approved):

Issue #12986, opened 2023-03-21
[DOC] Groupby apply JIT engine documentation example rendering incorrectly


beckernick:
The groupby.apply(..., engine="jit") [docs example](https://docs.rapids.ai/api/cudf/nightly/api_docs/api/cudf.core.groupby.groupby.groupby.apply) is not rendering correctly as code.

Perhaps due to the [docstring not using a codeblock](https://github.com/rapidsai/cudf/blob/branch-23.04/python/cudf/cudf/core/groupby/groupby.py#L1148-L1160) literal?

brandon-b-miller:
Taking a look at this - thanks!

PR #12994, opened 2023-03-22
Fix `GroupBy.apply` doc examples rendering

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/12986.

### DIFF
--- a/python/cudf/cudf/core/groupby/groupby.py
+++ b/python/cudf/cudf/core/groupby/groupby.py
@@ -1149,18 +1149,16 @@ class GroupBy(Serializable, Reducible, Scannable):
         initially those that contain reductions and arithmetic operations
         between results of those reductions:
 
-        ..code-block:: python
-
-          import cudf
-          df = cudf.DataFrame({'a':[1,1,2,2,3,3], 'b':[1,2,3,4,5,6]})
-          df.groupby('a').apply(
-              lambda group: group['b'].max() - group['b'].min(),
-              engine='jit'
-          )
-          a  None
-          0  1     1
-          1  2     1
-          2  3     1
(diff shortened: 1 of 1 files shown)


python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -1148,16 +1148,19 @@ def mult(df):
...
+          import cudf
+          df = cudf.DataFrame({'a':[1,1,2,2,3,3], 'b':[1,2,3,4,5,6]})
+          df.groupby('a').apply(
+              lambda group: group['b'].max() - group['b'].min(),
+              engine='jit'
+          )
+          a  None
+          0  1     1
+          1  2     1
+          2  3     1
```
galipremsagar, line 1163: After rendering, this format seems to be broken too, so we either can do this:

```suggestion

        >>> import cudf
        >>> df = cudf.DataFrame({'a':[1,1,2,2,3,3], 'b':[1,2,3,4,5,6]})
        >>> df.groupby('a').apply(
        ...   lambda group: group['b'].max() - group['b'].min(),
        ...   engine='jit'
        ... )
           a  None
        0  1     1
        1  2     1
        2  3     1
```

or this:

```suggestion

        .. code-block:: python
          import cudf
          df = cudf.DataFrame({'a':[1,1,2,2,3,3], 'b':[1,2,3,4,5,6]})
          df.groupby('a').apply(
              lambda group: group['b'].max() - group['b'].min(),
              engine='jit'
          )
          a  None
          0  1     1
          1  2     1
          2  3     1
```

galipremsagar (changes requested):

PR #12995, opened 2023-03-22
Fix gtest column utility comparator diff reporting

davidwendt:
## Description
Fixes a bug introduced in #12777 (by me) in `column_utilities.cu` that caused the difference reporting in a failed comparison in a gtest result (i.e. through `CUDF_TEST_EXPECT_COLUMNS_EQUIVALENT` for example) to report the incorrect row.
The logic was changed to improve compile time but incorrectly created indices of only 0s and 1s in the difference vector.
This PR fixes the logic to create the correct indices for the reporting logic.

## Checklist


ttnghia (approved):

ttnghia:
Just hit tested failures with lists column, caused by the issue that is fixed by this :+1: :
```
Failed
first difference: lhs[4] = 0, rhs[4] = 0
```

divyegala (approved):

Issue #12646, opened 2023-01-30
[BUG] Broadast of scalar inputs to the same lengths as other columns in DataFrame constructor


shwina:
In the constructor of `cudf.DataFrame`, when one or more input "columns" are provided as scalars, they are broadcast to the same length as other input columns:

```python
df = cudf.DataFrame({'a': [1, 2, 3], 'b': ['x', 'y', 'z'], 'c': 4})
print(df)
```

Output:

```
   a  b  c
0  1  x  4
1  2  y  4
2  3  z  4
```

However, if the scalar appears as the _first_ value in the input, cuDF complains about a length mismatch:

```python
df = cudf.DataFrame({'c': 4, 'a': [1, 2, 3], 'b': ['x', 'y', 'z']})  # ValueError
```

PR #12997, opened 2023-03-23
[REVIEW] Fix `DataFrame` constructor to broadcast scalar inputs properly

galipremsagar:
## Description
Fixes: #12646

This PR fixes an issue with `DataFrame` where broadcasting scalar inputs was order dependent.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -907,24 +907,24 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
         if index is None:
             num_rows = 0
             if data:
-                lengths = set()
-                for key, value in data.items():
-                    if is_scalar(value):
-                        data[key] = value
-                        lengths.add(1)
-                    else:
-                        value = column.as_column(
-                            value, nan_as_null=nan_as_null
+                keys, values, lengths = zip(
+                    *(
+                        (k, v, 1)
+                        if is_scalar(v)
+                        else (
+                            k,
+                            vc := as_column(v, nan_as_null=nan_as_null),
+                            len(vc),
                         )
-                        lengths.add(len(value))
-                        data[key] = value
-
-                if len(lengths - {1}) > 1:
-                    raise ValueError("All arrays must be of the same length")
-                else:
-                    num_rows = (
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/dataframe.py
```diff
@@ -907,14 +907,24 @@ def _init_from_dict_like(
         if index is None:
             num_rows = 0
             if data:
-                col_name = next(iter(data))
-                if is_scalar(data[col_name]):
-                    num_rows = num_rows or 1
+                lengths = set()
+                for key, value in data.items():
+                    if is_scalar(value):
+                        data[key] = value
```
wence-, line 913: This is unnecessary because `data[key]` is already `value`.

python/cudf/cudf/core/dataframe.py
```diff
@@ -907,14 +907,24 @@ def _init_from_dict_like(
...
+
+                if len(lengths - {1}) > 1:
+                    raise ValueError("All arrays must be of the same length")
                 else:
-                    data[col_name] = column.as_column(
-                        data[col_name], nan_as_null=nan_as_null
+                    num_rows = (
+                        1 if lengths == {1} else next(iter(lengths - {1}))
                     )
-                    num_rows = len(data[col_name])
```
wence-, line 917: Is this rewrite too tricksy?
```python
            keys, values, lengths = zip(
                *(
                    (k, v, 1)
                    if is_scalar(v)
                    else (k, vc := as_column(v, nan_as_null=nan_as_null), len(vc))
                    for k, v in data.items()
                )
            )
            data = dict(zip(keys, values))
            try:
                (num_rows,) = (lengths - {1}) or {1}
            except ValueError:
                raise ValueError("All arrays must be the same length")
```

I tend to write things like this because I (personally) find stateful modification in loops hard to reason about. But that is probably just me.

Actually you don't need to carry the keys along for the ride, but I tend to because I still don't trust dictionary iteration order :).

wence- (commented):
A minor suggestion, that needn't necessarily be acted on.

python/cudf/cudf/core/dataframe.py
galipremsagar: This approach looks good to me, updated 👍

PR #13001, opened 2023-03-23
Pin numba version

vyasr:
## Description

Changes in the upcoming numba 0.57 release will break cudf until we've had time to adapt them, which will not happen in 23.04.

## Checklist

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -340,7 +340,7 @@ dependencies:
         packages:
           - cachetools
           - cuda-python>=11.7.1,<12.0
-          - &numba numba>=0.56.2,<0.57
+          - &numba numba>=0.56.4,<0.57
           - nvtx>=0.2.1
           - packaging
           - rmm==23.4.*
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -340,7 +340,7 @@ dependencies:
         packages:
           - cachetools
           - cuda-python>=11.7.1,<12.0
-          - numba>=0.56.2
+          - &numba numba>=0.56.2,<0.57
           - nvtx>=0.2.1
           - packaging
           - rmm==23.4.*
@@ -493,4 +493,4 @@ dependencies:
       - output_types: [conda, requirements, pyproject]
         packages:
           - dask-cuda==23.4.*
-          - numba>=0.56.2
+          - *numba
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -44,7 +44,7 @@ dependencies:
 - nbsphinx
 - ninja
 - notebook
-- numba>=0.56.2
+- numba>=0.56.2,<0.57
 - numpy>=1.21
 - numpydoc
 - nvcc_linux-64=11.8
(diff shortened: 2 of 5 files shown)


bdice (approved):

dependencies.yaml
```diff
@@ -340,7 +340,7 @@ dependencies:
         packages:
           - cachetools
           - cuda-python>=11.7.1,<12.0
-          - numba>=0.56.2
+          - &numba numba>=0.56.2,<0.57
```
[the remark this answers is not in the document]
bdice: YAML anchor. The & defines it and the * reuses that content.

gmarkall (commented):
There were some CUDA-related fixes for extensions that went into Numba 0.56.4 - while we're here could / should we bump it to `>=0.56.4,<0.57`?

FWIW, `<0.57` looks like the correct upper bound to me from a Numba standpoint but I can't vouch for the changes (whether there's other locations that need changing in cudf).

dependencies.yaml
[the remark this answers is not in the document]
galipremsagar: I see, TIL. Thanks 👍

galipremsagar (approved):

bdice:
@gmarkall @brandon-b-miller @vyasr Can we cross-link an issue describing the problem with Record types / UDFs? Anything to document why we’re pinning in a little more detail, so that we can resolve it later when 0.57 is out and have everything link together.

gmarkall:
@bdice Created in #13002 to record the general idea - I don't have a working cuDF environment right now so the issue doesn't have e.g. the traceback and a complete environment description attached, but the problem is well-understood.

dependencies.yaml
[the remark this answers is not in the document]
vyasr: This creates a [YAML anchor](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) that allows it to be referenced later (see the other place I'm doing `*numba`)

dependencies.yaml
```diff
@@ -340,7 +340,7 @@ dependencies:
         packages:
           - cachetools
           - cuda-python>=11.7.1,<12.0
-          - numba>=0.56.2
+          - &numba numba>=0.56.4,<0.57
```
gmarkall, line 343: Probably a basic question, but why does this change into `& numba numba`?

dependencies.yaml
```diff
@@ -493,4 +493,4 @@ dependencies:
       - output_types: [conda, requirements, pyproject]
         packages:
           - dask-cuda==23.4.*
-          - numba>=0.56.2
+          - *numba
```
gmarkall, line 496: Is this related to the `&numba numba` above?

gmarkall (approved):
LGTM (from the perspective of the Numba version ranges).

dependencies.yaml
wence-: Yes, the `&numba ...` takes a reference to `...` and "stores" it in `numba` which can then be dereferenced and copied in with `*numba`.

wence- (approved):

dependencies.yaml
ajschmidt8: This is referred to as a YAML anchor. You can view the spec here: https://yaml.org/spec/1.2.2/#692-node-anchors.

There are some Google results that provide better examples than the official spec docs above.

In this case an anchor was used to keep the `dependencies.yaml` file DRY.

ajschmidt8 (approved):

PR #13006, opened 2023-03-24
Fix memcheck script to execute only _TEST files found in bin/gtests/libcudf

davidwendt:
The nightly runs of `compute-sanitizer` started failing due to the test script trying to execute some new extra files found in the `$CONDA_PREFIX"/bin/gtests/libcudf/` directory. This change ensures only files ending in `_TEST` are executed by `compute-sanitizer`.

For reference errors are here: https://github.com/rapidsai/cudf/actions/runs/4508267264/jobs/7936800047
Example:
```
Running compute-sanitizer on CTestTestfile.cmake
========= COMPUTE-SANITIZER
========= Error: Target application terminated before first instrumented API call
```

Follow on issue/PR could explore adding a special make option to execute `compute-sanitizer` only on test files.

## Checklist

### DIFF
--- a/ci/test_cpp_memcheck.sh
+++ b/ci/test_cpp_memcheck.sh
@@ -11,7 +11,7 @@ set +e
 rapids-logger "Memcheck gtests with rmm_mode=cuda"
 export GTEST_CUDF_RMM_MODE=cuda
 COMPUTE_SANITIZER_CMD="compute-sanitizer --tool memcheck"
-for gt in "$CONDA_PREFIX"/bin/gtests/libcudf/* ; do
+for gt in "$CONDA_PREFIX"/bin/gtests/libcudf/*_TEST ; do
     test_name=$(basename ${gt})
     if [[ "$test_name" == "ERROR_TEST" ]] || [[ "$test_name" == "STREAM_IDENTIFICATION_TEST" ]]; then
         continue


bdice (approved):

ajschmidt8 (commented):
@davidwendt, it looks like the new files that have showed up in that directory are these:

```
CTestTestfile.cmake
generate_ctest_json
run_gpu_test.cmake
```

These look like they're from https://github.com/rapidsai/cudf/pull/12451.

As part of the changes in that PR, @robertmaynard also switched up the way that the GTests are invoked for our regular `test_cpp.sh` script.

I'm wondering if his changes should also be adapted for this `test_cpp_memcheck.sh` script.

His method looks a bit more robust than the approach that we've been using previously.

davidwendt:
> I'm wondering if his changes should also be adapted for this `test_cpp_memcheck.sh` script.
> 
> His method looks a bit more robust than the approach that we've been using previously.

Yes, this is what I meant about making this work with make/cmake in a future PR.
The `test_cpp.sh` uses `ctest` which is a cmake utility.

ajschmidt8 (approved):

ajschmidt8:
> Yes, this is what I meant about making this work with make/cmake in a future PR. The `test_cpp.sh` uses `ctest` which is a cmake utility.

Got it. Sorry, missed that part. Approved to get this unblocked for now.

PR #13011, opened 2023-03-25
Fix OOB memory access in CSV reader when reading without NA values

vuule:
## Description
CSV reader uses a trie to read field with special values as nulls. The creation of the trie does not work correctly when there are not special values. This can happen when the NA filter is enabled, but the default NA values are removed, and user does not specify custom values. In this case, use of this trie leads to OOB memory access.
This PR fixes the trie creation to create an empty trie when there are not special values to look for.
Included a C++ test that crashes without the fix.

## Checklist


cpp/src/io/utilities/trie.cuh
```diff
@@ -83,8 +83,8 @@ trie create_serialized_trie(const std::vector<std::string>& keys, rmm::cuda_stre
 __host__ __device__ inline bool serialized_trie_contains(device_span<serial_trie_node const> trie,
                                                          device_span<char const> key)
 {
-  if (trie.data() == nullptr || trie.empty()) return false;
-  if (key.empty()) return trie.front().is_leaf;
+  if (trie.empty()) { return false; }
```
ttnghia, line 86: What's wrong with `trie.data() == nullptr`?

vyasr: I assume it's strictly redundant since a device span with `data() == nullptr` will always have size 0.

vyasr (approved):

hyperbolic2346 (approved):
🚀

ttnghia (approved):

PR #13029, opened 2023-03-28
[REVIEW] Fix branch version in CI scripts

galipremsagar:
## Description
This PR fixes missed version upgrades in scripts.

## Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -98,5 +98,5 @@ jobs:
       sha: ${{ inputs.sha }}
       package-name: dask_cudf
       # Test against latest dask/distributed/dask-cuda.
-      test-before: "pip install git+https://github.com/dask/dask.git@main git+https://github.com/dask/distributed.git@main git+https://github.com/rapidsai/dask-cuda.git@branch-23.04"
+      test-before: "pip install git+https://github.com/dask/dask.git@main git+https://github.com/dask/distributed.git@main git+https://github.com/rapidsai/dask-cuda.git@branch-23.06"
       test-unittest: "python -m pytest -v -n 8 ./python/dask_cudf/dask_cudf/tests"
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -134,5 +134,5 @@ jobs:
       build_type: pull-request
       package-name: dask_cudf
       # Install the cudf we just built, and also test against latest dask/distributed/dask-cuda.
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yaml
```diff
@@ -98,5 +98,5 @@ jobs:
       sha: ${{ inputs.sha }}
       package-name: dask_cudf
       # Test against latest dask/distributed/dask-cuda.
-      test-before: "pip install git+https://github.com/dask/dask.git@main git+https://github.com/dask/distributed.git@main git+https://github.com/rapidsai/dask-cuda.git@branch-23.04"
+      test-before: "pip install git+https://github.com/dask/dask.git@main git+https://github.com/dask/distributed.git@main git+https://github.com/rapidsai/dask-cuda.git@branch-23.06"
```
bdice, line 101: Is this automated correctly? It looks like line 104 of `update-version.sh` should've fixed this. Here's raft's code, for comparison:

https://github.com/rapidsai/raft/blob/22ebc72e8c0c2738a30a0bbf9b0efc52736eddd6/ci/release/update-version.sh#L64-L67

bdice (approved):

.github/workflows/test.yaml
galipremsagar: It is automated correctly in cudf:

https://github.com/rapidsai/cudf/blob/branch-23.04/ci/release/update-version.sh#L101

I verified running it locally and the script should've upgraded this automatically but looks like somehow it wasn't.

msadang (approved):
LGTM

.github/workflows/test.yaml
raydouglass: Looks like this is a BSD sed vs GNU sed issue. When I run this script during burndown, I use my Mac which uses BSD sed and it misses this change. When I updated to `gsed` in the `update-version.sh` script it worked as expected.

Let me think about how to handle this in the future.

ajschmidt8: I assume (though cannot confirm since I don't use a Mac) that the issue with that expression lies within this string: `branch-[^\"\s]\+`

Therefore we can probably just change this:
```sh
for FILE in .github/workflows/*.yaml; do
  sed_runner "s/dask-cuda.git@branch-[^\"\s]\+/dask-cuda.git@branch-${NEXT_SHORT_TAG}/g" ${FILE};
done
```

to this:

```sh
for FILE in .github/workflows/*.yaml; do
  sed_runner "s/dask-cuda.git@branch-[0-9][0-9].[0-9][0-9]/dask-cuda.git@branch-${NEXT_SHORT_TAG}/g" ${FILE};
done
```

PR #13043, opened 2023-03-30
Log cuIO warnings using the libcudf logger

vuule:
## Description
Log warning messages in the following scenarios:

- cuIO reader benchmark is ran without dropping the L3 cache.
- An nvCOMP call fails and we fall back to older API.
- Mangling of duplicated column names is disabled and some columns are dropped.
- Default compression fails in ORC writer and we fall back to uncompressed output.
- Parquet writer is writing a decimal column with precision < 10 as int64.

## Checklist


cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -269,6 +269,9 @@ size_t batched_compress_temp_size(compression_type compression,
       compression, num_chunks, max_uncomp_chunk_size, max_total_uncomp_size);
   } catch (...) {
     // Ignore errors in the expanded version; fall back to the old API in case of failure
+    CUDF_LOG_WARN(
+      "Error in CompressGetTempSizeEx call, falling back to CompressGetTempSize; this may increase "
```
harrism, line 273: Suggest not using the word Error in a warning message, to avoid confusion in searches.

```suggestion
      "CompressGetTempSizeEx call failed, falling back to CompressGetTempSize; this may increase "
```

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1347,6 +1347,7 @@ void CompressOrcDataStreams(uint8_t* compressed_data,
         [] __device__(compression_result & stat) { stat.status = compression_status::FAILURE; });
       // Since SNAPPY is the default compression (may not be explicitly requested), fall back to
       // writing without compression
+      CUDF_LOG_WARN("ORC writer: error during compression, writing uncompressed data");
```
harrism, line 1350:
```suggestion
      CUDF_LOG_WARN("ORC writer: compression failed, writing uncompressed data");
```

harrism (approved):
Suggest removing word "Error" from warning messages, otherwise looks good.

PR #13047, opened 2023-03-31
Reduce shared memory usage in gpuComputePageSizes by 50%

nvdbaranec:
In a multithreaded, multi-stream environment (Spark) we were experiencing a performance regression on some benchmark queries.  The culprit was gpu scheduling issues related to the `gpuComputePageSizes` kernel.   Dependent kernels (`gpuDecodePages`) were getting serialized because `gpuComputePageSizes` wasn't running alongside other streams well.

The fix was reducing shared memory usage in `gpuComputePageSizes`.  The kernel shares a lot of code and data structures with `gpuDecodePages` but doesn't actually use several of the large buffers that are stored in shared memory.  This PR refactors those buffers out so that they are only declared in the `gpuDecodePages` kernel, reducing the shared usage by 50% (3kb).

This clears up the performance issue on Spark.  I am currently experiencing build issues with cudf benchmarks so I'm marking this as do-not-merge until I can verify with them.

## Checklist


ttnghia (approved):
Nit: Recommend the following:
 * Polish comments more if possible, and
 * Run compute sanitizer to check if there is any issue.

vuule (approved):
Looks good.

nvdbaranec:
> Nit: Recommend the following:
> 
>     * Polish comments more if possible, and

Right. Forgot to update all the function comments.  Will fix.

PR #13054, opened 2023-04-04
Add except declaration in Cython interface for regex_program::create

### LINKED ISSUE #13052: [BUG] Exceptions thrown in `regcomp.cpp` seem to not propagate up to Cython (leads to SIGABRT)
Exceptions that are thrown in the C++ code in [`regcomp.cpp`](https://github.com/rapidsai/cudf/blob/branch-23.06/cpp/src/strings/regex/regcomp.cpp) seem to not propagate all the way up to Python and lead to a SIGABRT. For example:

```python
>>> import cudf
>>> s = cudf.Series(["a"])
>>> s.str.extract("{\}")
terminate called after throwing an instance of 'cudf::logic_error'
  what():  CUDF failure at:/cudf/cpp/src/strings/regex/regcomp.cpp:565: invalid regex pattern: nothing to repeat at position 0
fish: “python” terminated by signal SIGABRT (Abort)
```

Our [bindings](https://github.com/rapidsai/cudf/blob/bd1ef2959e64c9f44a709cd9aaaad5dc72611c23/python/cudf/cudf/_lib/cpp/strings/extract.pxd#L15) to the `extract` function are marked `except +`, which means that the C++ code generated by Cython will catch any C++ exceptions thrown by `cudf::strings::extract` and rethrow them as Python errors. For example, this all works as expected when the error is thrown from outside `regcomp.cpp`:

```python
>>> s.str.extract("blah")
RuntimeError                              Traceback (most recent call last)
...
...
RuntimeError: CUDF failure at: /cudf/cpp/src/strings/extract/extract.cu:98: Group indicators not found in regex pattern

```

cc: @davidwendt is there something internally within libcudf that may be preventing exceptions from `regcomp.cpp` from being propagated up to the Python caller?

davidwendt:
## Description
Add the `except +` declaration to the `cudf::strings::regex_program::create()` function in the Cython `regex_program.pxd` interface since invalid regex patterns are thrown by this call. This allows the normal Cython exception handling to pass the exception to the Python logic without aborting the process.

Closes #13052 

## Checklist

### DIFF
--- a/python/cudf/cudf/_lib/cpp/strings/regex_program.pxd
+++ b/python/cudf/cudf/_lib/cpp/strings/regex_program.pxd
@@ -1,4 +1,4 @@
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 
 from libcpp.memory cimport unique_ptr
 from libcpp.string cimport string
@@ -12,4 +12,7 @@ cdef extern from "cudf/strings/regex/regex_program.hpp" \
     cdef cppclass regex_program:
 
         @staticmethod
-        unique_ptr[regex_program] create(string pattern, regex_flags flags)
+        unique_ptr[regex_program] create(
+            string pattern,
+            regex_flags flags
+        ) except +
--- a/python/cudf/cudf/_lib/strings/contains.pyx
+++ b/python/cudf/cudf/_lib/strings/contains.pyx
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
 
 from cython.operator cimport dereference
 from libc.stdint cimport uint32_t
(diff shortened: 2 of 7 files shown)


bdice (approved):
Changes look good. Should we add a test?

bdice:
Proposed test:
```python
def test_invalid_regex():
    s = cudf.Series(["a"])
    with pytest.raises(ValueError):    # Might be the wrong error class, please verify
        s.str.extract("{\}")
```

davidwendt:
> Changes look good. Should we add a test?

Yes, thanks. 
Also, I was wondering about changing the libcudf exception to something more specific like `std::invalid_argument` for poorly formed regex pattern. I could do that in a follow up PR. Just requesting opinions.

shwina (approved):
LGTM, pending @bdice's suggestions for testing

bdice:
> Also, I was wondering about changing the libcudf exception to something more specific like `std::invalid_argument` for poorly formed regex pattern. I could do that in a follow up PR. Just requesting opinions.

I think that’d be fine. I am not up to date on the latest guidelines / changes for exception types so I don’t know exactly what to recommend. I know we’ve been pushing to improve that recently.

PR #13060, opened 2023-04-04
Introduce Numba CUDA Runtime API

brandon-b-miller:
This PR introduces the central object of concern for numba reference counting in CUDA, the `NRT_MemInfo` object, as well as a set of functions for creating and destroying them. Much of this is vendored and stripped down from [the numba nopython mode c++ files](https://github.com/numba/numba/blob/main/numba/core/runtime/nrt.cpp). The difference is I generally added `__device__` to the APIs, and removed logging and stats. If desired I can add these back in.

This PR also includes a shim utility for creating `NRT_MemInfo` instances with the `udf_string` destructor attached. MemInfo's created in this way should, upon reaching a `refct` of zero, deallocate using `udf_str_dtor`.


brandon-b-miller:
cc @gmarkall

wence-:
> Much of this is vendored and stripped down from [the numba nopython mode c++ files](https://github.com/numba/numba/blob/main/numba/core/runtime/nrt.cpp).

Numba is BSD 2-clause. But you've at the very least dropped that license information.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,86 @@
...
+  size_t size;
+};
+}
+
+typedef struct MemInfo NRT_MemInfo;
+
+// Globally needed variables
+struct NRT_MemSys {
+  /* Shutdown flag */
+  int shutting;
```
gmarkall, line 40: Can we get by without this? I don't think we need to worry about the interpreter shutting down when we're on-device.

gmarkall:
> removed logging and stats. If desired I can add these back in.

These are hardly used (if at all) in Numba these days, but were valuable debugging and testing tools early on, before the Numba runtime was robust and widely tested in use. I think it makes sense to keep them in from the outset but perhaps disabled by default by a #ifdef (or whatever mechanism is appropriate for cuDF C++ code) so that they're readily available when we're inevitably chasing down our early bugs.

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,15 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
 make_definition_idx(BlockIdxMax, float64, double);
 #undef make_definition_idx
 }
+
+__device__ void udf_str_dtor(void* udf_str, size_t size, void* dtor_info)
```
gmarkall, line 668: I wonder how obvious it is what's going on here - could it do with some explanatory comments? I know that the purpose is to cause the `udf_string` to free its data via `clear()` rather than calling `delete` on it because it was constructed with placement `new`, but I don't know if it comes across in the code in this PR.

I'm also a bit unclear why we call `clear()` rather than being explicit that we're trying to destroy the `udf_string` with something like `ptr->~udf_string()` - is this a question for @davidwendt?

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,15 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
...
+{
+  udf_string* ptr = reinterpret_cast<udf_string*>(udf_str);
+  ptr->clear();
```
gmarkall, line 671: Given @davidwendt's comment above, then I think this should be:
```suggestion
  ptr->~udf_string();
```

And I think then there' no explanatory comment needed, because it's clear this is just a Numba shim for the destructor.

python/cudf/udf_cpp/shim.cu
brandon-b-miller: Updated to call the destructor

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
brandon-b-miller: removed this

brandon-b-miller:
Hi all, an update here, specifically to `meminfo_from_new_udf_str`. The problem I was having up until this point is that the process of returning new string from any operation goes like this:
1. Numba allocates enough space for a `udf_string` on the stack
2. Numba passes the pointer obtained to the allocation in (1) to a shim function
3. The shim function uses the `udf_string` c++ APIs to build the new string at the pointed-to location from (2) using placement `new`
4. Numba returns the pointed-to struct out of the operation by value

BUT when we create a new `NRT_MemInfo` object, we need to do so around some kind of pointer. The original pointer to the stack allocation where the `udf_string` was constructed doesn't necessarily contain the `udf_string` though at the point at which the refcount hits zero.

The solution in this PR is to allocate an extra `sizeof(udf_string)` when the `NRT_MemInfo` is created and point that `NRT_MemInfo` to that small extra heap allocation. Then copy the `udf_string` into that memory. To be clear this is a copy of the `udf_string` struct but _not_ its data. Technically this creates a situation where two `udf_string` objects point to the same actual data allocation. But numba has no way outside of an `NRT_MemInfo` to access the destructor for its stack memory - to numba it's a simple `c` struct and thus there's never a double free.

Technically `udf_string` objects are mutable on the c++ level meaning the heap copy could go out of sync, but we don't _use_ `udf_string` in a mutable way - all our shim functions, even the ones that _could_ mutate an existing string and just return that, create new strings instead.

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,41 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
...
+
+// Only used to allocate the right amount of space, see below
+struct meminfo_and_str {
+  NRT_MemInfo mi;
+  udf_string st;
+};
+
+/*
+Create a new MemInfo object holding the reference count of a udf_string. When returning
+new strings, shim functions expect a pointer to a stack allocated buffer into which it
```
shwina, line 682: > shim functions expect a pointer to a stack allocated buffer

Is this right? Or is it that we happen to always pass pointers to stack allocated objects to the shim function?

shwina (approved):
LGTM pending a minor clarification

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,41 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
 make_definition_idx(BlockIdxMax, float64, double);
 #undef make_definition_idx
 }
+
+__device__ void udf_str_dtor(void* udf_str, size_t size, void* dtor_info)
+{
+  udf_string* ptr = reinterpret_cast<udf_string*>(udf_str);
```
vyasr, line 670:
```suggestion
  auto ptr = reinterpret_cast<udf_string*>(udf_str);
```

python/cudf/udf_cpp/shim.cu
vyasr: What buffers are we referring to here? Everything in the `udf_string` is heap allocated AFAIR.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
vyasr, line 1: I'm assuming that there's nothing to review in this file other than the addition of `__device__`.

python/cudf/udf_cpp/shim.cu
brandon-b-miller: This refers to the piece of memory containing the `udf_string` struct itself. These are allocated on the numba side before being passed in, for example [here](https://github.com/rapidsai/cudf/blob/branch-23.06/python/cudf/cudf/core/udf/strings_lowering.py#L228). More info on `alloca` [here](https://llvm.org/docs/LangRef.html#alloca-instruction). The important bit being

```
The ‘alloca’ instruction allocates sizeof(<type>)*NumElements bytes of memory on the runtime stack, returning a pointer of the appropriate type to the program
```

The `udf_string` consists of three members, all _technically_ integers and is usually 24 bytes IIRC:
- the length of the string
- the total number of bytes in the string
- a pointer to the heap allocation containing the string

So while one of the struct members is a pointer to a heap allocation the struct itself lives in a stack allocation. This is the reason so much fancy footwork is needed in this function. The `NRT_MemInfo` object must be created around a pointer of some kind. It can't be created around the pointer to the stack allocation containing the `udf_string`, because later on when it comes time to call the destructor it may not be there anymore.

One could use the heap pointer underneath the `udf_string` for this purpose technically, since the pointed-to allocation won't be going anywhere until explicitly freed. But in discussing this with @davidwendt we decided that it's not a great idea to be freeing pointers underneath a `udf_string` via some external mechanism that isn't exactly the `udf_string` dtor itself. For instance imagine in the future if `udf_string` was changed somehow to require the other members to correctly clean up all of its data. At that point just having a pointer but no way of getting to the rest of the `udf_string` won't be sufficient.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,106 @@
...
+
+extern "C" __device__ void NRT_MemInfo_init(
+  NRT_MemInfo* mi, void* data, size_t size, NRT_dtor_function dtor, void* dtor_info)
+{
+  mi->refct     = 1; /* starts with 1 refct */
+  mi->dtor      = dtor;
+  mi->dtor_info = dtor_info;
+  mi->data      = data;
+  mi->size      = size;
+  if (TheMSys.stats.enabled)
```
gmarkall, line 78: We have this stats functionality here, but it's not present in the other functions - is it an oversight that this is present, or that it's missing from the other functions? (e.g. `NRT_Free` has no `TheMSys.stats.free++;` in it).

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
gmarkall: It's a cut-down version of some of the functions in https://github.com/numba/numba/blob/main/numba/core/runtime/nrt.cpp - I think the changes to the copied code are:

- Some changes to stats recording
- Addition of the `__device__` keyword
- Replacement of `std::atomic` with `cuda::atomic`
- Simplifying out / removing support for custom allocators (there's no need for now, and if we do want to do something custom with allocation it likely won't fit into the existing abstraction anyway)

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,106 @@
...
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#include <cuda/atomic>
+
+typedef __device__ void (*NRT_dtor_function)(void* ptr, size_t size, void* info);
```
brandon-b-miller, line 29: These typedefs are lifted from [nrt.h](https://github.com/numba/numba/blob/ce9bdacc76816653e4f000e9210d45c667fa9f98/numba/core/runtime/nrt.h#L27). I added `__device__`.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
brandon-b-miller: oversight. Will put these back in where they originally were

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,106 @@
...
+  mi->dtor_info = dtor_info;
+  mi->data      = data;
+  mi->size      = size;
+  if (TheMSys.stats.enabled)
+  {
+    TheMSys.stats.mi_alloc++;
+  }
+}
+
+__device__ NRT_MemInfo* NRT_MemInfo_new(void* data,
```
brandon-b-miller, line 84: This is vendored from [nrt.cpp](https://github.com/numba/numba/blob/main/numba/core/runtime/nrt.cpp#L183) where debug statements have been removed and the call to `NRT_MemInfo_init` has been adjusted to align with its new signature that drops the external allocator.

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,41 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
...
+  NRT_MemInfo mi;
+  udf_string st;
+};
+
+/*
+Create a new MemInfo object holding the reference count of a udf_string. When returning
+new strings, shim functions expect a pointer to a stack allocated buffer into which it
+will construct the udf_string it returns. Since one can not safely build a MemInfo object
+around this stack memory, we store a copy of the udf_string itself next to the MemInfo
+where it can be later used to destroy the underlying data.
```
gmarkall, line 685: Does "it can be later used to destroy the underlying data." imply a usage of the `MemInfo` that's not as broad in scope as what a `MemInfo` actually does? Does

```suggestion
which manages the lifetime of the udf_string.
```

provide a more accurate representation of what's happening?

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,41 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
...
+    udf_string* st_ptr = &(mi_and_str->st);
+
+    NRT_MemInfo_init(mi_ptr, st_ptr, NULL, udf_str_dtor, NULL);
+
+    // copy the udf_string to the extra heap space
+    udf_string* in_str_ptr = reinterpret_cast<udf_string*>(udf_str);
+    memcpy(st_ptr, in_str_ptr, sizeof(udf_string));
+  }
+
+  *nb_retval = &(mi_and_str->mi);
```
gmarkall, line 702: Should this be inside the `if`? If `mi_and_str` is null, then this is an illegal memory access. Should also `*nb_retval` be set to a sentry value (perhaps a null or something recognizable) so it'll be easier to track down if an issue was due to a failed allocation?

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,41 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
...
+*/
+extern "C" __device__ int meminfo_from_new_udf_str(void** nb_retval, void* udf_str)
+{
+  // allocate enough room for both the meminfo and udf_string
+  meminfo_and_str* mi_and_str = (meminfo_and_str*)NRT_Allocate(sizeof(meminfo_and_str));
+  if (mi_and_str != NULL) {
+    auto mi_ptr        = &(mi_and_str->mi);
+    udf_string* st_ptr = &(mi_and_str->st);
+
+    NRT_MemInfo_init(mi_ptr, st_ptr, NULL, udf_str_dtor, NULL);
```
gmarkall, line 695: It's a bit confusing for the uninitiated reader to follow seeing the size being null - perhaps:
```suggestion
    // We pass a null size here because the udf_string actually exists on the heap, 
    // and tracks the size of the string data that it points to.
    NRT_MemInfo_init(mi_ptr, st_ptr, NULL, udf_str_dtor, NULL);
```

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,110 @@
...
+                                        size_t size,
+                                        NRT_dtor_function dtor,
+                                        void* dtor_info)
+{
+  NRT_MemInfo* mi = (NRT_MemInfo*)NRT_Allocate(sizeof(NRT_MemInfo));
+  if (mi != NULL) { NRT_MemInfo_init(mi, data, size, dtor, dtor_info); }
+  return mi;
+}
+
+extern "C" __device__ void NRT_Free(void* ptr)
```
brandon-b-miller, line 91: This is vendored from [nrt.cpp](https://github.com/numba/numba/blob/main/numba/core/runtime/nrt.cpp#L545) with the removal of debug print and the addition of `__device__`.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,110 @@
...
+{
+  TheMSys.allocator.free(ptr);
+  if (TheMSys.stats.enabled) { TheMSys.stats.free++; }
+}
+
+extern "C" __device__ void NRT_dealloc(NRT_MemInfo* mi) { NRT_Free(mi); }
```
brandon-b-miller, line 97: this is vendored from [nrt.cpp](https://github.com/numba/numba/blob/main/numba/core/runtime/nrt.cpp#L369) dropping the logic for the external allocator and debug print.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,110 @@
...
+{
+  NRT_MemInfo* mi = (NRT_MemInfo*)NRT_Allocate(sizeof(NRT_MemInfo));
+  if (mi != NULL) { NRT_MemInfo_init(mi, data, size, dtor, dtor_info); }
+  return mi;
+}
+
+extern "C" __device__ void NRT_Free(void* ptr)
+{
+  TheMSys.allocator.free(ptr);
+  if (TheMSys.stats.enabled) { TheMSys.stats.free++; }
```
gmarkall, line 94: There's an asymmetry in the stats in that there's nothing to increment `TheMSys.stats.alloc` - I think this is missing from the `NRT_Allocate()` function

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,110 @@
...
+}
+
+extern "C" __device__ void NRT_dealloc(NRT_MemInfo* mi) { NRT_Free(mi); }
+
+extern "C" __device__ void NRT_MemInfo_destroy(NRT_MemInfo* mi)
+{
+  NRT_dealloc(mi);
+  if (TheMSys.stats.enabled) { TheMSys.stats.mi_free++; }
+}
+extern "C" __device__ void NRT_MemInfo_call_dtor(NRT_MemInfo* mi)
```
brandon-b-miller, line 104: This is vendored from [nrt.cpp](https://github.com/numba/numba/blob/main/numba/core/runtime/nrt.cpp#L397) except we don't check in CUDA if `TheMSys.shutting` which I believe is used to track information about the python interpreter in numba nopython mode, @gmarkall can correct me if I am wrong here. We might be able to remove that member from our version of `TheMSys`.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
brandon-b-miller: @vyasr I added a couple of comments in the diff around the vendored functions that call out what I changed in the CUDA versions, let me know if they are helpful.

python/cudf/udf_cpp/shim.cu
gmarkall: (note - I hadn't got Brandon's previous comment visible when I posted mine, so ignore mine for now)

python/cudf/udf_cpp/shim.cu
gmarkall: I think I wrote "heap" when I meant "stack" there, apologies for confusion

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
gmarkall: That's correct, `shutting` is only used to avoid doing things that would be unsafe when the interpreter is in the middle of being torn down.

python/cudf/udf_cpp/shim.cu
brandon-b-miller: Updated this

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
brandon-b-miller: I added the stats machinery back into `NRT_Allocate` whose internals are now the internals of `NRT_Allocate_external` modulo the actual logic for an external allocator.

python/cudf/udf_cpp/shim.cu
vyasr: Agreed, we don't want to reach into the `udf_string` and rely on its internals. The design here makes sense now that you've explained it, I needed the pointers to the builder and the realization that we're talking about stack allocations within the context of the LLVM instructions rather than in the code actually written here. The explanation is super helpful, thanks.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
vyasr: The comments are very helpful, thank you!

vyasr (approved):
The extra info from the last round of review was very helpful to clarify things, I'm on board now.

brandon-b-miller:
After propagating the review changes forward to the final PR there turned out to be a missing initialization of the `msys` object that I added at https://github.com/rapidsai/cudf/pull/13060/commits/6a0aae839af82adbfe232c3fbe6b20e6d0f70687. Let me know if this still seems ok.

vyasr:
The initializer looks fine to me. I'll leave it up to Graham to give the final OK here. Also looks like style fixes are needed to get CI passing.

brandon-b-miller:
> The initializer looks fine to me. I'll leave it up to Graham to give the final OK here. Also looks like style fixes are needed to get CI passing.

correct. I think we'll need to exclude this file from the copyright check since it will complain about a nonstandard header.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,122 @@
...
+    cuda::atomic<size_t, cuda::thread_scope_device> mi_free;
+  } stats;
+};
+
+__device__ void* malloc_wrapper(size_t size) { return malloc(size); }
+
+__device__ void free_wrapper(void* ptr) { free(ptr); }
+
+/* The Memory System object */
+__device__ NRT_MemSys TheMSys = {
```
gmarkall, line 70: If this header is included in two separate translation units, will `TheMSys` be multiply-defined?

gmarkall (commented):
Aside from my question about multiple definitions of `TheMSys`, I think this looks good as a standalone unit. I haven't fully understood subsequent PRs fully, but I think once that query is resolved, approving this and narrowing the workload remaining to review is the way to go.

vyasr:
> Aside from my question about multiple definitions of `TheMSys`, I think this looks good as a standalone unit. I haven't fully understood subsequent PRs fully, but I think once that query is resolved, approving this and narrowing the workload remaining to review is the way to go.

Agreed, I've been ignoring the follow-up PRs until this one is merged.

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
brandon-b-miller: I added an include guard for this scenario in https://github.com/rapidsai/cudf/pull/13060/commits/d9001d3a390e081c9b20b35d67de63e4e523d4e7

gmarkall: I don't think an include guard will stop the symbol being defined twice in separate translation units - it can be included once in each TU.

brandon-b-miller: ah I see what you mean. Working on a fix for this.

gmarkall (approved):
I think this looks good now. I am assuming that `shim.cu` never gets added to the link twice because this would already be causing all sorts of havoc, and therefore defining `TheMSys` in it is safe.

wence- (commented):
A few small queries

python/cudf/udf_cpp/nrt/_nrt_cuda.cuh
```diff
@@ -0,0 +1,126 @@
...
+    cuda::atomic<size_t, cuda::thread_scope_device> alloc;
+    cuda::atomic<size_t, cuda::thread_scope_device> free;
+    cuda::atomic<size_t, cuda::thread_scope_device> mi_alloc;
+    cuda::atomic<size_t, cuda::thread_scope_device> mi_free;
+  } stats;
+};
+
+__device__ void* malloc_wrapper(size_t size) { return malloc(size); }
+
+__device__ void free_wrapper(void* ptr) { free(ptr); }
```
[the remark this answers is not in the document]
gmarkall: > In the spirit of IWYU, this header file should `#include` the header file that provides forward declarations for `malloc` and `free`.

I think that should be `cuda_runtime.h` (the actual declarations are in `crt/common_functions.h` but they warn not to use them directly and use `cuda_runtime.h` instead).

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,53 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
...
+Create a new MemInfo object holding the reference count of a udf_string. When returning
+new strings, shim functions expect a pointer to a stack allocated buffer into which it
+will construct the udf_string it returns. Since one can not safely build a MemInfo object
+around this stack memory, we store a copy of the udf_string itself next to the MemInfo
+which manages the lifetime of the udf_string.
+*/
+extern "C" __device__ int meminfo_from_new_udf_str(void** nb_retval, void* udf_str)
+{
+  // allocate enough room for both the meminfo and udf_string
+  meminfo_and_str* mi_and_str = (meminfo_and_str*)NRT_Allocate(sizeof(meminfo_and_str));
```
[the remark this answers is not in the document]
gmarkall: Calling `NRT_Allocate` will also modify the alloc / dealloc stats, which are used for debugging purposes.

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,53 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
...
+    auto mi_ptr        = &(mi_and_str->mi);
+    udf_string* st_ptr = &(mi_and_str->st);
+
+    // We pass a null size here because the udf_string actually exists on the stack
+    // and tracks the size of the string data that it points to.
+    NRT_MemInfo_init(mi_ptr, st_ptr, NULL, udf_str_dtor, NULL);
+
+    // copy the udf_string to the extra heap space
+    udf_string* in_str_ptr = reinterpret_cast<udf_string*>(udf_str);
+    memcpy(st_ptr, in_str_ptr, sizeof(udf_string));
```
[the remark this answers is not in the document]
gmarkall: Following through the code in later PRs, it looks like the placement new is no longer needed (in functions like `replace()` etc in `udf_cpp/shim.cu` because the `udf_string` pointed-to by `st_ptr` was constructed by the 

```cuda
__device__ inline udf_string replace(string_view source,
                                     string_view target,
                                     string_view replacement)
```

function in `replace.cuh`, and overwrites the data that would have been put there by the placement new - @brandon-b-miller Do you see this flow through the code in the code in its entirety in #13181?

python/cudf/udf_cpp/shim.cu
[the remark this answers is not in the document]
brandon-b-miller: Yes, this was just to do a single allocation rather than two.

python/cudf/udf_cpp/shim.cu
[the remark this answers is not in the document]
wence-: > 



> IIUC `udf_str` will have been constructed with a placement `new` where the call to it originates from Numba-generated code when lowering functions like `replace()`. Would it help in the context of this PR if the comment / documentation were explicit that `udf_str` should be pointing to a fully-initialized `udf_string`?

I think I'm happy that `udf_str` is fully initialised. But is `*st_ptr` a valid `udf_string` object after we've just done a `memcpy` into it? If it were a simple C struct I think the answer would be yes, but it's not a simple C struct I think.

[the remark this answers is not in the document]
davidwendt: Yes, it is valid.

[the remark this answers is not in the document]
wence-: Thanks @davidwendt.

python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +664,53 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
...
 #undef make_definition_idx
 }
+
+/*
+NRT CUDA functions
+*/
+
+__device__ NRT_MemSys TheMSys = {
+  .allocator = {(NRT_malloc_func)malloc_wrapper, NULL, (NRT_free_func)free_wrapper},
+  .stats     = {0, 0, 0, 0}};
```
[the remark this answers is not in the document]
gmarkall: Becase `malloc` is special - I haven't looked into the underlying implementation, but it's not a regular function. Taking its address directly results in a `ptxas` error:

```
ptxas fatal   : Cannot take address of function 'malloc'
```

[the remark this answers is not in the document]
wence-: Ah ok, thanks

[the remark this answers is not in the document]
davidwendt: Maybe worth a comment in the code for this. Perhaps where the wrappers are defined.

gmarkall, line 674: `stats` has 5 members - should this be:

```suggestion
  .stats     = {false, 0, 0, 0, 0}};
```

?

brandon-b-miller:
I've had some time to get back to this recently and wanted to summarize the state of play. When I was working on this a while ago, I encountered a bug in the memory allocation process that manifests under `compute-sanitizer` when running the `string_udf` tests in downstream PRs. From debugging done with @wence- and @gmarkall it looked like something was going wrong with the members of the `MemSys` struct that was causing them to appear uninitialized but we never fully narrowed it down.

We've noticed though that the failures go away when building cuDF with a cuda 12 environment which was not available at the time. I'm able to run the full string UDF test suite under `compute-sanitizer`, invoking the code we're adding in this PR,  with no errors. For me this provides at least a little peace of mind that we're not relying on something that's subtly illegal, as well as a direction towards which we can make progress in debugging while we merge the rest of the downstream work. 

I've updated the feature branch to the latest 23.10 and merged the latest into this PR as well. I will finish up addressing the remaining reviews that are leftover from the last round and then if it seems reasonable to the original reviewers, I'd like to suggest we merge to the feature branch and continue our cuda 11 `compute-sanitizer` debugging efforts in parallel with reviewing the downstream work.

python/cudf/udf_cpp/shim.cu
```diff
@@ -30,6 +30,8 @@
 #include <limits>
 #include <type_traits>
 
+#include <_nrt_cuda.cuh>
```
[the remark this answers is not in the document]
brandon-b-miller: I renamed this.

python/cudf/udf_cpp/shim.cu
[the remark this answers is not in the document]
brandon-b-miller: I added a comment about this.

python/cudf/udf_cpp/nrt/numba_cuda_runtime.cuh
```diff
@@ -0,0 +1,129 @@
+/*
+* SPDX-FileCopyrightText: Copyright (c) <2023> NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
bdice, line 2: The `<year>` is a placeholder. No literal `<>` needed.
```suggestion
* SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

wence- (approved):

gmarkall (approved):
This seems fine to merge to me - I have a couple of comments on the diff, but they're something we could just address / think about for a future PR rather than modifying this one again.

python/cudf/udf_cpp/nrt/numba_cuda_runtime.cuh
```diff
@@ -0,0 +1,129 @@
...
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef _NRT_CUDA_H
+#define _NRT_CUDA_H
```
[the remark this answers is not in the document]
wence-: Is it still the case that `#pragma once` provokes quadratic behaviour in GCC (it did in at least up to the 7 series IIRC).

python/cudf/udf_cpp/nrt/numba_cuda_runtime.cuh
```diff
@@ -0,0 +1,129 @@
+/*
+* SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+* SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
brandon-b-miller: Closing the loop here since there's been some offline discussion. There's a pending question here about how to distribute this license in a compliant manner since it won't be present in the PTX this file compiles to. We decided to leave this as a TODO so that it won't block technical work for now, but we will need to resolve this before we merge this to dev and start distributing nightlies.

PR #13062, opened 2023-04-04
Allow passing different types of physical element comparator to `cudf::detail::sorted_order`

ttnghia:
The experimental lexicographic comparator allows specifying different types of physical element comparator so that `NaN` values can be compared by:
 * Following the IEEE-754 standard: comparison involving a `NaN` produces undefined output, or
 * A `NaN` value is always equivalent to other `NaN`s and greater than all other non-`NaN` values.

This PR extends the detail version of `sorted_order` so that we can also pass in a type of physical element comparator. This is necessary if we want to use `detail::sorted_order` in places where different types of physical element comparator can be specified.

An immediate usage of this is to implement a customized version of `cudf::rank` to fully support nested types in lexicographic comparison.

### DIFF
--- a/cpp/src/sort/sort.cu
+++ b/cpp/src/sort/sort.cu
@@ -15,6 +15,7 @@
  */
 
 #include <cudf/column/column.hpp>
+#include <cudf/detail/gather.hpp>
 #include <cudf/detail/nvtx/ranges.hpp>
 #include <cudf/detail/sorting.hpp>
 #include <cudf/sorting.hpp>
--- a/cpp/src/sort/sort_column_impl.cuh
+++ b/cpp/src/sort/sort_column_impl.cuh
@@ -36,6 +36,7 @@ namespace detail {
  * This API offers fast sorting for primitive types. It cannot handle nested types and will not
  * consider `NaN` as equivalent to other `NaN`.
  *
+ * @tparam stable Whether to use stable sort
  * @param input Column to sort. The column data is not modified.
  * @param column_order Ascending or descending sort order
  * @param null_precedence How null rows are to be ordered
--- a/cpp/src/sort/sort_column_impl.cuh
+++ b/cpp/src/sort/sort_column_impl.cuh
@@ -33,6 +33,9 @@ namespace detail {
 /**
  * @brief Sort indices of a single column.
  *
+ * This API offers fast sorting for primitive types. It cannot handle nested types and will not
+ * consider `NaN` as equivalent to other `NaN`.
+ *
  * @param input Column to sort. The column data is not modified.
  * @param column_order Ascending or descending sort order
  * @param null_precedence How null rows are to be ordered
--- a/cpp/src/sort/sort_column_impl.cuh
+++ b/cpp/src/sort/sort_column_impl.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021-2022, NVIDIA CORPORATION.
+ * Copyright (c) 2021-2023, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


cpp/src/sort/sort_column_impl.cuh
```diff
@@ -16,14 +16,68 @@
...
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ * @return Sorted indices for the input column.
+ */
+template <bool stable>
+std::unique_ptr<column> sorted_order(column_view const& input,
+                                     order column_order,
+                                     null_order null_precedence,
+                                     rmm::cuda_stream_view stream,
+                                     rmm::mr::device_memory_resource* mr);
```
ttnghia, line 53: This is moved from `sort_impl.cuh`.

cpp/src/sort/sort_column_impl.cuh
```diff
@@ -16,14 +16,68 @@
...
+
+/**
+ * @brief Comparator functor needed for single column sort.
+ *
+ * @tparam Column element type.
+ */
+template <typename T>
+struct simple_comparator {
```
ttnghia, line 61: This is moved from `sort_impl.cuh`.

cpp/src/sort/sort.cu
```diff
@@ -14,15 +14,16 @@
  * limitations under the License.
  */
 
+#include <sort/sort_impl.cuh>
```
vyasr, line 17: I think this should jus be 
```suggestion
#include "sort_impl.cuh"
```

cpp/src/sort/stable_sort.cu
```diff
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-#include "sort_impl.cuh"
+#include <sort/sort_impl.cuh>
```
vyasr, line 17: Why was this include changed? I think the original version was correct.

vyasr: Specifically
> Use <> instead of "" unless the header is in the same directory as the source file.

from the [developer guide](https://docs.rapids.ai/api/libcudf/stable/developer_guide).

cpp/src/sort/sort_impl.cuh
```diff
@@ -16,76 +16,21 @@
...
-#include <cudf/detail/utilities/vector_factories.hpp>
-#include <cudf/table/experimental/row_operators.cuh>
-#include <cudf/table/table_device_view.cuh>
-#include <cudf/utilities/error.hpp>
-#include <cudf/utilities/traits.hpp>
-
-#include <rmm/cuda_stream_view.hpp>
-#include <rmm/device_uvector.hpp>
-#include <rmm/exec_policy.hpp>
+#include <sort/sort_column_impl.cuh>
```
vyasr, line 19:
```suggestion
#include "sort_column_impl.cuh"
```

cpp/src/sort/sort_column_impl.cuh
vyasr: I'm not sure that I see the benefit here. Before, `sort_column_impl.cuh` included `sort_impl.cuh`. Now, the situation is reversed. Why is one better than the other?

vyasr (changes requested):

cpp/src/sort/sort_column_impl.cuh
ttnghia: `sort_impl.cuh` is sorting for table, which in turn calls sorting for column `sort_column_impl.cuh`. Thus, logically the order should be `sort_impl.cuh` include `sort_column_impl.cuh`.

Similarly, `(stable_)sort_column.cu` include `sort_column_impl.cuh`. We don't need them to include `sort_impl.cuh` in the dependency chain.

vyasr (approved):

davidwendt (approved):

PR #13063, opened 2023-04-04
Prevent overflow with `skip_rows` in ORC and Parquet readers

vuule:
## Description
Use int64_t for `skip_rows` since source or combined sources can have more than two billion rows, and we should be able to read a range of rows even in that case.
Store `num_rows` as `std::optional`, instead of using special value (`-1`).
Reuse code with error-prone logic between ORC and Parquet.
Added unit tests for the tricky code above.

## Checklist


galipremsagar (approved):

brandon-b-miller (approved):

cpp/include/cudf/io/orc.hpp
```diff
@@ -178,7 +178,8 @@ class orc_reader_options {
   void set_stripes(std::vector<std::vector<size_type>> stripes)
   {
     CUDF_EXPECTS(stripes.empty() or (_skip_rows == 0), "Can't set stripes along with skip_rows");
-    CUDF_EXPECTS(stripes.empty() or (_num_rows == -1), "Can't set stripes along with num_rows");
+    CUDF_EXPECTS(stripes.empty() or not _num_rows.has_value(),
+                 "Can't set stripes along with num_rows");
```
PointKernel, line 182: If stripes are empty, the code will throw in both exception handling. Do you mean `and` here?
```suggestion
    CUDF_EXPECTS(stripes.empty() and not _num_rows.has_value(),
                 "Can't set stripes along with num_rows");
```

Also, docs to be updated:
```cpp
  * @throw If no rows are skipped or stripes are empty
  * @throw If all rows need to be read but stripes are empty
```

cpp/include/cudf/io/orc.hpp
```diff
@@ -187,8 +188,9 @@ class orc_reader_options {
    *
    * @param rows Number of rows
    */
-  void set_skip_rows(size_type rows)
+  void set_skip_rows(int64_t rows)
   {
+    CUDF_EXPECTS(rows >= 0, "skip_rows cannot be negative");
     CUDF_EXPECTS(rows == 0 or _stripes.empty(), "Can't set both skip_rows along with stripes");
```
PointKernel, line 194: docs for `throw`

cpp/include/cudf/io/orc.hpp
```diff
@@ -200,7 +202,8 @@ class orc_reader_options {
    */
   void set_num_rows(size_type nrows)
   {
-    CUDF_EXPECTS(nrows == -1 or _stripes.empty(), "Can't set both num_rows along with stripes");
+    CUDF_EXPECTS(nrows >= 0, "num_rows cannot be negative");
+    CUDF_EXPECTS(_stripes.empty(), "Can't set both num_rows along with stripes");
```
PointKernel, line 206: docs for `throw`

cpp/include/cudf/io/parquet.hpp
```diff
@@ -58,9 +58,9 @@ class parquet_reader_options {
   // List of individual row groups to read (ignored if empty)
```
PointKernel, line 58: `throw` docs to be added for multiple APIs across this file

cpp/src/io/orc/aggregate_orc_metadata.cpp
```diff
@@ -151,22 +153,29 @@ aggregate_orc_metadata::aggregate_orc_metadata(
   }
 }
 
-std::vector<metadata::stripe_source_mapping> aggregate_orc_metadata::select_stripes(
+std::tuple<int64_t, size_type, std::vector<metadata::stripe_source_mapping>>
```
PointKernel, line 156: Not necessarily a change request, but a struct would be more descriptive than a tuple

cpp/src/io/orc/aggregate_orc_metadata.hpp
```diff
@@ -62,7 +62,7 @@ class aggregate_orc_metadata {
 
  public:
   std::vector<metadata> per_file_metadata;
-  size_type const num_rows;
+  int64_t const num_rows;
```
PointKernel, line 65: Just to make sure I understand this correctly, the total number of rows must be `int64_t` since it may exceed 2B rows while the per stripe num of rows is still using `int32_t` (or `cudf::size_type`) since we have a hard limit of the stripe size?

cpp/src/io/orc/reader_impl.cu
```diff
@@ -933,8 +933,8 @@ std::unique_ptr<table> reader::impl::compute_timezone_table(
     {}, selected_stripes[0].stripe_info[0].second->writerTimezone, stream);
 }
 
-table_with_metadata reader::impl::read(size_type skip_rows,
-                                       size_type num_rows,
+table_with_metadata reader::impl::read(int64_t skip_rows,
```
PointKernel, line 936: Worth creating a `cudf::io::row_index_type` (please suggest a better name) alias instead of manually switching to `int64_t` all over the place.

cpp/include/cudf/io/parquet.hpp
```diff
@@ -136,14 +136,14 @@ class parquet_reader_options {
...
-  [[nodiscard]] size_type get_skip_rows() const { return _skip_rows; }
+  [[nodiscard]] int64_t get_skip_rows() const { return _skip_rows; }
 
   /**
    * @brief Returns number of rows to read.
    *
    * @return Number of rows to read
    */
-  [[nodiscard]] size_type get_num_rows() const { return _num_rows; }
+  [[nodiscard]] std::optional<size_type> const& get_num_rows() const { return _num_rows; }
```
PointKernel, line 146: Docs to be updated since the return type is changed.

cpp/src/io/orc/aggregate_orc_metadata.cpp
vuule: I modified it to be the same as the equivalent code in Parquet. I also prefer struct for return types, but it this case the benefit would be very limited. We only use the return values in a structured binding.

cpp/include/cudf/io/orc.hpp
vuule: `or` is correct here. The failure case is when `not stripes.empty() and _num_rows.has_value()`. We need to expect the opposite here, which is the expression above.
I'll add `@throw` details to the docs👍

cpp/src/io/orc/aggregate_orc_metadata.cpp
PointKernel: > We only use the return values in a structured binding.

Ah, good point.

cpp/src/io/orc/aggregate_orc_metadata.hpp
vuule: Good question
That got me looking into the ORC specs:
```
 // the number of rows in the stripe
 optional uint64 numberOfRows = 5;
 ```
So in theory we could have a stripe with >2B rows. I can look into this, but maybe it's better to leave it for a separate PR.

File: cpp/src/io/orc/reader_impl.cu (hunk 7 above)
Comment (vuule) [reply cpp/src/io/orc/reader_impl.cu:936]: Well. I tried to do this and ran into an issue with our file/include structure. This type would be needed on both `/include` and `/src` side, but there are no header files that are included in both.
I think we need to create a new header that only contains ORC "specs" and declare the type(s) there. I think this can be done in the same PR as https://github.com/rapidsai/cudf/pull/13063#discussion_r1160948892

Comment (PointKernel) [reply cpp/src/io/orc/reader_impl.cu:936]: Works for me. Thanks for looking into this.

File: cpp/include/cudf/io/orc.hpp [hunk 9]
```diff
@@ -58,9 +58,9 @@ class orc_reader_options {
   // List of individual stripes to read (ignored if empty)
   std::vector<std::vector<size_type>> _stripes;
   // Rows to skip from the start;
-  size_type _skip_rows = 0;
-  // Rows to read; -1 is all
-  size_type _num_rows = -1;
+  uint64_t _skip_rows = 0;
```
vyasr, line 61: Should we use an `int64_t` instead of a `uint64_t` for this? If we don't absolutely need the extra power of 2 then it might be better to stick with a signed integer since we do that for `size_type`.

cpp/include/cudf/io/orc.hpp
```diff
@@ -197,10 +207,14 @@ class orc_reader_options {
...
    * @param nrows Number of rows
+   *
+   * @throw cudf::logic_error if a negative value is passed
+   * @throw cudf::logic_error if stripes have been previously set
    */
   void set_num_rows(size_type nrows)
   {
-    CUDF_EXPECTS(nrows == -1 or _stripes.empty(), "Can't set both num_rows along with stripes");
+    CUDF_EXPECTS(nrows >= 0, "num_rows cannot be negative");
+    CUDF_EXPECTS(_stripes.empty(), "Can't set both num_rows along with stripes");
```
vyasr, line 217:
```suggestion
    CUDF_EXPECTS(_stripes.empty(), "Can't set both num_rows and stripes");
```

cpp/include/cudf/io/parquet.hpp
```diff
@@ -58,9 +58,9 @@ class parquet_reader_options {
   // List of individual row groups to read (ignored if empty)
   std::vector<std::vector<size_type>> _row_groups;
   // Number of rows to skip from the start
-  size_type _skip_rows = 0;
-  // Number of rows to read; -1 is all
-  size_type _num_rows = -1;
+  int64_t _skip_rows = 0;
```
vyasr, line 61: Why is this an `int64_t` when the orc uses `uint64_t`? Sticking to signed in both cases is probably preferable.

cpp/src/io/orc/aggregate_orc_metadata.cpp
```diff
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
-#include "aggregate_orc_metadata.hpp"
+#include <io/orc/aggregate_orc_metadata.hpp>
```
vyasr, line 17: We prefer quoted includes when it's in the same directory (the way it was before), right?

cpp/src/io/utilities/row_selection.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace cudf::io::detail {
+
+std::pair<uint64_t, size_type> skip_rows_num_rows_from_options(
+  uint64_t skip_rows_opt, std::optional<size_type> const& num_rows_opt, uint64_t num_source_rows)
+{
+  auto const rows_to_skip = std::min(skip_rows_opt, num_source_rows);
+  if (not num_rows_opt.has_value()) {
+    CUDF_EXPECTS(num_source_rows - rows_to_skip <= std::numeric_limits<size_type>::max(),
+                 "ORC reader can't read all rows from its input(s)");
```
vyasr, line 32: Clarifying the wording, the problem isn't that the user is necessarily trying to read all the lines, it's that the number of lines requested exceeds the cudf size_type limit
```suggestion
                 "The requested number of rows to read exceeds the largest cudf column size");
```

cpp/src/io/utilities/row_selection.cpp
```diff
@@ -0,0 +1,41 @@
...
+    return {rows_to_skip, num_source_rows - rows_to_skip};
+  }
+  // Limit the number of rows to the end of the input
+  return {rows_to_skip,
+          static_cast<size_type>(
+            std::min<uint64_t>(num_rows_opt.value(), num_source_rows - rows_to_skip))};
```
vyasr, line 38: Template param should be inferred, right?
```suggestion
            std::min(num_rows_opt.value(), num_source_rows - rows_to_skip))};
```

cpp/src/io/utilities/row_selection.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <io/utilities/row_selection.hpp>
+
+#include <cudf/utilities/error.hpp>
+
+#include <algorithm>
+#include <limits>
+
+namespace cudf::io::detail {
+
+std::pair<uint64_t, size_type> skip_rows_num_rows_from_options(
```
vyasr, line 26: One more instance of uint vs int.

cpp/src/io/utilities/row_selection.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <utility>
+
+namespace cudf::io::detail {
+/**
+ *@brief TODO.
+ *@param skip_rows_opt TODO
+ *@param num_rows_opt TODO
+ *@param num_source_rows TODO
+ *@return A std::pair containing the number of rows to skip and the number of rows to read
+ * @throw TODO
```
vyasr, line 31: Time TODO!!

cpp/src/io/parquet/reader_impl_helpers.cpp
```diff
@@ -336,54 +338,49 @@ std::vector<std::string> aggregate_reader_metadata::get_pandas_index_names() con
...
-  size_type row_count) const
+  int64_t skip_rows_opt,
+  std::optional<size_type> const& num_rows_opt) const
 {
   std::vector<row_group_info> selection;
+  auto [rows_to_skip, rows_to_read] = [&]() {
+    if (not row_group_indices.empty()) { return std::pair<int64_t, size_type>{}; }
+    auto const from_opts = cudf::io::detail::skip_rows_num_rows_from_options(
+      skip_rows_opt, num_rows_opt, get_num_rows());
+    return std::pair{static_cast<int64_t>(from_opts.first), from_opts.second};
```
vyasr, line 352: It looks like we can't return directly here solely because we need to cast the uint to int, is that right?

cpp/src/io/parquet/reader_impl_helpers.cpp
```diff
@@ -336,54 +338,49 @@ std::vector<std::string> aggregate_reader_metadata::get_pandas_index_names() con
...
-    return {row_start, row_count, std::move(selection)};
-  }
-
-  row_start = std::max(row_start, 0);
-  if (row_count < 0) {
-    row_count = std::min(get_num_rows(), std::numeric_limits<size_type>::max());
-  }
-  row_count = std::min(row_count, get_num_rows() - row_start);
-  CUDF_EXPECTS(row_count >= 0, "Invalid row count");
-  CUDF_EXPECTS(row_start <= get_num_rows(), "Invalid row start");
```
vyasr, line 372: I guess this is the logic now contained in `skip_rows_num_rows_from_options`?

cpp/src/io/orc/reader_impl.cu
```diff
@@ -956,7 +956,8 @@ table_with_metadata reader::impl::read(size_type skip_rows,
     return {std::make_unique<table>(), std::move(out_metadata)};
 
   // Select only stripes required (aka row groups)
-  const auto selected_stripes = _metadata.select_stripes(stripes, skip_rows, num_rows, stream);
+  const auto [rows_to_skip, rows_to_read, selected_stripes] =
```
vyasr, line 959: May as well fix while we're here
```suggestion
  auto const [rows_to_skip, rows_to_read, selected_stripes] =
```

python/cudf/cudf/_lib/orc.pyx
```diff
@@ -341,11 +341,12 @@ cdef orc_reader_options make_orc_reader_options(
         orc_reader_options.builder(src)
         .stripes(strps)
         .skip_rows(skip_rows)
-        .num_rows(num_rows)
         .timestamp_type(data_type(timestamp_type))
         .use_index(use_index)
         .build()
     )
+    if num_rows >= 0:
+        opts.set_num_rows(num_rows)
```
vyasr, line 349: Why the switch to calling the function instead of using the option?

cpp/tests/io/row_selection_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+  }
+}
+
+TEST_F(FromOptsTest, InputSize32BitOverflow)
+{
+  // Input number of rows too large to fit into cudf::size_type
+  // Test that we can still select rows from such input
+  auto const too_large_for_32bit = std::numeric_limits<int64_t>::max();
+
+  // no num_row
```
vyasr, line 77:
```suggestion
  // no num_rows
```

vyasr (changes requested):

cpp/include/cudf/io/orc.hpp
vuule: The ORC files store the number of rows as uint64.
I understand that it's unlikely that an ORC file with over int64::max rows would ever exist. The idea is to match the specs (hoping to at some point add a header with ORC types that we can use in places like this).

cpp/include/cudf/io/parquet.hpp
vuule: Same as with ORC, we're matching the Parquet specs, where the row count is stored as signed int64.
Unless you think this is very error-prone, I'd like us to start using the types as they are in the format specs. Having aliases for these would also go a long way to make the code more readable.

cpp/src/io/utilities/row_selection.hpp
vuule: :grimacing:

cpp/src/io/utilities/row_selection.cpp
vuule: We need to specify the template type. `num_rows_opt.value()` is `size_type` and `num_source_rows - rows_to_skip` is `uint64_t`.

python/cudf/cudf/_lib/orc.pyx
vuule: Because `num_rows` is now stored as `std::optional`. When the `optional` is `nullopt`, it denotes that user did not pass this option.

vuule: Because we don't want to set the option when user did not pass a value. Python layer passes -1 as "not set" (same as what we used before I switched to std::optional), which is why we use the `>= 0`condition here.

cpp/include/cudf/io/orc.hpp
vyasr: I was not aware that this was matching the spec. After some offline discussion, the general consensus on the team was that it was better to match the spec here, we should just add some comments indicating that is why we use this type.

PR #13066, opened 2023-04-05
Fix tests/identify_stream_usage.cpp

ahendriksen:
## Description
The identify_stream_usage test uses `strcmp` but not does not include `<cstring>`. This PR fixes that.

The missing include was surfaced by #13064, showing that the test relied on headers in `spdlog` to include `cstring`.


ahendriksen:
Thanks! That was quick. I thought the 'alphabetizing includes' was an automated commit.

PR #13069, opened 2023-04-05
Update `struct_minmax_util` to experimental row comparator

divyegala:
## Description
Par of #11844 

## Checklist


ttnghia:
I've already worked on it: https://github.com/rapidsai/cudf/pull/10811.

It was stalled because of performance regression. Still need to wait for more performance optimization before continuing.

divyegala:
> Still need to wait for more performance optimization before continuing.

@ttnghia 

@GregoryKimball and I discussed this and the performance regressions don't look significant to us. If you prefer, can you bring your PR up-to-date and we can run a set of microbenchmarks on it to determine if the minor cost is just a single run and amortized over multiple runs?

ttnghia:
I would prefer to do it after all the optimization mentioned above came in. By doing so we can have the performance regression (if any) being minimized.

vyasr:
I asked about moving forward on Nghia's PR right around when this one got made it looks like https://github.com/rapidsai/cudf/pull/10811#issuecomment-1496797197. Can we discuss a concrete plan for moving forward here? 
- What benchmarks do we need to see? The other PR already has some reported.
- What performance numbers do we need to hit?
- Have we done any investigation at all to understand the causes of the regression?

cpp/src/reductions/struct_minmax_util.cuh
```diff
@@ -21,7 +21,7 @@
 #include <cudf/detail/utilities/device_operators.cuh>
 #include <cudf/detail/utilities/vector_factories.hpp>
 #include <cudf/reduction/detail/reduction_operators.cuh>
-#include <cudf/table/row_operators.cuh>
+#include <cudf/table/experimental/row_operators.cuh>
```
ttnghia, line 24: Also please remove the header at line 20.

cpp/src/reductions/struct_minmax_util.cuh
```diff
@@ -35,18 +35,14 @@ namespace detail {
...
 
-  row_arg_minmax_fn(table_device_view const& table,
-                    bool has_nulls,
-                    null_order const* null_precedence,
-                    bool const is_arg_min)
-    : num_rows(table.num_rows()),
-      comp(nullate::DYNAMIC{has_nulls}, table, table, nullptr, null_precedence),
-      is_arg_min(is_arg_min)
+  row_arg_minmax_fn(size_type num_rows_, DeviceComparator comp_, bool const is_arg_min_)
+    : num_rows(num_rows_), comp(comp_), is_arg_min(is_arg_min_)
```
ttnghia, line 45:
```suggestion
    : num_rows{num_rows_}, comp{std::move(comp_)}, is_arg_min{is_arg_min_}
```

cpp/src/reductions/struct_minmax_util.cuh
ttnghia: There are also a lot of unused headers such as at line 28-29. Please also remove them. See https://github.com/rapidsai/cudf/pull/10811/files#diff-7f4becb7510a70126c28187c4ecfdfe6498dcf8ee897be829b50b403f4868fbcL21-L29.

cpp/src/reductions/struct_minmax_util.cuh
```diff
@@ -87,49 +83,39 @@ auto static constexpr DEFAULT_NULL_ORDER = cudf::null_order::BEFORE;
...
-      d_flattened_input_ptr{
-        table_device_view::create(flattened_input->flattened_columns(), stream)},
-      is_min_op(is_min_op),
-      has_nulls{has_nested_nulls(table_view{{input}})},
-      null_orders_dvec(0, stream)
+  rmm::cuda_stream_view stream;
+
+  comparison_binop_generator(column_view const& input_,
+                             rmm::cuda_stream_view stream_,
+                             bool is_min_op_)
```
ttnghia, line 94: Let's `stream` be the last param.

cpp/src/reductions/struct_minmax_util.cuh
```diff
@@ -87,49 +83,39 @@ auto static constexpr DEFAULT_NULL_ORDER = cudf::null_order::BEFORE;
...
+    : input_tview{cudf::table_view{{input_}}},
+      is_min_op(is_min_op_),
+      has_nulls{cudf::has_nested_nulls(cudf::table_view{{input_}})},
+      null_orders{std::vector<null_order>{DEFAULT_NULL_ORDER}},
```
ttnghia, line 98: I think this is the reason that fails many tests in this PR. In the meantime I don't have a clear way to fix it. Need to think further.

ttnghia:
@vyasr My potential optimization mentioned earlier has failed thus I don't block this anymore.

> * What benchmarks do we need to see? The other PR already has some reported.
Since this is touching struct min/max, we should run the benchmark on struct min/max on groupby scan and reduction. I ran such benchmark here so this PR just need to reran them: https://github.com/rapidsai/cudf/pull/10811.

> * What performance numbers do we need to hit?
We should not have significant performance regression with the benchmark above.

> * Have we done any investigation at all to understand the causes of the regression?
The cause of regression before was probably due to register usage as I investigated: https://github.com/rapidsai/cudf/pull/10811#issuecomment-1125576034.

vyasr:
@ttnghia can we close #10811 and use this PR instead, or would you prefer the reverse? Is there any difference, or code that needs to be ported from one to the other?

@divyegala let's repoint this PR at 23.08, don't think there's anything urgent to get in for this release here.

ttnghia:
@vyasr sure this should closes https://github.com/rapidsai/cudf/pull/10811.

vyasr:
@divyegala I've retargeted and merged latest 23.08. There are a few comments to address, then I think this PR should be close.

cpp/src/reductions/struct_minmax_util.cuh
divyegala: Only this test is failing in this PR now. Any ideas why?
```
StructReductionTest.StructReductionMinMaxWithNulls
```

divyegala:
@vyasr @ttnghia this PR is up to date now. There are failing tests for scans/reductions when NULLs are present, although I'm not really sure how to diagnose why they are occurring. Any pointers?

cpp/src/reductions/struct_minmax_util.cuh
ttnghia: I'm not sure. Can you print out the output of that test to see what it is?

divyegala:
```
[ RUN      ] StructReductionTest.StructReductionMinMaxWithNulls
/opt/conda/conda-bld/work/cpp/tests/utilities/column_utilities.cu:289: Failure
Expected equality of these values:
  lhs_null_count
    Which is: 1
  rhs_null_count
    Which is: 0
Google Test trace:
/opt/conda/conda-bld/work/cpp/tests/reductions/reduction_tests.cpp:2565:  <--  line of failure
```

divyegala:
```

Comment (ttnghia) [reply cpp/src/reductions/struct_minmax_util.cuh]: It seems that the `min` test fails, and `aab` is considered as smallest, and nulls are excluded (nulls are larger than non-nulls).

Comment (divyegala) [reply cpp/src/reductions/struct_minmax_util.cuh]: It looks like this was already a [pre-conditon](https://github.com/rapidsai/cudf/blob/aed7174eae6c6eb38fbf186938df44f88787cf29/cpp/src/reductions/struct_minmax_util.cuh#L113-L124) for `min` where top-level NULLs were pushed to `AFTER`

Comment (ttnghia) [reply cpp/src/reductions/struct_minmax_util.cuh]: Yes. So the correct output is `{null, null}` which contains nulls only in the children, not top level.
In other word: nulls are excluded (nulls are larger than non-nulls) in the children level.

Comment (divyegala) [reply cpp/src/reductions/struct_minmax_util.cuh]: Ah I see, I don't understand why this is happening. Fault with code I wrote or lex comparator?

Comment (ttnghia) [reply cpp/src/reductions/struct_minmax_util.cuh]: The previous approach flattens the input structs column into a table. The null order here will have size equal to number of table columns:
```
  if (is_min_op) {
      null_orders = flattened_input->null_orders();
```
So we could just change the first null order element and the top level null order will be different from all children level null order.

In the new approach, you always have just one input column and always have the null order array having one element. The row comparator will then use that unique null order for all top level and children level.

 I can think of only one solution to overcome this: continue flattening the input and modify the null order vector, then pass everything into the row comparator.

Comment (ttnghia) [reply cpp/src/reductions/struct_minmax_util.cuh]: And that is only needed for `min`. For `max`, all null order values are the same.

File: cpp/src/groupby/sort/group_scan_util.cuh [hunk 5]
```diff
@@ -26,6 +26,7 @@
 #include <cudf/detail/gather.hpp>
 #include <cudf/detail/iterator.cuh>
 #include <cudf/detail/null_mask.hpp>
+#include <cudf/detail/structs/utilities.hpp>
```
ttnghia, line 29: Why?

cpp/src/reductions/scan/scan_inclusive.cu
```diff
@@ -23,6 +23,7 @@
 #include <cudf/detail/gather.hpp>
 #include <cudf/detail/iterator.cuh>
 #include <cudf/detail/null_mask.hpp>
+#include <cudf/detail/structs/utilities.hpp>
```
ttnghia, line 26: Similar question here. Why do we need this?

cpp/src/groupby/sort/group_scan_util.cuh
divyegala: https://github.com/rapidsai/cudf/blob/dc3c1a884363c97e8a3232dd0e3b928937bf0d0c/cpp/src/groupby/sort/group_scan_util.cuh#L230

This was being transitively included before

cpp/src/reductions/scan/scan_inclusive.cu
divyegala: Same as above

cpp/src/reductions/struct_minmax_util.cuh
```diff
@@ -87,66 +78,82 @@ auto static constexpr DEFAULT_NULL_ORDER = cudf::null_order::BEFORE;
...
+                      is_min_op_,
+                      flattened_tview = flattened_input->flattened_columns(),
+                      orders          = null_orders,
+                      stream_]() {
+        if (is_min_op_) {
+          return cudf::experimental::row::lexicographic::self_comparator{
+            flattened_tview, {}, orders, stream_};
+        } else {
+          return cudf::experimental::row::lexicographic::self_comparator{
+            cudf::table_view{{input_}}, {}, orders, stream_};
```
ttnghia, line 129:
```suggestion
            cudf::table_view{{input_tview}}, {}, orders, stream_};
```

cpp/src/reductions/struct_minmax_util.cuh
```diff
@@ -87,66 +78,82 @@ auto static constexpr DEFAULT_NULL_ORDER = cudf::null_order::BEFORE;
...
-    }
-    // else: Don't need to generate nulls order to copy to device memory if we have all null orders
-    // are BEFORE (that happens when we have is_min_op == false).
   }
 
  public:
   auto binop() const
   {
-    return row_arg_minmax_fn(*d_flattened_input_ptr, has_nulls, null_orders_dvec.data(), is_min_op);
+    auto const device_comp = row_comparator.less<true>(cudf::nullate::DYNAMIC{has_nulls});
```
ttnghia, line 138: Does this mean we use `.less` all the time, even if we are finding `max`?

cpp/src/groupby/sort/group_scan_util.cuh
ttnghia: Oh but now `struct_minmax_util` already included it so we can remove this now.

cpp/src/reductions/struct_minmax_util.cuh
ttnghia: Then capture like this:
```
[input_tview = input_tview,...]
```

ttnghia (approved):

vyasr (changes requested):
Perhaps I'm missing something: I see that you removed the old tests verifying that this code path failed loudly, but shouldn't we add new tests verifying the now supported behavior?

vyasr (approved):
Approving after some offline discussions. The main new feature added by this PR is nesting lists with structs (whereas previous only pure struct nesting was supported), but in other such PRs we have not added testing of the list case either. We're relying on existing tests of the comparator to validate those specific different cases.

PR #13071, opened 2023-04-05
Assert for non-empty nulls

razajafri:
Added assert check to ColumnView creation to ensure there are no non-empty nulls at the time of creation. This also contains changes to bypass this check by setting `-da:ai.rapids.cudf.AssertEmptyNulls`. 

## Checklist


java/pom.xml
```diff
@@ -158,6 +158,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <junit.version>5.4.2</junit.version>
         <ai.rapids.refcount.debug>false</ai.rapids.refcount.debug>
+        <ai.rapids.allow.nonempty.nulls>true</ai.rapids.allow.nonempty.nulls>
```
ttnghia, line 161: `nulls` should stay first, as we are talking about null. So this should be `nulls.noempty` instead.

ttnghia: Or, this should be better: `ai.rapids.allow.nonempty_nulls`.

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
```diff
@@ -0,0 +1,18 @@
...
+  /*
+  We have to read the system property to see if we should by-pass this assert. Ideally we should
+  just use `-da:AssertNonEmptyNulls` as per Java standard, but the reason for doing this is because
+  surefire-plugin doesn't expose anyway to disable assertions at the class level or package level
+  instead we can only disable asserts at the ClassLoader level. Therefore, in order for the tests to
+  pass we need to set a flag from the pom to know we are running unit-tests to allow non-empty nulls
+ */
+  private static final boolean ALLOW_NON_EMPTY_NULLS =
+      Boolean.getBoolean("ai.rapids.allow.nonempty.nulls");
+  public static void assertHasEmptyNulls(ColumnView cv) {
```
ttnghia, line 13: `HasEmptyNulls` means the column has at least one empty nulls. We should make sure it has all empty nulls.
```suggestion
  public static void assertEmptyNulls(ColumnView cv) {
```

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
```diff
@@ -0,0 +1,18 @@
...
+    if (!ALLOW_NON_EMPTY_NULLS) {
+      assert !cv.hasNonEmptyNulls() : "Column has non-empty nulls";
```
ttnghia, line 15: I would recommend calling `cv.hasNonEmptyNulls()` only if the input is of type STRING or LIST. This would avoid unnecessarily executing native method if `cv` is not of type STRING/LIST.

razajafri:
@revans2 

Would you mind taking a look at this when you have time. This addresses the issue of imported data having non-empty nulls. I just had to add another system property for the unit-tests to pass. PTAL

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
```diff
@@ -0,0 +1,18 @@
+package ai.rapids.cudf;
+
+public class AssertEmptyNulls {
+  /*
+  We have to read the system property to see if we should by-pass this assert. Ideally we should
+  just use `-da:AssertNonEmptyNulls` as per Java standard, but the reason for doing this is because
```
revans2, line 6: I am confused by this. surefire does not let us disable assertions at a class level?

Is argLine not enough? https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#argLine or is there something else that I am missing.  For the scala testing I think there is something similar.

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
```diff
@@ -0,0 +1,18 @@
...
+  surefire-plugin doesn't expose anyway to disable assertions at the class level or package level
+  instead we can only disable asserts at the ClassLoader level. Therefore, in order for the tests to
+  pass we need to set a flag from the pom to know we are running unit-tests to allow non-empty nulls
+ */
+  private static final boolean ALLOW_NON_EMPTY_NULLS =
+      Boolean.getBoolean("ai.rapids.allow.nonempty.nulls");
+  public static void assertEmptyNulls(ColumnView cv) {
```
revans2, line 13: nit: The name of this feels off. assertTrue verifies that the result passed in is true.  Here assertEmptyNulls feels like it will throw an exception if there are not any EmptyNulls.

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
razajafri: > @razajafri Did you change this? cv.type.isNestedType is true for STRUCT and for LIST, but not for STRING. I think this is fine so long as strings will not mess up the experimental comparison code.

Nope, I didn't change this. I was assuming it will have the Strings. Structs can also have non-empty nulls so I assume we should test them too?

ttnghia: Struct doesn't have non-empty nulls by itself but its children can have. So yeah you should also recursively call the check on its children too.

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
razajafri: Thanks. I wasn't aware of this parameter. I was using the `enableAssertions` parameter which only supports "ea" and "da"

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
razajafri: How about `assertNullsAreEmpty()`

@ttnghia

ttnghia: Yeah that sounds good.

razajafri:
@revans2 @ttnghia 

I think I have addressed all your concerns, PTAL

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
```diff
@@ -0,0 +1,43 @@
...
+ *  that explicitly test with ColumnViews that contain non-empty nulls but more importantly, there
+ *  could be cases where an external system may not have a requirement of nulls being empty, so for
+ *  us to work with those systems, we can turn off this assert in the field.
+ */
+public class AssertEmptyNulls {
+  // This is the recommended way to check at runtime that assertions are enabled.
+  // https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html
+  private static boolean assertNullsAreEmpty = false;
+  static {
+      assert assertNullsAreEmpty = true ; // Intentional side effect
```
ttnghia, line 35:
```suggestion
      assert assertNullsAreEmpty == true ; // Intentional side effect
```

ttnghia: Or:
```suggestion
      assert assertNullsAreEmpty; // Intentional side effect
```

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  public static void assertNullsAreEmpty(ColumnView cv) {
+    if (cv.type.isNestedType() || cv.type.hasOffsets() && assertNullsAreEmpty) {
+      assert !cv.hasNonEmptyNulls() : "Column has non-empty nulls";
```
ttnghia, line 40: I think the static var `assertNullsAreEmpty` is not necessary because if assert is disabled for the entire class then both the first assert and the second assert will not run. We just need to have the second assert not run in that case.

razajafri: You are right. 

I think what I wanted to do originally was detect if asserts are off before running the unit tests. I don't know if we need to do that considering the pom has the `-da:ai.rapids.cudf.AssertEmptyNulls` but probably won't hurt to add that extra check just to be safe?

ttnghia (approved):

java/pom.xml
```diff
@@ -153,6 +153,7 @@
     </dependencies>
 
     <properties>
+        <argLine>-da:ai.rapids.cudf.AssertEmptyNulls</argLine>
```
jlowe, line 156: I'm confused what the goal of this PR is.  Java assertions are off by default.  There's an explicit enable of assertions for our unit tests run, but this appears to effectively disable the newly added assertions when we run those unit tests.  That seems to make it so we never assert in practice?

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package ai.rapids.cudf;
+
+/**
+ *  This class is a Helper class to assert there are no non-empty nulls in a ColumnView
+ *
+ *  The reason for the existence of this class is so that we can turn the asserts on/off when needed
+ *  by passing "-da:ai.rapids.cudf.AssertEmptyNulls". We need that behavior because we have tests
+ *  that explicitly test with ColumnViews that contain non-empty nulls but more importantly, there
```
revans2, line 26: This is exactly why we want the assert in place for tests. CUDF has a goal to never produce non-empty nulls if the input did not contain them.  If we or anyone else is interacting with an external system, we need to remove the non-empty nulls or risk data corruption.  This is here as a early warning to tell us that we have a problem that needs to be fixed. We should fix all of the issues to make this pass before we check it in.

revans2 (changes requested):

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
razajafri: @jlowe let's discuss on this thread. 

I get that CUDF has a goal to never produce or deal with non-empty nulls which is why we added this assert. We also have unit-tests in cudf java that introduce empty-nulls for tests. Now we either skip those tests or delete them, or we disable the asserts so we can test the CUDF functionality like `purgeNonEmptyNullsList()`. I chose to do the last option because I think that funcionality should be tested. What am I doing wrong here?

razajafri: I think alternatively we can have these unit-tests in a separate execution tag and disable the assertions only for the 4 tests. 

That sounds better I think

jlowe: That's why I was asking what the goal of the PR is.

> we disable the asserts so we can test the CUDF functionality

By disabling the asserts during all of the tests, we're effectively disabling the asserts all of the time because assertions are off by default.  In practice they're only on during unit tests.  That makes one wonder what purpose the new assertions have.

> I think alternatively we can have these unit-tests in a separate execution tag and disable the assertions only for the 4 tests.

That's one possibility, and that approach at least doesn't avoid ever exercising the new assertions.

razajafri:
CI failure has nothing to do with the changes in this PR.

revans2 (approved):
Just a nit.

java/src/main/java/ai/rapids/cudf/AssertEmptyNulls.java
```diff
@@ -0,0 +1,36 @@
...
+ *  could be cases where an external system may not have a requirement of nulls being empty, so for
+ *  us to work with those systems, we can turn off this assert in the field.
+ */
+public class AssertEmptyNulls {
+  public static void assertNullsAreEmpty(ColumnView cv) {
```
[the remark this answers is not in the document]
razajafri: Won't that imply it will assert only if NullsAreEmpty when we run the assert regardless? Referring back to your original example of `assertTrue` is not `assertIfTrue`. Do I make sense?

razajafri:
@revans2 I wanted to see what you think about this. 

Considering this is something that CUDF requires and assertions in Java aren't enabled by default, should we transition this to a hard check and throw an Exception instead of using an assert? We can do this at a later time too, I just wanted to know if we need a follow-on issue for this.

revans2:
> @revans2 I wanted to see what you think about this.
> 
> Considering this is something that CUDF requires and assertions in Java aren't enabled by default, should we transition this to a hard check and throw an Exception instead of using an assert? We can do this at a later time too, I just wanted to know if we need a follow-on issue for this.

We want the assertion off in production because checking if they are empty involves launching at least one kernel and is rather expensive. Having it be an assertion lets us verify that it is doing the right thing in tests, but we get the performance in production.  It also lets us enable it in production if we think there is a problem.  That said, in some cases where we know that we would get data corruption if we get a non-empty null, it is probably worth doing a check and then fixing the problem, at least until we feel comfortable enough that we have fixed all of the issues.

razajafri:
rebuild

PR #13061, opened 2023-04-04
Add string scalar support in AST

karthikeyann:
## Description
Adding string scalar support in AST.
A new generic scalar device view class is added in AST to support numeric, timestamp, duration and string scalars.

Register count did not change, and benchmark results are almost same.
Compile time - There is major increase in join.cu by 15%. Other files are in range of -2% to 7%
 
Addressed part of #8858

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13072, opened 2023-04-05
Add Java bindings for string literal support in AST

karthikeyann:
## Description
Depends on #13061
Add Java bindings for string scalar support in AST
Add unit test for string comparison - column vs column, column vs literal.

## Checklist


java/src/main/java/ai/rapids/cudf/ast/Literal.java
jlowe, line 1: 2023 copyrights

java/src/main/java/ai/rapids/cudf/ast/Literal.java
```diff
@@ -200,6 +200,15 @@ public static Literal ofDurationFromLong(DType type, Long value) {
     return ofDurationFromLong(type, value.longValue());
   }
 
+  /** Construct a string literal with the specified value or null. */
+  public static Literal ofString(String value) {
+    byte[] stringBytes = value.getBytes();
```
jlowe, line 205: This doesn't handle `value` being null.  Need to do something like this:
```suggestion
    if (value == null) {
      return ofNull(DType.STRING)
    }
    byte[] stringBytes = value.getBytes();
```

java/src/test/java/ai/rapids/cudf/ast/CompiledExpressionTest.java
jlowe, line 1: 2023 copyrights

java/src/test/java/ai/rapids/cudf/ast/CompiledExpressionTest.java
```diff
@@ -476,6 +476,65 @@ void testBinaryComparisonOperationTransform(BinaryOperator op, Integer[] in1, In
...
+        Arguments.of(BinaryOperator.GREATER, in1, in2, mapArray(in1, (a) -> a.compareTo(in2) > 0)),
+        Arguments.of(BinaryOperator.LESS_EQUAL, in1, in2, mapArray(in1, (a) -> a.compareTo(in2) <= 0)),
+        Arguments.of(BinaryOperator.GREATER_EQUAL, in1, in2, mapArray(in1, (a) -> a.compareTo(in2) >= 0)));
+  }
+
+  @ParameterizedTest
+  @MethodSource("createStringLiteralComparisonParams")
+  void testStringLiteralComparison(BinaryOperator op, String[] in1, String in2,
+                                               List<Boolean> expectedValues) {
+    Literal lit = Literal.ofString(in2);
```
jlowe, line 496: This doesn't test a null string literal, should add that as a test case.

I also recommend adding something like the `test*LiteralTransform` tests which are more focused on just handling the literal directly without other expressions and validates that the string data was translated properly.  The binary comparison test can hide a bad string translation and still pass the test by chance.

karthikeyann: For string column, direct transformation is not supported. only fixed width types are supported now.

jlowe (approved):

java/src/main/java/ai/rapids/cudf/ast/Literal.java
```diff
@@ -200,6 +200,18 @@ public static Literal ofDurationFromLong(DType type, Long value) {
...
   }
 
+  /** Construct a string literal with the specified value or null. */
+  public static Literal ofString(String value) {
+    if (value == null) {
+      return ofNull(DType.STRING);
+    }
+    byte[] stringBytes = value.getBytes();
+    byte[] serializedValue = new byte[stringBytes.length + Integer.BYTES];
+    ByteBuffer.wrap(serializedValue).order(ByteOrder.nativeOrder()).putInt(stringBytes.length);
```
mythrocks, line 210: :+1:

java/src/main/native/src/CompiledExpression.cpp
```diff
@@ -254,9 +263,29 @@ struct make_literal {
                                      std::move(scalar_ptr));
   }
 
+  /** Construct an AST literal from a string value */
+  template <typename T, std::enable_if_t<std::is_same_v<T, cudf::string_view>> * = nullptr>
```
mythrocks, line 267: I will look into JNI switching to using `CUDF_ENABLE_IF()`, afterwards.

java/src/main/native/src/CompiledExpression.cpp
```diff
@@ -56,12 +57,20 @@ class jni_serialized_ast {
...
   /** Read a multi-byte value from the serialized AST data buffer */
   template <typename T> T read() {
-    check_for_eof(sizeof(T));
-    // use memcpy since data may be misaligned
-    T result;
-    memcpy(reinterpret_cast<jbyte *>(&result), data_ptr, sizeof(T));
-    data_ptr += sizeof(T);
-    return result;
+    if constexpr (std::is_same_v<T, std::string>) {
+      auto const size = read<cudf::size_type>();
```
mythrocks, line 61: :+1:

mythrocks (approved):
Awesome. LGTM!

PR #13073, opened 2023-04-05
Add Python bindings for string literal support in AST

karthikeyann:
## Description
Depends on https://github.com/rapidsai/cudf/pull/13061
Add Python bindings for string scalar support in AST
Add unit test for string comparison - column vs column, column vs literal.

## Checklist

### DIFF
--- a/python/cudf/cudf/_lib/expressions.pyx
+++ b/python/cudf/cudf/_lib/expressions.pyx
@@ -1,10 +1,11 @@
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 
 from enum import Enum
 
 from cython.operator cimport dereference
 from libc.stdint cimport int64_t
 from libcpp.memory cimport make_unique, unique_ptr
+from libcpp.string cimport string
 
 from cudf._lib.cpp cimport expressions as libcudf_exp
 from cudf._lib.cpp.types cimport size_type
@@ -80,6 +81,7 @@ cdef class Literal(Expression):
         # TODO: Would love to find a better solution than unions for literals.
         cdef int intval
         cdef double doubleval
+        cdef string stringval
 
         if isinstance(value, int):
             self.c_scalar_type = scalar_type_t.INT
@@ -98,6 +100,15 @@ cdef class Literal(Expression):
             )
             self.c_obj = <expression_ptr> make_unique[libcudf_exp.literal](
                 <numeric_scalar[double] &>dereference(self.c_scalar.double_ptr)
+            )
+        elif isinstance(value, str):
+            self.c_scalar_type = scalar_type_t.STRING
+            stringval = value.encode()
+            self.c_scalar.string_ptr = make_unique[string_scalar](
+                stringval, True
+            )
+            self.c_obj = <expression_ptr> make_unique[libcudf_exp.literal](
+                <string_scalar &>dereference(self.c_scalar.string_ptr)
             )
 
 
--- a/python/cudf/cudf/core/_internals/expressions.py
+++ b/python/cudf/cudf/core/_internals/expressions.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 
 import ast
 import functools
@@ -115,7 +115,7 @@ class libcudfASTVisitor(ast.NodeVisitor):
         self.stack.append(ColumnReference(col_id))
 
     def visit_Constant(self, node):
-        if not isinstance(node, ast.Num):
+        if not isinstance(node, ast.Num) and not isinstance(node, ast.Str):
             raise ValueError(
                 f"Unsupported literal {repr(node.value)} of type "
                 "{type(node.value).__name__}"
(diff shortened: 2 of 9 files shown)


python/cudf/cudf/core/_internals/expressions.py
```diff
@@ -115,7 +115,7 @@ def visit_Name(self, node):
         self.stack.append(ColumnReference(col_id))
 
     def visit_Constant(self, node):
-        if not isinstance(node, ast.Num):
+        if not isinstance(node, ast.Num) and not isinstance(node, ast.Str):
```
vyasr, line 118: `isinstance` supports sequences:
```suggestion
        if not isinstance(node, (ast.Num, ast.Str)):
```

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -9810,6 +9810,9 @@ def df_eval():
             float,
         ),
         ("a_b_are_equal = (a == b)", int),
+        ("a > b", str),
+        ("a < '1'", str),
+        ('a == "1"', str),
```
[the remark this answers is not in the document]
karthikeyann: `+` operator on strings is not supported in both pandas and  cudf.
comparison operators is supported by both. (except the bug below).

created an issue  https://github.com/rapidsai/cudf/issues/13185 while testing all comparsion operators in cuDF.

python/cudf/cudf/_lib/expressions.pyx
```diff
@@ -80,6 +81,7 @@ cdef class Literal(Expression):
         # TODO: Would love to find a better solution than unions for literals.
         cdef int intval
         cdef double doubleval
+        cdef string stringval
```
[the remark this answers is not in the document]
karthikeyann: cleaned up the `Literal`.

PointKernel:
Removing libcudf review requests since it's no longer relevant

python/cudf/cudf/core/dataframe.py
```diff
@@ -7057,7 +7057,8 @@ def eval(self, expr: str, inplace: bool = False, **kwargs):
               Specifically, `&` must be used for bitwise operators on integers,
               not `and`, which is specifically for the logical and between
               booleans.
-            * Only numerical types are currently supported.
+            * Only numerical types are currently supported on all operators.
+            * String types are supported only on comparison operators.
```
vyasr, line 7061:
```suggestion
            * Only numerical types currently support all operators.
            * String types currently support comparison operators.
```

python/cudf/cudf/tests/test_dataframe.py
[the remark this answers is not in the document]
vyasr: LGTM, thanks.

vyasr (approved):

python/cudf/cudf/_lib/expressions.pyx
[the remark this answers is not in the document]
vyasr: Much better now, thanks!

python/cudf/cudf/_lib/expressions.pyx
```diff
@@ -77,27 +77,20 @@ class TableReference(Enum):
...
-        cdef int intval
-        cdef double doubleval
-
         if isinstance(value, int):
-            self.c_scalar_type = scalar_type_t.INT
-            intval = value
-            self.c_scalar.int_ptr = make_unique[numeric_scalar[int64_t]](
-                intval, True
-            )
+            self.c_scalar.reset(new numeric_scalar[int64_t](value, True))
```
shwina, line 81: - Shouldn't every `new` have a corresponding `delete`? Note that if we need to call `delete`, it can be done in the `__dealloc__` method of this class
- Can we continue to use a unique/shared pointer so we don't have to worry about defining `__dealloc__` or calling `delete`?

shwina: Oh I think I see what's going here, we're resetting the value of the unique pointer to a new `numeric_scalar`. But why do it this way instead of `make_unique`?

shwina (approved):

python/cudf/cudf/_lib/expressions.pyx
karthikeyann: `make_unique` created the `numeric_scalar` unique pointer. It can't be type casted and stored to generic base pointer. Hence this approach. I saw similar approach in another Cython file as well.
https://github.com/rapidsai/cudf/blob/branch-23.06/python/cudf/cudf/_lib/scalar.pyx#L250

shwina: Thanks! I approved the PR, but still have a question. Why prefer:

```
c_scalar.reset(new ....)
```

versus:

```
c_scalar = make_unique(...)
```

?

karthikeyann: It would require releasing the unique_ptr of derived pointer anyway to store in  `unique_ptr<scalar>` class.
`unique_ptr[scalar] c_scalar = unique_ptr(static_cast<scalar*>(make_unique(...).release()));`

Hence using the new operator directly.

shwina: Got it - thanks for clarifying! Definitely a TIL for me

>  Is 'releaseing the unique_ptr and typecasting' safer than using reset(new... ?

Do you mean is it safer from a Cython perspective? If so, the answer is that it should be no different than in C++. I would say whatever pattern you prefer in C++, we should use here as well.

Issue #13042, opened 2023-03-30
[FEA] Retry support for chunked Parquet writer


ttnghia:
Similar to https://github.com/rapidsai/cudf/issues/12792, we need to support retry for Parquet writer. The changes to Parquet writer should be implemented in a very similar way as was done for ORC writer in https://github.com/rapidsai/cudf/pull/12949.

PR #13076, opened 2023-04-06
Refactor Parquet chunked writer

ttnghia:
Similar to https://github.com/rapidsai/cudf/pull/12949, this refactors Parquet writer to support retry mechanism. The internal `writer::impl::write()` function is rewritten such that it is separated into multiple pieces:
 * A free function that performs compressing/encoding the input table into intermediate results. These intermediate results are totally independent of the writer.
 * After having the intermediate results in the previous step, these results will be actually applied to the output data sink to start the actual data writing.

Closes: 
 * https://github.com/rapidsai/cudf/issues/13042

Depends on:
 * https://github.com/rapidsai/cudf/pull/13206


cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1398,20 +1420,62 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+                                   size_t max_page_size_bytes,
+                                   size_type max_row_group_rows,
+                                   size_type max_page_size_rows,
+                                   int32_t column_index_truncate_length,
+                                   statistics_freq stats_granularity,
+                                   Compression compression,
+                                   dictionary_policy dict_policy,
+                                   size_t max_dictionary_size,
+                                   bool single_write_mode,
+                                   bool int96_timestamps,
```
vuule, line 1469: I wish we could pass the options object directly here :(
Maybe there's a case for an `encode_options` so we group most of these? Just not in this PR, IMO.

cpp/src/io/parquet/writer_impl.hpp
```diff
@@ -118,102 +117,32 @@ class writer::impl {
...
- private:
-  // Cuda stream to be used.
+  void write_parquet_data_to_sink(std::unique_ptr<aggregate_writer_metadata>& updated_agg_meta,
+                                  hostdevice_2dvector<gpu::EncColumnChunk> const& chunks,
+                                  rmm::device_uvector<gpu::EncPage> const& pages,
+                                  std::vector<size_t> const& global_rowgroup_base,
+                                  std::vector<int> const& first_rg_in_part,
+                                  std::vector<size_type> const& batch_list,
+                                  std::vector<int> const& rg_to_part,
+                                  size_type num_columns,
```
vuule, line 142: `chunks.size().second` should be the same as num_columns.
```suggestion
```

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1853,70 +1926,247 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+    // Sync before calling the next `encode_pages` which may alter the stats data.
+    if (need_sync) { stream.synchronize(); }
+  }
+
+  auto write_buff = pinned_buffer<uint8_t>{[](size_t size) {
+                                             uint8_t* ptr = nullptr;
+                                             CUDF_CUDA_TRY(cudaMallocHost(&ptr, size));
+                                             return ptr;
+                                           }(max_chunk_bfr_size),
+                                           cudaFreeHost};
```
vuule, line 1976: Maybe a good time to replace pinned_buffer with a `thrust::host_vector<uint8_t, rmm::mr::pinned_memory_resource>` ?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1853,70 +1926,247 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+void writer::impl::write_parquet_data_to_sink(
+  std::unique_ptr<aggregate_writer_metadata>& updated_agg_meta,
+  hostdevice_2dvector<gpu::EncColumnChunk> const& chunks,
+  rmm::device_uvector<gpu::EncPage> const& pages,
+  std::vector<size_t> const& global_rowgroup_base,
+  std::vector<int> const& first_rg_in_part,
+  std::vector<size_type> const& batch_list,
+  std::vector<int> const& rg_to_part,
+  size_type num_columns,
+  uint8_t* write_buff)
```
vuule, line 2133: how about bounce_buffer?
Also, this should be a host_span.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1970,14 +2218,12 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
           }
 
           _stream.synchronize();
-          _agg_meta->file(p).offset_indexes.push_back(offset_idx);
-          _agg_meta->file(p).column_indexes.push_back(column_idx);
+          _agg_meta->file(p).offset_indexes.emplace_back(std::move(offset_idx));
+          _agg_meta->file(p).column_indexes.emplace_back(std::move(column_idx));
```
vuule, line 2222: :+1:

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1503,7 +1553,7 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
   std::transform(partitions.begin(),
                  partitions.end(),
                  std::back_inserter(num_frag_in_part),
-                 [this, max_page_fragment_size](auto const& part) {
+                 [&, max_page_fragment_size](auto const& part) {
```
vuule, line 1556: why is the capture-default (&) needed here?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1853,70 +1926,247 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+                                           _max_row_group_size,
+                                           _max_page_size_bytes,
+                                           _max_row_group_rows,
+                                           _max_page_size_rows,
+                                           _column_index_truncate_length,
+                                           _stats_granularity,
+                                           _compression,
+                                           _dict_policy,
+                                           _max_dictionary_size,
+                                           _single_write_mode,
```
vuule, line 2099: if you don't, please change `_single_write_mode` to type `SingleWriteMode`. `bool` is just a step back here, since we already have a nice enum.

vuule (commented):
Incomplete review, these kind of changes are hard to parse.
Will do another round once we iterate on the suggestions so far.

cpp/src/io/parquet/writer_impl.hpp
```diff
@@ -118,102 +117,32 @@ class writer::impl {
...
-
- private:
-  // Cuda stream to be used.
+  void write_parquet_data_to_sink(std::unique_ptr<aggregate_writer_metadata>& updated_agg_meta,
+                                  hostdevice_2dvector<gpu::EncColumnChunk> const& chunks,
+                                  rmm::device_uvector<gpu::EncPage> const& pages,
+                                  std::vector<size_t> const& global_rowgroup_base,
+                                  std::vector<int> const& first_rg_in_part,
+                                  std::vector<size_type> const& batch_list,
+                                  std::vector<int> const& rg_to_part,
```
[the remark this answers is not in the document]
ttnghia: Sounds good.

cpp/src/io/parquet/writer_impl.cu
ttnghia: That is to replace `this` which is not possible for a free function. Probably it is even not needed.

cpp/src/io/parquet/writer_impl.cu
ttnghia: Good point. But I also want to change that for ORC writer too. Will do it in a separate PR.

nvdbaranec:
What is the high level view here?   So internally we now generate the data (`convert_table_to_parquet_data`) and then if no exception has been thrown, we write it with `write_parquet_data_to_sink`.  What is the case this allows the external user to handle, and how do they 'retry'?

ttnghia:
> I understand that out-of-memory is probably what we're interested in, but I'm not sure what the intended usage is here for the external caller.

The logic behind current retry at the caller (Spark) is:
 1. Call `write(table)` with `try/catch`.
 2. If it is success, the everything is done.
 3. If it failed with some exception thrown out, the input may be too big that caused OOM. So it will be split into two smaller chunks and each one will be written one after another, hoping that writing such smaller chunk will not cause OOM.
 4. Repeat step 2. and 3. if needed.

nvdbaranec:
Out of curiosity, why does this have to happen inside cudf?  Can't the Spark code just create a new writer and use that?

ttnghia:
> Out of curiosity, why does this have to happen inside cudf? Can't the Spark code just create a new writer and use that?
 
Of course we can copy exactly the Parquet writer code (a lot) into spark-rapids-jni, then do this refactor there.

However, this PR doesn't implement anything new. It is just a refactor. But it can support Spark need, so this have to happen in cudf to avoid code duplicate (a lot if doing so in spark-rapids-jni).

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1511,24 +1566,48 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
 
-  std::vector<size_t> const global_rowgroup_base = _agg_meta->num_row_groups_per_file();
+  std::unique_ptr<aggregate_writer_metadata> agg_meta;
+  if (!curr_agg_meta) {
+    agg_meta = std::make_unique<aggregate_writer_metadata>(
+      partitions, kv_meta, this_table_schema, num_columns, stats_granularity);
+  } else {
+    agg_meta = std::make_unique<aggregate_writer_metadata>(*curr_agg_meta);
+
+    //    verify the user isn't passing mismatched tables
```
abellina, line 1600: nit
```suggestion
    // verify the user isn't passing mismatched tables
```

vuule:
> Out of curiosity, why does this have to happen inside cudf? Can't the Spark code just create a new writer and use that?

The new writer would not be able to append to the existing file (only relevant for chunked writer). The existing writer can at least retry to close to write the footer.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1511,24 +1566,48 @@ void writer::impl::write(table_view const& table, std::vector<partition_info> co
...
+    CUDF_EXPECTS(agg_meta->schema_matches(this_table_schema),
+                 "Mismatch in schema between multiple calls to write_chunk");
+
+    agg_meta->update_files(partitions);
```
abellina, line 1604: I am wondering if agg_meta gets updated, and then later we OOM, should agg_meta be rolled back or thrown away? Or is the code after this guaranteed to not OOM?

And apologies if I misunderstand the way this is working. I am probably missing something

ttnghia: This `agg_meta` is not the writer's `agg_meta`. Instead, it is a copy from writer's `agg_meta`. The writer's `agg_meta` will be updated (assigned to this) after the processing step is done.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -921,37 +926,69 @@ gpu::parquet_column_device_view parquet_column_view::get_device_view(
...
-void writer::impl::init_row_group_fragments(
-  cudf::detail::hostdevice_2dvector<gpu::PageFragment>& frag,
-  device_span<gpu::parquet_column_device_view const> col_desc,
-  host_span<partition_info const> partitions,
-  device_span<int const> part_frag_offset,
-  uint32_t fragment_size)
+/**
+ * @brief Gather row group fragments
+ *
+ * This calculates fragments to be used in determining row group boundariesa.
```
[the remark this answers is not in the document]
ttnghia: I wonder why CI spell check doesn't catch it this time.

cpp/src/io/parquet/writer_impl.cu
abellina: Got it, thank you!

vuule (approved):
Looks great!

nvdbaranec (approved):

Issue #13037, opened 2023-03-30
[BUG] GroupBy.apply() that involve reductions return index with wrong values


shwina:
When the UDF passed to `groupby.apply()` involves a reduction, we return an incorrect index:

```python
In [16]: df = cudf.DataFrame({'a': [1, 1, 2, 2], 'b': [1, 2, 3, 4]})

In [17]: df.groupby('a').apply(lambda group: group.sum())
Out[17]: 
   a  b
a      
0  2  3
1  4  7
```

Pandas behavior:

```python
In [18]: df.to_pandas().groupby('a').apply(lambda group: group.sum())
Out[18]: 
   a  b
a      
1  2  3
2  4  7
```

shwina:
As a note, just changing the data here:

https://github.com/rapidsai/cudf/blob/173fde9d9ae335014a1aa6de60b417f6f245dcf2/python/cudf/cudf/tests/test_groupby.py#L1958

to:

```python
{"key": [3, 3, 1, 1, 2, 2, 2], "val": [0, 1, 2, 3, 4, 5, 6]} 
```

will fail the unit test.

shwina:
@brandon-b-miller do you think this is a safe assumption (at least given the current state and features we support in groupby-apply):

> if the size of the result is the same as the number of unique keys, we can return the unique keys in sorted order as the index

brandon-b-miller:
I'd say that if the result of applying the function to a single group is a scalar, then we can take the unique keys to be the index, but I don't think we make any guarantees that the ordering of those keys matches pandas - just that they're the same set of values overall. 

I can look into addressing this as part of my fix for https://github.com/rapidsai/cudf/issues/12987 as there seem to be several issues with the same section of code that all have to do with inspecting the output of our iterative groupby process to determine what kind of output object to assemble and what to do with it's index.

shwina:
Okay -- thank you! I reassigned to you if that's ok

Issue #12987, opened 2023-03-21
[BUG] Groupby.apply UDFs that return a column per group result in a dataframe with one column per row


beckernick:
In the 23.04 nightly and 23.02 release, groupby apply UDFs that return a column per group return a sparse dataframe with one column per row in the original dataframe. We also lose the index.

This behavior does not happen for UDFs executing reductions.

```python
import cudf
import pandas as pd

df = cudf.datasets.randomdata()
pdf = df.to_pandas()

def func(df):
    return df.x + df.y

print(pdf.groupby("id").apply(func).head())
print(df.groupby("id").apply(func).head())
id     
956   7   -1.571980
964   9    1.112649
988   5    0.151777
989   8    1.131481
1011  3    1.748853
dtype: float64
       0     1     2            3     4          5     6             7  \
id                                                                       
0   <NA>  <NA>  <NA>         <NA>  <NA>       <NA>  <NA>  -1.571979514   
1   <NA>  <NA>  <NA>         <NA>  <NA>       <NA>  <NA>          <NA>   
2   <NA>  <NA>  <NA>         <NA>  <NA>  0.1517768  <NA>          <NA>   
3   <NA>  <NA>  <NA>         <NA>  <NA>       <NA>  <NA>          <NA>   
4   <NA>  <NA>  <NA>  1.748852657  <NA>       <NA>  <NA>          <NA>   

              8            9  
id                            
0          <NA>         <NA>  
1          <NA>  1.112648516  
2          <NA>         <NA>  
3   1.131481029         <NA>  
4          <NA>         <NA>  
```

```
conda list | grep rapids
# packages in environment at /home/nicholasb/miniconda3/envs/rapids-23.04:
cudf_kafka                23.04.00a       py310_230321_g84ac1d5a87_208    rapidsai-nightly
cusignal                  23.04.00a230203 py38_ge62a377_4    rapidsai-nightly
libcucim                  23.04.00a       cuda11_230308_gf3ef616_28    rapidsai-nightly
libcudf                   23.04.00a       cuda11_230321_g84ac1d5a87_208    rapidsai-nightly
libcudf_kafka             23.04.00a       230321_g84ac1d5a87_208    rapidsai-nightly
libcugraph                23.04.00a       cuda11_230320_g6e11df43_91    rapidsai-nightly
libcugraph_etl            23.04.00a       cuda11_230320_g6e11df43_91    rapidsai-nightly
[...]

PR #13078, opened 2023-04-06
Fix bugs in iterative groupby apply algorithm

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/13037
Closes https://github.com/rapidsai/cudf/issues/12987


python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -44,6 +44,14 @@ def _quantile_75(x):
     return x.quantile(0.75)
 
 
+def _is_row_of(chunk, obj):
+    if isinstance(chunk, cudf.Series) and isinstance(obj, cudf.DataFrame):
+        if len(chunk.index) == len(obj._column_names):
+            if (chunk.index.to_pandas() == pd.Index(obj._column_names)).all():
+                return True
```
bdice, line 51: Let's just return the `and` of these three conditions. `return A and B and C`.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -1188,8 +1202,20 @@ def _iterative_groupby_apply(
...
+            # When the UDF is like df.sum(), the result for each
+            # group is a row-like "Series" where the index labels
+            # are the same as the original calling DataFrame
+            if _is_row_of(chunk_results[0], self.obj):
+                result = cudf.concat(chunk_results, axis=1).T
+                result.index = group_names
+                result.index.names = self.grouping.names
+            # When the UDF is like df.x + df.y, the result for each
+            # group is the same length as the original group
+            elif len(self.obj) == sum(len(chk) for chk in chunk_results):
```
bdice, line 1214: There's no `else` block here. Can we write this in a way that prevents `result` from being an unbound variable? Even if the `else` block is raising a descriptive error, it's better than letting `result` be an unbound variable that is needed later in the function.

bdice (changes requested):

PR #13081, opened 2023-04-06
Change cudf::test::make_null_mask to also return null-count

davidwendt:
## Description
Change the `cudf::test::make_null_mask` to return both the null-mask and the null-count. Callers can then use this null-count instead of `UNKNOWN_NULL_COUNT`. These changes include removing `UNKNOWN_NULL_COUNT` usage from the libcudf C++ test source code.

One side-effect found that strings column with all nulls can technically have no children but using `UNKNOWN_NULL_COUNT` allowed the check for this to be bypassed. Therefore many utilities started to fail when `UNKNOWN_NULL_COUNT` was removed. The factory was modified to remove the check which results in an offsets column and an empty chars column as children.

More code will likely need to be change when the `UNKNOWN_NULL_COUNT` is no longer used as a default parameter for factories and other column functions.

No behavior is changed. Since the `cudf::test::make_null_mask` is technically a public API, this PR could be marked as a breaking change as well.

## Checklist


cpp/tests/copying/scatter_list_tests.cpp
```diff
@@ -951,13 +958,13 @@ TYPED_TEST(TypedScatterListsTest, NullListsOfStructs)
...
     cudf::detail::make_counting_transform_iterator(0, [](auto i) { return i != 4; });
 
-  auto expected_lists =
-    cudf::make_lists_column(6,
-                            offsets_column{0, 3, 5, 9, 11, 11, 13}.release(),
-                            expected_structs.release(),
-                            1,
-                            cudf::test::detail::make_null_mask(expected_lists_null_mask_begin,
-                                                               expected_lists_null_mask_begin + 6));
+  std::tie(null_mask, null_count) = cudf::test::detail::make_null_mask(
```
mythrocks, line 961: Ah, we `std::tie` these because they're already defined previously?

cpp/tests/interop/from_arrow_test.cpp
```diff
@@ -188,8 +188,8 @@ TEST_F(FromArrowTest, StructColumn)
   cols2.push_back(std::move(int_col2));
   auto mask =
     cudf::bools_to_mask(cudf::test::fixed_width_column_wrapper<bool>{{true, true, false}});
-  auto sub_struct_col = cudf::make_structs_column(
-    num_rows, std::move(cols2), cudf::UNKNOWN_NULL_COUNT, std::move(*(mask.first)));
+  auto sub_struct_col =
+    cudf::make_structs_column(num_rows, std::move(cols2), mask.second, std::move(*(mask.first)));
```
mythrocks, line 192: Should we consider using structured bindings here, instead of `mask.first`/`mask.second`, for readability?

mythrocks (approved):
Sorry for the delayed review. LGTM! A couple of minor nitpicks.

cpp/include/cudf_test/column_utilities.hpp
```diff
@@ -254,26 +254,26 @@ std::pair<thrust::host_vector<T>, std::vector<bitmask_type>> to_host(column_view
...
+
+    // build std::string vector from chars and offsets
+    std::transform(
+      std::begin(h_offsets),
+      std::end(h_offsets) - 1,
+      std::begin(h_offsets) + 1,
+      host_data.begin(),
+      [&](auto start, auto end) { return std::string(h_chars.data() + start, end - start); });
+  }
+  return std::make_pair(std::move(host_data), bitmask_to_host(c));
```
mythrocks, line 276: As an aside, I wonder if it would be possible express this as:
```c++
return std::pair(std::move(host_data), bitmask_to_host(c));
```
Or even thus:
```c++
return {std::move(host_data), bitmask_to_host(c)};
```

davidwendt: Are these better somehow?

cpp/include/cudf_test/column_wrapper.hpp
```diff
@@ -266,12 +273,14 @@ std::vector<bitmask_type> make_null_mask_vector(ValidityIterator begin, Validity
...
 template <typename ValidityIterator>
-rmm::device_buffer make_null_mask(ValidityIterator begin, ValidityIterator end)
+std::pair<rmm::device_buffer, cudf::size_type> make_null_mask(ValidityIterator begin,
+                                                              ValidityIterator end)
 {
-  auto null_mask = make_null_mask_vector(begin, end);
-  return rmm::device_buffer{null_mask.data(),
-                            null_mask.size() * sizeof(decltype(null_mask.front())),
-                            cudf::get_default_stream()};
+  auto [null_mask, null_count] = make_null_mask_vector(begin, end);
```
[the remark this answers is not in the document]
davidwendt: Perhaps. Is that important to convey here?

cpp/tests/copying/scatter_list_tests.cpp
davidwendt: Yes.

cpp/include/cudf_test/column_wrapper.hpp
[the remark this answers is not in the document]
mythrocks: No, the const-ness isn't particularly important. Just commonly the default.

cpp/include/cudf_test/column_utilities.hpp
mythrocks: It's more concise. I assumed it was preferred because it avoids repeating the type in the return. (I've had this pointed out during reviews of my code.)

Please feel free to ignore. This might just be a stylistic preference.

cpp/include/cudf_test/column_utilities.hpp
```diff
@@ -254,26 +254,26 @@ std::pair<thrust::host_vector<T>, std::vector<bitmask_type>> to_host(column_view
...
+
+    // build std::string vector from chars and offsets
+    std::transform(
+      std::begin(h_offsets),
+      std::end(h_offsets) - 1,
+      std::begin(h_offsets) + 1,
+      host_data.begin(),
+      [&](auto start, auto end) { return std::string(h_chars.data() + start, end - start); });
+  }
+  return {std::move(host_data), bitmask_to_host(c)};
```
vyasr, line 276: The git diff here is a bit hard to read, but IIUC this whole change is just to preallocate the vector and then early return if you don't have any non-null entries to write, is that correct?

cpp/include/cudf_test/column_wrapper.hpp
```diff
@@ -266,12 +273,14 @@ std::vector<bitmask_type> make_null_mask_vector(ValidityIterator begin, Validity
...
+  auto d_mask                  = rmm::device_buffer{null_mask.data(),
+                                   null_mask.size() * sizeof(decltype(null_mask.front())),
```
vyasr, line 281: No strong opinion here, but aren't we basically guaranteed that `decltype(null_mask.front())` is `bitmask_type`? Can we ever have anything else?

cpp/tests/bitmask/bitmask_tests.cpp
```diff
@@ -575,12 +574,13 @@ TEST_F(CopyBitmaskTest, TestZeroOffset)
...
   }
-  auto input_mask = cudf::test::detail::make_null_mask(validity_bit.begin(), validity_bit.end());
+  auto input_mask =
+    std::get<0>(cudf::test::detail::make_null_mask(validity_bit.begin(), validity_bit.end()));
 
   int begin_bit         = 0;
   int end_bit           = 800;
-  auto gold_splice_mask = cudf::test::detail::make_null_mask(validity_bit.begin() + begin_bit,
-                                                             validity_bit.begin() + end_bit);
+  auto gold_splice_mask = std::get<0>(cudf::test::detail::make_null_mask(
```
vyasr, line 582: Our of curiosity, is there a reason you prefer `std::get<0>` to `.first` for pairs? For consistency with tuples?

vyasr (approved):
For a detail API this is used a ton in tests... In the end I suspect we're going to need to expose a number of our currently detail-only mask APIs publicly. Anyway this LGTM for now.

cpp/include/cudf_test/column_utilities.hpp
davidwendt: Yes, that is correct.

cpp/include/cudf_test/column_wrapper.hpp
davidwendt: That is strange. I'd not noticed it before. I think we should be using https://github.com/rapidsai/cudf/blob/a562a7e0ffbe3fcfa78488733ecb00beebeaa48b/cpp/include/cudf/null_mask.hpp#L57 here instead.

cpp/tests/bitmask/bitmask_tests.cpp
davidwendt: I like that this gets us the first value from the structured binding without an unused variable and without resorting to the vagueness of `.first` and `.second`.

PR #13090, opened 2023-04-07
[REVIEW] Enable `numeric_only` for row-wise ops

galipremsagar:
## Description
This PR enables `numeric_only` for row-wise ops to be on parity with pandas-2.0.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -5738,7 +5738,7 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
     # Stats
     #
     @_cudf_nvtx_annotate
-    def _prepare_for_rowwise_op(self, method, skipna):
+    def _prepare_for_rowwise_op(self, method, skipna, numeric_only):
         """Prepare a DataFrame for CuPy-based row-wise operations."""
 
         if method not in _cupy_nan_methods_map and any(
@@ -5752,26 +5752,23 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
             )
             raise ValueError(msg)
 
-        is_pure_dt = all(is_datetime_dtype(dt) for dt in self.dtypes)
-
-        if not is_pure_dt:
+        if numeric_only:
             filtered = self.select_dtypes(include=[np.number, np.bool_])
         else:
             filtered = self.copy(deep=False)
 
-        common_dtype = find_common_type(filtered.dtypes)
+        is_pure_dt = all(is_datetime_dtype(dt) for dt in self.dtypes)
 
-        if filtered._num_columns < self._num_columns:
-            # When we update our pandas compatibility target to 2.0, pandas
-            # will stop supporting numeric_only=None and users will have to
-            # specify True/False. At that time we should also top our implicit
-            # removal of non-numeric columns here.
-            assert Version(pd.__version__) < Version("2.0.0")
-            msg = (
-                "Row-wise operations currently only support int, float "
-                "and bool dtypes. Non numeric columns are ignored."
+        common_dtype = find_common_type(filtered.dtypes)
+        if (
+            not numeric_only
+            and is_string_dtype(common_dtype)
+            and any(not is_string_dtype(dt) for dt in filtered.dtypes)
+        ):
+            raise TypeError(
+                f"Cannot perform row-wise {method} across mixed-dtype columns,"
+                " try type-casting all the columns to same dtype."
             )
-            warnings.warn(msg)
 
         if not skipna and any(col.nullable for col in filtered._columns):
             mask = DataFrame(
@@ -5857,7 +5854,7 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
             )
             source = self._get_columns_by_label(numeric_cols)
             if source.empty:
-                return Series(index=cudf.StringIndex([]))
+                return Series(index=self.index)
 
         axis = source._get_axis_from_axis_arg(axis)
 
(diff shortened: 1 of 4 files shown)


galipremsagar:
This PR resolves around `75` pytest failures & adds `201` pytests  related to row-wise ops:

```bash
= 851 failed, 86378 passed, 2034 skipped, 992 xfailed, 165 xpassed in 532.31s (0:08:52) =
```

```bash
= 926 failed, 86177 passed, 2034 skipped, 992 xfailed, 165 xpassed in 545.17s (0:09:05) =
```

python/cudf/cudf/tests/test_array_ufunc.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
```
vyasr, line 1: Why did this change?

python/cudf/cudf/core/dataframe.py
```diff
@@ -5752,26 +5752,23 @@ def _prepare_for_rowwise_op(self, method, skipna):
...
-            # removal of non-numeric columns here.
-            assert Version(pd.__version__) < Version("2.0.0")
-            msg = (
-                "Row-wise operations currently only support int, float "
-                "and bool dtypes. Non numeric columns are ignored."
+        common_dtype = find_common_type(filtered.dtypes)
+        if (
+            not numeric_only
+            and is_string_dtype(common_dtype)
+            and any(not is_string_dtype(dt) for dt in filtered.dtypes)
```
vyasr, line 5766: Why is `is_string_dtype` the condition here? I get that we can't just verify that every column's dtype is the same as `common_dtype` since we need to allow for casting between different numeric types, but does this condition account for other column dtypes or are those handled separately?

galipremsagar: The actual intention was to capture `cudf.dtype(object)` dtype, `is_string_dtype` helps capture that. The reason being whenever there seems to be a mix of un-related types, the result is always `object` dtype or an error(incase of mix of struct & list types).
```python
In [3]: find_common_type(["float", "int"])
Out[3]: dtype('float64')

In [4]: find_common_type(["float", "int", "datetime64[ns]"])
Out[4]: dtype('O')
```

python/cudf/cudf/tests/test_array_ufunc.py
galipremsagar: This file was edited in https://github.com/rapidsai/cudf/pull/13083/ but looks like the style checker missed adding it locally and has done now.

python/cudf/cudf/core/dataframe.py
```diff
@@ -5752,26 +5752,23 @@ def _prepare_for_rowwise_op(self, method, skipna):
...
-        is_pure_dt = all(is_datetime_dtype(dt) for dt in self.dtypes)
-
-        if not is_pure_dt:
+        if numeric_only:
             filtered = self.select_dtypes(include=[np.number, np.bool_])
         else:
             filtered = self.copy(deep=False)
 
-        common_dtype = find_common_type(filtered.dtypes)
+        is_pure_dt = all(is_datetime_dtype(dt) for dt in self.dtypes)
```
galipremsagar, line 5760:
```suggestion
        is_pure_dt = all(is_datetime_dtype(dt) for dt in filtered.dtypes)
```

python/cudf/cudf/core/dataframe.py
vyasr: Ah I see, so basically if we can't compute a common dtype that isn't object we throw an error unless object is actually the result of a bunch of string columns in which case we're OK.

python/cudf/cudf/core/dataframe.py
```diff
@@ -5857,7 +5854,7 @@ def _reduce(
             )
             source = self._get_columns_by_label(numeric_cols)
             if source.empty:
-                return Series(index=cudf.StringIndex([]))
+                return Series(index=self.index)
```
vyasr, line 5857: Is this a pandas 2.0 change? You get an empty output but now with a nonempty index?

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -4910,38 +4912,41 @@ def test_df_constructor_dtype(dtype):
...
-    ):
         got = getattr(gdf, op)(**kwargs)
 
-    assert_eq(expected, got, check_exact=False)
+        if len(got) == 0:
+            assert_eq(
+                expected.reset_index(drop=True),
+                got.reset_index(drop=True),
+                check_exact=False,
+            )
```
vyasr, line 4947: So this is checking that new behavior I commented on above where you're setting the index even on an empty Series? Why does that need to be special-cased, doesn't our assertion capture that without needing to reset the index?

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -5044,10 +5000,12 @@ def test_rowwise_ops_nullable_dtypes_partial_null(op, expected):
 
     if op in ("var", "std"):
         got = getattr(gdf, op)(axis=1, ddof=0, skipna=False)
+        expected = getattr(gdf.to_pandas(), op)(axis=1, ddof=0, skipna=False)
     else:
         got = getattr(gdf, op)(axis=1, skipna=False)
+        expected = getattr(gdf.to_pandas(), op)(axis=1, skipna=False)
 
-    assert_eq(got.null_count, expected.null_count)
+    assert_eq(got.null_count, 2)
```
vyasr, line 5008: Can't we still use `expected.null_count()`? Or do we have to use a different API in 2.0 to account for NaNs vs pd.NA?

vyasr (approved):
Some questions, but assuming that my understanding of the code is correct everything looks right and only have a couple non-blocking small changes that could be made.

python/cudf/cudf/core/dataframe.py
galipremsagar: Yup, pandas 2.0 fixed an issue where the index was being dropped for empty outputs.

python/cudf/cudf/tests/test_dataframe.py
galipremsagar: `expected` was a `cudf.Series` previously, hence we could use `null_count`. But pandas doesn't expose any `null_count` equivalent public API/property.

python/cudf/cudf/tests/test_dataframe.py
galipremsagar: Thanks for pointing it out. Upon checking we actually don't need this change after the index change you pointed out above.

Issue #12973, opened 2023-03-17
[FEA] Consistently use prefix/suffix for member variables of `cudf::io::detail::orc::writer::impl`


ttnghia:
Currently, the member variables of `cudf::io::detail::orc::writer::impl` are very inconsistent:
 * Some have prefix `_`.
 * Some have suffix `_`.
 * Some don't have `_` as prefix or suffix.

We should make them consistent. The convention should be only using `_` as prefix.

ttnghia:
CC @vuule.

vuule:
Sounds good.

PR #13091, opened 2023-04-07
Cleanup ORC chunked writer

ttnghia:
This changes the internal variables of ORC chunked writer:
 * Renaming them to have a `_` prefix consistently.
 * Add `const` qualifier to some variables that are writer parameters.
 * Regroup them.

Closes: 
 * https://github.com/rapidsai/cudf/issues/12973


vyasr (approved):
The improved consistency is good, but isn't the leading underscore a bit redundant in C++ where the private access specifier already conveys that these are internal variables?

PR #13095, opened 2023-04-07
Use `.element()` instead of `.data()` for window range calculations

mythrocks:
## Summary
This change is in preparation for supporting `STRING` order-by columns in range window functions. It updates how order-by column values are accessed for range-based window functions. No user-facing APIs are altered.

## Description
In the staging step for executing window range queries, the boundaries of each row's window are calculated. This involves subtracting/adding the `preceding`/`following` values from each order-by column row, and then searching backwards/forwards for the boundary values.

The staging step has been using `column_device_view.data()` for accessing the order-by rows, an acceptable approach for when the order-by columns are numeric (e.g. `INT32`).

This approach fails when the order-by column is a `STRING`, because `.data()` is not defined for such columns. A better approach would be to use `.element()` to directly access the rows, because it has special handling for `STRING`, among other types, while continuing to work for numeric primitives.

## Future
In a followup to this change, support for `STRING` order-by columns will be added.

## Checklist


cpp/src/rolling/grouped_rolling.cu
```diff
@@ -269,6 +270,69 @@ __device__ T subtract_safe(T const& value, T const& delta)
...
+template <typename T>
+struct device_value_accessor {
+  column_device_view const col;  ///< column view of column in device
+
+  /**
+   * @brief constructor
+   *
+   * @param[in] _col column device view of cudf column
+   */
+  __device__ device_value_accessor(column_device_view const& _col) : col{_col}
```
ttnghia, line 309: Typical convention for using `_`:
 * Use `_var` style for private class/struct members.
 * Use `var_` or `var` style for function parameters.
 * Don't use `_` for public class/struct members.

```suggestion
  __device__ device_value_accessor(column_device_view const& col_) : col{col_}
```

ttnghia: if `col` is made private then:
```suggestion
...
private:
column_device_view const _col;
....
  __device__ device_value_accessor(column_device_view const& col) : _col{col}
```

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -752,14 +828,14 @@ std::unique_ptr<column> range_window_DESC(column_view const& input,
...
     // orderby[idx] not null. Search must exclude the null group.
     // If nulls_begin == group_start, either of the following is true:
     //  1. NULLS FIRST ordering: Search ends at group_end.
     //  2. NO NULLS: Search ends at group_end.
     // Otherwise, NULLS LAST ordering. Search ends at nulls_begin.
-    auto search_end = nulls_begin == group_start ? group_end : nulls_begin;
-
-    auto lowest_in_window = subtract_safe(d_orderby[idx], following_window);
+    auto search_end       = nulls_begin == group_start ? group_end : nulls_begin;
+    auto lowest_in_window = compute_lowest_in_window(d_orderby, idx, following_window);
```
ttnghia, line 838: `const`?

cpp/src/rolling/grouped_rolling.cu
mythrocks: Thanks. I've switched to using a `col_` for method parameter.

cpp/src/rolling/grouped_rolling.cu
mythrocks: Agreed. I've added const all over, now.

mythrocks:
Rerun tests

vyasr:
@mythrocks the world has changed since our GHA migration :) See https://docs.rapids.ai/resources/github-actions/#rerunning-failed-github-actions for how to rerun tests

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -269,6 +270,69 @@ __device__ T subtract_safe(T const& value, T const& delta)
...
+   */
+  __device__ T operator()(cudf::size_type i) const { return col.element<T>(i); }
+};
+
+template <typename T>
+using const_device_iterator =
+  thrust::transform_iterator<device_value_accessor<T>, thrust::counting_iterator<size_type>>;
+
+/// This is a stand-in for the `cudf::column_device_view::begin<T>()`, which is `__host__` only.
+/// For range window functions, one might need to iterate over the `oby` column, per row.
```
ttnghia, line 328: Is `oby` here a typo?

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -269,6 +270,69 @@ __device__ T subtract_safe(T const& value, T const& delta)
...
+template <typename T, CUDF_ENABLE_IF(cudf::column_device_view::has_element_accessor<T>())>
+[[nodiscard]] __device__ const_device_iterator<T> begin(cudf::column_device_view const& col)
```
ttnghia, line 330: Sorry I don't see this being used anywhere?

cpp/src/rolling/grouped_rolling.cu
mythrocks: That was shorthand for the order-by column. I have spelt it out, so that it's more readable.

cpp/src/rolling/grouped_rolling.cu
mythrocks: [Here](https://github.com/rapidsai/cudf/pull/13095/files#diff-43139648ae3ec9db09ede1dde96e6e5c88d67a32320a07c7a1dfa267ff918291R410) is one example. It is used in the preceding/following calculation instead of using `d_orderby = orderby_column.data<T>()`, so as to ultimately use `.element<T>()`.

PointKernel (approved):
LGTM

ttnghia (approved):

mythrocks:
Thank you for the reviews, @ttnghia, @PointKernel. I have merged this change.

Issue #13053, opened 2023-04-04
[FEA] Reference counting for memory owning objects in numba kernels


brandon-b-miller:
With the introduction of string datatypes inside user defined functions as well as the merging of the groupby JIT engine, there are now certain objects allowed within UDFs that are individually backed by a device allocation. This issue tracks the introduction of a CUDA version of the reference counting mechanism that powers numba's nopython mode for the CPU target. This mechanism will provide much more robust cleanup for the objects in question than the current approach which attempts to clean everything up in post-processing.

The below tasklist is a draft for now and may change as things are merged.

```[tasklist]
### Tasks
- [ ] https://github.com/rapidsai/cudf/pull/13060
- [ ] https://github.com/rapidsai/cudf/pull/13098
- [ ] https://github.com/rapidsai/cudf/pull/13125
- [ ] https://github.com/rapidsai/cudf/pull/13172
- [ ] https://github.com/rapidsai/cudf/pull/13181
```

vyasr:
@brandon-b-miller I'm closing this issue as resolved by https://github.com/rapidsai/cudf/pull/18453, but please reopen if there's further work needed.

PR #13098, opened 2023-04-10
Introduce `managed_udf_string` and matching API variants

brandon-b-miller:
Part of https://github.com/rapidsai/cudf/issues/13053
Depends on https://github.com/rapidsai/cudf/pull/13060

cc @davidwendt

### DIFF
--- a/python/cudf/udf_cpp/strings/include/cudf/strings/udf/udf_apis.hpp
+++ b/python/cudf/udf_cpp/strings/include/cudf/strings/udf/udf_apis.hpp
@@ -28,6 +28,7 @@ namespace strings {
 namespace udf {
 
 class udf_string;
+class managed_udf_string;
 
 /**
  * @brief Return a cudf::string_view array for the given strings column
@@ -58,6 +59,21 @@ std::unique_ptr<cudf::column> column_from_udf_string_array(udf_string* d_strings
                                                            cudf::size_type size);
 
 /**
+ * @brief Return a STRINGS column given an array of managed_udf_string objects
+ *
+ * This will make a copy of the strings in managed_strings in order to build
+ * the output column.
+ * The individual managed_udf_strings are also cleared freeing each of their internal
+ * device memory buffers.
+ *
+ * @param managed_strings Pointer to device memory of managed_udf_string objects
+ * @param size The number of elements in the managed_strings array
+ * @return A strings column copy of the managed_udf_string objects
+ */
+std::unique_ptr<cudf::column> column_from_managed_udf_string_array(
+  managed_udf_string* managed_strings, cudf::size_type size);
+
+/**
  * @brief Frees a vector of udf_string objects
  *
  * The individual udf_strings are cleared freeing each of their internal
@@ -67,6 +83,18 @@ std::unique_ptr<cudf::column> column_from_udf_string_array(udf_string* d_strings
  * @param size The number of elements in the d_strings array
  */
 void free_udf_string_array(udf_string* d_strings, cudf::size_type size);
+
+/**
+ * @brief Frees a vector of managed_udf_string objects
+ *
+ * The individual managed_udf_strings are cleared freeing each of their internal
+ * device memory buffers. In addition, the meminfo object allocated by numba using
+ * malloc is also freed.
+ *
+ * @param managed_strings Pointer to device memory of udf_string objects
+ * @param size The number of elements in the d_strings array
+ */
+void free_managed_udf_string_array(managed_udf_string* managed_strings, cudf::size_type size);
 
 }  // namespace udf
 }  // namespace strings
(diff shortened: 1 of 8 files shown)


python/cudf/udf_cpp/strings/include/cudf/strings/udf/udf_apis.hpp
```diff
@@ -57,6 +58,21 @@ std::unique_ptr<rmm::device_buffer> to_string_view_array(cudf::column_view const
 std::unique_ptr<cudf::column> column_from_udf_string_array(udf_string* d_strings,
                                                            cudf::size_type size);
 
+/**
+ * @brief Return a STRINGS column given an array of managed_udf_string objects
+ *
+ * This will make a copy of the strings in managed_strings in order to build
+ * the output column.
+ * The individual managed_udf_strings are also cleared freeing each of their internal
```
gmarkall, line 66: I think the docstring here says that the `managed_udf_string`s are cleared, but it doesn't look like the implementation does that. Am I misunderstanding the code or is the docstring inaccurate?

brandon-b-miller: You're right. The freeing functions clear these not the transform functions. I updated this.

python/cudf/udf_cpp/strings/src/strings/udf/udf_apis.cu
```diff
@@ -98,6 +129,24 @@ void free_udf_string_array(cudf::strings::udf::udf_string* d_strings,
...
+ */
+void free_managed_udf_string_array(cudf::strings::udf::managed_udf_string* managed_strings,
+                                   cudf::size_type size,
+                                   rmm::cuda_stream_view stream)
+{
+  thrust::for_each_n(rmm::exec_policy(stream),
+                     thrust::make_counting_iterator(0),
+                     size,
+                     [managed_strings] __device__(auto idx) {
+                       managed_strings[idx].udf_str.clear();
```
gmarkall, line 145: Should this be replaced with a call to the destructor, similarly to https://github.com/rapidsai/cudf/pull/13060/files#r1158371391 ?

python/cudf/udf_cpp/strings/src/strings/udf/udf_apis.cu
```diff
@@ -98,6 +129,24 @@ void free_udf_string_array(cudf::strings::udf::udf_string* d_strings,
...
+                       free(managed_strings[idx].meminfo);
```
gmarkall, line 146: Would this have been allocated with `NRT_MemInfo_new`? If so, should it be freed with a call to `NRT_MemInfo_destroy` - this would keep symmetry in the API usage, and ensure the alloc / free stats balance.

python/cudf/udf_cpp/strings/src/strings/udf/udf_apis.cu
brandon-b-miller: This follows the pattern from [the other API](https://github.com/rapidsai/cudf/blob/branch-23.06/python/cudf/udf_cpp/strings/src/strings/udf/udf_apis.cu#L98). But this is a good point nonetheless. 

@davidwendt should we be calling the full destructor in _both_ these cases?

python/cudf/udf_cpp/strings/src/strings/udf/udf_apis.cu
brandon-b-miller: this is a great suggestion and I'll make this change.

python/cudf/udf_cpp/strings/src/strings/udf/udf_apis.cu
davidwendt: From my understanding, the lifetime of the managed-udf-string is managed by the meminfo object so I don't think  you need this call at all, right? The meminfo's manage-udf-string will be cleaned up Numba?

brandon-b-miller: that is true except in the case where we want to return the string outside of the function, meaning the result of the full UDF computation is a new string. In that case we make sure the strings are _not_ cleaned up and still exist to be transformed into the final column.

Technically we do this by adding an `incref` to every string that gets placed into the output array, so at the end of the kernel the `refct` associated with those strings is always `1` still.

davidwendt: You just don't want to be calling the destructor twice. If you were to call the destructor here and then meminfo calls it again later when `refct==0` for the same object that would be bad. Are you also destroying the meminfo objects after they get converted to a strings column?

davidwendt: Oh wait, I see that you are destroying the `meminfo` right below this line! So yes, then calling the destructor here makes sense.

ajschmidt8 (approved):

PR #13099, opened 2023-04-10
Use make_empty_lists_column instead of make_empty_column(type_id::LIST)

### LINKED ISSUE #13096: [BUG] `make_empty_column` is called on `type_id::LIST` input
By definition, `make_empty_column` doesn't support nested types. However, it was called on `LIST` type in this code:
 * https://github.com/rapidsai/cudf/blob/52e8b5e40b4a728208d17814368535b810a5caa5/cpp/src/strings/split/split_record.cu#L49

The bug still doesn't show up yet, but we should fix it ASAP.

davidwendt:
## Description
Fixes bug where `cudf::make_empty_column(type_id::LIST)` is called and adds a gtests to check for this error.
The `make_empty_column` cannot accept a nested type because it requires a child type. The internal `make_empty_lists_column` is moved to the `lists_column_factories.hpp` header which is itself moved to the `cpp/include/cudf/lists/detail` directory since it only contains detail functions.

Closes #13096 

## Checklist


ttnghia (approved):

ajschmidt8 (approved):

vyasr (approved):
Just making sure that I understand, this is a code path that was never taken before? Otherwise we should have hit the `CUDF_EXPECTS` in `make_empty_column` that prohibits nested types, correct?

davidwendt:
> Just making sure that I understand, this is a code path that was never taken before? Otherwise we should have hit the `CUDF_EXPECTS` in `make_empty_column` that prohibits nested types, correct?

That is correct. I thought I had a test for this as well but missed this case.

PR #13100, opened 2023-04-10
Performance improvement for cudf::string_view::find functions

### LINKED ISSUE #12445: [BUG] Operations with `string_view` involving character counting are very inefficient
Due to the nature of strings columns that they only contain offsets to determine string boundaries, there is no way to know in `O(1)` how many characters were stored in each string. As such, when an operation requires to know the number of characters, an expensive sequential count is executed:
```
return thrust::count_if(
    thrust::seq, ptr, ptr + bytes, [](uint8_t chr) { return is_begin_utf8_char(chr); });
```

Because of that, a simple operation like `substr` can be extremely expensive. One of my tests with 1.4 million (customized) substring ops from just one source string can be slow like this:
```
|  test_size  | Samples |  CPU Time  | Noise |  GPU Time  | Noise |  Elem/s  | bytes_per_second | peak_memory_usage |
|-------------|---------|------------|-------|------------|-------|----------|------------------|-------------------|
|   2^8 = 256 |     34x | 451.036 ms | 1.62% | 450.978 ms | 1.62% |   8.287M |          8287194 |        35.142 MiB |
| 2^12 = 4096 |      1x |   96.397 s |  inf% |   96.393 s |  inf% | 620.353K |           620353 |       562.277 MiB |
```

When I just use the simple way of `memcpy` (which cannot deal with utf8 character, assuming that the input only has ASCII) then the performance went back to be reasonable:
```
|  test_size  | Samples | CPU Time  | Noise | GPU Time  | Noise |  Elem/s  | bytes_per_second | peak_memory_usage |
|-------------|---------|-----------|-------|-----------|-------|----------|------------------|-------------------|
|   2^8 = 256 |   1899x |  7.884 ms | 4.23% |  7.879 ms | 4.22% | 474.344M |        474344125 |        35.142 MiB |
| 2^12 = 4096 |    788x | 19.030 ms | 2.86% | 19.024 ms | 2.86% |   3.143G |       3143281474 |       562.277 MiB |
```

Summary for the tables above: using just `memcpy` for substring without character counting can improve performance from `96` seconds down to `19` milliseconds.

As such, the way of implementing strings column should be improved. We may need to have some efficient way to compute the number of characters in parallel (such as using `thrust::reduce_by_key`), or have a new way to store strings such that we can compute the number of characters in `O(1)` time. 

I don't have a very good idea of how to implement and use those efficiently for various internal `string_view` operations but they should be on the table. Otherwise, we will hit this performance issue more in the future.

davidwendt:
## Description
Improves performance of calls to `cudf::string_view::find` functions for longer strings by minimizing character and byte counting required for resolving the input parameters. The input values are in character positions/counts so range checks do not include bytes that may be in partial characters. Converting these to bytes requires some counting usually from the beginning of the string for any given character position. The code has been reworked to minimizing this counting as much as possible. This improves performance overall by 10% but greatly improves performance for long strings by up to 3x.

Closes #12445

## Checklist


davidwendt:
For `cudf::strings::find`
| rows | width | baseline (ms) | rework (ms) |  x |
|:----:|:----:| ----:| ----:| ----:|
| 4096 |  32   |     0.041 | 0.038 | 1.1 |
| 4096 |  128  |     0.075 | 0.060 | 1.3 |
| 4096 |  512  |     0.234 | 0.169 | 1.4 |
| 4096 |  2048 |     0.948 | 0.465 | 2.0 |
| 4096 |  8192 |      4.20 |  1.40 | 3.0 |
| 32768 |  32   |    0.044 | 0.040 | 1.1 |
| 32768 |  128  |    0.081 | 0.065 | 1.2 |
| 32768 |  512  |    0.254 | 0.197 | 1.3 |
| 32768 |  2048 |     1.32 | 0.523 | 2.5 |
| 32768 |  8192 |     4.73 |  1.48 | 3.2 |
| 262144 |  32   |   0.076 | 0.061 | 1.2 |
| 262144 |  128  |   0.215 | 0.177 | 1.2 |
| 262144 |  512  |    3.80 |  1.40 | 2.7 |
| 262144 |  2048 |    21.3 |  5.70 | 3.7 |
| 2097152 |  32   |  0.325 | 0.215 | 1.5 |
| 2097152 |  128  |   1.24 | 0.944 | 1.3 |
| 2097152 |  512  |   30.6 |  10.0 | 3.1 |
| 16777216 |  32   |  2.31 |  1.44 | 1.6 |

For `cudf::strings::find_multiple`
| rows | width | baseline (ms) | rework (ms) |  x |
|:----:|:----:| ----:| ----:| ----:|
| 4096 |  32 |    0.048 | 0.045 | 1.1 |
| 4096 |  128 |   0.094 | 0.073 | 1.3 |
| 4096 |  512 |   0.255 | 0.195 | 1.3 |
| 4096 |  2048 |  0.737 | 0.433 | 1.7 |
| 4096 |  8192 |   3.36 |  1.14 | 2.9 |
| 32768 |  32 |   0.053 | 0.048 | 1.1 |
| 32768 |  128 |  0.104 | 0.083 | 1.3 |
| 32768 |  512 |  0.302 | 0.230 | 1.3 |
| 32768 |  2048 |  1.19 | 0.585 | 2.0 |
| 32768 |  8192 |  4.13 |  1.51 | 2.7 |
| 262144 |  32 |  0.123 | 0.092 | 1.3 |
| 262144 |  128 | 0.359 | 0.239 | 1.5 |
| 262144 |  512 |  1.55 | 0.993 | 1.6 |
| 262144 |  2048 | 5.66 |  2.70 | 2.1 |
| 2097152 |  32 | 0.695 | 0.439 | 1.6 |
| 2097152 |  128 | 2.27 |  1.37 | 1.7 |
| 2097152 |  512 | 9.72 |  6.30 | 1.5 |
| 16777216 |  32 | 5.28 |  3.21 | 1.6 |

bdice (approved):
Looks good to me. Nice optimization!

ttnghia (approved):

PR #13102, opened 2023-04-10
Replace unnecessary uses of `UNKNOWN_NULL_COUNT`

vyasr:
## Description

This PR replaces uses of `cudf::UNKNOWN_NULL_COUNT` where the null count is either already known or trivially computed.

Contributes to #11968

## Checklist


cpp/src/binaryop/compiled/binary_ops.cu
```diff
@@ -81,7 +81,7 @@ scalar_as_column_view::return_type scalar_as_column_view::operator()<cudf::strin
                            1,
                            nullptr,
                            (bitmask_type const*)s.validity_data(),
-                           cudf::UNKNOWN_NULL_COUNT,
+                           static_cast<size_type>(s.is_valid()),
```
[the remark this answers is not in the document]
vyasr: Good catch, something that my current level of stream testing wouldn't catch but the newer preload checks will once we start testing with public APIs.

bdice:
It'd be nice to know if we're eliminating many "real-world" calls to the `null_count` kernel by supplying these values. Not sure how to measure that well, but perhaps benchmarks would show it?

cpp/src/binaryop/compiled/binary_ops.cu
```diff
@@ -81,7 +81,7 @@ scalar_as_column_view::return_type scalar_as_column_view::operator()<cudf::strin
                            1,
                            nullptr,
                            (bitmask_type const*)s.validity_data(),
-                           cudf::UNKNOWN_NULL_COUNT,
+                           static_cast<size_type>(s.is_valid(stream)),
```
ttnghia, line 84: This should be inverted:
```suggestion
                           static_cast<size_type>(!s.is_valid(stream)),
```

ttnghia: Also, please modify the live above it, using `static_cast` instead of C-style cast.

vyasr: Done.

cpp/src/binaryop/compiled/binary_ops.cu
```diff
@@ -80,8 +80,8 @@ scalar_as_column_view::return_type scalar_as_column_view::operator()<cudf::strin
   auto col_v = column_view(s.type(),
                            1,
                            nullptr,
-                           (bitmask_type const*)s.validity_data(),
-                           static_cast<size_type>(s.is_valid(stream)),
+                           static_cast<bitmask_type const*> s.validity_data(),
```
ttnghia, line 83: Sorry should be `reinterpret_cast`.
```suggestion
                           reinterpret_cast<bitmask_type const*>(s.validity_data()),
```

davidwendt (approved):

ttnghia (approved):

PR #13104, opened 2023-04-10
Explicitly compute null count in concatenate APIs

vyasr:
## Description

The total number of nulls in the output can be computed by summing the nulls in the input columns.

Contributes to #11968

## Checklist


vyasr:
@vuule locally this change passes all tests for concatenation itself but fails some tests in the Parquet reader. Would you be able to help me look into those? I am unsure if the reader is catching some edge cases that the test is missing, or if the Parquet code has some null sanitization bug that is causing the failures. I mention sanitization because the failure specifically occurs for list and struct types.

cpp/src/lists/copying/concatenate.cu
```diff
@@ -123,15 +124,22 @@ std::unique_ptr<column> concatenate(host_span<column_view const> columns,
...
     total_list_count, has_nulls ? mask_state::UNINITIALIZED : mask_state::UNALLOCATED);
+  cudf::size_type null_count{0};
   if (has_nulls) {
     cudf::detail::concatenate_masks(columns, static_cast<bitmask_type*>(null_mask.data()), stream);
+    null_count = std::transform_reduce(
+      columns.begin(),
+      columns.end(),
+      0,
+      [](auto x, auto y) { return x + y; },
+      [](auto col) { return col.null_count(); });
```
ttnghia, line 135: Avoid copying the column_view objs if not needed.
```suggestion
      [](auto const& col) { return col.null_count(); });
```

ttnghia (approved):

vuule:
> @vuule locally this change passes all tests for concatenation itself but fails some tests in the Parquet reader. Would you be able to help me look into those? I am unsure if the reader is catching some edge cases that the test is missing, or if the Parquet code has some null sanitization bug that is causing the failures. I mention sanitization because the failure specifically occurs for list and struct types.

@nvdbaranec

vuule:
Tests fail because chunked Parquet reader returns columns with incorrect null count, and the new concatenate relies on this info.
Opened https://github.com/rapidsai/cudf/pull/13111 with a potential fix.

vyasr:
Depends on #13111

Issue #13030, opened 2023-03-29
[BUG] `cudaErrorIllegalAddress` when trying to write a large list column to json


galipremsagar:
**Describe the bug**
When we try writing a dataframe that has `ListColumn` with large number of rows, we are getting a runtime error that says `cudaErrorIllegalAddress`.

**Steps/Code to reproduce bug**
with small number of rows it works fine:
```python
In [1]: import pandas as pd

In [2]: import cudf

In [3]: import numpy as np

In [4]: size = 10000
   ...: pdf = pd.DataFrame({
   ...:     "lists": [[1, 2], [3], [4, None], None, [1, 2]]*int(size/5),
   ...:     })
   ...: gdf = cudf.from_pandas(pdf)
   ...: gdf.to_json("cudf.json", engine="cudf", orient='records', lines=True)
```

Right as we increase the number of rows we see the issue:
```python
In [5]: size = 10000000
   ...: pdf = pd.DataFrame({
   ...:     "lists": [[1, 2], [3], [4, None], None, [1, 2]]*int(size/5),
   ...:     })
   ...: gdf = cudf.from_pandas(pdf)
   ...: gdf.to_json("cudf.json", engine="cudf", orient='records', lines=True)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[5], line 8
      2 pdf = pd.DataFrame({
      3     "lists": [[1, 2], [3], [4, None], None, [1, 2]]*int(size/5),
      4     })
      7 gdf = cudf.from_pandas(pdf)
----> 8 gdf.to_json("cudf.json", engine="cudf", orient='records', lines=True)

File /nvme/0/pgali/envs/cudfdev/lib/python3.10/contextlib.py:79, in ContextDecorator.__call__.<locals>.inner(*args, **kwds)
     76 @wraps(func)
     77 def inner(*args, **kwds):
     78     with self._recreate_cm():
---> 79         return func(*args, **kwds)

File /nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/cudf/core/frame.py:2548, in Frame.to_json(self, path_or_buf, *args, **kwargs)
   2543 @_cudf_nvtx_annotate
   2544 @ioutils.doc_to_json()
   2545 def to_json(self, path_or_buf=None, *args, **kwargs):
   2546     """{docstring}"""
-> 2548     return cudf.io.json.to_json(
   2549         self, path_or_buf=path_or_buf, *args, **kwargs
[...]

PR #13108, opened 2023-04-10
Fixes sliced list and struct column bug  in JSON chunked writer

karthikeyann:
## Description
Fixes the OOM access error while using chunked JSON writer on list columns.
The issue is present in struct columns also, which is fixed in this change.
Fixes https://github.com/rapidsai/cudf/issues/13030

## Checklist


cpp/src/io/json/write_json.cu
```diff
@@ -470,7 +472,15 @@ struct column_to_strings_fn {
...
-    auto col_string = operator()(column.child_begin(), column.child_end(), children_names);
+    std::vector<cudf::column_view> sliced_children;
+    std::transform(thrust::make_counting_iterator(0),
+                   thrust::make_counting_iterator(column.num_children()),
+                   std::back_inserter(sliced_children),
+                   [structs_view = structs_column_view{column}](auto const idx) {
+                     return structs_view.get_sliced_child(idx);
+                   });
+
+    auto col_string = operator()(sliced_children.begin(), sliced_children.end(), children_names);
```
ttnghia, line 483: We can use `transform_iterator`:
```
auto const child_it = make_counting_transform_iterator(0, [structs_view = structs_column_view{column}](auto const child_idx) { return structs_view.get_sliced_child(child_idx); });

auto const col_string = operator()(child_it, child_it + column.num_children(), children_names);
```

ttnghia (approved):

davidwendt (approved):

PR #13111, opened 2023-04-11
Fix `null_count` of columns returned by `chunked_parquet_reader`

vuule:
## Description
Chunked Parquet reader returns columns with incorrect null counts - the counts are cumulative sums that include all previous chunks.
Root cause is that `nesting_decode_cache` is not copied back to `nesting_decode` when `gpuDecodePageData` returns early, so previously computed null counts are only reset in the cache.
With this PR, we use RAII to make sure cached decode info is always copied back in `gpuDecodePageData`.
Also fixed `column_buffer::empty_like` to return zero null count and empty null mask.

## Checklist

### DIFF
--- a/cpp/src/io/parquet/page_data.cu
+++ b/cpp/src/io/parquet/page_data.cu
@@ -994,7 +994,10 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
   int chunk_idx;
 
   // Fetch page info
-  if (!t) s->page = *p;
+  if (!t) {
+    s->page         = *p;
+    s->nesting_info = nullptr;
+  }
   __syncthreads();
 
   if (s->page.flags & PAGEINFO_FLAGS_DICTIONARY) { return false; }
@@ -1018,9 +1021,6 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
           s->page.nesting_decode[thread_depth].start_depth;
         s->nesting_decode_cache[thread_depth].end_depth =
           s->page.nesting_decode[thread_depth].end_depth;
-        s->nesting_decode_cache[thread_depth].valid_count = 0;
-        s->nesting_decode_cache[thread_depth].value_count = 0;
-        s->nesting_decode_cache[thread_depth].null_count  = 0;
       }
       depth += blockDim.x;
     }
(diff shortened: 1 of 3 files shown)


vyasr:
I can't speak to the logic of the change, but I can confirm that merging these changes make tests pass on #13104.

cpp/src/io/parquet/page_data.cu
```diff
@@ -1018,23 +1018,25 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
...
 
   // zero counts
   int depth = 0;
   while (depth < s->page.num_output_nesting_levels) {
     int const thread_depth = depth + t;
     if (thread_depth < s->page.num_output_nesting_levels) {
-      s->nesting_info[thread_depth].valid_count = 0;
-      s->nesting_info[thread_depth].value_count = 0;
-      s->nesting_info[thread_depth].null_count  = 0;
+      s->page.nesting_decode[thread_depth].valid_count = 0;
```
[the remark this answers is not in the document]
vuule: Thanks. Then I suspect something is wrong with the back copy. I'll look into it.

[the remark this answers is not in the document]
vuule: Thank you for your help reworking the fix :)
Please take another look.

cpp/src/io/parquet/page_data.cu
```diff
@@ -1025,7 +1028,6 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
   if (!t) {
     s->nesting_info = can_use_decode_cache ? s->nesting_decode_cache : s->page.nesting_decode;
   }
-  __syncthreads();
```
nvdbaranec, line 1028: This `__syncthreads()` still needs to be here, `s->nesting_info` is used by other threads right below this

nvdbaranec (commented):
Looks good.  Just the one issue.  I'm going to run this on the spark integration tests before giving the thumbs up.

cpp/src/io/parquet/page_data.cu
vuule: Oops, missed reverting this.

nvdbaranec:
Spark integration tests passed.

nvdbaranec (approved):

cpp/src/io/parquet/page_data.cu
```diff
@@ -994,7 +994,10 @@ static __device__ bool setupLocalPageInfo(page_state_s* const s,
   int chunk_idx;
 
   // Fetch page info
-  if (!t) s->page = *p;
+  if (!t) {
+    s->page         = *p;
+    s->nesting_info = nullptr;
```
vyasr, line 999: So IIUC this is the key change that stops it from looking in the decode cache?

vyasr (approved):
LGTM!

cpp/src/io/parquet/page_data.cu
vuule: Oops, I didn't update the description.
This null prevents `null_count_back_copier` from copying `nesting_decode_cache` to `nesting_decode` when the nesting info hasn't been set up because of an early exit from `setupLocalPageInfo`.

Issue #13103, opened 2023-04-10
[FEA] Attempt to JIT `GroupBy.apply` functions by default and fall back to iterative algorithm


brandon-b-miller:
With https://github.com/rapidsai/cudf/pull/11452 we introduced a framework for JIT compiling groupby UDFs with numba, along with the `GroupBy.apply` `engine='jit'` kwarg. This is an o.k. approach since generally we are alright with introducing things that are a superset of the Pandas API. 

Recently we've discussed changing things so that when a user uses `GroupBy.apply` we try and JIT the UDF first and if it doesn't work, then fall back to the iterative method. This would provide a unified API with less to learn for users and no wondering if the UDF conforms to the restrictions on JIT apply. It also provides an easier internal interface for features that build on top of `GroupBy.apply`, such as `filter`. However it introduces JIT overhead to workflows that ultimately won't even use it. This is not ideal, but iterative groupby apply is pretty slow already.

vyasr:
I'm fine with introducing the extra overhead in the cases that won't JIT to help the cases that do JIT. We could mitigate the issue by introducing a new `engine='auto'` mode that does this, allowing users to opt into `engine='cudf'` if they know that they don't have a UDF that will compile successfully and want to avoid the overhead.

brandon-b-miller:
So would `engine='auto'` be the default?

vyasr:
I think so. If we wanted to err on the side of performance we could default to `jit`, but that seems likely to break many workflows.

PR #13113, opened 2023-04-11
Automatically select `GroupBy.apply` algorithm based on if the UDF is jittable 

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/13103


brandon-b-miller:
hi all, this should be ready for review.

python/cudf/cudf/core/udf/groupby_utils.py
```diff
@@ -202,3 +203,30 @@ def jit_groupby_apply(offsets, grouped_values, function, *args):
...
+            _supported_dtypes_from_frame(
+                frame, supported_types=SUPPORTED_GROUPBY_NUMPY_TYPES
+            ).items()
+        )
+    )
+    dataframe_group_type = _get_frame_groupby_type(
+        np_field_types, frame.index.dtype
+    )
+    try:
+        _get_udf_return_type(dataframe_group_type, func, args)
```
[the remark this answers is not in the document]
vyasr: Not necessary for this PR, but now that we're calling this function (`_get_udf_return_type`) more often is there a way to make it also cache the results like `_compile_udf` does? I suspect that this function isn't the cheapest.

vyasr (approved):
One minor suggestion and a question, but LGTM.

python/cudf/cudf/core/udf/groupby_utils.py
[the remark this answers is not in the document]
brandon-b-miller: this goes through `cudautils.compile_udf` which also caches, though it's not the same cache :(. One of the caches holds the result of compiling the UDF separately from any kernel for the purposes of getting the return type and the other holds the compilation result of the full kernel.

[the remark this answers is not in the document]
vyasr: Yup I realize that they're two different functions with one calling the other. I'm wondering whether we should add a cache for this function as well in case there is significant overhead in the parts of this function _aside_ from the call to `compile_udf` since now we're calling this one (possibly a lot) more frequently.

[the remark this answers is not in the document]
brandon-b-miller: ah I see what you mean.

I think it'd certainly be worth going through where we're caching things and doing some profiling while we assess where it's the most useful to do so at this point, given the proliferation of calls to numba's compilation machinery in our udf code. Especially as we eventually get on disk caching too. I'll raise a separate issue about this.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -1144,11 +1147,9 @@ def _jit_groupby_apply(
         self, function, group_names, offsets, group_keys, grouped_values, *args
     ):
         # Nulls are not yet supported
-        for colname in self.grouping.values._data.keys():
-            if self.obj._data[colname].has_nulls():
-                raise ValueError(
-                    "Nulls not yet supported with groupby JIT engine"
-                )
+        # TODO: don't check this twice under `engine='auto'`
```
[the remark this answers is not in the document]
brandon-b-miller: I thought this might be a perf hit but since each column caches its null count I think it's actually ok to just leave it.

wence- (approved):
Thanks!

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -1334,10 +1336,20 @@ def mult(df):
...
         """
+
+        if self.obj.empty:
+            return self.obj
         if not callable(function):
             raise TypeError(f"type {type(function)} is not callable")
         group_names, offsets, group_keys, grouped_values = self._grouped()
 
+        if engine == "auto":
+            if (not grouped_values._has_nulls) and _can_be_jitted(
```
bdice, line 1347: Should this null check be part of the definition of `_can_be_jitted`?

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -1252,7 +1249,7 @@ def apply(self, function, *args, engine="cudf"):
           on the grouped chunk.
         args : tuple
             Optional positional arguments to pass to the function.
-        engine: {'cudf', 'jit'}, default 'cudf'
+        engine: {'cudf', 'jit'}, default 'auto'
```
bdice, line 1252: For my knowledge: Is there a performance cost to the fallback? i.e. Does the JIT attempt have measurable overhead?

brandon-b-miller: Yes, it does have measurable overhead. One way of measuring it is with
```python
import cudf
df = cudf.DataFrame({
    'a':[0,1,1],
    'b':[1,2,3]
})

def func(grp):
    # binops can't be jitted without refcounting
    return grp + grp

grouped = df.groupby('a')

import cProfile
cProfile.run('grouped.apply(func)', sort='cumtime')
```

For this I get 
```
        1    0.000    0.000    0.213    0.213 groupby_utils.py:207(_can_be_jitted)
```

Meaning it's quite impactful. However if this becomes a problem users should be able to obtain the old behavior by just passing `engine='cudf'`.

python/cudf/cudf/core/groupby/groupby.py
brandon-b-miller: I kept it separate here because it's only the groupby pipeline that doesn't have null support. If we wanted to use this function for basic `apply` we'd want it to pass through frames with nulls.

python/cudf/cudf/core/groupby/groupby.py
vyasr: Full disclosure, we anticipated this and [I was OK with it](https://github.com/rapidsai/cudf/issues/13103#issuecomment-1502096723). I think the tradeoff is generally worthwhile. If we think it isn't then I think we'd just stop this work and remove 'auto' altogether.

Also I just noticed that 'auto' is not listed in the set of valid `engine` arguments here.

bdice: Just to clarify — your snippet above is saying the JIT attempt costs ~200ms? That sounds right to me. I would think a cache could also be used here if needed to prevent multiple failed attempts from paying the overhead for the same function.

I am supportive of this change, because when it does pay off, it’s a big win. Just want to make sure we’re putting in the appropriate amount of engineering effort to mitigate the downside risk.

wence-: > Also I just noticed that 'auto' is not listed in the set of valid engine arguments here.

Fixed this.

bdice (approved):

PR #13115, opened 2023-04-11
Adds checks to make sure json reader won't overflow

elstehle:
## Description
The JSON reader is currently using 32-bit offsets to index into the input's characters to lower memory footprint and for performance reasons. Hence, currently, if an input larger than `UINT_MAX` is read, the parser may return incorrect data. 

This PR adds a check that fails for inputs that could overflow. 

The longer term plan is to make the finite-state transducer stage reentrant and split up inputs larger than `UINT_MAX` into smaller chunks.

## Checklist

### DIFF
--- a/cpp/src/io/json/nested_json_gpu.cu
+++ b/cpp/src/io/json/nested_json_gpu.cu
@@ -70,6 +70,20 @@ struct tree_node {
 
   std::size_t num_children = 0;
 };
+
+/**
+ * @brief Verifies that the JSON input can be handled without corrupted data due to offset
+ * overflows.
+ *
+ * @param input_size The JSON inputs size in bytes
+ */
+void check_input_size(std::size_t input_size)
+{
+  // Transduce() writes symbol offsets that may be as large input_size-1
+  CUDF_EXPECTS(input_size == 0 ||
+                 (input_size - 1) <= std::numeric_limits<cudf::io::json::SymbolOffsetT>::max(),
+               "Given JSON input is too large");
+}
 }  // namespace
 
 namespace cudf::io::json {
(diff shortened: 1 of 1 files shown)


cpp/src/io/json/nested_json_gpu.cu
```diff
@@ -1076,20 +1082,25 @@ std::pair<rmm::device_uvector<PdaTokenT>, rmm::device_uvector<SymbolOffsetT>> ge
   rmm::cuda_stream_view stream,
   rmm::mr::device_memory_resource* mr)
 {
+  // Transduce() writes symbol offsets that may be as large json_in.size()-1
+  CUDF_EXPECTS(
+    json_in.size() == 0 || (json_in.size() - 1) <= std::numeric_limits<SymbolOffsetT>::max(),
+    "Given JSON input is too large");
```
vuule, line 1088: Maybe move this into a function to avoid code duplication

cpp/src/io/json/nested_json_gpu.cu
vuule, line 1: Could we have a test for this? What would be the peak memory use with a 2GB+1B `device_buffer_source`?

vuule (commented):
Thanks for the PR! 
Posted a few suggestions.

cpp/src/io/json/nested_json_gpu.cu
elstehle: Done 👍

cpp/src/io/json/nested_json_gpu.cu
elstehle: As per offline discussion, we decided that the risk of introducing a potentially flaky test that may OOM for `4 GB + 1 B` inputs is not worth it for testing such a simple check.

vuule (approved):

karthikeyann (approved):

PR #13117, opened 2023-04-11
Fix GPU_ARCHS setting in Java CMake build and CMAKE_CUDA_ARCHITECTURES in Python package build.

davidwendt:
## Description
Changes the `GPU_ARCHS` setting in `pom.xml` from `ALL` to `RAPIDS` per recent change in rapidsai/rapids-cmake/pull/397

The Python package build requires CUDA as of the addition of string UDFs, which added compilation of ptx code to the Python build. Therefore CMAKE_CUDA_ARCHITECTURES must be set appropriately even when only the Python package is being built. https://github.com/rapidsai/rapids-cmake/pull/397 requires a nonempty string value to be used if the variable is set at all. This PR updates build.sh to include the appropriate default when only the Python build is requested.

## Checklist

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -257,7 +257,7 @@ fi
 ################################################################################
 # Configure, build, and install libcudf
 
-if buildAll || hasArg libcudf || hasArg cudfjar; then
+if buildAll || hasArg libcudf || hasArg cudf || hasArg cudfjar; then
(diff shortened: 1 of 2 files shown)


bdice (approved):

bdice (approved):
Re-approving, but ~let's mention Python / cudf in the PR title rather than only Java.~ Done.

Issue #13109, opened 2023-04-10
[BUG] Left joins on struct key producing incorrect null results for right table


jlowe:
**Describe the bug**
After #12787 some RAPIDS Accelerator tests started failing for joins on struct keys.  See NVIDIA/spark-rapids#8061.  Examining the expected vs. actual data, there are nulls in the GPU result that are not expected, implying the row comparison is not working properly in some cases.

Oddly, an inner join on the same data seems to do the right thing.

**Steps/Code to reproduce bug**
I'm attaching two Parquet files, left and rightnotnull which are the the inputs to the following test.  There should be 4 rows that have matches in the join results, but the GPU produces zero rows that have matches instead, so the right gather map indicates (incorrectly) all null rows.

[left.gz](https://github.com/rapidsai/cudf/files/11193921/left.gz)
[rightnotnull.gz](https://github.com/rapidsai/cudf/files/11193922/rightnotnull.gz)

```
#include <iostream>

#include <cudf/join.hpp>
#include <cudf/io/parquet.hpp>

int main(int argc, char** argv) {
  auto left_table = cudf::io::read_parquet(
    cudf::io::parquet_reader_options::builder(cudf::io::source_info("left")));
  auto right_table = cudf::io::read_parquet(
    cudf::io::parquet_reader_options::builder(cudf::io::source_info("rightnotnull")));
  auto left_keys = cudf::table_view({left_table.tbl->get_column(0)});
  auto right_keys = cudf::table_view({right_table.tbl->get_column(0)});
  auto [gather_left, gather_right] = cudf::left_join(left_keys, right_keys);
  std::cout << "LEFT GATHER MAP:" << std::endl;
  for (int i = 0; i < gather_left->size(); i++) {
    std::cout << i << ": " << gather_left->element(i, rmm::cuda_stream_default) << std::endl;
  }
  std::cout << "RIGHT GATHER MAP:" << std::endl;
  for (int i = 0; i < gather_right->size(); i++) {
    std::cout << i << ": " << gather_right->element(i, rmm::cuda_stream_default) << std::endl;
  }
  return 0;
}
```

**Expected behavior**
Instead of producing all null results (i.e.: minint for all entries in the right gather map), there should be four not-null matches in the right gather map.  Change to an inner join to see the expected matches.

PR #11284, opened 2022-07-16
Fix inconsistency when hashing two tables in `cudf::detail::contains`

ttnghia:
When hashing elements in a column, the nullable information needs to be taken into account. This could produce different results for the hash values of the same elements if the nullable condition is changed. As such, whenever we need to compute hashing for more than one column/table in the same operation, we need to take into account the nullable of all columns/tables, not just the one that is being hashed.

This PR fixes a bug when hashing two tables using different nullable values in `cudf::detail::contains`, which led to incorrect results when there is only one nullable table in the input.

PR #13120, opened 2023-04-11
Fix hash join when the input tables have nulls on only one side

ttnghia:
This is very similar to https://github.com/rapidsai/cudf/pull/11284, which fixes a bug when only one input table has nulls while the other doesn't. This is due to the new experimental hasher producing different hash values depending on an input flag `has_nulls`. In order to properly use it, `has_nulls` must be computed by checking all the possible input tables, or set to a constant value (`true`).

Closes:
 * https://github.com/rapidsai/cudf/issues/13109


cpp/include/cudf/join.hpp
```diff
@@ -282,10 +282,13 @@ class hash_join {
    * undefined.
    *
    * @param build The build table, from which the hash table is built
+   * @param has_nulls Flag to indicate if the there exists any nulls in the `build` table or
+   *        any table that will be used later for join
    * @param compare_nulls Controls whether null join-key values should match or not
    * @param stream CUDA stream used for device memory operations and kernel launches
    */
   hash_join(cudf::table_view const& build,
+            std::optional<bool> has_nulls,
```
PointKernel, line 291: optional bool seems an overkill

cpp/src/join/hash_join.cu
```diff
@@ -381,11 +381,14 @@ hash_join<Hasher>::hash_join(cudf::table_view const& build,
 
   if (_is_empty) { return; }
 
+  auto const row_bitmask = std::move(
```
PointKernel, line 384: redundant move

cpp/tests/join/join_tests.cpp
```diff
@@ -1041,7 +1041,7 @@ TEST_F(JoinTest, EmptyRightTableInnerJoin)
   }
 
   {
-    cudf::hash_join hash_join(empty1, cudf::null_equality::EQUAL);
+    cudf::hash_join hash_join(empty1, true, cudf::null_equality::EQUAL);
```
PointKernel, line 1044: We should add an enum for this instead of a raw boolean type exposed to the public.

cpp/src/join/hash_join.cu
```diff
@@ -458,19 +463,21 @@ std::size_t hash_join<Hasher>::left_join_size(cudf::table_view const& probe,
   // Trivial left join case - exit early
   if (_is_empty) { return probe.num_rows(); }
 
+  CUDF_EXPECTS(_has_nulls || !cudf::has_nested_nulls(probe),
+               "Probe table has nulls while build table was not hashed with null check.");
```
PointKernel, line 467: docs for throw

cpp/tests/join/join_tests.cpp
ttnghia: Okay then I'll use `nullate::DYNAMIC`.

cpp/include/cudf/join.hpp
jlowe: Agreed.  If we're going to allow the user to avoid specifying this, I'd prefer to not break the API and instead overload this method, keeping the original and adding a new one that has the new nulls present/absent parameter.

cpp/tests/join/join_tests.cpp
divyegala: Can we take out `nullate::DYNAMIC` from `column_device_view.cuh` and put it in another header? We use it quite a lot and it doesn't seem to have any device-specific implementations anyway. This doesn't have to be in this PR.

cpp/src/join/hash_join.cu
```diff
@@ -574,7 +588,16 @@ hash_join::~hash_join() = default;
 hash_join::hash_join(cudf::table_view const& build,
                      null_equality compare_nulls,
                      rmm::cuda_stream_view stream)
-  : _impl{std::make_unique<const impl_type>(build, compare_nulls, stream)}
+  // If we cannot know beforehand about null existence then let's assume that there are nulls.
+  : hash_join(build, true /*has_nulls*/, compare_nulls, stream)
```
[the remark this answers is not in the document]
ttnghia: Either way. This constructor calls the other constructor which in turn initializes `_impl`.

cpp/tests/join/join_tests.cpp
ttnghia: I thought about that too. Where to put depends on discussion.

cpp/src/join/hash_join.cu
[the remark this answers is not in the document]
divyegala: Ah, okay. We have a `cudf::hash_join` and a `cudf::detail::hash_join` class so it is a bit difficult to follow. Thanks for clearing it up!

cpp/tests/join/join_tests.cpp
ttnghia: Should be out of scope of this PR. I'll open another one for it.

divyegala (approved):

cpp/include/cudf/detail/join.hpp
```diff
@@ -73,10 +73,10 @@ struct hash_join {
   hash_join& operator=(hash_join&&) = delete;
 
  private:
-  bool const _is_empty;                         ///< true if `_hash_table` is empty
-  rmm::device_buffer const _composite_bitmask;  ///< Bitmask to denote whether a row is valid
-  cudf::null_equality const _nulls_equal;       ///< whether to consider nulls as equal
-  cudf::table_view _build;                      ///< input table to build the hash map
+  bool const _is_empty;                    ///< true if `_hash_table` is empty
+  bool const _has_nulls;                   ///< true if there is any nulls
```
PointKernel, line 77:
```suggestion
  bool const _has_nulls;                   ///< true if nulls are present in either build table or probe table
```

PointKernel (approved):
LGTM

PR #13123, opened 2023-04-12
Use ARC V2 self-hosted runners for GPU jobs

jjacobelli:
This PR is updating the runner labels to use ARC V2 self-hosted runners for GPU jobs. This is needed to resolve the auto-scalling issues.


ajschmidt8 (approved):

bdice (commented):
Is there no longer a way to specify “use the latest driver version, any GPU is fine”?

bdice:
> arc-v2 only allows a single label per runner set.

Then does "latest" make sense, or should we always need to specify the driver version?

ajschmidt8:
> > arc-v2 only allows a single label per runner set.
> 
> Then does "latest" make sense, or should we always need to specify the driver version?

We will no longer be using driver-specific labels at all for arc-v2.

There are more details in the PR below (the PR was closed before the force-push that was recently done to that repository, but will be rebased and re-opened soon)

- https://github.com/rapidsai/docs/pull/357

PR #13124, opened 2023-04-12
Compute null-count in cudf::detail::slice

davidwendt:
## Description
Calculates the null-count in the `cudf::detail::slice()` function. This requires adding a stream parameter to the function and updating the callers to pass the stream. Also moved the function definition to the `slice.cu` file since there are only two possible values for the template parameter.

Labeling this with non-breaking since it is a detail function.

## Checklist

### DIFF
--- a/cpp/src/copying/slice.cu
+++ b/cpp/src/copying/slice.cu
@@ -49,16 +49,14 @@ ColumnView slice(ColumnView const& input,
     children.emplace_back(input.child(index));
   }
 
-  size_type nulls =
-    input.null_count() == 0 ? 0 : cudf::detail::null_count(input.null_mask(), begin, end, stream);
-
-  return ColumnView(input.type(),
-                    end - begin,
-                    input.head(),
-                    input.null_mask(),
-                    nulls,
-                    input.offset() + begin,
-                    children);
+  return ColumnView(
+    input.type(),
+    end - begin,
+    input.head(),
+    input.null_mask(),
+    input.null_count() ? cudf::detail::null_count(input.null_mask(), begin, end, stream) : 0,
+    input.offset() + begin,
+    children);
 }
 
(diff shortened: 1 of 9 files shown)


cpp/src/copying/slice.cu
```diff
@@ -31,6 +32,44 @@
...
+  CUDF_EXPECTS(end <= input.size(), "Slice range out of bounds.");
+
+  std::vector<ColumnView> children{};
+  children.reserve(input.num_children());
+  for (size_type index = 0; index < input.num_children(); index++) {
+    children.emplace_back(input.child(index));
+  }
+
+  size_type nulls =
+    input.null_count() == 0 ? 0 : cudf::detail::null_count(input.null_mask(), begin, end, stream);
```
divyegala, line 53: Can you explain what is happening here? If `input.null_count != 0`, why are we computing the null count? It reads to me as if it's already known. If there is some default initialization value here then perhaps this will read cleaner like this, implying that we are always choosing a known null count:
```suggestion
    input.null_count() == UNKNOWN_NULL_COUNT ? cudf::detail::null_count(input.null_mask(), begin, end, stream) : input.null_count();
```

ttnghia: This is the `slice` function which returns a segment view of the input in the range `[begin, end)`. The return null count (`nulls`) is counted nulls in that segment.

divyegala: Got it, yes. I just don't like the pattern `... == 0 ? 0 : ...`. Then my suggestion is
```suggestion
    input.null_count() ? cudf::detail::null_count(input.null_mask(), begin, end, stream) : 0;
```

ttnghia (approved):

divyegala (approved):

vyasr (approved):
LGTM

PR #13125, opened 2023-04-12
Enable NRT for the numba CUDA target on cuDF import

brandon-b-miller:
Depends on https://github.com/rapidsai/cudf/pull/13098

This PR introduces `_nrt_cuda.py` and adds it to `cudf/core/udf/__init__.py`. This makes it so that upon `cudf` import, several things happen:

- numba's `cuda_target.target_context.enable_nrt` is set to true. This has the effect of telling numba to inspect all of the types in the code it's compiling to see if they are backed by structs containing a `MemInfo` member. When a variable in the user's UDF is detected as being an instance of such a type, numba will insert `incref` and `decref` calls targeting this `MemInfo` in the compiled code. 
- Several functions defined in numba's IR which serve as higher level wrappers for `incref` and `decref` are defined and wrapped in a numba `library` object, which has the ability to produce PTX on demand.
- Numba's linker is further patched such that the PTX from the above library is linked to every kernel that is launched


python/cudf/cudf/core/udf/_nrt_cuda.py
```diff
@@ -0,0 +1,148 @@
...
+CUDATargetContext.enable_nrt = True
+# compile NRT functions written in lowering into a PTX string
+library = compile_nrt_functions(descriptor.cuda_target.target_context)
+_cuda_nrt_library = compile_nrt_functions(
+    descriptor.cuda_target.target_context
+)
+numba_cuda_runtime = _cuda_nrt_library._get_ptxes()[0].encode()
+
+
+def install_registry(self, registry):
```
brandon-b-miller, line 119: Replacing this method within numba's internals might not be necessary for this feature to work, will figure it out and post my findings here.

bdice:
@brandon-b-miller I may be missing context here -- can we define what NRT is in the PR description, and why we care about it? I understand this PR depends on another unmerged PR, so maybe I'm just unaware of the term.

brandon-b-miller:
> @brandon-b-miller I may be missing context here -- can we define what NRT is in the PR description, and why we care about it? I understand this PR depends on another unmerged PR, so maybe I'm just unaware of the term.

I updated this with a little context, I'm guessing there still might be gaps so let me know what's lacking :)

vyasr:
NRT is the numba runtime. It's how numba handles things like Python reference counting in nopython mode to ensure that you still get Pythonic behavior from the C objects used in nopython mode.

python/cudf/cudf/core/udf/_nrt_cuda.py
```diff
@@ -0,0 +1,148 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
```
shwina, line 1: Are we eventually hoping for this whole module to go away and just be a part of Numba?

python/cudf/cudf/core/udf/__init__.py
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) 2022-2023, NVIDIA CORPORATION.
 from . import (
+    _nrt_cuda,
```
shwina, line 3: Do we need this here?

gmarkall: Importing `_nrt_cuda` has side-effects, so it's needed somewhere during import. An argument could be made that the side-effects of its import should be moved into a function (e.g. `init_cuda_nrt()` or whatever) then the function called from this file like `_nrt_cuda.init_cuda_nrt()` so that the effect is more transparent.

This pattern of imports having side-effects that add to Numba's registries or otherwise change / augment it is a bit pervasive in Numba, but it wasn't a good idea and we'd do it differently if we started again - it may be better not to bring that pattern over into cuDF.

python/cudf/cudf/core/udf/_nrt_cuda.py
gmarkall: Eventually I would like to see that but there's a reasonable gap between "enough functionality for cuDF's purposes" and "the complete Numba runtime functionality" that would be needed for the CUDA target to support it in general. I see this as a step along the way to adding that support.

brandon-b-miller:
I managed to cut down `_nrt_cuda.py` a bit as I found we don't need to install any of these numpy registries nor hack any of numba's internals to get our refcounting working for strings. The removed code was mostly a holdover from the rest of our experiments refcounting general numpy arrays in kernels and so forth.

python/cudf/cudf/core/udf/__init__.py
brandon-b-miller: The important effect here is that `enable_nrt` is set to `True` in the cuda target context at import.

python/cudf/cudf/core/udf/_nrt_cuda.py
brandon-b-miller: A lot of this can be upstreamed eventually yes. With NRT enabled, our cuda target contexts would already have `enable_nrt=True` and presumably the extra PTX we're feeding numba's linker here would be present for all cuda kernel launches.

gmarkall: The extra PTX would at least be added when needed. I think if this logic were in Numba it would be much easier to add it exactly when it's required, rather than adding it to every link regardless.

ajschmidt8 (approved):

PR #13127, opened 2023-04-12
Pin curand version

### LINKED ISSUE #13126: [BUG] Compile issue with `curand.h`
Today I saw CI failed with this error:
```
In file included from $SRC_DIR/cpp/benchmarks/join/join_common.hpp:19,
                 from $SRC_DIR/cpp/benchmarks/join/left_join.cu:17:
$SRC_DIR/cpp/benchmarks/join/generate_input_tables.cuh:29:10: fatal error: curand.h: No such file or directory
   29 | #include <curand.h>
```

vyasr:
## Description

Merging the conda-forge curand recipe and building conda-forge packages has caused conda to choose a newer version of curand than what cudf currently supports (we cannot use the version from CUDA 12).

Closes #13126 

## Checklist

### DIFF
--- a/conda/recipes/libcudf/meta.yaml
+++ b/conda/recipes/libcudf/meta.yaml
@@ -388,7 +388,6 @@ outputs:
         - {{ pin_subpackage('libcudf_kafka', exact=True) }}
         - cudatoolkit {{ cuda_spec }}
         - libcurand {{ libcurand_run_version }}
-        - libcurand-dev {{ libcurand_run_version }}
     about:
       home: https://rapids.ai/
       license: Apache-2.0
--- a/conda/recipes/libcudf/meta.yaml
+++ b/conda/recipes/libcudf/meta.yaml
@@ -45,7 +45,7 @@ requirements:
     - cuda-nvrtc ={{ cuda_version }}
     - cuda-nvrtc-dev ={{ cuda_version }}
     - cuda-nvtx ={{ cuda_version }}
-    - libcurand-dev ={{libcurand_version}}
+    - libcurand-dev {{libcurand_version}}
     - libarrow {{ libarrow_version }}
     - dlpack {{ dlpack_version }}
     - librdkafka {{ librdkafka_version }}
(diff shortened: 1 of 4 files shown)


raydouglass (approved):

robertmaynard:
Fixes #https://github.com/rapidsai/cudf/issues/13126

robertmaynard (approved):

conda/recipes/libcudf/meta.yaml
```diff
@@ -45,7 +45,7 @@ requirements:
     - cuda-nvrtc ={{ cuda_version }}
     - cuda-nvrtc-dev ={{ cuda_version }}
     - cuda-nvtx ={{ cuda_version }}
-    - libcurand-dev
+    - libcurand-dev {{libcurand_version}}
```
bdice, line 48: We must pin both libcurand and libcurand-dev. Follow raft’s pinnings and naming scheme: https://github.com/rapidsai/raft/blob/a98295b516ef58bc855177077860bab2a2a76d77/conda/recipes/libraft/meta.yaml#L58
```suggestion
    - libcurand-dev {{ libcurand_version }}
```

vyasr: We're discussing this offline, but for context I had done the minimal fix here to unblock release while investigating exactly where we need libcurand. I'm fairly certain it's only needed in benchmarks.

vyasr: To be complete I went ahead and added the run pinnings for libcudf-tests. I'm guessing that nobody actually runs the benchmarks from the package, which is why no issues have ever been observed with the missing requirements.

conda/recipes/libcudf/meta.yaml
```diff
@@ -386,6 +387,8 @@ outputs:
         - {{ pin_subpackage('libcudf', exact=True) }}
         - {{ pin_subpackage('libcudf_kafka', exact=True) }}
         - cudatoolkit {{ cuda_spec }}
+        - libcurand {{ libcurand_run_version }}
+        - libcurand-dev {{ libcurand_run_version }}
```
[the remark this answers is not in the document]
vyasr: That's what I would have expected too, but they are currently included in raft so I matched that as requested. I haven't verified the contents of the CUDA 11 packages to know if they are properly split up in that respect. I can try it if you would like, but given that CI is almost done running at this point I would prefer to experiment on a separate PR going into 23.06 so that we can unblock the release.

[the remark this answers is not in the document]
vyasr: Never mind, I see that you requested additional changes. I'll just make them in this PR.

PR #13132, opened 2023-04-13
Add log messages about cuIO's nvCOMP and cuFile use

vuule:
## Description
TBD

## Checklist

### DIFF
--- a/cpp/src/io/comp/nvcomp_adapter.cpp
+++ b/cpp/src/io/comp/nvcomp_adapter.cpp
@@ -125,6 +125,16 @@ auto batched_decompress_async(compression_type compression, Args&&... args)
   }
 }
 
+std::string compression_type_name(compression_type compression)
+{
+  switch (compression) {
+    case compression_type::SNAPPY: return "Snappy";
+    case compression_type::ZSTD: return "Zstandard";
+    case compression_type::DEFLATE: return "Deflate";
+  }
+  return "compression_type(" + std::to_string(int(compression)) + ")";
+}
+
 size_t batched_decompress_temp_size(compression_type compression,
                                     size_t num_chunks,
                                     size_t max_uncomp_chunk_size,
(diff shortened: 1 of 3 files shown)


cpp/src/io/comp/nvcomp_adapter.cpp
```diff
@@ -439,8 +449,21 @@ feature_status_parameters::feature_status_parameters()
...
+struct hash_feature_status_inputs {
+  size_t operator()(feature_status_inputs const& fsi) const
+  {
+    // Outside of unit tests, the same `feature_status_parameters` value will always be passed
+    // within a run; for simplicity, only use `compression_type` for the hash
+    return std::hash<compression_type>{}(fsi.first);
+  }
+};
+
+using feature_status_memo_map =
```
ttnghia, line 462: Please add comment for it: What is this?

cpp/src/io/comp/nvcomp_adapter.hpp
```diff
@@ -55,6 +55,19 @@ struct feature_status_parameters {
   }
 };
 
+/**
+ * @brief Equality operator overload. Required to use `feature_status_parameters` as a map key.
+ */
+inline bool operator==(const feature_status_parameters& lhs, const feature_status_parameters& rhs)
```
ttnghia, line 61: east const.

cpp/src/io/comp/nvcomp_adapter.hpp
```diff
@@ -55,6 +55,19 @@ struct feature_status_parameters {
...
+{
+  return lhs.lib_major_version == rhs.lib_major_version and
+         lhs.lib_minor_version == rhs.lib_minor_version and
+         lhs.lib_patch_version == rhs.lib_patch_version and
+         lhs.are_all_integrations_enabled == rhs.are_all_integrations_enabled and
+         lhs.are_stable_integrations_enabled == rhs.are_stable_integrations_enabled and
+         lhs.compute_capability_major == rhs.compute_capability_major;
+}
```
ttnghia, line 69: I see that this cover all the members of `feature_status_parameters`. So the compiler should generate this operator `==` for you.

cpp/src/io/utilities/file_io_utilities.cpp
```diff
@@ -273,9 +273,18 @@ std::unique_ptr<cufile_input_impl> make_cufile_input(std::string const& filepath
 {
   if (cufile_integration::is_gds_enabled()) {
     try {
-      return std::make_unique<cufile_input_impl>(filepath);
+      auto const cufile_in = std::make_unique<cufile_input_impl>(filepath);
+      CUDF_LOG_INFO("File successfully opened for reading with GDS.");
```
ttnghia, line 277: Do we need this log? It succeeds, and nothing special here?

cpp/src/io/comp/nvcomp_adapter.hpp
vuule: yeah, it didn't work without the overload unfortunately.

cpp/src/io/utilities/file_io_utilities.cpp
vuule: The goal of the messages in this PR is to provide some transparency into what method is used for file IO. We have direct use of GDS (this branch), no GDS (host IO + memcpy), and kvikIO. I included messages that make this selection apparent. FWIW, this is not the default branch, so very few users/devs will see this message.
I guess we can derive that lack of messages means that GDS is used, but I would prefer to log this case as INFO (off by default). If we find that this is too verbose, I'll remove some of these.

ttnghia (approved):

karthikeyann (approved):

PR #13133, opened 2023-04-13
Update clang-format to 16.0.1.

bdice:
This PR updates the clang-format version used by pre-commit.


cpp/scripts/run-clang-format.py
bdice, line 1: Can we remove this file? Our style enforcement is centrally handled by pre-commit. I just don't want to delete if some folks happen to rely on it.

ttnghia (approved):

cpp/scripts/run-clang-format.py
ttnghia: I use format-on-save and never touch that file.

ttnghia:
The output of new clang version is much better IMO.

elstehle (approved):

cpp/scripts/run-clang-format.py
bdice: Ok. I will delete it.

PR #13138, opened 2023-04-14
[REVIEW] Raise error when there is a binary operation between certain `DataFrame` and `Series` objects

galipremsagar:
## Description

This PR raises an error when there is a binary operation performed between `DataFrame` & `Series` with unequal `columns` and `index` respectively.

This PR fixes 120 pytests:
```
= 833 failed, 86451 passed, 2034 skipped, 968 xfailed, 165 xpassed in 490.86s (0:08:10) =
```
on `pandas_2.0_feature_branch`:
```
= 953 failed, 86307 passed, 2034 skipped, 992 xfailed, 165 xpassed in 511.09s (0:08:31) =
```

## Checklist


mroeschke (approved):

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -5361,39 +5362,32 @@ def test_cov_nans():
...
+    pdf = gdf.to_pandas(nullable=True)
 
-        psr = gsr.to_pandas(nullable=True)
+    psr = gsr.to_pandas(nullable=True)
 
+    try:
         expect = op(pdf, psr)
+    except ValueError:
+        with pytest.raises(ValueError):
+            got = op(gdf, gsr).to_pandas(nullable=True)
```
vyasr, line 5379: Is this failing on the `op` or the `to_pandas` call? I assume the former, in which case we should be able to remove the latter altogether?

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -5361,39 +5362,32 @@ def test_cov_nans():
...
+        with pytest.raises(ValueError):
+            expect = op(psr, pdf)
+        with pytest.raises(ValueError):
+            got = op(gsr, gdf).to_pandas(nullable=True)
```
vyasr, line 5383: Same question as above re: to_pandas.

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -5361,39 +5362,32 @@ def test_cov_nans():
...
+    else:
         got = op(gdf, gsr).to_pandas(nullable=True)
-        assert_eq(expect, got, check_dtype=False)
+        assert_eq(expect, got, check_dtype=False, check_like=True)
```
vyasr, line 5386: Why do we need `check_like=True`? Isn't the whole point of the new error to prohibit binops between cases where the index and columns are unequal, whereas that's what `check_like=True` allows? Or is `check_like=True` still stricter than before because it allows reordering but not missing elements or something like that?

python/cudf/cudf/core/dataframe.py
```diff
@@ -1943,6 +1943,20 @@ def _make_operands_and_index_for_binop(
...
+                and (
+                    not self._data.to_pandas_index().equals(
+                        other.index.to_pandas()
+                    )
+                )
+            ):
+                raise ValueError(
+                    "Can only compare DataFrame & Series objects "
+                    "whose columns & index are same respectively, "
+                    "please re-index."
```
vyasr, line 1958:
```suggestion
                    "please reindex."
```

vyasr (changes requested):

vyasr (approved):

vyasr:
Needs a style fix, then should be good.

galipremsagar:
> Needs a style fix, then should be good.

The current style failures are of other files that are untouched in this PR. Part of the problem is style checking seems to be broken for feature branches locally. I'm planning to perform the style fixes all in a single pr later to this feature branch. So I will go ahead and merge this for now.

PR #13143, opened 2023-04-14
STRING order-by column for RANGE window functions

mythrocks:
## Description
This commit adds support to use `STRING` columns as the order-by column in range-based window functions.

## Context
A range-based window function allows a variable-width window to be defined around each row of an input column. The "range" describes the range of values in an order-by column.

Consider the following input:

```c++
agg_column = {  1,   1,   1,   1,   1,   1,   1 };   // The aggregation column.
oby_column = { "0", "1", "2", "3", "4", "5", "6" };  // The order-by column.
//                        ^
```
If the range is defined as `[preceding=∞, following=0]`, then for the row at index `2`, the window would be rows `0-2`.

The supported range bounds for `STRING` order-by column are:
1. `UNBOUNDED` (i.e. extending to the beginning/end of the column/group) 
2. `CURRENT ROW`, indicating that all equivalent values of the current row in the `oby_column`.

Since `STRING` columns cannot have "delta" values, only `UNBOUNDED` and `CURRENT ROW` based window bounds are supported, unlike for numeric types which could specify a specific scalar delta value as a range.

Depends on #13095.

## Checklist


cpp/src/rolling/grouped_rolling.cu
```diff
@@ -869,10 +877,22 @@ std::unique_ptr<column> grouped_range_rolling_window_impl(
   rmm::cuda_stream_view stream,
   rmm::mr::device_memory_resource* mr)
 {
+  /*
   auto preceding_value =
     detail::range_comparable_value<OrderByT>(preceding_window, orderby_column.type(), stream);
   auto following_value =
     detail::range_comparable_value<OrderByT>(following_window, orderby_column.type(), stream);
```
mythrocks, line 884: Argh. This needs cleaning up.

mythrocks: Resolved.

cpp/include/cudf/rolling/range_window_bounds.hpp
```diff
@@ -26,15 +26,27 @@ namespace cudf {
...
  */
 struct range_window_bounds {
  public:
+  /**
+   * The type of range_window_bounds.
+   */
+  enum class extent : int32_t {
+    CURRENT_ROW = 0,  /// Bounds defined as the first/last row that matches the current row.
+    BOUNDED,          /// Bounds defined as the first/last row that falls within
+                      /// a specified range from the current row
```
ttnghia, line 46:
```suggestion
                      /// a specified range from the current row.
```

cpp/include/cudf/rolling/range_window_bounds.hpp
```diff
@@ -26,15 +26,27 @@ namespace cudf {
...
  *      b) An `INT32` scalar, for use with an `INT32` orderby column
  *   2. "unbounded", indicating that the bounds stretch to the first/last
  *      row in the group.
+ *   3. "current row", indicating that the bounds end at the first/last
+ *      row in the group that match the value of the current row.
  */
 struct range_window_bounds {
  public:
+  /**
+   * The type of range_window_bounds.
```
ttnghia, line 41:
```suggestion
   * @brief The type of range_window_bounds.
```

cpp/include/cudf/rolling/range_window_bounds.hpp
```diff
@@ -26,15 +26,27 @@ namespace cudf {
...
+   */
+  enum class extent : int32_t {
```
ttnghia, line 43: In addition, I think the name `extent` is a bit misleading, since I was firstly confused when reading the code with it:
```suggestion
  enum class extent_type : int32_t {
```

cpp/include/cudf/rolling/range_window_bounds.hpp
mythrocks: I just noticed that yesterday. Thanks for *point*-ing that out.

mythrocks: Fixed.

cpp/tests/rolling/grouped_rolling_range_test.cpp
```diff
@@ -287,3 +288,321 @@ TYPED_TEST(GroupedRollingRangeOrderByDecimalTypedTest, UnboundedRanges)
...
+  auto const orderby =
+      strings_column{
+          "A", "A", "A", "B", "B", "B",
+          "C", "C", "C", "C",
+          "D", "D", "E", "E"
+      }.release();
+  // clang-format on
+
+  // Un-partitioned cases.
+  CUDF_TEST_EXPECT_COLUMNS_EQUIVALENT(
```
ttnghia, line 385: Nit: Please try to use the `_EQUAL` assert version if possible, so we can check the output exactly.

ttnghia (approved):

cpp/tests/rolling/grouped_rolling_range_test.cpp
mythrocks: Fixed.

mythrocks:
Thanks for the reviews, chaps. This is now merged.

PR #13144, opened 2023-04-16
Optimize JSON writer

karthikeyann:
## Description
Reduce benchmark runtime by skipping unrequired combinations
Optimize List and Struct conversion functions.
   - Creates array of string views and use `make_strings_column` to create the joined strings, and update offsets to create concatenated row string.

Update default rows_per_chunk in Cython.

512MB dataframe writing runtime reduced from 84s to 2.7s.
JSON_WRITER benchmark runtime reduced from ~15m to ~5m.

## Checklist


cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +182,235 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
...
+ */
+std::unique_ptr<column> join_list_of_strings(lists_column_view const& lists_strings,
+                                             string_view const list_prefix,
+                                             string_view const list_suffix,
+                                             string_view const element_separator,
+                                             string_view const element_narep,
+                                             //  string_view const list_narep,
+                                             rmm::cuda_stream_view stream,
+                                             rmm::mr::device_memory_resource* mr)
+{
```
karthikeyann, line 287: Algorithm used:
-  calculate #strviews per list using null mask, and list_offsets.
-  scan #strviews to get #strviews(scanned)
-  create label segments.
-  sublist_index= index - offsets[label]
-  #strviews(scanned)[label] + sublist_index = string_view index +1, +2
-  use above 2 to scatter element, element_seperator
-  scatter list_prefix, list_suffix to the right place using list_offsets
-  make_strings_column() and gather offsets, based on #strviews(scanned).

karthikeyann:
Nsight Profile:
Before:
image
After:
image

cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +183,208 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
...
+      }
+      // suffix
+      if (col == tbl.num_columns() - 1) {
+        d_strviews[this_index + 2] = str_pair(row_suffix.data(), row_suffix.size_bytes());
+      }
+    });
+  auto joined_col = make_strings_column(d_strviews, stream, mr);
+
+  // gather from offset and create a new string column
+  rmm::device_uvector<size_type> d_strview_offsets(strings_columns.num_rows() + 1, stream);
```
PointKernel, line 244: It seems `d_strview_offsets` can be created via `make_counting_transform_iterator` thus we can avoid materialize the data:
```cpp
auto const d_strview_offsets =
  make_counting_transform_iterator(0, [](auto i) { return i * num_str_views_per_row; });
```

cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +183,208 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
...
+ * @param stream CUDA stream used for device memory operations and kernel launches.
+ * @param mr Device memory resource used to allocate the returned column's device memory.
+ * @return New strings column with concatenated results.
+ */
+std::unique_ptr<column> join_list_of_strings(lists_column_view const& lists_strings,
+                                             string_view const list_prefix,
+                                             string_view const list_suffix,
+                                             string_view const element_separator,
+                                             string_view const element_narep,
+                                             //  string_view const list_narep,
```
PointKernel, line 285:
```suggestion
```

cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +183,208 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
...
+      } else {
+        // [ ]
+        d_strviews[d_strview_counts[idx]] = str_pair{list_prefix.data(), list_prefix.size_bytes()};
+        d_strviews[d_strview_counts[idx + 1] - 1] =
+          str_pair{list_suffix.data(), list_suffix.size_bytes()};
+      }
+    });
+
+  // scatter string and separator
+  auto labels = cudf::lists::detail::generate_labels(lists_strings, num_strings, stream, mr);
```
PointKernel, line 340:
```suggestion
  auto labels = cudf::lists::detail::generate_labels(lists_strings, num_strings, stream, rmm::mr::get_current_device_resource());
```
`labels` seems not to be returned by the function

cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +183,208 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
...
+    thrust::make_counting_iterator<size_type>(num_lists),
+    [col = *col_device_view,
+     list_prefix,
+     list_suffix,
+     d_strview_counts = d_strview_offsets.begin(),
+     d_strviews       = d_strviews.begin()] __device__(auto idx) {
+      if (col.is_null(idx)) {
+        d_strviews[d_strview_counts[idx]] = str_pair{nullptr, 0};
+        d_strviews[d_strview_counts[idx] + 1] =
+          str_pair{nullptr, 0};  // TODO could skip this to reduce memory usage.
```
PointKernel, line 330: Would this `TODO` be a separate PR?

cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +183,208 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
                            strings_columns.end(),
                            [](auto const& c) { return c.type().id() == type_id::STRING; }),
                "All columns must be of type string");
+  auto const num_str_views_per_row = strings_columns.num_columns() * 3 + 1;
```
PointKernel, line 186: Would be great if we can get rid of the magic number `3`

cpp/src/io/json/write_json.cu
```diff
@@ -589,29 +676,22 @@ std::unique_ptr<column> make_column_names_column(host_span<column_name_info cons
...
-  auto p_str_col_w_nl = cudf::strings::detail::join_strings(str_column_view,
-                                                            d_line_terminator,
-                                                            string_scalar("", false),
-                                                            stream,
-                                                            rmm::mr::get_current_device_resource());
-  strings_column_view strings_column{p_str_col_w_nl->view()};
-
-  auto total_num_bytes      = strings_column.chars_size();
-  char const* ptr_all_bytes = strings_column.chars_begin();
+  auto total_num_bytes      = str_column_view.chars_size() - skip_last_chars;
```
PointKernel, line 693:
```suggestion
  auto const total_num_bytes      = str_column_view.chars_size() - skip_last_chars;
```

cpp/src/io/json/write_json.cu
```diff
@@ -418,53 +517,40 @@ struct column_to_strings_fn {
...
+      lists_column_view(column), stream_, rmm::mr::get_current_device_resource());
+    auto const list_child_string = make_lists_column(
+      column.size(),
+      std::move(new_offsets),
+      std::move(child_string_with_null()),
+      column.null_count(),
+      cudf::detail::copy_bitmask(column, stream_, rmm::mr::get_current_device_resource()),
+      stream_);
+    auto l_pre = string_scalar{"["}, l_post = string_scalar{"]"}, sep = string_scalar{","},
+         elem_narep = string_scalar{"null"};
```
PointKernel, line 546: could be constexpr

cpp/src/io/json/write_json.cu
karthikeyann: I replaced `cudf::detail::gather` with `thrust::gather` after using iterator as suggested.

cpp/src/io/json/write_json.cu
karthikeyann: `string_scalar` has non-trivial constructor. It is non-literal type, so it can't be a constexpr.

cpp/src/io/json/write_json.cu
karthikeyann: Not a big change. it can be ignore for now.

cpp/src/io/json/write_json.cu
karthikeyann: updated.

cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +182,200 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
...
+  auto const num_str_views_per_row    = strings_columns.num_columns() * str_views_per_column + 1;
+  // Eg. { col1: value , col2: value , col3: value } = 1+ 3+3+(3-1) +1 = 10
+
+  auto tbl_device_view = cudf::table_device_view::create(strings_columns, stream);
+  auto d_column_names  = column_device_view::create(column_names, stream);
+
+  // (num_columns*3+1)*num_rows size (very high!) IDEA: chunk it here? but maximize parallelism?
+  auto const string_views_total = num_str_views_per_row * strings_columns.num_rows();
+  auto const total_rows         = strings_columns.num_rows() * strings_columns.num_columns();
+  using str_pair                = thrust::pair<const char*, size_type>;
```
davidwendt, line 195: Using `thrust::pair` but calling these string-views irks me a bit. Perhaps use string-pair or some other name for these?

cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +182,200 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
...
+  rmm::device_uvector<str_pair> d_strviews(string_views_total, stream);
+
+  // scatter row_prefix, row_suffix, column_name:, value, value_separator as string_views
+  thrust::for_each(
+    rmm::exec_policy(stream),
+    thrust::make_counting_iterator<size_type>(0),
+    thrust::make_counting_iterator<size_type>(total_rows),
+    [tbl       = *tbl_device_view,
```
davidwendt, line 203: This is a large lambda and capture group. Could it be made into a functor or do you feel it is better inline?

cpp/src/io/json/write_json.cu
davidwendt: Perhaps add this as a comment in the code?

karthikeyann: It was in the code earlier. I removed it. I will add it.

cpp/src/io/json/write_json.cu
```diff
@@ -25,33 +25,36 @@
 #include <cudf/column/column_device_view.cuh>
 #include <cudf/column/column_factories.hpp>
 #include <cudf/detail/copy.hpp>
+#include <cudf/detail/iterator.cuh>
```
davidwendt, line 28: Are all of these headers still needed? Offhand I see that `cudf/detail/valid_if.cuh` is not being used in this file as far as I can tell.

cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +182,200 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
...
+  // gather from offset and create a new string column
+  auto old_offsets = strings_column_view(joined_col->view()).offsets();
+  rmm::device_uvector<size_type> row_string_offsets(strings_columns.num_rows() + 1, stream, mr);
+  auto const d_strview_offsets = cudf::detail::make_counting_transform_iterator(
+    0, [num_str_views_per_row] __device__(size_type const i) { return i * num_str_views_per_row; });
+  thrust::gather(rmm::exec_policy(stream),
+                 d_strview_offsets,
+                 d_strview_offsets + row_string_offsets.size(),
+                 old_offsets.begin<size_type>(),
+                 row_string_offsets.begin());
```
davidwendt, line 255: This is a cool solution.

cpp/src/io/json/write_json.cu
davidwendt: There is a string-view version of the factory too if you wanted to use `string_view`s instead of `std::pair`. Was there something preventing you from using that one?
https://github.com/rapidsai/cudf/blob/267bc6a82dd7ddd241e2d722af08e2f5bf996e6f/cpp/include/cudf/column/column_factories.hpp#L407-L411

karthikeyann: It transforms to pairs. That's why I used pairs directly.

cpp/src/io/json/write_json.cu
karthikeyann: moved to lambda.

cpp/src/io/json/write_json.cu
karthikeyann: did a include cleanup once again.

cpp/src/io/json/write_json.cu
karthikeyann: used `string_view` itself. used `string_view` version of the factory too.

davidwendt (approved):

galipremsagar (approved):

cpp/src/io/json/write_json.cu
```diff
@@ -135,81 +134,51 @@ struct escape_strings_fn {
...
-  char* d_chars{};
+// Struct - scatter string_views of each element in a struct column
+struct struct_scatter_strings_fn {
+  table_device_view const tbl;
+  column_device_view const col_names;
+  size_type const strviews_per_column;
+  size_type const num_strviews_per_row;
+  string_view const row_prefix;       //{
+  string_view const row_suffix;       //} or }\n for json-lines
+  string_view const value_separator;  //,
```
bdice, line 145:
```suggestion
  string_view const row_prefix;       // {
  string_view const row_suffix;       // } or }\n for json-lines
  string_view const value_separator;  // ,
```

cpp/src/io/json/write_json.cu
```diff
@@ -254,38 +223,171 @@ std::unique_ptr<column> struct_to_strings(table_view const& strings_columns,
                            strings_columns.end(),
                            [](auto const& c) { return c.type().id() == type_id::STRING; }),
                "All columns must be of type string");
+  auto constexpr strviews_per_column = 3;  // (for each "column_name:", "value",  "separator")
+  auto const num_strviews_per_row    = strings_columns.num_columns() * strviews_per_column + 1;
+  // Eg. { col1: value , col2: value , col3: value } = 1+ 3+3+(3-1) +1 = 10
```
bdice, line 228:
```suggestion
  // e.g. {col1: value, col2: value, col3: value} = 1 + 3 + 3 + (3-1) + 1 = 10
```

bdice (approved):
Minor suggestions. Approving.

shwina (approved):

cpp/src/io/json/write_json.cu
karthikeyann: How is this not caught in clang-format?

davidwendt: Seems like it should. I noticed the new clang-format doing some odd things myself.

bdice: We noticed a bug in clang-format 16.0.1 relating to spacing around `*` (it made some multiplications look like pointers). We can revert to another version if you think there are issues with the version we've chosen.

davidwendt: I've not seen anything concerning that is worth reverting to an older version.

PR #13148, opened 2023-04-17
Fix gtests to always pass null-count to set_null_mask calls

davidwendt:
## Description
Fix calls to the `column::set_null_mask()` function to always pass the null-count value in gtests source files.

## Checklist

### DIFF
--- a/cpp/tests/reductions/rank_tests.cpp
+++ b/cpp/tests/reductions/rank_tests.cpp
@@ -215,7 +215,8 @@ TYPED_TEST(TypedRankScanTest, StructsWithNullPushdown)
   // First, verify that if the structs column has only nulls, all output rows are ranked 1.
   {
     // Null mask not pushed down to members.
-    struct_col->set_null_mask(create_null_mask(12, cudf::mask_state::ALL_NULL), struct_col->size());
+    struct_col->set_null_mask(create_null_mask(struct_col->size(), cudf::mask_state::ALL_NULL),
+                              struct_col->size());
     auto const expected_null_result = rank_result_col{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
     auto const expected_percent_rank_null_result =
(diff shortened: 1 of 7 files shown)


cpp/tests/reductions/rank_tests.cpp
```diff
@@ -214,8 +214,8 @@ TYPED_TEST(TypedRankScanTest, StructsWithNullPushdown)
 
   // First, verify that if the structs column has only nulls, all output rows are ranked 1.
   {
-    struct_col->set_null_mask(
-      create_null_mask(12, cudf::mask_state::ALL_NULL));  // Null mask not pushed down to members.
+    // Null mask not pushed down to members.
+    struct_col->set_null_mask(create_null_mask(12, cudf::mask_state::ALL_NULL), struct_col->size());
```
bdice, line 218: It seems fragile to have one length hardcoded as 12 and another written as a column size call. Should we be preferring the size call? Relatedly, can we do anything to not require/not provide this argument when the mask state is all nulls or no nulls?

davidwendt: I don't see how we can make this easier other than providing some additional member functions. I don't think adding these is worth doing for this especially for the very rare all-nulls case.

cpp/tests/transform/nans_to_null_test.cpp
```diff
@@ -27,11 +27,11 @@ template <typename T>
 struct NaNsToNullTest : public cudf::test::BaseFixture {
   void run_test(cudf::column_view const& input, cudf::column_view const& expected)
   {
-    auto got_mask = cudf::nans_to_nulls(input);
+    auto [null_mask, null_count] = cudf::nans_to_nulls(input);
     cudf::column got(input);
-    got.set_null_mask(std::move(*(got_mask.first)));
+    got.set_null_mask(std::move(*(null_mask)), null_count);
```
ttnghia, line 32:
```suggestion
    got.set_null_mask(std::move(*null_mask), null_count);
```

ttnghia (approved):

vyasr (approved):

PR #13152, opened 2023-04-17
[REVIEW] Add `no_default` and adapt `Series.reset_index` to differentiate `None` for `name` parameter

galipremsagar:
## Description

In `pandas-2.0` the behavior for `name` parameter has changed to actually name a column `0` if no value is passed to `name`. But if `name=None`, the column will be named `None` too:

```python
In [1]: import pandas as pd

In [2]: s = pd.Series([10, 11, 23], index=[2, 3, 5])

In [3]: s
Out[3]: 
2    10
3    11
5    23
dtype: int64

In [4]: s.reset_index()
Out[4]: 
   index   0
0      2  10
1      3  11
2      5  23

In [5]: s.reset_index(name=None)
Out[5]: 
   index  None
0      2    10
1      3    11
2      5    23
```

To achieve the same behavior in `cudf`, we had to introduce `no_default` value(which is same as pandas's `no_default` value). 

This also fixes 18 pytests:
```
= 965 failed, 86325 passed, 2044 skipped, 992 xfailed, 165 xpassed in 508.32s (0:08:28) =
```

```
= 983 failed, 86277 passed, 2034 skipped, 992 xfailed, 165 xpassed in 541.87s (0:09:01) =
```
## Checklist

### DIFF
--- a/python/cudf/cudf/__init__.py
+++ b/python/cudf/cudf/__init__.py
@@ -53,7 +53,6 @@ from cudf.core.index import (
     UInt64Index,
     interval_range,
 )
-from cudf.core.default import no_default
 from cudf.core.missing import NA
 from cudf.core.multiindex import MultiIndex
 from cudf.core.reshape import (
--- a/python/cudf/cudf/__init__.py
+++ b/python/cudf/cudf/__init__.py
@@ -53,6 +53,7 @@ from cudf.core.index import (
     UInt64Index,
     interval_range,
 )
+from cudf.core.default import no_default
 from cudf.core.missing import NA
 from cudf.core.multiindex import MultiIndex
 from cudf.core.reshape import (
--- a/python/cudf/cudf/api/extensions/__init__.py
+++ b/python/cudf/cudf/api/extensions/__init__.py
@@ -1,13 +1,15 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
 
 from cudf.api.extensions.accessor import (
     register_dataframe_accessor,
     register_index_accessor,
     register_series_accessor,
 )
+from pandas.api.extensions import no_default
 
 __all__ = [
     "register_dataframe_accessor",
     "register_index_accessor",
     "register_series_accessor",
+    "no_default",
 ]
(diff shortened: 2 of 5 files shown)


python/cudf/cudf/__init__.py
```diff
@@ -53,6 +53,7 @@
     UInt64Index,
     interval_range,
 )
+from cudf.core.default import no_default
```
bdice, line 56: Why are we exposing this in the top level API of cudf, rather than mirroring pandas in `cudf.api.extensions.no_default`?

bdice (changes requested):
Let's match the pandas API layout.

python/cudf/cudf/__init__.py
galipremsagar: Moved the `no_default` into `api.extensions` and removed it from top level API of `cudf` to closely mirror pandas now.

galipremsagar:
> Can we add tests for the cases newly supported by this PR (`None` names)?

We do here: https://github.com/rapidsai/cudf/pull/13152/files#diff-5d087b898461a7b9dbabf8cde1c72a8ded397b75fc9402b7108524f24de67adeR1411

bdice:
> > Can we add tests for the cases newly supported by this PR (`None` names)?
> 
> We do here: https://github.com/rapidsai/cudf/pull/13152/files#diff-5d087b898461a7b9dbabf8cde1c72a8ded397b75fc9402b7108524f24de67adeR1411

Sorry, I misread the filename and thought that was an implementation and not a test. My mistake. Approving now!

bdice (approved):

python/cudf/cudf/api/extensions/__init__.py
```diff
@@ -1,13 +1,15 @@
...
     register_index_accessor,
     register_series_accessor,
 )
+from pandas.api.extensions import no_default
 
 __all__ = [
     "register_dataframe_accessor",
     "register_index_accessor",
     "register_series_accessor",
+    "no_default",
```
bdice, line 14: Small request: Alphabetize this list?

galipremsagar: Done 👍 Sorry for the random order.

mroeschke (approved):

PR #13153, opened 2023-04-17
Allow compilation with any GTest version 1.10+

robertmaynard:
## Description
GTest max support for `Types` was removed in 1.11, so we remove the workarounds in cudf_gtest.

Since we need to support our custom `Types` and the GTest 1.11+ version rework the type_list_utilities to be generic and not depend on specific traits.

Also corrected the `<<` overloads for GTest printing so that they work with GTest 1.11.

## Checklist


robertmaynard:
> Question: Do we need the ostream code for debugging? Seems like a candidate for removal.

The ostream code is needed so that we can use the `cudf::meanvar` type with GTest macro's like `EXPECT_EQ` and `EXPECT_NEAR`.

ttnghia (approved):

PR #13163, opened 2023-04-18
Add support for DatetimeTZDtype and tz_localize

shwina:
## Description

TBD

Quick benchmark:

```python
psr = pd.Series(pd.date_range("1970-01-01", "1980-01-01", freq="1T"))
sr = cudf.from_pandas(psr)

In [6]: %timeit psr.dt.tz_localize("America/New_York", ambiguous="NaT", nonexistent="NaT")
1.55 s ± 9.91 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [7]: %timeit sr.dt.tz_localize("America/New_York", ambiguous="NaT", nonexistent="NaT")
9.98 ms ± 246 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

## Checklist


python/cudf/cudf/core/index.py
```diff
@@ -2364,7 +2364,12 @@ def isocalendar(self):
 
     @_cudf_nvtx_annotate
     def to_pandas(self, nullable=False):
-        nanos = self._values.astype("datetime64[ns]")
+        if isinstance(self.dtype, pd.DatetimeTZDtype):
+            nanos = self._values.astype(
+                pd.DatetimeTZDtype("ns", self.dtype.tz)
```
mroeschke, line 2369: As a pandas 2.0 note, would be good to keep "s", "ms", or "us" resolution here

python/cudf/cudf/core/column/datetime.py
```diff
@@ -517,6 +524,72 @@ def can_cast_safely(self, to_dtype: Dtype) -> bool:
...
+
+    def to_pandas(
+        self, index: pd.Index = None, nullable: bool = False, **kwargs
+    ) -> "cudf.Series":
+        return self._local_time.to_pandas().dt.tz_localize(
+            self.dtype.tz, ambiguous="NaT", nonexistent="NaT"
+        )
+
+    def to_arrow(self):
+        return self._local_time.to_arrow().cast(
```
mroeschke, line 568: I think you'll need to use `pa.assume_timezone` here. IIRC `cast` is equivalent to "tz_convert"  in pandas: https://arrow.apache.org/docs/python/generated/pyarrow.compute.assume_timezone.html

python/cudf/cudf/core/_internals/timezones.py
```diff
@@ -67,5 +78,135 @@ def _find_and_read_tzfile_tzdata(zone_name):
...
+        left_inclusive=True,
+        right_edges=nonexistent_end,
+        right_inclusive=False,
+    ).notnull()
+
+    return ambiguous, nonexistent
+
+
+def localize(
+    data: DatetimeColumn, zone_name: str, ambiguous, nonexistent
```
mroeschke, line 160: Are you planning on supporting `tz_localize(None)` in this PR? In pandas, it removes the timezone info for tz-aware data returning local timestamps

python/cudf/cudf/core/index.py
galipremsagar: Since this is currently targeting `branch-23.06` which doesn't have pandas-2.0 support this change seems okay, what @mroeschke suggesting is taken care of in `pandas_2.0_feature_branch`, so perhaps a note or todo here would be useful while I sync these changes with pandas-2.0 support.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -313,9 +313,13 @@ cdef class Column:
...
         if is_categorical_dtype(self.dtype):
             col = self.base_children[0]
+            data_dtype = col.dtype
+        elif is_datetime64tz_dtype(self.dtype):
+            col = self
+            data_dtype = _get_base_dtype(col.dtype)
         else:
             col = self
-        data_dtype = col.dtype
+            data_dtype = self.dtype
```
bdice, line 322: Let's use `col` or `self` consistently in the second line. Favoring `col` here:
```suggestion
            data_dtype = col.dtype
```

python/cudf/cudf/core/_internals/timezones.py
```diff
@@ -67,5 +78,135 @@ def _find_and_read_tzfile_tzdata(zone_name):
...
+    return DataFrame._from_columns(
+        transition_times_and_offsets, ["transition_times", "offsets"]
+    )
+
+
+def _find_ambiguous_and_nonexistent(
+    data: DatetimeColumn, zone_name: str
+) -> Tuple:
+    """
+    Recognize ambiguous and nonexistent timestamps for the given timezone.
```
bdice, line 101: Please document the parameters and two return values. I had to scroll to see `return ambiguous, nonexistent` before I understood what this was meant to do.

python/cudf/cudf/core/column/datetime.py
```diff
@@ -517,6 +524,72 @@ def can_cast_safely(self, to_dtype: Dtype) -> bool:
...
+            pa.timestamp(self.dtype.unit, str(self.dtype.tz))
+        )
+
+    @property
+    def values(self):
+        """
+        Return a CuPy representation of the DateTimeColumn.
+        """
+        raise NotImplementedError(
+            "DateTime Arrays is not yet implemented in cudf"
```
bdice, line 578:
```suggestion
            "DateTime Arrays are not yet implemented in cudf"
```

python/cudf/cudf/tests/series/test_datetimelike.py
```diff
@@ -1 +1,63 @@
...
+from cudf.testing._utils import assert_eq
+
+
+@pytest.fixture(params=["ns", "us", "ms", "s"])
+def unit(request):
+    return request.param
+
+
+@pytest.fixture(
+    params=["America/New_York", "Asia/Tokyo", "CET", "Etc/GMT+1", "UTC"]
```
bdice, line 17: I'd like to see if we can test _every_ time zone present on the system against pandas behavior for some function(s).

python/cudf/cudf/core/_internals/timezones.py
shwina: Ah good catch! I think that's very useful and I'll include it.

python/cudf/cudf/core/index.py
```diff
@@ -2490,6 +2495,55 @@ def round(self, freq):
...
+                      dtype='datetime64[ns, America/New_York]')
+
+        Ambiguous or nonexistent datetimes are converted to NaT.
+
+        >>> s = cudf.to_datetime(cudf.Series(['2018-10-28 01:20:00',
+        ...                                   '2018-10-28 02:36:00',
+        ...                                   '2018-10-28 03:46:00']))
+        >>> s.dt.tz_localize("CET")
+        0    2018-10-28 01:20:00.000000000
+        1                             <NA>
```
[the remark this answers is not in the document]
shwina: We use `<NA>` to indicate `NaT`. Note that unlike `NaN`, `NaT` is just a "regular" null value:


```python
In [5]: cudf.Series([1, 2, None, 3], dtype="datetime64[ns]")
Out[5]: 
0    1970-01-01 00:00:00.000000001
1    1970-01-01 00:00:00.000000002
2                             <NA>
3    1970-01-01 00:00:00.000000003
dtype: datetime64[ns]

```

python/cudf/cudf/core/column/datetime.py
shwina: Done -- thanks for pointing this out!

python/cudf/cudf/core/_internals/timezones.py
shwina: Done -- `.tz_localize(None)` now converts timezone-aware timestamps to timezone-naive ones.

python/cudf/cudf/core/_internals/timezones.py
shwina: Fixed

mroeschke (approved):
Nice! My comments have been addressed

shwina:
This PR is blocked on dropping Python 3.8 and adding Python 3.9 support in cuDF (the `zoneinfo` module was introduced in Python 3.9)

bdice:
Waiting for #13196.

python/cudf/cudf/core/_internals/timezones.py
```diff
@@ -67,5 +78,143 @@ def _find_and_read_tzfile_tzdata(zone_name):
...
+
+    Returns a tuple of columns, both of "bool" dtype and of the same
+    size as `data`, that respectivelt indicate ambiguous and
```
bdice, line 104:
```suggestion
    size as `data`, that respectively indicate ambiguous and
```

python/cudf/cudf/core/_internals/timezones.py
```diff
@@ -67,5 +78,143 @@ def _find_and_read_tzfile_tzdata(zone_name):
...
+    If no transitions occur, the tuple `(None, None)` is returned.
+    """
+    tz_data_for_zone = get_tz_data(zone_name)
+    transition_times = tz_data_for_zone["transition_times"]
+    offsets = tz_data_for_zone["offsets"].astype(
+        f"timedelta64[{data._time_unit}]"
+    )
+
+    if len(offsets) == 1:  # no transitions
+        return False, False
```
bdice, line 118: The docs say that "no transitions" returns `(None, None)`. How do we reconcile this?

python/cudf/cudf/core/column/datetime.py
```diff
@@ -517,6 +524,72 @@ def can_cast_safely(self, to_dtype: Dtype) -> bool:
...
+            self.dtype.tz, ambiguous="NaT", nonexistent="NaT"
+        )
+
+    def to_arrow(self):
+        return pa.compute.assume_timezone(
+            self._local_time.to_arrow(), str(self.dtype.tz)
+        )
+
+    @property
+    def values(self):
```
bdice, line 573: Should this property error implementation be inherited from `DateTimeColumn`? Or if you think it's possible that CuPy might support time-zone-naive data (the parent class but not this zone-aware class), should we mention "time zone" in the error message?

python/cudf/cudf/core/index.py
```diff
@@ -2490,6 +2495,57 @@ def round(self, freq):
 
         return self.__class__._from_data({self.name: out_column})
 
+    def tz_localize(self, tz, ambiguous="NaT", nonexistent="NaT"):
+        """
+        Localize timezone-naive data to timezone-aware data.
+
+        Parameters
+        ----------
+        tz: str
```
bdice, line 2504: I think there's usually supposed to be a space on both sides of the colon here.

> The colon must be preceded by a space, or omitted if the type is absent.

https://numpydoc.readthedocs.io/en/latest/format.html#parameters

```suggestion
        tz : str
```

python/cudf/cudf/core/series.py
```diff
@@ -4602,6 +4602,21 @@ def strftime(self, date_format, *args, **kwargs):
             data=str_col, index=self.series._index, name=self.series.name
         )
 
+    @copy_docstring(DatetimeIndex.tz_localize)
```
bdice, line 4605: This might be weird, but I would expect to copy the docstring from Series to Index, rather than Index to Series. I view the Series as the more fundamental object, but I don't know what precedents exist here in either class.

python/cudf/cudf/tests/series/test_datetimelike.py
```diff
@@ -1 +1,113 @@
...
+import pytest
+
+import cudf
+from cudf import date_range
+from cudf.testing._utils import assert_eq
+
+
+def _get_all_zones():
+    zones = []
+    for root, dirs, files in os.walk("/usr/share/zoneinfo"):
```
bdice, line 15: Just FYI, we ran into issues with `/etc/localtime` not existing in https://github.com/rapidsai/mambaforge-cuda/pull/16. The relation here is that I'm not 100% sure if you can assume `/usr/share/zoneinfo` is populated with data in super-minimalist Docker containers. What this means is that the result of this function could be an empty list. If you want that to raise an error/warning, you might want to add something explicit for that. I'm content with the current behavior or a warning. I might lean away from raising an error, but I'm not sure.

python/cudf/cudf/tests/series/test_datetimelike.py
bdice: Yes, this is a good solution. I think testing all timezones for a very limited subset of functions is a helpful way to make sure we're not making bad assumptions about the timezone database structure which may change over time or by system.

python/cudf/cudf/tests/series/test_datetimelike.py
```diff
@@ -1 +1,113 @@
...
+    got = s.dt.tz_localize(zone_name)
+    assert_eq(expect, got)
+
+
+@pytest.mark.parametrize("zone_name", ALL_TIME_ZONES)
+def test_localize_nonexistent(request, unit, zone_name):
+    request.applymarker(
+        pytest.mark.xfail(
+            condition=(zone_name == "America/Grand_Turk"),
+            reason="https://www.worldtimezone.com/dst_news/dst_news_turkscaicos03.html",  # noqa: E501
```
bdice, line 84: Were we able to identify if there is an inconsistency somewhere for Grand Turk and Metlakatla? Perhaps disagreements between system-provided timezone information and some timezone database in a Python package?

bdice: This also might be a candidate for skipping instead of xfailing, if we run tests with `xfail_strict`. This sounds like something where we could get an unexpected success (which would cause an error) due to Python/system software updates ...particularly if my hypothesis above is true.

bdice (commented):
Mostly minor comments. I'll approve once you've taken a look and addressed some of them. This is great work -- nice job @shwina.

python/cudf/cudf/tests/series/test_datetimelike.py
shwina: I honestly haven't gotten to the bottom of it. If we ever _do_ hit an unexpected success here, e.g., if something changes in either `zoneinfo` or `pytz`, I would love to be alerted of that! So, should we keep the `xfail`?

python/cudf/cudf/core/_internals/timezones.py
shwina: I changed the docs

shwina: (I wish I was the kind of programmer whose docs were more up-to-date than their code was)

python/cudf/cudf/core/column/datetime.py
shwina: Nope -- you're right, we should just inherit this

python/cudf/cudf/core/index.py
shwina: Fix'd

python/cudf/cudf/core/column/__init__.py
```diff
@@ -23,6 +23,7 @@
     serialize_columns,
 )
 from cudf.core.column.datetime import DatetimeColumn  # noqa: F401
+from cudf.core.column.datetime import DatetimeTZColumn  # noqa: F401
```
bdice, line 26: FYI: you can fix these "unused import" errors by defining `__all__`. If it's in `__all__`, it's "part of the API" of this module and thus flake8 sees it as used.

python/cudf/cudf/core/column/datetime.py
```diff
@@ -517,6 +524,63 @@ def can_cast_safely(self, to_dtype: Dtype) -> bool:
         else:
             return False
 
+    def _with_type_metadata(self, dtype):
```
galipremsagar, line 527: This might need an update to places here to include `DatetimeTZColumn`:

https://github.com/rapidsai/cudf/blob/branch-23.06/python/cudf/cudf/core/series.py#L1545-L1552
https://github.com/rapidsai/cudf/blob/branch-23.06/python/cudf/cudf/core/dataframe.py#L3801-L3815
https://github.com/rapidsai/cudf/blob/branch-23.06/python/cudf/cudf/core/dataframe.py#L1702-L1713

python/cudf/cudf/core/column/column.py
```diff
@@ -1554,6 +1554,17 @@ def build_column(
             offset=offset,
             null_count=null_count,
         )
+    elif is_datetime64tz_dtype(dtype):
+        if data is None:
+            raise TypeError("Must specify data buffer")
```
bdice, line 1559: `ValueError` seems more natural here unless you're matching something in pandas.
```suggestion
            raise ValueError("Must specify data buffer")
```

python/cudf/cudf/core/series.py
bdice: Sounds good. I agree with you!

python/cudf/cudf/tests/series/test_datetimelike.py
bdice: No strong feelings. Let's leave it as-is and not overengineer it just to get a skip message.

python/cudf/cudf/tests/series/test_datetimelike.py
bdice: Sounds good to me!

bdice (approved):
Great work. I had a couple small comments/replies but nothing blocking. Thanks.

python/cudf/cudf/core/column/datetime.py
galipremsagar: While we are at it, can we see if can totally remove the special casing in some those places and call `_with_type_metadata` always?

python/cudf/cudf/core/column/__init__.py
shwina: Isn't this more `explicit` though? I always thought the only "official" use for `__all__` was to define what happens when you do `from module import *`

galipremsagar (approved):

python/cudf/cudf/core/column/datetime.py
shwina: Done (and done!)

python/cudf/cudf/core/column/__init__.py
bdice: I think that defining your module’s public API (rather than not defining it) is the most `explicit` approach. I’m pro-`__all__` in Python libraries.

shwina: >  think that defining your module’s public API (rather than not defining it) is the most `explicit` approach. 

Hmm, but `__all__` doesn't do that AFAICT. Placing a name in `__all__` doesn't preclude it from appearing in e.g., tab completion options (naming it with a leading `_` does). 

Are there other tools than `flake8` that do something with `__all__`?

mroeschke: I think type checkers (mypy, pyright) will utilize `__all__` to see what APIs are "public"

bdice: Also, to your question about tab completion in IPython/etc., that's discussed here: https://github.com/ipython/ipykernel/issues/129#issuecomment-216870597

The community seems split on the issue, but some folks seem to think limiting to `__all__` by default would be good. Either way -- it seems that `__all__` is viewed as a source of truth for public APIs, and the real question is whether autocompletion should only show public APIs.

> It's configurable, with the option IPCompleter.limit_to__all__ ([config options docs](http://ipython.readthedocs.io/en/stable/config/options/terminal.html)).
>
> The reason we do this is that we often want to use IPython to poke about and debug modules, so we like to see what's really there, not just what's whitelisted as public API. We should, however, do a better job of prioritising completions, so that things in `__all__` come before things that aren't in it.

shwina: Thanks, both! I'm on board with defining `__all__`

shwina: I wonder if we can get a linter to enforce that `__all__` is defined (even if empty) always.

PR #13171, opened 2023-04-19
Enable mixed-dtype decimal/scalar binary operations

### LINKED ISSUE #13170: [BUG] ValueError when multiplying decimal columns of large precision with scalars
#13034 introduced a regression when attempting to multiply decimal columns and scalars:

```python
from decimal import Decimal

import cudf

cudf.Series([1, 2, 3, 4], dtype=cudf.Decimal128Dtype(38, 0)) * Decimal('100.0')
# ValueError: scale=-39 exceeds precision=38
```

The problematic change is this line:

https://github.com/rapidsai/cudf/commit/0b59fdad43fa5fd35287748a57d800e8db37c416#diff-ab0b6c0cb47275bf318fb8392675fe9f1e17ec9b0b31a5a2d8c9440d503d0c89R160

We shouldn't try to cast the scalar to the same data type as the input in every situation.

shwina:
Also closes #13170


python/cudf/cudf/core/column/decimal.py
```diff
@@ -157,7 +157,13 @@ def normalize_binop_value(self, other):
                 other = other.astype(self.dtype)
             return other
         elif is_scalar(other) and isinstance(other, (int, Decimal)):
-            return cudf.Scalar(Decimal(other), dtype=self.dtype)
+            other = Decimal(other)
+            metadata = other.as_tuple()
+            precision = max(len(metadata.digits), metadata.exponent)
+            scale = -metadata.exponent
+            return cudf.Scalar(
```
wence-, line 164: If `other` is `Decimal` should this not somehow end up going through the same normalisation as above with `cudf.Scalar` with decimal dtypes? Or does it not matter?

I'm pretty shaky on all the decimal stuff though, so ...

wence- (commented):
One query

wence-:
Merged trunk which will hopefully fix the wheelbuild fails.

PR #13172, opened 2023-04-19
Introduce `ManagedUDFString` on the python side

brandon-b-miller:
Depends on https://github.com/rapidsai/cudf/pull/13125

This PR introduces `ManagedUDFString` which is modeled as a struct containing an unmanaged `udf_string` and a pointer to an `NRT_MemInfo` object tracking that `udf_string`. 

It also introduces a function to be called when functions return new strings that copies the `udf_string` into heap allocated memory next to the tracking `NRT_MemInfo`. This allows the string to be free'd later reliably without needing to figure out where the stack copy lives at free time. Note that numba has no mechanism for freeing the memory _other_ than via the dtor attached to the `NRT_MemInfo`, so although there are two copies of the same `udf_string` pointing to the same string data, there is no way of causing a double free.


ajschmidt8 (approved):

PR #13057, opened 2023-04-04
Improve performance of slice_strings for long strings

davidwendt:
## Description
Improves on performance for longer strings with `cudf::strings::slice_strings()` API. The `cudf::string_view::substr` was reworked to minimize counting characters and the gather version of `make_strings_children` is used to build the resulting strings column. This version is already optimized for small and large strings.

Additionally, the code was refactored so the common case of `step==1 and start < stop`  can also make use of the gather approach. Common code was also grouped closer together to help navigate the source file better.

The `slice.cpp` benchmark was updated to better measure large strings with comparable slice boundaries. The benchmark showed performance improvement was up to 9x for larger strings with no significant degradation for smaller strings.

Reference #13048 and #12445

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13178, opened 2023-04-20
Fix slice_strings to return empty strings for stop < start indices

### LINKED ISSUE #13173: [BUG] slice_strings changed behavior when stop < start
**Describe the bug**
After #13057 the behavior of cudf::slice_strings has unexpectedly changed.  Previously when the stop column was less than the start column, an empty string was returned.  Now it appears the behavior is to return the substring from start to the end of the string.

The following program used to return:
```
sliced rows: 1
sliced nulls: 0
sliced result: 
```
and now returns:
```
sliced rows: 1
sliced nulls: 0
sliced result: c
```

```c++
#include <iostream>

#include <cudf/copying.hpp>
#include <cudf/scalar/scalar_factories.hpp>
#include <cudf/strings/slice.hpp>

  auto str_col = cudf::make_column_from_scalar(*cudf::make_string_scalar("abc"), 1);
  auto start_col = cudf::make_column_from_scalar(*cudf::make_fixed_width_scalar(2), 1);
  auto end_col = cudf::make_column_from_scalar(*cudf::make_fixed_width_scalar(0), 1);
  auto sliced = cudf::strings::slice_strings(cudf::strings_column_view(*str_col), *start_col, *end_col);
  std::cout << "sliced rows: " << sliced->size() << std::endl;
  std::cout << "sliced nulls: " << sliced->null_count() << std::endl;
  auto str_scalar = cudf::get_element(*sliced, 0);
  std::cout << "sliced result: " << static_cast<cudf::string_scalar*>(str_scalar.get())->to_string() << std::endl;
  return 0;
}
```
**Expected behavior**
Non-null strings sliced with stop < start should be empty strings.

davidwendt:
## Description

Fixes bug where `stop` value is less than `start` value in calls to `cudf::strings::slice_strings` should result in an empty string. Optimization in #13057 introduced this bug. Additional gtest was added to check for this condition.

Close #13173 

## Checklist


jlowe (approved):
Verified this fixes the substring integration test failure reported in NVIDIA/spark-rapids#8147

cpp/src/strings/slice.cu
```diff
@@ -59,7 +59,7 @@ struct substring_from_fn {
 
     auto const stop = stops[idx];
     auto const end  = (((stop < 0) || (stop > length)) ? length : stop);
-    return d_str.substr(start, end - start);
+    return start < end ? d_str.substr(start, end - start) : string_view{};
```
ttnghia, line 62: Should this be moved to the beginning of the function?

davidwendt: I don't see how. The `end` value is computed just above this line. This basically an edge case where the values provided are legal within the string but happen to be invalid compared to each other.

ttnghia: > The end value is computed just above this line.

Oh I see. I was trying to avoid calling `element<string_view>(idx)` before this but it seems that it is unavoidable.

ttnghia (approved):

mythrocks (approved):
LGTM!

PR #13181, opened 2023-04-20
Promote string UDF functions to return `ManagedUDFString` objects

brandon-b-miller:
Depends on https://github.com/rapidsai/cudf/pull/13172


ajschmidt8 (approved):

PR #13183, opened 2023-04-20
Use custom nvbench entry point to ensure `cudf::nvbench_base_fixture` usage

robertmaynard:
## Description
Previously we used the `nvbench::main` CMake target which would cause the same `main.o` to re-used across targets.
Since this object file was compiled without any knowledge of rmm or libcudf it wouldn't have our custom `NVBENCH_ENVIRONMENT` type. This caused our benchmarks to not execute with the rmm pool allocator.

## Checklist


davidwendt (approved):
Thanks for fixing this.

ttnghia (approved):

cpp/benchmarks/fixture/nvbench_main.cu
```diff
@@ -0,0 +1,22 @@
+/*
```
karthikeyann, line 1: Just curious, will this work, if this `nvbench_main.cu` is renamed to `nvbench_main.cpp`?

(Since compiling with g++ is faster than nvcc).

karthikeyann (approved):

cpp/benchmarks/fixture/nvbench_main.cu
robertmaynard: Yes, that works. Updated.
Thanks.

PR #13186, opened 2023-04-20
Refactor contiguous_split API into contiguous_split.hpp

abellina:
This PR moves contiguous_split specific APIs (pack/unpack/metadata/contiguous_split) out of `copying.hpp` and creates a new header `contiguous_split.hpp`.

I have built the cpp side, created doxygen docs, and built the python side and ran `pack_test`, but I could use more eyes there. 

I've marked this as breaking because APIs are moving from `copying.hpp` to `contiguous_split.hpp`

@nvdbaranec fyi


abellina:
Java tests passed, so I took this out of draft.

jlowe (approved):
Java approval

ttnghia:
Python builds fail because of unrelated (already known) reason.

java/src/main/native/src/ColumnVectorJni.cpp
```diff
@@ -19,6 +19,7 @@
 #include <arrow/api.h>
 #include <cudf/column/column_factories.hpp>
 #include <cudf/concatenate.hpp>
+#include <cudf/copying.hpp>
```
ttnghia, line 22: Why?

jlowe: Because ColumnViewJni.cpp is using methods from cudf/copying.hpp (e.g.: cudf::copy_if_else) and should have included it all along. It was implicitly relying on the inclusion of cudf/copying.hpp from cudf_jni_apis.hpp (and now that implicit include is gone).

ajschmidt8 (approved):

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ * @addtogroup column_copy
+ * @{
+ * @file
+ * @brief Table APIs for contiguous_split, pack, unpack, and metadadata
+ */
+
+/**
+ * @brief Column data in a serialized format
+ *
+ * @ingroup copy_split
```
ttnghia, line 37: I don't think we typically use such tag in cudf code.
```suggestion
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * Contains data from an array of columns in two contiguous buffers: one on host, which contains
+ * table metadata and one on device which contains the table data.
+ */
+struct packed_columns {
+  /**
+   * @brief Host-side metadata buffer used for reconstructing columns via unpack.
+   *
+   * @ingroup copy_split
```
ttnghia, line 46:
```suggestion
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+   */
+  struct metadata {
```
ttnghia, line 48: I see no reason to have this struct. It just wraps around a `std::vector<uint8_t>` so can we just use `std::vector<uint8_t>` directly?

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+   *
+   * @param md Host-side metadata buffer
+   * @param gd Device-side data buffer
+   */
+  packed_columns(std::unique_ptr<metadata>&& md, std::unique_ptr<rmm::device_buffer>&& gd)
+    : metadata_(std::move(md)), gpu_data(std::move(gd))
+  {
+  }
+
+  std::unique_ptr<metadata> metadata_;           ///< Host-side metadata buffer
```
ttnghia, line 92:
```suggestion
  std::vector<uint8_t> metadata_;           ///< Host-side metadata buffer
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+  std::unique_ptr<rmm::device_buffer> gpu_data;  ///< Device-side data buffer
```
ttnghia, line 93: We also don't need to wrap the device buffer in a smart pointer.
```suggestion
  rmm::device_buffer gpu_data;  ///< Device-side data buffer
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+};
+
+/**
+ * @brief The result(s) of a cudf::contiguous_split
+ *
+ * @ingroup copy_split
```
ttnghia, line 99:
```suggestion
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * Each table_view resulting from a split operation performed by contiguous_split,
+ * will be returned wrapped in a `packed_table`.  The table_view and internal
```
ttnghia, line 102:
```suggestion
 * will be returned wrapped in a `packed_table`. The table_view and internal
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ * column_views in this struct are not owned by a top level cudf::table or cudf::column.
+ * The backing memory and metadata is instead owned by the `data` field and is in one
+ * contiguous block.
+ *
+ * The user is responsible for assuring that the `table` or any derived table_views do
+ * not outlive the memory owned by `data`
```
ttnghia, line 108:
```suggestion
 * not outlive the memory owned by `data`.
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ */
+struct packed_table {
+  cudf::table_view table;  ///< Result table_view of a cudf::contiguous_split
+  packed_columns data;     ///< Column data owned
+};
+
+/**
+ * @brief Performs a deep-copy split of a `table_view` into a set of `table_view`s into a single
```
ttnghia, line 116:
```suggestion
 * @brief Performs a deep-copy split of a `table_view` into a single
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ * contiguous block of memory.
+ *
+ * @ingroup copy_split
```
ttnghia, line 119:
```suggestion
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ * in the `packed_table`. There is no top-level owning table.
+ *
+ * The returned views of `input` are constructed from a vector of indices, that indicate
+ * where each split should occur. The `i`th returned `table_view` is sliced as
+ * `[0, splits[i])` if `i`=0, else `[splits[i], input.size())` if `i` is the last view and
+ * `[splits[i-1], splits[i]]` otherwise.
+ *
+ * For all `i` it is expected `splits[i] <= splits[i+1] <= input.size()`
+ * For a `splits` size N, there will always be N+1 splits in the output
+ *
```
ttnghia, line 131:
```suggestion
 * For all `i` it is expected `splits[i] <= splits[i+1] <= input.size()`.
 * For a `splits` size N, there will always be N+1 splits in the output.
 *
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ * @endcode
+ *
+ *
+ * @throws cudf::logic_error if `splits` has end index > size of `input`.
+ * @throws cudf::logic_error When the value in `splits` is not in the range [0, input.size()).
+ * @throws cudf::logic_error When the values in the `splits` are 'strictly decreasing'.
+ *
+ * @param input View of a table to split
+ * @param splits A vector of indices where the view will be split
+ * @param[in] mr Device memory resource used to allocate the returned result's device memory
```
ttnghia, line 152:
```suggestion
 * @param mr Device memory resource used to allocate the returned result's device memory
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ * @return The set of requested views of `input` indicated by the `splits` and the viewed memory
+ * buffer.
```
ttnghia, line 154: Don't use period `.` inconsistently.
```suggestion
 * buffer
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ */
+std::vector<packed_table> contiguous_split(
+  cudf::table_view const& input,
+  std::vector<size_type> const& splits,
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Deep-copy a `table_view` into a serialized contiguous memory format
```
ttnghia, line 162:
```suggestion
 * @brief Deep-copy a `table_view` into a serialized contiguous memory format.
```

cpp/include/cudf/contiguous_split.hpp
ttnghia: Also, recommend to use `std::vector<std::byte>` instead.

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ *         and device memory respectively
+ */
+packed_columns pack(cudf::table_view const& input,
+                    rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Produce the metadata used for packing a table stored in a contiguous buffer.
+ *
+ * The metadata from the `table_view` is copied into a host vector of bytes which can be used to
+ * construct a `packed_columns` or `packed_table` structure.  The caller is responsible for
```
ttnghia, line 180:
```suggestion
 * construct a `packed_columns` or `packed_table` structure. The caller is responsible for
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * @brief Deep-copy a `table_view` into a serialized contiguous memory format
+ *
+ * The metadata from the `table_view` is copied into a host vector of bytes and the data from the
+ * `table_view` is copied into a `device_buffer`. Pass the output of this function into
+ * `cudf::unpack` to deserialize.
+ *
+ * @param input View of the table to pack
+ * @param[in] mr Optional, The resource to use for all returned device allocations
```
ttnghia, line 169:
```suggestion
 * @param mr Optional, The resource to use for all returned device allocations
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+/**
+ * @brief Produce the metadata used for packing a table stored in a contiguous buffer.
+ *
+ * The metadata from the `table_view` is copied into a host vector of bytes which can be used to
+ * construct a `packed_columns` or `packed_table` structure.  The caller is responsible for
+ * guaranteeing that that all of the columns in the table point into `contiguous_buffer`.
+ *
+ * @param table View of the table to pack
+ * @param contiguous_buffer A contiguous buffer of device memory which contains the data referenced
+ * by the columns in `table`
```
ttnghia, line 185:
```suggestion
 * @param contiguous_buffer A contiguous buffer of device memory which contains the data referenced
 *        by the columns in `table`
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ * @param buffer_size The size of `contiguous_buffer`
+ * @return Vector of bytes representing the metadata used to `unpack` a packed_columns struct
+ */
+packed_columns::metadata pack_metadata(table_view const& table,
```
ttnghia, line 189: Similar to my comment above, just use `std::vector<uint8_t/std::byte>` directly.

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+                                       uint8_t const* contiguous_buffer,
+                                       size_t buffer_size);
+
+/**
+ * @brief Deserialize the result of `cudf::pack`
```
ttnghia, line 194:
```suggestion
 * @brief Deserialize the result of `cudf::pack`.
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * No new device memory is allocated in this function.
+ *
+ * @param input The packed columns to unpack
+ * @return The unpacked `table_view`
+ */
+table_view unpack(packed_columns const& input);
+
+/**
+ * @brief Deserialize the result of `cudf::pack`
```
ttnghia, line 210:
```suggestion
 * @brief Deserialize the result of `cudf::pack`.
```

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,229 @@
...
+ * buffer `metadata`.
+ *
+ * It is the caller's responsibility to ensure that the `table_view` in the output does not outlive
+ * the data in the input.
+ *
+ * No new device memory is allocated in this function.
+ *
+ * @param metadata The host-side metadata buffer resulting from the initial pack() call
+ * @param gpu_data The device-side contiguous buffer storing the data that will be referenced by
+ * the resulting `table_view`
```
ttnghia, line 223:
```suggestion
 * @param gpu_data The device-side contiguous buffer storing the data that will be referenced by
 *        the resulting `table_view`
```

shwina (approved):
Python changes look good. Thanks!

cpp/include/cudf/contiguous_split.hpp
abellina: Since I am just moving code, I carried the doxygen groups because the old code used them (https://github.com/rapidsai/cudf/blob/branch-23.06/cpp/include/cudf/copying.hpp#L629)

When running doxygen you do get a module under Copying/Slicing that would group the logical slice `cudf::slice` and `cudf::contiguous_split` together as split operations (even though they are in separate headers), which is nice. I also see all kinds of other doxygen modules for factories, aggregates, scalar operations, lists, utilities, etc. It seems to be used either with `@ingroup` or `@addtogroup`.

I'd like to keep that indexing but if you think cuDF is moving away from this or the doxygen docs then I am happy to clean things up. We may want to add issues to clean up other parts of cuDF as well in that case.

cpp/include/cudf/contiguous_split.hpp
abellina: See my comment above as it's also for `@ingroup`.

cpp/include/cudf/contiguous_split.hpp
abellina: Yeah I just moved this code, but I see the point. I'll push a commit to remove the struct in favor of a `using`, which means the rest of the code that wants to use `metadata` as a type can still do so.

cpp/include/cudf/contiguous_split.hpp
abellina: See doxygen comment above.

cpp/include/cudf/contiguous_split.hpp
abellina: Yeah I just moved this code, but I see the point. I'll push a commit to remove the struct in favor of a `using`, which means the rest of the code that wants to use `metadata` as a type can still do so.

cpp/include/cudf/contiguous_split.hpp
davidwendt: The current usage of `@ingroup` and `@addtogroup` are to organize the Modules page as well as identify which declarations go with what header file. Please keep the indexing and tag usage as you described and implemented here.

cpp/include/cudf/contiguous_split.hpp
abellina: Curious about `@param[in]`. In this case, it is inconsistently used (e.g. `input` and `splits` should also be `@param[in]`). The reason I bring this up is because I see lots of usage of `@param[in]` in cuDF, and was wondering if this is going out of favor (being replaced by just `@param`)

davidwendt: In general, we drop the `[in]` especially when all parameters are `[in]`
The `[out]` is relatively rare and should be used to clear up any ambiguity.
https://github.com/rapidsai/cudf/blob/branch-23.06/cpp/doxygen/developer_guide/DOCUMENTATION.md#param

abellina: That sounds reasonable

cpp/include/cudf/contiguous_split.hpp
ttnghia: Such smart pointer is needed as the output of `continuous_split` so it can be used in Python (Cython requires the returned object to be on heap memory). So for simplicity, just take over the content of the input pointers:
```
 packed_columns(std::unique_ptr<metadata>&& md, std::unique_ptr<rmm::device_buffer>&& gd)
    : metadata_(std::move(*md)), gpu_data(std::move(*gd))
  {
  }
```

cpp/include/cudf/contiguous_split.hpp
abellina: I broke the python build with this. I think all I need to do is define the type using `ctypedef`, I am working on it.

cpp/include/cudf/contiguous_split.hpp
abellina: @ttnghia you marked this as resolved, but I haven't pushed it. Do you want me stick with `uint8_t`? I have been making the change locally and the main thing is I have to change other files in java/python and other source files on the cpp side. If we are OK with the bigger diff, I can spend time on moving us to this.

ttnghia: Oh sorry, accidentally marked as resolved :smile: . Yes, please address this.

cpp/include/cudf/contiguous_split.hpp
ttnghia: I think you can just remove this type completely and fix the python side, such as removing `cdef struct metadata`, replacing `metadata` with just `vector[uint8_t]`.

cpp/include/cudf/contiguous_split.hpp
ttnghia: That's OK, no problem.

cpp/include/cudf/contiguous_split.hpp
abellina: I have tried this, but it doesn't seem `std::byte` is defined in cpython:

```
cdef extern from "cudf/contiguous_split.hpp" namespace \
        "cudf::packed_columns" nogil:
    ctypedef vector[byte] metadata 
                       ^

cudf/python/cudf/cudf/_lib/cpp/contiguous_split.pxd:16:24: unknown type in template argument
```

Looking at `stddef` in cython, I see they didn't exposed that type: https://github.com/cython/cython/blob/1aef71764cd50e4cd2f638eac8e69fff3ced2224/Cython/Includes/libc/stddef.pxd

Should we leave as uint8_t? Or should I handle this differently? @ttnghia

abellina: @ttnghia I did remove the typedef for `metadata` and went directly with the `std::vector<uint8_t>` here: https://github.com/rapidsai/cudf/pull/13186/commits/d8dd74683e6e86f4de62456ae835720286bdd074

abellina:
@ttnghia Thanks for the reviews so far. I have addressed the things I could address, and commented on those I couldn't and am looking for any follow on steps here.

@shwina FYI, I had to change more of the python code and will need another check but I would wait until we settle on the c++ types first, since those are all at the interface level.

cpp/include/cudf/contiguous_split.hpp
abellina: Should be handled now with https://github.com/rapidsai/cudf/commit/d8dd74683e6e86f4de62456ae835720286bdd074

cpp/include/cudf/contiguous_split.hpp
ttnghia: Keeping `uint8_t` is fine. I want to move forward with a more standard way that libcudf devs prefer (`std::byte` instead of other types) but understand that it is may be out of scope of this PR.

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -58,13 +52,14 @@ struct packed_columns {
...
-  packed_columns(std::unique_ptr<metadata>&& md, std::unique_ptr<rmm::device_buffer>&& gd)
+  packed_columns(std::unique_ptr<std::vector<uint8_t>>&& md,
+                 std::unique_ptr<rmm::device_buffer>&& gd)
     : metadata_(std::move(md)), gpu_data(std::move(gd))
   {
   }
 
-  std::unique_ptr<metadata> metadata_;           ///< Host-side metadata buffer
-  std::unique_ptr<rmm::device_buffer> gpu_data;  ///< Device-side data buffer
+  std::unique_ptr<std::vector<uint8_t>> metadata_;  ///< Host-side metadata buffer
```
ttnghia, line 61:
```suggestion
  std::unique_ptr<std::vector<uint8_t>> metadata;  ///< Host-side metadata buffer
```

abellina: good eye.. will fix

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * Contains data from an array of columns in two contiguous buffers: one on host, which contains
+ * table metadata and one on device which contains the table data.
+ */
+struct packed_columns {
+  packed_columns()
+    : metadata_(std::make_unique<std::vector<uint8_t>>()),
+      gpu_data(std::make_unique<rmm::device_buffer>())
+  {
+  }
```
ttnghia, line 47:
```suggestion
  packed_columns() = default;
```

abellina: There is code depending on the default constructor to instantiate the `unique_ptrs` (since we need to keep these for now). If these were members directly of packed_columns I am with you, but as it is I can't remove.

cpp/include/cudf/contiguous_split.hpp
abellina: fixed

ttnghia (approved):

cpp/include/cudf/contiguous_split.hpp
```diff
@@ -0,0 +1,198 @@
...
+   * @param md Host-side metadata buffer
+   * @param gd Device-side data buffer
+   */
+  packed_columns(std::unique_ptr<std::vector<uint8_t>>&& md,
+                 std::unique_ptr<rmm::device_buffer>&& gd)
+    : metadata(std::move(md)), gpu_data(std::move(gd))
+  {
+  }
+
+  std::unique_ptr<std::vector<uint8_t>> metadata;  ///< Host-side metadata buffer
```
nvdbaranec, line 61: Just checking:  what is the intent behind removing the metadata struct and just converting it into a buffer?  Just clarity or some other reason?

ttnghia: I've requested to remove it.

The struct `metadata` just wraps around `std::vector` without anything else. There is also not any member functions doing anything on that `std::vector`. So why should we have that?

abellina: Should we introduce `typedef` or `using` so we can still do `packed_columns::metadata`? If we don't want to see `std::vector<uint8_t>` everywhere...

nvdbaranec: > I've requested to remove it.

The argue I'd make for keeping it is:  it is a return value from `pack_metadata` and by changing it to be a more generic blob of data you're losing a little bit of type safety.   That said, I don't think it's too big of a deal.

abellina: I think this one could go either way and the next PR introduces a class we can use to build such metadata. Perhaps we can continue on that one. For now, keeping the std::vector<uint8_t> exposed.

abellina:
@shwina I think this is ready for another python pass if you have a chance. I had to change the cython code again, and I don't trust myself.

nvdbaranec (approved):

shwina (approved):
Python changes LGTM

abellina:
I am noticing a JVM leak with this patch applied if I run with OOM injection logic we have added. I am working on identifying what is going on here.

abellina:
I triaged the JNI leak to: https://github.com/rapidsai/cudf/issues/13225

Essentially the cuDF version I tested without leaks vs. the new one in this PR missed some code that was added that causes creation of column vectors to be unsafe due to exceptions. That is unrelated to this change.

jlowe (approved):

abellina:
Thanks all for the reviews!

PR #13187, opened 2023-04-20
Fix `hostdevice_vector::subspan`

ttnghia:
This fixes a bug in `hostdevice_vector::subspan` that incorrectly constructs a `hostdevice_span` with the input parameters.

In particular, instead of passing the `count` parameter to `hostdevice_span` constructor, it (incorrectly) subtracts `offset` from `count`.


davidwendt (approved):

cpp/src/io/utilities/hostdevice_vector.hpp
```diff
@@ -164,13 +164,15 @@ class hostdevice_vector {
...
    *
-   * @return A typed hostdevice_span of the hostdevice_vector's data.
+   * @param offset The offset of the first element in the subspan
+   * @param count The number of elements in the subspan
+   * @return A typed hostdevice_span of the hostdevice_vector's data
    */
   [[nodiscard]] hostdevice_span<T> subspan(size_t offset, size_t count)
   {
     CUDF_EXPECTS(count >= offset, "End index cannot be smaller than the starting index.");
     CUDF_EXPECTS(count <= d_data.size(), "Slice range out of bounds.");
```
PointKernel, line 174: The naming of `count` is a bit misleading. According to the above code, it seems `offset` indicates the starting index and `count` refers to the end index. So `count - offset` should be correct. @ttnghia does your change break any existing unit tests?

ttnghia: No there were not any relevant unit tests for that function. I just added tests for it.

Oh these checks are also invalid. I'm going to update them.

PointKernel (approved):

bdice (approved):
Looks like a proper fix. Part of me wishes `subspan` was a free function rather than a method, so that `hostdevice_vector` would be closer in API to `std::vector`. If it were a method, I think it could be written for any container with `hostdevice` semantics (`hostdevice_vector` or `hostdevice_span`).

edit: I see there's a method `std::span<T>::subspan`, and this is based on that but applied to `hostdevice_vector`. This is taking the wrong lesson about container design from the STL, in my view, where a method could have been avoided in favor of a free function accepting any span-like object and returning a subspan that is a non-owning view subspan.

Issue #13193, opened 2023-04-21
[BUG] Changing the type of a scalar arg in a UDF results in an erroneous cache hit


brandon-b-miller:
**Describe the bug**
Suppose a user has a UDF defined as 
```python
def f(x, c):
    return x + c
```
and a series `s` that calls
```python
s.apply(f, args=(1,))
```
If the user later on calls
```python
s.apply(f, args=(1.5,))
```
cuDF tries to launch the kernel specialized for the first call again resulting in an error on the python side. The source of the bug is the fact that the types of the scalar args are not considered when generating the cache key.

**Steps/Code to reproduce bug**
```python
import cudf
import operator

op = operator.add

df = cudf.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
    
def f(row, k1, k2):
    x = op(row["a"], k1)
    y = op(x, k2)

    return y

df.apply(f, args=(1, 1))
df.apply(f, args=(1, 1.5))
```

**Expected behavior**
A new kernel shall be compiled and launched. 

**Environment details**
This is on 23.06.

**Additional context**
cc @charlesbluca

PR #13194, opened 2023-04-21
Add scalar argtypes to udf cache keys

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/13193


python/cudf/cudf/core/udf/utils.py
```diff
@@ -245,20 +245,22 @@ def _mask_get(mask, pos):
...
-def _generate_cache_key(frame, func: Callable, suffix="__APPLY_UDF"):
+def _generate_cache_key(frame, func: Callable, args, suffix="__APPLY_UDF"):
     """Create a cache key that uniquely identifies a compilation.
 
     A new compilation is needed any time any of the following things change:
     - The UDF itself as defined in python by the user
     - The types of the columns utilized by the UDF
     - The existence of the input columns masks
     """
+    scalar_argtys = tuple(typeof(arg) for arg in args)
```
bdice, line 256: Are `args` always scalars? Are there supported argument types that are unknown to `typeof`? (basic numerical types and strings appear to succeed, but lists, dicts, custom classes `class A: pass`, and `pd.Timestamp` fail).

```suggestion
    scalar_types = tuple(typeof(arg) for arg in args)
```

brandon-b-miller: We only support scalar args. But you're right there should be a check for this way further up the pipeline. Happy to add it as part of this PR.

brandon-b-miller: addressed in https://github.com/rapidsai/cudf/pull/13194/commits/f85ee4d45daca2958a080d743e69ad9de1918b21

galipremsagar (approved):

PR #13199, opened 2023-04-21
JNI changes for range-extents in window functions.

mythrocks:
This commit adds back-end JNI changes to support explicit range-extents for ranged window functions. The change is analogous to the addition of `cudf::range_window_bounds::extent_type`, in `libcudf`. This change will allow for `STRING` order-by columns for range window functions. It is required because without it, it would be impossible to differentiate between `UNBOUNDED` and `CURRENT ROW` for a window function over a `STRING` order-by column.

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/WindowOptions.java
+++ b/java/src/main/java/ai/rapids/cudf/WindowOptions.java
@@ -52,8 +52,6 @@ public class WindowOptions implements AutoCloseable {
   private final FrameType frameType;
   private final RangeExtentType precedingBoundsExtent;
   private final RangeExtentType followingBoundsExtent;
-//  private RangeExtentType precedingBoundsExtent = RangeExtentType.BOUNDED;
-//  private RangeExtentType followingBoundsExtent = RangeExtentType.BOUNDED;
 
   private WindowOptions(Builder builder) {
     this.minPeriods = builder.minPeriods;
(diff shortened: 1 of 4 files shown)


java/src/main/java/ai/rapids/cudf/WindowOptions.java
```diff
@@ -33,8 +50,10 @@ enum FrameType {ROWS, RANGE}
   private final int orderByColumnIndex;
   private final boolean orderByOrderAscending;
   private final FrameType frameType;
-  private final boolean isUnboundedPreceding;
-  private final boolean isUnboundedFollowing;
+  private final RangeExtentType precedingBoundsExtent;
+  private final RangeExtentType followingBoundsExtent;
+//  private RangeExtentType precedingBoundsExtent = RangeExtentType.BOUNDED;
+//  private RangeExtentType followingBoundsExtent = RangeExtentType.BOUNDED;
```
ttnghia, line 56: Need cleanup?

ttnghia (approved):

java/src/main/java/ai/rapids/cudf/WindowOptions.java
mythrocks: Thank you. Yes, I should've removed that.

mythrocks: Removed now.

revans2 (approved):

PR #11869, opened 2022-10-06
Fix writing of Parquet files with many fragments

etseidl:
## Description
This PR fixes an error that can occur when very small page sizes are used when writing Parquet files. #11551 changed from fixed 5000 row page fragments to a scaled value based on the requested max page size. For small page sizes, the number of fragments to process can exceed 64k. The number of fragments is used as the `y` dimension when calling `gpuInitPageFragments`, and when it exceeds 64k the kernel fails to launch, ultimately leading to an invalid memory access. 

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #3842, opened 2020-01-20
[REVIEW] ORC writer: add support for column statistics

OlivierNV:
stripe-level and file-level statistics, enabled by default.
resolves #3018

PR #13202, opened 2023-04-22
Clean up and simplify `gpuDecideCompression`

vuule:
## Description
Changed the block size to single warp, since only 32 threads are used in the kernel.
Simplify the kernel logic a bit and remove unnecessary atomic operations.

FWIW, the kernel is faster now; not important as it is a tiny part of E2E time.

## Checklist


ttnghia (approved):

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1315,56 +1315,44 @@ __global__ void __launch_bounds__(128, 8)
...
-    for (uint32_t page = t; page < num_pages; page += 32) {
-      auto& curr_page         = ck_g.pages[page];
-      uint32_t page_data_size = curr_page.max_data_size;
-      uncompressed_data_size += page_data_size;
-      if (auto comp_res = curr_page.comp_res; comp_res != nullptr) {
-        has_compression = true;
-        compressed_data_size += comp_res->bytes_written;
-        if (comp_res->status != compression_status::SUCCESS) { atomicAdd(&error_count, 1); }
-      }
+  __syncwarp();
```
bdice, line 1331: I think this should still be a `__syncthreads` to indicate that the whole block should synchronize? The fact that block size is equal to warp size here should be treated as a coincidence, not something required for correctness.

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1315,56 +1315,44 @@ __global__ void __launch_bounds__(128, 8)
...
+
+  uint32_t uncompressed_data_size = 0;
+  uint32_t compressed_data_size   = 0;
+  auto const num_pages            = ck_g.num_pages;
+  for (auto page = t; page < num_pages; page += warpSize) {
```
bdice, line 1336: Likewise, this should be described as a "block size" rather than a "warp size". Avoid assuming the block size is 32 for correctness.

bdice, line 1336: We don't use `warpSize` in libcudf kernels. We always use `cudf::detail::warp_size` instead. https://github.com/rapidsai/cudf/blob/52342789a367153bedbc804efdc0a54a7b6ed083/cpp/include/cudf/detail/utilities/cuda.cuh#L36

This is important for various reasons -- we cannot assume that hardware will have a constant warp size of 32 forever. While we're not in a particularly good place to refactor the code for non-32 warp sizes, this would at least help us track where warp size assumptions are being used.

This rule is violated in [one other place, in merge.cu](https://github.com/rapidsai/cudf/blob/52342789a367153bedbc804efdc0a54a7b6ed083/cpp/src/merge/merge.cu#L100) which we should fix.

See also:
- https://github.com/rapidsai/cudf/pull/11869#discussion_r998705107
- https://github.com/rapidsai/cudf/pull/3842#discussion_r373320116

cpp/src/io/parquet/page_enc.cu
bdice, line 1331: Can we write this kernel in a block-size agnostic way? Unlike `__syncthreads();`, using `__syncwarp();` assumes that `block_size == warp_size == 32`.

bdice (changes requested):

cpp/src/io/parquet/page_enc.cu
```diff
@@ -2167,7 +2155,7 @@ void EncodePages(device_span<gpu::EncPage> pages,
 
 void DecideCompression(device_span<EncColumnChunk> chunks, rmm::cuda_stream_view stream)
 {
-  gpuDecideCompression<<<chunks.size(), 128, 0, stream.value()>>>(chunks);
+  gpuDecideCompression<<<chunks.size(), 32, 0, stream.value()>>>(chunks);
```
karthikeyann, line 2158: As suggested by others, I suggest to replace magic numbers with `constexpr`

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1315,56 +1315,44 @@ __global__ void __launch_bounds__(128, 8)
   }
 }
 
-// blockDim(128, 1, 1)
-__global__ void __launch_bounds__(128) gpuDecideCompression(device_span<EncColumnChunk> chunks)
+// blockDim(32, 1, 1)
+__global__ void __launch_bounds__(32) gpuDecideCompression(device_span<EncColumnChunk> chunks)
```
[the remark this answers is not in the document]
vuule: wasn't aware of warp_size. I'll use it wherever possible.

cpp/src/io/parquet/page_enc.cu
vuule: That depends on how we would scale the parallelism with multiple warps. If any warps worked on a single `chunks` element, then, yes, we would need to syn all threads in the block. But, with multiple warps, IMO this kernel should actually have each warp would work on a separate `chunks` element. In this case we don't need to synchronize different warps and `__syncwarp` is still the right option.
I understand that my change left this ambiguous as warp size is used interchangeably for block and warp size. I'll try to make this clearer.

vuule: Modified the kernel to work with any number of warps in a block. The size can be adjusted via `constexpr` `decide_compression_warps_in_block`. Used warp_size as well, so we should be magic number-free now :)

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1315,56 +1315,56 @@ __global__ void __launch_bounds__(128, 8)
...
-  __shared__ typename warp_reduce::TempStorage temp_storage[2];
-  __shared__ volatile bool has_compression;
+  __shared__ typename warp_reduce::TempStorage temp_storage[decide_compression_warps_in_block][2];
 
-  uint32_t t                      = threadIdx.x;
-  uint32_t uncompressed_data_size = 0;
-  uint32_t compressed_data_size   = 0;
-  uint32_t num_pages;
+  auto const lane_id  = threadIdx.x % cudf::detail::warp_size;
+  auto const warp_id  = threadIdx.x / cudf::detail::warp_size;
```
vuule, line 1332: Question for the reviewers: Are there maybe helper functions for this? Looks very generic.

bdice: Not that I am aware of.

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1315,56 +1315,56 @@ __global__ void __launch_bounds__(128, 8)
...
-      if (auto comp_res = curr_page.comp_res; comp_res != nullptr) {
-        has_compression = true;
-        compressed_data_size += comp_res->bytes_written;
-        if (comp_res->status != compression_status::SUCCESS) { atomicAdd(&error_count, 1); }
+  __syncwarp();
+
+  uint32_t uncompressed_data_size = 0;
+  uint32_t compressed_data_size   = 0;
+  auto const num_pages            = ck_g[warp_id].num_pages;
+  for (auto page = lane_id; page < num_pages; page += warpSize) {
```
bdice, line 1346:
```suggestion
  for (auto page = lane_id; page < num_pages; page += cudf::detail::warp_size) {
```

bdice (approved):
One more fix for `warpSize`. Otherwise I think this is better!

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1315,56 +1315,56 @@ __global__ void __launch_bounds__(128, 8)
...
+// blockDim(decide_compression_block_size, 1, 1)
+__global__ void __launch_bounds__(decide_compression_block_size)
+  gpuDecideCompression(device_span<EncColumnChunk> chunks)
 {
-  // After changing the way structs are loaded from coop to normal, this kernel has no business
-  // being launched with 128 thread block. It can easily be a single warp.
-  __shared__ __align__(8) EncColumnChunk ck_g;
-  __shared__ __align__(4) unsigned int error_count;
+  __shared__ __align__(8) EncColumnChunk ck_g[decide_compression_warps_in_block];
+  __shared__ __align__(4) unsigned int compression_error[decide_compression_warps_in_block];
```
ttnghia, line 1327: Why do we align them manually? And why do we need to align them?

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1315,56 +1315,56 @@ __global__ void __launch_bounds__(128, 8)
...
-        has_compression = true;
-        compressed_data_size += comp_res->bytes_written;
-        if (comp_res->status != compression_status::SUCCESS) { atomicAdd(&error_count, 1); }
+  __syncwarp();
+
+  uint32_t uncompressed_data_size = 0;
+  uint32_t compressed_data_size   = 0;
+  auto const num_pages            = ck_g[warp_id].num_pages;
+  for (auto page = lane_id; page < num_pages; page += cudf::detail::warp_size) {
+    auto& curr_page           = ck_g[warp_id].pages[page];
```
ttnghia, line 1347: `const`?

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1315,56 +1315,56 @@ __global__ void __launch_bounds__(128, 8)
...
-      if (auto comp_res = curr_page.comp_res; comp_res != nullptr) {
-        has_compression = true;
-        compressed_data_size += comp_res->bytes_written;
-        if (comp_res->status != compression_status::SUCCESS) { atomicAdd(&error_count, 1); }
+  __syncwarp();
+
+  uint32_t uncompressed_data_size = 0;
+  uint32_t compressed_data_size   = 0;
+  auto const num_pages            = ck_g[warp_id].num_pages;
+  for (auto page = lane_id; page < num_pages; page += cudf::detail::warp_size) {
```
ttnghia, line 1346: `page_id` or `page_idx` should be a better name?

cpp/src/io/parquet/page_enc.cu
vuule: It allows more efficient access, at least in theory. I'm not the one who added the alignment, and I also haven't tested how this alignment impacts performance in practice.

PR #13205, opened 2023-04-24
Remove UNKNOWN_NULL_COUNT where it can be easily computed

vyasr:
## Description

Contributes to #11968

## Checklist


davidwendt (approved):

cpp/include/cudf/structs/structs_column_view.hpp
```diff
@@ -87,9 +90,12 @@ class structs_column_view : public column_view {
    * @throw cudf::logic error if this is an empty column
    *
    * @param index The index of the child column to return
+   * @param stream The stream on which to perform the operation. Uses the default CUDF
+   * stream if none is specified.
```
ttnghia, line 94:
```suggestion
   *        stream if none is specified.
```

cpp/include/cudf/structs/structs_column_view.hpp
```diff
@@ -87,9 +90,12 @@ class structs_column_view : public column_view {
...
    * @return The child column sliced relative to the parent's offset and size
    */
-  [[nodiscard]] column_view get_sliced_child(int index) const;
+  [[nodiscard]] column_view get_sliced_child(
+    int index, rmm::cuda_stream_view stream = cudf::get_default_stream()) const;
```
ttnghia, line 98: This is a breaking change. I recommend removing this default stream value and fixing all the downstream usages. Otherwise, it is very likely that a wrong stream will be used anywhere.

divyegala: Isn't it okay to have a default option for the stream in a public API?

ttnghia: I don't think so.

ttnghia: Oh wait, this is the public API? Humnn, then we may in trouble, since this API will be called a lot internally that requires explicit `stream`.

davidwendt: Public APIs can have defaults.
This PR should be addressing where this API is called within this repo.
I agree this should probably be labeled as a breaking change.

vyasr: Right, what we want for now is for the API to have a default stream since it's public but to ensure that all internal calls forward along the stream. I will make the necessary change now.

Since it's adding a parameter with a default it won't break existing calls to the function, so I chose not to mark it as breaking, but happy to modify that label if preferred.

divyegala: I would say it's breaking because existing calls should be modified in places where the right stream needs to be passed along.

vyasr: That's a great point, callers should loudly be made aware of this fact especially since the whole point is to remove implicit streams.

vyasr: OK I have updated all call sites and marked the PR as breaking.

divyegala (approved):

ttnghia (approved):
I'm curious: Can this be tested to check incorrect stream usage?

vyasr:
> I'm curious: Can this be tested to check incorrect stream usage?

Not quite yet, but we should be able to do so soon. There is some validation happening now, but I suspect that the validation you want to see is verifying that we're not accidentally falling back to `cudf::get_default_stream()` somewhere internally instead of passing the input stream through. We can't do that quite yet because at the moment the input stream is always `cudf::get_default_stream()`. Once we have stream parameters in public APIs we will be able to test this (or if we want to do it sooner, we could start testing detail APIs; that's a discussion we've had before and may again as we get closer to exposing streams).

PR #13212, opened 2023-04-25
Fix string_scalar stream usage in write_json.cu

davidwendt:
## Description
Fixes stream usage for `cudf::string_scalar` variables used in `write_json.cu`. The passed-in stream should be used instead of the default stream supplied by the `string_scalar` constructor. Also removed some unused variables. A couple of strings are hardcoded and so these were moved to device code as constants to avoid the host-to-copy that occurs within the `string_scalar` constructor.

## Checklist


cpp/src/io/json/write_json.cu
```diff
@@ -545,8 +542,10 @@ struct column_to_strings_fn {
       column.null_count(),
       cudf::detail::copy_bitmask(column, stream_, rmm::mr::get_current_device_resource()),
       stream_);
-    auto l_pre = string_scalar{"["}, l_post = string_scalar{"]"}, sep = string_scalar{","},
-         elem_narep = string_scalar{"null"};
+    auto l_pre      = string_scalar{"[", true, stream_};
+    auto l_post     = string_scalar{"]", true, stream_};
+    auto sep        = string_scalar{",", true, stream_};
+    auto elem_narep = string_scalar{"null", true, stream_};
```
karthikeyann, line 548: This could be moved and added as class member.

karthikeyann (commented):
Looks good to me 👍
Thank you for catching this!

cpp/src/io/json/write_json.cu
davidwendt: You mean use the `narep` class member here instead of `elem_narep`?

karthikeyann: No. I meant, all 4 scalars could be moved to the class. Right now, this scalar is initialized everytime on a list column call.

karthikeyann: Since we already have `include_nulls` in options to control null behaviour for structs and table, `narep` could be kept common for both struct and list. List will always include nulls, so list_narep cannot be null.

davidwendt: Ok. I made them share a `narep` scalar value. Let me know if that looks right.

karthikeyann (approved):
LGTM 👍

cpp/src/io/json/write_json.cu
```diff
@@ -730,15 +740,15 @@ void write_json(data_sink* out_sink,
...
+  string_scalar d_line_terminator_with_row_end{"}" + line_terminator, true, stream};
+  string_scalar d_line_terminator{line_terminator, true, stream};
 
   // write header: required for non-record oriented output
   // header varies depending on orient.
   // write_chunked_begin(out_sink, table, user_column_names, options, stream, mr);
   // TODO This should go into the write_chunked_begin function
   std::string const list_braces{"[]"};
-  string_scalar d_list_braces{list_braces};
+  string_scalar d_list_braces{list_braces, true, stream};
```
karthikeyann, line 751: can these be `const` too?

bdice (approved):

PR #13221, opened 2023-04-25
Set the null count in output columns in the CSV reader

vuule:
## Description
CSV reader used to keep the null count of the output columns as `UNKNOWN_NULL_COUNT`. This has performance implications on further use of these columns.
This PR adds a validity counter to convert_csv_to_cudf kernel. This counter is then used to set the correct null count in the columns.
Also added a test that explicitly checks the null counts, as it is never explicitly used in the CSV reader tests.

## Checklist


cpp/src/io/csv/reader_impl.cu
```diff
@@ -859,7 +867,7 @@ table_with_metadata read_csv(cudf::io::datasource* source,
         const std::string dblquotechar(2, parse_opts.quotechar);
         std::unique_ptr<column> col = cudf::make_strings_column(*out_buffers[i]._strings, stream);
         out_columns.emplace_back(
-          cudf::strings::replace(col->view(), dblquotechar, quotechar, -1, mr));
+          cudf::strings::detail::replace(col->view(), dblquotechar, quotechar, -1, stream, mr));
```
vuule, line 870: Unrelated change, just noticed that the version that the detail API needs to be used here to ensure that the right stream is used.

cpp/tests/io/csv_test.cpp
```diff
@@ -169,7 +169,8 @@ void check_float_column(cudf::column_view const& col_lhs,
 
   CUDF_TEST_EXPECT_COLUMN_PROPERTIES_EQUIVALENT(col_lhs,
                                                 (wrapper<T>{data.begin(), data.end(), validity}));
-  CUDF_EXPECTS(col_lhs.null_count() == 0, "All elements should be valid");
+  CUDF_EXPECTS(col_lhs.null_count() == 0 and col_rhs.null_count() == 0,
+               "All elements should be valid");
```
vuule, line 173: This is one test that was (kind of) checking the null count. With this additional check, it fails when the null count is incorrect.

davidwendt (approved):

ttnghia (approved):

PR #13224, opened 2023-04-26
Fix decimal scale reductions in `_get_decimal_type`

charlesbluca:
## Description

This PR modifies `_get_decimal_type` to follow the scale reduction steps outlined in:

- https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql?view=sql-server-ver16#remarks

It seems like before, we were only reducing scale with the formula used in case 1 of multiplication/division operations.

Noting that this is a breaking change in that before, we were intentionally avoiding overflow errors here but it seems like in some multiplication/division cases, this is unavoidable? According to the docs when `scale < 6 and (precision - scale) > 32`:

> The scale isn't changed if it's less than 6 and if the integral part is greater than 32. In this case, an overflow error might be raised if it can't fit into decimal(38, scale).

Not sure if we actually want to raise in this case or do something else.

EDIT:

After testing a reproducer that would result in overflow:

```python
import cudf

decimal_type = cudf.Decimal128Dtype(20, -20)
a = cudf.Series([int(1e20), int(2e20), int(3e20)], dtype=decimal_type)
b = cudf.Series([int(4e20), int(5e20), int(6e20)], dtype=decimal_type)

# results in precision=38, scale=-40
res = a * b
```

It seems like this does error on `branch-23.06`, so I think I was just confused by this comment:

https://github.com/rapidsai/cudf/blob/421ac6797a5a8cae129b11584254e7eb955eb1fe/python/cudf/cudf/core/column/decimal.py#L402-L405

## Checklist


vyasr:
@charlesbluca it looks like there's a failing test on this PR, can you take a look?

python/cudf/cudf/core/column/decimal.py
```diff
@@ -366,6 +366,20 @@ def _get_decimal_type(lhs_dtype, rhs_dtype, op):
     else:
         raise NotImplementedError()
 
+    # potentially reduce scale and precision
+    if precision > cudf.Decimal128Dtype.MAX_PRECISION:
+        if op in {"__add__", "__sub__"}:
```
brandon-b-miller, line 371: Can we combine the handling for each specific op to the section of code above so we do all the `scale` and `precision` calculations for each op happen in once place?

python/cudf/cudf/core/column/decimal.py
```diff
@@ -366,6 +366,20 @@ def _get_decimal_type(lhs_dtype, rhs_dtype, op):
     else:
         raise NotImplementedError()
 
+    # potentially reduce scale and precision
+    if precision > cudf.Decimal128Dtype.MAX_PRECISION:
```
brandon-b-miller, line 370: Can we make it so that we don't go through the top level `cudf` namespace here? It can make it harder to resolve circular import issues later on down the line.

python/cudf/cudf/core/column/decimal.py
```diff
@@ -402,15 +416,12 @@ def _get_decimal_type(lhs_dtype, rhs_dtype, op):
...
-    # with max possible scale & precision, see example of this demonstration
-    # here: https://learn.microsoft.com/en-us/sql/t-sql/data-types/
-    # precision-scale-and-length-transact-sql?view=sql-server-ver16#examples
-    scale = min(
-        scale, cudf.Decimal128Dtype.MAX_PRECISION - (precision - scale)
+    # if we've reached this point, we cannot create a decimal type without
+    # overflow; raise an informative error
+    raise ValueError(
+        f"Performing {op} between columns of type {repr(lhs_dtype)} and "
+        "{repr(rhs_dtype)} would result in overflow"
```
brandon-b-miller, line 423:
```suggestion
        f"{repr(rhs_dtype)} would result in overflow"
```

python/cudf/cudf/core/column/decimal.py
```diff
@@ -402,15 +416,12 @@ def _get_decimal_type(lhs_dtype, rhs_dtype, op):
...
-    # Instead of raising an overflow error, we create a `Decimal128Dtype`
-    # with max possible scale & precision, see example of this demonstration
-    # here: https://learn.microsoft.com/en-us/sql/t-sql/data-types/
-    # precision-scale-and-length-transact-sql?view=sql-server-ver16#examples
-    scale = min(
-        scale, cudf.Decimal128Dtype.MAX_PRECISION - (precision - scale)
+    # if we've reached this point, we cannot create a decimal type without
+    # overflow; raise an informative error
+    raise ValueError(
+        f"Performing {op} between columns of type {repr(lhs_dtype)} and "
```
brandon-b-miller, line 422: Do we actually need to call `repr` here? Shouldn't the printing action trigger `__repr__` automatically?

charlesbluca: Think the printing triggers `__str__`, which ends up giving an error message that doesn't include the scale/precision:

```
ValueError: Performing __mul__ between columns of type decimal128 and decimal128 would result in overflow
```

Thought that forcing use of `__repr__` here to show the scale and precision would be more informative, though open to suggestions for a better error message in this case (cc @shwina if you have thoughts)

python/cudf/cudf/core/column/decimal.py
```diff
@@ -357,12 +357,23 @@ def _get_decimal_type(lhs_dtype, rhs_dtype, op):
...
+        if precision > Decimal128Dtype.MAX_PRECISION:
+            precision = Decimal128Dtype.MAX_PRECISION
+            scale = Decimal128Dtype.MAX_PRECISION - max(p1 - s1, p2 - s2)
+    elif op in {"__mul__", "__div__"}:
+        if op == "__mul__":
+            scale = s1 + s2
+            precision = p1 + p2 + 1
+        else:
+            scale = max(6, s1 + p2 + 1)
+            precision = p1 - s1 + s2 + scale
```
charlesbluca, line 369: cc @brandon-b-miller not a fan of the nested conditionals here, did this so we wouldn't have to repeat the scale/precision reduction code below for both the `__mul__` and `__div__` - any preference here?

python/cudf/cudf/core/column/decimal.py
shwina: Honestly, we should probably include the scale/precision in both str and repr. But I think changing `__str__` is too large an undertaking for this PR so I'm good with this change.

python/cudf/cudf/core/column/decimal.py
brandon-b-miller: spent a little time trying to organize it differently and concluded this is about as good as it gets.

brandon-b-miller (approved):
LGTM

brandon-b-miller (approved):
LGTM

shwina (approved):
Thanks, @charlesbluca !

PR #13227, opened 2023-04-26
Remove default null-count parameter from cudf::make_strings_column factory

davidwendt:
## Description
Removes the `UNKNOWN_NULL_COUNT` default parameter from `cudf::make_strings_column` factory functions and updates the callers to always pass the null-count.

Labeling this with breaking since it is a public function.

## Checklist


cpp/benchmarks/string/factory.cu
```diff
@@ -63,7 +63,7 @@ static void BM_factory(benchmark::State& state)
 
   for (auto _ : state) {
     cuda_event_timer raii(state, true, cudf::get_default_stream());
-    cudf::make_strings_column(pairs);
+    cudf::make_strings_column(pairs, cudf::get_default_stream());
```
karthikeyann, line 66: will previous thrust call on `thrust::device` use the same stream as `cudf::get_default_stream()` ?

karthikeyann: The default value on the `stream` argument does not seem to be removed in this PR. should it be included too?

davidwendt: No, removing default values only applies to detail functions. We'll be addressing the stream parameter in public APIs like this one later.

harrism (approved):

karthikeyann (approved):

Issue #13230, opened 2023-04-26
[BUG] libcudf conda packages are shipping dependencies in the package


bdice:
**Describe the bug**
Currently, libcudf conda packages are shipping libraries/headers from libcudf dependencies. In particular:

libcudf's conda package is 548 MB (all reported sizes are unzipped). It currently ships libraries/headers from:
- kvikio (< 1 MB)
- libcudacxx (5.2 MB)
- nvbench (4.1 MB)
- nvcomp (46 MB)

**Expected behavior**
libcudf conda packages should not ship its dependencies, but should instead list other conda packages as dependencies. This will prevent files from outside libcudf from being shipped in libcudf packages.

### Specific proposals for resolution

**kvikio (easy)**
kvikio should be easy to fix. RAPIDS already produces conda packages of kvikio and just need to use those, rather than letting it be found via CPM and repackaged in libcudf.

**libcudacxx (medium difficulty)**
`libcudacxx` is part of CCCL, along with Thrust and CUB. Historically, librmm has shipped most of the CCCL headers in `include/rapids` because they've been found by CPM during the librmm build and repackaged there. However, this is not ideal. libcudacxx is an exception to librmm shipping a RAPIDS-vendored CCCL because it was not needed by librmm -- only libcudf. I suspect we're repackaging libcudacxx in many RAPIDS packages right now. The fix here is to adopt [rapids-core-dependencies](https://github.com/rapidsai/rapids-cmake/tree/branch-23.06/conda/recipes/rapids_core_dependencies) across all of RAPIDS, which should be mostly ready for use, if I recall correctly.

**nvbench (very difficult)**
Ideally, `nvbench` should be its own conda package, but we often require patched versions. Therefore, we should consider moving nvbench headers to a special subdirectory `include/rapids` as we do for CCCL (see https://github.com/rapidsai/rapids-cmake/pull/98). We should also do a more granular review to decide if (and where) we want to be shipping files like `bin/nvbench-ctl` or `lib/objects-Release/nvbench.main/main.cu.o`.

**nvcomp (medium difficulty)**
`nvcomp` should be shipping in a separate conda package (probably on either the rapidsai or nvidia channels). Then we make it a dependency of libcudf. It is the largest component of libcudf that is being repackaged (around 10% of libcudf's size).

cc: @vuule (with whom I discussed this)

jakirkham:
cc @madsbk

jakirkham:
[...]

PR #13231, opened 2023-04-26
Use libkvikio conda packages in libcudf, add explicit libcufile dependency.

bdice:
## Description
Provides a partial solution to #13230. During conda builds of `libcudf`, `libkvikio` is currently being fetched by CPM rather than supplied by a conda dependency. This means `libkvikio` headers are being shipped as part of `libcudf`'s packages, which is not desirable.

I also added `libcufile[-dev]` as an explicit dependency of libcudf. Note that this is only available on linux64 (amd64), not aarch64 (arm64). We should always make the cuFile library available at build time for conda packages on amd64.

The impacts of this change are:
- `libcudf` conda packages will no longer ship `libkvikio` headers (those headers will instead be supplied by `libkvikio` packages). This reduces the size of libcudf and prevents clobbering files from `libkvikio`.
- `libcudf` will have a dependency on `libcufile` from the `nvidia` channel on `linux64` (but not `aarch64`, since libcufile packages currently do not exist for `aarch64`).

This change does not impact builds outside of conda-build.

## Checklist


ajschmidt8 (approved):

bdice:
Verified that builds show the libkvikio conda package being used on arm64.
```
-- CPM: using local package KvikIO@23.06.0
```

Also adding the cuFile dependency caused `-DCUFILE_FOUND` to be added to every compile line on amd64. This meant it had to build without sccache and took a long time! However, that's also a confirmation that cuFile support should now be working in packages built by CI. 😄 cc: @madsbk @vuule

ci/release/update-version.sh
bdice, line 1: FYI @ajschmidt8 -- you requested some changes on [cuspatial](https://github.com/rapidsai/cuspatial/pull/1110#pullrequestreview-1407854018) and [ucxx](https://github.com/rapidsai/ucxx/pull/36#discussion_r1178429239) for recent changes to the `update-version.sh` script. I started to make those changes but cudf's update script is significant more complex than ucxx/cuspatial and I didn't want to set the wrong example (since so many packages look to cudf for guidance). If you want this to be updated in the near term, I would appreciate your help!

ajschmidt8: Sounds good.

There are a lot of process changes I want to make around these scripts in the future.

Realistically I probably won't start that initiative until `23.08` or even `23.10`.

bdice:
For the record, pip packages are built with cuFile support. I checked both CUDA 11.8 and CUDA 12.0 builds. AMD64 only, because cuFile isn't supported on ARM64.
```
  -- CPM: adding package KvikIO@23.06 (branch-23.06)
  -- Found cuFile: /usr/local/cuda/lib64/libcufile.so
```

vuule (approved):
FWIW, looks good to me :)
This is an important fix for cuFile integration, thank you @bdice !

PR #13232, opened 2023-04-26
Add metadata_builder helper class

abellina:
This PR introduces a helper class `metadata_builder` within the `cudf::detail` namespace, that allows `pack` and `contiguous_split` to create metadata for a packed buffer without having access to metadata internals (e.g. `serialized_column`).

The class makes it possible for callers to build metadata independent of the current implementations in `pack`, that are relying on `column_view`s being defined (actual valid cuDF columns). The chunked pack work will use this class to create metadata without needing to create a `column_view` in the first place, since it doesn't have a valid base pointer for the columns to be instantiated.

### DIFF
--- a/cpp/include/cudf/detail/contiguous_split.hpp
+++ b/cpp/include/cudf/detail/contiguous_split.hpp
@@ -67,7 +67,7 @@ class metadata_builder {
    * @brief Destructor that will be implemented as default, required because metadata_builder_impl
    * is incomplete at this stage.
    */
-  ~metadata_builder();
+  ~metadata_builder() = default;
 
   /**
    * @brief Add a column to this metadata builder.
--- a/cpp/include/cudf/detail/contiguous_split.hpp
+++ b/cpp/include/cudf/detail/contiguous_split.hpp
@@ -44,5 +44,70 @@ packed_columns pack(cudf::table_view const& input,
                     rmm::cuda_stream_view stream,
                     rmm::mr::device_memory_resource* mr);
 
+// opaque implementation of `metadata_builder` since it needs to use
+// `serialized_column`, which is only defined in pack.cpp
+class metadata_builder_impl;
+
+/**
+ * @brief Helper class that creates packed column metadata.
+ *
+ * This class is an interface to the opaque metadata that is used to
+ * describe `contiguous_split` and `pack` results.
+ */
+class metadata_builder {
+ public:
+  /**
+   * @brief Construct a new metadata_builder.
+   *
+   * @param num_root_columns is the number of top-level columns
+   */
+  explicit metadata_builder(size_type num_root_columns);
+
+  /**
+   * @brief Destructor that will be implemented as default, required because metadata_builder_impl
+   * is incomplete at this stage.
+   */
+  ~metadata_builder();
+
+  /**
+   * @brief Add a column to this metadata builder.
+   *
+   * Callers must call this function for the parent column and followed by any children,
+   * in the order maintained in the column/column_view.
+   *
+   * Example: given a table with a nested column "a" with 2 children, and a non-nested column "b":
+   *
+   *   1) add_column_to_meta(col_a)
+   *   2) add_column_to_meta(col_a_child_1)
+   *   3) add_column_to_meta(col_a_child_2)
+   *   4) add_column_to_meta(col_b)
+   *
+   * @param col_type column data type
+   * @param col_size column row count
+   * @param col_null_count column null count
+   * @param data_offset data offset from the column's base ptr,
+   *                    or -1 for an empty column
+   * @param null_mask_offset null mask offset from the column's base ptr,
+   *                    or -1 for a column that isn't nullable
+   * @param num_children number of chilren columns
+   */
+  void add_column_to_meta(data_type col_type,
+                          size_type col_size,
+                          size_type col_null_count,
+                          int64_t data_offset,
+                          int64_t null_mask_offset,
+                          size_type num_children);
+
+  /**
+   * @brief Builds the opaque metadata for all added columns.
+   *
+   * @returns A vector containing the serialized column metadata
+   */
+  std::vector<uint8_t> build();
+
+ private:
+  std::unique_ptr<metadata_builder_impl> impl;
+};
+
 }  // namespace detail
 }  // namespace cudf
--- a/cpp/include/cudf/detail/contiguous_split.hpp
+++ b/cpp/include/cudf/detail/contiguous_split.hpp
@@ -89,7 +89,7 @@ class metadata_builder {
    *                    or -1 for an empty column
    * @param null_mask_offset null mask offset from the column's base ptr,
    *                    or -1 for a column that isn't nullable
-   * @param num_children number of chilren columns
+   * @param num_children number of children columns
    */
   void add_column_to_meta(data_type col_type,
                           size_type col_size,
--- a/cpp/include/cudf/detail/contiguous_split.hpp
+++ b/cpp/include/cudf/detail/contiguous_split.hpp
@@ -61,7 +61,7 @@ class metadata_builder {
    *
    * @param num_root_columns is the number of top-level columns
    */
-  explicit metadata_builder(size_type num_root_columns);
+  explicit metadata_builder(size_type const num_root_columns);
 
   /**
    * @brief Destructor that will be implemented as default, required because metadata_builder_impl
@@ -77,10 +77,10 @@ class metadata_builder {
    *
    * Example: given a table with a nested column "a" with 2 children, and a non-nested column "b":
    *
-   *   1) add_column_to_meta(col_a)
-   *   2) add_column_to_meta(col_a_child_1)
-   *   3) add_column_to_meta(col_a_child_2)
-   *   4) add_column_to_meta(col_b)
+   *   1) add_column_info_to_meta(col_a)
+   *   2) add_column_info_to_meta(col_a_child_1)
+   *   3) add_column_info_to_meta(col_a_child_2)
+   *   4) add_column_info_to_meta(col_b)
    *
    * @param col_type column data type
    * @param col_size column row count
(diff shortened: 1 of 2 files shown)


abellina:
I have tested with this patch in the past, and I know the cpp tests passed before I posted the PR. I had a linker error with pyarrow when it was trying to dlopen one of its internal libraries (so not even to the tests). As a result I have been recreating the conda environment and still waiting on that. Marking this as reviewable for now as I think getting some :eyes: should be good and I don't expect new issues to show up in the python tests.

abellina:
> I have tested with this patch in the past, and I know the cpp tests passed before I posted the PR. I had a linker error with pyarrow when it was trying to dlopen one of its internal libraries (so not even to the tests). As a result I have been recreating the conda environment and still waiting on that. Marking this as reviewable for now as I think getting some eyes should be good and I don't expect new issues to show up in the python tests.

Local build finished and `test_pack` passed (CI also passed..)

cpp/src/copying/pack.cpp
```diff
@@ -208,6 +227,27 @@ table_view unpack(uint8_t const* metadata, uint8_t const* gpu_data)
   return table_view{get_columns(num_columns)};
 }
 
+metadata_builder::metadata_builder(size_type num_root_columns)
+  : impl(std::make_unique<metadata_builder_impl>())
+{
+  impl->add_column_to_meta(data_type{type_id::EMPTY}, num_root_columns, 0, -1, -1, 0);
+}
+
+metadata_builder::~metadata_builder() = default;
```
mythrocks, line 236: We can move this to the header.

cpp/include/cudf/detail/contiguous_split.hpp
```diff
@@ -44,5 +44,70 @@ packed_columns pack(cudf::table_view const& input,
...
+   *
+   * @param num_root_columns is the number of top-level columns
+   */
+  explicit metadata_builder(size_type num_root_columns);
+
+  /**
+   * @brief Destructor that will be implemented as default, required because metadata_builder_impl
+   * is incomplete at this stage.
+   */
+  ~metadata_builder();
```
mythrocks, line 70:
```suggestion
  ~metadata_builder() = default;
```

abellina:
Thanks @mythrocks, updated.

mythrocks (changes requested):
Sorry for the slow review. I'll look through this some more tomorrow.

cpp/include/cudf/detail/contiguous_split.hpp
```diff
@@ -44,5 +44,70 @@ packed_columns pack(cudf::table_view const& input,
...
+
+  /**
+   * @brief Add a column to this metadata builder.
+   *
+   * Callers must call this function for the parent column and followed by any children,
+   * in the order maintained in the column/column_view.
+   *
+   * Example: given a table with a nested column "a" with 2 children, and a non-nested column "b":
```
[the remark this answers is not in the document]
abellina: Good question. This metadata is really just saying that the column of cudf type string, but not specifying how exactly it is going to be encoded. `contiguous_split` special cases a string column and handles these underlying buffers.

For a list or struct types, in addition to the parent type and metadata, other children information is carried (children types, nullability, number of rows, ..)  these are first-class columns (like strings), so that's the level of detail in the metadata.

cpp/include/cudf/detail/contiguous_split.hpp
```diff
@@ -44,5 +44,70 @@ packed_columns pack(cudf::table_view const& input,
...
+   *
+   *   1) add_column_to_meta(col_a)
+   *   2) add_column_to_meta(col_a_child_1)
+   *   3) add_column_to_meta(col_a_child_2)
+   *   4) add_column_to_meta(col_b)
+   *
+   * @param col_type column data type
+   * @param col_size column row count
+   * @param col_null_count column null count
```
hyperbolic2346, line 87: Can there be an unknown null count? I know cudf supports such a thing, but I don't know if it makes sense in the metadata.

cpp/include/cudf/detail/contiguous_split.hpp
```diff
@@ -44,5 +44,70 @@ packed_columns pack(cudf::table_view const& input,
...
+ * describe `contiguous_split` and `pack` results.
+ */
+class metadata_builder {
+ public:
+  /**
+   * @brief Construct a new metadata_builder.
+   *
+   * @param num_root_columns is the number of top-level columns
+   */
+  explicit metadata_builder(size_type num_root_columns);
```
hyperbolic2346, line 64:
```suggestion
  explicit metadata_builder(size_type const num_root_columns);
```

cpp/include/cudf/detail/contiguous_split.hpp
```diff
@@ -44,5 +44,70 @@ packed_columns pack(cudf::table_view const& input,
...
+   * @param null_mask_offset null mask offset from the column's base ptr,
+   *                    or -1 for a column that isn't nullable
+   * @param num_children number of children columns
+   */
+  void add_column_to_meta(data_type col_type,
+                          size_type col_size,
+                          size_type col_null_count,
+                          int64_t data_offset,
+                          int64_t null_mask_offset,
+                          size_type num_children);
```
hyperbolic2346, line 99:
```suggestion
  void add_column_to_meta(data_type const col_type,
                          size_type const col_size,
                          size_type const col_null_count,
                          int64_t const data_offset,
                          int64_t const null_mask_offset,
                          size_type const num_children);
```

cpp/src/copying/pack.cpp
```diff
@@ -156,27 +156,46 @@ std::vector<uint8_t> pack_metadata(ColumnIter begin,
...
+class metadata_builder_impl {
+ public:
+  metadata_builder_impl() = default;
+
+  void add_column_to_meta(data_type col_type,
+                          size_type col_size,
+                          size_type col_null_count,
+                          int64_t data_offset,
+                          int64_t null_mask_offset,
+                          size_type num_children)
```
hyperbolic2346, line 177:
```suggestion
  void add_column_to_meta(data_type const col_type,
                          size_type const col_size,
                          size_type const col_null_count,
                          int64_t const data_offset,
                          int64_t const null_mask_offset,
                          size_type const num_children)
```

cpp/src/copying/pack.cpp
```diff
@@ -156,27 +156,46 @@ std::vector<uint8_t> pack_metadata(ColumnIter begin,
...
+  {
+    metadata.emplace_back(
+      col_type, col_size, col_null_count, data_offset, null_mask_offset, num_children);
+  }
+
+  std::vector<uint8_t> build()
```
hyperbolic2346, line 183:
```suggestion
  std::vector<uint8_t> build() const
```

cpp/src/copying/pack.cpp
```diff
@@ -208,6 +227,25 @@ table_view unpack(uint8_t const* metadata, uint8_t const* gpu_data)
   return table_view{get_columns(num_columns)};
 }
 
+metadata_builder::metadata_builder(size_type num_root_columns)
```
hyperbolic2346, line 230:
```suggestion
metadata_builder::metadata_builder(size_type const num_root_columns)
```

cpp/src/copying/pack.cpp
```diff
@@ -208,6 +227,25 @@ table_view unpack(uint8_t const* metadata, uint8_t const* gpu_data)
...
+  : impl(std::make_unique<metadata_builder_impl>())
+{
+  impl->add_column_to_meta(data_type{type_id::EMPTY}, num_root_columns, 0, -1, -1, 0);
```
hyperbolic2346, line 233:
```suggestion
  // first metadata entry is a stub indicating how many total (top level) columns
  // there are
  impl->add_column_to_meta(data_type{type_id::EMPTY}, num_root_columns, 0, -1, -1, 0);
```

cpp/src/copying/pack.cpp
```diff
@@ -208,6 +227,25 @@ table_view unpack(uint8_t const* metadata, uint8_t const* gpu_data)
...
+}
+
+void metadata_builder::add_column_to_meta(data_type col_type,
+                                          size_type col_size,
+                                          size_type col_null_count,
+                                          int64_t data_offset,
+                                          int64_t null_mask_offset,
+                                          size_type num_children)
```
hyperbolic2346, line 241:
```suggestion
void metadata_builder::add_column_to_meta(data_type const col_type,
                                          size_type const col_size,
                                          size_type const col_null_count,
                                          int64_t const data_offset,
                                          int64_t const null_mask_offset,
                                          size_type const num_children)
```

hyperbolic2346 (changes requested):
Mostly const additions requested.

cpp/src/copying/pack.cpp
```diff
@@ -156,27 +156,46 @@ std::vector<uint8_t> pack_metadata(ColumnIter begin,
...
+                          size_type num_children)
+  {
+    metadata.emplace_back(
+      col_type, col_size, col_null_count, data_offset, null_mask_offset, num_children);
+  }
+
+  std::vector<uint8_t> build()
+  {
+    // convert to anonymous bytes
+    std::vector<uint8_t> metadata_bytes;
```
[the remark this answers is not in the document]
abellina: So `std::byte` isn't defined in cython. @ttnghia had a similar suggestion in my prior PR https://github.com/rapidsai/cudf/pull/13186 and we ended up settling for `uint8_t` because otherwise the python code couldn't use this field.

cpp/src/copying/pack.cpp
```diff
@@ -156,27 +156,46 @@ std::vector<uint8_t> pack_metadata(ColumnIter begin,
...
-
-  return metadata_bytes;
+  return mb.build();
 }
 
+class metadata_builder_impl {
+ public:
+  metadata_builder_impl() = default;
+
+  void add_column_to_meta(data_type col_type,
```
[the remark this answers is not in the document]
abellina: you got it.

cpp/src/copying/pack.cpp
mythrocks: Seconded. I considered mentioning this as well. (There were no `std::move()`s in the `build()` code, so this should be safely `const`-able.)

cpp/src/copying/pack.cpp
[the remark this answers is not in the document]
mythrocks: Oh, wow. I hadn't considered that. TIL.

cpp/src/copying/pack.cpp
abellina: > so this should be safely const-able.

`const`ed it has been!

hyperbolic2346 (approved):

PR #13240, opened 2023-04-27
Make `is_decompression_disabled` and `is_compression_disabled` thread-safe

vuule:
## Description
Adds a mutex to guard the use of the memoization map for the status of nvCOMP features.

## Checklist

### DIFF
--- a/cpp/src/io/comp/nvcomp_adapter.cpp
+++ b/cpp/src/io/comp/nvcomp_adapter.cpp
@@ -21,6 +21,8 @@
 
 #include <nvcomp/snappy.h>
 
+#include <mutex>
+
 #define NVCOMP_DEFLATE_HEADER <nvcomp/deflate.h>
 #if __has_include(NVCOMP_DEFLATE_HEADER)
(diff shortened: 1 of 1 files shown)


ttnghia:
I wonder if we can just call `is_(de)compression_disabled_impl` directly without the caching mechanism, as these function is very cheap.

vuule:
> I wonder if we can just call `is_(de)compression_disabled_impl` directly without the caching mechanism, as these functions are very cheap.

Yes, we could. I only added this mechanism to be able to log the result of each query once. Otherwise we would either need a similar map for logging, or we would have to allow logs to be very repetitive.

abellina:
@vuule I ran with this and it did not reproduce.

abellina (approved):

abellina:
We are hitting all kinds of segfaults in CI that I think are all related to this issue: https://github.com/NVIDIA/spark-rapids/issues/8194, so the spark-rapids CI is currently blocked or limping along.

ttnghia (approved):

PR #13241, opened 2023-04-27
Introduce `pandas_compatible` option in `cudf`

galipremsagar:
## Description
This PR adds `mode.pandas_compatible` option in `cudf`, which if set to true will try to mimic pandas behavior as much as possible if there is an API inconsistency with pandas.

## Checklist


python/cudf/cudf/api/extensions/__init__.py
```diff
@@ -1,12 +1,14 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
 
 from cudf.api.extensions.accessor import (
     register_dataframe_accessor,
     register_index_accessor,
     register_series_accessor,
 )
+from pandas.api.extensions import no_default
```
galipremsagar, line 8: Note: We have `no_default` introduced in `pandas_2.0_feature_branch`: https://github.com/rapidsai/cudf/pull/13152, but not in `branch-23.06`.

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -3915,12 +3916,18 @@ def groupby(
...
         group_keys=False,
         squeeze=False,
         observed=False,
         dropna=True,
     ):
+        if sort is no_default:
+            if cudf.get_option("mode.pandas_compatible"):
+                sort = True
+            else:
+                sort = False
```
wence-, line 3929:
```suggestion
            sort = cudf.get_option("mode.pandas_compatible")
```

wence- (approved):
Tiny cleanup, otherwise looks good.

python/cudf/cudf/api/extensions/__init__.py
vyasr: Is it possible to cherry-pick that change over rather than doing it manually here? Would make your life easier when the final merge comes around, I suspect the current approach will have some (small) conflicts.

shwina: Since Prem is out for a while I'll take this over tomorrow and apply the commit(s).

galipremsagar: We merged the `no_default` change to `pandas_2.0_feature_branch` as a merge commit: https://github.com/rapidsai/cudf/commit/81565cfa490348892b7e883faf78a7000fa2ada0, so cherry-picking it will bring in additional `reset_index` changes that aren't intended for this branch yet. So the best way is to manually resolve the conflict in `pandas_2.0_feature_branch` after this is merged into `branch-23.06`, which anyway has to be done on a daily basis to be up-to-date with `branch-23.06`.

vyasr: OK that's fine with me, was just hoping to save some extra work later.

PR #13242, opened 2023-04-27
Fix the row index stream order in ORC reader

### LINKED ISSUE #11890: [BUG] cuDF ORC reader incorrectly reads file written by pyarrow

vuule:
## Description
Fixes #11890 

Use fixed order when reading row index streams.
The order is `PRESENT`, `DATA`, `SECONDARY`/`LENGTH` (maps to `DATA2`). 
Is any of these is absent in the column, relative order is maintained. Thus, the order is a sub-array of the one above.

Also simplified some logic related to stream order, as we do not need to pass it from the host. Instead, we only pass a bitmap to denote which streams are present.
Updated the xfail test, as it now passes :)

## Checklist


PointKernel (approved):

bdice (approved):
Approving Python test change.

cpp/src/io/orc/reader_impl.cu
```diff
@@ -213,16 +204,15 @@ size_t gather_stream_info(const size_t stripe_index,
...
-          chunk.skip_count          = idx.second;
-
-          if (idx.first == gpu::CI_DICTIONARY) {
+        auto& chunk           = chunks[stripe_index][col];
+        auto const index_type = get_stream_index_type(stream.kind);
+        if (index_type < gpu::CI_NUM_STREAMS) {
+          chunk.strm_id[index_type]  = stream_info.size();
+          chunk.strm_len[index_type] = stream.length;
+          // NOTE: skip_count field is temporarily used to track the presence of index streams
+          chunk.skip_count |= 1 << index_type;
```
vyasr, line 213: Is the `skip_count` now equivalent to what was being computed before? This change seems separate from the stream ordering, but also the skip_count looks like it would be subtly different now from before.

cpp/src/io/orc/stripe_init.cu
```diff
@@ -226,6 +226,26 @@ enum row_entry_state_e {
...
+{
+  constexpr std::array full_order = {CI_PRESENT, CI_DATA, CI_DATA2};
+
+  std::array<uint32_t, full_order.size()> partial_order;
+  auto curr_idx = 0;
+  for (auto index_type : full_order) {
+    // If the index type is present, add it to the partial order
+    if (index_types_bitmap & (1 << index_type)) { partial_order[curr_idx++] = index_type; }
+  }
+  return partial_order;
```
vyasr, line 246: Could you use `std::copy_if` instead here? It is order-preserving.

cpp/src/io/orc/stripe_init.cu
```diff
@@ -239,11 +259,14 @@ static uint32_t __device__ ProtobufParseRowIndexEntry(rowindex_state_s* s,
...
+  auto const stream_order                = index_order_from_index_types(s->chunk.skip_count);
 
   const uint8_t* cur      = start;
   row_entry_state_e state = NOT_FOUND;
-  uint32_t length = 0, strm_idx_id = s->chunk.skip_count >> 8, idx_id = 1, ci_id = CI_PRESENT,
-           pos_end = 0;
+  uint32_t length         = 0;
+  uint32_t idx_id         = 0;
+  uint32_t pos_end        = 0;
+  uint32_t ci_id          = CI_NUM_STREAMS;
```
vyasr, line 269: Hmm why did some of these values change?

cpp/src/io/orc/stripe_init.cu
```diff
@@ -226,6 +226,26 @@ enum row_entry_state_e {
...
 };
 
+/**
+ * @brief Calculates the order of index streams based on the index types present in the column.
+ *
+ * @param index_types_bitmap The bitmap of index types showing which index streams are present
+ *
+ * @return The order of index streams
+ */
+static auto __device__ index_order_from_index_types(uint32_t index_types_bitmap)
```
ttnghia, line 236: Technically all free functions are static.

cpp/src/io/orc/reader_impl.cu
vuule: Yes, `skip_count` value changed. It used to hold order of streams within a column. Each stream type was held in a byte of the skip_count value. Since we don't need the order during decode, I replaced this with a bitmap that uses stream type as bit "index".
In either case, this is a temporary (ab)use of this data member. I'm not a fan of this approach but did not change it in this PR, as I want to keep the changes minimal (this is a risky PR IMO). We can replace this reuse of `skip_count` in a future PR.

cpp/src/io/orc/stripe_init.cu
vuule: Great question!
`idx_id` was one because of the... complicated... way `strm_idx_id` was implemented. New implementation counts from zero (as expected).
`ci_id` is now `CI_NUM_STREAMS` to denote an invalid value. `CI_PRESENT` is a valid stream type in this context so I changed this to a value that makes sure the execution flow is as expected. `ci_id` should be set to the actual stream type before it's used. Unfortunately, we do need to keep track of the stream type outside of the loop.

cpp/src/io/orc/stripe_init.cu
vuule: Maybe I'm misunderstanding the comment, but `static` is used here to give the function internal linkage - it is not called outside of this translation unit. Maybe this is not relevant for `__device__` functions, but we use this extensively in cuIO.

cpp/src/io/orc/stripe_init.cu
vuule: great suggestion, done.

cpp/src/io/orc/stripe_init.cu
vyasr: Perhaps @ttnghia is confusing the concept of [static member variables](https://en.cppreference.com/w/cpp/language/static) with usage of `static` as a [storage class specifier](https://en.cppreference.com/w/cpp/language/storage_duration)? They are two completely different concepts that unfortunately reuse the same keyword. Otherwise I am not sure what the comment is referring to.

vyasr (approved):

cpp/src/io/orc/stripe_init.cu
vuule: Yes, if the input is valid and the code is correct, line 316 will be executed before `ci_id` is used. If that's not the case, I'd rather it have an invalid value than an incorrect one :)

cpp/src/io/orc/stripe_init.cu
ttnghia: > but static is used here to give the function internal linkage

This makes sense. Otherwise, I don't see why it needs to be `static`. Sorry for the confusion.

Typically we can just put the function into an anonymous namespace for that purpose.

PR #13243, opened 2023-04-28
Fix structs_column_wrapper ctors to copy input column wrappers

### LINKED ISSUE #13200: [BUG] `cudf::test::print` seg faults on nested structs
**Describe the bug**
When used to print out nested structs, the `cudf::test::print` utility triggers a segmentation fault rather than printing in some cases.

Add these two lines:
```
    cudf::test::print(output, std::cerr);                                                                                                                                                                  
    //cudf::test::print(expected_structs, std::cerr);        
```
just before [this assertion](https://github.com/rapidsai/cudf/blob/branch-23.06/cpp/tests/structs/utilities_tests.cpp#L683) in the `NestedStruct_Sliced` test. The result should be fine, until the second line is commented out. Then the code will seg fault.

**Expected behavior**
No seg fault

davidwendt:
## Description

Fixes the `cudf::test::structs_column_wrapper` constructors that accept `column_wrapper` instances to copy the underlying column rather than move it. The constructor signatures do not provide any indication that the passed wrappers will be destroyed. Also, the other nested column wrapper constructors make copies unless the parameters are decorated with move declarations (i.e. `&&`). 

Closes #13200

## Checklist

### DIFF
--- a/cpp/include/cudf_test/column_wrapper.hpp
+++ b/cpp/include/cudf_test/column_wrapper.hpp
@@ -1807,7 +1807,9 @@ class structs_column_wrapper : public detail::column_wrapper {
     std::transform(child_column_wrappers.begin(),
                    child_column_wrappers.end(),
                    std::back_inserter(child_columns),
-                   [&](auto column_wrapper) { return column_wrapper.get().release(); });
+                   [&](auto const& column_wrapper) {
+                     return std::make_unique<cudf::column>(column_wrapper.get());
+                   });
     init(std::move(child_columns), validity);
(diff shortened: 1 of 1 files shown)


ttnghia (approved):

vyasr:
> This is great. There is a -lot- of test code out there that has to do some ugly code jiggling to get around this.

@nvdbaranec should we open an issue to track that cleanup? Not a high priority task, but it would be nice to know it could be done (I don't have the context for these tests but it sounds like you have some in mind?).

PR #13245, opened 2023-04-28
Allow empty input table in ast `compute_column`

### LINKED ISSUE #13236: [BUG]  AST fails with zero-row inputs
**Describe the bug**

AST can't handle inputs with zero rows because it can't get the right numbers of kernel's blocks for this case. This is an edge case, but it should be covered as pandas do. 

```
C++ exception with description "CUDF failure at: /opt/mambaforge/conda-bld/libcudf-ext_1681746453367/work/cpp/include/cudf/detail/utilities/cuda.cuh:66: num_blocks must be > 0" thrown in the test body.
```
```
import cudf
df = cudf.DataFrame({'A': [], 'B': []})
df.eval('A + B')
```

`RuntimeError: CUDF failure at: /opt/conda/conda-bld/work/cpp/include/cudf/detail/utilities/cuda.cuh:66: num_blocks must be > 0`

**Expected behavior**

For the example above it should return a column with zero rows. 

```
import pandas as pd
df = pd.DataFrame({'A': [], 'B': []})
df.eval('A + B')
Out[10]: Series([], dtype: float64)
```
 - Environment location: conda
 -  branch-23.04, origin/branch-23.04

wence-:
## Description

If the input table is empty we can still produce an output of
appropriate type by letting the parser run and inferring types.
In this case though, we do not want to actually run a kernel,
since we would try and launch with zero blocks which is not
allowed.

Closes #13236.

## Checklist

### DIFF
--- a/cpp/src/transform/compute_column.cu
+++ b/cpp/src/transform/compute_column.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2022, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2023, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


vuule (approved):

vyasr (approved):
Thanks for the patch!

Issue #13254, opened 2023-04-30
[BUG] cuDF inner join and iloc seemingly broken


danieldemarch:
Assume ticks is a list of cuDF DataFrames. If it matters, these cuDF DataFrames were made by converting pandas DataFrames to cuDF.

```python
dfa= ticks[i].copy()
dfb= ticks[j].copy()

dfc= dfa.join(dfb, how='inner', rsuffix="2").copy()

indy = int(len(dfc)*0.65)

dummy = dfc.iloc[:indy].copy()

print(dummy.tail(5))
```
The .copy() can be removed optionally, I get the same behavior with/without them.

This behavior produces two errors. First, the dataframes are all sorted in chronological order before the inner join. Post-inner join, the sort is wrong. It's like halfway sorted but with random insertions - e.g. a bunch of 04/04/2023 timestamps in a row and then a random o4/03/2023, before back to 04/04/2023.

Second, the "dummy = dfc.iloc[:indy].copy()" produces inconsistent results when running the same block of code twice. It just doesn't work. It seems to cycle between 5-6 different outcomes if I repeatedly run the cell.

Sticking with vanilla pandas does not produce either of the above errors. It's purely the act of converting to cuDF DataFrames that produces this bug.

Very disappointed. This is the sort of thing unit testing is supposed to catch. If it's some bizarre setting on the cuDF DataFrame - no setting should produce this outcome.

shwina:
Thanks for reporting, @danieldemarch. Is this reproducible with any data or specific to a certain dataset?

shwina:
It's worth noting that by default, cuDF's merge/join operations do not guarantee output ordering or sortedness (https://docs.rapids.ai/api/cudf/legacy/api_docs/api/cudf.dataframe.merge).

Does using `sort=True` in the call to `join` help?

danieldemarch:
Pardon - so by default, arrays will be essentially randomly reordered after a merge?

shwina:
Correct. Please see this section of our documentation that provides more details about result ordering: https://docs.rapids.ai/api/cudf/stable/user_guide/pandas-comparison/#result-ordering

vyasr:
There has also been a lot of discussion on this in https://github.com/rapidsai/cudf/issues/1781 (and https://github.com/rapidsai/cudf/issues/5286, although that's referring to a different function but the same sorting-related behavior).

danieldemarch:
Thanks much!

wence-:
[...]

PR #13255, opened 2023-05-01
Fix and clarify notes on result ordering

shwina:
I noticed when answering #13254 that the code example in this section of our documentation was incorrect and the text itself could use some improving.


docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -82,9 +82,16 @@ using `.from_arrow()` or `.from_pandas()`.
 
 ## Result ordering
 
-By default, `join` (or `merge`) and `groupby` operations in cuDF
-do *not* guarantee output ordering.
-Compare the results obtained from Pandas and cuDF below:
+In Pandas, `join` and `groupby` operations provide certain guarantees
+about the order of rows in the result returned.  In Pandas `join`, the
```
wence-, line 86:
```suggestion
about the order of rows in the result returned.  In a Pandas `join`, the
```

docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -82,9 +82,16 @@ using `.from_arrow()` or `.from_pandas()`.
...
+order of join keys are either preserved or sorted lexicographically by
```
wence-, line 87:
```suggestion
order of join keys is either preserved or sorted lexicographically by
```
(subject of the sentence is "order" not "join keys")

But I think I don't quite understand the explanation of this restriction (how can ordering be either preserved or sorted by default for a join key?).

docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -82,9 +82,16 @@ using `.from_arrow()` or `.from_pandas()`.
...
+default.  `groupby` sorts the group keys, and preserves the order of
+rows within each group. In some cases, disabling this option in Pandas
+can yield better performance.
+
+By contrast, cuDF's default behavior is to return rows in a
+non-deterministic order to maximize performance.  Compare the results
+obtained from Pandas and cuDF below:
```
wence-, line 94: Maybe note that for a groupby _within_ a group the ordering of the input is preserved?

docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -107,10 +114,11 @@ Compare the results obtained from Pandas and cuDF below:
  10  640.00
 ```
 
-To match Pandas behavior, you must explicitly pass `sort=True`:
+If you require that results be returned in a predictable order, you
+must pass the `sort=True` option explicitly:
```
wence-, line 118: Is it worth mentioning about the index being useful to write ordering-independent code?

wence- (commented):
Minor nitpicks

docs/cudf/source/user_guide/pandas-comparison.md
shwina: Do we _need_ to make that guarantee? Internally, we do use a stable sort when doing a sort-based groupby but can we leave that as an implementation detail?

docs/cudf/source/user_guide/pandas-comparison.md
shwina: Done!

docs/cudf/source/user_guide/pandas-comparison.md
wence-: Oh probably not, but I don't know the details of what we guarantee as an API statement, and what we provide as an implementation detail. So let us leave it as is here.

docs/cudf/source/user_guide/pandas-comparison.md
wence-: OK, let us therefore write something like:

> In a Pandas `join`, the order of join keys is (depending on the particular style of join being performed) either preserved or sorted lexicographically.

vyasr:
@shwina @wence- the docs have been updated from underneath this PR, so it needs some conflict resolution. Do you want to fix up the wording to something that you both like?

docs/cudf/source/user_guide/pandas-comparison.md
vyasr: Added this.

wence- (approved):

PR #13262, opened 2023-05-01
Clean up buffers in case AssertionError

### LINKED ISSUE #13225: [BUG] JNI GPU memory leak when assertions are enabled
This line: https://github.com/rapidsai/cudf/blob/branch-23.06/java/src/main/java/ai/rapids/cudf/ColumnVector.java#L153, creates a new `OffHeapState` using a `DeviceMemoryBuffer` (`contiguousBuffer`). This `OffHeapState` owns the contiguous buffer, hence it `incRefCount`s it. 

Since https://github.com/rapidsai/cudf/pull/13071 was added, the construction of `ColumnView` can throw because `ColumnView.hasNonEmptyNulls` performs GPU memory allocations (and we could get an OOM right there). The very next thing that happens in line https://github.com/rapidsai/cudf/blob/branch-23.06/java/src/main/java/ai/rapids/cudf/ColumnVector.java#L153 with the call to `super`  (following the chain), is the call to `ColumnView`'s constructor with the contiguous buffer that has been `incRefCount`ed and the exception that can happen here.

In addition to an OOM, you could get `AssertionError`, again with the same effect of a dangling contiguous buffer.

We need to make sure that the ColumnView constructor's clean this up. I am not 100% sure on the ownership of `OffHeapState` re: `ColumnView`, e.g. should it close the offHeap state blindly..

razajafri:
In case `ColumnView` throws an `AssertionError` which it can after #13071, we should close out the open buffers. 

fixes #13225 

## Checklist


java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -68,7 +75,14 @@ protected ColumnView(ColumnVector.OffHeapState state) {
     type = DType.fromNative(ColumnView.getNativeTypeId(viewHandle), ColumnView.getNativeTypeScale(viewHandle));
     rows = ColumnView.getNativeRowCount(viewHandle);
     nullCount = ColumnView.getNativeNullCount(viewHandle);
-    AssertEmptyNulls.assertNullsAreEmpty(this);
+    try {
+      AssertEmptyNulls.assertNullsAreEmpty(this);
+    } catch (AssertionError ae) {
+      // This constructor is only called from ColumnVector so calling close will call
+      // close in ColumnVector which will close out the offHeapState
```
ttnghia, line 82: That means calling derived class method in base class constructor. I suspect that it may be incorrect or won't produce the desired effect. What I'm seeing is only the `close` method of this class is called, then the exception is propagated to the derived class without calling `close` of the derived class.

razajafri: > That means calling derived class method in base class constructor.

No, not always. Only if the object is of the `ColumnVector` type, if the object is of ColumnView type then the `close()` in `ColumnView` will be called.
> What I'm seeing is only the close method of this class is called, then the exception is propagated to the derived class without calling close of the derived class

```
ColumnView c = new ColumnVector()
c.close();
```
^ this will call the `close()` in `ColumnVector`

```
ColumnView c = new ColumnView()
c.close();
```
^ this will call `close()` in `ColumnView`
In both cases we will get the desired results

ttnghia: Yeah, but the example above is showing that `close()` is called on an already-constructed object. What I mentioned above is `close()` called during object construction. When an instance of the derived class is being constructed, the constructor of the base class is called first (`super(....)`) in the constructor of the derived class. Since the construction of the derived class is not yet completed, polymorphism doesn't take place.

ttnghia: It seems that the same rule should be followed in Java: https://stackoverflow.com/questions/3404301/whats-wrong-with-overridable-method-calls-in-constructors

abellina: +1 we should not rely on close. In this case, `ColumnView.close` is also a noop for a straight `ColumnView` at this time, as it checks that `offHeap` is null, and only then it takes action.

abellina:
You want to treat this case differently since it is a special error on construction.

In the viewHandle constructor case, the code you have seems fine but I would also 0 it out for good measure.

In the OffHeapState constructor case, the state of concern is `offHeap`. This is created externally in `ColumnVector` and handed down to the view. I think the only thing you want to do is to close `offHeap` state directly  if you are going to throw. I would also 0 out viewHandle for good measure, it's a pointer to nothing after `offHeap` closes.

I would put comments on both constructors that this is now happening and add a `throws AssertionError` for each constructor that now could throw that to make it extra clear and we don't forget this in the future.

java/src/main/java/ai/rapids/cudf/ColumnView.java
razajafri: @ttnghia @abellina thank you for taking a look at this. 

I lost sight in the second case that it is a constructor and will not behave properly. Will fix

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -43,32 +43,50 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
...
+   * @throws AssertionError if the address points to a nested-type view with non-empty nulls
    */
   ColumnView(long address) {
     this.viewHandle = address;
     this.type = DType.fromNative(ColumnView.getNativeTypeId(viewHandle), ColumnView.getNativeTypeScale(viewHandle));
     this.rows = ColumnView.getNativeRowCount(viewHandle);
     this.nullCount = ColumnView.getNativeNullCount(viewHandle);
     this.offHeap = null;
-    AssertEmptyNulls.assertNullsAreEmpty(this);
+    try {
```
abellina, line 57: This covers the single `ColumnView` case. Note that there is one more case we should cover (that I can see in cuDF): `ColumnView.splitAsViews`. This function iterates over handles and instantiates the `ColumnView`s in a loop.

If one of those iterations asserts, then we leak the handles.

I think what can be done in that case is to declare a `Throwable` object (but not instantiated) before going into the loop, and set it to the first assertion in a try/catch. If you see an assertion while instantiating CV `i`, you know you have dealt with this particular handle, so what you need to do is then close out the remaining handles at: `[i+1, nativeHandles.length)`,  and `[0, i-1]`.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -43,32 +43,50 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
...
     viewHandle = state.getViewHandle();
     type = DType.fromNative(ColumnView.getNativeTypeId(viewHandle), ColumnView.getNativeTypeScale(viewHandle));
     rows = ColumnView.getNativeRowCount(viewHandle);
     nullCount = ColumnView.getNativeNullCount(viewHandle);
-    AssertEmptyNulls.assertNullsAreEmpty(this);
+    try {
+      AssertEmptyNulls.assertNullsAreEmpty(this);
+    } catch (AssertionError ae) {
+      // cleanup offHeap
+      offHeap.clean(false);
```
abellina, line 86: Lets set `offHeap = null` as well. 

For this constructor, I see `ColumnView.slice` being used the same way that  `ColumnView.splitAsViews` was above, so we need to take the same treatment here and close out any handles not closed by the constructor.

I also see `Table.convertToRows` and `Table.convertToRowsFixedWidthOptimized`. Note that the `Table` constructor (`Table(long[] cudfColumns)`) has protections against this already, so that's a pattern to follow.

razajafri: > Lets set offHeap = null as well

can't do that, it's a final variable. I would want to keep it final, we are cleaning the `offHeap` and since this is a constructor, the object will not be instantiated so there is no way for anyone to access the `offHeap`.

java/src/main/java/ai/rapids/cudf/ColumnView.java
razajafri: Why just handle this specific case? Why not just catch the `Throwable` and just close everything out and exit the method after rethrowing?

razajafri:
@abellina do you have anything else to add? I think I have addressed all your concerns

java/src/main/java/ai/rapids/cudf/ColumnView.java
abellina: I am fine if you want to change this to catch Throwable, I think you are looking at the other method calls in this constructor as they could throw as well? As long as you think they can throw, it would be a good change IMHO.

java/src/main/java/ai/rapids/cudf/ColumnView.java
abellina: Ah.. missed that. Then that's fine +1.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -649,8 +667,17 @@ public final ColumnVector ifElse(Scalar trueValue, Scalar falseValue) {
...
     long[] nativeHandles = slice(this.getNativeView(), indices);
     ColumnVector[] columnVectors = new ColumnVector[nativeHandles.length];
-    for (int i = 0; i < nativeHandles.length; i++) {
-      columnVectors[i] = new ColumnVector(nativeHandles[i]);
+    try {
+      for (int i = 0; i < nativeHandles.length; i++) {
+        columnVectors[i] = new ColumnVector(nativeHandles[i]);
+      }
+    } catch (Throwable t) {
+      for (ColumnView columnView: columnVectors) {
```
abellina, line 675: In addition to closing the column vectors/views in the array, we also have pending `nativeHandles`, sorry I just caught this one. In order to really make safer, we should also need to close the handles which are leaked. Should probably create a helper handles -> vectors, or handles -> views, if we don't have those already, since all the other places you changed are similar.

I would change the loop like so:

```
try {
  for (int i = 0; i < nativeHandles.length; i++) {
    columnVectors[i] = new ColumnVector(nativeHandles[i]);
    nativeHandles[i] = 0; // mark the handle as consumed
  }
} catch (Throwable t) {
  for (ColumnVector cv : columnVectors) {
    if (cv != null) {
      cv.close();
    }
  }
  for (long handle : nativeHandles) { 
    if (handle != 0) {
      ColumnVector.deleteCudfColumn(handle);
    }
  }
  throw t;
}
```

razajafri: Good catch! But why not close all the `nativeHandles`? We got them as a result of `slice` They should all be cleaned out since we have encountered an error.

razajafri: I believe the same thing should happen in `splitAsViews`

abellina: Yes, but since you have handed some handles to the vector/view, the expectation is that the vector/view will close the handle they were given. So you don't want to double delete these.

razajafri:
@abellina thank you for taking a look and catching things that I had missed. I have covered the missed cases I think. I have also created a helper method in ColumnView and also replaced places where it should be called instead of the code we were using before. Please take a look again when you have a chance

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -95,13 +95,7 @@ public Table(long[] cudfColumns) {
...
       this.rows = columns[0].getRowCount();
     } catch (Throwable t) {
-      for (int i = 0; i < cudfColumns.length; i++) {
-        if (this.columns[i] != null) {
-          this.columns[i].close();
-        } else {
-          ColumnVector.deleteCudfColumn(cudfColumns[i]);
-        }
-      }
+      ColumnView.cleanupColumnViews(cudfColumns, this.columns);
```
abellina, line 98: need to zero-out `cudfColumns[i]` as we iterate in the loop that creates the column vectors. I would put this try catch around the creation of the column vectors for `this.columns`.

The views later, that's a different set of failures I believe.

razajafri: I don't want to change too much of this function, and the only reason I touched it was to refactor the cleanup code into a separate method.

razajafri: I will push an update to the zeroing out of the cudfColumns pointers

java/src/main/java/ai/rapids/cudf/BitVectorHelper.java
```diff
@@ -71,7 +71,7 @@ private static void shiftSrcLeftAndWriteToDst(HostMemoryBuffer src, HostMemoryBu
   /**
    * This method returns the length in bytes needed to represent X number of rows
    * e.g. getValidityLengthInBytes(5) => 1 byte
-   * getLengthInBytes(7) => 1 byte
+   * getValidityLengthInBytes(7) => 1 byte
```
abellina, line 74: pls update copyrights.

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
```diff
@@ -6677,6 +6678,36 @@ void testApplyBooleanMaskFromListOfStructure() {
...
+    List<Integer> list2 = Arrays.asList(7, 8, 9);
+    List<Integer> list3 = null;
+    try (ColumnVector input = ColumnVectorTest.makeListsColumn(DType.INT32, list0, list1, list2, list3);
+         BaseDeviceMemoryBuffer baseValidityBuffer = input.getDeviceBufferFor(BufferType.VALIDITY);
+         BaseDeviceMemoryBuffer baseOffsetBuffer = input.getDeviceBufferFor(BufferType.OFFSET);
+         HostMemoryBuffer newValidity = HostMemoryBuffer.allocate(BitVectorHelper.getValidityAllocationSizeInBytes(3))) {
+
+      newValidity.copyFromDeviceBuffer(baseValidityBuffer);
+      BitVectorHelper.setNullAt(newValidity, 1);
+      DeviceMemoryBuffer validityBuffer = DeviceMemoryBuffer.allocate(BitVectorHelper.getValidityAllocationSizeInBytes(3));
```
abellina, line 6694: wrap this one in try-with-resources

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
```diff
@@ -6677,6 +6678,36 @@ void testApplyBooleanMaskFromListOfStructure() {
     }
   }
 
+  @Test
+  void testColumnViewWithNonEmptyNullsIsCleared() {
```
abellina, line 6682: could you add comments in this test on what it is doing? specifically what about this is a "non-empty-nulls?

Right now I know you are building a column, but I do not know the significance of say `Optional.of(2L)`. It would be great  to comment on anything important here.

razajafri: I have added comments. I hope this makes it clear what we are doing

java/src/main/java/ai/rapids/cudf/ColumnVector.java
```diff
@@ -119,7 +119,10 @@ public ColumnVector(DType type, long rows, Optional<Long> nullCount,
     incRefCountInternal(true);
   }
 
-  private static OffHeapState makeOffHeap(DType type, long rows, Optional<Long> nullCount,
+  /**
+   * This method is internal and exposed purely for testing purpopses
```
abellina, line 123:
```suggestion
   * This method is internal and exposed purely for testing purposes
```

abellina (approved):
I just had one typo change request, otherwise LGTM

revans2 (approved):

PR #13271, opened 2023-05-02
Numba 0.57 compatibility fixes

gmarkall:
## Description

For compatibility with Numba 0.57 (required for full CUDA 12.0 support) the following changes are necessary:

- We can no longer use `Record` types directly, because Numba 0.57 assumes that it can convert a `Record` back to a NumPy dtype using its internal mechanism. The internal mechanism doesn't recognize cuDF extension types and has no provision for adding support for new types. To work around this, we directly encode a correspondence between our `Row` type (directly derived from a `Record`) and the NumPy `object` dtype. An `object` dtype is acceptable for this use because we use our row types for type checking and inference, but not code generation. Because code generation is not needed, we don't need to have an accurate representation of the low-level layout of a `Row` in the typing.
- A deprecated address space conversion function was removed from Numba. We switch to using another method that performs the address space conversion itself.
- The data layout moved from the `nvvm` module on to an `NVVM` object. Rather than attempting to get this from Numba, we just define it in cuDF, since there is only one possible data layout in CUDA anyway. This allows the same code path to work with all Numba versions (and removes a previous workaround from it having once moved before).

These changes should be compatible with both Numba 0.56 (the minimum required version at present) and 0.57 (can be used, will become a requirement soon).

Tested locally with Numba 0.57.0rc1 (which doesn't differ from the recently-tagged release in any way material to CUDA).

## Checklist


bdice:
@gmarkall (cc: @vyasr @divyegala for pip wheels): I see this PR is compatible with 0.56. Will we need to bump the minimum version to 0.57 for CUDA 12 packages? For now we only want pip packages that support CUDA 12, so the [blockage on the conda-forge release of numba 0.57](https://github.com/conda-forge/numba-feedstock/pull/115#issuecomment-1532308552) is not an issue yet.

python/cudf/cudf/core/udf/utils.py
```diff
@@ -240,6 +240,26 @@ def _construct_signature(frame, return_type, args):
...
+    # mechanism (`numba.np_support.as_dtype). This is not the case for cuDF
+    # extension types that might be the column types (e.g. masked types, string
+    # types or group types).
+    #
+    # We use this type for type inference and type checking, but not in code
+    # generation. For this use case, it is sufficient to provide a dtype for a
+    # row that corresponds to any Python object.
+    @property
+    def dtype(self):
+        return np.dtype("object")
```
vyasr, line 257: Is it at all useful to try the parent's behavior first and see if it gets a useful dtype, then fall back to object if not?

python/cudf/cudf/core/udf/utils.py
```diff
@@ -356,14 +376,20 @@ def _post_process_output_col(col, retty):
     return as_column(col, retty)
 
 
+# The only supported data layout in NVVM.
+# See: https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html?#data-layout
+_nvvm_data_layout = (
+    "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-"
+    "i128:128:128-f32:32:32-f64:64:64-v16:16:16-v32:32:32-"
+    "v64:64:64-v128:128:128-n16:32:64"
+)
```
vyasr, line 385: Why does this need to be hardcoded now? Would the old logic using `nvvm.data_layout` not have picked it up automatically?

python/cudf/cudf/core/udf/strings_lowering.py
```diff
@@ -126,9 +125,8 @@ def cast_string_literal_to_string_view(context, builder, fromty, toty, val):
     sv = cgutils.create_struct_proxy(string_view)(context, builder)
 
     # set the empty strview data pointer to point to the literal value
-    s = context.insert_const_string(builder.module, fromty.literal_value)
-    sv.data = context.insert_addrspace_conv(
-        builder, s, nvvm.ADDRSPACE_CONSTANT
+    sv.data = context.insert_string_const_addrspace(
```
vyasr, line 128: Is this a new function?

gmarkall: It's not, it should really have been used in the first place.

python/cudf/cudf/core/udf/utils.py
gmarkall: There's only one possible datalayout for CUDA, and the place where Numba keeps the datalayout is an implementation detail, so to avoid this breaking underneath us again we should really just keep our own copy of it.

python/cudf/cudf/core/udf/utils.py
gmarkall: I don't think it's a good idea - presently the dtype we provide is not used, but if it starts being used it wouldn't necessarily be correct to use the parent's dtype anyway - this will break more loudly if it breaks, whereas an issue resulting from using a parent-supplied dtype will look more like weird code-generation bugs.

vyasr (approved):

bdice:
> Might it be an issue for our testing / CI though? Won't we need Numba 0.57 in conda-forge to be able to generally develop and test against 0.57?

Yes, you're correct - after merging this, we'll probably apply a patch that _only_ uses numba 0.57 for CUDA 12 wheels until conda-forge has numba 0.57, too. @vyasr @divyegala

bdice (approved):
All my questions about the major changes here were addressed in @vyasr's review and in my comments on the PR as a whole. Thanks so much for this, @gmarkall!

PR #13272, opened 2023-05-02
Fix integer overflow in `partition` `scatter_map` construction

### LINKED ISSUE #7513: [BUG] scatter_by_map does not work correctly
scatter_by_map is supposed to "use map_index to determine the destination of each row of the original DataFrame."
The output should be a list of frames with all the records for a specific index value.
This used to work in earlier version (ex. 0.10)
Now however, the dataframes in the list have multiple different values in the index column

I am having trouble coming up with a small reproducible example.
What I have is a cudf DataFrame with many numeric columns and a single 'id' column:

```
> df.dtypes
vp_m             float64
va_m             float64
vb_m             float64
vc_m             float64
vp_a             float64
va_a             float64
vb_a             float64
vc_a             float64
ip_m             float64
ia_m             float64
ib_m             float64
ic_m             float64
ip_a             float64
ia_a             float64
ib_a             float64
ic_a             float64
f                float64
df               float64
utc       datetime64[ns]
status             int32
id                object
dtype: object
```

I call:
```
id_df_list = df.scatter_by_map(df['id'])
[print(x['id'].unique()) for x in id_df_list]
```
and I get the output of length 42, which is correct in my case, but the assignments are wrong in that each dataframe contains multiple different values in 'id' column.

Output:
```
0    B337
1    B567
2    B608
3    B939
Name: id, dtype: object
0    B193
1    B608
Name: id, dtype: object
0    B161
1    B193
2    B326
3    B703
Name: id, dtype: object
....
```

I am on ubuntu 18, cudf 0.15.
Python 3.8.6
Conda install

wence-:
## Description

Although the input partition map may be any valid integral type, the
intermediate scatter map should have the same type as a valid row
index (and the offsets histogram), namely `size_type`. Previously, the
scatter map was created with the same integer type as the partition
map, which can result in integer overflow, and incorrect results, when
the partition map is a narrow integral type and the input table has
more rows than the width of the type.

On the Python side, this adds validation of the user
partition_map input to ensure that all entries are in range,
avoiding potential out of bounds memory accesses.

Closes #7513.

## Checklist

### DIFF
--- a/python/cudf/cudf/tests/test_sorting.py
+++ b/python/cudf/cudf/tests/test_sorting.py
@@ -386,7 +386,14 @@ def test_dataframe_sort_values_kind(nelem, dtype, kind):
     assert_eq(sorted_df["b"].values, bb[sorted_index])
 
 
-def test_dataframe_scatter_by_map_7513():
-    df = DataFrame({"id": [1, 2, 1, 2], "val": [0, 1, 2, 3]})
+@pytest.mark.parametrize("ids", [[-1, 0, 1, 0], [0, 2, 3, 0]])
+def test_dataframe_scatter_by_map_7513(ids):
+    df = DataFrame({"id": ids, "val": [0, 1, 2, 3]})
     with pytest.raises(ValueError):
         df.scatter_by_map(df["id"])
+
+
+def test_dataframe_scatter_by_map_empty():
+    df = DataFrame({"a": [], "b": []})
+    scattered = df.scatter_by_map(df["a"])
+    assert len(scattered) == 0


davidwendt (approved):
Approving for C++

cpp/src/partitioning/partitioning.cu
```diff
@@ -693,7 +693,7 @@ struct dispatch_map_type {
 
     // Unfortunately need to materialize the scatter map because
     // `detail::scatter` requires multiple passes through the iterator
-    rmm::device_uvector<MapType> scatter_map(partition_map.size(), stream);
+    rmm::device_uvector<size_type> scatter_map(partition_map.size(), stream);
```
ttnghia, line 696: If we can use `size_type`, we can construct a `column_view` from it and use the `scatter` API taking that `column_view` map instead (line 710 below).

ttnghia: Yeah, after changing the header like that then just this:
```
    auto scattered =
-      cudf::detail::scatter(t, scatter_map.begin(), scatter_map.end(), t, stream, mr);
+     cudf::detail::scatter(t, scatter_map, t, stream, mr);
```

cpp/src/partitioning/partitioning.cu
```diff
@@ -16,9 +16,10 @@
 
 #include <cudf/column/column_factories.hpp>
 #include <cudf/copying.hpp>
+#include <cudf/detail/gather.cuh>
```
wence-, line 19: Just to note I need this header now because this file uses `gather_bitmask` which isn't exposed via the `gather.hpp` header, only `gather.cuh`.

ttnghia (approved):

PointKernel (approved):

wence-:
Merged trunk so hopefully the unrelated python test fails will be gone.

ttnghia:
I think the failed python tests are relevant:
```
/opt/conda/envs/test/lib/python3.9/site-packages/cudf/core/dataframe.py:2290: in scatter_by_map
    partitioned_columns, output_offsets = libcudf.partitioning.partition(
/opt/conda/envs/test/lib/python3.9/contextlib.py:79: in inner
    return func(*args, **kwds)
partitioning.pyx:59: in cudf._lib.partitioning.partition
    ???
/opt/conda/envs/test/lib/python3.9/site-packages/cudf/core/scalar.py:279: in __bool__
    return bool(self.value)
....
```

wence-:
> > I think the failed python tests are relevant:
> > ```
> > /opt/conda/envs/test/lib/python3.9/site-packages/cudf/core/dataframe.py:2290: in scatter_by_map
> >     partitioned_columns, output_offsets = libcudf.partitioning.partition(
> > /opt/conda/envs/test/lib/python3.9/contextlib.py:79: in inner
> >     return func(*args, **kwds)
> > partitioning.pyx:59: in cudf._lib.partitioning.partition
> >     ???
> > /opt/conda/envs/test/lib/python3.9/site-packages/cudf/core/scalar.py:279: in __bool__
> >     return bool(self.value)
> > ....
> > ```
> 
> Ah, I didn't think about nulls in my error checking. Will look tomorrow.

OK, should be fixed...

shwina (approved):

Issue #584, opened 2018-08-29
Add clang-tidy for automatic linting


andrewseidl:
## Feature request

In similar spirit of #117, would be good to start using [`clang-tidy`](https://clang.llvm.org/extra/clang-tidy/) as a linter on the codebase to catch some errors, style violations, etc.

Arrow's setup would be a good place to start, which includes a [CMake target](https://github.com/apache/arrow/blob/44c2fa7d7d4e7a1c1645c85f87d557c4fc510c33/cpp/CMakeLists.txt#L552-L560) to run the tool and a [custom version of `run-clang-tidy`](https://github.com/apache/arrow/blob/master/cpp/build-support/run-clang-tidy.sh) which supports ignoring some files.

vyasr:
@jrhemstad @harrism Now that we've merged rapidsai#6695 I think there is interest in enabling clang-tidy on the libcudf code base so I'm reopening this issue. If we'd prefer to consolidate discussion on #479 we can close this again and reopen that, but offhand it seems like #479 grew into a much more involved discussion about defining a style guide and enabling clang-tidy with some minimal configuration seems like a simpler goal to start with.

harrism:
Agree! @teju85 is interested in having a shared cross-rapids clang-tidy config. This makes sense as we have more or less converged on a shared RAPIDS clang-format config.

teju85:
Thank you Mark for adding me to this discussion. Yes, I'm certainly interested in a more uniform tidy config across RAPIDS!

BTW, we already have clang-tidy running as part of CI for cuml for almost a year now (though not super stringent, yet). cuML CI is [here](https://github.com/rapidsai/cuml/blob/branch-21.10/ci/checks/style.sh#L92). However, when it was introduced, we were at clang v8.0.1 which wasn't particularly great with .cu/.cuh files! So, the [tidy wrapper script](https://github.com/rapidsai/cuml/blob/branch-21.10/cpp/scripts/run-clang-tidy.py) had to take in a bunch of hacks :(. If someone can pick up this flow, improve and generalize it for RAPIDS-wide usage, that'd be awesome.

Finally, I'm not particularly attached to these scripts or a particular style (eg: google, cpp-core, etc) and will happily accept if there's a better solution for RAPIDS. However, I'd like for atleast all the key devs to be involved and their requirements heard before finalizing on a tidy style guide. Because in my personal opinion, tidying has a bigger benefit in terms of code uniformity and readability, especially for folks working across RAPIDS projects.

[...]

PR #13275, opened 2023-05-02
Fix unused variables and functions

karthikeyann:
## Description
Removed unused variables and functions.
Removed src/io/utilities/type_conversion.cpp/hpp
Rename some variables which are shadowed by other variables to avoid confusion.

## Checklist


ttnghia (approved):

vuule:
I'm surprised some of these didn't trigger warning-as-error.
How did you catch these?

karthikeyann:
> I'm surprised some of these didn't trigger warning-as-error. How did you catch these?

by running `cppcheck`.

vuule (approved):
nice!

bdice (approved):
@karthikeyann After these changes, is the code “clean” according to cppcheck or are there lots of false positives? Feel free to add a pre-commit hook if it’s relatively clean and not too slow to run. Search cppcheck here, there are several options: https://pre-commit.com/hooks.html

vyasr:
Last I checked none of the available cppcheck hooks actually installed cppcheck, they required it to already be installed on the system (unlike the clang-format wheels that are available). It would be great if that had changed, though.

karthikeyann:
@bdice cppcheck still has many false positives, especially in tests/ code. Not ready now. will check if more fixes and configuration can make it to add as a pre-commit hook. [clang-tidy](https://github.com/rapidsai/cudf/issues/584) is an option too.

@vyasr I installed cppcheck using conda. It's available in pypi https://pypi.org/project/cppcheck-codequality/. If it is pip installable, it can be a pre-commit hook right? (without the need for conda)

bdice:
@vyasr Ah, right. I forgot that cppcheck itself is not pip-installable. Ignore my proposal. @karthikeyann That pip package seems like a utility for converting reports, not `cppcheck` itself.

PR #13277, opened 2023-05-02
Fix parquet schema interpretation issue

### LINKED ISSUE #13237: [BUG] libcudf fails to load old Parquet array encoding properly
**Describe the bug**
Older Parquet files may have arrays encoded in different ways, including a group followed immediately by a repeated primitive rather than the typical group->group list->primitive.  libcudf is failing to load this older encoding properly, loading the data as LIST of LIST of INT32 rather than LIST of INT32.

Attached is a sample file which can be loaded with the following test program:
```c++
#include <iostream>

#include <cudf/io/parquet.hpp>
#include <cudf/lists/lists_column_view.hpp>
#include <cudf/types.hpp>

  auto table = cudf::io::read_parquet(
    cudf::io::parquet_reader_options::builder(cudf::io::source_info("pq891392009.parquet")));
  if (table.tbl->num_columns() != 1) {
    std::cerr << "Incorrect column count" << std::endl;
  }
  auto c0 = table.tbl->get_column(0);
  if (c0.type().id() != cudf::type_id::LIST) {
    std::cerr << "Top level column type mismatch, expected LIST" << std::endl;
    return 1;
  }
  auto lc = cudf::lists_column_view(c0);
  auto child = lc.child();
  if (child.type().id() != cudf::type_id::INT32) {
    std::cerr << "child column type mismatch, expected INT32" << std::endl;
    return 1;
  }
  return 0;
}
```
[pq891392009.parquet.gz](https://github.com/rapidsai/cudf/files/11347506/pq891392009.parquet.gz)

**Expected behavior**
libcudf should load the file as a LIST of INT32 child column.  Both the parquet-cli and Spark CPU can load this file properly.

From parquet-cli:
```
$ java -cp 'target/*:target/dependency/*' org.apache.parquet.cli.Main meta /tmp/dbg/pq891392009.parquet

File path:  /tmp/dbg/pq891392009.parquet
Created by: RAPIDS Spark Plugin
Properties: (none)
Schema:
message spark_schema {
  required group f (LIST) {
    repeated int32 array;
  }
}


Row group 0:  count: 1  43.00 B records  start: 4  total: 43 B
--------------------------------------------------------------------------------
         type      encodings count     avg size   nulls   min / max
f.array  INT32     _   _     2         21.50 B    0       "0" / "1"

$ java -cp 'target/*:target/dependency/*' org.apache.parquet.cli.Main cat /tmp/dbg/pq891392009.parquet
{"f": [0, 1]}
```

and from Spark CPU:
```
scala> spark.read.parquet("/tmp/dbg/pq891392009.parquet").printSchema
root                                                                            
 |-- f: array (nullable = true)
 |    |-- element: integer (containsNull = true)


scala> spark.read.parquet("/tmp/dbg/pq891392009.parquet").show
+------+
|     f|
+------+
|[0, 1]|
+------+
```

### LINKED ISSUE #13239: [BUG] libcudf chunked Parquet reader hangs when loading Parquet with older array encoding
**Describe the bug**
Potentially relates to #13237.  Using the same Parquet test file from #13237, using the chunked Parquet reader to load the file will hang from the caller's perspective, never returning and consuming 100% of a CPU core.

Using the same test file from #13237, the following C++ program will reproduce the hang.
```c++
#include <iostream>

#include <cudf/io/parquet.hpp>
#include <cudf/lists/lists_column_view.hpp>
#include <cudf/types.hpp>

  std::cerr << "Creating reader" << std::endl;
  auto reader = cudf::io::chunked_parquet_reader(1L << 31, 
    cudf::io::parquet_reader_options::builder(cudf::io::source_info("pq891392009.parquet")));
  while (reader.has_next()) {
    std::cerr << "Reading batch" << std::endl;
    auto chunk = reader.read_chunk();
  }
  std::cerr << "Done reading" << std::endl;
  return 0;
}
```

**Expected behavior**
libcudf calls should not hang/crash on valid inputs.

hyperbolic2346:
## Description
There is a bug reading parquet files that have a specific encoding. The typical list is 
```
<list-repetition> group <name> (LIST) {
  repeated group list {
    <element-repetition> <element-type> element;
  }
}
```
but it can also be
```
<list-repetition> group <name> (LIST) {
  repeated group list {
    repeated int32 name
  }
}
```
This second case was failing and this fixes that. The issue was the `one_level_list` was returning true for the name, which resulted in another nesting level. The fix is to only return true if the immediate parent is not a list.

closes #13237
closes #13239

## Checklist


PointKernel (approved):
LGTM

Thanks for the fix

mythrocks:
I've been trying to wrap my head around Parquet's handling for legacy LIST representations as well.

I think we are agreed on the first one:
```
<list-repetition> group <name> (LIST) {
  repeated group list {
    <element-repetition> <element-type> element;
  }
}
```
This conforms to the first section of the [`LogicalType` spec](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#backward-compatibility-rules):
> _It is required that the repeated group of elements is named `list` and that its element field is named `element`._

So this should be interpreted as `LIST<element_type>`.

Could I please have clarification on the second one?
```
<list-repetition> group <name> (LIST) {
  repeated group list {
    repeated int32 name
  }
}
```
The backward compatibility rules are:
```
1. If the repeated field is not a group, then its type is the element type and elements are required.
2. If the repeated field is a group with multiple fields, then its type is the element type and elements are required.
3. If the repeated field is a group with one field and is named either `array` or uses the `LIST`-annotated group's name with `_tuple` appended then the repeated type is the element type and elements are required.
4. Otherwise, the repeated field's type is the element type with the repeated field's repetition.
```
Wouldn't this case fall into the fourth category? In which case, should it not be interpreted as before the change? (i.e. `LIST<LIST<INT32>>`?)

nvdbaranec (approved):

cpp/tests/io/parquet_test.cpp
```diff
@@ -5376,4 +5376,64 @@ TEST_F(ParquetWriterTest, UserNullabilityInvalid)
...
+    0x6b, 0x20, 0x50, 0x6c, 0x75, 0x67, 0x69, 0x6e, 0x19, 0x1c, 0x1c, 0x00, 0x00, 0x00, 0x9f, 0x00,
+    0x00, 0x00, 0x50, 0x41, 0x52, 0x31};
+
+  auto reader = cudf::io::chunked_parquet_reader(
+    1L << 31,
+    cudf::io::parquet_reader_options::builder(
+      cudf::io::source_info{reinterpret_cast<const char*>(list_bytes), sizeof(list_bytes)}));
+  int iterations = 0;
+  while (reader.has_next() && iterations < 10) {
+    auto chunk = reader.read_chunk();
```
mythrocks, line 5434: Hmm. I wonder why the compiler does not issue a warning for this line. Do we not need a `[[nodiscard]]`?
Alternatively:
```c++
(void)reader.read_chunk();
```

cpp/tests/io/parquet_test.cpp
```diff
@@ -5376,4 +5376,64 @@ TEST_F(ParquetWriterTest, UserNullabilityInvalid)
...
+  // read single level list reproducing parquet file
+  cudf::io::parquet_reader_options read_opts = cudf::io::parquet_reader_options::builder(
+    cudf::io::source_info{reinterpret_cast<const char*>(list_bytes), sizeof(list_bytes)});
+  auto table = cudf::io::read_parquet(read_opts);
+
+  auto c0 = table.tbl->get_column(0);
+  EXPECT_TRUE(c0.type().id() == cudf::type_id::LIST);
+
+  auto lc    = cudf::lists_column_view(c0);
+  auto child = lc.child();
```
mythrocks, line 5406: These might be `const`-able, actually.

mythrocks (approved):
:rocket: Thanks for the fix, @hyperbolic2346. LGTM!

shwina (approved):

cpp/tests/io/parquet_test.cpp
mythrocks: I'm resolving this concern. It's really a nitpick.

mythrocks:
Thank you for this fix, @hyperbolic2346!

PR #13260, opened 2023-05-01
Implement a chunked_pack API

abellina:
## Description
This PR introduces a "chunked pack" API built on top of `contiguous_split`. This API is used when we want to copy a cuDF `table_view` over the wire or to the host in a contiguous layout (aka `contiguous_split`), but without user provided memory. As a result this API does not allocate any buffers for GPU data and instead it uses a provided user buffer to perform the `contiguous_split` in chunks. 

Luckily, `contiguous_split` already had a subdivision of work that we are now calling "batches". Each batch is up to 1MB of data from the source table. As such, one can think of this function as copying as many batches as will fit in a user buffer (or "chunking the batches"). The API follows other chunked interfaces in cuDF with a `has_next` and a `next`, with the difference that in this case `next` takes a `device_span`, and the user can provide any `device_span` as long as the size is the same as the size specified during construction.

When thrust and scratch space is required on the GPU, this PR makes use of the memory resource passed, even to the point of using the second argument of `exec_policy` which is the memory resource. I found this while testing that `exec_policy` defaults to the per-device resource, and in this case I really wanted to pass a pooled memory resource, outside of our normal async memory resource to set aside this memory ahead of time.

Most of the changes are about moving things around. You'll see 3 structs with the name "packed" (e.g. `packed_split_indices_and_src_buf_info`). These are here to group together state that `contiguous_split` needs to work, but now because of `chunked_pack` we need to keep around as well for the subsequent calls to `next`. These structs are also packed in memory, which is an optimization contiguous_split had already done to reduce the number of d2h/h2d copies. This PR did need to add a state object (`contiguous_split_state`) that now `contiguous_split` leverages. It also makes use of the `metadata_builder` which we added in a prior PR.

This PR does not include the JNI changes needed for this to work on the java side, I'll post that separately. I figured this was too big already (and if people have suggestions on "chunking" this PR up, I am happy to do that). 

@nvdbaranec spent a great deal of time documenting `contiguous_split` for me and he suggested a path to get this done that I just followed (thank you!!)

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13278, opened 2023-05-02
JNI api for cudf::chunked_pack

abellina:
This PR is the standalone JNI side of https://github.com/rapidsai/cudf/pull/13260. Therefore it doesn't build in as is, but I am putting this up as draft for the Java reviewers to start taking a look.

This implements a `ChunkedPack` java class mirroring the interface of `cudf::chunked_pack`. This is an iterator-like class that can be used to invoke `cudf::pack` (aka `cudf::contigous_split` without splits) over several iterations against a bounce buffer.

In order to create a `ChunkedPack`, the user calls `makeChunkedPack` from a `Table` instance. During this call the user can also pass an `RmmDeviceMemoryResource` to be used internally by `cudf::chunked_pack` exclusively for scratch/temporary data.

## Checklist


java/src/main/java/ai/rapids/cudf/ChunkedPack.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ */
+
+package ai.rapids.cudf;
+
+/**
+ * JNI interface to cudf::chunked_pack.
+ * 
+ * ChunkedPack is an iterator-like interface with the familiar `hasNext` and `next`
+ * interface. `next` should be used in a loop until `hasNext` returns false.
```
revans2, line 24: nit: can we say more like 
```
`ChunkedPack` has an Iterator-like API with the familiar `hasNext` and `next` methods.
```

java/src/main/native/src/PackedColumnMetadataJni.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  // Convert local reference to global so it cannot be garbage collected.
+  Packed_columns_meta_jclass = static_cast<jclass>(env->NewGlobalRef(cls));
+  if (Packed_columns_meta_jclass == nullptr) {
+    return false;
+  }
+  return true;
+}
+
+jobject packed_column_metadata_from(JNIEnv *env, std::unique_ptr<std::vector<uint8_t>> meta) {
```
revans2, line 52: Why do we create a jobect and do all of the reflection/etc if we could just return a jlong to java and let it call the constructor in java?

java/src/main/java/ai/rapids/cudf/ChunkedPack.java
abellina: done

java/src/main/native/src/PackedColumnMetadataJni.cpp
abellina: yeap that is much easier, fixed.

jbrennan333 (commented):
This looks good.

revans2 (approved):

ttnghia (approved):

java/src/main/native/src/CudfJni.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2022, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2023, NVIDIA CORPORATION.
```
revans2, line 2: Is this needed? We are only changing the copyright date???

java/src/main/native/src/cudf_jni_apis.hpp
```diff
@@ -51,7 +51,6 @@ convert_table_for_return(JNIEnv *env, std::unique_ptr<cudf::table> &&table_resul
 //
 
 bool cache_contiguous_table_jni(JNIEnv *env);
-
```
revans2, line 54: why is this change needed?

revans2 (approved):
Just two curious changes that appear to not be needed.

java/src/main/native/src/CudfJni.cpp
abellina: Sorry about that, it is fixed now.

java/src/main/native/src/cudf_jni_apis.hpp
abellina: Fixed

PR #13284, opened 2023-05-03
Update cupy dependency

vyasr:
Switch to requiring cupy 12


jakirkham:
This seems like a good idea

Do we want to update other libraries as well?

AyodeAwe (approved):

shwina (approved):

PR #13285, opened 2023-05-04
Fix approach to detecting assignment for gte/lte operators

### LINKED ISSUE #13185: [BUG] cuDF expression evaluation of some comparison operators creates AssertionError
**Describe the bug**
cuDF expression evaluation created `AssertionError` for less equal, greater equal and not equal operators.

```python
import cudf
df = cudf.DataFrame({'a': 1, 'b': 's'})
df.to_pandas().eval("a>=2")
df.to_pandas().eval("a<=2")
df.to_pandas().eval("a!=2")

#these statements error ou
df.eval("a>=2")
df.eval("a<=2")
df.eval("a!=2")
```

vyasr:
## Description

Resolves #13185 

## Checklist


vyasr:
CC @karthikeyann

python/cudf/cudf/core/dataframe.py
```diff
@@ -7133,12 +7133,12 @@ def eval(self, expr: str, inplace: bool = False, **kwargs):
             )
 
         # Have to use a regex match to avoid capturing "=="
-        includes_assignment = re.search("[^=]=[^=]", expr) is not None
+        includes_assignment = re.search("[^=><]=[^=><]", expr) is not None
```
bdice, line 7136: I’m a little puzzled. Why is the second part of the expression altered? I don’t think you need to catch =< or => here since those aren’t legal gte/lte operators as far as I know.
```suggestion
        includes_assignment = re.search("[^=><]=[^=]", expr) is not None
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -7133,12 +7133,12 @@ def eval(self, expr: str, inplace: bool = False, **kwargs):
...
 
         # Check if there were multiple statements. Filter out empty lines.
         statements = tuple(filter(None, expr.strip().split("\n")))
         if len(statements) > 1 and any(
-            re.search("[^=]=[^=]", st) is None for st in statements
+            re.search("[^=><]=[^=><]", st) is None for st in statements
```
bdice, line 7141: Given that this is repeated and slightly nontrivial, I’d like to use a helper function or a name for the regex pattern.

python/cudf/cudf/core/dataframe.py
vyasr: You're right, this was just from a quick find-and-replace.

python/cudf/cudf/core/dataframe.py
vyasr: Done.

bdice (approved):

Issue #13287, opened 2023-05-04
[FEA] Refactor `transform_lists_of_structs` in `row_operators.cu`


ttnghia:
As discussed in https://github.com/rapidsai/cudf/pull/12953/files#r1176816066, the function `transform_lists_of_structs` in `row_operators.cu` should be either split into two separate functions or simplified its implementation (by using https://github.com/rapidsai/cudf/issues/12932). Since https://github.com/rapidsai/cudf/issues/12932 is no longer relevant, we are left with just one option which is to split it into two separate functions, each processing one table input vs two tables input.

PR #13288, opened 2023-05-04
Refactor `transform_lists_of_structs` in `row_operators.cu`

ttnghia:
This split the functions `transform_lists_of_structs` in `row_operators.cu` into separate functions or simplified their implementation. From one function that can process both cases of having either one or two input columns, now we have two functions, each one processes one case.

Closes https://github.com/rapidsai/cudf/issues/13287.


cpp/src/table/row_operators.cu
```diff
@@ -393,6 +393,83 @@ namespace lexicographic {
...
+ * exactly the same as sorting `input` regardless of the sorting order (ASC or DESC).
+ *
+ * @param input The input column to compute ranks
+ * @param column_null_order The flag indicating how nulls compare to non-null values
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @return The output rank columns
+ */
+auto compute_ranks(column_view const& input,
+                   null_order column_null_order,
+                   rmm::cuda_stream_view stream)
```
bdice, line 461: The functions changed in this file (and perhaps others not changed here) should take an `mr` that is used for the memory they allocate. This is consistent with the [developer guide section on output memory](https://github.com/rapidsai/cudf/blob/branch-23.06/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#output-memory), which includes `detail` APIs.

cpp/src/table/row_operators.cu
```diff
@@ -707,8 +716,39 @@ preprocessed_table::create(table_view const& lhs,
     decompose_structs(rhs, decompose_lists_column::NO, column_order, null_precedence);
 
   // Transform any (nested) lists-of-structs column into lists-of-integers column.
-  auto [transformed_lhs, transformed_rhs_opt, transformed_columns_lhs, transformed_columns_rhs] =
-    transform_lists_of_structs(decomposed_lhs, decomposed_rhs, new_null_precedence_lhs, stream);
+  std::vector<std::unique_ptr<column>> transformed_columns_lhs;
+  std::vector<std::unique_ptr<column>> transformed_columns_rhs;
+  auto const [transformed_lhs,
+              transformed_rhs] = [&,
```
bdice, line 722: Same question about lambda captures here.

bdice (changes requested):
A few requests -- overall this is great and the changes look much cleaner.

cpp/src/table/row_operators.cu
```diff
@@ -672,9 +662,28 @@ std::shared_ptr<preprocessed_table> preprocessed_table::create(
   auto [decomposed_input, new_column_order, new_null_precedence, verticalized_col_depths] =
     decompose_structs(input, decompose_lists_column::NO, column_order, null_precedence);
 
-  // Unused variables are generated for rhs table which is not available here.
-  [[maybe_unused]] auto [transformed_input, unused_0, transformed_columns, unused_1] =
-    transform_lists_of_structs(decomposed_input, std::nullopt, new_null_precedence, stream);
+  // Transform any (nested) lists-of-structs column into lists-of-integers column.
+  std::vector<std::unique_ptr<column>> transformed_columns;
+  auto const transformed_input =
+    [&, &decomposed_input = decomposed_input, &new_null_precedence = new_null_precedence] {
```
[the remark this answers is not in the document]
ttnghia: Unfortunately I have to rely on that to capture the structure binding names. Before C++20, structure binding names are not considered as variables and can't be captured implicitly.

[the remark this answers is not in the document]
bdice: Oh, I see. That's fine then. I'll resolve this.

divyegala (approved):
Thanks for this PR! The code looks in great shape now.

cpp/src/table/row_operators.cu
```diff
@@ -393,6 +393,87 @@ namespace lexicographic {
 
 namespace {
 
+/**
+ * @brief Replace child of the input lists column by a new child column.
+ *
+ * If the input is not sliced, just replace the input child by the new_child.
+ * Otherwise, we have to generate new offsets and replace both offsets/child of the input by the
+ * new ones. This is because the new child was generated by ranking and always has zero
+ * offset thus cannot replace the input child if it is sliced.
```
vyasr, line 402:
```suggestion
 * Otherwise, we have to generate new offsets and replace both offsets/child of the input by the
 * Otherwise, we have to generate new offsets and replace both the offsets and the child of the
 * input by the new ones. This is because the new child was generated by ranking and always
 * has zero offset, so it cannot replace the input child if it is sliced.
```

cpp/src/table/row_operators.cu
```diff
@@ -393,6 +393,87 @@ namespace lexicographic {
...
+    cudf::lists::detail::get_normalized_offsets(lists_column_view{input}, stream, mr));
+  return make_output(out_cols.back()->view(), new_child);
+}
+
+/**
+ * @brief Compute ranks of the input column.
+ *
+ * Dense rank type is used for compute ranking of the input for later lexicographic comparison.
+ *
+ * Consider this example: `input = [ [{0, "a"}, {3, "c"}], [{0, "a"}, {2, "b"}] ]`.
```
vyasr, line 443:
```suggestion
 * `Dense` rank type must be used for compute ranking of the input for later lexicographic comparison.
 *
 * To understand why, consider: `input = [ [{0, "a"}, {3, "c"}], [{0, "a"}, {2, "b"}] ]`.
```

cpp/src/table/row_operators.cu
```diff
@@ -393,6 +393,87 @@ namespace lexicographic {
...
+ * If first rank is used, `transformed_input = [ [0, 3], [1, 2] ]`. Comparing them will lead
+ * to the result row(0) < row(1) which is incorrect.
+ * With dense rank, `transformed_input = [ [0, 2], [0, 1] ]`, producing the correct output for
+ * lexicographic comparison.
+ *
+ * In addition, since the being ranked input column is always a nested child column instead of
```
vyasr, line 449:
```suggestion
 * In addition, since the input column being ranked is always a nested child column instead of
```

cpp/src/table/row_operators.cu
```diff
@@ -393,6 +393,87 @@ namespace lexicographic {
...
+ * top-level column, the column order for ranking should be fixed to the same value
```
vyasr, line 450:
```suggestion
 * a top-level column, the column order for ranking should be fixed to the same value
```

cpp/src/table/row_operators.cu
```diff
@@ -402,126 +483,119 @@ namespace {
...
+  std::vector<std::unique_ptr<column>> out_cols;
+
+  if (input.type().id() == type_id::LIST) {
+    auto const child = cudf::lists_column_view{input}.get_sliced_child(stream);
+
+    // Found a lists-of-structs column.
+    if (child.type().id() == type_id::STRUCT) {
+      out_cols.emplace_back(compute_ranks(child, column_null_order, stream, mr));
+      auto transformed = replace_child(input, out_cols.back()->view(), out_cols, stream, mr);
+      return {std::move(transformed), std::move(out_cols)};
```
vyasr, line 509: If you directly put the `replace_child` call into the pair, you can remove the `std::move` and rely on RVO right?

cpp/src/table/row_operators.cu
```diff
@@ -402,126 +483,119 @@ namespace {
...
+        transform_lists_of_structs(child, column_null_order, stream, mr);
+
+      // Only transform the current column if its child has been transformed.
+      if (out_cols_child.size() > 0) {
+        out_cols.insert(out_cols.end(),
+                        std::make_move_iterator(out_cols_child.begin()),
+                        std::make_move_iterator(out_cols_child.end()));
+        auto transformed = replace_child(input, new_child, out_cols, stream, mr);
+        return {std::move(transformed), std::move(out_cols)};
+      }
```
vyasr, line 524: Same question as above about RVO and move semantics.

cpp/src/table/row_operators.cu
```diff
@@ -402,126 +483,119 @@ namespace {
...
+  // In such situations, lhs.type().id() can still be type_id::STRUCT. However, any
+  // structs-of-lists should be decomposed into empty struct type `Struct<>` before being
+  // processed by this function so we do nothing here.
+
+  // Passthrough: nothing changed.
+  return {input, std::move(out_cols)};
+}
+
+/**
+ * @copydoc transform_lists_of_structs(column_view const&, null_order, rmm::cuda_stream_view)
```
vyasr, line 539: This signature is no longer correct since you've added an `mr` parameter above. However, it also seems like it would be incorrect because the other function has an `input` column view whereas this function has `lhs` and `rhs`, so using copydoc will end up showing 3 input columns.

vyasr (changes requested):
Big improvement here, thanks! Mostly small changes requested.

cpp/src/table/row_operators.cu
```diff
@@ -402,126 +483,119 @@ namespace {
...
  *         during the transformation process
  */
 std::tuple<column_view,
-           std::optional<column_view>,
+           column_view,
            std::vector<std::unique_ptr<column>>,
            std::vector<std::unique_ptr<column>>>
 transform_lists_of_structs(column_view const& lhs,
-                           std::optional<column_view> const& rhs_opt,
+                           column_view const& rhs,
```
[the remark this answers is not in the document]
ttnghia: Yes. Previously we had the same function `transform(lhs, optional rhs)` processing all cases. Now we have two functions `transform(lhs)` and `transform(lhs, rhs)`.

cpp/src/table/row_operators.cu
vyasr: @ttnghia I see that this conversation was resolved but no action was taken. Was that intentional because it doesn't work, or do you have unpushed local changes?

ttnghia: Sorry I made changes but forgot to push :smile:

vyasr: I figured :slightly_smiling_face: no worries

vyasr (approved):

cpp/src/table/row_operators.cu
```diff
@@ -402,126 +484,126 @@ namespace {
...
+
+      auto const ranks =
+        compute_ranks(concatenated_children->view(), column_null_order, stream, mr);
+      auto const ranks_slices = cudf::detail::slice(
+        ranks->view(),
+        {0, child_lhs.size(), child_lhs.size(), child_lhs.size() + child_rhs.size()},
+        stream);
+
+      out_cols_lhs.emplace_back(std::make_unique<column>(ranks_slices.front()));
+      out_cols_rhs.emplace_back(std::make_unique<column>(ranks_slices.back()));
```
davidwendt, line 597: This makes a copy so the stream should be passed
```suggestion
      out_cols_lhs.emplace_back(std::make_unique<column>(ranks_slices.front(), stream));
      out_cols_rhs.emplace_back(std::make_unique<column>(ranks_slices.back(), stream));
```

ttnghia: Oh didn't realize that. Also need to pass `mr`.

bdice (approved):

Issue #13253, opened 2023-04-30
[FEA] Support the `case=` argument to `str.contains(...)`


shwina:
In Pandas `.str.contains()` function accepts a `case=` argument that ignores case when searching for substrings:

```python
>>> s = pd.Series(['a', 'A', 'b'])
>>> s.str.contains('a')
0     True
1    False
2    False
dtype: bool
>>> s.str.contains('a', case=False)
0     True
1     True
2    False
dtype: bool
```

We don't currently support the `case=` argument, but I believe it can easily be supported by calling `lower()` (or alternatively `upper()`) on both the input string Series as well as the substring before the call to `contains()`.

davidwendt:
This approach likely work if `regex=False`. The description does not indicate the intention specifically.
Be careful here with the target input string since some regex character classes are different depending on the case of the flag. Specifically `\d, \s, \w` vs `\D, \S, \W` so changing the case on the regex pattern without accounting for those would give incorrect results. https://docs.rapids.ai/api/libcudf/stable/md_regex.html

PR #13290, opened 2023-05-04
Support the case=False argument to str.contains

shwina:
Closes https://github.com/rapidsai/cudf/issues/13253

### DIFF
--- a/python/cudf/cudf/core/column/string.py
+++ b/python/cudf/cudf/core/column/string.py
@@ -744,7 +744,7 @@ class StringMethods(ColumnMethods):
         4    False
         dtype: bool
 
-        The ``pat`` may also be a list of strings in which case
+        The ``pat`` may also be a sequence of strings in which case
         the individual strings are searched in corresponding rows.
 
         >>> s2 = cudf.Series(['house', 'dog', 'and', '', ''])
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/column/string.py
```diff
@@ -767,22 +765,31 @@ def contains(
...
+
+        if is_scalar(pat):
             if regex:
                 result_col = libstrings.contains_re(self._column, pat, flags)
             else:
-                result_col = libstrings.contains(
-                    self._column, cudf.Scalar(pat, "str")
-                )
+                if case is False:
+                    input_column = libstrings.to_lower(self._column)
```
mroeschke, line 778: Nit: Guessing libstrings doesn't have a `casefold`-like method right? https://docs.python.org/3/library/stdtypes.html#str.casefold

shwina: @davidwendt is (or was this ever) on our radar?

davidwendt: No.

mroeschke (approved):
LGTM

vyasr:
Is this PR waiting on anything?

PR #13292, opened 2023-05-04
Throw error if UNITIALIZED is passed to cudf::state_null_count

davidwendt:
## Description
Changes `cudf::state_null_count` to throw a `std::invalid_argument` error which `mask_state::UNINITIALIZED` is passed. This change is part of removing `UNKNOWN_NULL_COUNT` from cudf/libcudf.

## Checklist

### DIFF
--- a/cpp/src/copying/copy.cpp
+++ b/cpp/src/copying/copy.cpp
@@ -129,7 +129,7 @@ std::unique_ptr<column> allocate_like(column_view const& input,
                                   size,
                                   rmm::device_buffer(size * size_of(input.type()), stream, mr),
                                   detail::create_null_mask(size, allocate_mask, stream, mr),
-                                  state_null_count(allocate_mask, input.size()));
+                                  0);
 }
 
 }  // namespace detail
--- a/cpp/include/cudf/null_mask.hpp
+++ b/cpp/include/cudf/null_mask.hpp
@@ -36,6 +36,8 @@ namespace cudf {
  * @brief Returns the null count for a null mask of the specified `state`
  * representing `size` elements.
  *
+ * @throw std::invalid_argument if state is UNINITIALIZED
+ *
  * @param state The state of the null mask
  * @param size The number of elements represented by the mask
  * @return The count of null elements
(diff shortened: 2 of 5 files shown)


cpp/src/copying/copy.cpp
```diff
@@ -129,7 +129,7 @@ std::unique_ptr<column> allocate_like(column_view const& input,
                                   size,
                                   rmm::device_buffer(size * size_of(input.type()), stream, mr),
                                   detail::create_null_mask(size, allocate_mask, stream, mr),
```
bdice, line 131: Is it still legal for `detail::create_null_mask` to accept `UNINITIALIZED`? I think that this `allocate_like` is permitting `UNINITIALIZED` as a value but then claims the null count is zero.

Ref: https://github.com/davidwendt/cudf/blob/f115335590990267333cf07cf6aa8c35ceaa8645/cpp/src/bitmask/null_mask.cu#L91-L95

bdice (changes requested):
Overall looks great, but I'm not sure if we're handling `UNINITIALIZED` properly in some places now.

cpp/src/copying/copy.cpp
davidwendt: Yes, it is still legal to create an uninitialized mask and then the null-count would be uninitialized as well. Technically the entire output column from `allocate_like` is uninitialized (the data buffer is garbage too) so this is no different than what we have today. The caller still needs to fill in the data and the null-mask/count with real values.

bdice: I looked at the [docs for this function](https://docs.rapids.ai/api/libcudf/nightly/group__column__copy.html#ga735c8dd32ebfbf05638095afa51eba9d). The docs are clear that the data is uninitialized. It would be a bit clearer to note the requirements for setting a null count if the null mask is uninitialized, too.

davidwendt:
> > This change is part of removing UNKNOWN_NULL_COUNT from cudf/libcudf.
> 
> For my own education, what is the idea behind the `UNKNOWN_NULL_COUNT` removal?

Valid question. I neglected to reference the appropriate issue https://github.com/rapidsai/cudf/issues/11968 where we discuss removing this to help with moving to a stream based interface.

ttnghia (approved):

vyasr:
@bdice @PointKernel do you have any changes you want to see here, or is it good to go?

bdice (approved):
Thanks @davidwendt!

PR #13294, opened 2023-05-04
Optionally return compression statistics from ORC and Parquet writers

### LINKED ISSUE #13252: [FEA] Return compression statistics from ORC and Parquet writers
In ORC and Parquet writers, compression is performed separately on many data blocks. If compression fails for some of these blocks, writers output such blocks as uncompressed data, instead of throwing an exception (ORC does throw in some cases).
This may lead to unexpectedly large files. Currently there is no transparency into this process so it's hard for users to understand the resulting file size.

The proposal is to add 
```
struct writer_compression_statistics{
  size_t compressed_num_bytes;
  size_t failed_num_bytes;
  size_t skipped_num_bytes;
  double effective_compression_ratio;
};
```
Users can pass a shared pointer to a `writer_compression_statistics` object as an option to a writer. Once writing is complete, this object will the populated by the writer.

Benefits of this approach over returning `writer_compression_statistics` object directly from `write_xyz`: no API break, and no overhead in the writer when the feature is not used.
```[tasklist]
### Tasks
- [x] Main PR: https://github.com/rapidsai/cudf/pull/13294
- [ ] JNI work
```

vuule:
## Description
Partially address #13252

Adds an option to ORC and Parquet writers to take a shared pointer to a `writer_compression_statistics` objects. This object is then updated with the compression statistics from the written file.
In the case of chunked writing, statistics object is updated after each file. If a writing operation fails, statistics are not modified.
One statistics object can be reused in multiple write calls, but it should not be used concurrently in multiple threads.

## Checklist


cpp/include/cudf/io/orc.hpp
```diff
@@ -648,6 +661,16 @@ class orc_writer_options {
   {
     _user_data = std::move(metadata);
   }
+
+  /**
+   * @brief Sets the pointer to the output compression statistics.
+   *
+   * @param comp_stats Pointer to compression statistics to be updated after writing
+   */
+  void set_compression_statistics(std::shared_ptr<writer_compression_statistics> const& comp_stats)
```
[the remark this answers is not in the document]
vuule: I got tripped up by the shared_ptr, but, yes, we should be able to pass by value + move here as well.

cpp/src/io/comp/statistics.cu
```diff
@@ -0,0 +1,62 @@
...
+    return thrust::transform_reduce(
+      rmm::exec_policy(stream),
+      zipped_begin,
+      zipped_end,
+      [status] __device__(auto tup) {
+        return thrust::get<1>(tup).status == status ? thrust::get<0>(tup).size() : 0;
+      },
+      0ul,
+      thrust::plus<size_t>());
+  };
```
mythrocks, line 54: Super! 
I had assumed that there was a two-sequence version of `transform_reduce()`, like there is for `transform()`.
I'll bear this in mind.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1334,6 +1336,10 @@ void encode_pages(hostdevice_2dvector<gpu::EncColumnChunk>& chunks,
                                 d_chunks_in_batch.flat_view().size_bytes(),
                                 cudaMemcpyDefault,
                                 stream.value()));
+
+  if (comp_stats.has_value()) {
+    comp_stats.value() += collect_compression_statistics(comp_in, comp_res, stream);
+  }
```
mythrocks, line 1342: I'm digging this `operator +=()` business. :]
Question: I'm trying to understand the etiquette for operator usage. Would the following be less readable?
```suggestion
  if (comp_stats) {
    *comp_stats += collect_compression_statistics(comp_in, comp_res, stream);
  }
```

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -2018,6 +2032,14 @@ void writer::impl::init_state()
   std::fill_n(_current_chunk_offset.begin(), _current_chunk_offset.size(), sizeof(file_header_s));
 }
 
+void writer::impl::update_compression_statistics(
+  std::optional<writer_compression_statistics> const& compression_stats)
+{
+  if (compression_stats.has_value() and _compression_statistics != nullptr) {
```
mythrocks, line 2038: Again, for my edumacation, would the following be less readable?
```suggestion
  if (compression_stats and _compression_statistics) {
```

cpp/tests/io/orc_test.cpp
```diff
@@ -1705,4 +1705,87 @@ TEST_F(OrcReaderTest, ZstdMaxCompressionRate)
   CUDF_TEST_EXPECT_TABLES_EQUAL(expected, result.tbl->view());
 }
 
+TEST_F(OrcWriterTest, CompStats)
+{
+  auto table = create_random_fixed_table<int>(1, 100000, true);
+
+  auto stats = std::make_shared<cudf::io::writer_compression_statistics>();
```
mythrocks, line 1712: Nitpick: `const`-able, though that might not really matter in the test.

mythrocks (approved):
Super. I just have a couple of nitpicks, and some clarifying questions.

cpp/include/cudf/io/orc.hpp
```diff
@@ -560,6 +562,17 @@ class orc_writer_options {
     return _user_data;
   }
 
+  /**
+   * @brief Returns a pointer to the user-provided compression statistics.
+   *
+   * @return Compression statistics
+   */
+  [[nodiscard]] std::shared_ptr<writer_compression_statistics> const& get_compression_statistics()
```
ttnghia, line 570: `std::shared_ptr` can be copied.
```suggestion
  [[nodiscard]] std::shared_ptr<writer_compression_statistics> get_compression_statistics()
```

cpp/include/cudf/io/orc.hpp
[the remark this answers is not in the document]
ttnghia: I assume that the statistic is generated internally by the writer, isn't it? If so, why do we need this function?

cpp/include/cudf/io/orc.hpp
```diff
@@ -919,6 +957,17 @@ class chunked_orc_writer_options {
...
```
ttnghia, line 965:
```suggestion
  [[nodiscard]] std::shared_ptr<writer_compression_statistics> get_compression_statistics()
```

cpp/include/cudf/io/parquet.hpp
```diff
@@ -670,6 +672,17 @@ class parquet_writer_options {
    */
   [[nodiscard]] auto get_max_page_fragment_size() const { return _max_page_fragment_size; }
 
+  /**
+   * @brief Returns a pointer to the user-provided compression statistics.
+   *
+   * @return Compression statistics
+   */
+  [[nodiscard]] std::shared_ptr<writer_compression_statistics> const& get_compression_statistics()
```
ttnghia, line 680:
```suggestion
  [[nodiscard]] std::shared_ptr<writer_compression_statistics> get_compression_statistics()
```

cpp/include/cudf/io/parquet.hpp
```diff
@@ -1204,6 +1242,17 @@ class chunked_parquet_writer_options {
...
```
ttnghia, line 1250:
```suggestion
  [[nodiscard]] std::shared_ptr<writer_compression_statistics> get_compression_statistics()
```

cpp/include/cudf/io/types.hpp
```diff
@@ -100,6 +100,104 @@ enum statistics_freq {
...
+   * @brief Returns the compression ratio for the successfully compressed blocks.
+   *
+   * Returns nan if there were no successfully compressed blocks.
+   *
+   * @return double The ratio between the size of the compression inputs and the size of the
+   * compressed output.
+   */
+  [[nodiscard]] auto compression_ratio() const noexcept
+  {
+    return static_cast<double>(num_compressed_bytes()) / _num_compressed_output_bytes;
```
ttnghia, line 191: Handing for div by zero?

cpp/src/io/comp/statistics.cu
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+#include "gpuinflate.hpp"
+
+#include <rmm/exec_policy.hpp>
+#include <thrust/transform_reduce.h>
+
+namespace cudf::io {
+
+writer_compression_statistics collect_compression_statistics(
```
ttnghia, line 24: Why not `shared_ptr` anymore?

cpp/src/io/comp/statistics.cu
```diff
@@ -0,0 +1,62 @@
...
+    [] __device__(auto& res) {
+      return res.status == compression_status::SUCCESS ? res.bytes_written : 0;
+    },
+    0ul,
+    thrust::plus<size_t>());
+
+  auto input_size_with_status = [inputs, results, stream](compression_status status) {
+    auto zipped_begin =
+      thrust::make_zip_iterator(thrust::make_tuple(inputs.begin(), results.begin()));
+    auto zipped_end = thrust::make_zip_iterator(thrust::make_tuple(inputs.end(), results.end()));
```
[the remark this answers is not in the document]
ttnghia:
```suggestion
auto const zipped_end = zipped_begin + input.size();
```

cpp/src/io/orc/orc_gpu.hpp
```diff
@@ -350,21 +351,26 @@ void CompactOrcDataStreams(device_2dspan<StripeStream> strm_desc,
...
-                            CompressionKind compression,
-                            uint32_t comp_blk_size,
-                            uint32_t max_comp_blk_size,
-                            uint32_t comp_block_align,
-                            device_2dspan<StripeStream> strm_desc,
-                            device_2dspan<encoder_chunk_streams> enc_streams,
-                            device_span<compression_result> comp_res,
-                            rmm::cuda_stream_view stream);
+std::optional<writer_compression_statistics> CompressOrcDataStreams(
+  uint8_t* compressed_data,
```
ttnghia, line 363: span?

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1291,16 +1291,18 @@ void CompactOrcDataStreams(device_2dspan<StripeStream> strm_desc,
...
-                            CompressionKind compression,
-                            uint32_t comp_blk_size,
-                            uint32_t max_comp_blk_size,
-                            uint32_t comp_block_align,
-                            device_2dspan<StripeStream> strm_desc,
-                            device_2dspan<encoder_chunk_streams> enc_streams,
-                            device_span<compression_result> comp_res,
-                            rmm::cuda_stream_view stream)
+std::optional<writer_compression_statistics> CompressOrcDataStreams(
+  uint8_t* compressed_data,
```
ttnghia, line 1295: span?

cpp/src/io/orc/writer_impl.hpp
```diff
@@ -324,6 +324,17 @@ class writer::impl {
...
                                 std::vector<StripeInformation>& stripes);
 
+  /**
+   * @brief Update writer-level statistics with data from the current table.
+   *
+   * @param num_rows Number of rows in the current table
+   * @param single_table_stats Statistics data from the current table
+   * @param compression_stats Compression statistics from the current table
+   */
+  void update_statistics(size_type num_rows,
```
ttnghia, line 334: Why add this declaration here? Can we make this a free function?

cpp/src/io/parquet/writer_impl.hpp
```diff
@@ -93,6 +93,14 @@ class writer::impl {
    */
   void init_state();
 
+  /**
+   * @brief Updates writer-level statistics with data from the current table.
+   *
+   * @param compression_stats Optional compression statistics from the current table
+   */
+  void update_compression_statistics(
```
ttnghia, line 101: Similar question, can we make this a free function?

cpp/include/cudf/io/types.hpp
vuule: We let C++ handle it and return a `nan`. I added a note about this behavior in the docs of `compression_ratio`.
The alternative would be an optional, but I think `nan` fits better here.

cpp/src/io/parquet/writer_impl.cu
vuule: I prefer `has_value` as it's more explicit/reads like English.
`value()` vs `*`, don't have a strong preference. I think I just like to make it obvious that the object in question is not a pointer.

cpp/include/cudf/io/orc.hpp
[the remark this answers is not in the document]
vuule: The statistics object is created by the user (see tests). That's why we have this shared lifetime management. Basically users registers an object for the writer to write the statistics into.
FWIW, this also means that a statistics object is reusable :)

cpp/src/io/comp/statistics.cu
vuule: This function creates a `writer_compression_statistics object`, no need to know how its lifetime might be managed after it's created. If we think of this as a factory, we could return a `unique_ptr`, but I don't think it's beneficial here.

cpp/src/io/comp/statistics.cu
```diff
@@ -0,0 +1,62 @@
...
+  device_span<device_span<uint8_t const> const> inputs,
+  device_span<compression_result const> results,
+  rmm::cuda_stream_view stream)
+{
+  // bytes_written on success
+  auto const output_size_successful = thrust::transform_reduce(
+    rmm::exec_policy(stream),
+    results.begin(),
+    results.end(),
+    [] __device__(auto& res) {
```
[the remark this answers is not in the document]
vuule: We could include `const` here, but it wouldn't have any impact because results is a `device_span<compression_result const>` and its elements are always accessed as `const`. So in the current code `auto` is deduced as `compression_result const`. With your change it would be deduced as `compression_result`, but the type of `res` does not change.
Seems like personal preference, let me know what you think.

cpp/src/io/parquet/writer_impl.hpp
vuule: It changes the internal state of the `impl` object, I think it fits better as a member function.

cpp/include/cudf/io/types.hpp
```diff
@@ -100,6 +100,104 @@ enum statistics_freq {
...
+  }
+
+ private:
+  size_t _num_compressed_bytes = 0;  ///< The number of bytes that were successfully compressed
+  size_t _num_failed_bytes     = 0;  ///< The number of bytes that failed to compress
+  size_t _num_skipped_bytes    = 0;  ///< The number of bytes that were skipped during compression
+  size_t _num_compressed_output_bytes = 0;  ///< The number of bytes in the compressed output
```
ttnghia, line 198: Let's add `std::` prefix as it was suggested throughout libcudf.

ttnghia (approved):

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1136,7 +1136,7 @@ __global__ void __launch_bounds__(256)
                            device_span<device_span<uint8_t const>> inputs,
                            device_span<device_span<uint8_t>> outputs,
                            device_span<compression_result> results,
-                           uint8_t* compressed_bfr,
+                           device_span<uint8_t> compressed_bfr,
```
hyperbolic2346, line 1139: 🚀

hyperbolic2346 (approved):
Looks good to me, thanks!

Issue #12422, opened 2022-12-19
[FEA] When libcudf throws exception, it should also print out the stack trace


ttnghia:
Currently, when an exception happens, it will be thrown with an error message. All we can see from it is just something like this:
```
RuntimeError: CUDF failure at:/opt/conda/conda-bld/work/cpp/include/cudf/utilities/type_dispatcher.hpp:531: Invalid type_id.
```

We have no idea which APIs have been called and cause that exception. It would be much much better if we can print out the stack trace like this:
```
 0# void throw_with_trace<std::logic_error>(std::logic_error const&) at ../example/throwing_st.cpp:22
 1# oops(int) at ../example/throwing_st.cpp:38
 2# bar(int) at ../example/throwing_st.cpp:54
 3# foo(int) at ../example/throwing_st.cpp:59
 4# bar(int) at ../example/throwing_st.cpp:49
 5# foo(int) at ../example/throwing_st.cpp:59
 6# main at ../example/throwing_st.cpp:76
 7# 0x00007FAC113BEF45 in /lib/x86_64-linux-gnu/libc.so.6
 8# 0x0000000000402ED9
```
Ref: https://www.boost.org/doc/libs/develop/doc/html/stacktrace/getting_started.html#stacktrace.getting_started.exceptions_with_stacktrace

We can adopt one of the very well-known opensource stack trace libraries such as https://github.com/boostorg/stacktrace or https://github.com/bombela/backward-cpp or write our own stack trace code. I believe that having the ability to print the call stack would significantly benefit our developers.

vyasr:
I'm -1 on using a Boost solution. We ripped it out intentionally in #7932, it's a large dependency and I wouldn't want to make RAPIDS any harder to build than it already is. 

I'm not opposed to providing a stack trace in general, though. A simpler solution for this would be nice. FWIW I adapted a much more hacked together solution in the stream identification library: https://github.com/rapidsai/cudf/blob/branch-23.02/cpp/tests/utilities/identify_stream_usage/identify_stream_usage.cpp#L45.

ttnghia:
That's great. Should we extract that code out and make it a common utility function so `CUDF_EXPECT` can call it? I was afraid that implementing it ourselves would be difficult, didn't realize that we already have it.

vyasr:
[...]

PR #13298, opened 2023-05-05
Add stacktrace into cudf exception types

ttnghia:
This implements stacktrace and adds a stacktrace string into any exception thrown by cudf. By doing so, the exception carries information about where it originated, allowing the downstream application to trace back with much less effort.

Closes https://github.com/rapidsai/cudf/issues/12422.

### Example:
```
#0: cudf/cpp/build/libcudf.so : std::unique_ptr<cudf::column, std::default_delete<cudf::column> > cudf::detail::sorted_order<false>(cudf::table_view, std::vector<cudf::order, std::allocator<cudf::order> > const&, std::vector<cudf::null_order, std::allocator<cudf::null_order> > const&, rmm::cuda_stream_view, rmm::mr::device_memory_resource*)+0x446
#1: cudf/cpp/build/libcudf.so : cudf::detail::sorted_order(cudf::table_view const&, std::vector<cudf::order, std::allocator<cudf::order> > const&, std::vector<cudf::null_order, std::allocator<cudf::null_order> > const&, rmm::cuda_stream_view, rmm::mr::device_memory_resource*)+0x113
#2: cudf/cpp/build/libcudf.so : std::unique_ptr<cudf::column, std::default_delete<cudf::column> > cudf::detail::segmented_sorted_order_common<(cudf::detail::sort_method)1>(cudf::table_view const&, cudf::column_view const&, std::vector<cudf::order, std::allocator<cudf::order> > const&, std::vector<cudf::null_order, std::allocator<cudf::null_order> > const&, rmm::cuda_stream_view, rmm::mr::device_memory_resource*)+0x66e
#3: cudf/cpp/build/libcudf.so : cudf::detail::segmented_sort_by_key(cudf::table_view const&, cudf::table_view const&, cudf::column_view const&, std::vector<cudf::order, std::allocator<cudf::order> > const&, std::vector<cudf::null_order, std::allocator<cudf::null_order> > const&, rmm::cuda_stream_view, rmm::mr::device_memory_resource*)+0x88
#4: cudf/cpp/build/libcudf.so : cudf::segmented_sort_by_key(cudf::table_view const&, cudf::table_view const&, cudf::column_view const&, std::vector<cudf::order, std::allocator<cudf::order> > const&, std::vector<cudf::null_order, std::allocator<cudf::null_order> > const&, rmm::mr::device_memory_resource*)+0xb9
#5: cudf/cpp/build/gtests/SORT_TEST : ()+0xe3027
#6: cudf/cpp/build/lib/libgtest.so.1.13.0 : void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*)+0x8f
#7: cudf/cpp/build/lib/libgtest.so.1.13.0 : testing::Test::Run()+0xd6
#8: cudf/cpp/build/lib/libgtest.so.1.13.0 : testing::TestInfo::Run()+0x195
#9: cudf/cpp/build/lib/libgtest.so.1.13.0 : testing::TestSuite::Run()+0x109
#10: cudf/cpp/build/lib/libgtest.so.1.13.0 : testing::internal::UnitTestImpl::RunAllTests()+0x44f
#11: cudf/cpp/build/lib/libgtest.so.1.13.0 : bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*)+0x87
#12: cudf/cpp/build/lib/libgtest.so.1.13.0 : testing::UnitTest::Run()+0x95
#13: cudf/cpp/build/gtests/SORT_TEST : ()+0xdb08c
#14: /lib/x86_64-linux-gnu/libc.so.6 : ()+0x29d90
#15: /lib/x86_64-linux-gnu/libc.so.6 : __libc_start_main()+0x80
#16: cudf/cpp/build/gtests/SORT_TEST : ()+0xdf3d5
```

### Usage

In order to retrieve a stacktrace with fully human-readable symbols, some compiling options must be adjusted. To make such adjustment convenient and effortless, a new cmake option (`CUDF_BUILD_STACKTRACE_DEBUG`) has been added. Just set this option to `ON` before building cudf and it will be ready to use.

For downstream applications, whenever a cudf-type exception is thrown, it can retrieve the stored stacktrace and do whatever it wants with it. For example:
```
try {
  // cudf API calls
} catch (cudf::logic_error const& e) {
  std::cout << e.what() << std::endl;
  std::cout << e.stacktrace() << std::endl;
  throw e;
} 
// similar with catching other exception types
```

### Follow-up work

The next step would be patching `rmm` to attach stacktrace into `rmm::` exceptions. Doing so will allow debugging various memory exceptions thrown from libcudf using their stacktrace.

### Note:
 * This feature doesn't require libcudf to be built in Debug mode.
 * The flag `CUDF_BUILD_STACKTRACE_DEBUG` should not be turned on in production as it may affect code optimization. Instead, libcudf compiled with that flag turned on should be used only when needed, when debugging cudf throwing exceptions.
 * This flag removes the current optimization flag from compiling (such as `-O2` or `-O3`, if in Release mode) and replaces by `-Og` (optimize for debugging).
 * If this option is not set to `ON`, the stacktrace will not be available. This is to avoid expensive stracktrace retrieval if the throwing exception is expected.


cpp/CMakeLists.txt
```diff
@@ -94,13 +101,17 @@ message(VERBOSE "CUDF: Use a file cache for JIT compiled kernels: ${JITIFY_USE_C
...
+  "CUDF: Replace the current optimization flags by the options '-rdynamic -Og -NDEBUG' (useful for debugging with stacktrace retrieval): ${CUDF_BUILD_STACKTRACE_DEBUG}"
+)
 message(
   VERBOSE
   "CUDF: Disable warnings generated from deprecated declarations: ${DISABLE_DEPRECATION_WARNINGS}"
 )
 message(
   VERBOSE
-  "CUDF: Enable the -lineinfo option for nvcc (useful for cuda-memcheck / profiler: ${CUDA_ENABLE_LINEINFO}"
+  "CUDF: Enable the -lineinfo option for nvcc (useful for cuda-memcheck / profiler: ${CUDA_ENABLE_LINEINFO})"
```
davidwendt, line 114:
```suggestion
  "CUDF: Enable the -lineinfo option for nvcc (useful for cuda-memcheck / profiler): ${CUDA_ENABLE_LINEINFO}"
```

java/src/test/java/ai/rapids/cudf/CudaTest.java
jlowe, line 1: 2023 copyrights

java/src/test/java/ai/rapids/cudf/CudaFatalTest.java
jlowe, line 1: 2023 copyrights

cpp/cmake/Modules/ConfigureCUDA.cmake
```diff
@@ -43,5 +43,16 @@ endif()
...
 if(CMAKE_BUILD_TYPE MATCHES Debug)
   message(VERBOSE "CUDF: Building with debugging flags")
+  list(APPEND CUDF_CXX_FLAGS -rdynamic)
   list(APPEND CUDF_CUDA_FLAGS -Xcompiler=-rdynamic)
 endif()
+
+if(CUDF_BUILD_STACKTRACE_DEBUG)
+  message(VERBOSE "CUDF: Building with flags '-rdynamic -Og -NDEBUG'")
+  set(CMAKE_CXX_FLAGS_RELEASE "") # Remove the current optimization level
+  set(CMAKE_CUDA_FLAGS_RELEASE "") # Remove the current optimization level
```
bdice, line 53: Are there other flags not related to optimization that this could be overwriting?

bdice (commented):
Minor suggestions. This looks helpful for debugging.

cpp/src/utilities/stacktrace.cpp
```diff
@@ -0,0 +1,97 @@
...
+  constexpr int max_stack_depth = 64;
+  void* stack[max_stack_depth];
+
+  auto const depth   = backtrace(stack, max_stack_depth);
+  auto const modules = backtrace_symbols(stack, depth);
+
+  if (modules == nullptr) { return "No stacktrace could be captured!"; }
+
+  std::stringstream ss;
+  size_t func_name_size = 256;
```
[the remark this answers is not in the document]
ttnghia: That variable will be modified later in `__cxa_demangle` (line 72 below).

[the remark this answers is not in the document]
bdice: This is awkward because the value is not used after modification. According to https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.3/a01696.html, you can pass `nullptr` and it won't write anything. So make this `constexpr` and pass `nullptr` instead of `&func_name_size` below.

cpp/src/utilities/stacktrace.cpp
```diff
@@ -0,0 +1,82 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cudf/detail/utilities/stacktrace.hpp>
+
+#include <cxxabi.h>
+#include <execinfo.h>
+#include <stdlib.h>
+#include <string.h>
```
davidwendt, line 22:
```suggestion
#include <cstdlib>
#include <cstring>
```

cpp/src/utilities/stacktrace.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  // Skip one more depth to avoid including the stackframe of this function.
+  auto const skip_depth = 1 + (capture_last_frame == capture_last_stackframe::YES ? 0 : 1);
+  for (auto i = skip_depth; i < depth; ++i) {
+    // Each modules[i] string contains a mangled name in the format like following:
+    // `module_name(function_name+0x012) [0x01234567890a]`
+    // We need to extract function name and function offset.
+    char* begin_func_name   = strstr(modules[i], "(");
+    char* begin_func_offset = strstr(modules[i], "+");
+    char* end_func_offset   = strstr(modules[i], ")");
```
davidwendt, line 49:
```suggestion
    char* begin_func_name   = std::strstr(modules[i], "(");
    char* begin_func_offset = std::strstr(modules[i], "+");
    char* end_func_offset   = std::strstr(modules[i], ")");
```

cpp/cmake/Modules/ConfigureCUDA.cmake
```diff
@@ -43,5 +43,16 @@ endif()
...
+  list(APPEND CUDF_CXX_FLAGS -rdynamic -Og)
+  list(APPEND CUDF_CUDA_FLAGS -Xcompiler=-rdynamic -Xcompiler=-Og)
+  list(APPEND CUDF_CXX_DEFINITIONS NDEBUG)
```
robertmaynard, line 56: `NDEBUG` is already defined when in release mode via the default flags, lets not go with this approach as it also has problem when a user does a debug build as it will remove asserts from working.

ajschmidt8 (approved):

cpp/include/cudf/utilities/error.hpp
```diff
@@ -29,13 +31,35 @@ namespace cudf {
...
+  std::string const _stacktrace;  //!< The whole stacktrace stored as one string.
+};
+
 /**
  * @brief Exception thrown when logical precondition is violated.
  *
  * This exception should not be thrown directly and is instead thrown by the
  * CUDF_EXPECTS macro.
  */
-struct logic_error : public std::logic_error {
```
jrhemstad, line 38: Folks should be aware that this is changing the size of the exception type even when compiling in non-stacktrace mode.

That isn't to say it's a good idea to _not_ inherit from `stacktrace_recorder` in non-stacktrace mode because there would be the possibility for weird ABI issues with different builds of libcudf having different sizes for exception types. 

The ideal way to do this without changing the size of the exception type would be to include the stack trace in the `what()` string such that there is only a single `string` member.

ttnghia: The "what()" string is accessed for logging. If we extend it, it may be very large for downstream applications to print out every time. This issue has been mentioned before, please see https://github.com/rapidsai/cudf/issues/12422#issuecomment-1371387818.

ttnghia: There is another solution for this concern: Only modify the exception structs with compiler check `#ifdef CUDF_BUILD_STACKTRACE_DEBUG`. The `char* stacktrace()` will always be an empty string if that option is not defined. How do you think about it?

cpp/include/cudf/utilities/error.hpp
```diff
@@ -29,13 +31,35 @@ namespace cudf {
...
+struct stacktrace_recorder {
+  stacktrace_recorder()
+    // Exclude the current stackframe, as it is this constructor.
+    : _stacktrace{cudf::detail::get_stacktrace(cudf::detail::capture_last_stackframe::NO)}
+  {
+  }
+
+ public:
+  /**
+   * @brief Get the stored stacktrace captured during object construction
```
ttnghia, line 46:
```suggestion
   * @brief Get the stored stacktrace captured during object construction.
```

robertmaynard (approved):
CMake changes look good

robertmaynard (approved):
CMake changes look good

cpp/include/cudf/utilities/error.hpp
davidwendt: Does the recorder need to be part of the signature? 
The stack trace does not seem to have any real association with the exception object.
Could it not just be free function used like this?

```
try {
  // cudf API calls
} catch (cudf::logic_error const& e) {
  std::cout << e.what() << std::endl;
  std::cout << stacktrace() << std::endl;
  throw e;
} 
```

ttnghia: Unfortunately not. `stacktrace()` gives you the call stack up to the point inside `stacktrace()`. From there, you cannot trace back to where the exception was created.

harrism:
> CMake changes look good

**REALLY** good.

cpp/include/cudf/utilities/error.hpp
vyasr: An alternative would be for each exception type to own a `stacktrace_recorder` that is constructed in its constructor. That approach would require ignoring 2 frames instead of just 1, though, and it would be more verbose. I don't have a strong opinion that that is a necessary change here.

cpp/src/utilities/stacktrace.cpp
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  free(modules);
+
+  return ss.str();
+#else
+#ifdef CUDF_BUILD_STACKTRACE_DEBUG
+  return "Stacktrace is only supported when built with a GNU compiler.";
+#else
+  return "libcudf was not build with stacktrace support.";
```
vyasr, line 83:
```suggestion
  return "libcudf was not built with stacktrace support.";
```

vyasr (changes requested):
Apologies for taking so long to review this. The changes look solid, just a couple of small requests. Can we insert a call to `get_stacktrace` in [`stream_checking_resource_adaptor::verify_stream`](https://github.com/rapidsai/cudf/blob/branch-23.08/cpp/include/cudf_test/stream_checking_resource_adaptor.hpp#L164)? I had originally wanted to extract this function anyway for that purpose since I originally implemented the trace to help find improper stream usage.

vyasr (approved):

cpp/CMakeLists.txt
```diff
@@ -94,13 +101,17 @@ message(VERBOSE "CUDF: Use a file cache for JIT compiled kernels: ${JITIFY_USE_C
 message(VERBOSE "CUDF: Build and statically link Arrow libraries: ${CUDF_USE_ARROW_STATIC}")
 message(VERBOSE "CUDF: Build and enable S3 filesystem support for Arrow: ${CUDF_ENABLE_ARROW_S3}")
 message(VERBOSE "CUDF: Build with per-thread default stream: ${CUDF_USE_PER_THREAD_DEFAULT_STREAM}")
+message(
+  VERBOSE
+  "CUDF: Replace the current optimization flags by the options '-rdynamic -Og -NDEBUG' (useful for debugging with stacktrace retrieval): ${CUDF_BUILD_STACKTRACE_DEBUG}"
```
ttnghia, line 106:
```suggestion
  "CUDF: Replace the current optimization flags by the options '-rdynamic -Og' (useful for debugging with stacktrace retrieval): ${CUDF_BUILD_STACKTRACE_DEBUG}"
```

java/src/main/native/include/jni_utils.hpp
```diff
@@ -843,18 +879,14 @@ inline void jni_cuda_check(JNIEnv *const env, cudaError_t cuda_status) {
...
       return ret_val;                                                                              \
     }                                                                                              \
-  }
-
-#define CATCH_STD_CLASS(env, class_name, ret_val)                                                  \
-  catch (const rmm::out_of_memory &e) {                                                            \
     auto what =                                                                                    \
         std::string("Could not allocate native memory: ") + (e.what() == nullptr ? "" : e.what()); \
-    JNI_CHECK_THROW_NEW(env, cudf::jni::OOM_CLASS, what.c_str(), ret_val);                         \
+    JNI_THROW_NEW(env, cudf::jni::OOM_CLASS, what.c_str(), ret_val);                               \
```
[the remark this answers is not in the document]
ttnghia: Yes, this macro `JNI_THROW_NEW` doesn't do the check thus I explicitly added that check into this context (line 884-886) above.

java/src/main/native/include/jni_utils.hpp
```diff
@@ -877,7 +910,11 @@ inline void jni_cuda_check(JNIEnv *const env, cudaError_t cuda_status) {
       JNI_CHECK_CUDA_ERROR(env, cudf::jni::CUDA_FATAL_ERROR_CLASS, cuda_error, ret_val);           \
```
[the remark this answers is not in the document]
ttnghia: Good point. I'll change that.

java/src/main/native/include/jni_utils.hpp
```diff
@@ -863,21 +895,26 @@ inline void jni_cuda_check(JNIEnv *const env, cudaError_t cuda_status) {
...
     /* Like cudf::detail::throw_cuda_error, it is nearly certain that a fatal error  */            \
     /* occurred if the second call doesn't return with cudaSuccess. */                             \
     cudaGetLastError();                                                                            \
     auto const last = cudaFree(0);                                                                 \
     if (cudaSuccess != last && last == cudaDeviceSynchronize()) {                                  \
-      auto msg = e.what() == nullptr ? std::string{""} : e.what();                                 \
-      auto cuda_error = cudf::fatal_cuda_error{msg, last};                                         \
-      JNI_CHECK_CUDA_ERROR(env, cudf::jni::CUDA_FATAL_ERROR_CLASS, cuda_error, ret_val);           \
+      /* Throw CudaFatalException since the thrown exception is unrecoverable CUDA error */        \
+      ex_class = cudf::jni::CUDA_FATAL_ERROR_CLASS;                                                \
```
ttnghia, line 910: Oh no, I was wrong with this. Should call `JNI_CHECK_CUDA_ERROR` to record the cuda error code. Working on this now...

ttnghia: Done.

jlowe (approved):

Issue #13024, opened 2023-03-28
[FEA] Performance issue with the Parquet reader for very large schemas (especially when containing strings)


nvdbaranec:
For parquet files that contain very large schemas with strings (either large numbers of columns, or large numbers of nested columns) we pay a very heavy price postprocessing the string data after the core decode kernels runs.  

Essentially, the "decode" process for strings is just emitting a large array of pointer/size pairs that are then passed to other cudf functions to reconstruct actual columns.    The problem is that we are doing this with no batching - each output string column results in an entire cudf function call (`make_strings_column`) with multiple internal kernel calls each.  In situations with thousands of columns, this gets very expensive.

![image](https://user-images.githubusercontent.com/56695930/228312025-67ea3177-9d67-4e84-84e5-cb317c895001.png)

In the image above, the green span represents the time spent in the decode kernel and the time spent in all of the `make_strings_column` calls afterwards.  The time is totally dominated by the many many calls to `make_strings_column` (the red span).  

Ideally, we would have some kind of batched interface to `make_strings_column`  (`make_strings_columns` ?) that can do the work for the thousands of output columns coalesced into fewer kernels.  

On a related note, the area under the blue line represents a similar problem involving preprocessing the file (thousands of calls to `thrust::reduce` and `thrust::exclusive_scan_by_key`).   This has been largely addressed by this PR  https://github.com/rapidsai/cudf/pull/12931

etseidl:
@nvdbaranec I'm curious if #13302 had any impact on the file used to generate the profile above.

GregoryKimball:
@vuule recently conducted some experiments using an internal stream pool to hide latencies during column buffer allocation. Perhaps evaluating string types with larger column counts would show a bigger signal.

mhaseeb123:
Investigated this issue by profiling the overhead of many `make_strings_columns` on `read_parquet` using a custom NVTX range. Observed that 1024 calls of `make_strings_column`s take roughly ~1.5ms to complete in the current state which seems negligible compared to the decode time (roughly 4ms for table size of 64MB across 1024 cols, 17ms for table size of 1GB across 1024 string cols and 178ms for table size of 12GB across 1024 string cols).

![Image](https://github.com/user-attachments/assets/600498d4-1c6c-400e-b57f-95a6ca2b4161)

[...]

PR #13302, opened 2023-05-05
Write string data directly to column_buffer in Parquet reader

etseidl:
## Description

The current Parquet reader decodes string data into a list of {ptr, length} tuples, which are then used in a gather step by `make_strings_column`. This gather step can be time consuming, especially when there are a large number of string columns. This PR addresses this by changing the decode step to write char and offset data directly to the `column_buffer`, which can then be used directly, bypassing the gather step.

The image below compares the new approach to the old. The green arc at the top (82ms) is `gpuDecodePageData`, and the red arc (252ms) is the time spent in `make_strings_column`.  The green arc below (25ms) is `gpuDecodePageData`, the amber arc (22ms) is a new kernel that computes string sizes for each page, and the magenta arc (106ms) is the kernel that decodes string columns.
flat_edited

NVbench shows a good speed up for strings as well.  There is a jump in time for the INTEGRAL benchmark, but little to no change for other data types.  The INTEGRAL time seems to be affected by extra time spent in `malloc` allocating host memory for a `hostdevice_vector`. This `malloc` always occurs, but for some reason in this branch it takes much longer to return.

This is comparing to @nvdbaranec's branch for #13203.
```
|  data_type  |      io       |  cardinality  |  run_length  |   Ref Time |   Cmp Time |        Diff |   %Diff |  
|-------------|---------------|---------------|--------------|------------|------------|-------------|---------| 
|  INTEGRAL   | DEVICE_BUFFER |       0       |      1       |  14.288 ms |  14.729 ms |  440.423 us |   3.08% |   
|  INTEGRAL   | DEVICE_BUFFER |     1000      |      1       |  13.397 ms |  13.997 ms |  600.596 us |   4.48% |   
|  INTEGRAL   | DEVICE_BUFFER |       0       |      32      |  11.831 ms |  12.354 ms |  522.485 us |   4.42% |   
|  INTEGRAL   | DEVICE_BUFFER |     1000      |      32      |  11.335 ms |  11.854 ms |  518.791 us |   4.58% |   
|    FLOAT    | DEVICE_BUFFER |       0       |      1       |   8.681 ms |   8.715 ms |   34.846 us |   0.40% |   
|    FLOAT    | DEVICE_BUFFER |     1000      |      1       |   8.473 ms |   8.472 ms |   -0.680 us |  -0.01% |   
|    FLOAT    | DEVICE_BUFFER |       0       |      32      |   7.217 ms |   7.192 ms |  -25.311 us |  -0.35% |   
|    FLOAT    | DEVICE_BUFFER |     1000      |      32      |   7.425 ms |   7.422 ms |   -3.162 us |  -0.04% |   
|   STRING    | DEVICE_BUFFER |       0       |      1       |  50.079 ms |  42.566 ms |-7513.004 us | -15.00% |   
|   STRING    | DEVICE_BUFFER |     1000      |      1       |  16.813 ms |  14.989 ms |-1823.660 us | -10.85% |   
|   STRING    | DEVICE_BUFFER |       0       |      32      |  49.875 ms |  42.443 ms |-7432.718 us | -14.90% |   
|   STRING    | DEVICE_BUFFER |     1000      |      32      |  15.312 ms |  13.953 ms |-1358.910 us |  -8.87% |   
|    LIST     | DEVICE_BUFFER |       0       |      1       |  80.303 ms |  80.688 ms |  385.916 us |   0.48% |   
|    LIST     | DEVICE_BUFFER |     1000      |      1       |  71.921 ms |  72.356 ms |  435.153 us |   0.61% |   
|    LIST     | DEVICE_BUFFER |       0       |      32      |  61.658 ms |  62.129 ms |  471.022 us |   0.76% |   
|    LIST     | DEVICE_BUFFER |     1000      |      32      |  63.086 ms |  63.371 ms |  285.608 us |   0.45% |   
|   STRUCT    | DEVICE_BUFFER |       0       |      1       |  66.272 ms |  61.142 ms |-5130.639 us |  -7.74% |   
|   STRUCT    | DEVICE_BUFFER |     1000      |      1       |  40.217 ms |  39.328 ms | -888.781 us |  -2.21% |   
|   STRUCT    | DEVICE_BUFFER |       0       |      32      |  63.660 ms |  58.837 ms |-4822.647 us |  -7.58% |   
|   STRUCT    | DEVICE_BUFFER |     1000      |      32      |  38.080 ms |  37.104 ms | -976.133 us |  -2.56% | 
```

May address #13024 

~Depends on #13203~

## Checklist


etseidl:
The intention behind the changes to `column_buffer` was to not require code changes to the other readers, while not adding another field for the char data. This approach may be a bit too "extra" and could be simplified by accepting the slight overhead of unused fields and controlling behavior with template args.

etseidl:
Future enhancements would include combining the string size computations introduced here with those performed when doing chunked reads/nested schemas, and perhaps overlapping the string decode kernel with the decode kernel for other types.

I'd also note that this change will be beneficial to implementing decoders for DELTA_LENGTH_BYTE_ARRAY and DELTA_BYTE_ARRAY (see #12948).

etseidl:
Here's what I'm talking about with the nvbench results. In both windows I've highlighted the `cudaMallocHost` that results from the instantiation of the `page_nesting_decode_info` (`hostdevice_vector<gpu::PageNestingDecodeInfo>`) in `reader::impl::allocate_nesting_info`. I'm totally baffled as to what is going on here. The same behavior is exhibited in each iteration of the benchmark, but doesn't happen for other data types.

Screenshot from 2023-05-10 15-51-11

etseidl:
@nvdbaranec and @vuule could you please give this a look? :pray:

vuule:
Oh dear, I thought this PR was large because of its dependency. How wrong I was.

etseidl:
> Oh dear, I thought this PR was large because of its dependency. How wrong I was.

Sorry. In my defense a lot of it is just refactoring moving shared functions to a new file.

vuule (commented):
Flushing the minor comments so far. Still hung up on `column_buffer` design, feels like there's a cleaner option, just can't quite figure it out.

cpp/src/io/parquet/page_data.cu
```diff
@@ -2001,12 +823,14 @@ __global__ void __launch_bounds__(decode_block_size) gpuDecodePageData(
...
 
+  // string cols handled elsewhere
+  if (is_string_col(pages[blockIdx.x], chunks)) { return; }
+
   page_state_s* const s          = &state_g;
   page_state_buffers_s* const sb = &state_buffers;
   int page_idx                   = blockIdx.x;
   int t                          = threadIdx.x;
   int out_thread0;
-  [[maybe_unused]] null_count_back_copier _{s, t};
```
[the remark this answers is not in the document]
etseidl: I'll have to revisit this, but I was getting strange errors with this in place that away when I switched to invoking directly. Maybe something to do with two kernels running at the same time that both hit that destructor.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -63,6 +97,9 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
     auto data                  = chunk_nested_data.host_ptr(chunk_off);
     chunks[c].column_data_base = chunk_nested_data.device_ptr(chunk_off);
 
+    auto str_data                = chunk_nested_str_data.host_ptr(chunk_off);
+    chunks[c].column_string_base = chunk_nested_str_data.device_ptr(chunk_off);
```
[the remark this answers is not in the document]
etseidl: I can give that a try. I still want to figure out some clever way to not need to copy pages back and forth at the top where I'm calculating the string buffer sizes.

cpp/src/io/parquet/page_data.cu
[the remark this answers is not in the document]
etseidl: I put it back...it seems to be working again.

cpp/src/io/parquet/page_data.cu
```diff
@@ -47,395 +27,6 @@ namespace gpu {
...
-      __shared__ typename WarpReduce::TempStorage temp_storage;
-      // note: str_len will only be valid on thread 0.
-      str_len += WarpReduce(temp_storage).Sum(len);
-    }
-
-    pos += batch_len;
-  }
-  return {pos, str_len};
-}
-
```
ttnghia, line 438: This file still contains `gpuDecodeRleBooleans`. So you extract some `gpuDecode*` functions out of here but leave some remaining. Please either extract all.

In addition, you also should consider adding `page_decode.cu` and leave only the declaration of these functions in a `page_decode.hpp` header instead. Including the full function definition in the header will duplicate code and bloat the compiled binary.

etseidl: The moved functions are all device functions.  My (probably out of date) understanding is that absent the use of the `-dc` compiler option, you can't call device functions in a separate compilation unit. I tried doing this initially and got 'Unresolved extern function' errors during compilation. If there is a way to do this please point me to an example and I'll happily convert.

I only moved the functions used by both `page_data.cu` and `page_string_decode.cu` to limit the bloat you reference.

ttnghia: > you can't call device functions in a separate compilation unit

Oh sorry I forgot about that. But then we still need to move all them out for consistency. Code bloats only if you call an inline function multiple times. Here we have to do that due to the restriction of device code mentioned above so there should be no difference.

cpp/src/io/parquet/reader_impl.cpp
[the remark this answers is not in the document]
vuule: I only see one H2D copy of `chunk_nested_str_data`. Where is the data copied back?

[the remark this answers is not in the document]
etseidl: it isn't.  `chunk_nested_str_data` is just pointers to device memory for each page.  After the pointers are set, they aren't used on the host.

The round trip I was talking about is up around line 59.

cpp/src/io/utilities/column_buffer.hpp
```diff
@@ -108,25 +85,111 @@ struct column_buffer {
...
+
+  rmm::device_buffer _string_data{};
+};
+
+/**
+ * @brief Class for holding device memory buffers to column data that eventually
+ * will be used to create a column.
+ */
+template <class string_policy>
+struct column_buffer : string_policy {
```
vuule, line 157: I think we can remove the need to separate column_buffer and column_buffer_base classes with CRTP. If I understand the purpose of current column_buffer correctly, this should work:
```
template <typename string_policy>
class column_buffer {
// implement non-string stuff

// I think SFINAE can be used here to compile-time disable some members
void string_data() {
        static_cast<T*>(this)->string_data();
}
...

std::vector<column_buffer<string_policy>> children;
};

class column_buffer_with_pointers : column_buffer<column_buffer_with_pointers>

**Review by vuule — COMMENTED** [review]
Few comments on the pattern used for `column_buffer`.
As discussed offline, I might experiment with this and share the code instead of us going back and forth with my nitpicky ideas.

File: cpp/src/io/utilities/column_buffer.hpp (hunk 4 above)
Comment (etseidl) [reply cpp/src/io/utilities/column_buffer.hpp:157]: Curious how you keep repeating your desire for CRTP :rofl: If you don't get to it this week I'll give it a try in a bit (might get pulled onto other work today).

Comment (vuule) [reply cpp/src/io/utilities/column_buffer.hpp:157]: I'll try to make a commit locally today.

**Comment by etseidl** [comment]
Latest nvbench after reworking offsets calculation and moving the column size calculation to the device.  This is comparing to the latest 23.06.
```
|  data_type  |      io       |  cardinality  |  run_length  |   Ref Time |   Cmp Time |         Diff |   %Diff |  
|-------------|---------------|---------------|--------------|------------|------------|--------------|---------|
|  INTEGRAL   | DEVICE_BUFFER |       0       |      1       |  14.582 ms |  15.141 ms |   558.786 us |   3.83% |   
|  INTEGRAL   | DEVICE_BUFFER |     1000      |      1       |  13.916 ms |  14.272 ms |   356.535 us |   2.56% |   
|  INTEGRAL   | DEVICE_BUFFER |       0       |      32      |  12.381 ms |  12.688 ms |   306.657 us |   2.48% |   
|  INTEGRAL   | DEVICE_BUFFER |     1000      |      32      |  11.941 ms |  12.390 ms |   449.714 us |   3.77% |   
|    FLOAT    | DEVICE_BUFFER |       0       |      1       |   8.943 ms |   8.914 ms |   -28.594 us |  -0.32% |   
|    FLOAT    | DEVICE_BUFFER |     1000      |      1       |   8.771 ms |   8.850 ms |    78.382 us |   0.89% |   
|    FLOAT    | DEVICE_BUFFER |       0       |      32      |   7.442 ms |   7.565 ms |   123.829 us |   1.66% |   
|    FLOAT    | DEVICE_BUFFER |     1000      |      32      |   7.637 ms |   7.709 ms |    72.101 us |   0.94% |
|   DECIMAL   | DEVICE_BUFFER |       0       |      1       |  19.061 ms |  19.108 ms |    47.267 us |   0.25% |   
|   DECIMAL   | DEVICE_BUFFER |     1000      |      1       |   7.575 ms |   7.614 ms |    39.594 us |   0.52% |   
|   DECIMAL   | DEVICE_BUFFER |       0       |      32      |   9.294 ms |   9.277 ms |   -16.626 us |  -0.18% |   
|   DECIMAL   | DEVICE_BUFFER |     1000      |      32      |   6.298 ms |   6.339 ms |    41.699 us |   0.66% |   
|  TIMESTAMP  | DEVICE_BUFFER |       0       |      1       |  25.766 ms |  25.820 ms |    54.299 us |   0.21% |   
|  TIMESTAMP  | DEVICE_BUFFER |     1000      |      1       |   8.130 ms |   8.188 ms |    58.665 us |   0.72% |   
|  TIMESTAMP  | DEVICE_BUFFER |       0       |      32      |  10.680 ms |  10.712 ms |    32.218 us |   0.30% |   
|  TIMESTAMP  | DEVICE_BUFFER |     1000      |      32      |   7.089 ms |   7.139 ms |    49.849 us |   0.70% |   
|  DURATION   | DEVICE_BUFFER |       0       |      1       |  10.691 ms |  10.725 ms |    33.323 us |   0.31% |   
|  DURATION   | DEVICE_BUFFER |     1000      |      1       |   7.823 ms |   7.888 ms |    64.737 us |   0.83% |   
|  DURATION   | DEVICE_BUFFER |       0       |      32      |   7.834 ms |   7.878 ms |    44.310 us |   0.57% |   
|  DURATION   | DEVICE_BUFFER |     1000      |      32      |   6.831 ms |   6.881 ms |    50.430 us |   0.74% |   
|   STRING    | DEVICE_BUFFER |       0       |      1       |  50.794 ms |  41.035 ms | -9759.577 us | -19.21% |   
|   STRING    | DEVICE_BUFFER |     1000      |      1       |  17.030 ms |  13.426 ms | -3603.722 us | -21.16% |   
|   STRING    | DEVICE_BUFFER |       0       |      32      |  50.821 ms |  41.117 ms | -9704.550 us | -19.10% |   
|   STRING    | DEVICE_BUFFER |     1000      |      32      |  15.429 ms |  12.609 ms | -2819.955 us | -18.28% |   
|    LIST     | DEVICE_BUFFER |       0       |      1       |  80.453 ms |  80.650 ms |   196.955 us |   0.24% |   
|    LIST     | DEVICE_BUFFER |     1000      |      1       |  72.462 ms |  72.608 ms |   145.647 us |   0.20% |   
|    LIST     | DEVICE_BUFFER |       0       |      32      |  62.211 ms |  62.328 ms |   116.965 us |   0.19% |   
|    LIST     | DEVICE_BUFFER |     1000      |      32      |  63.325 ms |  63.394 ms |    69.201 us |   0.11% |   
|   STRUCT    | DEVICE_BUFFER |       0       |      1       |  63.418 ms |  56.766 ms | -6651.363 us | -10.49% |   
|   STRUCT    | DEVICE_BUFFER |     1000      |      1       |  37.288 ms |  34.966 ms | -2321.975 us |  -6.23% |   
|   STRUCT    | DEVICE_BUFFER |       0       |      32      |  61.149 ms |  54.457 ms | -6691.907 us | -10.94% |   
|   STRUCT    | DEVICE_BUFFER |     1000      |      32      |  35.232 ms |  32.623 ms | -2608.674 us |  -7.40% |   
```

File: cpp/src/io/parquet/reader_impl.cpp [hunk 5]
```diff
@@ -231,7 +271,7 @@ reader::impl::impl(std::size_t chunk_read_limit,
   // Don't need to do it if we read the file all at once.
   if (_chunk_read_limit > 0) {
     for (auto const& buff : _output_buffers) {
-      _output_buffers_template.emplace_back(column_buffer::empty_like(buff));
+      _output_buffers_template.emplace_back(inline_column_buffer::empty_like(buff));
```
PointKernel, line 274: Is there a difference between `inline_column_buffer` and `gather_column_buffer` in this case?

cpp/src/io/utilities/column_buffer.cpp
```diff
@@ -265,6 +296,34 @@ std::unique_ptr<column> empty_like(column_buffer& buffer,
...
+                                                         rmm::cuda_stream_view stream,
+                                                         rmm::mr::device_memory_resource* mr);
+
+template std::unique_ptr<column> empty_like<pointer_type>(pointer_column_buffer& buffer,
+                                                          column_name_info* schema_info,
+                                                          rmm::cuda_stream_view stream,
+                                                          rmm::mr::device_memory_resource* mr);
+
+template class column_buffer_base<pointer_type>;
+template class column_buffer_base<string_type>;
```
PointKernel, line 328:
```suggestion
```

should be a leftover

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -0,0 +1,1291 @@
...
+#pragma once
+
+#include "parquet_gpu.hpp"
+#include "rle_stream.cuh"
+
+#include <cuda/std/tuple>
+#include <io/utilities/block_utils.cuh>
+
+namespace cudf::io::parquet::gpu {
+namespace {
```
PointKernel, line 26: An anonymous namespace is not recommended in a header file.
```suggestion
```

PointKernel (commented):
Partial reviews. Looking into page decoding details...

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -37,12 +37,31 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
...
 
+  // Check to see if there are any string columns present. If so, then we need to get size info
+  // for each string page. This size info will be used to pre-allocate memory for the column,
+  // allowing the page decoder to write string data directly to the column buffer, rather than
+  // doing a gather operation later on.
+  // TODO: This step is somewhat redundant if size info has already been calculated (nested schema,
+  // chunked reader).
+  auto is_string_col = [](gpu::ColumnChunkDesc const& chunk) {
+    return (chunk.data_type & 7) == BYTE_ARRAY && (chunk.data_type >> 3) != 4 &&
+           chunk.converted_type != DECIMAL;
```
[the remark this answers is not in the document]
etseidl: I added the check for DECIMAL because I noticed [this](https://github.com/rapidsai/cudf/blob/5e12c25ca40c41f21fc80d4cd36713c514fabaa4/cpp/src/io/parquet/page_data.cu#L1085) and [this](https://github.com/rapidsai/cudf/blob/5e12c25ca40c41f21fc80d4cd36713c514fabaa4/cpp/src/io/parquet/page_data.cu#L2112). It appears that DECIMAL can be encoded as either BYTE_ARRAY or FIXED_LEN_BYTE_ARRAY. I only want to count up the BYTE_ARRAY columns that are UTF8.

cpp/src/io/utilities/column_buffer.hpp
```diff
@@ -62,86 +62,184 @@ inline rmm::device_buffer create_data(data_type type,
...
+  {
+  }
 
   // instantiate a column of known type with a specified size.  Allows deferred creation for
   // preprocessing steps such as in the Parquet reader
-  void create(size_type _size, rmm::cuda_stream_view stream, rmm::mr::device_memory_resource* mr);
+  void create(size_type _size, rmm::cuda_stream_view stream, rmm::mr::device_memory_resource* mr)
+  {
+    size = _size;
+    _mr  = mr;
```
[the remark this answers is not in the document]
etseidl: > Shall we use a consistent naming convention to improve the code? I'm fine with either all data members with `_` prefix or all input with `_` prefix. The goal is to avoid such a mix use.

I agree, but the reason for the split right now is to distinguish between public and private data members.  I would like to make all the members private and prefixed, but that change had a huge ripple effect that I think should be addressed in a separate PR.

cpp/src/io/utilities/column_buffer.cpp
```diff
@@ -26,43 +26,68 @@
...
-
-    // list columns store a buffer of int32's as offsets to represent
-    // their individual rows
-    case type_id::LIST: _data = create_data(data_type{type_id::INT32}, size, stream, mr); break;
-
-    // struct columns store no data themselves.  just validity and children.
-    case type_id::STRUCT: break;
-
-    default: _data = create_data(type, size, stream, mr); break;
+  if (type.id() == type_id::STRING) {
```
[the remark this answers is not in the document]
etseidl: leftover from an earlier iteration.  thanks.

cpp/src/io/parquet/reader_impl.cpp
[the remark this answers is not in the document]
PointKernel: Oh, shame on me. It was me adding this feature and somehow I forgot binary can be read binary as decimals during the review.

cpp/src/io/utilities/column_buffer.hpp
```diff
@@ -62,86 +62,184 @@ inline rmm::device_buffer create_data(data_type type,
...
-                rmm::mr::device_memory_resource* mr)
-    : column_buffer(_type, _is_nullable)
-  {
-    create(_size, stream, mr);
-  }
+template <class string_policy>
+std::unique_ptr<column> make_column(column_buffer_base<string_policy>& buffer,
+                                    column_name_info* schema_info,
+                                    std::optional<reader_column_schema> const& schema,
+                                    rmm::cuda_stream_view stream);
```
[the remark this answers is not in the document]
etseidl: I'm fine with changing this, but it seems like the original intent was to use the memory resource used to create the `column_buffer` for subsequent allocations. Do we want to change this here or as a separate PR?

cpp/src/io/utilities/column_buffer.cpp
etseidl: The forced template instantiations are actually needed.  The `using` defs are just sugar.  Would you prefer fully qualified names for the rest?

PointKernel: Ah, you are right. Using aliases is totally fine. Please ignore my previous comment.

cpp/src/io/utilities/column_buffer.hpp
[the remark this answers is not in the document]
PointKernel: According to https://github.com/rapidsai/cudf/blob/branch-23.08/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#output-memory, any function allocates the data and returns it to the user/invoker should take a memory resource parameter (`mr`) and uses that `mr` to allocate the data. i.e., in this case, the returned `column` should be allocated with the input `mr`.

cpp/src/io/utilities/column_buffer.hpp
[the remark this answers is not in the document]
PointKernel: > should be addressed in a separate PR.

works for me

cpp/src/io/utilities/column_buffer.hpp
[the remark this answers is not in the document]
etseidl: I wound up just removing the mr from `column_buffer`. I think that will lead to less confusion.

cpp/src/io/utilities/column_buffer.cpp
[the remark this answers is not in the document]
PointKernel: A runtime assert to avoid misuse would be great but not mandatory. I'm fine with either way.

cpp/src/io/parquet/reader_impl.cpp
etseidl: Somewhere in all the refactoring we lost the delete of the copy constructor and the definition of the move constructor. Should those go back in?  @vuule

vuule: Oh yeah, that one's on me :grimacing: 
I think it would be enough to delete the copy constructor/assignment in the base class.
Why do we need to define the move constructor for these classes?

PointKernel: > Parquet only uses inline_column_buffer

A stupid question: in parquet, string data can be decoded via `string_index_pair` (or the `gather_column_buffer`), right?

cpp/src/io/parquet/page_decode.cuh
PointKernel: Thanks for the fix!

cpp/src/io/parquet/reader_impl.cpp
etseidl: > Why do we need to define the move constructor for these classes?

STL throws fits if copy is deleted and move isn't defined, it seems.  But it seems happy if both are in the base class...I had tried putting both in the derived classes.

cpp/src/io/utilities/column_buffer.hpp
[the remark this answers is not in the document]
etseidl: And then reverted.  @vuule pointed me to #12364.  The lack of passing in the memory resource is very deliberate.

vuule (commented):
Misc comments, some possibly outdated at this point.
Parts of the new kernels seem to be very much in line with the current code, so I didn't look into those in detail.
Great PR overall, thank you @etseidl !

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -0,0 +1,1157 @@
...
+    // - so we will end up ignoring the first two input rows, and input rows 2..n will
+    //   get written to the output starting at position 0.
+    //
+    if (!has_repetition) { dst_pos -= s->first_row; }
+
+    // need to do this before we branch on src_pos/dst_pos so we don't deadlock
+    // choose a character parallel string copy when the average string is longer than a warp
+    using cudf::detail::warp_size;
+    auto const use_char_ll =
+      s->page.num_valids > 0 && (s->page.str_bytes / s->page.num_valids) >= warp_size;
```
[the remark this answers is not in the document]
etseidl: It can. But divide by zero is ok on device?  I added a print statement when `num_valids` is zero, but removing the check for `num_valids > 0` didn't crash.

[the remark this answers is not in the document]
vuule: Disregard, I completely missed the `s->page.num_valids > 0` part of the expression, you got it covered :D
I'm actually not sure what would happen if an int is divided by zero on the device.

vuule (approved):

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -37,12 +37,31 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
       return cursum + _metadata->get_output_nesting_depth(chunk.src_col_schema);
     });
 
+  // Check to see if there are any string columns present. If so, then we need to get size info
+  // for each string page. This size info will be used to pre-allocate memory for the column,
+  // allowing the page decoder to write string data directly to the column buffer, rather than
+  // doing a gather operation later on.
+  // TODO: This step is somewhat redundant if size info has already been calculated (nested schema,
+  // chunked reader).
+  auto is_string_col = [](gpu::ColumnChunkDesc const& chunk) {
```
hyperbolic2346, line 46: This function is duplicated in `page_decode.cuh` and seems easy to get wrong and bug fix only one. Is there a way we can combine these?

hyperbolic2346 (commented):
Great insight to lead to this PR. I have a few architectural questions, which I would imagine are just my lack of understanding.

cpp/src/io/parquet/reader_impl.cpp
etseidl: Should be able to do an inline or constexpr shared version. I'll give it a go.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -120,8 +147,14 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
   chunks.host_to_device(_stream);
   chunk_nested_valids.host_to_device(_stream);
   chunk_nested_data.host_to_device(_stream);
+  chunk_nested_str_data.host_to_device(_stream);
 
+  // TODO: explore launching these concurrently with a stream pool
```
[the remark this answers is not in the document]
etseidl: This was an offline suggestion from @nvdbaranec for #12948, where I was once doing just what you suggest in `gpuDecodePageData`. If I am remembering and understanding correctly, branching to call the different decoders increases the shared memory requirements for all, even if the other kernels aren't called.  This approach was seen as a stop-gap solution.  It would ultimately be nice to split the pages up into batches, with each batch being sent to the proper decoder kernel. Down the road there will be additional kernels for the delta encodings.

nvdbaranec (commented):
Can you point out the major highlights for the changes outside `page_string_decode.cu`.  My initial feeling here is a lot of the green and red I'm seeing is mostly just code moving around :)

cpp/src/io/parquet/reader_impl.cpp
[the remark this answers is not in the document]
nvdbaranec: Another thing to consider here:  by calling these on the same stream we are losing a decent amount of parallelism.  These kernels won't execute at the same time, so it's possible to get into a situation like:

- GPU has 100 SMs
- 50 pages contain string data
- 50 pages contain int data

At any given time, we will only be using half the SMs.   I think it would be worth (as a followup) to either figure out how to conglomerate shared memory usage so that these can be run in one invocation, or to investigate using multiple streams.   

I'm doing something like with the an upcoming PR.  Fundamentally:

This:
```
__global__ kernel()
{
    if(a){
       path_a();
   } else {
      path_b();
   }
}

void path_a()
{
    __shared__ x;
}

void path_b()
{
    __shared__ y;
}
```

Becomes:

```
struct shared_stuff{
  ...
};

__global__ kernel()
{
  __shared__ shared_stuff s;
   if(a){
      path_a(s);
   } else {
      path_b(s);
   }
}

void path_a(shared_stuff s)
{
}

void path_b(shared_stuff s)
{
}
```

Since the shared stuff for a and b is pretty much the same (buffers for rep/def levels, some cub scratch space, etc) it ends up removing most duplication.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -0,0 +1,1125 @@
...
+ * @tparam level_t Type used to store decoded repetition and definition levels
+ */
+template <int lvl_buf_size, typename level_t>
+__global__ void __launch_bounds__(decode_block_size) gpuDecodeStringPageData(
+  PageInfo* pages, device_span<ColumnChunkDesc const> chunks, size_t min_row, size_t num_rows)
+{
+  __shared__ __align__(16) page_state_s state_g;
+  __shared__ __align__(16) page_state_buffers_s state_buffers;
+  __shared__ __align__(4) size_type last_offset;
+
```
nvdbaranec, line 659: This data and the rep/def buffers below seem identical to the data in the V2 function.   We should wrap these in a struct that gets declared once (as shared) in the root DecodeStringPageData function and then passed as a reference to various subfunctions.

nvdbaranec (changes requested):

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -0,0 +1,1125 @@
...
+  // if no repetition we haven't calculated start/end bounds and instead just skipped
+  // values until we reach first_row. account for that here.
+  if (!has_repetition) { value_count -= s->first_row; }
+
+  auto const offptr = reinterpret_cast<size_type*>(nesting_info_base[leaf_level_index].data_out);
+  block_excl_sum<decode_block_size>(offptr, value_count, s->page.str_offset);
+}
+
+__global__ void __launch_bounds__(preprocess_block_size)
+  gpuComputePageOffsets(device_span<ColumnChunkDesc const> chunks, device_span<size_t> col_sizes)
```
nvdbaranec, line 1014: This smells like something that could be done as a `thrust::exclusive_scan_by_key()`.   We did something just like this recently to eliminate a similar nested loop:

https://github.com/rapidsai/cudf/blob/0c68ca15011774541985f46fb566e07c634e1615/cpp/src/io/parquet/reader_impl_preprocess.cu#LL1754C2-L1754C2

I think if you structured your keys as:

`A A A A A B B B B B C C C C C ...`

where

```
|-------|  size of each set of keys = (# of chunks * max # of pages in any chunk)
A A A A A

and the # of sets would be the number of columns.
```

```
elements_per_key = chunks.size() * (max pages in any chunk)
num_keys = elements_per_key * num_columns
key_iter(int index) = {
   return index / elements_per_key  //  col_index
}
value_iter(int index) = {
   // the logic here may not be exactly right but you get the idea
   col_index = index / elements_per_key        // col_index

   intra_index = index % elements_per_key    // index within each chunk/page group
   chunk_index = intra_index / max pages     // chunk index
   chunk = chunks[chunk_index];
   page_index = intra_index % max pages     // page index
  
   if(!is_string_col(chunk)){ return 0; }
   if(page_index > chunk.max_num_pages) { return 0; }
   if(chunk.pages[page_index].flags & DICTIONARY) { return 0; }
   return chunk.page_info[page_index].str_bytes;
}
thrust::scan_by_key(key_iter, key_iter + num_keys, value_iter, output)
```

After the scan you should have what you're looking for, and the whole thing becomes fully parallel - no nested looping.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -0,0 +1,1125 @@
...
+ *
+ * @param pages All pages to be decoded
+ * @param chunks All chunks to be decoded
+ * @param min_rows crop all rows below min_row
+ * @param num_rows Maximum number of rows to read
+ * @tparam lvl_buf_size Size of the buffer used when decoding repetition and definition levels
+ * @tparam level_t Type used to store decoded repetition and definition levels
+ */
+template <int lvl_buf_size, typename level_t>
+__global__ void __launch_bounds__(preprocess_block_size) gpuComputePageStringSizes(
```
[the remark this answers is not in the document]
etseidl: TBH `gpuComputePageSizes` does stuff I still can't quite wrap my head around.  That said, it's only called for lists or when using custom row bounds or using the chunked reader. Early on I tried detecting if `gpuComputePageSizes` had been called and then calling it if it hadn't, but was not sure about what effect calling it had on other decoding state.  Also, the string counting bits don't seem to pay attention to num_rows/skip_rows, and I _believe_ there's a bit of a bug in `gpuDecodeDictionaryIndices` [here](https://github.com/rapidsai/cudf/blob/0c68ca15011774541985f46fb566e07c634e1615/cpp/src/io/parquet/page_data.cu#L344).  It's been awhile, but I think `target_pos` doesn't take into account null values, so the dictionary decoding winds up running off past the end of data.  So long story short, I couldn't easily figure out how to get the exact string sizes I needed short of implementing it myself. Your `rle_stream` came to the rescue, making this step so fast it wasn't a blocking issue in my mind, and I figured when the dust settles we could figure out how to merge the two counts kernels into one.

cpp/src/io/parquet/page_string_decode.cu
etseidl: I discussed this offline with @vuule, but the V2 kernel is not currently called, but will likely be the only version once `rle_stream` for decoding is online. I've left it in but `if constexpr`'d out to keep the bit rot away. I'm ok with just getting rid of at this point since there shouldn't be much changing at this point.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -0,0 +1,1125 @@
...
+ *
+ * @param pages List of pages
+ * @param chunks List of column chunks
+ * @param min_row Row index to start reading at
+ * @param num_rows Maximum number of rows to read
+ * @tparam lvl_buf_size Size of the buffer used when decoding repetition and definition levels
+ * @tparam level_t Type used to store decoded repetition and definition levels
+ */
+template <int lvl_buf_size, typename level_t>
+__global__ void __launch_bounds__(decode_block_size) gpuDecodeStringPageData(
```
nvdbaranec, line 653: Couple of things about these two functions `gpuDecodeStringPageData` and `gpuDecodeStringPageDataV2`.  As they are right now, these functions will be bottlenecked similar to the non-string path.  1 warp traversing all the level data.  I have a PR almost ready that parallelizes gpuDecodePageData with rle_stream.   

This will undo those wins for string data unless they also get converted to use rle_stream.  But it's a fair amount of work to do this conversion and we now have two additional functions that will need to get it done.  Is there any possibility of somehow merging `gpuDecodeStringPageData` and `gpuDecodeStringPageDataV2` so that they're just a single function with some branching?

cpp/src/io/parquet/page_string_decode.cu
etseidl: 🤯 

Thanks!  I tried and failed to come up with a way to do this from the host side with thrust calls. :(  I'll see if I can get this to work. (Might be bugging you offline)

cpp/src/io/parquet/page_string_decode.cu
etseidl: V2 was written with `rle_stream` in mind.  I envision a loop of:
```
 loop
   decode batch of level data
   decode dict if necessary
   write string data
end loop
```
V2 currently does the first two steps concurrently with 2 warps, and then does step 3 using all 4 warps. I hoping it won't be too much work to modify V2 to work with rle_stream 2.0.

nvdbaranec: I just saw your other reply that the V2 function probably ends up going away.  That reduces the stickiness.

nvdbaranec: Looking at `gpuDecodeStringPageData` I strongly suspect it might be possible to graft the actual decoding work here directly into the new `gpuDecodePageData() `.   The rest of the function structure looks mostly the same and I took the chunk of work that does the actual value copying and separated it into a new function `gpuDecodeValues()`.  It might be possible to just call into this string copying stuff from there. 

It's a big enough change that we probably don't want to hold this PR up on it.  It can be a followup (not a small amount of work though).

nvdbaranec: > V2 was written with `rle_stream` in mind. I envision a loop of:
> 
> ```
>  loop
>    decode batch of level data
>    decode dict if necessary
>    write string data
> end loop
> ```

Yeah this is exactly what we need.   A quick preview of the new gpuDecodePageData

```
while (processed < s->page.num_input_values) {
    int const this_processed = decoders[level_type::DEFINITION].decode_next(t);
    __syncthreads();

    int const this_valid = gpuUpdateValidityOffsetsAndRowIndicesFlat<lvl_buf_size, level_t>(processed + this_processed,
                                                                              s,
                                                                              &shared,
                                                                              def,
                                                                              t);
    __syncthreads();
  
    // if we have dictionary data, decode it now.
    if(s->dict_base){
      dict_stream.decode_next(t, this_valid);
      __syncthreads();
    }    

    // decode the values themselves
    gpuDecodeValues(s, sb, valid, valid + this_valid, t);
    __syncthreads();
  }
```

So if the string copying can be transplanted into gpuDecodeValues (which is basically everything in the last else statement in the existing decode function) then we're all good.

etseidl:
> Can you point out the major highlights for the changes outside `page_string_decode.cu`. My initial feeling here is a lot of the green and red I'm seeing is mostly just code moving around :)

All of the red bits yanked from `page_data.cu` were pasted into `page_decode.cuh`. The only changes being adding `inline` to functions that didn't already have that, the addition of `is_string_col`, and a modification to `is_bounds_page`. I was finding pages being labeled as 'bounds pages' that weren't because of the inclusive comparison operators.  Removing the '=', however, broke page spanning rows, so now I test for repetition and use different bounds accordingly.

Beyond that, `PageInfo` added a few fields for string handling, and there are the major changes to `column_buffer` to support the two different methods of handling string columns.

cpp/src/io/parquet/page_string_decode.cu
etseidl: Do you have your branch in github yet so I could take a peek? I'm ok with putting this work on hold and seeing how it fits with the new decoder. Plus there's also #13437 in play, and the delta decoding which relies heavily on this PR to avoid 2X the string copies.

cpp/src/io/parquet/page_string_decode.cu
etseidl: @nvdbaranec please check what I've done. In particular the change [here](https://github.com/etseidl/cudf/blob/59bfd1be94bba846599c8989d0a91f338ee558b8/cpp/src/io/parquet/reader_impl_preprocess.cu#L1259). When I tried reusing `page_keys`, the ReorderedColumns reader test failed.  Switching from `src_col_schema` to `src_col_index` fixed that.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -37,13 +37,37 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
...
+    gpu::ComputePageStringSizes(
+      pages, chunks, skip_rows, num_rows, _file_itm_data.level_type_size, _stream);
+
+    col_sizes = calculate_page_string_offsets();
+
+    // check for overflow
+    if (std::any_of(col_sizes.begin(), col_sizes.end(), [](size_t sz) {
+          return sz > std::numeric_limits<size_type>::max();
+        })) {
+      CUDF_FAIL("String column exceeds 2GB limit");
```
etseidl, line 59: This should throw `std::overflow_error` like other instances of exceed the 2GB string data limit.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -122,7 +156,13 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
   chunk_nested_valids.host_to_device_async(_stream);
   chunk_nested_data.host_to_device_async(_stream);
 
+  // TODO: explore launching these concurrently with a stream pool
   gpu::DecodePageData(pages, chunks, num_rows, skip_rows, _file_itm_data.level_type_size, _stream);
+  if (has_strings) {
+    chunk_nested_str_data.host_to_device_async(_stream);
```
nvdbaranec, line 162: So I have one last request.  The lingering concern here is running these separate kernels on the same stream.  In situations where we have small-ish files with a mixed number of types - strings and ints, running these kernels on the same stream potentially impacts performance seriously.  

If we have a GPU with 100 SMs, and we have 50 string pages and 50 int pages, normally they would all run at once.  But with this new arrangement, 50 would run, then another 50, halving gpu utilization.

The existing cudf benchmarks are not going to expose this because they favor generating inputs with huge numbers of pages, so the GPU will always be saturated.  So what I'd like to see is this:

- Add a benchmark that creates a reasonably large file with a similar to above profile.  Strings, ints.  40 pages of each (a T4 has 80 SMs so this would be a nice universal test).  You should be able to force this arrangement with the writer via `.max_page_size_bytes` and `.max_page_size+rows`

- Do a before/after comparison with your stuff.  There should be a pretty clear difference.

- As a step 2, at the top of decode_page_data, add this little static initializer for a stream pool, and then pull a stream out of the pool to invoke the string kernel via `.get_stream` (make sure to synchronize on this stream at a rational point alongside the core stream.  
```
auto& get_stream_pool(){
   static auto pool = rmm::cuda_stream_pool(10);
    return pool;
}
```

There should be a clear performance win (including some clear differences in an nsys trace).  Based on some results I've had with recent optimization attempts, running multiple, type-specific kernels is something we might be interested in pursuing further.  This string stuff might be a nice guinea pig.

etseidl: Pushed some changes...still working on some.  Results so far:
```
# parquet_read_io_small_mixed without stream pool

|    io    |  cardinality  |  run_length  |  num_string_cols  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |       Diff |   %Diff |
|----------|---------------|--------------|-------------------|------------|-------------|------------|-------------|------------|---------|
| FILEPATH |       0       |      1       |         1         |   4.925 ms |       5.29% |   5.005 ms |       1.74% |  79.316 us |   1.61% |
| FILEPATH |     1000      |      1       |         1         |   2.307 ms |       2.72% |   3.356 ms |       2.53% |   1.049 ms |  45.45% |
| FILEPATH |       0       |      32      |         1         |   4.749 ms |       4.27% |   4.794 ms |       1.49% |  45.341 us |   0.95% |
| FILEPATH |     1000      |      32      |         1         |   1.357 ms |       2.93% |   1.942 ms |       6.34% | 585.435 us |  43.15% |
| FILEPATH |       0       |      1       |         2         |   5.483 ms |       1.72% |   5.516 ms |       2.80% |  32.711 us |   0.60% |
| FILEPATH |     1000      |      1       |         2         |   2.308 ms |       2.81% |   3.356 ms |       2.33% |   1.049 ms |  45.44% |
| FILEPATH |       0       |      32      |         2         |   5.267 ms |       1.82% |   5.304 ms |       1.44% |  37.500 us |   0.71% |
| FILEPATH |     1000      |      32      |         2         |   1.469 ms |       2.24% |   1.937 ms |       1.46% | 468.230 us |  31.88% |
| FILEPATH |       0       |      1       |         3         |   5.948 ms |       1.39% |   5.983 ms |       1.22% |  35.133 us |   0.59% |
| FILEPATH |     1000      |      1       |         3         |   2.338 ms |       7.68% |   3.362 ms |       2.08% |   1.023 ms |  43.78% |
| FILEPATH |       0       |      32      |         3         |   5.797 ms |       2.04% |   5.809 ms |       1.24% |  11.800 us |   0.20% |
| FILEPATH |     1000      |      32      |         3         |   1.583 ms |       2.43% |   1.934 ms |       1.53% | 351.543 us |  22.21% |

# parquet_read_io_small_mixed with stream pool

|    io    |  cardinality  |  run_length  |  num_string_cols  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |        Diff |   %Diff |
|----------|---------------|--------------|-------------------|------------|-------------|------------|-------------|-------------|---------|
| FILEPATH |       0       |      1       |         1         |   4.925 ms |       5.29% |   4.935 ms |       4.67% |    9.449 us |   0.19% |
| FILEPATH |     1000      |      1       |         1         |   2.307 ms |       2.72% |   2.267 ms |       3.16% |  -39.907 us |  -1.73% |
| FILEPATH |       0       |      32      |         1         |   4.749 ms |       4.27% |   4.757 ms |       3.02% |    8.061 us |   0.17% |
| FILEPATH |     1000      |      32      |         1         |   1.357 ms |       2.93% |   1.532 ms |       2.18% |  175.445 us |  12.93% |
| FILEPATH |       0       |      1       |         2         |   5.483 ms |       1.72% |   5.204 ms |       8.58% | -279.011 us |  -5.09% |
| FILEPATH |     1000      |      1       |         2         |   2.308 ms |       2.81% |   2.263 ms |       3.27% |  -44.884 us |  -1.95% |
| FILEPATH |       0       |      32      |         2         |   5.267 ms |       1.82% |   5.217 ms |       1.94% |  -49.412 us |  -0.94% |
| FILEPATH |     1000      |      32      |         2         |   1.469 ms |       2.24% |   1.612 ms |       2.43% |  143.197 us |   9.75% |
| FILEPATH |       0       |      1       |         3         |   5.948 ms |       1.39% |   5.910 ms |       3.03% |  -37.821 us |  -0.64% |
| FILEPATH |     1000      |      1       |         3         |   2.338 ms |       7.68% |   2.268 ms |       3.91% |  -69.983 us |  -2.99% |
| FILEPATH |       0       |      32      |         3         |   5.797 ms |       2.04% |   5.862 ms |       1.54% |   65.267 us |   1.13% |
| FILEPATH |     1000      |      32      |         3         |   1.583 ms |       2.43% |   1.625 ms |       1.60% |   42.474 us |   2.68% |
```
As expected times are bad for the cases where the GPU is left half idle because of the serial nature of decoding the pages.  Using a stream pool to overlap the decoding kernels does much better (but still a little worse than the old code).

etseidl: Screenshot from 2023-06-21 11-40-00

Here's the profile from the 1000/32/2 benchmark.

etseidl: I tried to streamline `setupLocalPageInfo` as @nvdbaranec suggested offline, but there's some wrinkle I'm missing. One of the unit tests fails with null counts larger than expected. I'll give this another go.

etseidl: @vuule can you look at how I did the stream pool?  I wound up adding it to the impl object rather than using a static one.

etseidl: latest benchmark with the stream pool in place:
```
# parquet_read_decode

|  data_type  |      io       |  cardinality  |  run_length  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |         Diff |   %Diff |
|-------------|---------------|---------------|--------------|------------|-------------|------------|-------------|--------------|---------|
|  INTEGRAL   | DEVICE_BUFFER |       0       |      1       |  14.446 ms |       0.75% |  14.836 ms |       0.99% |   390.704 us |   2.70% |
|  INTEGRAL   | DEVICE_BUFFER |     1000      |      1       |  13.672 ms |       0.44% |  14.000 ms |       0.50% |   327.931 us |   2.40% |
|  INTEGRAL   | DEVICE_BUFFER |       0       |      32      |  12.078 ms |       0.50% |  12.566 ms |       0.48% |   488.227 us |   4.04% |
|  INTEGRAL   | DEVICE_BUFFER |     1000      |      32      |  11.556 ms |       0.45% |  12.051 ms |       0.50% |   494.633 us |   4.28% |
|    FLOAT    | DEVICE_BUFFER |       0       |      1       |   8.755 ms |       0.38% |   8.772 ms |       0.36% |    16.405 us |   0.19% |
|    FLOAT    | DEVICE_BUFFER |     1000      |      1       |   8.589 ms |       0.50% |   8.545 ms |       0.53% |   -44.250 us |  -0.52% |
|    FLOAT    | DEVICE_BUFFER |       0       |      32      |   7.287 ms |       0.55% |   7.290 ms |       0.70% |     2.525 us |   0.03% |
|    FLOAT    | DEVICE_BUFFER |     1000      |      32      |   7.549 ms |       1.39% |   7.544 ms |       2.35% |    -4.273 us |  -0.06% |
|   DECIMAL   | DEVICE_BUFFER |       0       |      1       |  19.095 ms |       0.50% |  18.840 ms |       0.75% |  -254.786 us |  -1.33% |
|   DECIMAL   | DEVICE_BUFFER |     1000      |      1       |   7.569 ms |       0.65% |   7.501 ms |       0.58% |   -67.905 us |  -0.90% |
|   DECIMAL   | DEVICE_BUFFER |       0       |      32      |   9.118 ms |       0.77% |   9.116 ms |       0.50% |    -1.569 us |  -0.02% |
|   DECIMAL   | DEVICE_BUFFER |     1000      |      32      |   6.291 ms |       0.71% |   6.230 ms |       0.67% |   -61.431 us |  -0.98% |
|  TIMESTAMP  | DEVICE_BUFFER |       0       |      1       |  25.775 ms |       0.57% |  25.213 ms |       0.50% |  -561.543 us |  -2.18% |
|  TIMESTAMP  | DEVICE_BUFFER |     1000      |      1       |   8.146 ms |       0.57% |   8.046 ms |       0.49% |   -99.693 us |  -1.22% |
|  TIMESTAMP  | DEVICE_BUFFER |       0       |      32      |  10.686 ms |       1.56% |  10.580 ms |       0.95% |  -106.155 us |  -0.99% |
|  TIMESTAMP  | DEVICE_BUFFER |     1000      |      32      |   7.087 ms |       0.57% |   7.018 ms |       0.71% |   -68.621 us |  -0.97% |
|  DURATION   | DEVICE_BUFFER |       0       |      1       |  10.709 ms |       0.50% |  10.771 ms |       0.50% |    62.400 us |   0.58% |
|  DURATION   | DEVICE_BUFFER |     1000      |      1       |   7.828 ms |       0.42% |   7.889 ms |       0.65% |    60.604 us |   0.77% |
|  DURATION   | DEVICE_BUFFER |       0       |      32      |   7.860 ms |       0.41% |   7.902 ms |       0.50% |    41.143 us |   0.52% |
|  DURATION   | DEVICE_BUFFER |     1000      |      32      |   6.859 ms |       1.31% |   6.897 ms |       0.54% |    37.835 us |   0.55% |
|   STRING    | DEVICE_BUFFER |       0       |      1       |  50.841 ms |       0.49% |  41.080 ms |       0.50% | -9760.994 us | -19.20% |
|   STRING    | DEVICE_BUFFER |     1000      |      1       |  17.094 ms |       0.59% |  13.336 ms |       0.63% | -3757.784 us | -21.98% |
|   STRING    | DEVICE_BUFFER |       0       |      32      |  50.890 ms |       0.68% |  41.009 ms |       0.50% | -9880.564 us | -19.42% |
|   STRING    | DEVICE_BUFFER |     1000      |      32      |  15.565 ms |       0.24% |  12.407 ms |       0.71% | -3157.489 us | -20.29% |
|    LIST     | DEVICE_BUFFER |       0       |      1       |  80.768 ms |       0.40% |  80.828 ms |       0.20% |    60.150 us |   0.07% |
|    LIST     | DEVICE_BUFFER |     1000      |      1       |  72.315 ms |       0.17% |  72.743 ms |       0.28% |   428.268 us |   0.59% |
|    LIST     | DEVICE_BUFFER |       0       |      32      |  62.031 ms |       0.29% |  62.582 ms |       0.36% |   551.235 us |   0.89% |
|    LIST     | DEVICE_BUFFER |     1000      |      32      |  62.994 ms |       0.20% |  63.540 ms |       0.23% |   546.548 us |   0.87% |
|   STRUCT    | DEVICE_BUFFER |       0       |      1       |  63.745 ms |       0.26% |  56.472 ms |       0.22% | -7273.554 us | -11.41% |
|   STRUCT    | DEVICE_BUFFER |     1000      |      1       |  37.924 ms |       0.37% |  34.097 ms |       0.32% | -3827.127 us | -10.09% |
|   STRUCT    | DEVICE_BUFFER |       0       |      32      |  61.452 ms |       0.61% |  54.311 ms |       0.68% | -7140.755 us | -11.62% |
|   STRUCT    | DEVICE_BUFFER |     1000      |      32      |  35.541 ms |       0.36% |  31.730 ms |       0.14% | -3810.204 us | -10.72% |
```

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -122,7 +159,16 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
...
+  auto stream1 = _stream_pool.get_stream();
+  gpu::DecodePageData(pages, chunks, num_rows, skip_rows, _file_itm_data.level_type_size, stream1);
+  if (has_strings) {
+    auto stream2 = _stream_pool.get_stream();
+    chunk_nested_str_data.host_to_device_async(stream2);
+    gpu::DecodeStringPageData(
+      pages, chunks, num_rows, skip_rows, _file_itm_data.level_type_size, stream2);
+    stream2.synchronize();
+  }
+  stream1.synchronize();
```
nvdbaranec, line 171: I think we should always guarantee that we call the first kernel with the passed in stream. So stream1 here should always just be `_stream`.  Then we can pull off other streams from the pool as needed.  

Also, I think we need a `_stream.synchronize()` before we launch these kernels because of the `host_to_device_async()` calls above.

etseidl: As feared, using `_stream` for the first kernel tanks performance for the small mixed case. Is this ok?
```
## parquet_read_io_small_mixed

pooled stream1

|    io    | cardinality | run_length | num_string_cols | Samples | CPU Time | Noise | GPU Time | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|----------|-------------|------------|-----------------|---------|----------|-------|----------|-------|------------------|-------------------|-------------------|
| FILEPATH |        1000 |          1 |               2 |   1312x | 2.267 ms | 7.41% | 2.262 ms | 7.41% |     237354738759 |        10.487 MiB |         1.019 MiB |
| FILEPATH |        1000 |         32 |               2 |   2544x | 1.604 ms | 2.02% | 1.599 ms | 1.97% |     335678427608 |         9.585 MiB |       126.455 KiB |

_stream for first kernel
|    io    | cardinality | run_length | num_string_cols | Samples | CPU Time | Noise | GPU Time | Noise | bytes_per_second | peak_memory_usage | encoded_file_size |
|----------|-------------|------------|-----------------|---------|----------|-------|----------|-------|------------------|-------------------|-------------------|
| FILEPATH |        1000 |          1 |               2 |   2064x | 3.361 ms | 2.30% | 3.356 ms | 2.29% |     159987070458 |        10.487 MiB |         1.019 MiB |
| FILEPATH |        1000 |         32 |               2 |   1280x | 1.920 ms | 2.04% | 1.915 ms | 1.98% |     280400202774 |         9.585 MiB |       126.455 KiB |
```

vuule: @nvdbaranec why do you think that the first kernel should use the user stream? IMO it's fine to fork to different streams in the middle of the read call.
Although, the more I think about this, the less sense I can make of the user stream parameter overall (given that read_parquet is synchronous in practice).

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -663,38 +663,19 @@ __global__ void __launch_bounds__(decode_block_size) gpuDecodeStringPageData(
   page_state_buffers_s* const sb = &state_buffers;
   int const page_idx             = blockIdx.x;
   int const t                    = threadIdx.x;
+  [[maybe_unused]] null_count_back_copier _{s, t};
```
nvdbaranec, line 666: How does this avoid the race condition when two separate kernels visit the same page? Won't one of them erroneously zero the page out that another may have written a valid value to?

etseidl: Only one invocation should make it past the filter. That one will zero out the null count and then the back copier will copy it back to the page. @vuule added the logic to make the back copy a no-op if the setup returns early.

nvdbaranec: Ah, I see.  Checking to see if the nesting_info pointer is null.

vuule:
oops, missing a cmake review

vyasr (approved):
CMake approval

PR #13307, opened 2023-05-08
Exposure Tracked Buffer (first step towards unifying copy-on-write and spilling)

madsbk:
The first step towards unifying copy-on-write and spillable buffers. 

This PR re-implement copy-on-write by introducing a `ExposureTrackedBuffer` and `BufferSlice`. The idea is that when `copy-on-write` (and in a follow-up PR later, when `spill`) is enabled, we use `BufferSlice` throughout cudf. 
`BufferSlice` is a _view_ of a `ExposureTrackedBuffer` that implements copy-on-write semantics by tracking the number of `BufferSlice` that points to the same `ExposureTrackedBuffer`. 

## Checklist

cc. @shwina, @vyasr, @galipremsagar, @wence-


docs/cudf/source/developer_guide/library_design.md
```diff
@@ -325,30 +325,26 @@ This section describes the internal implementation details of the copy-on-write
...
-This means that all `CopyOnWriteBuffer`s that point to the same device memory are contained in the same `WeakSet` (corresponding to the same `ptr` key) in `CopyOnWriteBuffer._instances`.
-This data structure is then used to determine whether or not to make a copy when a write operation is performed on a `Column` (see below).
-If multiple buffers point to the same underlying memory, then a copy must be made whenever a modification is attempted.
+The core copy-on-write implementation relies on the factory function `as_tenable_buffer` and the two classes `TenableBuffer` and `BufferSlice`.
+
+`TenableBuffer` is a subclass of the regular `Buffer` that tracks its "expose" status of its underlying memory. We say that the buffer has been exposed if the device pointer (integer or void*) has been accessed outside of cudf, in which case we have no control over knowing if the data is being modified by a third-party. Additionally, `TenableBuffer` also maintains [weak references](https://docs.python.org/3/library/weakref.html) to every existing `BufferSlice` that points to its underlying memory.
```
wence-, line 330: How about:

> A `TenableBuffer` is a subclass of the regular `Buffer` that tracks internal and external references to its underlying memory. Internal references are tracked by maintaining [weak references](https://docs.python.org/3/library/weakref.html) to every `BufferSlice` of the underlying memory. External references are tracked through "exposure" status of the underlying memory. A buffer is considered exposed if the device pointer (integer or `void*`) has been handed out to a library outside of cudf. In this case, we have no way of knowing if the data are being modified by a third party.

WDYT?

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -325,30 +325,26 @@ This section describes the internal implementation details of the copy-on-write
...
+
+`BufferSlice` is a subclass of `TenableBuffer` that represents a _slice_ of the memory underlying a tenable buffer.
```
wence-, line 332: Is subclassing the right thing? (I haven't yet read the implementation). It seems like every `BufferSlice` `has-a` `TenableBuffer`, but that doesn't necessarily imply an `is-a` relationship.

If it really is a subclass, do we actually need both classes, or can there just be `TenableBuffer` objects that either own data, or are views of existing data.

docs/cudf/source/developer_guide/library_design.md
```diff
@@ -325,30 +325,26 @@ This section describes the internal implementation details of the copy-on-write
...
 
 ### Obtaining a read-only object
 
 A read-only object can be quite useful for operations that will not
-mutate the data. This can be achieved by calling `._get_cuda_array_interface(readonly=True)`, and creating a `SimpleNameSpace` object around it.
-This will not trigger a deep copy even if the `CopyOnWriteBuffer`
-has weak references. This API should only be used when the lifetime of the proxy object is restricted to cudf's internal code execution. Handing this out to external libraries or user-facing APIs will lead to untracked references and undefined copy-on-write behavior. We currently use this API for device to host
+mutate the data. This can be achieved by calling `.get_ptr(mode="read")`, and use `cuda_array_interface_wrapper` to wrap a `__cuda_array_interface__` object around it.
```
wence-, line 346:
```suggestion
mutate the data. This can be achieved by calling `.get_ptr(mode="read")`, and using `cuda_array_interface_wrapper` to wrap a `__cuda_array_interface__` object around it.
```

python/cudf/cudf/_lib/column.pyx
```diff
@@ -554,10 +554,7 @@ cdef class Column:
                     owner=data_owner,
                     exposed=True,
                 )
-                if isinstance(data_owner, CopyOnWriteBuffer):
-                    data_owner.get_ptr(mode="write")
```
wence-, line 558: Why do we no longer have to mark the `data_owner` as exposed in the `TenableBuffer` case?

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -169,7 +169,7 @@ def _from_host_memory(cls: Type[T], data: Any) -> T:
         # Create from device memory
         return cls._from_device_memory(buf)
 
-    def _getitem(self, offset: int, size: int) -> Buffer:
+    def _getitem(self: T, offset: int, size: int) -> T:
```
wence-, line 172: Throughout, use `from typing_extensions import Self` and then
```suggestion
    def _getitem(self: Self, offset: int, size: int) -> Self:
```

ref: (PEP 673)[https://peps.python.org/pep-0673/]

python/cudf/cudf/core/buffer/tenable_buffer.py
```diff
@@ -0,0 +1,302 @@
...
+    In order to implement copy-on-write and spillable buffers, we need the
+    ability to detect external access to the underlying memory. We say that
+    the buffer has been exposed if the device pointer (integer or void*) has
+    been accessed outside of TenableBuffer. In this case, we have no control
+    over knowing if the data is being modified by a third-party.
+
+    Attributes
+    ----------
+    _exposed
+        The current expose status of the buffer. Notice, once the expose status
```
wence-, line 109:
```suggestion
        The current exposure status of the buffer. Notice, once the exposure status
```

python/cudf/cudf/core/buffer/tenable_buffer.py
```diff
@@ -0,0 +1,302 @@
...
+        self._owner = base
+        self._passthrough_attributes = passthrough_attributes
+        base._slices.add(self)
+
+    def __getattr__(self, name):
+        if name in self._passthrough_attributes:
+            return getattr(self._base, name)
+        raise AttributeError(
+            f"{self.__class__.__name__} object has no attribute {name}"
+        )
```
wence-, line 210: Aha, here's the `has-a` vs. `is-a` problem.

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -275,19 +255,26 @@ def get_ptr(self, *, mode) -> int:
...
+        self, *, offset: int = 0, size: Optional[int] = None
+    ) -> memoryview:
         """Read-only access to the buffer through host memory."""
-        host_buf = host_memory_allocation(self.size)
+        size = self._size if size is None else size
+        host_buf = host_memory_allocation(size)
         rmm._lib.device_buffer.copy_ptr_to_host(
-            self.get_ptr(mode="read"), host_buf
+            self.get_ptr(mode="read") + offset, host_buf
         )
```
wence-, line 278: Does anyone do error checking if we ask for a larger memoryview than is allocated?

python/cudf/cudf/core/buffer/tenable_buffer.py
```diff
@@ -0,0 +1,302 @@
...
+
+        Parameters
+        ----------
+        deep : bool, default True
+            The semantics when copy-on-write is disabled:
+                - If True, returns a deep copy of the underlying Buffer data.
+                - If False, returns a shallow copy of the Buffer pointing to
+                  the same underlying data.
+            The semantics when copy-on-write is enabled:
+                - Always a deep copy of the underlying Buffer data.
```
wence-, line 248: I thought with C-o-W, we always get a shallow copy here. (Unless the data are exposed).

python/cudf/cudf/core/buffer/tenable_buffer.py
```diff
@@ -0,0 +1,302 @@
...
+        Returns
+        -------
+        Buffer
+            Buffer representing the same device memory as `data`
+        """
+
+        if len(self._base._slices) > 1:
+            # If this is not the only slice pointing to `self._base`, we
+            # point to a new deep copy of the base.
+            t = self.copy(deep=True)
```
wence-, line 291: Why do we not just copy `self._base` directly?

python/cudf/cudf/core/buffer/tenable_buffer.py
```diff
@@ -0,0 +1,302 @@
...
+        -------
+        BufferSlice
+            A slice pointing to either a new or the existing base buffer
+            depending on the expose status of the base buffer and the
+            copy-on-write option (see above).
+        """
+        if deep or not cudf.get_option("copy_on_write"):
+            base_copy = self._base.copy(deep=deep)
+        else:
+            base_copy = self._base.copy(deep=self.exposed)
```
wence-, line 260: Maybe rewrite as:
```
if cudf.get_option("copy_on_write"):
    base_copy = self._base.copy(deep=deep or self.exposed)
else:
    base_copy = self._base.copy(deep=deep)
```

Which I _think_ makes it clearer that the exposure status only matters for the CoW case.

python/cudf/cudf/core/buffer/tenable_buffer.py
```diff
@@ -0,0 +1,302 @@
...
+
+        if len(self._base._slices) > 1:
+            # If this is not the only slice pointing to `self._base`, we
+            # point to a new deep copy of the base.
+            t = self.copy(deep=True)
+            self._base = t._base
+            self._offset = t._offset
+            self._size = t._size
+            self._owner = t._base
+            self._base._slices.add(self)
```
wence-, line 296: `self._base._slices` will temporarily contain `self` and `t` but `t` will disappear when it goes out of scope...

wence- (commented):
A first pass, haven't fully managed to go through all the logic yet

docs/cudf/source/developer_guide/library_design.md
madsbk: In a follow-up PR, `SpillableBuffer` will inherent from `TenableBuffer` and `BufferSlice._base` will point to `SpillableBuffer`.

madsbk: Maybe it is better if `BufferSlice` inherent from `Buffer` ?

wence-: It kind of seems like what we want is [ad-hoc](https://en.wikipedia.org/wiki/Ad_hoc_polymorphism), rather than [subtype](https://en.wikipedia.org/wiki/Subtyping), polymorphism (`BufferSlice`s are not Liskov-substitutable for `TenableBuffer`s everywhere [in particular when constructing new buffer slices]). In which case, is this the time to introduce a buffer [`Protocol`](https://peps.python.org/pep-0544/)?

python/cudf/cudf/core/buffer/tenable_buffer.py
madsbk: As far as i understand, from the user's point of view everything is a deep copy ?

wence-: Ah, I see what you mean. Perhaps something along the lines of:

> Always a deep copy of the underlying Buffer data which is either eager (if the buffer is exposed) or lazy (if not)

wence-: Maybe that doesn't actually clarify things because we've introduced new terminology.

docs/cudf/source/developer_guide/library_design.md
madsbk: A buffer Protocol might make sense, but then I think we should do it in a follow up PR

python/cudf/cudf/core/buffer/buffer.py
madsbk: I guess we need `mypy v1.0.0` for `Self` support: https://github.com/python/mypy/issues/11871
I am getting:
```
python/cudf/cudf/core/buffer/buffer.py:173: error: Variable "typing_extensions.Self" is not valid as a type
python/cudf/cudf/core/buffer/buffer.py:173: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
python/cudf/cudf/core/buffer/buffer.py:186: error: Variable "typing_extensions.Self" is not valid as a type
python/cudf/cudf/core/buffer/buffer.py:186: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
python/cudf/cudf/core/buffer/buffer.py:198: error: Variable "typing_extensions.Self" is not valid as a type
python/cudf/cudf/core/buffer/buffer.py:198: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
python/cudf/cudf/core/buffer/tenable_buffer.py:262: error: Value of type Self? is not indexable
Found 4 errors in 2 files (checked 149 source files)
```

madsbk: Maybe submit a small PR that replaces all `self: T` and bump the mypy version?

python/cudf/cudf/core/buffer/buffer.py
madsbk: no, I don't think so

python/cudf/cudf/core/buffer/spillable_buffer.py
```diff
@@ -364,7 +364,7 @@ def spill_lock(self, spill_lock: SpillLock) -> None:
             self.spill(target="gpu")
             self._spill_locks.add(spill_lock)
 
-    def get_ptr(self, *, mode) -> int:
+    def get_ptr(self, mode: str = "write") -> int:
```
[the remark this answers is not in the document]
madsbk: Good point, fixed

python/cudf/cudf/core/buffer/tenable_buffer.py
```diff
@@ -0,0 +1,302 @@
...
+    def exposed(self) -> bool:
+        return self._exposed
+
+    def mark_exposed(self) -> None:
+        """Mark the buffer as "exposed" permanently"""
+        self._exposed = True
+
+    @classmethod
+    def _from_device_memory(
+        cls: Type[T], data: Any, *, exposed: bool = False
```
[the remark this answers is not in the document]
madsbk: Good catch, fixed

python/cudf/cudf/core/buffer/tenable_buffer.py
```diff
@@ -0,0 +1,302 @@
...
+        self._base = base
+        self._offset = offset
+        self._size = size
+        self._owner = base
+        self._passthrough_attributes = passthrough_attributes
+        base._slices.add(self)
+
+    def __getattr__(self, name):
+        if name in self._passthrough_attributes:
+            return getattr(self._base, name)
```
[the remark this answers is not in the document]
madsbk: Good catch, fixed

python/cudf/cudf/core/buffer/tenable_buffer.py
madsbk: This is to support `SpillableBuffer` when it also uses `BufferSlice`

python/cudf/cudf/core/buffer/tenable_buffer.py
madsbk: I think that is ok

python/cudf/cudf/core/buffer/tenable_buffer.py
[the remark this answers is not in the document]
madsbk: It is to make it compatible with `Buffer`. We could default to `None` and raise an error if it isn't set?

docs/cudf/source/developer_guide/library_design.md
madsbk: I like it :)
Done

docs/cudf/source/developer_guide/library_design.md
madsbk: Alternatively, we could have `Buffer` use `BufferSlice` so that `BufferSlice` is the only buffer object used in the rest of cudf.
In any case, I think we should wait until after the next PR that makes `SpillableBuffer` use  `BufferSlice` and `TenableBuffer`

python/cudf/cudf/core/buffer/tenable_buffer.py
madsbk: Agree, fixed

python/cudf/cudf/core/buffer/buffer.py
wence-: If you merge trunk you can now do this, I opened https://github.com/rapidsai/cudf/pull/13346 for the other cases.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -534,13 +534,16 @@ cdef class Column:
                     rmm.DeviceBuffer(ptr=data_ptr,
                                      size=(size+offset) * dtype_itemsize)
                 )
-            elif column_owner and isinstance(data_owner, CopyOnWriteBuffer):
-                # TODO: In future, see if we can just pass on the
-                # CopyOnWriteBuffer reference to another column
-                # and still create a weak reference.
-                # With the current design that's not possible.
-                # https://github.com/rapidsai/cudf/issues/12734
```
vyasr, line 542: I guess ExposureTrackedBuffer handles this properly?

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+
+    if type(data) is klass:
+        return data
+    if hasattr(data, "owner"):
+        return get_owner(data.owner, klass)
+    return None
+
+
+def as_exposure_tracked_buffer(
```
vyasr, line 50: Can we define this in utils.py since that's where as_buffer is defined?

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
```diff
@@ -0,0 +1,311 @@
...
+from typing_extensions import Self
+
+import cudf
+from cudf.core.buffer.buffer import Buffer, get_ptr_and_size
+from cudf.utils.string import format_bytes
+
+T = TypeVar("T", bound="ExposureTrackedBuffer")
+
+
+def get_owner(data, klass: Type[T]) -> Optional[T]:
```
vyasr, line 26: This function seems a bit odd. If `klass` is None then `get_owner` is always `None`? Is this functionality going to change once SpillableBuffer goes away and spilling relies directly on ExposureTrackedBuffer? Or is it going to stick around this way because of slices?

vyasr: OK from reading below it looks like this is necessary because we'll always be working with BufferSlice objects.

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
```diff
@@ -0,0 +1,311 @@
...
+    a BufferSlice that points to all of the new `subclass` is returned
+
+    It is illegal for an exposure tracked buffer to own another exposure
+    tracked buffer. When representing the same memory, we should have a single
+    exposure tracked buffer and multiple buffer slices.
+
+    Developer Notes
+    ---------------
+    This function always returns slices thus all buffers in cudf will use
+    `BufferSlice` when copy-on-write is enabled. The slices implements
```
vyasr, line 69:
```suggestion
    `BufferSlice` when copy-on-write is enabled. The slices implement
```

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
```diff
@@ -0,0 +1,311 @@
...
+    copy-on-write by trigging deep copies when write access is detected
+    and multiple slices points to the same exposure tracked buffer.
+
+    Parameters
+    ----------
+    data : buffer-like or array-like
+        A buffer-like or array-like object that represent C-contiguous memory.
```
vyasr, line 76:
```suggestion
        A buffer-like or array-like object that represents C-contiguous memory.
```

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
```diff
@@ -0,0 +1,311 @@
...
+    BufferSlice
+        A buffer slice that points to a ExposureTrackedBuffer (or `subclass`),
+        which in turn wraps `data`.
+    """
+
+    if not hasattr(data, "__cuda_array_interface__"):
+        if exposed:
+            raise ValueError("cannot created exposed host memory")
+        return cast(
+            BufferSlice, ExposureTrackedBuffer._from_host_memory(data)[:]
```
vyasr, line 93: Is the eventual plan for Buffer.__getitem__ to return a `BufferSlice`? If not, it might be cleaner to override the method in `ExposureTrackedBuffer`. I know the whole point of the `_getitem`/`__getitem__` split is to help share some functionality, but the typing confusion here indicates that there are potentially incorrect types that could result from that approach (obviously we can coerce the code into behaving correctly, but it makes it much harder to write intrinsically type-safe code if the type annotations aren't sufficiently valid).

python/cudf/cudf/tests/test_copying.py
```diff
@@ -114,7 +114,8 @@ def test_series_setitem_partial_slice_cow_on():
 
         new_slice = actual[2:]
         assert (
-            new_slice._column.base_data._ptr == actual._column.base_data._ptr
+            new_slice._column.base_data._base._ptr
```
vyasr, line 117: Maybe once we've merged spilling and cow we can find a cleaner way to test this? Might be nice to add a TODO. I don't have a particular suggestion, but it's a bit unfortunate that our only option right now is to reach multiple times into the internals. Perhaps there isn't really a good alternative though.

vyasr (changes requested):
A few small change requests, but in general this looks like the right direction. Thanks @madsbk! I can see from this PR how you would integrate spilling, but also appreciate you splitting up the work this way to make the changes incrementally.

python/cudf/cudf/_lib/column.pyx
madsbk: yes

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
madsbk: I like to wait doing this until `SpillableBuffer` also becomes a `ExposureTrackedBuffer`.
At the moment, `as_buffer()`, imports spillable and exposure tracked buffers as orthogonal types:
```python
from cudf.core.buffer.exposure_tracked_buffer import as_exposure_tracked_buffer
from cudf.core.buffer.spillable_buffer import SpillLock, as_spillable_buffer
```

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
madsbk: I agree, this is a very valid point!

I think the clean design is to make cudf always work on `BufferSlice` even when COW and spilling is disabled. Then we get a clean class hierarchy: 
- COW & Spilling disable:
`BufferSlice -> Buffer`
- COW enabled:
`BufferSlice -> ExposureTrackedBuffer -> Buffer`
- Spilling enabled (when is has been unified with COW in a follow-up PR):
`BufferSlice -> SpillableBuffer -> ExposureTrackedBuffer -> Buffer`

The downside is that this approach is a bit more intrusive in the default case where COW and spilling is disabled. I think it is worth it but what do you guys think?

cc. @wence- @shwina

python/cudf/cudf/tests/test_copying.py
vyasr: Could you add a TODO for this @madsbk?

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
vyasr: OK got it, that's fine with me.

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
vyasr: I agree that it is probably worth it. I would stage that as work to be done after the COW and spilling unification is complete and we can reevaluate in the context of a cleaner architecture.

vyasr (approved):

python/cudf/cudf/tests/test_copying.py
madsbk: done

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -233,35 +233,15 @@ def owner(self) -> Any:
...
-            ),
+            "data": (self.get_ptr(mode="write"), False),
             "shape": (self.size,),
             "strides": None,
             "typestr": "|u1",
             "version": 0,
         }
 
-    def get_ptr(self, *, mode) -> int:
+    def get_ptr(self, *, mode: Literal["read", "write"]) -> int:
```
wence-, line 244: No need to change this, but just to note that `Literal` doesn't get handled very well by type-checkers if the argument comes from a variable, rather than a literal value (unless the variable is marked with `: Final`). Since they don't do dataflow analysis

https://mypy-play.net/?mypy=latest&python=3.11&gist=6541ee12e80daeb4b0837563d98dc442

wence- (approved):
Very minor typo fix, thanks!

madsbk:
Thanks all for the reviews

Issue #13154, opened 2023-04-17
[FEA] Add overload for `set_null_mask` that accept result of `make_null_mask` directly


ttnghia:
Currently, we have the `make_null_mask` API returning both null mask and null count:
```
template <typename ValidityIterator>
std::pair<rmm::device_buffer, cudf::size_type> make_null_mask(ValidityIterator begin,
                                                              ValidityIterator end)
```

However, the `set_null_mask` function can't accept that result. As such, we have to pass the result of `make_null_mask` into `set_null_mask` by two steps:
```
auto [null_mask, null_count] =
    cudf::test::detail::make_null_mask(valid_iter, valid_iter + input.size());
  output.set_null_mask(std::move(null_mask), null_count);
```

This is a bit inconvenient. We can do that in just one step like this:
```
output.set_null_mask(cudf::test::detail::make_null_mask(valid_iter, valid_iter + input.size()));
```

In order to do that, we need to have a new overload of `set_null_mask` that accept the return type of `make_null_mask` such as:
```
void set_null_mask(std::pair<rmm::device_buffer, cudf::size_type>&&) { ... }
```

ttnghia:
CC @davidwendt.

PR #13311, opened 2023-05-09
Remove null mask and null count from column_view constructors

vyasr:
## Description

This is a breaking change that removes default values. Removing the default null count value of `UNKNOWN_NULL_COUNT` is necessary since we are removing `UNKNOWN_NULL_COUNT` altogether. A potential alternative was setting the default to 0, which would correspond to a default null mask of `nullptr` (the current default). However, that change is potentially more dangerous if callers were previously setting the null mask to something other than `nullptr` without setting the null count and relying on the behavior of `UNKNOWN_NULL_COUNT`. Removing the default parameter altogether is therefore the safer option here.

Contributes to #11968.

## Checklist


davidwendt (approved):

cpp/include/cudf/column/column_view.hpp
```diff
@@ -289,7 +289,7 @@ class column_view_base {
                    size_type size,
                    void const* data,
                    bitmask_type const* null_mask = nullptr,
-                   size_type null_count          = UNKNOWN_NULL_COUNT,
+                   size_type null_count          = 0,
```
ttnghia, line 292: Do we have any validity check for null count? If we specify some `null_mask`, but leave `null_count` default to `0`, will the column_view be constructed as-is?

ttnghia: Answer to myself: We don't have any validity check. So in such cases, we will have invalid `null_count`. I would suggest to add such check, which mean we have to execute the kernel to count nulls upon column_view construction.

vuule: That's a good point. With this API change, it becomes easier to create a column with nulls and zero null count. Ideally we would not have a default `null_count` when `null_mask` is passed.

vuule: One way to do this would be to tie the two parameters
```
struct null_mask{
  bitmask_type const* mask;
  size_type count;
};
```
instead of passing these dependent parameters separately. But something like this would be a **very** breaking change.

ttnghia: That sounds reasonable. We can make such breaking change at the end after completely removing `UNKNOWN_NULL_COUNT`.

GregoryKimball:
@etseidl I'd like to make sure you are aware of this work. Would you please let me know if you are using `mutable_column_view` in your application? If so, I recommend checking the null mask and null count handling to make sure the data structures stay consistent.

cpp/include/cudf/column/column_view.hpp
ttnghia: I am thinking of removing the default value for both null mask and null count to avoid potential bug. It will not be breaking changes, as the callers can just pass in an empty null mask. How do you guys think about it?

vyasr: I had discussed this recently with @davidwendt and thought that @ttnghia had actually already proposed what @vuule did above. Looking through recent issues, I suspect he was referring to #13154. That issue is a subset of what David and I had discussed, which was indeed redefining the null mask everywhere as a pair containing both pieces of information. That would enforce that you could never read or write one without the other. Whether we actually want to go quite that far, or just modify select APIs, remains to be seen. I definitely think that is the right direction for us to take though.

@vuule @ttnghia what do you think is a minimal acceptable change for:
1. Moving this PR forward
2. The 23.06 release
3. The long-term API changes

vuule: This PR can potentially break user code, when they create a column with nulls and default null count.
We need an API that takes both mask and count, or none. AFAICT, we can't get to that API without an API break. Which means we can't remove `UNKNOWN_NULL_COUNT` without an API break.

I'm not too experienced with non-cuIO API changes, but it seems like we'd have to deprecate these functions and offer an alternative that's safe to use in libcudf that has no `UNKNOWN_NULL_COUNT`. Then we can remove the old functions in the next release.

ttnghia: I'm fine with moving merging PR as-is to avoid more friction, assuming that we will address all our concerns above in the follow up PR. Merging this as-is will open a loophole for bug, so we will need to close it by the follow up (breaking change) work ASAP.

vyasr: libcudf doesn't promise deprecations, we only say we'll make a best effort. In this instance I think that ship has already failed since David and I have had at least one API-breaking change merged. Our hope at this stage is to get all the breaking changes that we want around `UNKNOWN_NULL_COUNT` merged into this release. So I'm inclined to follow Nghia's recommendation of merging this PR ASAP, then making follow-ups immediately to address the outstanding issues.

vuule: What is the plan for the follow-up changes?

vyasr: After this and #13341 are merged, we'll be ready to remove `UNKNOWN_NULL_COUNT`. Then we have two choices for the immediate follow-up:
- @vuule's suggestion: Couple the two parameters (mask and count) for some set of APIs by changing to require a pair as input
- @ttnghia's suggestion: Remove the default null mask parameter
and I suppose there is the question of whether we think this is a must for 23.06.

@davidwendt what do you think?

vuule: Unless we have other potential uses for a `null_mask_with_known_null_count`, it's probably better to just remove the default parameters.

vyasr:
To be clear, @etseidl this particular PR won't affect anything major regarding the mutable column view. The problematic case will be in a subsequent PR (related to this overall project). The relevant change will affect the following current behavior:
1. Create and populate a column
2. Create a mutable view of that column
3. Modify the null mask via the mutable view in place

Currently, the creation of the mutable view (step 2) unsets the null count in the column, so the next time that the column's null count is queried it is recomputed based on the mask. In the future, that lazy computation of the count will be removed, so the user will be required to set the null count as part of step 3 (setting the mask alone will be insufficient).

cpp/include/cudf/column/column_view.hpp
davidwendt: I thought that changing the signature could be something we would do in a future release. I think I'm convinced that changing to remove the default parameters or introducing the pair type would be more responsible. I  suspect removing the defaults may be the easier path.

vyasr: I was able to remove the default parameter without too much fuss. The total diff is still fairly small, so I think that was the right call to do in this PR.

vyasr: @vuule @ttnghia does that change look good to you, or is there anything else you'd like to see in this PR?

ttnghia: Yes, that's good. Thanks :+1:

vuule: Please update the PR description as well :)

cpp/src/io/csv/writer_impl.cu
```diff
@@ -251,7 +251,7 @@ struct column_to_strings_fn {
     return cudf::strings::detail::from_timestamps(
       column,
       format,
-      strings_column_view(column_view{data_type{type_id::STRING}, 0, nullptr}),
+      strings_column_view(column_view{data_type{type_id::STRING}, 0, nullptr, nullptr, 0}),
```
vuule, line 254: we don't have a way to create an empty column view?

davidwendt: Maybe something like this?
```
strings_column_view(make_empty_column(type_id::STRING)->view())
```

cpp/src/binaryop/compiled/binary_ops.cu
```diff
@@ -50,8 +50,11 @@ struct scalar_as_column_view {
   return_type operator()(scalar const& s, rmm::cuda_stream_view, rmm::mr::device_memory_resource*)
   {
     auto& h_scalar_type_view = static_cast<cudf::scalar_type_t<T>&>(const_cast<scalar&>(s));
-    auto col_v =
-      column_view(s.type(), 1, h_scalar_type_view.data(), (bitmask_type const*)s.validity_data());
+    auto col_v               = column_view(s.type(),
+                             1,
+                             h_scalar_type_view.data(),
+                             (bitmask_type const*)s.validity_data(),
```
bdice, line 56: Can we eliminate the C-style cast here? This one isn't safe to do as a `static_cast` because it's effectively `bool *` to `uint32_t *` and I want that to be super obvious.
```suggestion
                             reinterpret_cast<bitmask_type const*>(s.validity_data()),
```

cpp/src/dictionary/detail/concatenate.cu
```diff
@@ -245,7 +245,8 @@ std::unique_ptr<column> concatenate(host_span<column_view const> columns,
   std::vector<column_view> indices_views(columns.size());
   std::transform(columns.begin(), columns.end(), indices_views.begin(), [](auto cv) {
     auto dict_view = dictionary_column_view(cv);
-    if (dict_view.is_empty()) return column_view{data_type{type_id::UINT32}, 0, nullptr};
+    if (dict_view.is_empty())
+      return column_view{data_type{type_id::UINT32}, 0, nullptr, nullptr, 0};
```
bdice, line 249: I strongly prefer braces for readability in cases where conditionals break lines.
```suggestion
    if (dict_view.is_empty()) {
      return column_view{data_type{type_id::UINT32}, 0, nullptr, nullptr, 0};
    }
```

cpp/include/cudf/column/column_view.hpp
vyasr: Good call. Updated the description and the PR title.

cpp/src/binaryop/compiled/binary_ops.cu
vyasr: Sure. To be clear, this code predates my PR, but it's small enough that I'm OK with squeezing the change in here nonetheless.

ttnghia (approved):

cpp/include/cudf/column/column_view.hpp
```diff
@@ -288,9 +288,9 @@ class column_view_base {
   column_view_base(data_type type,
                    size_type size,
                    void const* data,
-                   bitmask_type const* null_mask = nullptr,
-                   size_type null_count          = UNKNOWN_NULL_COUNT,
-                   size_type offset              = 0);
+                   bitmask_type const* null_mask,
+                   size_type null_count,
```
vuule, line 292: null_mask and null_count are still documented as optional (same goes for the other two functions).

vuule (changes requested):

cpp/include/cudf/column/column_view.hpp
vyasr: Good catch, thanks!

vuule (approved):

PR #13312, opened 2023-05-09
Simplify null count checking in column equality comparator

vyasr:
## Description

This PR removes an extra code path used for checking the equality of the null count when verifying if columns are equivalent (not equal). The purpose of this code path was to verify a specific definition of equivalence for columns containing unsanitized nulls, i.e. by ignoring the stored null count and directly verifying the validity of the underlying null mask. This is no longer necessary because we required sanitized null masks to be output from all libcudf APIs now (see the "libcudf expects nested types to have sanitized null masks" section in the [developer guide](https://docs.rapids.ai/api/libcudf/stable/developer_guide)), and this requirement will be enforced with the merge of #14559.

## Checklist

### DIFF
--- a/cpp/tests/copying/get_value_tests.cpp
+++ b/cpp/tests/copying/get_value_tests.cpp
@@ -756,7 +756,7 @@ TYPED_TEST(ListGetStructValueTest, NestedGetNonNullEmpty)
     this->make_test_lists_column(3, {0, 1, 1, 2}, std::move(list_column), {1, 1, 1});
 
   auto expected_data =
-    this->make_test_lists_column(0, {0}, this->zero_length_struct().release(), {1});
+    this->make_test_lists_column(0, {0}, this->zero_length_struct().release(), {});
 
   cudf::size_type index = 1;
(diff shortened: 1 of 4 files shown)


vyasr:
This looks to be safe. As discussed offline with @nvdbaranec, this special function for null counting is to handle non-empty nulls, which we no longer support. The remaining todo here is to add a test with an `EXPECT_FAIL` for a column containing non-empty nulls to ensure that our assertion functions like `expect_columns_equivalent` will fail when provided such a column, ensuring that we won't accidentally have tests passing on such inputs in the future.

hyperbolic2346 (approved):

bdice (approved):

GregoryKimball:
@vyasr do you expect a performance impact from this simplification?

GregoryKimball:
>@vyasr do you expect a performance impact from this simplification?
No, this is only a test change

vyasr:
Confirming the last comment, which I discussed with Greg offline.

PR #13315, opened 2023-05-09
Correctly set bitmask size in `from_column_view`

### LINKED ISSUE #13305: [BUG] Calling to_pandas() on a dask dataframe containing a 'struct' column can result in a raised PyArrow exception
**Describe the bug**
Calling `to_pandas()` on a dask dataframe created from a cuDF dataframe containing a 'struct' column can result in a raised PyArrow exception:

```shell
/opt/anaconda/envs/morpheus/lib/python3.10/site-packages/cudf/core/column/struct.py:56: in to_arrow
    return pa.StructArray.from_buffers(
pyarrow/array.pxi:1110: in pyarrow.lib.Array.from_buffers
    ???
pyarrow/array.pxi:1504: in pyarrow.lib.Array.validate
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pyarrow.lib.ArrowInvalid: Buffer #0 too small in array of type struct<> and length 10: expected at least 2 byte(s), got 0

pyarrow/error.pxi:100: ArrowInvalid
```

Using this example file `azure_ad_logs.json`: https://github.com/nv-morpheus/Morpheus/blob/branch-23.07/tests/tests_data/azure_ad_logs.json

```python
def test_cudf_struct_type_conversion_dask():
    import cudf
    import dask_cudf

    cdf = cudf.read_json("azure_ad_logs.json")
    ddf = dask_cudf.from_cudf(cdf, npartitions=1)

    result = ddf['properties'].compute().to_pandas()

    print(result)
```

**Expected behavior**
A pandas dataframe should be returned.

Conda environment
```shell
grep 'cudf|arrow|dask'
arrow                     1.2.3                    pypi_0    pypi
arrow-cpp                 10.0.1          ha770c72_14_cuda    conda-forge
cudf                      23.02.00        cuda_11_py310_g5ad4a85b9d_0    rapidsai
dask                      2023.1.1           pyhd8ed1ab_0    conda-forge
dask-core                 2023.1.1           pyhd8ed1ab_0    conda-forge
dask-cuda                 23.2.1             pyhd8ed1ab_1    conda-forge
dask-cudf                 23.02.00        cuda_11_py310_g5ad4a85b9d_0    rapidsai
libarrow                  10.0.1          h255618e_14_cuda    conda-forge
libcudf                   23.02.00        cuda11_g5ad4a85b9d_0    rapidsai
pyarrow                   10.0.1          py310hc81d9b2_14_cuda    conda-forge
```

**Environment details**

**Additional context**

wence-:
## Description

An empty struct column (dtype of StructDtype({})) has no children, and
hence a base_size of zero. However, it may still have a non-zero size
and non-empty null mask. When slicing such a column, the mask size
must be transferred over correctly by inspecting the size and offset
of the owning column. Previously, we incorrectly determined the sliced
column to have a mask buffer of zero bytes in this case.

Closes #13305.

## Checklist

### DIFF
--- a/python/cudf/cudf/_lib/column.pyx
+++ b/python/cudf/cudf/_lib/column.pyx
@@ -522,7 +522,8 @@ cdef class Column:
         if data_ptr:
             if data_owner is None:
                 data = as_buffer(
-                    rmm.DeviceBuffer(ptr=data_ptr, size=base_nbytes)
+                    rmm.DeviceBuffer(ptr=data_ptr,
+                                     size=(size+offset) * dtype_itemsize)
                 )
             elif column_owner and isinstance(data_owner, CopyOnWriteBuffer):
                 # TODO: In future, see if we can just pass on the


python/cudf/cudf/_lib/column.pyx
```diff
@@ -155,7 +155,11 @@ cdef class Column:
             )
 
         if value is not None:
-            required_size = bitmask_allocation_size_bytes(self.base_size)
+            # base_size may be zero for a struct column with no
+            # children, but the column may have non-zero length.
+            required_size = bitmask_allocation_size_bytes(
+                max(self.base_size, self._size)
```
wence-, line 161: Is this correct, or does one also need to handle the offset of the column?

python/cudf/cudf/tests/test_struct.py
```diff
@@ -392,3 +392,13 @@ def test_struct_with_null_memory_usage():
 
     s[2:4] = None
     assert s.memory_usage() == 272
+
+
+@pytest.mark.parametrize("zlice", [slice(0, 3), slice(1, 4)])
```
bdice, line 397: Can we use `slice_` instead of… _zlice_ ?

wence-: Done.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -605,7 +609,9 @@ cdef class Column:
             else:
                 mask = as_buffer(
                     data=mask_ptr,
-                    size=bitmask_allocation_size_bytes(base_size),
+                    size=bitmask_allocation_size_bytes(
+                        max(size+offset, base_size)
+                    ),
```
wence-, line 614: Having looked again, I think `size + offset` (rounded up to bitmask allocation size) is always the correct size for the null mask data pointer. Here is my rationale. `column_view.null_mask()` returns a `bitmask_type *` but without taking into account any `offset` the `column_view` may have. So if I have a `column_view` with size `10` and offset `200`, the bitmask pointer will point to a buffer of size `round_up(10 + 200)`, and the bits starting at `200` will be relevant for this view.

So, suggestion
```suggestion
                    size=bitmask_allocation_size_bytes(size+offset),
```

python/cudf/cudf/_lib/column.pyx
```diff
@@ -155,7 +155,11 @@ cdef class Column:
             )
 
         if value is not None:
-            required_size = bitmask_allocation_size_bytes(self.base_size)
+            # base_size may be zero for a struct column with no
+            # children, but the column may have non-zero length.
+            required_size = bitmask_allocation_size_bytes(
+                max(self.base_size, self._size)
+            )
```
wence-, line 162: As below, I think both of these options are wrong. I think that the required size for the bitmask must be _at least_ `bitmask_allocation_size_bytes(self._size + self._offset)`.

So, suggestion:

```suggestion
            # bitmask size must be relative to offset = 0 data.
            required_size = bitmask_allocation_size_bytes(self._size + self._offset)
```

python/cudf/cudf/_lib/column.pyx
wence-: I think the column offset needs to be handled.

wence- (commented):
Another self-review, please check my working here...

python/cudf/cudf/_lib/column.pyx
```diff
@@ -155,7 +155,11 @@ cdef class Column:
             )
 
         if value is not None:
-            required_size = bitmask_allocation_size_bytes(self.base_size)
+            # base_size may be zero for a struct column with no
+            # children, but the column may have non-zero length.
```
shwina, line 159: Wait.. I thought that the length of a column could only ever be <= its `base_size`?

wence-:
```
In [1]: import cudf
cudf.Se
In [2]: cudf.Series([None, None, None, None], dtype=cudf.StructDtype({}))
Out[2]: 
0    None
1    None
2    None
3    None
dtype: struct

In [3]: series = cudf.Series([None, None, None, None], dtype=cudf.StructDtype({}))

In [4]: series._column.base_size
Out[4]: 0

In [5]: series._column._size
Out[5]: 4

In [6]: len(series._column)
Out[6]: 4

In [7]: series._column.base_children
Out[7]: ()
```

wence-: It's not a particularly interesting column (it's a very heavyweight way of representing a bitmask), but it's constructable.

shwina (commented):
I worry that in the future someone may use `base_size` when they should be using `size + offset`, and I _really_ would like for those both to be equal always.

Would redefining the `base_size` of a struct column as follows fix this bug? (and would it break anything else?)

```python

class StructColumn:

    ...

    @property
    def base_size(self):
        if not self.base_children:
            # this can happen if the column is empty
            # OR if the column is all nulls
            return self.null_count
        else:
            return len(self.base_children[0])
```

wence-:
> ```python
>     @property
>     def base_size(self):
>         if not self.base_children:
>             # this can happen if the column is empty
>             # OR if the column is all nulls
>             return self.null_count
> ```
This branch is wrong, it should be `self._size + self._offset`
> ```python
>         else:
>             return len(self.base_children[0])
> ```

I think this branch is right.

So perhaps this would do the trick:
```diff
diff --git a/python/cudf/cudf/core/column/struct.py b/python/cudf/cudf/core/column/struct.py
index 6838d71164..f8ae4ab846 100644
--- a/python/cudf/cudf/core/column/struct.py
+++ b/python/cudf/cudf/core/column/struct.py
@@ -29,9 +29,11 @@ class StructColumn(ColumnBase):
     @property
     def base_size(self):
         if not self.base_children:
-            return 0
+            return self.size + self.offset
         else:
-            return len(self.base_children[0])
+            size = len(self.base_children[0])
+            assert size == self.size + self.offset, "Unpossible"
+            return size
 
     def to_arrow(self):
         children = [
```

python/cudf/cudf/tests/test_struct.py
```diff
@@ -392,3 +392,13 @@ def test_struct_with_null_memory_usage():
 
     s[2:4] = None
     assert s.memory_usage() == 272
+
+
+@pytest.mark.parametrize("indices", [slice(0, 3), slice(1, 4)])
+def test_struct_empty_children_nulls_slice(indices):
+    values = [None, {}, {}, None]
+
+    s = cudf.Series([None, {}, {}, None])
```
bdice, line 401: Reuse the input from the previous line if it’s equivalent.
```suggestion
    s = cudf.Series(values)
```

bdice (approved):
This looks fine to me! Do you need to cover test cases with non-contiguous indices for iloc or is it only affecting “proper slices” with step=1?

shwina:
> if one slices a column, s.iloc[10:20] then size + offset is 20, but len(self.base_children[0]) is the length of the original s

Hmm, that's right. Sorry, I got this very wrong.

That being said, would you agree that the `base_size` of an all-null struct column should be non-zero? Going back to:

>```python
>    @property
>    def base_size(self):
>        if not self.base_children:
>            # this can happen if the column is empty
>            # OR if the column is all nulls
>            return self.null_count
>```
> This branch is wrong, it should be self.size + self.offset

When does this branch return the wrong answer?

wence-:
> This looks fine to me! Do you need to cover test cases with non-contiguous indices for iloc or is it only affecting “proper slices” with step=1?

It is only "proper slices" where the result of the slice shares data with the sliced column, so yes, only "proper" indices need covered.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -521,8 +522,7 @@ cdef class Column:
         if data_ptr:
             if data_owner is None:
                 data = as_buffer(
-                    rmm.DeviceBuffer(ptr=data_ptr,
-                                     size=(size+offset) * dtype_itemsize)
+                    rmm.DeviceBuffer(ptr=data_ptr, size=base_nbytes)
```
wence-, line 525: This is wrong sorry, will revert this hunk.

python/cudf/cudf/core/column/struct.py
```diff
@@ -28,10 +28,10 @@ class StructColumn(ColumnBase):
 
     @property
     def base_size(self):
-        if not self.base_children:
-            return 0
-        else:
+        if self.base_children:
```
wence-, line 31: Flipped the conditional...

python/cudf/cudf/_lib/column.pyx
```diff
@@ -599,7 +600,7 @@ cdef class Column:
                     mask = as_buffer(
                         rmm.DeviceBuffer(
                             ptr=mask_ptr,
-                            size=bitmask_allocation_size_bytes(size+offset)
+                            size=bitmask_allocation_size_bytes(base_size)
```
wence-, line 603: This is fine because if `column_owner` is None then `base_size == size + offset`.

PR #13323, opened 2023-05-09
Use std::overflow_error when output would exceed column size limit

### LINKED ISSUE #12925: [FEA] Throw `std::overflow` exception when an output column size exceeds cudf limit
In Spark, we want to handle the cases when the libcudf APIs throw exception due to their output column having sizes exceeding cudf limit. We want to differentiate if the thrown exception was really due to such reason (so we can split the input by half and try again), or due to something else (such as invalid input type or inconsistent data type etc). Unfortunately, currently only one type of exception (`cudf::logic_error`) is being thrown.

We want to have a new specific exception type with a name like `cudf::output_size_exceed_limit`. This exception type will be used to throw in the case mentioned above. Adding this new exception type is similar to https://github.com/rapidsai/cudf/pull/12426 which introduced a new exception type to throw when the input data type is invalid.

davidwendt:
## Description
Replaces generic `cudf::logic_error` exception with `std::overflow_error` where appropriate in libcudf.
Since this changes what is thrown in certain APIs, I think this technically is a breaking change.

Closes #12925 

## Checklist


divyegala (approved):

ttnghia (approved):

cpp/src/join/hash_join.cu
```diff
@@ -375,8 +375,9 @@ hash_join<Hasher>::hash_join(cudf::table_view const& build,
 {
   CUDF_FUNC_RANGE();
   CUDF_EXPECTS(0 != build.num_columns(), "Hash join build table is empty");
-  CUDF_EXPECTS(build.num_rows() < cudf::detail::MAX_JOIN_SIZE,
-               "Build column size is too big for hash join");
+  CUDF_EXPECTS(build.num_rows() <= cudf::detail::MAX_JOIN_SIZE,
```
bdice, line 378: Do comparisons to the `size_type` max always need to be done as type `std::size_t`? I don't see how this could ever be false, especially if both operands are `size_type`.

cpp/src/text/subword/subword_tokenize.cu
```diff
@@ -165,9 +165,11 @@ tokenizer_result subword_tokenize(cudf::strings_column_view const& strings,
 {
   CUDF_EXPECTS(stride <= max_sequence_length,
                "stride must be less than or equal to max_sequence_length");
-  CUDF_EXPECTS(max_sequence_length * max_rows_tensor <
-                 static_cast<std::size_t>(std::numeric_limits<cudf::size_type>::max()),
-               "max_sequence_length x max_rows_tensor is too large for cudf output column size");
+  CUDF_EXPECTS(
+    max_sequence_length <=
+      (static_cast<uint64_t>(std::numeric_limits<cudf::size_type>::max()) / max_rows_tensor),
```
bdice, line 170: Why use `uint64_t` here instead of `std::size_t` for both operands?

cpp/tests/filling/repeat_tests.cpp
```diff
@@ -271,3 +271,21 @@ TEST_F(RepeatErrorTestFixture, CountHasNulls)
   // input_table.has_nulls() == true
   EXPECT_THROW(auto ret = cudf::repeat(input_table, count), cudf::logic_error);
 }
+
+TEST_F(RepeatErrorTestFixture, Overflow)
+{
+  auto input = cudf::test::fixed_width_column_wrapper<int32_t>(
+    thrust::make_counting_iterator(0), thrust::make_counting_iterator(0) + 100);
+  cudf::table_view input_table{{input}};
+  // TODO explain why div 2
```
bdice, line 280: Does this TODO need to be addressed?

bdice (changes requested):
Do we need to do overflow comparisons with operands as `std::size_t` always? Should we use a common detail function to do these comparisons, like:

```cpp
#include <limits>

template<typename T>
bool check_column_size_overflow(T size) {
    return static_cast<std::size_t>(size) <= static_cast<std::size_t>(std::numeric_limits<cudf::size_type>::max());
}
```

cpp/src/text/subword/subword_tokenize.cu
davidwendt: I suppose `size_t == uint64_t` so why would there be a preference to use `size_t` here?
Knowing that `max_sequence_length` and `max_rows_tensor` are both `uint32_t` seemed more natural to use `uint64_t`.

bdice: My preference is to use `std::size_t` because, though equivalent to `uint64_t` in most places, it has two benefits:
1. We know it's safe by definition (https://en.cppreference.com/w/cpp/types/size_t)
2. We can use the same type everywhere rather than `std::size_t` in some places and `uint64_t` in others. A common spelling in all overflow checks in libcudf, ideally defined by a detail overflow-checking function as mentioned in my PR comment, would be my preferred solution.

cpp/src/join/hash_join.cu
bdice: We're comparing a `size_type` to the value `constexpr size_type MAX_JOIN_SIZE{std::numeric_limits<size_type>::max()};`. This can never be false, so this `EXPECTS` cannot raise.

bdice (approved):
Approving to unblock this because I think it's a net improvement. But I'm still in favor of doing this work with a helper function that makes this logic less repetitive across libcudf.

cpp/src/join/hash_join.cu
davidwendt: You're right. This makes no sense and we should be able to remove `MAX_JOIN_SIZE` completely. I traced the history and this constant has existed for a very long time. It is possible it was intended to be smaller (or maybe at least different) than `size_type` at one time -- perhaps when prototyping. And the value has not changed since it was added in 2019. So I think we can safely remove it now.

bdice:
Note: While looking at the `MAX_JOIN_SIZE` issue, I think there's some cleanup of constants on the Python side that needs to be written in terms of `size_type_dtype`. https://github.com/rapidsai/cudf/blob/707a48078470b3037a32b8c503bb4df1f682df52/python/cudf/cudf/_lib/__init__.py#L43-L46

I'll file a follow-up there.

Issue #13301, opened 2023-05-05
[BUG] Avoid out of bounds integers in tests


vyasr:
**Describe the bug**
A number of cudf tests create numpy or cupy arrays with data that is out of bounds for the specified dtype. [Here is one easy example](https://github.com/rapidsai/cudf/blob/branch-23.06/python/cudf/cudf/tests/test_column.py#L401) where 453 is larger than the largest representable uint8 (2^8=256) and so the result overflows. We should not be doing this. Not only is it bad practice, but [numpy 1.24 has started warning about this behavior](https://numpy.org/doc/stable/release/1.24.0-notes.html#conversion-of-out-of-bound-python-integers), so it will eventually be removed. Note that this is an important issue for us because we treat warnings as errors in our test suite, so fixing this bug is effectively a blocker for updating to a newer numpy version.

**Steps/Code to reproduce bug**
Installing numpy 1.24 into a conda environment with numba 0.56 is not possible because numba constrains the numpy version, and at present cudf is constrained to numba<0.57. However, practical support was added in #13271, so a reasonable environment may be created by removing the numba and numpy dependencies (https://github.com/rapidsai/cudf/blob/branch-23.06/conda/environments/all_cuda-118_arch-x86_64.yaml#L50) and then pip installing numba>=0.57 and numpy==1.24.3 into that environment. Building cudf in that environment will reproduce the issue.

**Expected behavior**
Test suite should pass

vyasr:
We introduced an upper bound to the numpy pinnings in #13289 in order to temporarily avoid this problem. However, we are working to resolve it ASAP and remove the pinning.

PR #13325, opened 2023-05-10
Changes to support Numpy >= 1.24

shwina:
## Description
Closes https://github.com/rapidsai/cudf/issues/13301

## Checklist


python/cudf/cudf/__init__.py
```diff
@@ -96,6 +97,7 @@
 
     _setup_numba_linker(_PTX_FILE)
 
+    patch_numba_codegen_if_needed()
```
shwina, line 100: This hack will go away

vyasr:
Can you remove the numpy upper bound pinnings in dependencies.yaml and the cudf meta.yaml? Then in the CUDA 12.0 CI let's make sure that at least one run installs numpy 1.24 and still passes tests.

python/cudf/cudf/__init__.py
vyasr: I think it can be removed from this PR itself. Now that we've verified numpy 1.24 support, I would recommend removing the numba-related changes in this PR that you're using in order to allow running numba 0.57 (which is necessary to use numpy 1.24). We'll still get it tested because our CUDA 12 wheel builds will patch us to use 0.57 anyway (but with CUDA 12 we don't use cubinlinker/ptxcompiler so we don't need any edits for those). Then when we bump our numba to 0.57 tests should pass thanks to this PR.

python/cudf/cudf/tests/test_csv.py
```diff
@@ -150,8 +150,8 @@ def make_all_numeric_extremes_dataframe():
         np_type = pdf_dtypes[gdf_dtype]
         if np.issubdtype(np_type, np.integer):
             itype = np.iinfo(np_type)
-            extremes = [0, +1, -1, itype.min, itype.max]
-            df[gdf_dtype] = np.array(extremes * 4, dtype=np_type)[:20]
+            extremes = [itype.min, itype.max]
```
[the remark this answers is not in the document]
vyasr: Could you elaborate? Is this a task you want to accomplish in this PR?

python/cudf/cudf/tests/test_column.py
```diff
@@ -398,8 +398,8 @@ def test_column_view_string_slice(slc):
             cudf.core.column.as_column([], dtype="uint8"),
         ),
         (
-            cp.array([453], dtype="uint8"),
-            cudf.core.column.as_column([453], dtype="uint8"),
+            cp.array([255], dtype="uint8"),
```
vyasr, line 401: I guess it doesn't matter what value we choose here? Just wondering if it's important to use 453-256.

python/cudf/cudf/tests/test_numerical.py
```diff
@@ -194,6 +194,7 @@ def test_to_numeric_downcast_int(data, downcast):
     assert_eq(expected, got)
 
 
+@pytest.mark.filterwarnings("ignore:invalid value encountered in cast")
```
vyasr, line 197: Instead of applying this to the whole test, can we just wrap the `pd.to_numeric` call? This doesn't affect the `cudf.to_numeric` call, does it?

Also, should we be handling the warning conditionally? i.e. I assuming this happens when trying to downcast a signed to an unsigned type or something?

python/cudf/cudf/__init__.py
vyasr: I have one other question on this PR, would wait to make changes here until everything else is resolved in case you need to run more tests locally.

vyasr (commented):
Success! https://github.com/rapidsai/cudf/actions/runs/4953167580/jobs/8860521944?pr=13325#step:10:828 shows that numpy 1.24 is now getting installed into the 12.0.1 wheel build, and tests are passing

python/cudf/cudf/tests/test_column.py
shwina: I don't think it matters. I use `255` just because it's `np.iinfo(uint8).max'

python/cudf/cudf/utils/queryutils.py
```diff
@@ -137,7 +137,7 @@ def query_compile(expr):
         key "args" is a sequence of name of the arguments.
     """
 
-    funcid = f"queryexpr_{np.uintp(hash(expr)):x}"
+    funcid = f"queryexpr_{np.uintp(abs(hash(expr))):x}"
```
wence-, line 140: This is incorrect, `hash` returns in the semi-open interval `[-2**63, 2**63)`, but `abs` folds this to the closed interval `[0, 2**63]` (so you alias just shy of 50% of the values). Instead, you want to shift, I suspect, and then you don't need numpy in the loop at all:

```suggestion
    funcid = f"queryexpr_{hash(expr) + 2**63:x}"
```

That said, strings are hashable, so this seems like a weird way of constructing a cache key (it's somehow deliberately making it more likely that you get hash collisions and produce the wrong value).

I would have thought that this would do the trick:

```
@functools.cache
def query_compile(expr):
    name  = "queryexpr" # these are only looked up locally so names can collide
    info = query_parser(expr)
    fn = query_builder(info, name)
    args = info["args"]
    devicefn = cudf.jit(device=True)(fn)
    kernel = _wrap_query_expr(f"kernel_{name}", devicefn, args)
    info["kernel"] = kernel
    return info
```

wence- (changes requested):
It seems like there are some strange things going on in the changes in the tests?

python/cudf/cudf/tests/test_csv.py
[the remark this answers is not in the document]
shwina: `np.uint8(-1)` in particular raises a deprecation warning. But the reason I gave up on this test was because I couldn't figure out why this was happening:

```python
In [2]: np.array([-1]).astype("uint64")
Out[2]: array([18446744073709551615], dtype=uint64)

In [3]: np.array([18446744073709551615]).astype("uint64")
Out[3]: array([18446744073709551615], dtype=uint64)

In [4]: np.array([-1, 18446744073709551615]).astype("uint64")
<ipython-input-4-03014ed268fc>:1: RuntimeWarning: invalid value encountered in cast
  np.array([-1, 18446744073709551615]).astype("uint64")
Out[4]: array([18446744073709551615,                    0], dtype=uint64)
```

I've gone ahead and filtered out that warning from this test.

python/cudf/cudf/tests/test_unaops.py
```diff
@@ -79,9 +79,13 @@ def generate_valid_scalar_unaop_combos():
 
 @pytest.mark.parametrize("slr,dtype,op", generate_valid_scalar_unaop_combos())
 def test_scalar_unary_operations(slr, dtype, op):
-    slr_host = cudf.dtype(dtype).type(slr)
+    slr_host = np.array([slr])[0].astype(cudf.dtype(dtype))
     slr_device = cudf.Scalar(slr, dtype=dtype)
 
+    if op.__name__ == "neg" and np.dtype(dtype).kind == "u":
+        # TODO: what do we want to do here?
+        return
```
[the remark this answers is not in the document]
shwina: Well:

```
In [2]: -np.uint16(1)
<ipython-input-2-5156426c8f88>:1: RuntimeWarning: overflow encountered in scalar negative
  -np.uint16(1)
Out[2]: 65535
```

Should we just go ahead and ignore that warning in this test? (I've resorted to doing that in most other cases)

vyasr:
@shwina Is this PR waiting on a re-review from me and/or @wence- ?

shwina:
@vyasr just resolved the conflicts here. Yeah I think it could use a quick skim

PR #13327, opened 2023-05-10
Move `meta` calculation in `dask_cudf.read_parquet`

rjzamora:
## Description
Implements `ArrowDatasetEngine._create_dd_meta` so that a cudf-backed `meta` object is created before the full "partitioning plan" is generated. This is important in the new dask expression project (dask-expr), where we need accurate metadata **before** the full partitioning plan is generated.

## Checklist


wence- (changes requested):
Tiny nitpick

Issue #13324, opened 2023-05-09
[FEA] Apply row-wise filtering for `filters` in `cudf/dask_cudf.read_parquet`


rjzamora:
**Is your feature request related to a problem? Please describe.**
The current `cudf/dask_cudf.read_parquet` APIs accept a `filters` argument. These filters are only used to drop data at the row-group level. This means that `cudf.read_parquet(path, filters=[("x", "==", 10]))` is not guaranteed to produce the same result as `df = cudf.read_parquet(path); df[df["x"] == 10]`.

Although it makes sense to use parquet statistics to filter out data at the row-group level, I feel that cudf should enforce the filters on all rows before returning the data to the user (even if those rows still needed to be read in to memory first).

**Describe the solution you'd like**
Although it would be nice to apply filters in cuio/libcudf, it seems perfectly reasonable to simply convert DNF-formatted filter expressions into cudf/python operations, and apply those operations on the data before returning.

**Describe alternatives you've considered**
Alternative is the status quo: Passing in filters provides no guarantee that the returned data will satisfy the provided filters.

**Additional context**
The primary motivation here is the new Dask Expressions (dask-expr) library. It is **much** easier to implement a predicate-pushdown optimization if a `Filter` expression can be completely absorbed by a `ReadParquet` expression by converting the distinct filtering operation into a `ReadParquet` argument.

Issue #12512, opened 2023-01-10
[BUG] read_parquet/read_orc with filters do not filter specific rows


ayushdg:
**Describe the bug**
When using `cudf.read_parquet` or `read_orc` with the filters argument to filter out rows based on certain predicates, the methods today just filter out reading row groups (or stripes) that can be completely eliminated based on the given condition, but does return all rows from the read row groups without applying the given filters again. This behavior can be confusing to users assuming that all the relevant data has already been filtered out and is contrary to how dask, dask-cuDF and PyArrow behave today.

Example:

Data:
```
Col Name: A
Row Group 0: 1,5,1
Row Group 1: 5,5,5
```
```python
cudf.read_parquet("data", filters=[('a','!=',5)])
```
Would return 1 , 5, 1 which is all elements from RG0 (RG1 gets filtered out).
Expected output would be 1,1

**Steps/Code to reproduce bug**
```python
df = cudf.DataFrame()

In [6]: df["a"] = [1,5]*2500 + [5]*5000

In [7]: df.to_parquet("rg_test.parquet", row_group_size_rows=5000)

In [8]: cudf.read_parquet("rg_test.parquet")
[10000 rows x 1 columns]

In [9]: cudf.read_parquet("rg_test.parquet", filters=[("a", "!=", 5)])
[5000 rows x 1 columns]
```

**Expected behavior**
The 5's from row group 0 also get filtered returning only 1's, which is inline with how pyarrow, dask/dask-cudf return return the result.

**Environment overview (please complete the following information)**
 - Environment location: bare-metal
 - Method of cuDF install: conda
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**
Please run and paste the output of the `cudf/print_env.sh` script here, to gather any other relevant environment details

**Additional context**

GregoryKimball:
Hello I would like to update this issue now that we have support for libcudf ASTs in cuDF's `DataFrame.eval`. I propose that we add a filtering step to `cudf.read_parquet` if the `filters` argument is present, similar to the following approach:

```
filters = [
    [('a', '>', 7),('b', '>', 15)],
    [('a', '<', 2)],
]
df = cudf.read_parquet('test.parquet', filters=filters)

assert isinstance(filters, list) and len(filters) > 0, "Invalid filters"
if isinstance(filters[0], tuple):
    filters = [filters]
[...]

PR #13334, opened 2023-05-10
Add row-wise `post_filters` step to `read_parquet`

rjzamora:
Closes https://github.com/rapidsai/cudf/issues/13324

## Description
This PR adds a `post_filters` argument to `cudf.read_parquet`, which is set equal to the `filters` argument by default. When this argument is set, the specified DNF (disjunctive normal form) filter expression will be applied to the in-memory `cudf.DataFrame` object after IO is performed.

The overal result of this PR is that the behavior of `cudf.read_parquet` becomes more consistent with that of `pd.read_parquet` in the sense that the default result will now enforce filters at a row-wise granularity for both libraries.

### Note on the "need" for distinct `filters` and `post_filters` arguments

My hope is that `post_filters` will eventually go away. However, I added a distinct argument for two general reasons:

1. PyArrow does not yet support `"is"` and `"is not"` operands in `filters`.  Therefore, we can not pass along **all** filters from `dask`/`dask_cudf` down to `cudf.read_parquet` using the existing `filters` argument, because we rely on pyarrow to filter out row-groups (note that dask implements its own filter-conversion utility to avoid this problem). I'm hoping pyarrow will eventually adopt these comparison types (xref: https://github.com/apache/arrow/issues/34504)
2. When `cudf.read_parquet` is called from `dask_cudf.read_parquet`, row-group filtering will have already been applied. Therefore, it is convenient to specify that you only need cudf to provide the post-IO row-wise filtering step. Otherwise, we are effectively duplicating some metadata processing.

My primary concern with adding `post_filters` is the idea that row-wise filtering *could* be added at the cuio/libcudf level in the future. In that (hypothetical) case, `post_filters` wouldn't really be providing any value, but we would probably deprecate it without much pain (if any).

## Additional Context

This feature is ultimately **needed** to support general predicate-pushdown optimizations in Dask Expressions (dask-expr). This is because the high-level optimization logic becomes much simpler when a filter-based operation of a `ReadParquet` expression can be iteratively "absorbed" into the root `ReadParquet` expression.

## Checklist


rjzamora:
cc @charlesbluca (for both dask-expr and dask-sql visibility)

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +565,118 @@ def read_parquet(
...
 
+    # Can re-set the index before returning if we filter
+    # out rows from a DataFrame with a default RangeIndex
+    # (to reduce memory usage)
+    post_filters = post_filters or filters
+    reset_index = post_filters and (
+        isinstance(df.index, cudf.RangeIndex)
+        and df.index.name is None
+        and df.index.start == 0
+        and df.index.step == 1
```
wence-, line 576: Is the thought that if the index is not a default rangeindex then it must be intentional and so we must not reset it? How are we to know that the "default" rangeindex is not intentional?

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +565,118 @@ def read_parquet(
...
+                        )
+                    conjunctions.append(
+                        df[column].isna() if op == "is" else ~df[column].isna()
+                    )
+                else:
+                    # Conventional comparison operator
+                    conjunctions.append(_comparisons[op](df[column], value))
+
+            disjunctions.append(
+                operator.and_(*conjunctions)
```
wence-, line 661: `and_` has type `bool -> bool -> bool` so this only works if you have exactly two conjunctions.

You want (handles length-1 conjunctions as well):

```
disjunctions.append(functools.reduce(operator.and_, conjunctions))
```

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +565,118 @@ def read_parquet(
...
+                if len(conjunctions) > 1
+                else conjunctions[0]
+            )
+
+        return df[
+            operator.or_(*disjunctions)
+            if len(disjunctions) > 1
+            else disjunctions[0]
```
wence-, line 669: Again here, wants to be `functools.reduce(operator.or_, disjunctions)`

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +565,118 @@ def read_parquet(
...
+    )
+
+    # Apply post_filters (if any are defined)
+    df = _apply_post_filters(df, post_filters)
+
+    # Return final cudf.DataFrame
+    return df.reset_index(drop=True) if reset_index else df
+
+
+def _apply_post_filters(df, filters):
```
wence-, line 586: Can I propose a rewrite like this, which I think separates the handler logic from the conjunction/disjunction a little more clearly? WDYT?

(Probably needs some minor modifications for py 3.9 compat (with the type-annotations)).

```python
from functools import reduce, partial
import operator
import numpy as np


def handle_in(column, value):
    if not isinstance(value, (list, set, tuple)):
        raise TypeError("Value of 'in' filter must be a " "list, set, or tuple.")
    return reduce(operator.or_, (operator.eq(column, value) for v in value))


def handle_is(column, value, *, negate):
    if value not in {np.nan, None}:
        raise TypeError("Value of 'is' or 'is not' filter " "must be np.nan or None.")
    return ~column.isna() if negate else column.isna()


def _apply_post_filters(df, filters: list[tuple | list[tuple]]):
    # Apply DNF filters to an in-memory DataFrame
    #
    # Disjunctive normal form (DNF) means that the inner-most
    # tuple describes a single column predicate. These inner
    # predicates are combined with an AND conjunction into a
    # larger predicate. The outer-most list then combines all
    # of the combined filters with an OR disjunction.

    if not filters:
        # No filters to apply
        return df

    handlers = {
        "==": operator.eq,
        "!=": operator.ne,
        "<": operator.lt,
        "<=": operator.le,
        ">": operator.gt,
        ">=": operator.ge,
        "in": handle_in,
        "is": partial(handle_is, negate=False),
        "is not": partial(handle_is, negate=True),
    }

    try:
        # Disjunction loop
        #
        # All elements of `disjunctions` shall be combined with
        # an `OR` disjunction (operator.or_)
        disjunctions = []
        expressions = [f if isinstance(f, list) else [f] for f in filters]
        for expr in expressions:
            conjunction = reduce(
                operator.and_,
                (handlers[op](df[column], value) for (column, op, value) in expr),
            )
            disjunctions.append(conjunction)

        return reduce(operator.or_, disjunctions)
    except (KeyError, TypeError):
        warnings.warn(...)
        return df
```

wence-: The reduction over the disjunctions could be merged in (so no looping) but I think it's a little less readable (would be something like):
```python
        return reduce(
            operator.or_,
            (
                reduce(
                    operator.and_,
                    (handlers[op](df[column], value) for (column, op, value) in expr),
                )
                for expr in expressions
            ),
        )
```

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -541,13 +539,13 @@ def test_parquet_read_filtered_multiple_files(tmpdir):
...
-        ([("x", "==", 0), ("z", "!=", 0)], 2),
+        ([[("x", "==", 0)], [("z", "==", 0)]], 2),
         ([("x", "==", 0), ("z", "==", 0)], 0),
+        ([("x", "==", 0), ("z", "!=", 0)], 1),
         ([("y", "==", "c"), ("x", ">", 8)], 0),
-        ([("y", "==", "c"), ("x", ">=", 5)], 2),
-        ([[("y", "==", "c")], [("x", "<", 3)]], 6),
+        ([("y", "==", "c"), ("x", ">=", 5)], 1),
+        ([[("y", "==", "c")], [("x", "<", 3)]], 5),
+        ([[("y", "==", "c")], [("x", "in", (0, 9))]], 4),
```
wence-, line 548: Can you add some tests with more than two predicates being anded together (and being ored together)?

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +565,118 @@ def read_parquet(
         **kwargs,
     )
 
+    # Can re-set the index before returning if we filter
+    # out rows from a DataFrame with a default RangeIndex
+    # (to reduce memory usage)
+    post_filters = post_filters or filters
```
wence-, line 571: So if there are no `post_filters` we apply `filters` twice (because the parquet reader also does?) Otherwise only once?

wence- (changes requested):
I think some of the core logic can be tidied up a bit, and am not sure this works correctly for more than two ANDed or ORed filters.

python/cudf/cudf/io/parquet.py
rjzamora: I don't have strong feeling about how we handle the default index. I just know that pandas/pyarrow avoids carrying around a filtered index if/when the original DataFrame would have had a default index anyway.

python/cudf/cudf/io/parquet.py
rjzamora: Ah right - I did completely forget that `operator.and_`/`or_` are binops

python/cudf/cudf/io/parquet.py
rjzamora: `filters` would have already been used to drop entire row-groups, but it would not have been used for row-wise filtering.  Therefore, we assume that users will want the same filters to be applied post-IO to make sure that any previously-missed rows are filtered out.  In contrast, when dask_cudf is calling this function, they will **only** want the post-IO filtering, because dask already drops unnecessary row-groups at graph-construction time.

GregoryKimball:
(Linking the discussion in #12512)

python/cudf/cudf/io/parquet.py
rjzamora: Thanks for the suggestion. I revised the code to look more like your example (there were some minor bugs, and I could't quite get the type annotations right yet - so left them out). Wouldn't mind going with the loop-free code, but didn't get a chance to try it yet.

rjzamora:
Small update: Removed the `post_filters` argument in favor of calling `_apply_post_filters` manually in `dask_cudf`.

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +567,116 @@ def read_parquet(
...
+
+def _normalize_filters(filters: list | None) -> List[List[tuple]] | None:
+    # Utility to normalize and validate the `filters`
+    # argument to `read_parquet`
+    if filters is not None:
+        msg = (
+            f"filters must be None, or non-empty List[Tuple] "
+            f"or List[List[Tuple]]. Got {filters}"
+        )
+        if not filters or not isinstance(filters, list):
```
wence-, line 582: Is it easier to accept empty containers as "no filters" and normalise them to `None` (rather than requiring specifically `None` as empty filters).

So:
```
if filters:
   ... validate
else:
   return None
```

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +567,116 @@ def read_parquet(
...
+            raise TypeError(msg)
+
+        def _validate_predicate(item):
+            if not isinstance(item, tuple) or len(item) != 3:
+                raise TypeError(
+                    f"Predicate must be Tuple[str, str, Any], "
+                    f"got {predicate}."
+                )
+
+        filters = filters if isinstance(filters[0], list) else [filters]
```
wence-, line 592: OK, so now we definitively have a `list-of-lists`.

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +567,116 @@ def read_parquet(
...
+        for conjunction in filters:
+            if not conjunction or not isinstance(conjunction, list):
```
wence-, line 594: OK, so each entry must be a non-empty list.

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +567,116 @@ def read_parquet(
...
+                raise TypeError(msg)
+            for predicate in conjunction:
+                _validate_predicate(predicate)
```
wence-, line 597: And each entry in that non-empty list must be a 3-tuple of appropriate type.

wence-, line 597: tl;dr: nothing to do here.

Could write:
```
if not all(isinstance(item, tuple) and len(item) == 3 for item in conjunction):
   raise TypeError("Every predicate must be ...")
```
But I guess then it's hard to point at the bad one, unless one abuses the walrus operator like so:

```
if not all(isinstance(item := pred, tuple) and len(pred) == 3 for pred in conjunctions):
   raise TypeError(... item)
```
Which is kind of ugly.

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +567,116 @@ def read_parquet(
...
+
+def _handle_in(column, value, *, negate):
+    if not isinstance(value, (list, set, tuple)):
+        raise TypeError(
+            "Value of 'in' or 'not in' filter must be a list, set, or tuple."
+        )
+    if negate:
+        return reduce(operator.and_, (operator.ne(column, v) for v in value))
+    else:
+        return reduce(operator.or_, (operator.eq(column, v) for v in value))
```
wence-, line 610: Ah, I hadn't really thought when writing this, let's do:

```suggestion
    return ~column.isin(value) if negate else column.isin(value)
```

To avoid scanning over the column once per value and then reducing.

wence- (changes requested):
Mostly commentary, but a few minor cleanups.

python/cudf/cudf/io/parquet.py
bdice: Regardless, the previous code should be spelled with `any`/`all` instead of reducing with or/and. Use builtin reductions when possible! And use `==` / `!=` instead of calling the (in)equality operator literally.

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +567,116 @@ def read_parquet(
...
+            if not conjunction or not isinstance(conjunction, list):
+                raise TypeError(msg)
+            for predicate in conjunction:
+                _validate_predicate(predicate)
+
+    return filters
+
+
+def _handle_in(column, value, *, negate):
+    if not isinstance(value, (list, set, tuple)):
```
bdice, line 603: Are you looking for `collections.abc.Iterable` here? Or another ABC?

python/cudf/cudf/io/parquet.py
```diff
@@ -561,6 +567,116 @@ def read_parquet(
...
+    # (to reduce memory usage)
+    reset_index = (
+        isinstance(df.index, cudf.RangeIndex)
+        and df.index.name is None
+        and df.index.start == 0
+        and df.index.step == 1
+    )
+
+    try:
+        selection: cudf.Series = reduce(
```
bdice, line 658: Use any/all here.

bdice (changes requested):

python/cudf/cudf/io/parquet.py
wence-: That doesn't work, `any/all` want to produce a single boolean, this is producing a boolean _mask_.

python/cudf/cudf/io/parquet.py
wence-: As below, `any` and `all` want to produce a single boolean, this is producing a boolean mask column, so that can't be done (although yes, `column == v` is clearer).

python/cudf/cudf/io/parquet.py
rjzamora: We could use probably `collections.abc.Iterable`, but I was just copying the instance check used by dask (for consistency).

python/cudf/cudf/io/parquet.py
rjzamora: Right - I don't expect the number of predicates to get very large here. It seems reasonable to just call `_validate_predicate` in a loop.

python/cudf/cudf/io/parquet.py
rjzamora: I modified things to use code like `column == v`. However, just a note that there is probably no performance advantage in doing this over the simpler `operator` approach. We still need to wrap the conditional in another function to reduce.

python/cudf/cudf/io/parquet.py
bdice: Great, that's a good reason to keep the status quo here.

python/cudf/cudf/io/parquet.py
bdice: Oh gosh, I apologize. I didn't realize these were acting on Series. 😪 What you have now is reasonable. I'll re-review this.

bdice: I made two significant mistakes in my previous review comment, for which I am terribly sorry.

1. I was perfectly fine with the use of `operator.lt` in the handler dispatch. The only thing I wanted to replace was the reduction. I wasn't clear about that in my comment - the handler dispatch is an ideal use of `operator`.
2. My proposal of replacing `reduce(operator.or_, ...))` with `any(...)` is wrong. It was not a logical `or` that you intended, but _bitwise or_ `|` [which is the action of `operator.or_`](https://docs.python.org/3/library/operator.html#operator.or_). I mistakenly thought it was representing a logical or, which is the reduction operator of `any`.

I'd consider going back to the previous design here, since my suggestions made it worse. 🙃 It was partly a misreading of the code, and partly incorrect information about `operator` in my brain (which I didn't realize until digging deeper).

python/cudf/cudf/io/parquet.py
bdice: I made an error in my comment, discussed here: https://github.com/rapidsai/cudf/pull/13334#discussion_r1194493655

python/cudf/cudf/io/parquet.py
rjzamora: >for which I am terribly sorry.

:)  No worries - I really appreciate having your eyes on this @bdice !

>I'd consider going back to the previous design here

Sure - I agree that operator makes for this case. I can roll that piece back.

bdice (approved):
Thanks @rjzamora. I learned a thing or two from this review! Appreciate your effort here.

wence- (approved):
Thanks @rjzamora !

shwina:
@rjzamora do you expect the work done here to be roughly the same for `orc`?

rjzamora:
>do you expect the work done here to be roughly the same for orc?

Yes, the `_apply_post_filters` code added here should be independent of the IO mechanism, since it is a post-IO filtering operation. We could easily move it out of the `parquet.py` file and use it for ORC (and even CSV).

shwina:
Thanks, I'll go ahead and open a PR to do the same for ORC then!

PR #13335, opened 2023-05-11
[Java] Purge non-empty nulls when setting validity 

razajafri:
## Description
When setting the validity buffer in `mergeAndSetValidity` method, we check to see two things 
1. Purge any non-empty nulls that may have arise due to setting values to null which were previously non-null
2. If we override the existing null values in the validity buffer then throw an IllegalStateException. 

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/ColumnView.java
+++ b/java/src/main/java/ai/rapids/cudf/ColumnView.java
@@ -33,6 +33,7 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
   static {
     NativeDepsLoader.loadNativeDeps();
   }
+  private static boolean IF_ELSE = Boolean.getBoolean("ai.rapids.cudf.test.ifelse");
 
   public static final long UNKNOWN_NULL_COUNT = -1;
 
@@ -814,17 +815,66 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
   }
 
   /**
+   * Create a deep copy of the column while replacing the validity mask. The resultant validity mask
+   * is the bitwise merge of the validity masks in the columns given as arguments. Validity mask is true
+   * when data is not null and false when it's null.
+   *
+   * The result will be sanitized to not contain any non-empty nulls in case of nested types
+   *
+   * @param mergeOp binary operator (BITWISE_AND and BITWISE_OR only)
+   * @param columns array of columns whose null masks are merged, must have identical number of rows.
+   * @return the new ColumnVector with merged null mask.
+   * @throws IllegalStateException if existing nulls are set to valid data in the validity buffer
+   *
+   */
+  public final ColumnVector mergeAndSetValidityUsingIfElse(BinaryOp mergeOp, ColumnView... columns) {
+    System.out.println("************ if else *******");
+    assert mergeOp == BinaryOp.BITWISE_AND || mergeOp == BinaryOp.BITWISE_OR : "Only BITWISE_AND and BITWISE_OR supported right now";
+
+    if (columns.length == 0) {
+      return copyToColumnVector();
+    }
+
+    ColumnVector shouldNotBeNull = columns[0].isNotNull();
+    try {
+      for (int i = 1; i < columns.length; i++) {
+        try (ColumnVector tmp = columns[i].isNotNull();
+             ColumnVector toClose = shouldNotBeNull) {
+          shouldNotBeNull = toClose.binaryOp(mergeOp, tmp, DType.BOOL8);
+        }
+      }
+
+      // Now make sure the original nulls are kept
+      try (ColumnVector originalIsNotNull = isNotNull();
+           ColumnVector finalShouldNotBeNull = shouldNotBeNull.and(originalIsNotNull);
+           Scalar nullScalar = Scalar.fromNull(this.getType())) {
+        return finalShouldNotBeNull.ifElse(this, nullScalar);
+      }
+    } finally {
+      shouldNotBeNull.close();
+    }
+  }
+
+  /**
    * Create a deep copy of the column while replacing the null mask. The resultant null mask is the
    * bitwise merge of null masks in the columns given as arguments.
    * The result will be sanitized to not contain any non-empty nulls in case of nested types
    *
-   * @param mergeOp binary operator, currently only BITWISE_AND is supported.
+   * @param mergeOp binary operator (BITWISE_AND and BITWISE_OR only)
    * @param columns array of columns whose null masks are merged, must have identical number of rows.
    * @return the new ColumnVector with merged null mask.
    * @throws IllegalStateException if existing nulls are set to valid data in the validity buffer
    *
    */
   public final ColumnVector mergeAndSetValidity(BinaryOp mergeOp, ColumnView... columns) {
+    if (IF_ELSE) {
+      return mergeAndSetValidityUsingIfElse(mergeOp, columns);
+    } else {
+      return mergeAndSetValidityOld(mergeOp, columns);
+    }
+  }
+
+  public final ColumnVector mergeAndSetValidityOld(BinaryOp mergeOp, ColumnView... columns) {
     assert mergeOp == BinaryOp.BITWISE_AND || mergeOp == BinaryOp.BITWISE_OR : "Only BITWISE_AND and BITWISE_OR supported right now";
     long[] columnViews = new long[columns.length];
     long size = getRowCount();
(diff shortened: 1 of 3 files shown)


razajafri:
Tests pass locally when I run `ci/test_java.sh` in `rapidsai/ci:cuda11.8.0-ubuntu22.04-py3.10`

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running ai.rapids.cudf.ColumnViewNonEmptyNullsTest
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.066 s - in ai.rapids.cudf.ColumnViewNonEmptyNullsTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  03:17 min
[INFO] Finished at: 2023-05-11T05:38:07Z
[INFO] ------------------------------------------------------------------------

```

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -832,7 +835,35 @@ public final ColumnVector mergeAndSetValidity(BinaryOp mergeOp, ColumnView... co
...
-    return new ColumnVector(bitwiseMergeAndSetValidity(getNativeView(), columnViews, mergeOp.nativeId));
+    ColumnVector potentialRes = new ColumnVector(bitwiseMergeAndSetValidity(getNativeView(), columnViews, mergeOp.nativeId));
+
+    /**
+     * We need to sanitize the vector to make sure of two things
+     * 1. Nulls set in the new vector don't have non-empty nested types
+     * 2. Nulls in 'this' stay nulls otherwise bad things will happen
+     */
+    ColumnVector finalRes;
+    if (potentialRes.getType().isNestedType() || potentialRes.getType().hasOffsets() && potentialRes.hasNonEmptyNulls()) {
```
revans2, line 846: hasNonEmptyNulls is expensive.  Would be better to use https://github.com/rapidsai/cudf/blob/6f3f50721e585b55bd263cc36926eb2c99c5f811/cpp/include/cudf/detail/copy.hpp#LL270C6-L270C29

But we don't have JNI for it.  Might be simpler to just put this part into C++ for the bitwiseMergeAndSetValidity.

Also if we return a ColumnVector with non-empty nulls in it from C++, won't an assertion error be throw? https://github.com/rapidsai/cudf/blob/6f3f50721e585b55bd263cc36926eb2c99c5f811/java/src/main/java/ai/rapids/cudf/ColumnView.java#L55

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -816,10 +816,13 @@ public final ColumnVector normalizeNANsAndZeros() {
   /**
    * Create a deep copy of the column while replacing the null mask. The resultant null mask is the
    * bitwise merge of null masks in the columns given as arguments.
+   * The result will be sanitized to not contain any non-empty nulls in case of nested types
    *
    * @param mergeOp binary operator, currently only BITWISE_AND is supported.
```
revans2, line 821: nit: Code says BITWISE_OR is supported, but docs to not.  (This is not your code, just noticed it)

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -832,7 +835,35 @@ public final ColumnVector mergeAndSetValidity(BinaryOp mergeOp, ColumnView... co
...
+    if (potentialRes.getType().isNestedType() || potentialRes.getType().hasOffsets() && potentialRes.hasNonEmptyNulls()) {
+      // purge non-empty nulls in the vector
+      finalRes = potentialRes.purgeNonEmptyNulls();
+      potentialRes.close();
+    } else {
+      finalRes = potentialRes;
+    }
+
+    // Verify the nulls in the original vector remain null
+    if (getNullCount() > 0) {
```
revans2, line 855: This feels kind of expensive to do. The point of this API originally was just to do something that would be very fast, and it has slowly changed over time to end up being rather expensive.  First off we never have to do this if 

Can we benchmark this to see what the performance is?  Because we are close to burndown for CUDF can you also file a follow on issue to try and find a faster way to do something similar?

If/else and coalesce all provide similar results but we know that they are safe.  Perhaps switching to them would be better.

```
// But switch the and to an or depending on the operation being done.
try (ColumnVector shouldOutputNotBeNull = columns[0].isNotNull()) {
  for (int i = 1; i < columns.length; i++) {
    try (ColumnVector tmpIsNotNull = column[i].isNotNull()) {
      ColumnVector tmpCombined = tmpIdNotNull.and(shouldOutputNotBeNull)
      shouldOutputNotBeNull.close()
      shouldOutputNotBeNull = tmpCombined
    }
  }
  return shouldOutputNotBeNull.ifElse(this, null_scalar)
}
```

java/src/test/java/ai/rapids/cudf/ColumnViewNonEmptyNullsTest.java
```diff
@@ -41,11 +42,10 @@ void testAndNullReconfigureNulls() {
          ColumnVector intResult = v1.mergeAndSetValidity(BinaryOp.BITWISE_AND, v0);
          ColumnVector v2 = ColumnVector.fromStrings("0", "100", "1", "2", "MIN_VALUE", "3");
          ColumnVector stringResult = v2.mergeAndSetValidity(BinaryOp.BITWISE_AND, v0, v1);
-         ColumnVector stringExpected = ColumnVector.fromStrings("0", "100", null, null, "MIN_VALUE", null);
-         ColumnVector noMaskResult = v2.mergeAndSetValidity(BinaryOp.BITWISE_AND)) {
+         ColumnVector stringExpected = ColumnVector.fromStrings("0", "100", null, null, "MIN_VALUE", null)) {
       assertColumnsAreEqual(v0, intResult);
       assertColumnsAreEqual(stringExpected, stringResult);
-      assertColumnsAreEqual(v2, noMaskResult);
+      assertThrows(IllegalStateException.class, () -> v2.mergeAndSetValidity(BinaryOp.BITWISE_AND));
```
revans2, line 48: I am really confused here.  If you don't pass in any children, then there is nothing to change about the validity. Why is this an exception and not a special case copy?

java/src/test/java/ai/rapids/cudf/ColumnViewNonEmptyNullsTest.java
```diff
@@ -41,11 +42,10 @@ void testAndNullReconfigureNulls() {
          ColumnVector intResult = v1.mergeAndSetValidity(BinaryOp.BITWISE_AND, v0);
          ColumnVector v2 = ColumnVector.fromStrings("0", "100", "1", "2", "MIN_VALUE", "3");
          ColumnVector stringResult = v2.mergeAndSetValidity(BinaryOp.BITWISE_AND, v0, v1);
```
revans2, line 44: nit: Can we get some documentation updates about what BITWISE_AND and BITWISE_OR do?  It is not really intuitive because we often thing about this in terms of a null mask, but it is a validity mask, so BITWISE_AND is the same as

```
try (ColumnVector shouldOutputNotBeNull = columns[0].isNotNull()) {
  for (int i = 1; i < columns.length; i++) {
    try (ColumnVector tmpIsNotNull = column[i].isNotNull()) {
      ColumnVector tmpCombined = tmpIdNotNull.and(shouldOutputNotBeNull)
      shouldOutputNotBeNull.close()
      shouldOutputNotBeNull = tmpCombined
    }
  }
  return shouldOutputNotBeNull.ifElse(this, null_scalar)
}
```

That might be a bit long though. If you feel it is too long feel free to ignore this. Or if you have a better way to describe it that works too.

Perhaps just something where we change it from null masks to validity buffers in the comments.

```
 /**
   * Create a deep copy of the column while replacing the validity buffers. The resultant validity buffer is the
   * bitwise merge of validity in the columns given as arguments. Please note that validity is the opposite of 
   * a null.  So BITWISE_AND will put a null in the output if either input values is a null, and BITWISE_OR
   * will put a null value in the output only if all input values are null.
   * The result will be sanitized to not contain any non-empty nulls in case of nested types
   *
```

java/src/test/java/ai/rapids/cudf/ColumnViewNonEmptyNullsTest.java
```diff
@@ -60,14 +60,13 @@ void testOrNullReconfigureNulls() {
...
-         ColumnVector stringExpected = ColumnVector.fromStrings("0", "100", "1", "2", "MIN_VALUE", null);
-         ColumnVector noMaskResult = v3.mergeAndSetValidity(BinaryOp.BITWISE_OR)) {
+         ColumnVector stringExpected = ColumnVector.fromStrings("0", "100", "1", "2", "MIN_VALUE", null)) {
       assertColumnsAreEqual(v0, intResultV0);
       assertColumnsAreEqual(v1, intResultV0V1);
       assertColumnsAreEqual(v1, intResultMulti);
       assertColumnsAreEqual(v2, intResultv0v1v2);
       assertColumnsAreEqual(stringExpected, stringResult);
-      assertColumnsAreEqual(v3, noMaskResult);
+      assertThrows(IllegalStateException.class, () ->  v3.mergeAndSetValidity(BinaryOp.BITWISE_OR));
```
revans2, line 69: Same comments here...

razajafri:
I am running benchmarks on a different version of this implementation based on suggestions from here. Will post an update soon

java/src/main/java/ai/rapids/cudf/ColumnView.java
ttnghia: cudf C++ API now guarantees that the output lists columns will never have non-empty nulls. The only exception is in-place set validity API.

razajafri:
I have run some numbers on my local workstation using different implementations. The script just reads a parquet file with a list[int] column and casts it to a string. The casting in turn calls the `mergeAndSetValidity` method 

|Baseline|Ifelse|Binop|CPP Modification|Input Size|
|-|-|-|-|-|
|5,975 ms| 7,025 ms| 6,013 ms| 5,936 ms| 5.8 MiB|
|26,365 ms| 35,874 ms| 26,335 ms| 26,284 ms| 52.43 MiB|

It looks like the modifications in this PR, listed as `cpp modification` above, has no performance degradation.

revans2 (approved):

PR #13337, opened 2023-05-11
cuDF numba cuda 12 updates

brandon-b-miller:
Summary of changes:

- Removed some old code that is only used for `numba<0.54` which hasn't been supported for a while now.
- Removed some old code that is only used when `cubinlinker` is not present, which is has been a hard requirement for a while now as well. 
- Created a file `_numba.py` and moved into this file all of the machinery used to configure numba upon cuDF import. This includes functions for determining which toolkit version was used to build the PTX file our UDFs rely on as well as the functions for potentially putting numba into MVC mode if necessary.
- Created a file `_ptxcompiler.py` which vendors the driver/runtime version checking machinery from ptxcompiler in case we're in a cuda 12 environment that doesn't have it
- Changed the code to issue a warning in cuda 12+ MVC situations that the library will likely not work
- The version of the toolkit used to determine if MVC is required is now determined from the `cc=60` PTX file which is [always built](https://github.com/rapidsai/cudf/blob/branch-23.06/python/cudf/udf_cpp/CMakeLists.txt#L85-L87). This is to avoid needing to query the device compute capability through numba's `cuda` module. This needs to be avoided during numba's setup because if `numba.cuda` is imported before numba's config is modified, the config options will have no effect. 

Closes https://github.com/rapidsai/cudf/issues/13351
Closes https://github.com/rapidsai/cudf/issues/13339


python/cudf/cudf/utils/_numba_setup.py
```diff
@@ -0,0 +1,144 @@
...
+    else:
+        return None
+
+
+def _get_ptx_file(path, prefix):
+    if "RAPIDS_NO_INITIALIZE" in os.environ:
+        # cc=60 ptx is always built
+        cc = int(os.environ.get("STRINGS_UDF_CC", "60"))
+    else:
+        from numba import cuda
```
brandon-b-miller, line 41: Runtime imports can be expensive but importing `cuda` must be avoided as a side effect of importing the module in which this function resides. Another option could be passing the `cuda` module as an optional argument here.

python/cudf/cudf/utils/_numba_setup.py
```diff
@@ -0,0 +1,148 @@
...
+        from ptxcompiler.patch import NO_DRIVER, safe_get_versions
+
+        versions = safe_get_versions()
+        if versions != NO_DRIVER:
+            driver_version, runtime_version = versions
+            ptx_toolkit_version = _get_cuda_version_from_ptx_file(path)
+            maybe_patch_numba_linker(
+                driver_version, runtime_version, ptx_toolkit_version
+            )
+    except ImportError:
```
bdice, line 89: Do this except immediately after the import, and return instead of passing. Then unindent all following logic.

python/cudf/cudf/utils/_numba_setup.py
```diff
@@ -0,0 +1,148 @@
...
+        pass
+
+
+def maybe_patch_numba_linker(
+    driver_version, runtime_version, ptx_toolkit_version
+):
+    # Numba thinks cubinlinker is only needed if the driver is older than
+    # the ctk, but when PTX files are present, it might also need to patch
```
bdice, line 97: Just wanting to be precise in our comments here. Do we specifically mean the runtime rather than the “toolkit”?
```suggestion
    # the CUDA runtime, but when PTX files are present, it might also need to patch
```

python/cudf/cudf/utils/_numba_setup.py
```diff
@@ -0,0 +1,148 @@
...
+    # because those PTX files may newer than the driver as well
```
bdice, line 98:
```suggestion
    # because those PTX files may be compiled by a CUDA version that is newer than the driver as well
```

python/cudf/cudf/utils/_numba_setup.py
```diff
@@ -0,0 +1,148 @@
...
+        else:
+            raise ValueError("Could not read CUDA version from ptx file.")
+    version = ver_line.strip("\n").split(" ")[1]
+    # from ptx_docs/release_notes above:
+    ver_map = {
+        "7.5": (11, 5),
+        "7.6": (11, 6),
+        "7.7": (11, 7),
+        "7.8": (11, 8),
+        "8.0": (12, 0),
```
bdice, line 139: Do we need to expand this to include 12.1? Users could run CUDA 12 pip wheels on 12.1.

bdice (changes requested):
Thanks for doing this. I have attached  comments.

python/cudf/cudf/utils/_numba_setup.py
bdice: (This follows the principle that “try” should have a minimal scope of errors it could catch.)

python/cudf/cudf/__init__.py
```diff
@@ -80,35 +80,19 @@
     read_text,
 )
 from cudf.options import describe_option, get_option, set_option
+from cudf.utils._numba_setup import _setup_numba
```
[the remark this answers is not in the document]
brandon-b-miller: Addressed this 👍

python/cudf/cudf/utils/_numba_setup.py
```diff
@@ -0,0 +1,148 @@
...
+def _setup_numba():
+    """
+    Configure numba for use with cuDF. This consists of potentially
+    putting numba into enhanced compatibility mode based on the user
+    driver and runtime versions as well as the version of the cuda
+    toolkit used to build the PTX files shipped with the user cuDF
+    package. It also sets any other config options within numba that
+    are desired for cuDF's operation.
+    """
+    _setup_numba_linker(ANY_PTX_FILE)
```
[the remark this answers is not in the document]
brandon-b-miller: This was originally a glob followed by zero indexing the list. It could be again if we wanted it to be...

python/cudf/cudf/utils/_numba_setup.py
```diff
@@ -0,0 +1,148 @@
...
+    if filtered_archs:
+        return max(filtered_archs, key=lambda y: y[0])
+    else:
+        return None
+
+
+def _get_ptx_file(path, prefix):
+    if "RAPIDS_NO_INITIALIZE" in os.environ:
+        # cc=60 ptx is always built
+        cc = int(os.environ.get("STRINGS_UDF_CC", "60"))
```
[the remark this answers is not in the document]
brandon-b-miller: This env var can be used to set the PTX file to be used from dask without having to query it through numba, which can interfere with network bringup. See https://github.com/rapidsai/cudf/pull/12056 for details. 

Is there any concern that renaming this right now could break existing workflows?

python/cudf/cudf/utils/_numba_setup.py
brandon-b-miller: I agree with this sentiment. Similarly to above however I think it's better to tackle this particular update separately as this function is a move. That said the change should be simple and I can commit to having it in for this release.

python/cudf/cudf/utils/_numba_setup.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    Configure numba for use with cuDF. This consists of potentially
+    putting numba into enhanced compatibility mode based on the user
+    driver and runtime versions as well as the version of the cuda
+    toolkit used to build the PTX files shipped with the user cuDF
+    package. It also sets any other config options within numba that
+    are desired for cuDF's operation.
+    """
+    _setup_numba_linker(ANY_PTX_FILE)
+    config.CUDA_LOW_OCCUPANCY_WARNINGS = 0
```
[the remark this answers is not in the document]
brandon-b-miller: This will require a few internal updates that are tangential to this PR. I raised an issue about this. https://github.com/rapidsai/cudf/issues/13351

dependencies.yaml
```diff
@@ -350,7 +350,7 @@ dependencies:
         packages:
           - cachetools
           - cuda-python>=11.7.1,<12.0
-          - &numba numba>=0.56.4,<0.57
+          - &numba numba>=0.57
```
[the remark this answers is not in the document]
vyasr: I would recommend removing the actual numba update from this PR and simply merging the rest when we're confident that everything works. We don't even know for sure that numba 0.57 will make it in time for the 23.06 RAPIDS release, and I don't want this PR lingering waiting for that. I'd prefer to get the fixes merged so that the update is painless when we're ready for it.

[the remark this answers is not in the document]
brandon-b-miller: I don't think this PR will pass tests without numba 0.57, because without it numba won't be be set into CEC/MVC mode via the config option.

python/cudf/cudf/utils/_numba_setup.py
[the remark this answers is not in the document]
wence-: Thanks.

python/cudf/cudf/utils/_setup_numba.py
```diff
@@ -0,0 +1,157 @@
...
+    driver_version, runtime_version, ptx_toolkit_version
+):
+    # Numba thinks cubinlinker is only needed if the driver is older than
+    # the CUDA runtime, but when PTX files are present, it might also need
+    # to patch because those PTX files may be compiled by a CUDA version
+    # that is newer than the driver as well
+    if (driver_version < ptx_toolkit_version) or (
+        driver_version < runtime_version
+    ):
+        config.NUMBA_CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY = 1
```
wence-, line 111: If we are going to pause this PR for numba 0.57 to be available, then I would prefer that we refactor to avoid this kind of "action-at-a-distance" here (rather than as a followup).

OTOH, if that will hold things up too much, let's not.

bdice: @wence- Great observation -- and thanks @brandon-b-miller for the context. I agree this would be nice to do contextually for the properties that we can control. Only occupancy warnings is sufficient for me.

brandon-b-miller: added something in https://github.com/rapidsai/cudf/pull/13337/commits/96b6f01fda82f3dcc074959fc51975d0cdc3c881

python/cudf/cudf/core/udf/utils.py
```diff
@@ -306,6 +253,16 @@ def _compile_or_get(
...
     """
+    # runtime check for CEC mode which is disabled for CUDA 12 for now
+    if cuda.cudadrv.driver.get_version() == (
+        12,
+        0,
+    ) and cuda.cudadrv.runtime.get_version() > (12, 0):
+        raise ValueError(
+            "Minor version compatibility not yet supported for "
+            "CUDA driver versions newer than 12.0"
+        )
```
vyasr, line 264: I believe this block resolves https://github.com/rapidsai/cudf/issues/13339, right? If so, can you update the PR description accordingly?

python/cudf/cudf/utils/_setup_numba.py
```diff
@@ -0,0 +1,157 @@
...
+
+def _setup_numba():
+    """
+    Configure numba for use with cuDF. This consists of potentially
+    putting numba into enhanced compatibility mode based on the user
+    driver and runtime versions as well as the version of the CUDA
+    Toolkit used to build the PTX files shipped with the user cuDF
+    package. It also sets any other config options within numba that
+    are desired for cuDF's operation.
+    """
```
vyasr, line 20: This docstring probably belongs in `_setup_numba_linker` more than here, no?

python/cudf/cudf/utils/_setup_numba.py
```diff
@@ -0,0 +1,157 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+import glob
+import os
+
+from numba import config
+from numba.cuda.cudadrv.driver import Linker
```
vyasr, line 7: In `__init__`.py we say that `_setup_numba` needs to be executed before `from numba import cuda` in order to ensure that configuration is set before the import. Doesn't this import cause problems then because it will execute the contents of `numba/cuda/__init__.py`?

vyasr (changes requested):
Couple of small questions left.

python/cudf/cudf/utils/_setup_numba.py
brandon-b-miller: Ah, this was true for the environment variable method in numba 0.57, but not true with the linker patching method. I'll update the comment.

python/cudf/cudf/core/udf/utils.py
brandon-b-miller: Done!

vyasr (approved):

python/cudf/cudf/__init__.py
```diff
@@ -80,35 +80,17 @@
...
-    from cubinlinker.patch import patch_numba_linker_if_needed
-except ImportError:
-    pass
-else:
-    # Patch Numba to support CUDA enhanced compatibility.
-    # cuDF requires a stronger set of conditions than what is
-    # checked by patch_numba_linker_if_needed due to the PTX
-    # files needed for JIT Groupby Apply and string UDFs
-    from cudf.core.udf.utils import _PTX_FILE, _setup_numba_linker
+_setup_numba()
```
bdice, line 87: Any magic in the import order is usually good to explicitly describe for readers.

```suggestion
# Must be called before importing numba.cuda.
_setup_numba()
```

Is it dangerous / invalid if a user imports `numba.cuda` in their own code before importing cudf?
- If so, will users clearly understand the problem (i.e. is a helpful error raised)?
- If not, why is this special import ordering needed? (Could it be written above, with the other `numba` import?)

python/cudf/cudf/utils/_setup_numba.py
```diff
@@ -0,0 +1,160 @@
...
+    cuda_ver = ver_map.get(version)
+    if cuda_ver is None:
+        raise ValueError(
+            f"Could not map PTX version {version} to a CUDA version"
+        )
+
+    return cuda_ver
+
+
+class CUDFNumbaConfig:
```
bdice, line 154: I don't like that this class is in a file named `_setup_numba`. It's not something used at import time, but at runtime when calling Numba. Should this be in a "utils"? Maybe rename this file to `utils/_numba.py` since it contains both import logic and runtime logic? Otherwise I would like the scope of `_setup_numba.py` to _only_ be import-time logic as implied by the name.

Also make it private.
```suggestion
class _CUDFNumbaConfig:
```

python/cudf/cudf/tests/test_dataframe_copy.py
```diff
@@ -159,8 +160,8 @@ def test_kernel_deep_copy():
     gdf = DataFrame.from_pandas(pdf)
     cdf = gdf.copy(deep=True)
     sr = gdf["b"]
-
-    add_one[1, len(sr)](sr._column.data_array_view(mode="write"))
+    with CUDFNumbaConfig():
+        add_one[1, len(sr)](sr._column.data_array_view(mode="write"))
```
bdice, line 164: This test surprises me. I expected that `CUDFNumbaConfig` would be treated as an implementation detail and thus not needed in our tests. (Ideally it should be `_CUDFNumbaConfig` to make it private.)

Do we _need_ to be testing a raw `@cuda.jit` function with reference to the private `_column` member? Is testing `apply` with a UDF the same thing, or does it matter that we are testing a raw column with a raw Numba JIT kernel?

python/cudf/cudf/__init__.py
vyasr: My interpretation of @brandon-b-miller's response [here](https://github.com/rapidsai/cudf/pull/13337#discussion_r1197041433) was that in fact there is no longer an explicit ordering issue because we are using the numba config API rather than an environment variable. If there is still an ordering issue then I remain confused about my question in the other thread because it seems like we would be breaking the ordering there.

python/cudf/cudf/core/udf/utils.py
```diff
@@ -306,6 +253,16 @@ def _compile_or_get(
     we then obtain the return type from that separate compilation and
     use it to allocate an output column of the right dtype.
     """
+    # runtime check for CEC mode which is disabled for CUDA 12 for now
```
gmarkall, line 256: Is checking here sufficient to catch an attempt to use CEC in enough cases? For example, indexing a series will result in a kernel launch that will elide this check and fail with the PTX version error. Running:

```
NUMBA_CUDA_LOG_LEVEL=DEBUG python -c "import cudf; cudf.Series([1, 2, 3])[0]"
```

gives output that includes:

```
== CUDA [1913] DEBUG -- call driver api: cuLaunchKernel
```

brandon-b-miller: @gmarkall this is happening because various uses of `iloc` within cuDF leverage a numba kernel [here](https://github.com/rapidsai/cudf/blob/branch-23.06/python/cudf/cudf/utils/cudautils.py#L84). 

It didn't occur to me until now that numba may need to be in CEC mode to launch kernels at all if the driver is from cuda 12.0 and the ctk is version 12.1, I assume this is because it's attempting to link PTX generated by the version of NVRTC shipped with the 12.1 ctk.

I'm a little worried this could be a bigger issue because not having working indexing more or less means that the library is broken entirely in the 12.0/12.1 configuration. 

cc @shwina @vyasr

python/cudf/cudf/utils/_numba_setup.py
gmarkall: The toolkit component of concern is NVVM. We only use the runtime to check what version NVVM is (assuming if we find the runtime, that it's the same version as NVVM) because NVVM provides no way to check what version it is. I find it less confusing to refer to "the toolkit" as it's the collection of components including the runtime and NVVM, amongst other things.

python/cudf/cudf/utils/_numba_setup.py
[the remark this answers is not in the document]
gmarkall: As an aside, it feels like there should be a kwarg for the jit decorator to indicate that it's OK to call a given kernel with low occupancy. I've created a Numba issue for this feature request: https://github.com/numba/numba/issues/8970.

python/cudf/cudf/utils/_setup_numba.py
```diff
@@ -0,0 +1,160 @@
...
+    if regular_result is None:
+        raise RuntimeError(
+            "This cuDF installation is missing the necessary PTX "
+            f"files that are <={cc}."
+        )
+    else:
+        return regular_result[1]
+
+
+def _setup_numba_linker(path):
```
gmarkall, line 73: Should this be

```suggestion
def _setup_numba():
```

and remove the definition above? I found it hard to see why the `CC_60_PTX_FILE` was passed into this function.

python/cudf/cudf/utils/_setup_numba.py
```diff
@@ -0,0 +1,160 @@
...
+        if driver_version < (12, 0):
+            ptx_toolkit_version = _get_cuda_version_from_ptx_file(path)
+            # Numba thinks cubinlinker is only needed if the driver is older
+            # than the CUDA runtime, but when PTX files are present, it might
+            # also need to patch because those PTX files may be compiled by
+            # a CUDA version that is newer than the driver as well
+            if (driver_version < ptx_toolkit_version) or (
+                driver_version < runtime_version
+            ):
+                if _numba_version_ok:
```
gmarkall, line 104: If the Numba version is not OK (e.g. if it's Numba 0.57), should we be setting the config variable to enable MVC here instead: https://numba.readthedocs.io/en/latest/cuda/minor_version_compatibility.html#enabling-mvc-support ?

gmarkall: (I may need to do another ptxcompiler release now that https://github.com/rapidsai/ptxcompiler/pull/32 is merged)

gmarkall (changes requested):
Some additional comments / questions on the diff.

python/cudf/cudf/core/udf/utils.py
gmarkall: Yes, it does need to be in CEC mode to launch kernels generated with a 12.1 toolkit on a 12.0 driver at all. But I think in past discussions we've concluded that we're not supporting CEC at all for CUDA 12 in 23.06. My concern here was that we'll end up giving the user the slightly cryptic message about an unsupported PTX version in this case, rather than the helpful message you've added below in this function.

brandon-b-miller: If that's the case then I would imagine we need a little bit more then a runtime check here, we really need cuDF to throw a CEC related error whenever theres _any_ kind of attempt at usage beyond just importing the library.

python/cudf/cudf/tests/test_dataframe_copy.py
brandon-b-miller: These seem like pretty old tests that are responsible for making sure shallow/deep copies behave as expected when some kind of action is taken that would modify the data inplace, in this case it looks like a numba kernel was chosen. I think I can sneak in a PR to update these.

brandon-b-miller: This was resolved separately in https://github.com/rapidsai/cudf/pull/13385

python/cudf/cudf/__init__.py
brandon-b-miller: So this is still an item of concern because we've pivoted back to putting numba 0.57 into this PR. Currently the call to `_setup_numba` is placed very early on in cudf's `__init__` even before `validate_setup()` is called, as `validate_setup()` calls `rmm._gpu.runtimeGetVersion()` which actually imports `numba.cuda` as a side effect.

The order of imports _does_ matter here. In addition there's @bdice 's observation that a user importing `numba.cuda` somewhere in their workflow _before_ `cudf` is imported is likely to break cuDF. So certainly this is an issue we need to come to some consensus about before merging. 

One option I can think of that could work is detecting this case dynamically somehow and issuing a warning to the user that instructs them to set `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY=1` in their environment. This should set the variable globally before the user's import.

python/cudf/cudf/__init__.py
```diff
@@ -1,7 +1,12 @@
 # Copyright (c) 2018-2023, NVIDIA CORPORATION.
 
+# this must be called before numba.cuda is imported, because
```
bdice, line 3:
```suggestion
# _setup_numba _must be called before numba.cuda is imported, because
```

python/cudf/cudf/__init__.py
```diff
@@ -1,7 +1,12 @@
...
+# it sets the numba config variable responsible for enabling
+# MVC. Setting it after importing cuda has no effect.
```
bdice, line 5:
```suggestion
# Minor Version Compatibility. Setting it after importing numba.cuda has no effect.
```

python/cudf/cudf/core/udf/utils.py
```diff
@@ -306,6 +253,8 @@ def _compile_or_get(
...
```
bdice, line 256: Avoid acronyms here for clarity.
```suggestion
    # runtime check for CUDA Enhanced Compatibility mode which is disabled for CUDA 12 for now
```

python/cudf/cudf/utils/_numba.py
```diff
@@ -0,0 +1,196 @@
...
+import subprocess
+import sys
+import warnings
+
+from numba import config
+
+CC_60_PTX_FILE = os.path.dirname(__file__) + "/../core/udf/shim_60.ptx"
+_NO_DRIVER = (math.inf, math.inf)
+
+CMD = """\
```
bdice, line 15: Use a descriptive name.
```suggestion
CHECK_CUDA_VERSION_CMD = """\
```

python/cudf/cudf/utils/_numba.py
```diff
@@ -0,0 +1,196 @@
...
+    based on the user driver and runtime versions as well as the
+    version of the CUDA Toolkit used to build the PTX files shipped
+    with the user cuDF package.
+    """
+    try:
+        # By default, ptxcompiler will not be installed with CUDA 12
+        # packages. This is ok, because in this situation putting
+        # numba in enhanced compatibility mode is not necessary.
+        from ptxcompiler.patch import NO_DRIVER, safe_get_versions
+    except ModuleNotFoundError:
```
bdice, line 109: To my knowledge, there's nothing prohibiting a user from installing `ptxcompiler` via pip or conda when using CUDA 12. If the absence of `ptxcompiler` is meant to be a proxy for using CUDA 12, we need a different strategy here.

python/cudf/cudf/utils/_numba.py
```diff
@@ -0,0 +1,196 @@
...
+    with open(path) as ptx_file:
+        for line in ptx_file:
+            if line.startswith(".version"):
+                ver_line = line
+                break
+        else:
+            raise ValueError("Could not read CUDA version from ptx file.")
+    version = ver_line.strip("\n").split(" ")[1]
+    # from ptx_docs/release_notes above:
+    ver_map = {
```
bdice, line 173: How is the list of versions decided on? We support older runtimes and drivers than 11.5. We need to document how this list is made and when it should be modified.

bdice (changes requested):

python/cudf/cudf/utils/_numba.py
```diff
@@ -0,0 +1,196 @@
...
+print(f'{drv_major} {drv_minor} {run_major} {run_minor}')
+"""
+
+
+def _get_versions():
+    """
+    This function is mostly vendored from ptxcompiler and is used
+    to check the system CUDA driver and runtime versions in its absence.
+    """
+    cp = subprocess.run([sys.executable, "-c", CMD], capture_output=True)
```
bdice, line 32: Does ptxcompiler use a subprocess currently? I am surprised by this because this tends to be costly. If possible, we should avoid launching a subprocess at cudf import time.

bdice: Is there a way to use any other tool (cuda-python, rmm, ...) to know this information? I assume the goal is to avoid importing `numba.cuda` in this process...

brandon-b-miller: Yes, `ptxcompiler` uses a subprocess. This command is vendored directly from it [here](https://github.com/rapidsai/ptxcompiler/blob/main/ptxcompiler/patch.py#L136-L145). While doing all of our setup before importing `numba.cuda` is _one_ goal, another requirement is that the process for getting the versions doesn't `cuinit`, otherwise it can interfere with the way dask initializes its networking.

brandon-b-miller: As an aside it would be nice if we had some kind of small independent package we could list as a dependency that served as a source of truth for the driver and runtime versions, which we could also configure to avoid forking a subprocess, calling cuinit, etc. This would avoid a bunch of reinventing the wheel across cuda-python, numba, rmm, etc....

brandon-b-miller: I tried putting a 
```python
from cuda import cuda, cudart
cuda.cuDriverGetVersion()
cudart.cudaRuntimeGetVersion()
```
At the top of `cudf`'s `__init__.py` and ran `test_no_cuinit.py`, which passed. So I suppose it's a clean approach from the `cuInit` perspective. Happy to refactor as such here and see if things work.

python/cudf/cudf/utils/_numba.py
brandon-b-miller: These are the versions of NVVM used to produce the PTX files that we ship. At the point at which we introduced the first feature that relied on shipping a PTX file, we were building all of our packages with cuda 11.5. so it was assumed we wouldn't need ever need to look up a version earlier then that. Although I suppose one could imagine building from source on an earlier cuda version. 

I am adding CUDA versions back through 11.0 and 12.1 to be safe. 

We would want to update this every time we start building user facing packages with a new CUDA version that isn't in the table. I am happy to put that info somewhere, not sure where would be best though.

python/cudf/cudf/utils/_numba.py
jakirkham: Think this is referring to issue ( https://github.com/NVIDIA/cuda-python/issues/16 )

Namely the version cuda-python returns is hard-coded at build time as opposed to querying it at run time. So the version it returns may not reflect what the user has installed on their system (and will instead reflect how the binary was built). The issue explains why this is happening currently (though it still presents an issue with intended usage)

There's more details in the threads in PR ( https://github.com/rapidsai/rmm/pull/946 ) as well

python/cudf/cudf/utils/_numba.py
bdice: Exactly what I was looking for. Summarize that and put it in a comment. 😄

“CUDA 11.5 or higher is required to build cudf. This dictionary maps from supported versions of NVVM to the PTX version it produces.”

I would leave off CUDA versions prior to 11.5, because we cannot use them for other reasons (no int128 support, required for 128-bit decimal types).

brandon-b-miller:
I moved a little of the logic around here while vendoring the few extra pieces of ptxcompiler I think we need to keep things safe to `_ptxcompiler.py`. I think the way the logic reads is a lot simpler now and I'm hoping one more round of review should be enough to merge this.

python/cudf/cudf/utils/_numba.py
```diff
@@ -0,0 +1,168 @@
...
+
+    versions = safe_get_versions()
+    if versions != NO_DRIVER:
+        driver_version, runtime_version = versions
+        if driver_version >= (12, 0) and runtime_version > driver_version:
+            warnings.warn(
+                f"Using CUDA toolkit version {runtime_version} with CUDA "
+                f"driver version {driver_version} requires minor version "
+                "compatibility, which is not yet supported for CUDA "
+                "driver versions newer than 12.0. It is likely that many "
```
gmarkall, line 90: The current wording sounds like it refers to driver 12.1 onwards:

```suggestion
                "driver versions 12.0 and above. It is likely that many "
```

python/cudf/cudf/core/udf/utils.py
gmarkall: Is the comment here now outdated / vestigial?

brandon-b-miller: Yes. To summarize we moved towards warning on import of cuDF that it will likely be fragile rather than throwing when calling each and every API that could fail. I'll remove this comment.

python/cudf/cudf/core/udf/utils.py
brandon-b-miller: Resolving this due to removing this comment.

python/cudf/cudf/utils/_numba.py
brandon-b-miller: To close the loop on this I decided in https://github.com/rapidsai/cudf/pull/13337/commits/5cb0ce6d67b2f2c8813f3140cbbf3311ca4c7743 to just vendor the three functions to `_ptxcompiler.py`. 

To summarize the way it works now, the user either has `ptxcompiler` or doesn't. If they do, use it to get the versions (this launches a subprocess). If they don't, use the vendored functions to do so (this also launches the same subprocess using the same command). 

We are constrained to use a subprocess for three reasons:
1. We can not call `cuinit` on `cudf`'s import without interfering with dask. 
2. We can not directly use `numba.cuda` to obtain the versions because we need to finish configuring numba before `numba.cuda` is imported.
3. We can not utilize `cuda.cuda` due to constraint described above

Even _with_ the approach above we need to leave in place a mechanism to _disable_ the spawning of a subprocess in HPC environments where it is not safe to do so. In this case the user may provide an environment variable which disables the check and two more environment variables specifying the driver and runtime versions manually, and these are parsed and returned instead. These environment variables are the same ones, between `ptxcompiler` and our vendored version.

python/cudf/cudf/utils/_numba.py
```diff
@@ -0,0 +1,169 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+import glob
+import os
+import warnings
+
+from numba import config
+
+CC_60_PTX_FILE = os.path.dirname(__file__) + "/../core/udf/shim_60.ptx"
```
bdice, line 9: This needs to use `os.path.join`.

python/cudf/cudf/utils/_ptxcompiler.py
```diff
@@ -0,0 +1,105 @@
...
+
+import math
+import os
+import subprocess
+import sys
+import warnings
+
+NO_DRIVER = (math.inf, math.inf)
+
+CMD = """\
```
bdice, line 23: Can we name this a proper name? `NUMBA_CHECK_VERSION_CMD` or similar.

bdice (approved):
Looks fine to me. I discussed with @brandon-b-miller and I feel like I understand more of the constraints. I am not sure that we have the _best_ solution but I'm not sure how to improve it at this point.

I requested that we run a manual test, installing the CUDA 12 wheel from the CI artifacts on a system with driver 12.0 and a Docker image with runtime 12.1. This is supposed to raise a warning, I think, and I'd like to see that warning occur in manual testing (we don't have a CI configuration where this can be tested). @brandon-b-miller Please report back the result or let me know if you need help doing this.

gmarkall (approved):
This looks good. The vendoring of small parts of the ptxcompiler patch module look fine.

shwina (approved):

jakirkham:
Thanks all! 🙏

Really impressive work! 👏

PR #13338, opened 2023-05-11
Refactor `iloc.__getitem__` to handle more corner-cases

### LINKED ISSUE #13013: [BUG] iloc indexing with a Categorical not possible
**Describe the bug**

```python
import cudf
import pandas as pd

f = pd.Series([0, 1, 2])
index1 = pd.Categorical([1])
f.iloc[index1] # => Length 1 series with the second row
index2 = pd.Categorical([0, 2])
f.iloc[index2] # => Length 2 series with first and third rows

cf = cudf.from_pandas(f)
cf.iloc[index1] # => TypeError: int() argument must be a string, a bytes-like object or a real number, not 'Categorical'
cf.iloc[index2] # => NotImplementedError: Unknown indexer <class 'cudf.core.column.categorical.CategoricalColumn'>
```

The former error is because `_is_scalar_or_zero_d_array` treats categoricals of length one as scalars (prior to https://github.com/vyasr/cudf/commit/62511222d4dd72fa078c47b5c283c185111b79f9 _all_ categoricals were treated as scalars). Which, as observed above, is in conflict with pandas.

The latter is an error because casting into a an integer array doesn't work, and so the indexer ends up not knowing what to do.

**Expected behavior**

I'm not sure, the implicit casting to integers for iloc-based indexing of an object that might well _not_ contain integers feels like an accident of pandas history. The pandas documentation [says (of `iloc` indexing)](https://pandas.pydata.org/docs/dev/user_guide/indexing.html):

> The .iloc attribute is the primary access method. The following are valid inputs:
> [...]
> -    A list or array of integers [4, 3, 0].

I think a categorical is _not_ a list of integers, so this should be a `TypeError`, but it just happens to work because `np.asarray(whatever).astype(int)` works as long as `whatever` is coercible to int (this is _broad_ e.g. `f.iloc[[1.5]]` works!?).

@mroeschke do you happen to know if `iloc`-based indexing with categoricals is "definitely supported" or "accident of history"?

### LINKED ISSUE #13015: [BUG] loc and iloc indexing with boolean mask does not handle invalid length correctly
**Describe the bug**

When the indexing array passed to `iloc` (or `loc`, which I don't like because it's not label-based in this case) is boolean, the idea is that it should select those rows for which the mask is true [NA values are treated as false, which AFAICT is an unnecessary stipulation because there's no way to construct a pandas boolean array with NA values in it].

If the length of the mask array _does not match_ the length of the frame, we should obtain an `IndexError`. However, cudf produces an empty frame in this case.

```python
import pandas as pd
import cudf

s = pd.Series([1, 2, 3])
s.iloc[np.asarray([0, 1], dtype="bool")] # => IndexError
c = cudf.from_pandas(s)
c.iloc[np.asarray([0, 1], dtype="bool")] # => empty series
```

**Expected behavior**

Should probably match pandas.

### LINKED ISSUE #13267: [BUG] Series and DataFrame iloc indexing does not handle `Ellipsis`
**Describe the bug**

`Ellipsis` (or `...`) is used in multi-dimensional indexing to say "eat as many axes as needed as if you had `slice(None)` in that slot". Pandas does this, but cudf indexing does not (instead we get an error).

```python
import cudf
import pandas as pd
import numpy as np
cdf = cudf.from_pandas(df)

df.iloc[..., 0]
# 0    0
# 1    2
# Name: 0, dtype: int64

cdf.iloc[..., 0]
# => TypeError ellipsis is not iterable
```

**Expected behavior**

Should match pandas/numpy.

### LINKED ISSUE #13268: [BUG] Series and DataFrame loc indexing does not handle `Ellipsis`
**Describe the bug**

This is the same as #13267 really, but `loc`-based indexing goes down a different code path, so I will end up fixing it separately.

```python
import cudf
import pandas as pd
import numpy as np
cdf = cudf.from_pandas(df)

df.loc[..., 0]
# 0    0
# 1    2
# Name: 0, dtype: int64

cdf.loc[..., 0] # => TypeError
```

**Expected behavior**

Should match pandas.

wence-:
## Description

This refactors `iloc.__getitem__` to handle more corner cases
consistently. I don't dive through quite as many layers (there's
less similarity between `iloc` and `loc` indexing than the code
previously assumed). I think there will be opportunities for more
refactoring once I implement `__setitem__` but for now, let's see
if this looks ok.

- Closes #13015
- Closes #13013
- Closes #13267
- Closes #13268

## Checklist

I thought I would get a bit further, but multiindex-based indexing is horrible (so `.loc` will have to wait a bit longer).


wence-:
Probably punt this to 23.08

wence-:
In the interests of making reviewing easier, I'm splitting the indexing rework into more pieces. This one just adds tests for most of the things we know don't work correctly at present.

python/cudf/cudf/tests/test_indexing.py
```diff
@@ -1741,3 +1742,319 @@ def test_boolean_mask_columns_iloc_series():
...
+        return request.param
+
+    @pytest.fixture
+    def index(self, order, dtype):
+        if dtype == "string":
+            index = ["a", "h", "f", "z"]
+        elif dtype == "int":
+            index = [-1, 10, 7, 14]
+        elif dtype == "float":
+            index = [-1.5, 7.10, 2.4, 11.2]
```
bdice, line 2003: It seems like these lists should just _be_ the parameters, rather than having a parameter `dtype`. It doesn't seem that `dtype` is used otherwise.

python/cudf/cudf/tests/test_indexing.py
```diff
@@ -1741,3 +1742,319 @@ def test_boolean_mask_columns_iloc_series():
...
+    df = pd.DataFrame(value, index=timestamps, columns=["value"])
+    cdf = cudf.from_pandas(df)
+    if index_type == "single":
+        index = pd.Timestamp(
+            datetime.strptime("2021-03-12 03:00", "%Y-%m-%d %H:%M")
+        )
+    elif index_type == "slice":
+        index = slice(start, end, None)
+    else:
+        raise ValueError("Invalid index type")
```
vyasr, line 1764: Probably overkill for a test function where you control all the inputs.

python/cudf/cudf/tests/test_indexing.py
```diff
@@ -1741,3 +1742,319 @@ def test_boolean_mask_columns_iloc_series():
 
     with pytest.raises(NotImplementedError):
         cdf.iloc[:, mask]
+
+
+@pytest.mark.parametrize("index_type", ["single", "slice"])
+def test_loc_timestamp_8585(index_type):
```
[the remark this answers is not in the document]
vyasr: Oh I see the normal index case works, it's the MultiIndex case in the next test that's a problem.

python/cudf/cudf/tests/test_indexing.py
```diff
@@ -1741,3 +1742,319 @@ def test_boolean_mask_columns_iloc_series():
...
+def test_iloc_13269():
+    df = pd.DataFrame(np.arange(4).reshape(2, 2))
+    cdf = cudf.from_pandas(df)
+
+    expect = df.loc[:, [0, 1, 0]]
+    actual = cdf.loc[:, [0, 1, 0]]
+    assert_eq(expect, actual)
+
+
+def test_loc_13270():
```
vyasr, line 1855: For the cases that are fixed, could we use descriptive names rather than issue numbers? I didn't even immediately realize that's what the numbers were.

vyasr (approved):
Couple of small requests, but nothing blocking. Thank you for diligently constructing all the necessary test cases here, it's super valuable.

python/cudf/cudf/tests/test_indexing.py
wence-: I've updated the names (with trailing `issue_XXX`) and added a link as a comment to the relevant issue for future spelunking.

python/cudf/cudf/tests/test_indexing.py
wence-: It was belt-and-braces since with the `pytest.param` stuff the parameter is half a screen away and a bit inconspicuous. Also, I like total functions :)

python/cudf/cudf/tests/test_indexing.py
```diff
@@ -1741,3 +1742,319 @@ def test_boolean_mask_columns_iloc_series():
...
+        index = pd.date_range(start, end, periods=2)
+    else:
+        raise ValueError("Invalid index type")
+    expect = df.loc[index]
+    actual = cdf.loc[index]
+    assert_eq(expect, actual)
+
+
+@pytest.mark.xfail(reason="https://github.com/rapidsai/cudf/issues/8693")
+def test_loc_8693():
```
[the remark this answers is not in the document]
wence-: Renamed and added links to the issues in comments.

python/cudf/cudf/tests/test_indexing.py
wence-: I use the dtype to conditionally apply a marker in the actual test:

```
        if not (order == "increasing" and dtype in {"int", "float"}):
            request.applymarker(
                pytest.mark.xfail(
                    reason="https://github.com/rapidsai/cudf/issues/12833"
                )
            )
```

So it's somewhat easier to just know that rather than pulling the dtype out of the index on the dataframe.

PR #13340, opened 2023-05-11
Update mypy to 1.3

### LINKED ISSUE #13336: [FEA] Update to mypy >= v1
Mypy v1 is out, which (amongst other things) introduces support for [PEP 673](https://peps.python.org/pep-0673/) "The Self type". Being able to use this will make typing a lot of our factory class methods simpler.

wence-:
## Description
Bring us up to date with the current mypy release. Big diff, mostly due to the no-implicit-optional rule that mypy now enforces (see commit message of 040f5773c37d89cdbc0baf57fa70a4b3dcab4fc9 for details). That commit is automated so probably best to review commit-by-commit for the substantive changes.

- Closes #13336 

## Checklist

### DIFF
--- a/python/cudf/cudf/core/column/lists.py
+++ b/python/cudf/cudf/core/column/lists.py
@@ -5,6 +5,7 @@ from typing import List, Optional, Sequence, Tuple, Union
 
 import numpy as np
 import pyarrow as pa
+from typing_extensions import Self
 
 import cudf
 from cudf._lib.copying import segmented_gather
@@ -261,7 +262,7 @@ class ListColumn(ColumnBase):
         # Call libcudf to format the list column
         return format_list_column(lc, separators)
 
-    def _transform_leaves(self, func, *args, **kwargs):
+    def _transform_leaves(self, func, *args, **kwargs) -> Self:
         # return a new list column with the same nested structure
         # as ``self``, but with the leaf column transformed
         # by applying ``func`` to it
--- a/python/cudf/cudf/core/column/struct.py
+++ b/python/cudf/cudf/core/column/struct.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 from functools import cached_property
+from typing import Optional
 
 import pandas as pd
 import pyarrow as pa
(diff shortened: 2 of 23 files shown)


vyasr:
This closes issue #13336

wence-:
> This closes issue #13336, right?

Yes, which I had somehow forgotten I had written as I was making this branch

python/cudf/cudf/core/column/lists.py
```diff
@@ -261,7 +262,7 @@ def as_string_column(
         # Call libcudf to format the list column
         return format_list_column(lc, separators)
 
-    def _transform_leaves(self, func, *args, **kwargs):
+    def _transform_leaves(self, func, *args, **kwargs) -> Self:
```
vyasr, line 265: IIUC this change was what motivated this update. What is the difference between using this and a bound generic like `T = TypeVar("T", bound="DataFrame")` as we've usually done?

python/cudf/cudf/core/column/struct.py
```diff
@@ -57,7 +58,9 @@ def to_arrow(self):
             pa_type, len(self), buffers, children=children
         )
 
-    def to_pandas(self, index: pd.Index = None, **kwargs) -> "pd.Series":
+    def to_pandas(
+        self, index: Optional[pd.Index] = None, **kwargs
+    ) -> "pd.Series":
```
vyasr, line 63: I think we can remove the quotes since we're importing pandas unconditionally here anyway.
```suggestion
    ) -> pd.Series:
```

vyasr, line 63: I think we can remove the quotes since we're importing pandas unconditionally here anyway.
```suggestion
    ) -> pd.Series:
```

vyasr (approved):
One question and one suggestion, but generally LGTM. All the Optional changes seem fine.

python/cudf/cudf/core/column/lists.py
wence-: [PEP 673](https://peps.python.org/pep-0673/) has all the gory details. tl;dr this is new syntax for the old `T = TypeVar("T", bound="Foo")` setup, but more natural (one doesn't also have to annotate `self : T`, and it also works with classmethods where previously you would have written `def(cls: type[T], ...) -> T`.

python/cudf/cudf/core/column/struct.py
wence-: I guess I can, but maybe in a separate changeset?

bdice:
@wence- Huge thanks for this. This update will make it easier to keep other pre-commit hooks updated on a semi-regular basis with `pre-commit autoupdate`. I have wanted to explore doing that update once every release or two but mypy was by far the hardest update to satisfy.

python/cudf/cudf/core/column/lists.py
vyasr: Agreed, no need to do it all in this PR.

PR #13341, opened 2023-05-11
Remove default UNKNOWN_NULL_COUNT from cudf::column member functions

davidwendt:
## Description
Remove the default parameters for null-mask and null-count from `cudf::column` constructors and `set_null_mask` member functions.

Reference #13311 

## Checklist


cpp/src/interop/from_arrow.cu
```diff
@@ -169,7 +169,7 @@ struct dispatch_to_cudf_column {
 
 std::unique_ptr<column> get_empty_type_column(size_type size)
 {
-  return std::make_unique<column>(data_type(type_id::EMPTY), size, rmm::device_buffer{});
+  return make_empty_column(type_id::EMPTY);
```
vyasr, line 172: This change is the reason that the tests are currently failing. The Python layer relies on being able to construct an empty column with a finite size. You can see the difference with this minimal example.
Before the change:
```
(rapids) rapids@compose:~/cudf$ python -c "import cudf; print(cudf.core.column.as_column([cudf.NA]))"
<cudf.core.column.string.StringColumn object at 0x7f5da8778240>
1 nulls
dtype: object
```
And after:
```
(rapids) rapids@compose:~/cudf$ python -c "import cudf; print(cudf.core.column.as_column([cudf.NA]))"
<cudf.core.column.string.StringColumn object at 0x7f64504cc040>
0 nulls
dtype: object
```

The C++ layer has [a check preventing empty columns from having a null mask](https://github.com/rapidsai/cudf/blob/branch-22.06/cpp/src/column/column_view.cpp#L51), but it doesn't check on the null _count_. Up until now, the Python code has been relying on operating with a somewhat broken intermediate state: an empty column with nonzero size, no null mask, and an unknown null count, being constructed in a case when in fact the data is all invalid. However, you can also see that the null count is being computed correctly in the above example, so I'm not quite sure where the EMPTY column is getting converted into a reasonable string column. 

I need to do a bit more digging to figure out exactly what the best solution is here.

galipremsagar (approved):

cpp/src/interop/from_arrow.cu
```diff
@@ -169,7 +170,9 @@ struct dispatch_to_cudf_column {
 
 std::unique_ptr<column> get_empty_type_column(size_type size)
 {
-  return std::make_unique<column>(data_type(type_id::EMPTY), size, rmm::device_buffer{});
+  // this abomination is required by cuDF
```
vyasr, line 173: Might as well explain the reason for the abomination :joy: 
```suggestion
  // this abomination is required by cuDF Python, which needs to handle
  // [PyArrow null arrays](https://arrow.apache.org/docs/python/generated/pyarrow.NullArray.html)
  // of finite length
```

cpp/tests/copying/concatenate_tests.cpp
```diff
@@ -194,8 +188,14 @@ TEST_F(StringColumnTest, ConcatenateTooManyColumns)
 
 struct TableTest : public cudf::test::BaseFixture {};
 
+template <typename T>
+using column_wrapper = cudf::test::fixed_width_column_wrapper<T>;
```
vyasr, line 192: Did this just move from above? Also, couldn't it be used in some of the lines above (e.g. 78-80)? I'm not suggesting expanding the diff of this PR for unrelated changes, more not sure what the purpose of this change was here. If it's intended to improve the scoping of the alias then perhaps we should save that for a future PR where it's used more consistently throughout the file?

cpp/tests/copying/concatenate_tests.cpp
```diff
@@ -404,7 +404,7 @@ TEST_F(OverflowTest, OverflowTest)
       static_cast<cudf::size_type>(static_cast<uint32_t>(512) * 1024 * 1024);
 
     // struct
-    std::vector<std::unique_ptr<column>> children;
+    std::vector<std::unique_ptr<cudf::column>> children;
```
vyasr, line 407: Did we remove a `using namespace cudf` or `namespace cudf {` somewhere that I'm missing in this file? There seem to be a large number of changes unrelated to the column API change, and I don't see why they would even be necessary based on the focus of the PR. Perhaps those changes should be split into a separate PR?

vyasr (changes requested):
Mostly looks great to me! However IIUC about 50% of the changes are in the concatenate_tests.cpp file, and they don't seem relevant. That ratio is too high for me, so unless I'm misunderstanding why that's here I would like those changes split into a separate PR.

cpp/tests/copying/concatenate_tests.cpp
davidwendt: The offending lines removed are 46-47 above.
```
using column = cudf::column;
using column_view = cudf::column_view;
```
I could not just let that be.

davidwendt: I did not realize how big a change that turned out to be. I can certainly do that change in a separate PR.

vyasr (approved):
LGTM, thanks!

mythrocks (approved):
LGTM!

ChuckHastings:
Shouldn't this have been marked BREAKING?  It broke cugraph.

The constructor for `cudf::column` now requires all 4 parameters to be passed, so any code that relied on the default for the fourth parameter is now broken.

vyasr:
Yes, this should have been marked as breaking. Apologies for the oversight. @raydouglass will updating the labels now have any effect, or will we need to modify the changelog etc manually to account for this now?

raydouglass:
> will updating the labels now have any effect

Yes, the changelog is generated from scratch each time, so if the labels are updated the next PR that merges will update the entire changelog.

ChuckHastings:
I'm not pressed about retroactively worrying about this, personally.  The breaking label helps me by alerting me in the slack channel that changes might come up that might affect my development.  In this case it was pretty easy to find what the culprit was (the error message clearly indicated a problem in this constructor... not too many PRs merged recently caused a change in the constructor.

I don't know how the labels propagate into release documentation.  If the labels have an effect on that, I suppose it would be helpful to get that tagged correctly so that it's represented in the release documentation.  If that's all still manual then it's probably less important.

vyasr:
Sounds good, updated the labels.

Issue #6004, opened 2020-08-17
[FEA] Profiling duplicate reading of metadata


calebwin:
The row-group-level filtered reading for Parquet that is introduced by #5843 creates an issue of duplicate metadata (metadata is stored in the footers of Parquet files) reading in the case when filters are specified. Arrow is used to read metadata and select a subset of data to read given user-provided filters [4] . Information about this subset is then passed to libcudf which reads in the subset [5]. The issue is that metadata gets read twice - first when Arrow reads metadata to do filtering and second when libcudf reads data.

This issue was initially raised here [1].

# What to profile

- [ ] Perf penalty of reading metadata using Arrow for filtering in the same vein as [2] but with datasets of varying # of files
- [ ] Perf penalty of parsing metadata buffer [3] as fraction of total time Arrow spends reading metadata

# What to determine

- [ ] Determine whether or not perf penalty of the additional reading of metadata using Arrow is significant
- [ ] Determine whether the duplicate reading should be resolved by passing metadata struct (steps to implement [6]) or metadata buffer (which is then parsed into metadata struct in libcudf) (steps to implement [7]) from Arrow `Dataset` to libcudf reader functions

# Relevant discussion/code

[1] https://github.com/rapidsai/cudf/pull/5843#discussion_r467191621
[2] https://github.com/rapidsai/cudf/pull/5843#issuecomment-673566456
[3] https://github.com/apache/arrow/blob/2e6009621011d7df43882aa883905b84d1647018/cpp/src/parquet/file_reader.cc#L532
[4] https://github.com/rapidsai/cudf/pull/5843/files#diff-deac873508aaa12ca2e7c0a2c9035230R316
[5] https://github.com/rapidsai/cudf/pull/5843/files#diff-deac873508aaa12ca2e7c0a2c9035230R359-R375
[6] https://github.com/rapidsai/cudf/pull/5843#issuecomment-674437264
[7] https://github.com/rapidsai/cudf/pull/5843#issuecomment-674437709

GregoryKimball:
Hello @wence- , now that #15028 is merged, would you please let me know if cuDF-python is still reading parquet row group metadata using pyarrow? Or is that step completely removed?

wence-:
> Hello @wence- , now that #15028 is merged, would you please let me know if cuDF-python is still reading parquet row group metadata using pyarrow? Or is that step completely removed?

[...]

PR #13348, opened 2023-05-13
[FEA] AST filtering in parquet reader

karthikeyann:
## Description

The plan to support AST based filter predicate pushdown in parquet. This PR adds predicate pushdown on row group filtering. 

The statistics of columns of each row group are loaded to a device column, and AST filter is applied on min, max of each column to select the row groups to read. The user given AST needs to be converted to another AST to be applied on min, max values of each column ('Statistics AST'). After the row groups are parsed, the user given AST is applied on the output columns to filter any remaining rows in the row groups.
New `column_name_reference` is introduced to help the users create AST's that reference columns by name, as the user may or may not have the column indices information before reading. Since AST engine takes only column index reference, a transformation is applied to the user given AST. So, 2 new AST transformation classes are introduced: 
1. `named_to_reference_converter` - Converts column name references to column index references
2. `stats_expression_converter` - Converts the above output table filtering AST to 'Statistics AST'.

Note: This column_name_reference only supported for predicate pushdown filtering, but not supported for other AST operations such as transform, joins etc.

- [x] #13472 
- [x] Convert column chunk min, max to cudf type column.
- [x] Add AST filter interface to parquet reader options
- [x] Convert AST to Statistics AST
- [x] Apply statistics AST on Stats values to get row_groups
- [x] Apply AST as filter on output columns.

Depends on #13472

## Checklist


cpp/src/io/parquet/compact_protocol_writer.cpp
```diff
@@ -201,7 +201,20 @@ size_t CompactProtocolWriter::write(const ColumnChunkMetaData& s)
...
+
+size_t CompactProtocolWriter::write(const Statistics& s)
+{
+  CompactProtocolFieldWriter c(*this);
+  if (s.max.size() != 0) { c.field_binary(1, s.max); }
+  if (s.min.size() != 0) { c.field_binary(2, s.min); }
+  if (s.null_count != -1) { c.field_int(3, s.null_count); }
+  if (s.distinct_count != -1) { c.field_int(4, s.distinct_count); }
+  if (s.max_value.size() != 0) { c.field_binary(5, s.max_value); }
+  if (s.min_value.size() != 0) { c.field_binary(6, s.min_value); }
```
GregoryKimball, line 217: I'm surprised to see a writer change here - was there something incorrect about the way we've been writing row group statistics?

karthikeyann: No. The statistics are pre-encoded during page encoding itself and stored as binary blob. I changed the datatype from binary blob to `Statistics` struct. So, this change is encoding that struct again.

cpp/src/io/parquet/parquet.hpp
```diff
@@ -227,7 +227,7 @@ struct ColumnChunkMetaData {
   int64_t index_page_offset = 0;  // Byte offset from beginning of file to root index page
   int64_t dictionary_page_offset =
     0;  // Byte offset from the beginning of file to first (only) dictionary page
```
karthikeyann, line 229:
```suggestion
    0;                    // Byte offset from the beginning of file to first (only) dictionary page
```

GregoryKimball:
I expect this will close #6004

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+      CUDF_EXPECTS(cudf::ast::detail::ast_operator_arity(op) == 2,
+                   "Only binary operations are supported on column reference");
+      CUDF_EXPECTS(dynamic_cast<ast::literal const*>(&operands[1].get()) != nullptr,
+                   "Second operand of binary operation with column reference must be a literal");
+      v->accept(*this);
+      auto const col_index = v->get_column_index();
+      switch (op) {
+        /* transform to stats conditions. op(col, literal)
+        col1 == val --> vmin <= val && vmax >= val
+        col1 != val --> !(vmin == val && vmax == val)
```
bdice, line 271: I think this should be range-based? It should be the opposite of `==`.
```suggestion
        col1 != val --> vmin > val || vmax < val
```

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+        }
+        case ast_operator::NOT_EQUAL: {
+          auto const& vmin = _col_ref.emplace_back(col_index * 2);
+          auto const& vmax = _col_ref.emplace_back(col_index * 2 + 1);
+          auto const& op1  = _operators.emplace_back(ast_operator::NOT_EQUAL, vmin, vmax);
+          auto const& op2 =
+            _operators.emplace_back(ast_operator::NOT_EQUAL, vmax, operands[1].get());
+          _operators.emplace_back(ast_operator::LOGICAL_OR, op1, op2);
+          break;
+        }
```
bdice, line 295: See above.
```suggestion
        case ast_operator::NOT_EQUAL: {
          auto const& vmin = _col_ref.emplace_back(col_index * 2);
          auto const& vmax = _col_ref.emplace_back(col_index * 2 + 1);
          auto const& op1  =
            _operators.emplace_back(ast_operator::GREATER, vmin, operands[1].get());
          auto const& op2 =
            _operators.emplace_back(ast_operator::LESS, vmax, operands[1].get());
          _operators.emplace_back(ast_operator::LOGICAL_OR, op1, op2);
          break;
        }
```

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+        // can these be combined in any way?
```
bdice, line 296: I'm happy with this implementation. I wouldn't change it.
```suggestion
```

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+  }
+
+  /**
+   * @brief Returns the AST to apply on Column chunk statistics.
+   *
+   * @return AST operation expression
+   */
+  [[nodiscard]] std::reference_wrapper<ast::expression const> get_stats_expr() const
+  {
+    // if(!_stats_expr.has_value()) _stats_expr = ast::operation(ast_operator::IDENTITY, expr)
```
bdice, line 329: Unused?
```suggestion
```

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+      predicate.null_count() == predicate.size()) {
+    return std::nullopt;
+  }
+  size_type is_required_idx = 0;
+  for (size_t src_idx = 0; src_idx < input_row_group_indices.size(); ++src_idx) {
+    std::vector<size_type> filtered_row_groups;
+    for (auto const rg_idx : input_row_group_indices[src_idx]) {
+      if ((!validity_it[is_required_idx]) || is_row_group_required[is_required_idx]) {
+        filtered_row_groups.push_back(rg_idx);
+        // std::cout << "Read [src_idx, rg_idx]=[" << src_idx << "," << rg_idx << "]\n";
```
bdice, line 441:
```suggestion
```

bdice (changes requested):
First pass, partial review. I mostly looked at `predicate_pushdown.cpp`.

cpp/include/cudf/ast/expressions.hpp
```diff
@@ -474,6 +502,48 @@ class operation : public expression {
   std::vector<std::reference_wrapper<expression const>> const operands;
 };
 
+/**
+ * @brief A expression referring to data from a column in a table.
+ */
+class column_name_reference : public expression {
```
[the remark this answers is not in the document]
karthikeyann: column names are not available before reading the parquet file (metadata) before calling the reader. That's why this is added.

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+        typename ToType::duration{static_cast<typename ToType::rep>(value)});
+    } else if constexpr (std::is_same_v<ToType, string_view>) {
+      return ToType{nullptr, 0};
+    } else {
+      return static_cast<ToType>(value);
+    }
+  }
+  // uses storage type as T
+  template <typename T>
+  static T convert(uint8_t const* stats_val, size_t stats_size, Type const type)
```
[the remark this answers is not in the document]
karthikeyann: Type is `cudf::io::parquet::Type`, will update to full namespace path.

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+  std::pair<std::unique_ptr<column>, std::unique_ptr<column>> operator()(
+    size_t col_idx, cudf::data_type dtype) const
+  {
+    // List, Struct, Dictionary types are not supported
+    if constexpr (is_compound<T>() && !std::is_same_v<T, string_view>)
+      CUDF_FAIL("Compound types does not have statistics");
+    else {
+      // Local struct to hold host columns
+      struct host_column {
+        thrust::host_vector<T> val;
```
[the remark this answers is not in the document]
karthikeyann: because `std::vector<bool>` uses bits to store each bool instead of a byte. 🙁

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+  template <typename T>
+  std::pair<std::unique_ptr<column>, std::unique_ptr<column>> operator()(
+    size_t col_idx, cudf::data_type dtype) const
+  {
+    // List, Struct, Dictionary types are not supported
+    if constexpr (is_compound<T>() && !std::is_same_v<T, string_view>)
+      CUDF_FAIL("Compound types does not have statistics");
+    else {
+      // Local struct to hold host columns
+      struct host_column {
```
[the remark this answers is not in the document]
karthikeyann: Performance is not the reason. The number of values to decode is less. The metadata is in host, and the decoding of binary blob to respective type also happens in host. That's the reason for `host_column` creation.
Alternative design might be to copy the binary blobs to GPU and decode in device. Considering less number of values to decode, I don't see much value in parallelising the decoding part. besides, null decoding may be little tricky which will incur more host code & device code. So, using GPU-resident column classes will still incur significant glue-code to make it work.

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+      return {min.to_device(dtype, stream), max.to_device(dtype, stream)};
+    }
+  }
+};
+
+/**
+ * @brief Converts AST expression to StatsAST for comparing with column statistics
+ * This is used in row group filtering based on predicate.
+ * statistics min value of a column is referenced by column_index*2
+ * statistics min value of a column is referenced by column_index*2+1
```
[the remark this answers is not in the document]
karthikeyann: good catch.

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+    using cudf::ast::ast_operator;
+    auto const operands = expr.get_operands();
+    auto op             = expr.get_operator();
+
+    if (auto* v = dynamic_cast<ast::column_reference const*>(&operands[0].get())) {
+      // First operand should be column reference, second should be literal.
+      CUDF_EXPECTS(cudf::ast::detail::ast_operator_arity(op) == 2,
+                   "Only binary operations are supported on column reference");
+      CUDF_EXPECTS(dynamic_cast<ast::literal const*>(&operands[1].get()) != nullptr,
+                   "Second operand of binary operation with column reference must be a literal");
```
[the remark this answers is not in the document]
karthikeyann: We can't compare 2 columns. columns can be compared against literal, and the their results can be combined using other operators.

cpp/src/io/parquet/predicate_pushdown.cpp
karthikeyann: I tried the same - `vmin > val or vmax < val`. It didn't work.

cpp/src/io/parquet/predicate_pushdown.cpp
karthikeyann: https://github.com/karthikeyann/cudf/commit/2fa536328f22de5b16e22656a2afd8edb1dfdcee
Didn't work.

cpp/src/io/parquet/predicate_pushdown.cpp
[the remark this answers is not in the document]
karthikeyann: It's an enum.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -359,58 +386,65 @@ table_with_metadata reader::impl::read_chunk_internal(bool uses_custom_row_bound
...
       out_columns.emplace_back(make_column(_output_buffers[i], &col_name, metadata, _stream));
     } else {
       out_columns.emplace_back(make_column(_output_buffers[i], nullptr, metadata, _stream));
     }
   }
 
-  // Add empty columns if needed.
-  return finalize_output(out_metadata, out_columns);
+  // Add empty columns if needed. Filter output columns based on filter.
+  return finalize_output(out_metadata, out_columns, filter);
```
hyperbolic2346, line 397: So this reads every column and culls at the end?

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,501 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "reader_impl_helpers.hpp"
+
+#include <cudf/ast/detail/expression_parser.hpp>  // possibly_null_value defined here, but used in operators.hpp
```
hyperbolic2346, line 18: Why isn't this in operators.hpp then?

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,501 @@
...
+    switch (type) {
+      case BOOLEAN: return targetType<T>(*reinterpret_cast<bool const*>(stats_val));
+      case INT32: return targetType<T>(*reinterpret_cast<int32_t const*>(stats_val));
+      case INT64: return targetType<T>(*reinterpret_cast<int64_t const*>(stats_val));
+      case INT96:  // Deprecated
+        return targetType<T>(static_cast<__int128_t>(reinterpret_cast<int64_t const*>(stats_val)[0])
+                               << 32 |
+                             reinterpret_cast<int32_t const*>(stats_val)[2]);
+      case FLOAT:
+        if constexpr (std::is_floating_point_v<T>)
```
hyperbolic2346, line 75: This check isn't included on the integer or boolean values. Why do we need to check it here? If the caller asks for a float out of a bool pointer it would be undefined, right? If it is an alignment issue, why not check the int64?

cpp/src/io/parquet/predicate_pushdown.cpp
hyperbolic2346: Did it fail on certain types like floats or timestamps? I had the same though as Bradley here. Why won't this comparison work?

etseidl: It's because the point of the predicate is to filter out row groups (or ultimately pages) that cannot satisfy the predicate.  In the case of `col1 != val`, the only time you could eliminate the row group entirely would be if the row group is a single value equal to `val`, in which case `min` and `max` will both be equal to `val`.  In any other case you'd have to test each row individually for the inequality.

cpp/src/io/parquet/reader_impl.cpp
karthikeyann: yes.

cpp/src/io/parquet/predicate_pushdown.cpp
karthikeyann: updated.

cpp/src/io/parquet/predicate_pushdown.cpp
karthikeyann: moved to type specializations for different types.

cpp/include/cudf/ast/expressions.hpp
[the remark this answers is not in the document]
bdice: Can we fetch metadata from the file first, before reading the rest of the file? I am very hesitant to introduce this logic into the AST feature. It is not well suited to the rest of libcudf which is generally unaware of column names.

cpp/src/io/parquet/predicate_pushdown.cpp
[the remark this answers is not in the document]
bdice: Can we change that code to use an `enum class` in a separate PR?

cpp/src/io/parquet/predicate_pushdown.cpp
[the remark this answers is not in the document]
bdice: Ouch. Let's document that decision here in a comment.

cpp/src/io/parquet/predicate_pushdown.cpp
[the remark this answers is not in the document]
bdice: Is that a limitation of our implementation, or is that a limitation of the Parquet predicate pushdown specification? (Is there such a specification?)

cpp/src/io/parquet/predicate_pushdown.cpp
bdice: Assuming that `vmin <= vmax`, I think that my suggestion is still valid -- and it's more obvious that this is the opposite of the equality case.

Reasoning: `vmin > val` implies `vmax > val`, so we are checking `vmax > val && vmax < val`. That's probably equivalent to `vmax != val`. Similarly for `vmax` --> `vmin > val && vmin < val`, then apply DeMorgan's law. Do you agree?

The case I'm worried about is NaN, because `x > NaN && x < NaN` isn't equivalent to `x != NaN`. I'm not sure which one is the right choice here from the quick glance and would need to think more about it. We should carefully assess and test the behaviors of NaN values (an entire row group of NaN, a row group containing NaNs, target value set to NaN, target value set to a non-NaN, etc.). I need to look back over the tests to see if these cases are covered.

cpp/include/cudf/ast/expressions.hpp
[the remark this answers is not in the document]
bdice: Another alternative would be to create an arbitrary mapping of input column names to integers (assign the first name `0`, second name `1`, and increment from there). Then we could reindex the column references at a later time once the true indices for those column names are known.

cpp/src/io/parquet/reader_impl_helpers.cpp
```diff
@@ -342,8 +342,18 @@ std::tuple<int64_t, size_type, std::vector<row_group_info>>
...
+  host_span<data_type const> output_dtypes,
+  std::optional<std::reference_wrapper<ast::expression const>> filter) const
 {
+  std::optional<std::vector<std::vector<size_type>>> filtered_row_group_indices;
+  if (filter.has_value()) {
+    filtered_row_group_indices =
+      filter_row_groups(row_group_indices, output_dtypes, filter.value());
+    if (filtered_row_group_indices.has_value())
+      row_group_indices =
+        host_span<std::vector<size_type> const>(filtered_row_group_indices.value());
```
bdice, line 355: Always wrap the body of a conditional or loop in braces.
```suggestion
    if (filtered_row_group_indices.has_value()) {
      row_group_indices =
        host_span<std::vector<size_type> const>(filtered_row_group_indices.value());
    }
```

cpp/src/io/parquet/reader_impl_helpers.hpp
```diff
@@ -195,4 +215,63 @@ class aggregate_reader_metadata {
...
+
+ private:
+  std::vector<std::reference_wrapper<ast::expression const>> visit_operands(
+    std::vector<std::reference_wrapper<ast::expression const>> operands);
+
+  table_metadata const& metadata;
+  std::unordered_map<std::string, size_type> column_name_to_index;
+  std::optional<std::reference_wrapper<ast::expression const>> _stats_expr;
+  std::list<ast::column_reference> _col_ref;
+  std::list<ast::operation> _operators;
```
bdice, line 274: Maybe add a comment about why we choose `std::list` here. I can't remember if I had a comment like this elsewhere in the AST code. In my understanding it's to prevent invalidation of the reference wrappers that would occur if a `std::vector` is resized?

cpp/src/io/parquet/reader_impl_helpers.hpp
```diff
@@ -195,4 +215,63 @@ class aggregate_reader_metadata {
...
+  named_to_reference_converter(std::optional<std::reference_wrapper<ast::expression const>> expr,
+                               table_metadata const& metadata)
+    : metadata(metadata)
+  {
+    if (!expr.has_value()) return;
+    // create map for column name.
+    std::transform(metadata.schema_info.begin(),
+                   metadata.schema_info.end(),
+                   std::inserter(column_name_to_index, column_name_to_index.end()),
+                   [i = 0](const auto& data) mutable { return make_pair(data.name, i++); });
```
bdice, line 233: East const?
```suggestion
                   [i = 0](auto const& data) mutable { return make_pair(data.name, i++); });
```

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5602,584 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+  auto literal           = cudf::ast::literal(literal_value);
+  auto col_ref_0         = cudf::ast::column_reference(0);
+  auto filter_expression = cudf::ast::operation(cudf::ast::ast_operator::LESS, col_ref_0, literal);
+
+  auto predicate = cudf::compute_column(*written_table, filter_expression);
+  EXPECT_EQ(predicate->view().type().id(), cudf::type_id::BOOL8)
+    << "Predicate filter should return a boolean";
+  auto expected = cudf::apply_boolean_mask(*written_table, *predicate);
+  // To make sure AST filters out some elements
+  EXPECT_NE(written_table->num_rows(), expected->num_rows());
```
bdice, line 5626: We should enforce that the expected number of rows is _less_ than the number of rows in the written table, not just "not equal."

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5602,584 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+  columns.push_back(col2.release());
+
+  return std::pair{cudf::table{std::move(columns)}, filepath};
+}
+
+TEST_F(ParquetReaderTest, FilterIdentity)
+{
+  auto [src, filepath] = create_parquet_with_stats("FilterIdentity.parquet");
+
+  // Filtering AST - table[0] < RAND_MAX/2
```
bdice, line 5669: This comment seems wrong.
```suggestion
  // Filtering AST - identity function, always true.
```

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5602,584 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+  CUDF_TEST_EXPECT_TABLES_EQUAL(*result.tbl, *expected);
+}
+
+// Test for Types - numeric, chrono, string.
+template <typename T>
+struct ParquetReaderPredicatePushdownTest : public ParquetReaderTest {};
+
+// These chrono types are not supported because parquet writer does not have a type to represent
+// them.
+using UnSupportedChronoTypes =
```
bdice, line 5733:
```suggestion
using UnsupportedChronoTypes =
```

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5602,584 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+    cudf::io::table_input_metadata expected_metadata(written_table);
+    expected_metadata.column_metadata[0].set_name("col0");
+    expected_metadata.column_metadata[1].set_name("col1");
+    expected_metadata.column_metadata[2].set_name("col2");
+
+    const cudf::io::parquet_writer_options out_opts =
+      cudf::io::parquet_writer_options::builder(cudf::io::sink_info{filepath}, written_table)
+        .metadata(&expected_metadata)
+        .row_group_size_rows(8000);
+    // .stats_level(cudf::io::statistics_freq::STATISTICS_ROWGROUP);
```
bdice, line 5760: Is this supposed to be commented?

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5602,584 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+}
+
+TYPED_TEST(ParquetReaderPredicatePushdownTest, FilterTyped)
+{
+  using T = TypeParam;
+
+  auto const [src, filepath] = create_parquet_typed_with_stats<T>("FilterTyped.parquet");
+  auto const written_table   = src.view();
+
+  // Filtering AST - table[0] < 100
```
bdice, line 5779: Is this comment correct?

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5602,584 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+  auto literal_value = []() {
+    if constexpr (cudf::is_timestamp<T>())
+      return cudf::timestamp_scalar<T>(T(typename T::duration(10000)));   // i (0-20,000)
+    else if constexpr (cudf::is_duration<T>())
+      return cudf::duration_scalar<T>(T(10000));                          // i (0-20,000)
+    else if constexpr (std::is_same_v<T, cudf::string_view>)
+      return cudf::string_scalar("000010000");                            // i (0-20,000)
+    else
+      return cudf::numeric_scalar<T>((100 - 100 * std::is_signed_v<T>));  // i/100 (-100-100/ 0-200)
```
bdice, line 5788: Always use braces.
```suggestion
    if constexpr (cudf::is_timestamp<T>()) {
      return cudf::timestamp_scalar<T>(T(typename T::duration(10000)));   // i (0-20,000)
    } else if constexpr (cudf::is_duration<T>()) {
      return cudf::duration_scalar<T>(T(10000));                          // i (0-20,000)
    } else if constexpr (std::is_same_v<T, cudf::string_view>) {
      return cudf::string_scalar("000010000");                            // i (0-20,000)
    } else {
      return cudf::numeric_scalar<T>((100 - 100 * std::is_signed_v<T>));  // i/100 (-100-100/ 0-200)
    }
```

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5602,584 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+  auto ref_filter        = cudf::ast::operation(cudf::ast::ast_operator::LESS, col_ref_0, literal);
+
+  // Reading with Predicate Pushdown
+  cudf::io::parquet_reader_options read_opts =
+    cudf::io::parquet_reader_options::builder(cudf::io::source_info{filepath})
+      .filter(filter_expression);
+  auto result       = cudf::io::read_parquet(read_opts);
+  auto result_table = result.tbl->view();
+
+  // Expected result
```
bdice, line 5803: I would prefer to compute the expected value earlier, before reading the other file from disk.

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5602,584 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+}
+
+TEST_F(ParquetReaderTest, FilterMultiple2)
+{
+  // multiple conditions on same column.
+  using T = cudf::string_view;
+
+  auto const [src, filepath] = create_parquet_typed_with_stats<T>("FilterMultiple2.parquet");
+  auto const written_table   = src.view();
+  // 0-8000, 8001-16000, 16001-20000
```
bdice, line 5858: This comment seems out of date. Please check all the comments in this test file since most of the filters don't appear to match their comments.

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5602,584 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+    T constexpr add_value = 110;
+    auto add_v            = cudf::numeric_scalar(add_value, true);
+    auto add_lit          = cudf::ast::literal(add_v);
+    auto expr_8 = cudf::ast::operation(cudf::ast::ast_operator::ADD, filter_col1, add_lit);
+
+    auto si      = cudf::io::source_info(filepath);
+    auto builder = cudf::io::parquet_reader_options::builder(si).filter(expr_8);
+    EXPECT_THROW(cudf::io::read_parquet(builder), cudf::logic_error);
+
+    // Expected result
```
bdice, line 6126: Why does this test have an expected result if the expectation is to throw?

bdice (commented):
Review comments on files other than `predicate_pushdown.cpp`. I'll do a separate review for that file.

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,521 @@
...
+    if constexpr (cudf::is_timestamp<ToType>()) {
+      return static_cast<ToType>(
+        typename ToType::duration{static_cast<typename ToType::rep>(value)});
+    } else if constexpr (std::is_same_v<ToType, string_view>) {
+      return ToType{nullptr, 0};
+    } else {
+      return static_cast<ToType>(value);
+    }
+  }
+  // uses storage type as T
```
bdice, line 62: I'm surprised clang-format doesn't require blank lines between these definitions...
```suggestion

  // uses storage type as T
```

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,521 @@
...
+  template <typename T, CUDF_ENABLE_IF(cudf::is_dictionary<T>() or cudf::is_nested<T>())>
+  static T convert(uint8_t const* stats_val, size_t stats_size, cudf::io::parquet::Type const type)
+  {
+    CUDF_FAIL("unsupported type for stats casting");
+  }
+  template <typename T, CUDF_ENABLE_IF(cudf::is_boolean<T>())>
```
bdice, line 68:
```suggestion

  template <typename T, CUDF_ENABLE_IF(cudf::is_boolean<T>())>
```

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,521 @@
...
+  static T convert(uint8_t const* stats_val, size_t stats_size, cudf::io::parquet::Type const type)
+  {
+    CUDF_EXPECTS(type == BOOLEAN, "Invalid type and stats combination");
+    return targetType<T>(*reinterpret_cast<bool const*>(stats_val));
+  }
+
+  template <typename T,
+            CUDF_ENABLE_IF((!cudf::is_boolean<T>() and cudf::is_integral<T>()) or
```
bdice, line 76: Intent is a little clearer if you write this as "integral and not boolean".
```suggestion
            CUDF_ENABLE_IF((cudf::is_integral<T>() and !cudf::is_boolean<T>()) or
```

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,505 @@
...
+                                    ? colchunk.meta_data.statistics.min_value
+                                    : colchunk.meta_data.statistics.min;
+          auto const& max_value = colchunk.meta_data.statistics.max_value.size() > 0
+                                    ? colchunk.meta_data.statistics.max_value
+                                    : colchunk.meta_data.statistics.max;
+          // translate binary data to Type then to <T>
+          min.set_index(stats_idx, min_value, colchunk.meta_data.type);
+          max.set_index(stats_idx, max_value, colchunk.meta_data.type);
+          // if constexpr(std::is_integral_v<T> && sizeof(T)<16)
+          //   std::cout << min.val[stats_idx] << "," << max.val[stats_idx] << "\n";
```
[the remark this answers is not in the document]
bdice: This comment was resolved but the change was not applied.

bdice (commented):
A few more comments on `predicate_pushdown.cpp`. This completes my review pass for now.

cpp/src/io/parquet/reader_impl_helpers.hpp
karthikeyann: That's right. Using `std::list` or `std::deque` to avoid reference invalidation.

cpp/tests/io/parquet_test.cpp
karthikeyann: Expected result will also throw. The Expected result throw test is added to show where is the logic_error comes from.

cpp/tests/io/parquet_test.cpp
karthikeyann: Great suggestion. Didn't realize `EXPECT_LT` is available.

cpp/include/cudf/ast/expressions.hpp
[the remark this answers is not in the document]
karthikeyann: I thought about that too. But this would leave the user with overhead of managing this ordering. Column name reference seems easier to construct for the user.
Besides, they can still do that with the existing code, by using column_reference.

cpp/src/io/parquet/predicate_pushdown.cpp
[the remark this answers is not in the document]
karthikeyann: a separate PR would be nice. Looks like read_parquet_metadata would like to expose more information. So, this entire header might move from src/ to cudf/include/

cpp/src/io/parquet/predicate_pushdown.cpp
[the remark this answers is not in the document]
karthikeyann: It's a limitation of our implementation. It's not part of parquet specification. column to column comparison based filtering using the min, max statistics will yield lesser filtering of row groups typically. Other parquet libraries also doesn't implement column-to-column comparison based filtering.
It's an interesting feature. If requested by a user, we can do implement that too.

There is no such specification in parquet filtering. No specfication on how to reference a column too. Each library seems to do their own format.

cpp/include/cudf/ast/expressions.hpp
```diff
@@ -474,6 +502,48 @@ class operation : public expression {
...
+   *
+   * @param column_name Name of this column in the table metadata (provided when the expression is
+   * evaluated).
+   */
+  column_name_reference(std::string column_name) : column_name(std::move(column_name)) {}
+
+  /**
+   * @brief Get the column name.
+   *
+   * @return This name of the column reference
```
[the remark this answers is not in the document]
karthikeyann: updated

cpp/src/io/parquet/reader_impl_helpers.cpp
karthikeyann: added braces on all if conditions in the changes in this PR.

cpp/tests/io/parquet_test.cpp
karthikeyann: cross checked all conditions and updated all comments.

cpp/tests/io/parquet_test.cpp
karthikeyann: updated all comments

cpp/tests/io/parquet_test.cpp
karthikeyann: removed it.

cpp/tests/io/parquet_test.cpp
karthikeyann: done;

cpp/src/io/parquet/predicate_pushdown.cpp
karthikeyann: updated. nice suggestion.

cpp/include/cudf/ast/expressions.hpp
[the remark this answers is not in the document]
bdice: We discussed this offline. We want to provide users with a friendly API (using string names). Unfortunately that introduces some complexity to the visitor pattern, since AST filtering supports names but libcudf (and the AST transform) do not support names. We're settling for some comments explaining the conflicting goals, since there wasn't an obvious way to make this better (balancing code duplication, user friendliness, etc.).

cpp/tests/io/parquet_test.cpp
karthikeyann: moved all expected result above the test code.

cpp/tests/io/parquet_test.cpp
karthikeyann: updated all test filter comments.

hyperbolic2346 (commented):
Just one other question from me right now.

cpp/src/io/parquet/predicate_pushdown.cpp
etseidl: `vmin > val || vmax < val` will not suffice in this case, because it does not handle the case `vmin < val < vmax`.  It would have to be `vmin > val || vmin < val || vmax < val || vmax > val` to work, which is far more prolix than `vmin != val || vmax != val`.  Consider a  row group with values `{1, 2, 3}` and predicate `val != 2`.  Neither `vmin > val` nor `vmax < val` will evaluate to `true`, so the row group will not be visited.  But this row group clearly contains valuues that satisfy the predicate `val != 2`.

The NaN case does warrant more investigation. I'm not at all sure how the statistics would handle that.

bdice: Oof. Thanks, I see it now.

Let's try to add tests that involve NaN values.

karthikeyann: NAN values are stored as nulls in statistics, which means, its binary blob is empty. if a binary is empty, we read that row group anyway. so, it gets read and filtered at the end too.

cpp/src/io/parquet/reader_impl.cpp
karthikeyann: @hyperbolic2346 since chunked reader does not support user specifying the row group too, filtering is not supported by chunked row group reader.

hyperbolic2346 (approved):
No further requests from me. I'm ok once Bradley signs off on this.

bdice:
Thanks Mike — I’ll review again tomorrow.

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,527 @@
...
+  auto predicate_col =
+    cudf::compute_column(stats_table, stats_ast.get(), rmm::mr::get_current_device_resource());
+  auto predicate = predicate_col->view();
+  CUDF_EXPECTS(predicate.type().id() == cudf::type_id::BOOL8,
+               "Filter expression must return a boolean column");
+
+  auto num_bitmasks = num_bitmask_words(predicate.size());
+  std::vector<bitmask_type> host_bitmask(num_bitmasks, ~bitmask_type{0});
+  if (predicate.nullable()) {
+    CUDF_CUDA_TRY(cudaMemcpy(host_bitmask.data(),
```
bdice, line 438: Stream usage is not handled very well in this function. Currently we're shifting over all public APIs to accept streams (detail APIs should already take streams). I would like to establish the same conventions for the new APIs in this file.

Can we accept a stream in this function, and use it throughout? If not, can we at least declare a variable like `auto stream = cudf::get_default_stream()` and use it?

This line should use `cudaMemcpyAsync` and the rest of this function should use the same stream, where needed.

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,527 @@
...
+std::reference_wrapper<ast::expression const> named_to_reference_converter::visit(
+  ast::operation const& expr)
+{
+  auto const operands = expr.get_operands();
+  auto op             = expr.get_operator();
+  auto new_operands   = visit_operands(operands);
+  if (cudf::ast::detail::ast_operator_arity(op) == 2)
+    _operators.emplace_back(op, new_operands.front(), new_operands.back());
+  else if (cudf::ast::detail::ast_operator_arity(op) == 1)
+    _operators.emplace_back(op, new_operands.front());
```
bdice, line 510: Always use braces.
```suggestion
  if (cudf::ast::detail::ast_operator_arity(op) == 2) {
    _operators.emplace_back(op, new_operands.front(), new_operands.back());
  } else if (cudf::ast::detail::ast_operator_arity(op) == 1) {
    _operators.emplace_back(op, new_operands.front());
  }
```

bdice (changes requested):
Some comments -- mostly minor changes but I definitely want to get stream conventions figured out before approving. I know that our I/O code might not be the best example of stream conventions today, but we need to get the entirety of libcudf moving in that direction. New APIs should accept and leverage streams correctly.

cpp/tests/io/parquet_test.cpp
```diff
@@ -5618,4 +5620,648 @@ TEST_F(ParquetWriterTest, NoNullsAsNonNullable)
...
+  {
+    auto filter_col1 = cudf::ast::column_reference(3);
+    T constexpr low  = 100;
+    auto lov         = cudf::numeric_scalar(low, true);
+    auto low_lot     = cudf::ast::literal(lov);
+    auto expr        = cudf::ast::operation(cudf::ast::ast_operator::LESS, filter_col1, low_lot);
+
+    auto builder = cudf::io::parquet_reader_options::builder(si).filter(expr);
+    EXPECT_THROW(cudf::io::read_parquet(builder), cudf::logic_error);
+    // FIXME: compute_column does not throw logic error
```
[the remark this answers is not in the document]
karthikeyann: This is an issue in `compute_column`. I will create a minimal repro and file an issue.

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,527 @@
...
+  // integral but not boolean, and fixed_point, and chrono.
+  template <typename T,
+            CUDF_ENABLE_IF((cudf::is_integral<T>() and !cudf::is_boolean<T>()) or
+                           cudf::is_fixed_point<T>() or cudf::is_chrono<T>())>
+  static T convert(uint8_t const* stats_val, size_t stats_size, cudf::io::parquet::Type const type)
+  {
+    switch (type) {
+      case INT32: return targetType<T>(*reinterpret_cast<int32_t const*>(stats_val));
+      case INT64: return targetType<T>(*reinterpret_cast<int64_t const*>(stats_val));
+      case INT96:  // Deprecated
```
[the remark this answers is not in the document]
karthikeyann: deprecated in parquet spec.

cpp/src/io/parquet/predicate_pushdown.cpp
```diff
@@ -0,0 +1,527 @@
...
+      case DOUBLE: return targetType<T>(*reinterpret_cast<double const*>(stats_val));
+      default: CUDF_FAIL("Invalid type and stats combination");
+    }
+  }
+
+  template <typename T, CUDF_ENABLE_IF(std::is_same_v<T, string_view>)>
+  static T convert(uint8_t const* stats_val, size_t stats_size, cudf::io::parquet::Type const type)
+  {
+    switch (type) {
+      case BYTE_ARRAY:
```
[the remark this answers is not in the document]
karthikeyann: I thought about this and experiemented a bit while reviewing another PR.
If first case statement has no statements upto next case statement, the compiler does not show any warnings even after warning is enabled. That's why it was not added. But, I will add them anyway.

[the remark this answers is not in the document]
bdice: I believe `-Wimplicit-fallthrough` is the warning flag. We try to use this universally in libcudf, from past conversations I’ve had on the topic. Implicit fallthrough is considered by some to be a design defect in C/C++.

cpp/src/io/parquet/predicate_pushdown.cpp
karthikeyann: cleaned up stream and mr usage.

cpp/src/io/parquet/reader_impl.cpp
karthikeyann: no error is thrown if user specifies the row groups for chunked reader object creation. it's silently ignored. So, the same convention is followed here. @hyperbolic2346 is this behaviour okay?

cpp/src/io/parquet/reader_impl_helpers.hpp
```diff
@@ -195,4 +215,64 @@ class aggregate_reader_metadata {
...
+  named_to_reference_converter(std::optional<std::reference_wrapper<ast::expression const>> expr,
+                               table_metadata const& metadata)
+    : metadata(metadata)
+  {
+    if (!expr.has_value()) return;
+    // create map for column name.
+    std::transform(metadata.schema_info.begin(),
+                   metadata.schema_info.end(),
+                   std::inserter(column_name_to_index, column_name_to_index.end()),
+                   [i = 0](auto const& data) mutable { return make_pair(data.name, i++); });
```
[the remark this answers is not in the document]
karthikeyann: updated to use zip iterator.

bdice (approved):
It looks like all the requested changes in my previous review were addressed. I think this is close enough to approve, but we need to address the conversation about compatibility with the chunked reader in https://github.com/rapidsai/cudf/pull/13348/files#r1271118219 before merging.

raydouglass (approved):

cpp/src/io/parquet/reader_impl.cpp
nvdbaranec: The chunked reader and row-group specification should work together fine (though I don't think we have a test specifically for it).  The only input to the chunked reader is "number of bytes in a chunk" so it doesn't care at all if the row groups being inspected in the first place are the entire file or part of it.

PR #13350, opened 2023-05-15
Use _compile_or_get in JIT groupby apply

brandon-b-miller:
During https://github.com/rapidsai/cudf/pull/13337 I realized we should have been using this function in groupby, so caching logic and any runtime checks for strings and groupby don't have to be duplicated.


python/cudf/cudf/core/udf/utils.py
```diff
@@ -308,7 +310,7 @@ def _compile_or_get(frame, func, args, kernel_getter=None):
         raise TypeError("only scalar valued args are supported by apply")
 
     # check to see if we already compiled this function
-    cache_key = _generate_cache_key(frame, func, args)
+    cache_key = _generate_cache_key(frame, func, args, suffix=suffix)
```
vyasr, line 313: Is it important for us to distinguish between functions compiled for groupby-apply vs normal apply?

brandon-b-miller: Yes, it is. The reason is that the cache key is generated based on the dtypes of the input frame and the function itself. It's possible you could get a cache hit for a groupby function where you wanted an apply function, which I am guessing would cause a segfault. So this parameter breaks any degeneracies.

vyasr (approved):

PR #13355, opened 2023-05-16
Ensure columns have valid null counts in CUDF JNI.

### LINKED ISSUE #13353: [FEA] JNI should calculate null count when it is unknown
Java currently allows column views to be constructed with unknown null counts.  To prepare for #11968, the first step is to ensure libcudf is not called with unknown null counts from Java.

The cudf Java JNI code in ColumnViewJni.cpp should check if the Java interface was called with an unknown null count.  If so, it should compute the null count on the validity buffer and pass the now known null count to the column_view constructor.

Alternatively we could rip out the ability to pass an unknown null count in the Java API.  However that is a much more involved and risky change at this point for 23.06 and affects downstream applications.

mythrocks:
## Description

Fixes #13353.
Depends on #13345.
    
In preparation for #11968, this change ensures that columns constructed from CUDF JNI do not have their null counts set to `UNKNOWN_NULL_COUNT` (i.e. `-1`). In cases where the caller invokes JNI functions with `UNKNOWN_NULL_COUNT`, the JNI layer computes the concrete null count from the validity mask, and sets this value in the column.

The current Java API remains unchanged; there should be no impact to user code.
    
The option to specify an optional null count through the Java API will likely be removed at a later date.
    
Signed-off-by: MithunR <mythrocks@gmail.com>

## Checklist


mythrocks:
This diff includes the changes in #13345, but it applies only to the Java API.

jlowe (approved):

java/src/main/native/src/ColumnViewJni.cpp
```diff
@@ -1893,6 +1893,11 @@ JNIEXPORT jlong JNICALL Java_ai_rapids_cudf_ColumnView_makeCudfColumnView(
       j_null_count = 0;
     }
 
+    if (j_null_count == cudf::UNKNOWN_NULL_COUNT) {
```
ttnghia, line 1896: Since we are removing this constant, can we use this?
```suggestion
    if (j_null_count < 0) {
```

davidwendt: We will eventually be removing the `cudf::UNKNOWN_NULL_COUNT` constant -- hopefully in this release since there should be no usage or reliance on it. If you still need the constant for the internal JNI transition, it may be best create your own constant.

java/src/main/native/src/row_conversion.cu
```diff
@@ -2258,10 +2258,17 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const &input,
...
-                                                std::move(string_data_cols[string_idx]),
-                                                std::move(*string_data.null_mask.release()),
-                                                cudf::UNKNOWN_NULL_COUNT);
+        auto const null_count = [&] {
+          auto const &null_mask = *string_data.null_mask;
+          return null_mask.data() ?
+                     cudf::null_count(static_cast<bitmask_type const *>(null_mask.data()), 0,
+                                      num_rows) :
+                     0;
+        }();
```
ttnghia, line 2267: Probably this is just this?
```
auto const null_count = string_row_offset_columns[string_idx]->null_count();
```

vyasr:
https://github.com/rapidsai/cudf/pull/13345 is merged, so @mythrocks you should be able to rebase on the latest 23.06 now.

@jlowe Mithun mentioned that for some partitioning applications Spark really doesn't need to know the null count of partitioned columns because the columns are ultimately reassembled and therefore only the total count/mask matters. As a future optimization, I noted that in those cases it may make sense to eventually switch to passing around the column's device buffers instead of the columns themselves if avoiding the extra intermediate null counts has a significant performance benefit.

python/cudf/cudf/_lib/cpp/null_mask.pxd
```diff
@@ -17,6 +17,12 @@ cdef extern from "cudf/null_mask.hpp" namespace "cudf" nogil:
         column_view view
     ) except +
 
+    cdef device_buffer copy_bitmask "cudf::copy_bitmask" (
+        const bitmask_type * mask,
+        size_type begin_bit,
+        size_type end_bit
+    ) except +
+
```
mythrocks, line 25: Incorrect merge. I will remove this change.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -24,7 +24,7 @@ from libcpp.memory cimport make_unique, unique_ptr
 from libcpp.utility cimport move
 from libcpp.vector cimport vector
 
-from rmm._lib.device_buffer cimport DeviceBuffer
+from rmm._lib.device_buffer cimport DeviceBuffer, device_buffer
```
mythrocks, line 27: Incorrect merge. I will remove this change.

java/src/main/native/src/ColumnViewJni.cpp
mythrocks: Agreed. I've changed this as per @ttnghia's suggestion.

java/src/main/native/src/row_conversion.cu
mythrocks: Agreed. Fixed.

Note: As per discussion with @hyperbolic2346, this code is likely cruft. I'll raise a separate issue to have `row_conversion.*` removed from CUDF.

python/cudf/cudf/_lib/cpp/null_mask.pxd
mythrocks: Removed.

python/cudf/cudf/_lib/column.pyx
mythrocks: Removed.

ttnghia (approved):

mythrocks:
The tests have passed. I'm not sure what's to be done about the code coverage thingie.

mythrocks:
Thanks for the reviews, all. I've merged this change.

PR #13363, opened 2023-05-16
Add log messages about kvikIO compatibility mode

vuule:
## Description
Log whether kvikIO's compatibility mode is on for a given file input/output.

## Checklist

### DIFF
--- a/cpp/src/io/utilities/data_sink.cpp
+++ b/cpp/src/io/utilities/data_sink.cpp
@@ -38,6 +38,8 @@ class file_sink : public data_sink {
 
     if (detail::cufile_integration::is_kvikio_enabled()) {
       _kvikio_file = kvikio::FileHandle(filepath, "w");
+      CUDF_LOG_INFO("Writing a file using kvikIO, with compatibility mode {}.",
+                    _kvikio_file.is_compat_mode_on() ? "on" : "off");
     } else {
       _cufile_out = detail::make_cufile_output(filepath);
     }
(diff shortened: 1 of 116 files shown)


cpp/src/io/utilities/data_sink.cpp
```diff
@@ -38,6 +38,8 @@ class file_sink : public data_sink {
 
     if (detail::cufile_integration::is_kvikio_enabled()) {
       _kvikio_file = kvikio::FileHandle(filepath, "w");
+      CUDF_LOG_INFO("Writing a file using kvikIO, with compatibility mode {}.",
+                    _kvikio_file.is_compat_mode_on() ? "on" : "off");
```
ttnghia, line 42: Should we also log for the `else` branch?

cpp/src/io/utilities/datasource.cpp
```diff
@@ -42,6 +42,8 @@ class file_source : public datasource {
   {
     if (detail::cufile_integration::is_kvikio_enabled()) {
       _kvikio_file = kvikio::FileHandle(filepath);
+      CUDF_LOG_INFO("Reading a file using kvikIO, with compatibility mode {}.",
```
ttnghia, line 45: Similarly.

cpp/src/io/utilities/data_sink.cpp
vuule: No need to log anything in the else branch because it calls `make_cufile_output`, which includes the logging for that path (whether GDS was used or not).
Same goes for the other file, just with `make_cufile_input`.

ttnghia (approved):

hyperbolic2346 (approved):

PR #13372, opened 2023-05-17
Remove UNKNOWN_NULL_COUNT

### LINKED ISSUE #11968: [FEA] Require null counts on column construction

vyasr:
## Description

This is the final PR for removing `UNKNOWN_NULL_COUNT` and the implicit kernel launch in the `null_count` methods of `column` and `column_view`. 

Depends on #13355 and #13341.

Closes #11968

## Checklist


vyasr:
There are a couple of Java failures left that I can reproduce locally. Will attempt to debug them myself, but may need to pull in some advice.

cpp/tests/groupby/var_tests.cpp
```diff
@@ -167,3 +167,34 @@ TYPED_TEST(groupby_var_test, dictionary)
...
+  auto result1 = gb_obj.aggregate(requests);
+
+  // This agg forces a sort groupby.
+  auto agg2 = cudf::make_quantile_aggregation<cudf::groupby_aggregation>({0.25});
+  requests[0].aggregations.push_back(std::move(agg2));
+
+  auto result2 = gb_obj.aggregate(requests);
+
+  EXPECT_EQ(result1.second[0].results[0]->null_count(), result2.second[0].results[0]->null_count());
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(*result1.second[0].results[0], *result2.second[0].results[0]);
```
davidwendt, line 199: The `CUDF_TEST_EXPECT_COLUMNS_EQUAL` includes comparing the null counts so I don't think you need the `EXPECT_EQ`.

cpp/include/cudf/column/column_view.hpp
```diff
@@ -511,10 +498,6 @@ class mutable_column_view : public detail::column_view_base {
    * @brief Construct a `mutable_column_view` from pointers to device memory for
    *the elements and bitmask of the column.
```
davidwendt, line 499:
```suggestion
   * the elements and bitmask of the column.
```

cpp/src/groupby/sort/group_std.cu
```diff
@@ -20,12 +20,14 @@
 #include <cudf/column/column_factories.hpp>
 #include <cudf/column/column_view.hpp>
 #include <cudf/detail/aggregation/aggregation.hpp>
+#include <cudf/detail/null_mask.hpp>
```
davidwendt, line 23:
```suggestion
```
Probably not needed?

cpp/tests/groupby/var_tests.cpp
vyasr: Ah yes this was a copy-paste from an earlier iteration.

cpp/src/groupby/sort/group_std.cu
vyasr: Yup this was for `count_nulls` before I switched to the `atomicAdd` approach.

mythrocks (approved):
I'm +1 on the JNI changes. 

Note that `spark-rapids` does not depend on this code. `row_conversion.cu` will be removed as part of https://github.com/rapidsai/cudf/issues/13374. The splash damage here should be limited.

cpp/src/column/column.cu
```diff
@@ -117,37 +117,16 @@ mutable_column_view column::mutable_view()
...
-size_type column::null_count() const
-{
-  CUDF_FUNC_RANGE();
-  if (_null_count <= cudf::UNKNOWN_NULL_COUNT) {
-    _null_count = cudf::detail::null_count(
-      static_cast<bitmask_type const*>(_null_mask.data()), 0, size(), cudf::get_default_stream());
-  }
-  return _null_count;
-}
+size_type column::null_count() const { return _null_count; }
```
ttnghia, line 129: This may be moved to the header for better inline.

cpp/src/column/column_view.cpp
```diff
@@ -63,15 +63,7 @@ column_view_base::column_view_base(data_type type,
...
-// If null count is known, returns it. Else, compute and return it
-size_type column_view_base::null_count() const
-{
-  if (_null_count <= cudf::UNKNOWN_NULL_COUNT) {
-    _null_count = cudf::detail::null_count(
-      null_mask(), offset(), offset() + size(), cudf::get_default_stream());
-  }
-  return _null_count;
-}
+size_type column_view_base::null_count() const { return _null_count; }
```
ttnghia, line 66: This may be moved to the header for better inline.

cpp/src/column/column.cu
vyasr: Done.

ttnghia (approved):

karthikeyann (approved):

PR #13378, opened 2023-05-18
JNI: Remove cleaned objects in memory cleaner

### LINKED ISSUE #13412: [FEA] Remove cleaned objects in memory cleaner
JNI `MemoryCleaner` holds a set of `CleanerWeakReference` which can go to very large size when doing NDS testing while full GC is not excuted frequently.
For the set, please refer to:  
```
  private static final Set<CleanerWeakReference> all =
      Collections.newSetFromMap(new ConcurrentHashMap()); // We want to be thread safe
```
https://github.com/rapidsai/cudf/blob/v23.06.00a/java/src/main/java/ai/rapids/cudf/MemoryCleaner.java#L148-L149
I verified that the full GC will shrink the set. But in practice the full GC may be not executed for a long time.
And I verified there is no leak on the resources that the set references.

Not sure if this large size set can cause OOM, but we can remove the closed items as soon as possible to save memory.

The `MemoryCleaner.Cleaner` already has `isClean`,  because the cleaned object will not be used again.
We can safely remove the cleaned items.

Refer to:
https://github.com/rapidsai/cudf/blob/v23.06.00a/java/src/main/java/ai/rapids/cudf/MemoryCleaner.java#L144

Update the thread code to remove cleaned items
https://github.com/rapidsai/cudf/blob/v23.06.00a/java/src/main/java/ai/rapids/cudf/MemoryCleaner.java#L182-L207

**Additional context**
The K8s will kill executor if the memory of the Pod running this executor exceeds the definded value which is typically `spark.executor.memory` + `spark.executor.memoryOverhead` and report OOM which is a kind of K8s OOM.
Not sure if the large size of CleanerWeakReference is the cause.

Also reported in https://github.com/NVIDIA/spark-rapids/issues/8305

res-life:
## Description
closes #13412
Remove weak references of cleaned resources when a resource is cleaned.
The cleaned objects are never leaked, it's safe to remove the weak references. 
This is to reduce the memory usage.

## Checklist
  No need.
 No need.


res-life:
build

java/src/main/java/ai/rapids/cudf/MemoryCleaner.java
```diff
@@ -199,6 +200,12 @@ static void setDefaultGpu(int defaultGpuId) {
             log.error("CAUGHT EXCEPTION WHILE TRYING TO CLEAN " + next, t);
           }
           all.remove(next);
+        } else {
+          long currentTime = System.currentTimeMillis();
+          if (currentTime - lastRemoveCleanedObjectsTime > 10 * 1000L) { // every 10 seconds
+            all.removeIf(e -> e.cleaner.isClean());
```
jlowe, line 206: Why are we polling to clean objects when we can proactively remove things from `all` when they are properly closed?  Saves from doing a full scan of potentially very many live objects in `all` and should be more efficient.

There's a couple of ways to do this.  One is to implement equals and hashcode for `CleanerWeakReference` so we can remove our cleaner from all (by building a temporary instance to do the remove).  More efficiently would be to rework the `all` and Cleaner setups so instead of `CleanerWeakReference` pointing to `Cleaner`, we remove `CleanerWeakReference` entirely and have `Cleaner` have the `isRmmBlocker` field a normal `WeakReference` to the object it is associated with.  `all` would simply be a hash set of the cleaner instances, and then it's trivial for a cleaner to remove itself from `all` when it cleans.

java/src/main/java/ai/rapids/cudf/MemoryCleaner.java
```diff
@@ -199,6 +200,12 @@ static void setDefaultGpu(int defaultGpuId) {
             log.error("CAUGHT EXCEPTION WHILE TRYING TO CLEAN " + next, t);
           }
           all.remove(next);
+        } else {
+          long currentTime = System.currentTimeMillis();
+          if (currentTime - lastRemoveCleanedObjectsTime > 10 * 1000L) { // every 10 seconds
```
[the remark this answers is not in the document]
res-life: Now proactively remove.

java/src/main/java/ai/rapids/cudf/MemoryCleaner.java
res-life: Done.
Now proactively remove.

res-life:
build

java/src/main/java/ai/rapids/cudf/MemoryCleaner.java
res-life: Use cleaner id as the key of the Map. The id is unique and can represent a unique resource.

res-life:
The pipeline shows: Base branch is not under active development
Will retargate to branch-23.08

res-life:
build

java/src/main/java/ai/rapids/cudf/MemoryCleaner.java
jlowe, line 1: copyrights need to be updated

jlowe (approved):

revans2 (approved):
looks good to me.

res-life:
build

mythrocks (approved):

res-life:
build

PR #13386, opened 2023-05-18
[JNI] Pass this ColumnVector to the onClosed event handler

abellina:
Putting this up as an optimization for ColumnVector.EventHandler. 

As it stands today, code in the spark-rapids plugin that wants to use this is having to java objects that encapsulate state about the columns around reference counting and what represents the bag of `ColumnVector`s that are spillable at any given time. If we pass the `ColumnVector` instance in the event handler, one of these objects can be removed, and we can simplify the implementation in the plugin. I am putting this up as draft while my local tests pass, but I think it should be fairly straightforward.

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/ColumnVector.java
+++ b/java/src/main/java/ai/rapids/cudf/ColumnVector.java
@@ -50,10 +50,11 @@ public final class ColumnVector extends ColumnView {
      *
      * @note the callback is invoked with this `ColumnVector`'s lock held.
      *
+     * @param cv - a reference to the ColumnVector we are closing
      * @param refCount - the updated ref count for this ColumnVector at the time
      *                 of invocation
      */
-    void onClosed(int refCount);
+    void onClosed(ColumnVector cv, int refCount);
   }
 
   private static final Logger log = LoggerFactory.getLogger(ColumnVector.class);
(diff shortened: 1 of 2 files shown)


abellina:
Marked as non-breaking because code that would use it is not checked in to spark-rapids yet.

jlowe (approved):

revans2 (approved):

java/src/main/java/ai/rapids/cudf/ColumnVector.java
```diff
@@ -50,10 +50,11 @@ public interface EventHandler {
      *
      * @note the callback is invoked with this `ColumnVector`'s lock held.
      *
+     * @param cv - a reference to the ColumnVector we are closing
      * @param refCount - the updated ref count for this ColumnVector at the time
      *                 of invocation
      */
-    void onClosed(int refCount);
+    void onClosed(ColumnVector cv, int refCount);
```
ttnghia, line 57: I assume that an `EventHandler` gets access directly to the current `ColumnVector` object.

abellina: Not in this case. When this is invoked in in the java class that has implemented `EventHandler`, `this` is the reference to this specifc event handler. In order to get access to `ColumnVector` we would have to subclass `ColumnVector` and then override a protected (and default implemented) `onClosed`. We may gravitate towards something like that to make spill-ability really transparent but I don't think we have thought through that yet.

abellina: But by passing `cv` here, we do get access to the public interface of `ColumnVector` within the callback.

PR #13395, opened 2023-05-19
Require numba >= 0.57

brandon-b-miller:
This PR sets the lower bound for `numba` to version 0.57.

### DIFF
--- a/conda/recipes/cudf/meta.yaml
+++ b/conda/recipes/cudf/meta.yaml
@@ -53,7 +53,7 @@ requirements:
     - cython >=0.29,<0.30
     - scikit-build >=0.13.1
     - setuptools
-    - numba >=0.56.4,<0.57
+    - numba >=0.57
     - dlpack >=0.5,<0.6.0a0
     - pyarrow =11
     - libcudf ={{ version }}
@@ -65,7 +65,7 @@ requirements:
     - typing_extensions
     - pandas >=1.3,<1.6.0dev0
     - cupy >=12.0.0
-    - numba >=0.56.4,<0.57
+    - numba >=0.57
     - numpy >=1.21,<1.24  # Temporarily upper bound numpy to avoid overflow deprecations
     - {{ pin_compatible('pyarrow', max_pin='x.x.x') }}
     - libcudf {{ version }}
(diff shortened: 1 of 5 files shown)


dependencies.yaml
```diff
@@ -507,4 +507,4 @@ dependencies:
       - output_types: [conda, requirements, pyproject]
         packages:
           - dask-cuda==23.6.*
-          - *numba
+          - numba>=0.57
```
bdice, line 510: Leave this `*numba` as it was, and use `&numba` above. That uses a YAML anchor to enforce that the value is copied.

bdice:
Please also remove this line: https://github.com/rapidsai/cudf/blob/209ab6e188ff6ae4b5ac4d4c9cf9a5970091038a/ci/release/apply_wheel_modifications.sh#L29

conda/recipes/cudf/meta.yaml
```diff
@@ -53,7 +53,7 @@ requirements:
     - cython >=0.29,<0.30
     - scikit-build >=0.13.1
     - setuptools
-    - numba >=0.56.4,<0.57
+    - numba >=0.57
```
bdice, line 56: Is numba actually a build-time requirement? It's not listed in `[build-system.requires]` for wheels. If not, please remove.

gmarkall:
The notebooks tests seemed to have failed because they didn't get the latest ptxcompiler - it looks like 0.7.0 came from the rapidsai-nightly channel:

```
   + ptxcompiler                0.7.0  cuda_11_py310_gaec9cc8_44            rapidsai-nightly/linux-64           6MB
```

I'm surprised we have ptxcompiler packages in this channel at all - is there a reason we don't just get them from conda-forge?

gmarkall (approved):
This looks good to me (assuming CI eventually passes once the right ptxcompiler packages are picked up)

conda/recipes/cudf/meta.yaml
bdice: I dealt with this separately in #13406, since it looks like this PR might be closed in favor of #13337.

jakirkham:
Closing in favor of PR ( https://github.com/rapidsai/cudf/pull/13395 )

Changes from that PR are needed to resolve the errors seen here

PR #13396, opened 2023-05-19
Refactor ORC reader

ttnghia:
This rewrites the implementation ORC reader, preparing for the next changes in ORC chunked reader:
 * Remove the `stream` parameter from the `read()` API. Instead, `stream` and `mr` are stored upon reader construction.
 * Remove the inclusion of unused headers.
 * Remove unused variable(s).
 * Rename and re-order member variables.
 * Move (almost) all member functions into free functions. This reduces header dependency between TUs for any future changes.
 * Re-order the functions by their called order.

There is not any new functionality was added or removed.


cpp/include/cudf/io/detail/orc.hpp
```diff
@@ -70,11 +70,9 @@ class reader {
    * @brief Reads the entire dataset.
    *
    * @param options Settings for controlling reading behavior
-   * @param stream CUDA stream used for device memory operations and kernel launches.
-   *
    * @return The set of columns along with table metadata
    */
-  table_with_metadata read(orc_reader_options const& options, rmm::cuda_stream_view stream);
+  table_with_metadata read(orc_reader_options const& options);
```
ttnghia, line 75: Remove `stream` parameter. It was already stored during reader construction.

cpp/src/io/utilities/row_selection.hpp
```diff
@@ -27,14 +27,14 @@ namespace cudf::io::detail {
...
+ * @param skip_rows skip_rows as passed by the user
+ * @param num_rows_opt Optional num_rows as passed by the user
  * @param num_source_rows number of rows in the ORC file(s)
  * @return A std::pair containing the number of rows to skip and the number of rows to read
  *
  * @throw std::overflow_exception The requested number of rows exceeds the column size limit
  */
 std::pair<uint64_t, size_type> skip_rows_num_rows_from_options(
-  uint64_t skip_rows_opt, std::optional<size_type> const& num_rows_opt, uint64_t num_source_rows);
+  uint64_t skip_rows, std::optional<size_type> const& num_rows_opt, uint64_t num_source_rows);
```
ttnghia, line 38: Rename to make clear that this is not an optional.

cpp/src/io/orc/aggregate_orc_metadata.cpp
```diff
@@ -155,19 +155,18 @@ aggregate_orc_metadata::aggregate_orc_metadata(
 
 std::tuple<int64_t, size_type, std::vector<metadata::stripe_source_mapping>>
 aggregate_orc_metadata::select_stripes(
-  std::vector<std::vector<size_type>> const& user_specified_stripes,
-  int64_t skip_rows_opt,
+  uint64_t skip_rows,
   std::optional<size_type> const& num_rows_opt,
+  std::vector<std::vector<size_type>> const& user_specified_stripes,
```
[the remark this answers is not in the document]
vuule: why reorder?

cpp/src/io/utilities/row_selection.hpp
vuule: OMG, took me a minute :)
`_opt` is not for optional, it stands for option :D 
Maybe both `_opt` suffixes can be removed, they don't seem to work as intended.

cpp/src/io/orc/reader_impl.hpp
```diff
@@ -91,143 +63,27 @@ class reader::impl {
...
-   * @return Timezone table with timestamp offsets
-   */
-  std::unique_ptr<table> compute_timezone_table(
-    std::vector<cudf::io::orc::metadata::stripe_source_mapping> const& selected_stripes,
-    rmm::cuda_stream_view stream);
-
- private:
-  rmm::mr::device_memory_resource* _mr = nullptr;
-  std::vector<std::unique_ptr<datasource>> _sources;
+  std::vector<std::unique_ptr<datasource>> const _sources;  // Unused but owns data for others
```
vuule, line 78: This comment does not match

cpp/src/io/orc/reader_impl.hpp
```diff
@@ -91,143 +63,27 @@ class reader::impl {
...
 
-  bool _use_index{true};
-  bool _use_np_dtypes{true};
-  std::vector<std::string> decimal128_columns;
-  data_type _timestamp_type{type_id::EMPTY};
-  reader_column_meta _col_meta{};
+  data_type const _timestamp_type;  // Override output timestamp resolution
+  bool const _use_index;            // Enable or disable attempt to use row index for parsing
+  bool const _use_np_dtypes;        // Enable or disable the conversion to numpy-compatible dtypes
+  std::vector<std::string> const _decimal128_columns;   // Control decimals conversion
```
vuule, line 85: I'm mildly against const members as they impact copy/move. I'm not sure if there's a guideline on this.

cpp/src/io/orc/reader_impl.cu
```diff
@@ -462,28 +403,27 @@ rmm::device_buffer reader::impl::decompress_stripe_data(
...
 void update_null_mask(cudf::detail::hostdevice_2dvector<gpu::ColumnDesc>& chunks,
-                      std::vector<column_buffer>& out_buffers,
+                      host_span<column_buffer> out_buffers,
                       rmm::cuda_stream_view stream,
                       rmm::mr::device_memory_resource* mr)
 {
-  auto const num_stripes = chunks.size().first;
-  auto const num_columns = chunks.size().second;
+  const auto num_stripes = chunks.size().first;
+  const auto num_columns = chunks.size().second;
```
vuule, line 423: const east got reverted, presumably merge conflict :)

vuule (commented):
I'm glad this PR is kept simple. Even this much is hard to review because of code shuffle. It should make future PRs much easier on the eye :)

cpp/src/io/orc/reader_impl.hpp
ttnghia: That will prevent copying, but not moving. I think we should not allow copying but we currently don't do that explicitly so leaving `const` members is just fine. In addition, members storing input parameters will not be changed thus marking them as `const` should be better to communicate their intention (store just unchanged input parameters).

cpp/src/io/orc/reader_impl.hpp
ttnghia: Actually this is correct. `_sources` are not used anywhere but members of `_metadata` has pointers refer to it and will be accessed later.

ttnghia: I'll update the comment.

cpp/src/io/orc/aggregate_orc_metadata.cpp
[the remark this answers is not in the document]
ttnghia: Sorry had to reorder for some reason but that reason is outdated so I revert this back.

cpp/src/io/utilities/row_selection.hpp
ttnghia: Done.

cpp/src/io/orc/reader_impl.hpp
vuule: AFAIK it's the other way around. An object can't be moved from when it has const members because move generally modifies the members (e.g. moving a vector leaves the moved from object empty). So const members prevent moves, and objects have to be copied instead. Not a big reason, but still makes the situation iffy.

cpp/src/io/orc/aggregate_orc_metadata.cpp
```diff
@@ -156,18 +156,17 @@ aggregate_orc_metadata::aggregate_orc_metadata(
 std::tuple<int64_t, size_type, std::vector<metadata::stripe_source_mapping>>
 aggregate_orc_metadata::select_stripes(
   std::vector<std::vector<size_type>> const& user_specified_stripes,
-  int64_t skip_rows_opt,
+  uint64_t skip_rows,
   std::optional<size_type> const& num_rows_opt,
```
vuule, line 160: _opt suffix need to be removed here as well

cpp/src/io/orc/reader_impl.hpp
vuule: That's surprising, did not realize we always use sources through metadata objects :)

vuule (approved):

cpp/src/io/orc/reader_impl.cu
```diff
@@ -413,33 +342,45 @@ rmm::device_buffer reader::impl::decompress_stripe_data(
...
+    // Check if any block has been failed to decompress.
+    // Not using `thrust::any` or `thrust::count_if` to defer stream sync.
+    thrust::for_each(
+      rmm::exec_policy(stream),
+      thrust::make_counting_iterator(std::size_t{0}),
+      thrust::make_counting_iterator(inflate_res.size()),
+      [results           = inflate_res.begin(),
+       any_block_failure = any_block_failure.device_ptr()] __device__(auto const idx) {
+        if (results[idx].status != compression_status::SUCCESS) { *any_block_failure = true; }
+      });
```
mythrocks, line 355: :clap:

mythrocks: Reading the new code, it isn't clear that this does not care about synchronization; the old code had a comment indicating that this needn't be atomic. Could we pull that comment in here?

mythrocks: Additionally, there might be value moving the definition of `any_block_failure` closer to this loop.

mythrocks: Additionally, I wonder what the current state of `thrust::any_of()`'s performance is. Should we consider `thrust::count_if()` or another stand-in for `any_of()`? 

Unless I'm mistaken, all the kernel needs to return here is a single bool. The `host_device_vector[1]` seems unnecessary.

ttnghia: This check is called multiple times for multiple data blocks but the output boolean value is only needed once after processing all blocks (which means it needs only 1 stream sync) so we can avoid multiple unnecessary stream sync.

mythrocks: Let me see if I understand:

Why is `any_block_failure` a `host_device_vector` of size `1`? Why have an output variable at all? 

Why not use, say, `thrust::any_of()` to check `results[idx].status != SUCCESS`? (Because `any_of()` is slow?)

How about `count_if()`? Because that still requires a thread-sync before reducing the result?

ttnghia: We need to do the check async for each block. Rather than sync after processing each block, we try to avoid sync altogether until we process all blocks. Then, we sync the check result and verify it on host.

`host_device_vector` of size `1` is just to store one boolean value (set to `false` if there is there is failure).

Both `any_of` and `count_if` sync stream after each block. As I said above, we don't want that.

mythrocks: Ok, thanks for confirming. Makes sense.

I would still move the `any_block_failure` definition closer to its first use. I would also copy the original comment here:
```c++
// Doesn't need to be atomic
```

ttnghia: Done.

cpp/src/io/orc/reader_impl.cu
```diff
@@ -294,6 +289,12 @@ rmm::device_buffer decompress_stripe_data(
                                  decompressor.GetLog2MaxCompressionRatio(),
                                  stream);
 
+  // Value for checking whether we decompress successfully.
+  // It doesn't need to be atomic as there is no race condition: we only write `false` if needed.
```
mythrocks, line 293: :clap:

mythrocks (approved):
LGTM! It certainly reads much better now.

PR #13402, opened 2023-05-22
Handle some corner-cases in indexing with boolean masks

### LINKED ISSUE #13015: [BUG] loc and iloc indexing with boolean mask does not handle invalid length correctly

### LINKED ISSUE #13265: [BUG] DataFrame iloc indexing is incorrect for boolean mask of columns
**Describe the bug**

```python
import cudf
import pandas as pd
import numpy as np
df = pd.DataFrame(np.zeros((2, 2)))
cdf = cudf.from_pandas(df)

df.iloc[:, [True, True]]
#     0    1
# 0  0.0  0.0
# 1  0.0  0.0

cdf.iloc[:, [True, True]]
#      1
# 0  0.0
# 1  0.0
```

This happens because the indexing in pandas treats a boolean mask specially in both the row and column slot, however cudf only does so in the row slot.

There's another (related) bug in terms of the number of columns that are returned but that is separate (#13266).

**Expected behavior**

Should probably match pandas.

### LINKED ISSUE #13270: [BUG] DataFrame loc indexing is incorrect for boolean mask of columns
**Describe the bug**
This is similar to #13265 but goes down a different code path so will be fixed separately.

```python
import cudf
import pandas as pd
import numpy as np
df = pd.DataFrame(np.zeros((2, 2)), columns=["A", "B"])
cdf = cudf.from_pandas(df)

df.loc[:, [True, True]]
#     A    B
# 0  0.0  0.0
# 1  0.0  0.0

cdf.loc[:, [True, True]] # => KeyError
```

This happens because the indexing in pandas treats a boolean mask specially in both the row and column slot, however cudf only does so in the row slot.

There's another (related) bug in terms of the number of columns that are returned but that is separate (#13269).

**Expected behavior**

Should probably match pandas.

wence-:
## Description
These must be treated specially and not accidentally converted to integers before indexing.

- Closes #13015
- Closes #13265
- Closes #13270

Introduces tighter guards for loc-based indexing with Series where in some circumstances one must align the indices of the indexed object and the indexer. These now raise NotImplementedError rather than returning incorrect answers.

## Checklist


wence-:
Note, this was discussed as probably wanted for 23.06 at the last cudf meeting; I haven't yet added tests but will do so on Wednesday.

python/cudf/cudf/core/column_accessor.py
```diff
@@ -371,6 +373,18 @@ def get_labels_by_index(self, index: Any) -> tuple:
...
+            if bn != (n := len(self.names)):
+                raise IndexError(
+                    f"Boolean mask has wrong length: {bn} not {n}"
+                )
+            if isinstance(index, (pd.Series, cudf.Series)):
+                # Don't allow iloc indexing with series
+                raise NotImplementedError(
+                    "Cannot use Series object for mask iloc indexing"
+                )
+            # TODO: Doesn't handle on-device columns
```
wence-, line 386: But the old code didn't either...

python/cudf/cudf/core/column_accessor.py
```diff
@@ -371,6 +373,18 @@ def get_labels_by_index(self, index: Any) -> tuple:
             return self.names[start:stop:step]
         elif pd.api.types.is_integer(index):
             return (self.names[index],)
+        elif (bn := len(index)) > 0 and all(map(is_bool, index)):
+            if bn != (n := len(self.names)):
+                raise IndexError(
+                    f"Boolean mask has wrong length: {bn} not {n}"
+                )
+            if isinstance(index, (pd.Series, cudf.Series)):
```
shwina, line 381: Would it be any better to replace this `if` with a `try`?

```python
try:
    index = list(index)  # assuming we want something we can coerce to a list
except <something>:
    raise NotImplementedError(...)  # or TypeError perhaps?
 ```

wence-: In fact, one also can't use a Series as a boolean mask for `iloc` indexing of rows, but changing that would break too much for this PR, I think.

shwina: Gotcha

shwina (approved):

Issue #11595, opened 2022-08-25
[FEA] Handle selected functions that don't have bytecode in `Series.apply`


brandon-b-miller:
**Is your feature request related to a problem? Please describe.**
During discussion in https://github.com/rapidsai/cudf/pull/11578 we realized that users might try and pass several commonly used python functions that don't necessarily have bytecode directly to `Series.apply`. For instance, using `float` dispatches to a C-level `__init__` method and thus has no bytecode for numba to build a function out of. 

**Describe the solution you'd like**
It would be nice if this worked:
```python
cudf.Series([1,2,3]).apply(float)
# [1.0, 2.0, 3.0]
```
We should be able to use 
- int
- float
- bool

We could also add:
- Numpy dtypes
- string?

**Describe alternatives you've considered**
Users should just call `astype` but this seems like plausible usage. 

**Additional context**
Came up during https://github.com/rapidsai/cudf/pull/11578#pullrequestreview-1082926062

bdice:
This isn't really about `float`, or casting at all. It's about _any_ Python builtin or function (with C implementation) that doesn't have a `__code__` attribute.

bdice:
A similar issue arose in https://github.com/rapidsai/cudf/pull/11319#discussion_r963912420:

> Can you add a test using len directly? Or is a __code__ attribute required as noted in https://github.com/rapidsai/cudf/issues/11595?
> ```python
> run_udf_test(data, len, "int64")
> ```

> Right - we'd need some separate handling for this. Adding the test yields
> ```python
> AttributeError: 'builtin_function_or_method' object has no attribute '__defaults__'
> ```

PR #13408, opened 2023-05-22
Add `abs` function to apply

brandon-b-miller:
From a question on the GoAI slack, this PR adds the absolute value function to `Series` and `DataFrame` apply.


bdice (approved):
LGTM. @brandon-b-miller Can we please file an issue with a list of other operators we might be missing? We should check against Python's builtins, or maybe the `math` module or `numpy` operators?

python/cudf/cudf/core/udf/masked_lowering.py
```diff
@@ -333,6 +333,27 @@ def masked_scalar_cast_impl(context, builder, sig, args):
...
+    input = cgutils.create_struct_proxy(sig.args[0])(
+        context, builder, value=args[0]
+    )
+    result = cgutils.create_struct_proxy(sig.return_type)(context, builder)
+
+    result.valid = input.valid
+    with builder.if_then(input.valid):
+        result.value = context.compile_internal(
+            builder,
+            lambda x: abs(x),
```
wence-, line 347: Is it possible to just write `abs` here ($\eta$-reduction)

python/cudf/cudf/core/udf/masked_typing.py
```diff
@@ -411,6 +411,29 @@ def generic(self, args, kws):
...
+@cuda_decl_registry.register_global(abs)
+class MaskedScalarAbsoluteValue(AbstractTemplate):
+    """
+    Typing for the builtin function abs. Returns the same
+    type as input except for boolean values which are converted
+    to integer.
+
+    This follows the expected result from the builtin abs function
+    which differs from numpy - np.abs returns a bool whereas abs
+    itself performs the cast.
```
wence-, line 423: I am not sure we want this special-casing. `abs(pd.Series([0, 1], dtype=bool))` produces a series with dtype bool.

wence- (changes requested):
Minor nits

python/cudf/cudf/core/udf/masked_lowering.py
bdice: It might not be. I would defer to @brandon-b-miller here but I remember raising a similar question someplace else with Numba.

brandon-b-miller: what @bdice mentions is correct. We need to wrap the actual `abs` call in an expression that as a whole exposes its bytecode as a `__code__` attribute. Switching to `abs` results in the following:
```python
'builtin_function_or_method' object has no attribute '__code__'
```

python/cudf/cudf/core/udf/masked_typing.py
brandon-b-miller: True, but we're trying to mimic the behavior of pandas, which executes the real `abs` function in a loop over the scalars in the data:

```python
import pandas as pd
sr = pd.Series([0, 1], dtype='bool')
sr.apply(lambda x: abs(x))
```
this gives me 
```python
0    0
1    1
dtype: int64
```

brandon-b-miller:
> LGTM. @brandon-b-miller Can we please file an issue with a list of other operators we might be missing? We should check against Python's builtins, or maybe the `math` module or `numpy` operators?

raised https://github.com/rapidsai/cudf/issues/13462

python/cudf/cudf/core/udf/masked_typing.py
wence-: Oh, sure. I mean this is horrible, but so it goes.

python/cudf/cudf/tests/test_udf_masked_ops.py
```diff
@@ -802,6 +802,18 @@ def func(x):
     run_masked_udf_series(func, data, check_dtype=False)
 
 
+@pytest.mark.parametrize(
+    "data", [[1, -1, 0, cudf.NA], [1, 0, -1, cudf.NA], [True, False, cudf.NA]]
```
wence-, line 806: Please can you test "interesting" corner cases.

In particular:

1. Input series should be both signed and unsigned integer dtypes, and also at least one float type
2. Values should include `numpy.iinfo(dtype).min` and `numpy.iinfo(dtype).max` (for integer dtypes) and `ffinfo.min/max` along with `nan` and `inf` and `-inf` for floats.

wence- (changes requested):
Apologies for the tardiness here. Requesting some more edge-case tests since my hunch is that the typing rules as implemented will DTRT in some cases.

python/cudf/cudf/tests/test_udf_masked_ops.py
brandon-b-miller: I added these test cases and indeed the overflow case where `abs(np.iinfo('int64').min)` won't fit in an `int64` fails. 

To get around this I've added code that upcasts to an unsigned type of the same bit width. Let me know if you think this is an appropriate solution.

python/cudf/cudf/core/udf/masked_typing.py
```diff
@@ -411,6 +411,29 @@ def generic(self, args, kws):
...
+
+    def generic(self, args, kws):
+        if isinstance(args[0], MaskedType):
+            if isinstance(args[0].value_type, (StringView, UDFString)):
+                # reject string types
+                return
+            if isinstance(args[0], types.Boolean):
+                return nb_signature(MaskedType(types.int64), args[0])
+            else:
+                return nb_signature(args[0], args[0])
```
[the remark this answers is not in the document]
brandon-b-miller: This should be addressed in https://github.com/rapidsai/cudf/pull/13408/commits/bab9298349a5e4599ef15802cb2c5d74d7e3ae9c.

python/cudf/cudf/core/udf/masked_typing.py
```diff
@@ -411,6 +413,33 @@ def generic(self, args, kws):
...
+
+    def generic(self, args, kws):
+        if isinstance(args[0], MaskedType):
+            if isinstance(args[0].value_type, (StringView, UDFString)):
+                # reject string types
+                return
+            else:
+                return_type = self.context.resolve_function_type(
+                    self.key, (args[0].value_type,), kws
+                ).return_type
```
brandon-b-miller, line 436: This line uses numba's internals to obtain the result of typing `abs` if the inputs were scalars.

python/cudf/cudf/tests/test_udf_masked_ops.py
```diff
@@ -802,6 +802,42 @@ def func(x):
...
+        ),
+        [False, True, False, cudf.NA],
+    ],
+)
+def test_masked_udf_abs(data):
+    data = cudf.Series(data)
+    data[0] = cudf.NA
+
+    def func(x):
+        return abs(x)
```
wence-, line 836: Does this also work if I, as a _user_  were to run `run_masked_udf_series(abs, data, check_dtype=False)`,  I suppose yes.

python/cudf/cudf/core/udf/masked_typing.py
wence-: I realise this might be opening pandora's box (again) here, but I'll just point out that pandas behaves inconsistently for the `INT_MIN` case depending on where or not you spelt it `s.apply(abs)` or `s.apply(np.abs)`. This is because `np.abs` preserves dtype, and (AFAICT) invokes C-level UB when called on `INT_MIN`. In contrast, `abs` will convert to `PyObject`, perform arbitrary width integer computation and get the right answer, which pandas will massage into an unsigned type.

I am happy to return a column with an unsigned dtype though.

wence- (approved):
Thanks @brandon-b-miller , just making a note about further inconsistencies here...

python/cudf/cudf/core/udf/masked_typing.py
wence-: (I think we are OK to return an unsigned type in both cases TBH). `abs(INT_MIN)` is UB so we can "do what we like"

python/cudf/cudf/tests/test_udf_masked_ops.py
brandon-b-miller: Sadly not. This is an edge case that has come up a few times where we need to wrap builtin functions inside an object that has accessible bytecode. See more at https://github.com/rapidsai/cudf/issues/11595.

PR #13411, opened 2023-05-22
Inline Cython exception handler

vyasr:
## Description

I originally placed the exception handler into a separate C++ header file that could be included by the Cython header because I figured that reading C++ inlined in Cython would be more confusing to devs. Unfortunately, the current approach complicates the build system due to the need to ensure that the directory containing the C++ header is always in the include path, which becomes problematic depending on where the files including the exception handler are (anywhere outside of `_lib` becomes problematic). Inlining is the simplest solution to this problem.

## Checklist


python/cudf/cudf/_lib/exception_handler.pxd
bdice, line 1: Is there a way to catch _only_ libcudf-specific exceptions and then use Cython's own handling to map the rest of the STL exceptions to Python?

bdice (commented):
One question, otherwise this looks fine.

shwina (approved):

python/cudf/cudf/_lib/exception_handler.pxd
vyasr: Unfortunately not the way that it's currently set up. There's no guarantee that Cython's built in exception handler will be defined at the point when it would need to be called in our custom exception handler. Otherwise you could potentially manage it with some sort of rethrow. [There's an open Cython issue to improve this support](https://github.com/cython/cython/issues/5388).

vyasr:
@bdice I'm just going to trigger a merge based on your previous comment and my response. Happy to revisit in a follow-up if you'd like.

bdice (approved):
Approving. All looks good.

PR #11656, opened 2022-09-06
Implement Python drop_duplicates with cudf::stable_distinct.

brandon-b-miller:
Depends on https://github.com/rapidsai/cudf/pull/13392.

Closes https://github.com/rapidsai/cudf/issues/11638
Closes https://github.com/rapidsai/cudf/issues/12449
Closes https://github.com/rapidsai/cudf/pull/11230
Closes https://github.com/rapidsai/cudf/issues/5286

This PR re-implements Python's `DataFrame.drop_duplicates` / `Series.drop_duplicates` to use the `stable_distinct` algorithm.

This fixed a large number of issues with correctness (ordering the same way as pandas) and also improves performance by eliminating a sorting step.

As a consequence of changing the behavior of `drop_duplicates`, a lot of refactoring was needed. The `drop_duplicates` function was used to implement `unique()`, which cascaded into changes for several groupby functions, one-hot encoding, `np.unique` array function dispatches, and more. Those downstream functions relied on the sorting order of `drop_duplicates` and `unique`, which is _not_ promised by pandas.

PR #13415, opened 2023-05-23
Use <NA> instead of null to match pandas.

bdice:
## Description
This PR updates the behavior of one-hot encoding to use `<NA>` for null values rather than `null`. This aligns with pandas behavior.

This change is backported from #13174 since it applies to pandas 1.x as well as pandas 2.0.

## Checklist

### DIFF
--- a/python/cudf/cudf/_lib/transform.pyx
+++ b/python/cudf/cudf/_lib/transform.pyx
@@ -163,7 +163,7 @@ def one_hot_encode(Column input_column, Column categories):
         move(c_result.second),
         owner=owner,
         column_names=[
-            x if x is not None else 'null' for x in pylist_categories
+            x if x is not None else '<NA>' for x in pylist_categories
         ]
     )
     return encodings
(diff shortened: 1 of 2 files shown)


galipremsagar (approved):

bdice:
This PR can target 23.08 but we don't have forward mergers from 23.06 to 23.08 yet and this depends on recent changes on `branch-23.06`. Once `branch-23.08` is up to date, we can move the target branch.

bdice:
This changes behavior (column names are different) so I marked it as `breaking`.

@shwina @galipremsagar Do you think this should go in for 23.06 or 23.08? We have the opportunity to exactly match pandas behavior by merging this (in conjunction with the merged changes from #11656).

galipremsagar:
> This changes behavior (column names are different) so I marked it as `breaking`.
> 
> @shwina @galipremsagar Do you think this should go in for 23.06 or 23.08? We have the opportunity to exactly match pandas behavior by merging this (in conjunction with the merged changes from #11656).

Okay to target for 23.06 since this is a bug fix. But will wait on @shwina to comment too.

shwina (approved):
Going to go ahead and merge

PR #13419, opened 2023-05-23
Refactor sort_by_values to sort_values, drop indices from return values.

### LINKED ISSUE #13391: [FEA] Change `Column.sort_by_values` to `Column.sort_values`.

cuDF Python has a method `Column.sort_by_values` that returns the sorted values and their indices. Unfortunately we never use both the values and the indices, only one or the other. We also have `Column.argsort` which returns just the indices.

I propose we change `Column.sort_by_values` to `Column.sort_values` and drop the indices. This reduces the complexity of the implementation. If users need both values and indices, they can compute the `argsort` and then use it as a gather map to `take`.

bdice:
## Description
This PR closes #13391.

## Checklist


python/cudf/cudf/core/column/column.py
```diff
@@ -926,16 +926,16 @@ def get_slice_bound(self, label: ScalarLike, side: str, kind: str) -> int:
...
         ascending: bool = True,
         na_position: str = "last",
-    ) -> Tuple[ColumnBase, "cudf.core.column.NumericalColumn"]:
+    ) -> ColumnBase:
         col_inds = self.as_frame()._get_sorted_inds(
             ascending=ascending, na_position=na_position
         )
-        col_keys = self.take(col_inds)
-        return col_keys, col_inds
+        col_keys = self.take(col_inds, check_bounds=False)
```
bdice, line 937: We should try to _directly_ sort, rather than converting to a frame, getting the sorted indices, and then gathering.

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -145,6 +146,53 @@ def order_by(list columns_from_table, object ascending, str na_position):
     return Column.from_unique_ptr(move(c_result))
 
 
+@acquire_spill_lock()
+def sort(
```
bdice, line 150: I was shocked we didn't have Cython bindings for `cudf::sort` yet. I suspect there are lots of places in cudf where we need to replace an "argsort and gather" with a plain table sort. That would eliminate a layer of Python / kernel overhead.

python/cudf/cudf/core/column/column.py
```diff
@@ -1380,7 +1378,7 @@ def _return_sentinel_column():
         # values of `self`:
         order = arange(len(self))
         order = order.take(left_gather_map, check_bounds=False).argsort()
-        codes = codes.take(order)
+        codes = codes.take(order, check_bounds=False)
```
bdice, line 1381: Reviewing uses of `take` to avoid bounds-checking would be a good thing anywhere that cudf constructs the indices (and they're known to be safe) rather than taking user input.

python/cudf/cudf/_lib/sort.pyx
wence-: I imagine that there are also a bunch of cases where `argsort + gather` is actually a `sort_by_key` in disguise.

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -145,6 +146,53 @@ def order_by(list columns_from_table, object ascending, str na_position):
...
+    column_order = column_order or [True] * ncol
+    null_precedence = null_precedence or ["first"] * ncol
+    for asc, null in zip(column_order, null_precedence):
+        c_column_order.push_back(order.ASCENDING if asc else order.DESCENDING)
+        if asc ^ (null == "first"):
+            c_null_precedence.push_back(null_order.AFTER)
+        elif asc ^ (null == "last"):
+            c_null_precedence.push_back(null_order.BEFORE)
+        else:
+            raise ValueError(f"Invalid null precedence {null}")
```
wence-, line 184: This bit of code is now replicated three times in this file, is it time to factor it out? (`order_by` doesn't take an iterable of `null_precedences`, but one could use `zip(column_order, repeat(na_position))` in that case.

wence- (changes requested):
Looks good, I think we have an opportunity to avoid repetition though.

python/cudf/cudf/_lib/sort.pyx
bdice: Yes, I'm a strong +1 for refactoring this. I had considered doing so but hadn't assessed it fully. I'll work on that.

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -145,6 +146,53 @@ def order_by(list columns_from_table, object ascending, str na_position):
...
+            cpp_sort(
+                values_view,
+                c_column_order,
+                c_null_precedence,
+            )
+        )
+    return columns_from_unique_ptr(move(result))
+
+
 def segmented_sort_by_key(
```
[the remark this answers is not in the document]
bdice: Yes, that seems relevant enough here.

python/cudf/cudf/core/column/column.py
wence-: #13456

python/cudf/cudf/_lib/sort.pyx
bdice: @wence- Can we temporarily skip this refactor and do it in your work-in-progress branch?

bdice: @wence- I added commit f2eef2022edd3bdc8b22e6fb45f94d9d329b9778 which adapts your changes.

python/cudf/cudf/core/dataframe.py
```diff
@@ -7639,7 +7639,7 @@ def _get_union_of_indices(indexes):
     else:
         merged_index = cudf.core.index.GenericIndex._concat(indexes)
         merged_index = merged_index.drop_duplicates()
-        _, inds = merged_index._values.sort_by_values()
+        inds = merged_index._values.argsort()
         return merged_index.take(inds)
```
wence-, line 7643:
```suggestion
        return merged_index.take(inds, check_bounds=False)
```

wence-: Although, am I missing something here? 

Why is

```python
foo.take(foo.argsort())
```
not the same as
```
foo.sort_values()
```
?

wence-: Ah, `Index.sort_values` is just implemented as this anyway, so this is a future enhancement.

wence- (approved):
Thanks @bdice

python/cudf/cudf/core/column/column.py
wence-, line 1381:
```suggestion
        codes = codes.take(order)
```

Sorry, I didn't even check if `take` supports `check_bounds` (which it doesn't)...

`tests/test_dataframe.py::test_dataframe_init_from_series_list[None-data5-False] - TypeError: take() got an unexpected keyword argument 'check_bounds'`

python/cudf/cudf/core/dataframe.py
```diff
@@ -7639,8 +7639,8 @@ def _get_union_of_indices(indexes):
     else:
         merged_index = cudf.core.index.GenericIndex._concat(indexes)
         merged_index = merged_index.drop_duplicates()
-        _, inds = merged_index._values.sort_by_values()
-        return merged_index.take(inds)
+        inds = merged_index._values.argsort()
+        return merged_index.take(inds, check_bounds=False)
```
wence-, line 7643:
```suggestion
        return merged_index.take(inds)
```

wence- (commented):
Sorry, `take` doesn't support `check_bounds`, fixing...

PR #13423, opened 2023-05-23
Enforce Groupby.__iter__ deprecation and miscellaneous pytest fixes

galipremsagar:
## Description

This PR:

- [x] Enforces deprecation in `GroupBy.__iter__` 
- [x] Fixes miscellaneous pytest failures due to already existing differences in cudf vs pandas & nuly introduced `inferred_type` in Index.

## Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -5860,9 +5861,11 @@ def _reduce(
             )
             source = self._get_columns_by_label(numeric_cols)
             if source.empty:
-                return Series(index=self.index)
-
-        axis = source._get_axis_from_axis_arg(axis)
+                return Series(
+                    index=cudf.Index([], dtype="str")
```
mroeschke, line 5865: Curious why it should be `dtype="str"` and not `dtype=self.index.dtype`?

galipremsagar: Yes, it should be `self.columns.dtype`. I fixed it now 👍

python/cudf/cudf/core/dataframe.py
```diff
@@ -5860,9 +5861,11 @@ def _reduce(
             )
             source = self._get_columns_by_label(numeric_cols)
             if source.empty:
-                return Series(index=self.index)
-
-        axis = source._get_axis_from_axis_arg(axis)
+                return Series(
+                    index=self._data.to_pandas_index()[:0]
```
mroeschke, line 5865: The `[:0]` is to get the empty `Index` with the correct type?

galipremsagar: Yup

mroeschke (approved):

Issue #13426, opened 2023-05-24
[BUG] groupby apply on a functools.partial func fails


ayushdg:
**Describe the bug**
When using groupby apply on a function wrapped via functools partial I run into `AttributeError: 'functools.partial' object has no attribute '__code__'`

**Steps/Code to reproduce bug**
```python
import cudf
from functools import partial

df = cudf.DataFrame({"a":[1,2,3]})

def func(df):
    return df.sum()

df.groupby("a").apply(func) # works

func2 = partial(func)
df.groupby("a").apply(func2) # fails
```

**Expected behavior**
No failures as this use to work in previous nightlies/23.04

**Environment overview (please complete the following information)**
 - 23.06 nightlies
 - conda installed
 
**Environment details**
Please run and paste the output of the `cudf/print_env.sh` script here, to gather any other relevant environment details

**Additional context**
It's most likely related to the latest numba 0.57 update since it started happening very recently seen in dask-sql [tests](https://gpuci.gpuopenanalytics.com/job/dask/job/dask-sql/job/prb/job/dask-sql-prb/2794/CUDA_VER=11.5,LINUX_VER=ubuntu18.04,PYTHON_VER=3.9,RAPIDS_VER=23.06/testReport/junit/tests.unit/test_queries/test_gpu_query_q53_sql_/).

bdice:
Here's the traceback.

```
Traceback (most recent call last):
  File "/raid/bdice/compose-environments/rapids1/t.py", line 12, in <module>
    df.groupby("a").apply(func2) # fails
  File "/raid/bdice/compose-environments/rapids1/compose/etc/conda/cuda_11.8/envs/rapids/lib/python3.10/site-packages/nvtx/nvtx.py", line 101, in inner
    result = func(*args, **kwargs)
  File "/raid/bdice/compose-environments/rapids1/cudf/python/cudf/cudf/core/groupby/groupby.py", line 1354, in apply
    if (not grouped_values._has_nulls) and _can_be_jitted(
  File "/raid/bdice/compose-environments/rapids1/cudf/python/cudf/cudf/core/udf/groupby_utils.py", line 221, in _can_be_jitted
    _get_udf_return_type(dataframe_group_type, func, args)
  File "/raid/bdice/compose-environments/rapids1/compose/etc/conda/cuda_11.8/envs/rapids/lib/python3.10/site-packages/nvtx/nvtx.py", line 101, in inner
    result = func(*args, **kwargs)
[...]

PR #13429, opened 2023-05-24
Reject functions without bytecode from `_can_be_jitted` in GroupBy Apply

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/13426


bdice (commented):
Can we add a test for the case in the issue this closes? We should not allow a regression that breaks past behavior even if it falls back to the old implementation.

python/cudf/cudf/core/udf/groupby_utils.py
```diff
@@ -209,6 +209,9 @@ def _can_be_jitted(frame, func, args):
     by attempting to compile just the function to PTX using the
     target set of types
     """
+    if not hasattr(func, "__code__"):
+        # numba requires bytecode to be present to proceed
```
bdice, line 213:
```suggestion
        # Numba requires bytecode to be present to proceed.
        # See https://github.com/numba/numba/issues/4587
```

brandon-b-miller:
> Can we add a test for the case in the issue this closes? We should not allow a regression that breaks past behavior even if it falls back to the old implementation.

What do you think of the test added in https://github.com/rapidsai/cudf/pull/13429/commits/51633570a2c755a2150786a7304bf572bcee0440 ?

mroeschke (approved):

bdice (approved):

PR #13433, opened 2023-05-24
Patch numba if it is imported first to ensure minor version compatibility works.

bdice:
## Description
`cudf` conditionally sets the Numba configuration option `CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY` if it is needed for the current driver/runtime. Using numba 0.57.0, if `numba.cuda` is imported before `cudf`, this option is read too late, and the necessary imports of ptxcompiler/cubinlinker are not available in `numba.cuda.cudadrv.driver`. This PR patches Numba by adding those imports late if (and only if) MVC is needed and `numba.cuda` has already been imported.

See also: https://github.com/numba/numba/issues/8977

## Checklist


brandon-b-miller:
we should be able to test this with
```python
import subprocess
import sys

import pytest

skip = True
try:
    from ptxcompiler.patch import NO_DRIVER, safe_get_versions

    versions = safe_get_versions()
    if versions != NO_DRIVER:
        driver_version, runtime_version = versions
        if driver_version < (12, 0) and runtime_version > driver_version:
            skip = False
except ModuleNotFoundError:
    pass

test = """
import numba.cuda
import cudf
from cudf.utils._numba import _CUDFNumbaConfig, _patch_numba_mvc

from numba import config
config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY=1

_patch_numba_mvc()

@numba.cuda.jit
def test_kernel(x):
    id = numba.cuda.grid(1)
    if id < len(x):
        x[id] += 1

s = cudf.Series([1,2,3])
with _CUDFNumbaConfig():
    test_kernel.forall(len(s))(s)
"""


@pytest.mark.skipif(False, reason="MVC Not Required")
def test_numba_mvc_enabled():
    cp = subprocess.run([sys.executable, "-c", test], capture_output=True)
    assert cp.returncode == 0

```

python/cudf/cudf/tests/test_numba_import.py
```diff
@@ -0,0 +1,44 @@
...
+import pytest
+
+skip = True
+try:
+    from ptxcompiler.patch import NO_DRIVER, safe_get_versions
+
+    versions = safe_get_versions()
+    if versions != NO_DRIVER:
+        driver_version, runtime_version = versions
+        if driver_version < (12, 0) and runtime_version > driver_version:
```
gmarkall, line 14: Do we need the `runtime_version > driver_version` check, or can we test this in more scenarios by unconditionally enabling MVC and testing as long as we have a CUDA 11 runtime version? (and would it be worth doing so, or are we happy this just getting testing in an actual MVC scenario in CI?)

bdice: Fixed in c81e967.

python/cudf/cudf/tests/test_numba_import.py
```diff
@@ -0,0 +1,48 @@
...
+except ModuleNotFoundError:
+    pass
+
+TEST_NUMBA_MVC_ENABLED = """
+import numba.cuda
+import cudf
+from cudf.utils._numba import _CUDFNumbaConfig, _patch_numba_mvc
+
+from numba import config
+config.CUDA_ENABLE_MINOR_VERSION_COMPATIBILITY=1
```
brandon-b-miller, line 25: I think we decided this wasn't necessary?
```suggestion
```

gmarkall (commented):
This looks good to me (modulo Brandon's comment).

gmarkall (approved):

python/cudf/cudf/tests/test_numba_import.py
bdice: Correct. Committing this change and adding `/merge`.

bdice:
CI failed with:
```
FAILED python/dask_cudf/dask_cudf/tests/test_groupby.py::test_groupby_cumulative[null-cumsum-True] - AssertionError: Series are different
```
I'm rerunning in hopes that this is a transient failure.

Issue #13414, opened 2023-05-23
[BUG] Partitioned Writing/Reading Failure


quasiben:
We are seeing a bug with partitioned writing we _think_ was introduced in https://github.com/rapidsai/cudf/pull/13076

```python
import cudf
input_path = "./data.parquet"
output_path = "./partition_on_frame.parquet"

no_part_df = cudf.read_parquet(input_path)
print(len(no_part_df))

# Write with partition_on
## This creates invalid files
no_part_df.to_parquet(output_path,partition_cols=["output_partition_id"])

### below fails
shuffled_df = cudf.read_parquet(output_path)
print(len(shuffled_df))
```

I'm trying to get a smaller dataset together -- at the moment the smallest parquet which seems to manifest this bug is ~400MB and is available on internal lab machines.

GregoryKimball:
I've learned a few things about this failure. So far my observations are also consistent with #13076 as the blame diff.

The offending file includes a column "text" that contains strings from 100-900k length. If we slice these strings to make them shorter, the writer can succeed.

```
editing the "text" column:
df['text'] = df['text'].str.slice(0, value)

result:
pass with 320500
pass with 320750
pass with 320875
pass with 320938
pass with 320969
fail with 320970 CUDF failure at: /nfs/repo/cudf-23.06/cpp/src/io/parquet/reader_impl_helpers.cpp:188: Incorrect data source
fail with 320971 CUDF failure at: /nfs/repo/cudf-23.06/cpp/src/io/parquet/reader_impl_preprocess.cu:537: Error during decompression
fail with 320973 CUDF failure at: /nfs/repo/cudf-23.06/cpp/src/io/parquet/reader_impl_preprocess.cu:537: Error during decompression
fail with 320977 CUDF failure at: /nfs/repo/cudf-23.06/cpp/src/io/parquet/reader_impl_preprocess.cu:1468: Encountered malformed parquet page data (unexpected row count in page data)
fail with 320985 CUDF failure at: /nfs/repo/cudf-23.06/cpp/src/io/parquet/reader_impl_preprocess.cu:1468: Encountered malformed parquet page data (unexpected row count in page data)
```

The repro creates 7967 parquet files as part of the partitioning on `output_partition_id`. Several of these files throw errors when read individually:
```
reader_impl_preprocess.cu:1661: There is no page to parse
3, 0, 8, 2, 6
reader_impl_preprocess.cu:1468: Encountered malformed parquet page data (unexpected row count in page data)
[...]

PR #13438, opened 2023-05-24
Fix batch processing for parquet writer

ttnghia:
In parquet writer, the input table is divided into multiple batches (at 1GB limit), each batch is processed and flushed to sink one after another. The buffers storing data for processing each batch are reused among batches. This is to reduce peak GPU memory usage.

Unfortunately, in order to support retry mechanism, we have to have separate buffers for each batch. This is equivalent to always having one batch. The benefit of batch processing is stripped away.  In https://github.com/rapidsai/cudf/pull/13076, we expect to keep data for all batches but failed to do that, causing a bug reported in https://github.com/rapidsai/cudf/issues/13414.

This PR fixes the issue introduced in https://github.com/rapidsai/cudf/pull/13076. And since we have to strip away the benefit of batch processing, peak memory usage may go up.

Flag this as `breaking` because peak GPU memory usage may go up and cause the downstream application to crash.


ttnghia:
Verify that this python test was successful with the data file provided in https://github.com/rapidsai/cudf/issues/13414:
```
import cudf

input_path ="batch_0_part_0.parquet"
output_path = "repro.parquet"

df = cudf.read_parquet(input_path)
df.to_parquet(output_path, partition_cols=["output_partition_id"])
shuffled_df = cudf.read_parquet(output_path)
```

etseidl:
There's an existing parquet test I wrote when I fixed the column indexes to work with multiple batches, but it only checked the file metadata, not the actual contents.  Modifying that test as below detects the buffer reuse.
```
diff --git a/cpp/tests/io/parquet_test.cpp b/cpp/tests/io/parquet_test.cpp
index a7199ad56e..8b6cd98afd 100644
--- a/cpp/tests/io/parquet_test.cpp
+++ b/cpp/tests/io/parquet_test.cpp
@@ -4095,6 +4095,12 @@ TEST_F(ParquetWriterTest, LargeColumnIndex)
       EXPECT_TRUE(compare_binary(ci.max_values[0], stats.max_value, ptype, ctype) >= 0);
     }
   }
+
+  cudf::io::parquet_reader_options read_args =
+  auto result = cudf::io::read_parquet(read_args);
+
+  CUDF_TEST_EXPECT_TABLES_EQUAL(expected, result.tbl->view());
 }
 
 TEST_F(ParquetWriterTest, CheckColumnOffsetIndex)
```

ttnghia:
@etseidl proposed an alternative solution:
```diff
diff --git a/cpp/src/io/parquet/writer_impl.cu b/cpp/src/io/parquet/writer_impl.cu
index 05d42cd9e2..2f4918f8f2 100644
--- a/cpp/src/io/parquet/writer_impl.cu
+++ b/cpp/src/io/parquet/writer_impl.cu
@@ -1782,7 +1782,7 @@ auto convert_table_to_parquet_data(table_input_metadata& table_meta,
   // Initialize batches of rowgroups to encode (mainly to limit peak memory usage)
   std::vector<size_type> batch_list;
   size_type num_pages          = 0;
-  size_t max_bytes_in_batch    = 1024 * 1024 * 1024;  // 1GB - TODO: Tune this
+  size_t max_bytes_in_batch    = std::numeric_limits<int64_t>::max();
   size_t max_uncomp_bfr_size   = 0;
   size_t max_comp_bfr_size     = 0;
   size_t max_chunk_bfr_size    = 0;
```

This may be better in term of performance, since that can reduce the number of kernel calls. I need more opinions on this before adopting it.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1833,17 +1833,23 @@ auto convert_table_to_parquet_data(table_input_metadata& table_meta,
...
   uint32_t const num_stats_bfr =
     (stats_granularity != statistics_freq::STATISTICS_NONE) ? num_pages + num_chunks : 0;
-  rmm::device_buffer uncomp_bfr(max_uncomp_bfr_size, stream);
-  rmm::device_buffer comp_bfr(max_comp_bfr_size, stream);
-  rmm::device_buffer col_idx_bfr(column_index_bfr_size, stream);
+  std::vector<rmm::device_buffer> uncomp_bfr;
+  std::vector<rmm::device_buffer> comp_bfr;
+  std::vector<rmm::device_buffer> col_idx_bfr;
+  for (std::size_t i = 0; i < batch_list.size(); ++i) {
+    uncomp_bfr.emplace_back(max_uncomp_bfr_size, stream);
```
etseidl, line 1840: This can lead to a large overuse of memory if there are unbalanced batches. For instance, in testing I had a file that did one batch of 14 row groups, and a second batch of a single row group, all sized the same. For that case this will have two buffers large enough for 14 rowgroups, 28 total, instead of the 15 needed.  Could you add a note here to that effect?

ttnghia: It seems that the alternative solution (https://github.com/rapidsai/cudf/pull/13438#issuecomment-1561999150) should be better. With that, we won't have overuse memory.

etseidl:
> @etseidl proposed an alternative ~solution~:

bandaid :rofl:

vuule:
There should still be a peak memory use benefit from batching, as we can at least free the buffers for uncompressed data before we process the next batch. This is somewhat speculative as I haven't looked into this part of the implementation and verified if this is actually how the writer works.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1779,10 +1779,15 @@ auto convert_table_to_parquet_data(table_input_metadata& table_meta,
     std::fill_n(std::back_inserter(rg_to_part), num_rg_in_part[p], p);
   }
 
+  // Batch processing is no longer supported.
+  // This line disables batch processing (so batch size will no longer be limited at 1GB as before).
+  // TODO: All the relevant code will be removed in the follow-up work:
+  // https://github.com/rapidsai/cudf/issues/13440
```
vuule, line 1785: Not sure if this needs to be said, but please don't remove this code without further discussion. It could play a role in future performance improvements.

vuule (approved):

GregoryKimball:
Hello @wence-, thank you for your help studying #13414. Would you please take a look at this proposed solution?

wence-:
> Hello @wence-, thank you for your help studying #13414. Would you please take a look at this proposed solution?

I can confirm this fixes the read/write issues from @etseidl's minimal reproducer.

That is, I can successfully run:
```python
import pandas as pd
import random
import cudf
df = pd.DataFrame([(random.randbytes(16).hex(), random.randbytes(16).hex()) for _ in range(30_000_000)], columns=["a", "b"])
cdf = cudf.from_pandas(df)

cdf.to_parquet("cudf.pq")

cdf2 = cudf.read_parquet("cudf.pq")
df2 = pd.read_parquet("cudf.pq")

assert (cdf == cdf2).all().all() # => True
assert (df == df2).all().all() # => True
```

I can also round-trip the repro dataset from #13414, but like @ttnghia don't know exactly how to check for correctness.

cc: @quasiben and @VibhuJawa to check too.

GregoryKimball:
I can confirm that now we can round-trip the repro dataset from #13414. The only difference is that the `output_partition_id` roundtrips into a `category` dtype instead of the starting `int32` dtype.

quasiben:
Confirmed that repro dataframes are equal (after ignoring indexes and sorting)

quasiben (approved):

PR #13442, opened 2023-05-25
Preserve Index and grouped columns in `Groupby.nth`

galipremsagar:
In pandas-2.0 `groupby.nth` behavior has changed: https://pandas.pydata.org/docs/whatsnew/v2.0.0.html#dataframegroupby-nth-and-seriesgroupby-nth-now-behave-as-filtrations

This PR enables preserving the callers index in the end result and returns grouping columns as part of the result.

This PR fixes all 12 pytests in `python/cudf/cudf/tests/test_groupby.py::test_groupby_nth`

## Checklist

### DIFF
--- a/python/cudf/cudf/core/groupby/groupby.py
+++ b/python/cudf/cudf/core/groupby/groupby.py
@@ -821,8 +821,8 @@ class GroupBy(Serializable, Reducible, Scannable):
         col_idxs = [
             i for i, x in enumerate(self.obj._data.names) if x in grouping_cols
         ]
-        result = result.sort_index()
-        sizes = self.size().sort_index()
+        result = result
+        sizes = self.size().reindex(result.index)
 
         result = result[sizes > n]
         for idx, name, col in zip(
--- a/python/cudf/cudf/core/groupby/groupby.py
+++ b/python/cudf/cudf/core/groupby/groupby.py
@@ -800,40 +800,17 @@ class GroupBy(Serializable, Reducible, Scannable):
         Return the nth row from each group.
         """
 
-        new_df = self.obj.copy(deep=False)
-        new_df["__groupbynth_order__"] = range(0, len(self.obj))
-        gb = new_df.groupby(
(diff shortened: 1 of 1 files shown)


python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -799,10 +799,42 @@ def nth(self, n):
...
+            )
+        )
+        col_idxs = [
+            i for i, x in enumerate(self.obj._data.names) if x in grouping_cols
+        ]
+        result = result.sort_index()
         sizes = self.size().sort_index()
 
-        return result[sizes > n]
+        result = result[sizes > n]
```
mroeschke, line 827: Would it be more performant to do this filter then call `sort_index`?

galipremsagar: `sort_index` was being done to have a matching comparison in `sizes>n`, I removed the sorting of index twice and reduced it to reindexing only once.

mroeschke (approved):

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -799,10 +799,19 @@ def nth(self, n):
         """
         Return the nth row from each group.
         """
-        result = self.agg(lambda x: x.nth(n)).sort_index()
-        sizes = self.size().sort_index()
 
-        return result[sizes > n]
+        self.obj["__groupbynth_order__"] = range(0, len(self.obj))
+        result = self.obj.groupby(self.grouping.keys).agg(lambda x: x.nth(n))
```
shwina, line 804: It would be helpful if we added a comment above this line explaining why we are doing a groupby again.

shwina (approved):
Approving, with a small request to add a comment

python/cudf/cudf/core/groupby/groupby.py
galipremsagar: Done 👍

PR #13459, opened 2023-05-26
Fix cudf::repeat logic when count is zero

### LINKED ISSUE #13458: [BUG] cudf::repeat crashes with SIGFPE when count == 0
**Describe the bug**
After #13323 calling cudf::repeat with a scalar count of 0 will crash the process with SIGFPE due to a divide by zero.  The problem is an assertion was added that involves dividing by `count`, but `count` can be zero at that point which will result in a divide by zero.

Call cudf::repeat with a scalar count of zero.

**Expected behavior**
The process should not crash and instead an empty table should be returned when count == 0.

davidwendt:
## Description
Fixes logic in `cudf::repeat` to check for `count==0` before using it to check for possible out-of-bounds output size result.
Added a specific gtest for this as well.

Closes #13458 

## Checklist


bdice (approved):
I would support this fix going into 23.06. It's very low risk and fixes a crash.

vuule (approved):

jlowe (approved):
I verified this fixes the crashes we saw in our Spark integration test.  Thanks for the quick turnaround, @davidwendt!

Note that we would like to see this in 23.06 to unblock our release pipelines there.

bdice:
Oops, I made a mistake (and fixed it). I changed the target branch to 23.06 but we do need to rebase this over `branch-23.06`. Luckily it's only a single commit so this is not too tough. @davidwendt Can you rebase? cc: @GregoryKimball since this fix is needed during freeze.

hyperbolic2346 (approved):

davidwendt:
My git foo is weak so I'll need help rebasing this PR.

PR #13463, opened 2023-05-26
Remove unused max_rows_tensor parameter from subword tokenizer

davidwendt:
## Description
Removes the `max_rows_tensor` parameter is from the `nvtext::subword_tokenize` API since it is no longer required. The parameter was intended to size the temporary working memory for the internal functions. Since some general rework it was no longer used but never removed from the API.
Also updates the Python/Cython calls which had been hard-coding a default value anyway.

Reference issue #13458 found this issue.

## Checklist


python/cudf/cudf/_lib/nvtext/subword_tokenize.pyx
```diff
@@ -36,8 +36,7 @@ def subword_tokenize_inmem_hash(
     uint32_t max_sequence_length=64,
     uint32_t stride=48,
     bool do_lower=True,
-    bool do_truncate=False,
-    uint32_t max_rows_tensor=500
+    bool do_truncate=False
```
bdice, line 39: You can leave trailing commas on the last parameter. In cases like this, it has no effect on Python parsing but does reduce the diff from two lines to one.
```suggestion
    bool do_truncate=False,
```

python/cudf/cudf/_lib/nvtext/subword_tokenize.pyx
```diff
@@ -52,8 +51,7 @@ def subword_tokenize_inmem_hash(
                 max_sequence_length,
                 stride,
                 do_lower,
-                do_truncate,
-                max_rows_tensor
+                do_truncate
```
bdice, line 54:
```suggestion
                do_truncate,
```

bdice (approved):

vyasr (approved):

mroeschke (approved):

PR #13482, opened 2023-05-31
Fix chunked Parquet reader benchmark

vuule:
## Description
Currently, chunked Parquet reader benchmark creates the chunked reader object once and reuses it for all iterations.
After the first iteration the source is fully read so each subsequent iteration returns a single, empty, chunk. 

This PR fixes the use of the chunked reader object.
The creation of the object is included in the benchmark timing.

## Checklist

### DIFF
--- a/cpp/benchmarks/io/parquet/parquet_reader_input.cpp
+++ b/cpp/benchmarks/io/parquet/parquet_reader_input.cpp
@@ -140,7 +140,6 @@ void BM_parquet_read_chunks(
 
   cudf::io::parquet_reader_options read_opts =
     cudf::io::parquet_reader_options::builder(source_sink.make_source_info());
-  auto reader = cudf::io::chunked_parquet_reader(byte_limit, read_opts);
 
   auto mem_stats_logger = cudf::memory_stats_logger();
   state.set_cuda_stream(nvbench::make_cuda_stream_view(cudf::get_default_stream().value()));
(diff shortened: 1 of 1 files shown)


ttnghia (approved):

cpp/benchmarks/io/parquet/parquet_reader_input.cpp
```diff
@@ -149,8 +148,9 @@ void BM_parquet_read_chunks(
                try_drop_l3_cache();
 
                timer.start();
+               auto reader = cudf::io::chunked_parquet_reader(byte_limit, read_opts);
```
PointKernel, line 151:
```suggestion
               auto reader = cudf::io::chunked_parquet_reader(byte_limit, read_opts);
               timer.start();
```
Do we want to move this line before `timer.start()` to exclude the reader construction time?

vuule: I've thought about this and decided to include it. The reader object is not reusable, so IMO the time it takes to create it should be taken into account as part of the time needed to read a file. There's no way for the user to offset this cost.
In addition, the constructor might be doing a lot of preprocessing (e.g. to determine the chunks).
CC @nvdbaranec

PointKernel: >  There's no way for the user to offset this cost.

good point

PointKernel (approved):
LGTM

PR #13489, opened 2023-06-01
Fix valid count computation in offset_bitmask_binop kernel

### LINKED ISSUE #13479: [BUG] Intermittent failures in `groupby`  cumulative scans when keys contain nulls.
**Describe the bug**

Running

```python
from itertools import count

import cudf
import dask_cudf
import numpy as np
import rmm
from cudf.testing._utils import assert_eq

if __name__ == "__main__":
    state = np.random.get_state()
    rmm.reinitialize(pool_allocator=False)
    for i in count():
        oldstate = np.random.get_state()
        size = 10_000
        gdf_original = cudf.DataFrame(
            {
                "xx": np.random.randint(0, 5, size=size),
                "x": np.random.normal(size=size),
                "y": np.random.normal(size=size),
            },
        )

        # insert nulls into the key column at random.
        gdf_original["xx"] = gdf_original.xx.mask(
            np.random.choice([True, False], size=gdf_original.xx.shape)
        )
        pdf = gdf_original.to_pandas(nullable=False)
        gdf = gdf_original
        gdf_grouped = gdf.groupby("xx")
        cudf_result = gdf_grouped.cumsum()

        # Although we don't look at the data, this seems pretty
        # crucial to provoking the issue.
        # Notice this never touches the gdf_original data! And the
        # computation is _done_ by the time we're here, so one
        # suspicion is that there is garbage left lying around for the
        # next iteration.
        ddf = dask_cudf.from_cudf(cudf.from_pandas(pdf), npartitions=5).persist()
        ddf_grouped = ddf.groupby("xx")
        dask_cudf_result = ddf_grouped.cumsum().compute(scheduler="sync")

        pandas_result = pdf.groupby("xx").cumsum()
        print(i)
        # This occasionally fails
        assert_eq(cudf_result, pandas_result, check_dtype=False)

        # This was for checking when we removed the reindex call in groupby._mimic_pandas_order

        # mask = ~pdf.xx.isna()
        # pandas_result_no_nulls = pandas_result.loc[mask]
        # assert_eq(cudf_result.sort_index(), pandas_result_no_nulls, check_dtype=False)
```

with a recent enough cudf nightly sometimes produces assertion errors when checking correctness. The way this exhibits is that a few entries in the grouped dataframe columns are marked as NULL when they should not be. Post-mortem debugging, if one re-executes the offending bad code it tends to produce the correct result. One normally has to run a few times, interrupting the script to see the failure.

This has sometimes been causing the nightly actions to fail, the first is https://github.com/rapidsai/cudf/actions/runs/5065561229/jobs/9094279670 which is the first nightly that contained #13372.

Some existing investigation with @shwina provides the following information:

1. #13389 fixed this code so that it would run at all, and introduces a `dataframe.reindex` call inside `_mimic_pandas_order`. If we remove the `reindex` call (applying this patch
    ```diff
    diff --git a/python/cudf/cudf/core/groupby/groupby.py b/python/cudf/cudf/core/groupby/groupby.py
    index b7faed1dfc..e1a84897f4 100644
    --- a/python/cudf/cudf/core/groupby/groupby.py
    +++ b/python/cudf/cudf/core/groupby/groupby.py
    @@ -13,6 +13,7 @@ import numpy as np
     import pandas as pd

     import cudf
    +import cudf._lib as libcudf
     from cudf._lib import groupby as libgroupby
     from cudf._lib.null_mask import bitmask_or
     from cudf._lib.reshape import interleave_columns
    @@ -2290,12 +2291,12 @@ class GroupBy(Serializable, Reducible, Scannable):
                 ri = cudf.RangeIndex(0, len(self.obj))
                 result.index = cudf.Index(ordering)
                 # This reorders and expands
    -            result = result.reindex(ri)
    +            # result = result.reindex(ri)
             else:
                 # Just reorder according to the groupings
                 result = result.take(ordering.argsort())
    -        # Now produce the actual index we first thought of
    -        result.index = self.obj.index
    +            # Now produce the actual index we first thought of
    +            result.index = self.obj.index
             return result
    ```
   And uncommenting the alternate error checking code in the bug script, we _do not_ see failures.

2. `reindex` goes through `join` and hence `gather`. However, we also tried reimplementing the reordering using `scatter`, like so:
    ```diff
    diff --git a/python/cudf/cudf/core/groupby/groupby.py b/python/cudf/cudf/core/groupby/groupby.py
    index b7faed1dfc..64d9ef2a8f 100644
    --- a/python/cudf/cudf/core/groupby/groupby.py
    +++ b/python/cudf/cudf/core/groupby/groupby.py
    @@ -13,6 +13,7 @@ import numpy as np
     import pandas as pd

     import cudf
    +import cudf._lib as libcudf
     from cudf._lib import groupby as libgroupby
     from cudf._lib.null_mask import bitmask_or
     from cudf._lib.reshape import interleave_columns
    @@ -2287,10 +2288,18 @@ class GroupBy(Serializable, Reducible, Scannable):
                 # Scan aggregations with null/nan keys put nulls in the
                 # corresponding output rows in pandas, to do that here
                 # expand the result by reindexing.
    -            ri = cudf.RangeIndex(0, len(self.obj))
    -            result.index = cudf.Index(ordering)
    -            # This reorders and expands
    -            result = result.reindex(ri)
    +            null_result_columns = [
    +                cudf.core.column.column_empty_like(
    +                    c, masked=True, newsize=len(self.obj)
    +                )
    +                for c in result._data.columns
    +            ]
    +            new_result_columns = libcudf.copying.scatter(
    +                [*result._data.columns], ordering, null_result_columns
    +            )
    +            result = result._from_columns_like_self(
    +                new_result_columns, column_names=result._column_names
    +            )
             else:
                 # Just reorder according to the groupings
                 result = result.take(ordering.argsort())
    ```

    and still observe the bug.

davidwendt:
## Description
Fixes the valid count calculation in the `offset_bitmask_binop` kernel when the mask may contain slack bits. The slack bits must be accounted for in the final bitmask word to correctly compute the valid count.

Closes #13479 

## Checklist

### DIFF
--- a/cpp/include/cudf/detail/null_mask.cuh
+++ b/cpp/include/cudf/detail/null_mask.cuh
@@ -66,7 +66,6 @@ __global__ void offset_bitmask_binop(Binop op,
                                      size_type source_size_bits,
                                      size_type* count_ptr)
 {
-  // constexpr auto const word_size{detail::size_in_bits<bitmask_type>()};
   auto const tid = threadIdx.x + blockIdx.x * blockDim.x;
 
   auto const last_bit_index  = source_size_bits - 1;
--- a/cpp/include/cudf/detail/null_mask.cuh
+++ b/cpp/include/cudf/detail/null_mask.cuh
@@ -91,7 +91,10 @@ __global__ void offset_bitmask_binop(Binop op,
     if (destination_word_index == last_word_index) {
       // mask out any bits not part of this word
       auto const num_bits_in_last_word = intra_word_index(last_bit_index);
-      destination_word &= set_least_significant_bits(num_bits_in_last_word + 1);
+      if (num_bits_in_last_word <
+          static_cast<size_type>(detail::size_in_bits<bitmask_type>() - 1)) {
+        destination_word &= set_least_significant_bits(num_bits_in_last_word + 1);
+      }
     }
 
     destination[destination_word_index] = destination_word;
--- a/cpp/include/cudf/detail/null_mask.cuh
+++ b/cpp/include/cudf/detail/null_mask.cuh
@@ -69,6 +69,10 @@ __global__ void offset_bitmask_binop(Binop op,
   constexpr auto const word_size{detail::size_in_bits<bitmask_type>()};
   auto const tid = threadIdx.x + blockIdx.x * blockDim.x;
 
+  size_type const last_bit_index  = source_size_bits - 1;
+  size_type const num_slack_bits  = word_size - (last_bit_index % word_size) - 1;
+  size_type const last_word_index = cudf::word_index(last_bit_index);
+
   size_type thread_count = 0;
 
   for (size_type destination_word_index = tid; destination_word_index < destination.size();
@@ -88,15 +92,8 @@ __global__ void offset_bitmask_binop(Binop op,
 
     destination[destination_word_index] = destination_word;
     thread_count += __popc(destination_word);
-  }
-
-  // Subtract any slack bits from the last word
-  if (tid == 0) {
-    size_type const last_bit_index = source_size_bits - 1;
-    size_type const num_slack_bits = word_size - (last_bit_index % word_size) - 1;
-    if (num_slack_bits > 0) {
-      size_type const word_index = cudf::word_index(last_bit_index);
-      thread_count -= __popc(destination[word_index] & set_most_significant_bits(num_slack_bits));
+    if (destination_word_index == last_word_index) {
+      thread_count -= __popc(destination_word & set_most_significant_bits(num_slack_bits));
     }
   }
 
--- a/cpp/include/cudf/detail/null_mask.cuh
+++ b/cpp/include/cudf/detail/null_mask.cuh
@@ -69,9 +69,8 @@ __global__ void offset_bitmask_binop(Binop op,
   constexpr auto const word_size{detail::size_in_bits<bitmask_type>()};
   auto const tid = threadIdx.x + blockIdx.x * blockDim.x;
 
-  size_type const last_bit_index  = source_size_bits - 1;
-  size_type const num_slack_bits  = word_size - (last_bit_index % word_size) - 1;
-  size_type const last_word_index = cudf::word_index(last_bit_index);
+  auto const last_bit_index  = source_size_bits - 1;
+  auto const last_word_index = cudf::word_index(last_bit_index);
 
   size_type thread_count = 0;
 
@@ -90,11 +89,14 @@ __global__ void offset_bitmask_binop(Binop op,
                                                          source_begin_bits[i] + source_size_bits));
     }
 
+    if (destination_word_index == last_word_index) {
+      // mask out any bits not part of this word
+      auto const num_bits_in_last_word = word_size - intra_word_index(last_bit_index) - 1;
+      destination_word &= set_least_significant_bits(word_size - num_bits_in_last_word);
+    }
+
     destination[destination_word_index] = destination_word;
     thread_count += __popc(destination_word);
-    if (destination_word_index == last_word_index) {
-      thread_count -= __popc(destination_word & set_most_significant_bits(num_slack_bits));
-    }
   }
 
   using BlockReduce = cub::BlockReduce<size_type, block_size>;
--- a/cpp/include/cudf/detail/null_mask.cuh
+++ b/cpp/include/cudf/detail/null_mask.cuh
@@ -66,7 +66,7 @@ __global__ void offset_bitmask_binop(Binop op,
                                      size_type source_size_bits,
                                      size_type* count_ptr)
 {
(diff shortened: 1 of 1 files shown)


cpp/include/cudf/detail/null_mask.cuh
```diff
@@ -69,6 +69,10 @@ __global__ void offset_bitmask_binop(Binop op,
   constexpr auto const word_size{detail::size_in_bits<bitmask_type>()};
   auto const tid = threadIdx.x + blockIdx.x * blockDim.x;
 
+  size_type const last_bit_index  = source_size_bits - 1;
+  size_type const num_slack_bits  = word_size - (last_bit_index % word_size) - 1;
+  size_type const last_word_index = cudf::word_index(last_bit_index);
```
vyasr, line 74: FWIW I think these can all be `auto`. 
```suggestion
  auto const last_bit_index  = source_size_bits - 1;
  auto const num_slack_bits  = word_size - (last_bit_index % word_size) - 1;
  auto const last_word_index = cudf::word_index(last_bit_index);
```

vyasr (approved):
Thanks for finding this @davidwendt! Also great work getting repros @wence- @shwina.

Approving with one minor suggestion for improvement.

cpp/include/cudf/detail/null_mask.cuh
```diff
@@ -88,15 +92,8 @@ __global__ void offset_bitmask_binop(Binop op,
...
-  }
-
-  // Subtract any slack bits from the last word
-  if (tid == 0) {
-    size_type const last_bit_index = source_size_bits - 1;
-    size_type const num_slack_bits = word_size - (last_bit_index % word_size) - 1;
-    if (num_slack_bits > 0) {
-      size_type const word_index = cudf::word_index(last_bit_index);
-      thread_count -= __popc(destination[word_index] & set_most_significant_bits(num_slack_bits));
+    if (destination_word_index == last_word_index) {
```
bdice, line 95: Would it be simpler / safer to count _only_ the bits we know to be non-slack, rather than counting all bits and subtracting the slack bits?
```
thread_count += __popc(destination_word_index == last_word_index
                           ? destination_word & set_least_significant_bits(
                                                    word_size - num_slack_bits)
                           : destination_word);
```

(I would refactor `num_slack_bits` to be `num_bits_in_last_word` to eliminate the subtraction)

bdice: It would also look a bit more like the implementation here: https://github.com/rapidsai/cudf/blob/ebc68df255b9f304ee19b969645a10d5ab9f8bea/cpp/src/bitmask/null_mask.cu#L119-L121

cpp/include/cudf/detail/null_mask.cuh
```diff
@@ -69,6 +69,10 @@ __global__ void offset_bitmask_binop(Binop op,
   constexpr auto const word_size{detail::size_in_bits<bitmask_type>()};
   auto const tid = threadIdx.x + blockIdx.x * blockDim.x;
 
+  size_type const last_bit_index  = source_size_bits - 1;
+  size_type const num_slack_bits  = word_size - (last_bit_index % word_size) - 1;
```
bdice, line 73: This can be computed with `intra_word_index`, I think. Let's reuse functions and avoid doing the modular math by hand.

cpp/include/cudf/detail/null_mask.cuh
bdice: To clarify my pair of suggestions: maybe we should write this above:
```cpp
auto const num_bits_in_last_word = cudf::detail::intra_word_index(last_bit_index);
```
and then here, write:
```cpp
thread_count += __popc(destination_word_index == last_word_index ? destination_word & set_least_significant_bits(num_bits_in_last_word) : destination_word);
```

davidwendt: Opted to mask the word itself to keep it to one if-statement on no ternary.

bdice: This change is even better. Not only does it fix the count, it also zeros the slack bits in the result. 👍

cpp/include/cudf/detail/null_mask.cuh
```diff
@@ -66,9 +66,12 @@ __global__ void offset_bitmask_binop(Binop op,
                                      size_type source_size_bits,
                                      size_type* count_ptr)
 {
-  constexpr auto const word_size{detail::size_in_bits<bitmask_type>()};
+  // constexpr auto const word_size{detail::size_in_bits<bitmask_type>()};
```
bdice, line 69: This can be removed.
```suggestion
```

bdice (approved):
New changes look great. One commented line can be deleted. Otherwise let's get this fix merged! Very nice work @davidwendt. 💯

wence-:
Just to note this latest version is still good on our reproducers for me.

wence- (approved):

PR #13472, opened 2023-05-30
Parse column chunk metadata statistics in parquet reader

karthikeyann:
## Description
Parses column chunk metadata statistics in parquet reader.
This change parses the statistics blob as `Statistics` struct.
update writer to use Statistics

Prerequite for #13348

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13493, opened 2023-06-01
Use east const in src files 

karthikeyann:
## Description
Use east const in _cpp/src/_ files
used clang-format `QualifierAlignment: Right` to do the clean up.

since https://clang.llvm.org/docs/ClangFormatStyleOptions.html#qualifieralignment has warning that 
`Setting QualifierAlignment to something other than Leave, COULD lead to incorrect code formatting due to incorrect decisions made due to clang-formats lack of complete semantic information. As such extra care should be taken to review code changes made by the use of this option.`
So it's not added to .clang-format

Split to multiple PRs for easy reviewing #13491, #13492, #13493, #13494

## Checklist


davidwendt:
Please, no.
Was there a discussion on this?

karthikeyann:
No. we haven't.
I started this after a review comment about east const in https://github.com/rapidsai/cudf/pull/13472#discussion_r1211106637
During review, I expected to get opinion from reviewers on what's preferred vs what's not.

ttnghia (approved):

vuule:
I'm pro-east const and thought that was the convention we used for the new code in libcudf.
But, given @davidwendt 's reaction, we should probably discuss this in the next standup.

vyasr:
I would assume David's reaction was more about the batch replacement PR rather than opposition to east const itself, but agree that we should discuss before proceeding.

cpp/src/io/orc/stripe_data.cu
```diff
@@ -1108,7 +1108,7 @@ __global__ void __launch_bounds__(block_size)
   } temp_storage;
 
   orcdec_state_s* const s = &state_g;
-  const bool is_nulldec   = (blockIdx.y >= num_stripes);
+  bool const is_nulldec   = (blockIdx.y >= num_stripes);
   const uint32_t column   = blockIdx.x;
   const uint32_t stripe   = (is_nulldec) ? blockIdx.y - num_stripes : blockIdx.y;
   const uint32_t chunk_id = stripe * num_columns + column;
```
vuule, line 1114: I wonder why it skipped these

cpp/src/io/orc/stripe_data.cu
```diff
@@ -1403,7 +1403,7 @@ __global__ void __launch_bounds__(block_size)
   }
   __syncthreads();
   // Struct doesn't have any data in itself, so skip
-  const bool is_valid       = s->chunk.type_kind != STRUCT;
+  bool const is_valid       = s->chunk.type_kind != STRUCT;
   const size_t max_num_rows = s->chunk.column_num_rows;
```
vuule, line 1407: here as well. Are there known cases where clang-format does not apply the const style?

vuule (commented):
A very exciting PR to review :D
@karthikeyann have you noticed any errors in the tools changes?

cpp/src/io/orc/stripe_data.cu
karthikeyann: Looks like, this is what clang-format warning was about.
There are other places where similar west const got replaced. So, this is a tool issue.

cpp/src/io/comp/debrotli.cu
```diff
@@ -422,7 +422,7 @@ static __device__ void ext_heap_free(void* ptr,
                                      uint32_t ext_heap_size)
 {
   uint32_t len              = (bytes + 0xf) & ~0xf;
-  volatile auto* heap_ptr   = (volatile uint32_t*)ext_heap_base;
+  auto volatile* heap_ptr   = (uint32_t volatile*)ext_heap_base;
```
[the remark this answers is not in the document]
karthikeyann: It's for specifiers and qualifiers (e.g. const/volatile).
- storage-class specifiers: [typedef](https://en.cppreference.com/w/c/language/typedef), [auto, register, static, extern, _Thread_local](https://en.cppreference.com/w/c/language/storage_duration)
- type qualifiers: [const](https://en.cppreference.com/w/c/language/const), [volatile](https://en.cppreference.com/w/c/language/volatile), [restrict](https://en.cppreference.com/w/c/language/restrict), [_Atomic](https://en.cppreference.com/w/c/language/atomic)
- alignment specifiers: [_Alignas](https://en.cppreference.com/w/c/language/_Alignas)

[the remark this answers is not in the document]
karthikeyann: removed the volatile changes.

vuule (approved):

Issue #12736, opened 2023-02-08
[FEA] Create a context manager for `cudf.set_option`


galipremsagar:
**Is your feature request related to a problem? Please describe.**
A context manager would be nice to have for options in `cudf`, instead of setting and unsetting options using `cudf.set_option` twice.

without context manager:
```python
cudf.set_option("abc", True)
df.add(df1)
cudf.set_option("abc", False)
```

With context manager:
```python
with cudf.option_context("abc", True):
    df.add(df1)
```

bdice:
For additional context, this feature already exists in pandas. https://pandas.pydata.org/docs/reference/api/pandas.option_context.html

PR #13505, opened 2023-06-02
Enable writing column names with mixed dtype in parquet writer when `mode.pandas_compatible=True`

galipremsagar:
## Description
This PR enables writing a dataframe that has column names that are of mixed types to a parquet file when pandas compatibility mode is enabled(`mode.pandas_compatible=True`).

## Checklist

### DIFF
--- a/python/cudf/cudf/_lib/parquet.pyx
+++ b/python/cudf/cudf/_lib/parquet.pyx
@@ -361,9 +361,12 @@ def write_parquet(
 
     for i, name in enumerate(table._column_names, num_index_cols_meta):
         if not isinstance(name, str):
-            raise ValueError("parquet must have string column names")
-
-        tbl_meta.get().column_metadata[i].set_name(name.encode())
+            if cudf.get_option("mode.pandas_compatible"):
+                tbl_meta.get().column_metadata[i].set_name(str(name).encode())
+            else:
+                raise ValueError("parquet must have string column names")
+        else:
+            tbl_meta.get().column_metadata[i].set_name(name.encode())
         _set_col_metadata(
             table[name]._column,
             tbl_meta.get().column_metadata[i],
--- a/python/cudf/cudf/tests/test_parquet.py
+++ b/python/cudf/cudf/tests/test_parquet.py
@@ -30,7 +30,6 @@ from cudf.testing import dataset_generator as dg
 from cudf.testing._utils import (
     TIMEDELTA_TYPES,
     assert_eq,
-    assert_exceptions_equal,
     set_random_null_mask_inplace,
 )
 
@@ -2523,16 +2522,41 @@ def test_parquet_writer_decimal(decimal_type, data):
     assert_eq(gdf.to_pandas(nullable=True), got)
 
 
+# TODO: Make use of set_option context manager
+# once https://github.com/rapidsai/cudf/issues/12736
+# is resolved.
+@contextmanager
+def pandas_compatible(on):
+    original_compat_setting = cudf.get_option("mode.pandas_compatible")
+    cudf.set_option("mode.pandas_compatible", on)
+    yield
+    cudf.set_option("mode.pandas_compatible", original_compat_setting)
+
+
 def test_parquet_writer_column_validation():
-    df = cudf.DataFrame({1: [1, 2, 3], "1": ["a", "b", "c"]})
+    df = cudf.DataFrame({1: [1, 2, 3], "a": ["a", "b", "c"]})
     pdf = df.to_pandas()
 
-    assert_exceptions_equal(
-        lfunc=df.to_parquet,
-        rfunc=pdf.to_parquet,
-        lfunc_args_and_kwargs=(["cudf.parquet"],),
-        rfunc_args_and_kwargs=(["pandas.parquet"],),
-    )
+    with pandas_compatible(on=True):
+        with pytest.warns(UserWarning):
+            df.to_parquet("cudf.parquet")
+
+    if PANDAS_GE_200:
+        with pytest.warns(UserWarning):
+            pdf.to_parquet("pandas.parquet")
+
+        assert_eq(
+            pd.read_parquet("cudf.parquet"),
+            cudf.read_parquet("pandas.parquet"),
+        )
+        assert_eq(
+            cudf.read_parquet("cudf.parquet"),
+            pd.read_parquet("pandas.parquet"),
+        )
+
+    with pandas_compatible(on=False):
+        with pytest.raises(ValueError):
+            df.to_parquet("cudf.parquet")
 
 
 def test_parquet_writer_nulls_pandas_read(tmpdir, pdf):
--- a/python/cudf/cudf/_lib/utils.pyx
+++ b/python/cudf/cudf/_lib/utils.pyx
@@ -174,7 +174,7 @@ cpdef generate_pandas_metadata(table, index):
             for col in table._columns
         ],
         df=table,
-        column_names=col_names,
+        column_names=map(str, col_names),
         index_levels=index_levels,
         index_descriptors=index_descriptors,
         preserve_index=index,


python/cudf/cudf/_lib/parquet.pyx
```diff
@@ -361,9 +361,12 @@ def write_parquet(
 
     for i, name in enumerate(table._column_names, num_index_cols_meta):
         if not isinstance(name, str):
-            raise ValueError("parquet must have string column names")
-
-        tbl_meta.get().column_metadata[i].set_name(name.encode())
+            if cudf.get_option("mode.pandas_compatible"):
+                tbl_meta.get().column_metadata[i].set_name(str(name).encode())
+            else:
+                raise ValueError("parquet must have string column names")
```
bdice, line 367: I think we should have a very strong reason to differ from pandas behavior here. Do we raise an error here as a form of "user protection" to prevent unwanted outcomes from non-string column names? I would favor matching pandas and always casting to a string if that's what is expected by pandas.

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -2523,16 +2522,41 @@ def test_parquet_writer_decimal(decimal_type, data):
     assert_eq(gdf.to_pandas(nullable=True), got)
 
 
+# TODO: Make use of set_option context manager
+# once https://github.com/rapidsai/cudf/issues/12736
```
bdice, line 2526: We should prioritize https://github.com/rapidsai/cudf/issues/12736 as a requirement for pandas 2.0 support rather than take on technical debt here. It's not a hard feature to implement, from what I can tell. You already implemented it in this context manager, I think. 😄 Just might need some tweaks.

python/cudf/cudf/_lib/parquet.pyx
bdice: Or is this behavior that should be gated by a pandas version check?

bdice (changes requested):

python/cudf/cudf/tests/test_parquet.py
galipremsagar: Implemented https://github.com/rapidsai/cudf/pull/13511 to address it in `branch-23.08` itself since it is also a pandas-1.x feature.

python/cudf/cudf/_lib/parquet.pyx
galipremsagar: @vuule and rest of the IO team thought it'd be best to not allow this behavior and have this behavior of converting the column names to string only in pandas compatibility mode. I agree we need to generally match pandas by default but this can become quite erroneous for us very quickly because cases when trying to write this type of dataframe even pandas exhibit un-defined behavior:
```python
In [10]: df = cudf.DataFrame({1: [1, 2, 3], "1": ["a", "b", "c"]})

In [11]: df
Out[11]: 
   1  1
0  1  a
1  2  b
2  3  c
```

galipremsagar: > Or is this behavior that should be gated by a pandas version check?

Not quite in favor of this because it's not quite a compelling one that we have to behave differently because of a different pandas version installed.

python/cudf/cudf/tests/test_parquet.py
galipremsagar: Updated the code to utilize `cudf.option_context`, hence resolving this comment.

python/cudf/cudf/_lib/parquet.pyx
vuule: > Do we raise an error here as a form of "user protection" to prevent unwanted outcomes from non-string column names?

Yes. As @galipremsagar mentioned, casting all column names to strings creates an opportunity for errors. Alternatively, we could check if casting is safe, but this check would not be trivially cheap.
We are far from 100% Pandas compatible in cuIO; there are options that would completely kill performance and some that don't really translate to GPU. The fact that there's an option to be compatible here sounds great to me.

I'll gladly discuss this further on Monday.

bdice: Got it. This is well justified, I just didn’t know that context. We should make sure these kinds of differences are documented clearly. It relates loosely to the section on column names, so maybe describe it there or in the API docs for affected functions if it’s not already mentioned: https://docs.rapids.ai/api/cudf/stable/user_guide/pandas-comparison/#column-names

galipremsagar: Thanks @bdice & @vuule ! I have documented this difference in behavior and how to work-around it here: https://github.com/rapidsai/cudf/pull/13505/commits/d9fc2f51b4123e068af4391b0659d8655d686b58

docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -139,6 +139,27 @@ module, which allow you to compare values up to a desired precision.
 Unlike Pandas, cuDF does not support duplicate column names.
 It is best to use unique strings for column names.
 
+## Writing dataframe to parquet with mixed column names
+
+When there is a dataframe with mixed column names, pandas type-casts each
+column name to `string` before writing to parquet file. `cudf` raises an
```
bdice, line 145:
```suggestion
column name to `str` before writing to a Parquet file. `cudf` raises an
```

docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -139,6 +139,27 @@ module, which allow you to compare values up to a desired precision.
...
+error by default if this is attempted. However, to achieve similar behavior
+as pandas you can enable pandas compatibility mode option, which will
```
bdice, line 147:
```suggestion
as pandas you can enable the `mode.pandas_compatible` option, which will
```

docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -139,6 +139,27 @@ module, which allow you to compare values up to a desired precision.
...
+>>> import cudf
+>>> df = cudf.DataFrame({1: [1, 2, 3], "1": ["a", "b", "c"]})
+>>> df.to_parquet("df.parquet")
+
+Traceback (most recent call last):
+ValueError: parquet must have string column names
+>>> cudf.set_option("mode.pandas_compatible", True)
+>>> df.to_parquet("df.parquet")
+
+UserWarning: The DataFrame has column names of mixed type. They will be converted to strings and not roundtrip correctly.
```
bdice, line 160: “Round trip” is a little bit of an informal term, I think? It may not be needed here.
```suggestion
UserWarning: The DataFrame has column names of non-string type. They will be converted to strings on write.
```

docs/cudf/source/user_guide/pandas-comparison.md
bdice: Please check that code snippets are using the right number of backticks. I think one backtick just italicizes the text…

docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -139,6 +139,27 @@ module, which allow you to compare values up to a desired precision.
...
+
+When there is a DataFrame with non-string column names, pandas casts each
+column name to `str` before writing to a Parquet file. `cudf` raises an
+error by default if this is attempted. However, to achieve similar behavior
+as pandas you can enable the `mode.pandas_compatible` option, which will
+enable `cudf` to cast the column names to `str` just like pandas.
+
+```python
+>>> import cudf
+>>> df = cudf.DataFrame({1: [1, 2, 3], "1": ["a", "b", "c"]})
```
galipremsagar, line 152:
```suggestion
>>> df = cudf.DataFrame({1: [1, 2, 3], 2: ["a", "b", "c"]})
```

docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -139,6 +139,27 @@ module, which allow you to compare values up to a desired precision.
...
+>>> df.to_parquet("df.parquet")
+
+Traceback (most recent call last):
+ValueError: parquet must have string column names
```
galipremsagar, line 156:
```suggestion
ValueError: Writing a Parquet file requires string column names
```

docs/cudf/source/user_guide/pandas-comparison.md
```diff
@@ -139,6 +139,27 @@ module, which allow you to compare values up to a desired precision.
...
+as pandas you can enable pandas compatibility mode option, which will
+enable `cudf` to type-cast the column names to `string` just like pandas.
+
+```python
+>>> import cudf
+>>> df = cudf.DataFrame({1: [1, 2, 3], "1": ["a", "b", "c"]})
+>>> df.to_parquet("df.parquet")
+
+Traceback (most recent call last):
+ValueError: parquet must have string column names
```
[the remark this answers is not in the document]
galipremsagar: Done 👍

docs/cudf/source/user_guide/pandas-comparison.md
galipremsagar: Verified that one backtick is bringing the code font from the deployed docs: https://docs.rapids.ai/api/cudf/nightly/user_guide/pandas-comparison/#no-true-object-data-type

bdice: Of course. This is Markdown and not RST. My fault.

bdice (approved):

PR #13507, opened 2023-06-02
Change build.sh to use pip install instead of setup.py

vyasr:
## Description

`python setup.py` has been deprecated for years and will eventually be removed. It has numerous issues, most significant of which is that it is incompatible with many modern build systems that support only PEP 517-based builds.

This is a breaking change for build.sh users who do not install. Modern builders no longer support running entirely in place. They do support editable installs, however, so `pip install -e` remains an option for which support could be added later.

## Checklist

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -32,7 +32,7 @@ HELP="$0 [clean] [libcudf] [cudf] [cudfjar] [dask_cudf] [benchmarks] [tests] [li
    custreamz                     - build the custreamz Python package
    -v                            - verbose build mode
    -g                            - build for debug
-   -n                            - no install step
+   -n                            - no install step (does not affect Python)
    --allgpuarch                  - build for all supported GPU architectures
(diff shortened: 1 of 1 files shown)


vyasr:
Waiting to hear a bit more feedback from build.sh users before proceeding here.

davidwendt:
I use the `python setup.py` regularly locally and my concern is how will `-DFIND_CUDF_CPP=ON` be supported to use an existing built/installed libcudf.

vyasr:
Answered this question offline, but porting here for reviewers:

All CMake arguments may be passed using environment variables instead of directly in the command line invocation. scikit-build supports either `SKBUILD_CONFIGURE_OPTIONS` or `CMAKE_ARGS`, so for instance
```
SKBUILD_CONFIGURE_OPTIONS="-FIND_CUDF_CPP=ON" pip install .`
```
will work just as well. That is already what we do for building wheels.

bdice (approved):

PR #13511, opened 2023-06-05
Implement `cudf.option_context`

### LINKED ISSUE #12736: [FEA] Create a context manager for `cudf.set_option`

galipremsagar:
## Description
Resolves #12736 
This PR implements `cudf.option_context` that is similar to `pd.option_context`: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.option_context.html

## Checklist


python/cudf/cudf/options.py
```diff
@@ -114,6 +115,40 @@ def set_option(name: str, val: Any):
...
+class option_context(ContextDecorator):
+    """
+    Context manager to temporarily set options in the `with` statement context.
+
+    You need to invoke as ``option_context(pat, val, [(pat, val), ...])``.
+
+    Examples
+    --------
+    >>> from cudf import option_context
+    >>> with option_context('display.max_rows', 10, 'display.max_columns', 5):
```
bdice, line 127: I tried to read the API docs and answer this question but I couldn't. Does a cudf option for `display.max_rows` exist? Do the options propagate from cudf to pandas somehow? I know pandas' options for `display.max_rows` will affect cudf, but wasn't sure about the opposite. If this doesn't exist, we should change the docstring example to something that exists.

Also we need to explicitly list all cudf options someplace in our docs. I didn't find it when I looked. Does such a list exist?

python/cudf/cudf/options.py
```diff
@@ -114,6 +115,40 @@ def set_option(name: str, val: Any):
...
+    ...     pass
+    """
+
+    def __init__(self, *args) -> None:
+        if len(args) % 2 != 0 or len(args) < 2:
+            raise ValueError(
```
bdice, line 133: Can we add tests of `cudf.option_context` itself? e.g. set options, get options, ensure that the "undo" works, that errors are raised properly, etc.

bdice (changes requested):
Nice. Two requests, otherwise LGTM.

python/cudf/cudf/options.py
```diff
@@ -114,6 +115,40 @@ def set_option(name: str, val: Any):
...
+
+    Examples
+    --------
+    >>> from cudf import option_context
+    >>> with option_context('display.max_rows', 10, 'display.max_columns', 5):
+    ...     pass
+    """
+
+    def __init__(self, *args) -> None:
+        if len(args) % 2 != 0 or len(args) < 2:
```
wence-, line 132:
```suggestion
    def __init__(self, *args) -> None:
        if len(args) % 2 != 0:
```
Perhaps.

Rationale: this is a silly API because as a caller you have to special-case an empty options list if you want to use the context manager. That is, suppose you have some high-level code that accepts arbitrary options (including none), and you want them set when you are running some cudf/pandas code. You must do:

```python
options = ...
if len(options):
    manager = cudf.options_context(*options)
else:
   manager = contextmanager.nullcontext()
with manager:
    run_code
```

Rather than the much more natural.

```
with cudf.options_context(*options):
    ...
```

python/cudf/cudf/options.py
```diff
@@ -114,6 +115,40 @@ def set_option(name: str, val: Any):
...
+            raise ValueError(
+                "Need to invoke as option_context(pat, val, "
+                "[(pat, val), ...])."
+            )
+
+        self.ops = list(zip(args[::2], args[1::2]))
```
wence-, line 138:
```suggestion
        self.ops = tuple(zip(args[::2], args[1::2]))
```

python/cudf/cudf/options.py
```diff
@@ -114,6 +115,40 @@ def set_option(name: str, val: Any):
...
+
+    def __enter__(self) -> None:
+        self.undo = [(pat, get_option(pat)) for pat, _ in self.ops]
```
wence-, line 141:
```suggestion
        self.undo = tuple((pat, get_option(pat)) for pat, _ in self.ops)
```

python/cudf/cudf/options.py
```diff
@@ -114,6 +115,40 @@ def set_option(name: str, val: Any):
...
+
+        for pat, val in self.ops:
+            set_option(pat, val)
+
+    def __exit__(self, *args) -> None:
+        if self.undo:
+            for pat, val in self.undo:
+                set_option(pat, val)
```
wence-, line 149:
```suggestion
        for pat, val in self.undo:
            set_option(pat, val)
```

python/cudf/cudf/options.py
wence-: > I tried to read the API docs and answer this question but I couldn't. Does a cudf option for display.max_rows exist? Do the options propagate from cudf to pandas somehow? I know pandas' options for display.max_rows will affect cudf, but wasn't sure about the opposite. If this doesn't exist, we should change the docstring example to something that exists.

Given that we rely on pandas, should we migrate the cudf options to just _using_ the pandas options object, and namespace any cudf-specific options behind `cudf.WHATEVER`. That way we delete code (good), and don't have to worry about keeping things in sync.

> Also we need to explicitly list all cudf options someplace in our docs. I didn't find it when I looked. Does such a list exist?

I think no.

wence- (commented):
Perhaps my complaints are better directed as upstream enhancement requests.

python/cudf/cudf/options.py
bdice: @wence- I think we decided against reusing pandas' options directly because we want to ensure that cudf has control of its own options if pandas decides to change its API, or place restrictions on what constitutes a valid key or value. There is also an implicit expectation that if cudf uses pandas' options, it also _supports_ all pandas options -- which we do not.

python/cudf/cudf/options.py
bdice: For what it's worth -- I had some of the same suggestions/questions in mind as @wence- here (and in other comments) but I deferred to matching the `pandas` code. This behavior is identical to what pandas already has. I am fine with applying the suggestions, though.

python/cudf/cudf/options.py
wence-: Thanks, makes sense.

python/cudf/cudf/options.py
wence-: Yeah, this suggestion is forward-compatible with pandas usage, so I think it is fine to diverge.

python/cudf/cudf/options.py
galipremsagar: I updated the docstring with cudf specific parameters. Also, added the available options section to our docs.

python/cudf/cudf/options.py
galipremsagar: I changed the behavior of `with cudf.options_contex()` to reset all `cudf` options to default.

python/cudf/cudf/options.py
galipremsagar: Done 👍

python/cudf/cudf/options.py
bdice: Sorry for being unclear, I was posing that mostly as a hypothetical. I would recommend adhering to the same behavior as pandas. Just layer over what is already set, without clearing the user options that exist.

python/cudf/cudf/options.py
```diff
@@ -319,3 +320,49 @@ def _integer_and_none_validator(val):
...
+            self.undo = tuple(
+                (pat, option.value) for pat, option in _OPTIONS.items()
+            )
+            old_options_copy = _OPTIONS.copy()
+            for name, option in old_options_copy.items():
+                set_option(name, option.default)
+        else:
+            self.undo = tuple((pat, get_option(pat)) for pat, _ in self.ops)
+            for pat, val in self.ops:
+                set_option(pat, val)
```
galipremsagar, line 364:
```suggestion
        self.undo = tuple((pat, get_option(pat)) for pat, _ in self.ops)
        for pat, val in self.ops:
            set_option(pat, val)
```

python/cudf/cudf/options.py
```diff
@@ -319,3 +320,49 @@ def _integer_and_none_validator(val):
...
+    """
+    Context manager to temporarily set options in the `with` statement context.
+
+    You need to invoke as ``option_context(pat, val, [(pat, val), ...])``.
+
+    .. pandas-compat::
+            One notable difference from Pandas is when `option_context` is
+            created with an empty input i.e., `option_context()` all the
+            options of `cudf` are reset to default values with-in
+            the scope of the context.
```
galipremsagar, line 335:
```suggestion
```

python/cudf/cudf/tests/test_options.py
```diff
@@ -45,33 +46,74 @@ def validator(x):
...
+    with cudf.option_context():
+        assert cudf.get_option("default_integer_bitwidth") is None
+        assert cudf.get_option("default_float_bitwidth") is None
+        assert cudf.get_option("spill") is os.getenv("CUDF_SPILL", False)
+        assert cudf.get_option("copy_on_write") is os.getenv(
+            "CUDF_COPY_ON_WRITE", False
+        )
+        assert cudf.get_option("mode.pandas_compatible") is False
+    assert cudf.get_option("default_integer_bitwidth") == 32
+    cudf.set_option("default_integer_bitwidth", prev_setting)
```
galipremsagar, line 94:
```suggestion
def test_empty_option_context():
    prev_setting = cudf.get_option("default_integer_bitwidth")
    cudf.set_option("default_integer_bitwidth", 32)
    with cudf.option_context():
        assert cudf.get_option("default_integer_bitwidth") == 32
        assert cudf.get_option("default_float_bitwidth") is None
        assert cudf.get_option("spill") is os.getenv("CUDF_SPILL", False)
        assert cudf.get_option("copy_on_write") is os.getenv(
            "CUDF_COPY_ON_WRITE", False
        )
        assert cudf.get_option("mode.pandas_compatible") is False
    assert cudf.get_option("default_integer_bitwidth") == 32
    cudf.set_option("default_integer_bitwidth", prev_setting)
```

python/cudf/cudf/options.py
galipremsagar: No worries, reverted that change 👍

python/cudf/cudf/tests/test_options.py
```diff
@@ -45,33 +46,74 @@ def validator(x):
...
+
+
+def test_option_context():
+    prev_setting = cudf.get_option("mode.pandas_compatible")
+    cudf.set_option("mode.pandas_compatible", True)
+    assert cudf.get_option("mode.pandas_compatible") is True
+    with cudf.option_context(
+        "mode.pandas_compatible", False, "default_integer_bitwidth", 32
```
wence-, line 102: Same comment about default integer bitwidth here too. Can we use a value that is guaranteed not the same as the one that is set (e.g. by saying `old_val = get_option(...)` `option_context(..., old_val + 1)`

wence- (approved):
Some minor fixes in the tests, otherwise LGTM, thanks!

python/cudf/cudf/tests/test_options.py
galipremsagar: Done 👍

python/cudf/cudf/tests/test_options.py
```diff
@@ -45,33 +46,74 @@ def validator(x):
...
+    ):
+        assert cudf.get_option("mode.pandas_compatible") is False
+        assert cudf.get_option("default_integer_bitwidth") == 32
+        cudf.set_option("mode.pandas_compatible", True)
+        assert cudf.get_option("mode.pandas_compatible") is True
+    assert cudf.get_option("mode.pandas_compatible") is True
+    cudf.set_option("mode.pandas_compatible", prev_setting)
```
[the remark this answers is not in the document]
galipremsagar: Added a check 👍

python/cudf/cudf/tests/test_options.py
```diff
@@ -45,33 +46,74 @@ def validator(x):
...
+        expected = (
+            "odd_option:\n\tAn odd option.\n\t[Default: 1] [Current: 1]\n"
+            "even_option:\n\tAn even option.\n\t[Default: 0] [Current: 0]\n"
+        )
+        assert expected == s.read()
+
+
+def test_empty_option_context():
+    prev_setting = cudf.get_option("default_integer_bitwidth")
+    cudf.set_option("default_integer_bitwidth", 32)
```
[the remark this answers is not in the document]
galipremsagar: Made a change to work-around that problem.

python/cudf/cudf/tests/test_options.py
```diff
@@ -45,33 +47,96 @@ def validator(x):
...
+        expected = (
+            "odd_option:\n\tAn odd option.\n\t[Default: 1] [Current: 1]\n"
+            "even_option:\n\tAn even option.\n\t[Default: 0] [Current: 0]\n"
+        )
+        assert expected == s.read()
+
+
+def test_empty_option_context():
+    prev_setting = cudf.get_option("default_integer_bitwidth")
+    set_default_integer_bitwidth = random.choice(
```
bdice, line 85: Don’t randomly select. Write a deterministic test. You could parameterize over multiple values or set the value multiple times in the same test to ensure that at least one of them is not the default.

bdice: And maybe we only need to test one option here, not pandas compat and bitwidth and spilling and etc.

bdice (approved):
The tests with random option values are a little convoluted. I would try to simplify them. Otherwise LGTM. I know this is blocking other work so I’ll approve and let you resolve the test simplification question as you wish.

python/cudf/cudf/tests/test_options.py
galipremsagar: Good idea, Simplified it with parametrization.

PR #11645, opened 2022-09-02
Support DECIMAL order-by for RANGE window functions

mythrocks:
## Description
CUDF grouped RANGE window functions currently support only
integral types and timestamps as the ORDER BY (OBY) column.

This commit adds support for DECIMAL types (i.e. decimal32,
decimal64, and decimal128) to be used as the ORDER BY
column in RANGE window functions.

This feature allows `spark-rapids` to address https://github.com/NVIDIA/spark-rapids/issues/6400.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13512, opened 2023-06-05
Floating point order-by columns for RANGE window functions

mythrocks:
## Description
This commit adds support for `FLOAT32` and `FLOAT64` order-by columns for RANGE-based window functions.

## Background
Up until this commit, order-by columns for RANGE window functions were allowed to be integral numerics, timestamps, or strings (for unbounded/current rows).

With this commit, window functions will be permitted to run on floating point value ranges. E.g. This supports windows defined with floating point deltas, like `rows with values exceeding the current row by no more than 3.14f`.

This is in the same vein as the support for `DECIMAL` (#11645) and `STRING` (#13143).

## Checklist


cpp/src/rolling/range_window_bounds.cpp
```diff
@@ -30,36 +30,28 @@ namespace {
...
+    } else if constexpr (cudf::is_duration<T>()) {
+      return std::make_unique<duration_scalar<T>>(
+        static_cast<duration_scalar<T> const&>(range_scalar_));
+    } else {
+      assert(not detail::is_supported_range_type<T>());
+      CUDF_FAIL(
+        "Unsupported range type. "
+        "Only Durations, fixed-point, floating point, and non-boolean integral range types are "
+        "allowed.");
+    }
```
davidwendt, line 52: I prefer the original specialization approach over this if-else approach. Is there some reason that just adding
```
  template <typename T, CUDF_ENABLE_IF(cudf::is_floating_point<T>())>
  std::unique_ptr<scalar> operator()(scalar const& range_scalar_) const
  {
    return std::make_unique<numeric_scalar<T>>(
      static_cast<numeric_scalar<T> const&>(range_scalar_));
  }
```
would not work here?

ttnghia: I agree that this chain of `if/else` looks worse than the previous approach :smiley:

mythrocks: I've changed it back now.

cpp/src/rolling/range_window_bounds.cpp
```diff
@@ -45,7 +46,9 @@ struct range_scalar_constructor {
       static_cast<duration_scalar<T> const&>(range_scalar_));
   }
 
-  template <typename T, CUDF_ENABLE_IF(std::is_integral_v<T> && not cudf::is_boolean<T>())>
+  template <typename T,
+            CUDF_ENABLE_IF(std::is_floating_point_v<T> ||
```
ttnghia, line 50: I believe that this needs to be removed.

cpp/src/rolling/range_window_bounds.cpp
```diff
@@ -58,8 +61,14 @@ struct range_scalar_constructor {
     return std::make_unique<fixed_point_scalar<T>>(
       static_cast<fixed_point_scalar<T> const&>(range_scalar_));
   }
-};
 
+  template <typename T, CUDF_ENABLE_IF(std::is_floating_point_v<T>())>
+  std::unique_ptr<scalar> operator()(scalar const& range_scalar_) const
+  {
+    return std::make_unique<fixed_point_scalar<T>>(
```
ttnghia, line 68: `fixed_point_scalar` for floating-point?
```suggestion
    return std::make_unique<fixed_point_scalar<T>>(
```

mythrocks: I'm re-checking. I thought I'd fixed that. Thanks for finding that.

mythrocks: Yep. This is messed up. I'll post a fix. 
Odd that this passed tests. I would have thought this would result in ambiguity.

cpp/src/rolling/range_window_bounds.cpp
mythrocks: You are right. I'll remove this shortly.

cpp/src/rolling/range_window_bounds.cpp
mythrocks: Fixed now.

cpp/src/rolling/range_window_bounds.cpp
mythrocks: Removed.

ttnghia (approved):

cpp/src/rolling/range_window_bounds.cpp
```diff
@@ -45,7 +46,9 @@ struct range_scalar_constructor {
       static_cast<duration_scalar<T> const&>(range_scalar_));
   }
 
-  template <typename T, CUDF_ENABLE_IF(std::is_integral_v<T> && not cudf::is_boolean<T>())>
+  template <typename T,
+            CUDF_ENABLE_IF(std::is_floating_point_v<T> ||
+                           (std::is_integral_v<T> && not cudf::is_boolean<T>()))>
```
davidwendt, line 51:
```suggestion
  template <typename T,
            CUDF_ENABLE_IF(cudf::is_numeric<T> && not cudf::is_boolean<T>())>
```
This has a somewhat nice symmetry to it -- `if T then static_cast<T_scalar>`

mythrocks: Sorry, that should have been `cudf::is_numeric<T>()`, not `cudf::is_numeric<T>`. Will fix shortly.

mythrocks: Fixed.

cpp/src/rolling/detail/range_window_bounds.hpp
```diff
@@ -98,7 +101,8 @@ void assert_non_negative([[maybe_unused]] T const& value)
 
 template <typename RangeT,
           typename RepT,
-          CUDF_ENABLE_IF(std::is_integral_v<RangeT> && !cudf::is_boolean<RangeT>())>
+          CUDF_ENABLE_IF(std::is_floating_point_v<RangeT> ||
+                         (std::is_integral_v<RangeT> && !cudf::is_boolean<RangeT>()))>
```
davidwendt, line 105:
```suggestion
          CUDF_ENABLE_IF(cudf::is_numeric<RangeT>() && !cudf::is_boolean<RangeT>())>
```
Could these also use the `cudf::is_numeric` instead? There are 3 other patterns similar to this in this file.

cpp/src/rolling/range_window_bounds.cpp
```diff
@@ -35,7 +35,8 @@ struct range_scalar_constructor {
   {
     CUDF_FAIL(
       "Unsupported range type. "
-      "Only Durations, fixed-point, and non-boolean integral range types are allowed.");
+      "Only Durations, fixed-point, floating point, and non-boolean integral range types are "
+      "allowed.");
```
davidwendt, line 39:
```suggestion
      "Only Durations, fixed-point, and non-boolean numeric range types are allowed.");
```
We could perhaps use `numeric` here as well.

cpp/src/rolling/detail/range_window_bounds.hpp
mythrocks: Good idea. Fixed.

cpp/src/rolling/range_window_bounds.cpp
mythrocks: Fixed here as well.

cpp/src/rolling/detail/range_window_bounds.hpp
```diff
@@ -64,7 +64,9 @@ struct range_type_impl {
 template <typename ColumnType>
 struct range_type_impl<
   ColumnType,
-  std::enable_if_t<std::is_integral_v<ColumnType> && !cudf::is_boolean<ColumnType>(), void>> {
+  std::enable_if_t<std::is_floating_point_v<ColumnType> ||
+                     (std::is_integral_v<ColumnType> && !cudf::is_boolean<ColumnType>()),
+                   void>> {
```
davidwendt, line 69:
```suggestion
  std::enable_if_t<cudf::is_numeric<ColumnType>() && !cudf::is_boolean<ColumnType>(),
                   void>> {
```
One more?

davidwendt (approved):

cpp/src/rolling/detail/range_window_bounds.hpp
mythrocks: D'oh! Sorry, mate. Will fix.

mythrocks: Fixed.

mythrocks:
Thank you for reviewing, chaps. I've merged this change.

PR #13513, opened 2023-06-06
Move stream-related test configuration to CMake

vyasr:
## Description

Instead of configuring test settings with environment variables in the bash script for CI, these changes allow all that configuration to happen directly in CMake, which is more robust and also makes the test configuration portable. With these changes any execution of the tests with `ctest` will use the preload library with all the other necessary environment variables configured as well. Running the test executables directly, on the other hand, will skip stream validation but still run correctly (except for the one test that explicitly tests the stream validation preload lib itself). These changes are now possible because our CI test scripts now run ctest instead of the executables (as of #12451) and due to full support for generator expression propagation even with the parallel testing feature (as of https://github.com/rapidsai/rapids-cmake/pull/410).

## Checklist


.github/CODEOWNERS
```diff
@@ -9,6 +9,8 @@ python/dask_cudf/  @rapidsai/cudf-dask-codeowners
 
 #cmake code owners
 cpp/CMakeLists.txt               @rapidsai/cudf-cmake-codeowners
+cpp/tests/CMakeLists.txt         @rapidsai/cudf-cmake-codeowners
```
robertmaynard, line 12: Are you planning on reverting this change? I presumed you added it to get a `cudf-cmake-codeowners` ping. We removed this from code owners to reduce amount of reviews for tests/benchmarks since they generally are just adding new files

vyasr: I didn't realize that was an intentional removal. I'll drop them.

vyasr: Removed.

Issue #10706, opened 2022-04-21
Remove cudautils.py


vyasr:
Currently the contents of cudautils.py can be grouped into three sets of functionality:
1. The first set of 7 functions are all targeted at supporting `find_first` and `find_last`, which in turn are used respectively by the `find_first_value` and `find_last_value` methods of Column objects.
2. The second set of functions 4 around window sizes are used by the rolling calculations in rolling.py.
3. The third pair of 2 functions is used for UDF compilation.

The functions in group 3 are necessary but should probably be moved to `core/udf/utils.py`. The remaining functions should be possible to remove altogether, but they may require some additional functionality to be added to libcudf. For instance, `find_first_value` and `find_last_value` may be possible to implement using libcudf's `lower_bound` followed by an equality check, and similarly `find_last_value` could be replaced with `upper_bound`, accounting for all of group 1. For group 2 functions I am not certain if the necessary functionality exists in libcudf's roling aggregations, so we may need to do a little more engineering there first.

vyasr:
@bdice I think you did some recent investigations here, please add any color that I'm missing.

bdice:
I agree with the proposal above. There may be a catch with circular imports for group 3 / UDF helpers.

brandon-b-miller:
I think `find_first` is used in the `iloc` codepath. That part of the code needs to be reworked anyways since it's not producing the right result in some cases.

Let me know if I may be of assistance untangling anything in the UDF area @bdice

vyasr:
> I think `find_first` is used in the `iloc` codepath. That part of the code needs to be reworked anyways since it's not producing the right result in some cases.

Yup, I think that `Column.find_first_value` gets invoked by the `loc_to_iloc` translation. There's also #10657 about trying to unify DataFrame and Series loc/iloc further, so maybe all of the relevant issues could be resolved in one fell swoop.

vyasr:
Assigning to @brandon-b-miller and @wence- for now. I think that Lawrence's work on rolling should make the rolling functions obsolete. After that we can reassess the first set.

PR #13517, opened 2023-06-06
Remove numba kernels from `find_index_of_val`

brandon-b-miller:
This PR replaces a few kernels in `cudautils` with equivalent logic composed of libcudf operations. This should have the effect of removing numba from our indexing codepath. 

Contributes to #10706


python/cudf/cudf/utils/cudautils.py
```diff
@@ -79,23 +30,27 @@ def find_index_of_val(arr, val, mask=None, compare="eq"):
...
+            if arr.dtype in ("float32", "float64"):
+                if np.isnan(val):
+                    locations = is_non_nan(arr)
                 else:
-                    gpu_mark_found_int.forall(found.size)(
-                        arr, val, found, arr.size
-                    )
+                    locations = arr != val
+            else:
+                locations = arr != val
```
galipremsagar, line 47: Could be simplified as?:
```
            if arr.dtype in ("float32", "float64") and np.isnan(val):
                locations = is_non_nan(arr)
            else:
                locations = arr != val
```

python/cudf/cudf/utils/cudautils.py
```diff
@@ -79,23 +30,27 @@ def find_index_of_val(arr, val, mask=None, compare="eq"):
...
-            elif compare == "lt":
-                gpu_mark_lt.forall(found.size)(arr, val, found, arr.size)
-            else:
-                if arr.dtype in ("float32", "float64"):
-                    gpu_mark_found_float.forall(found.size)(
-                        arr, val, found, arr.size
-                    )
+    arr = cudf.core.column.as_column(arr, nan_as_null=False)
+    target = cudf.core.column.arange(0, arr.size, dtype="int32")
+    if arr.size > 0:
```
bdice, line 35: It looks like `found` is never defined if this conditional is not met, which I suspect would yield a `NameError` in the return line.

python/cudf/cudf/utils/cudautils.py
```diff
@@ -16,56 +17,6 @@
...
-    if i < arr.size:
-        if arr[i] < val:
-            out[i] = i
-        else:
-            out[i] = not_found
-
-
 def find_index_of_val(arr, val, mask=None, compare="eq"):
     """
     Returns the indices of the occurrence of *val* in *arr*
```
bdice, line 22: A broader question: why is this function designed like this? The return value isn't what I expect for "Returns the indices of the occurrence of *val* in *arr*".

```python
>>> s = cudf.Series([0, 3, 3, 2, 5, 3])
>>> cudf.utils.cudautils.find_index_of_val(s, 3)
<cudf.core.column.numerical.NumericalColumn object at 0x...>
[
  6,
  1,
  2,
  6,
  6,
  5
]
dtype: int32
```

I expected `[1, 2, 5]` to be the result, not `[6, 1, 2, 6, 6, 5]`. Returning a full column that is the length of the input array indicates that we might have room to improve here.

python/cudf/cudf/utils/cudautils.py
```diff
@@ -79,23 +30,27 @@ def find_index_of_val(arr, val, mask=None, compare="eq"):
...
-                gpu_mark_gt.forall(found.size)(arr, val, found, arr.size)
-            elif compare == "lt":
-                gpu_mark_lt.forall(found.size)(arr, val, found, arr.size)
-            else:
-                if arr.dtype in ("float32", "float64"):
-                    gpu_mark_found_float.forall(found.size)(
-                        arr, val, found, arr.size
-                    )
+    arr = cudf.core.column.as_column(arr, nan_as_null=False)
+    target = cudf.core.column.arange(0, arr.size, dtype="int32")
```
bdice, line 34: Can we construct `target` later, closer to where it is needed?

python/cudf/cudf/utils/cudautils.py
bdice: We definitely need to refactor this more broadly, after removing numba. We only use `find_index_of_val` in `find_first` and `find_last`, which are in turn only used in one place each:
- https://github.com/rapidsai/cudf/blob/b8c0501d184a63c59cd70aca189b4887ba4f13fd/python/cudf/cudf/core/column/numerical.py#L611
- https://github.com/rapidsai/cudf/blob/b8c0501d184a63c59cd70aca189b4887ba4f13fd/python/cudf/cudf/core/column/numerical.py#L624

I think we have taken on much more complexity here than we really need.

bdice (changes requested):

python/cudf/cudf/utils/cudautils.py
bdice: Do we even need to check the dtype, or is it enough to check `np.isnan(val)`?

bdice: If so, we could write a one-liner:
```python
locations = is_non_nan(arr) if np.isnan(val) else arr != val
```

bdice:
@brandon-b-miller Thank you SO MUCH for doing this. This is an area of cudf that I've been wanting to improve for ages.

python/cudf/cudf/utils/cudautils.py
brandon-b-miller: This works great!

python/cudf/cudf/utils/cudautils.py
brandon-b-miller: Done!

python/cudf/cudf/utils/cudautils.py
brandon-b-miller: I think these are just fairly old functions for which this is one of the last usages. It's usage in the indexing codepath isn't even right, as it's used to produce results that deviate from pandas in some cases. I expect the function will ultimately be removed entirely as part of @wence- 's broader refactor of indexing, making this PR more of a stopgap as part of the wider MVC effort around organizing numba features. If that's the case then we may as well just wait for the bulk of this to be removed.

bdice: Great. Removing these entirely is my hope as well.

bdice (approved):

galipremsagar (approved):

python/cudf/cudf/utils/cudautils.py
wence-: > I expected [1, 2, 5] to be the result, not [6, 1, 2, 6, 6, 5]. Returning a full column that is the length of the input array indicates that we might have room to improve here.

Hmm, yes, agreed. But I am rewriting anyway as Brandon says.

PR #13520, opened 2023-06-07
Use sizes_to_offsets_iterator in cudf::gather for strings

davidwendt:
## Description
Changes `cudf::gather` specialization for strings to use the `sizes_to_offsets_iterator` to build the offsets column replacing 3 kernel calls with 2.
Also adds a strings-gather benchmark using nvbench.

## Checklist


davidwendt:
This did not improve performance as much as I had hoped.

cpp/include/cudf/strings/detail/gather.cuh
```diff
@@ -294,59 +293,31 @@ std::unique_ptr<cudf::column> gather(strings_column_view const& strings,
...
-                                       mr);
+  auto const offsets_view = out_offsets_column->view();
+  auto out_chars_column   = gather_chars(
+    d_strings->begin<string_view>(), begin, end, offsets_view, total_bytes, stream, mr);
 
   return make_strings_column(output_count,
                              std::move(out_offsets_column),
                              std::move(out_chars_column),
-                             0,
+                             0,  // caller sets these
```
bdice, line 320: Design question: the caller is currently expected to set the validity mask of the output. Can you explain precisely what this means? Are null inputs currently replaced by empty strings in the output? Is the caller supposed to apply the same gather map to the input null mask, then assign that mask/count to the output column? It makes me wonder if we should use a fancier iterator to handle mask gathering at the same time as string gathering, reducing the number of kernel launches. Does that negatively impact performance?

davidwendt: This is just a specialization of gather for this type.  There is some common code before/after the dispatcher that handles nulls. This specialization is not called without that dispatch logic. This particular implementation could build the nulls but there is no reason to do it twice. We could make each specialization handle the nulls but that could duplicate the bitmask computation across many functions.
I think it would be difficult to make a fancy iterator for gathering individual bits perform well over the current bitmask gather kernel.  Either way, I think this is a bigger discussion and out of scope for this PR.

bdice: Great! That’s exactly what I needed to know. Thanks for the context.

davidwendt:
Performance results appear to be mostly noise but also mostly better.
```
| width |    rows  |   Ref Time |   New Time |        Diff |   %Diff |
|-------|----------|------------|------------|-------------|---------|
|  32   |     4096 |  70.995 us |  69.658 us |   -1.337 us |  -1.88% |
|  64   |     4096 |  73.853 us |  71.359 us |   -2.494 us |  -3.38% |
|  128  |     4096 |  69.864 us |  68.240 us |   -1.624 us |  -2.32% |
|  256  |     4096 |  69.582 us |  68.212 us |   -1.370 us |  -1.97% |
|  512  |     4096 |  69.955 us |  68.676 us |   -1.279 us |  -1.83% |
| 1024  |     4096 |  71.891 us |  70.623 us |   -1.268 us |  -1.76% |
| 2048  |     4096 |  76.055 us |  74.704 us |   -1.351 us |  -1.78% |
| 4096  |     4096 |  89.793 us |  87.576 us |   -2.217 us |  -2.47% |
|  32   |    32768 |  76.501 us |  77.473 us |    0.972 us |   1.27% |
|  64   |    32768 |  85.991 us |  81.154 us |   -4.837 us |  -5.62% |
|  128  |    32768 |  83.309 us |  78.000 us |   -5.309 us |  -6.37% |
|  256  |    32768 |  82.419 us |  80.655 us |   -1.764 us |  -2.14% |
|  512  |    32768 |  82.855 us |  86.613 us |    3.759 us |   4.54% |
| 1024  |    32768 | 103.508 us | 100.596 us |   -2.912 us |  -2.81% |
| 2048  |    32768 | 134.220 us | 130.953 us |   -3.267 us |  -2.43% |
| 4096  |    32768 | 195.035 us | 192.203 us |   -2.833 us |  -1.45% |
|  32   |   262144 | 127.028 us | 122.849 us |   -4.179 us |  -3.29% |
|  64   |   262144 | 162.247 us | 160.995 us |   -1.253 us |  -0.77% |
|  128  |   262144 | 169.674 us | 167.918 us |   -1.756 us |  -1.03% |
|  256  |   262144 | 177.069 us | 175.267 us |   -1.802 us |  -1.02% |
|  512  |   262144 | 208.686 us | 206.245 us |   -2.441 us |  -1.17% |
| 1024  |   262144 | 298.197 us | 296.576 us |   -1.621 us |  -0.54% |
| 2048  |   262144 | 534.013 us | 531.678 us |   -2.334 us |  -0.44% |
| 4096  |   262144 |   1.016 ms |   1.013 ms |   -3.433 us |  -0.34% |
|  32   |  2097152 | 531.373 us | 511.363 us |  -20.010 us |  -3.77% |
|  64   |  2097152 | 847.085 us | 818.949 us |  -28.136 us |  -3.32% |
|  128  |  2097152 | 829.635 us | 805.675 us |  -23.960 us |  -2.89% |
|  256  |  2097152 | 888.701 us | 865.128 us |  -23.573 us |  -2.65% |
|  512  |  2097152 |   1.112 ms |   1.085 ms |  -27.146 us |  -2.44% |
|  32   | 16777216 |   3.864 ms |   3.722 ms | -142.410 us |  -3.69% |
|  64   | 16777216 |   6.395 ms |   6.226 ms | -169.344 us |  -2.65% |
```

ttnghia (approved):

harrism (approved):
Even if it isn't faster, it's simpler!

PR #13521, opened 2023-06-07
Fix some memoization logic in groupby/sort/sort_helper.cu

davidwendt:
## Description
Fixes some logic in the `groupby/sort/sort_helper.cu` to improve memoization safety. Some functions cache their results so they may be called several times. This change ensures the cached variable is fully formed before returning it so that parallel calls do not get a partially built result. The code also simplifies some of the logic by replacing some thrust calls with equivalent `cudf::detail` function calls that return columns.

Found while working on #13489 

## Checklist


PointKernel (approved):
Much cleaner. Thanks!

cpp/src/groupby/sort/sort_helper.cu
```diff
@@ -170,22 +161,23 @@ sort_groupby_helper::index_vector const& sort_groupby_helper::group_offsets(
...
                                      thrust::counting_iterator<size_type>(size),
-                                     _group_offsets->begin(),
+                                     group_offsets->begin(),
                                      permuted_row_equality_comparator(d_key_equal, sorted_order));
   }
 
-  size_type num_groups = thrust::distance(_group_offsets->begin(), result_end);
-  _group_offsets->set_element(num_groups, size, stream);
-  _group_offsets->resize(num_groups + 1, stream);
+  size_type num_groups = thrust::distance(group_offsets->begin(), result_end);
```
ttnghia, line 176: This will most likely lead to a down cast which may cause compiler warning in the future. So please use `std::size_t` or `auto const` type for `num_groups`.
```suggestion
  auto const num_groups = thrust::distance(group_offsets->begin(), result_end);
```

cpp/src/groupby/sort/sort_helper.cu
```diff
@@ -170,22 +161,23 @@ sort_groupby_helper::index_vector const& sort_groupby_helper::group_offsets(
...
+  group_offsets->set_element(num_groups, size, stream);
+  group_offsets->resize(num_groups + 1, stream);
 
+  _group_offsets = std::move(group_offsets);
```
ttnghia, line 180: Please also add a comment clarifying why `_group_offsets` is not modified from the beginning. I can understand this only after reading the PR description.

cpp/src/groupby/sort/sort_helper.cu
```diff
@@ -170,22 +161,23 @@ sort_groupby_helper::index_vector const& sort_groupby_helper::group_offsets(
...
-                                     _group_offsets->begin(),
+                                     group_offsets->begin(),
                                      permuted_row_equality_comparator(d_key_equal, sorted_order));
   }
 
-  size_type num_groups = thrust::distance(_group_offsets->begin(), result_end);
-  _group_offsets->set_element(num_groups, size, stream);
-  _group_offsets->resize(num_groups + 1, stream);
+  size_type num_groups = thrust::distance(group_offsets->begin(), result_end);
+  group_offsets->set_element(num_groups, size, stream);
```
ttnghia, line 177: Can we use `set_element_async`?

cpp/src/groupby/sort/sort_helper.cu
```diff
@@ -195,17 +187,18 @@ sort_groupby_helper::index_vector const& sort_groupby_helper::group_labels(
...
   // Get group labels for future use in segmented sorting
-  _group_labels = std::make_unique<index_vector>(num_keys(stream), stream);
-
-  auto& group_labels = *_group_labels;
-  if (num_keys(stream) == 0) return group_labels;
+  auto group_labels = std::make_unique<index_vector>(num_keys(stream), stream);
+
+  if (num_keys(stream)) {
+    cudf::detail::label_segments(group_offsets(stream).begin(),
+                                 group_offsets(stream).end(),
```
ttnghia, line 194: I have no idea if the compiler in the future can execute these two lines in parallel in multiple threads, which in turn fully executes the `group_offsets(stream)` function twice when the `_group_offsets` vector was not generated. IMO this is better to safeguard us from such situation:
```suggestion
   auto const& offsets = group_offsets(stream);
    cudf::detail::label_segments(offsets.begin(), offsets.end(),
```

davidwendt: This is the premise of the PR in general. I'm not sure if it is possible but thought it best for the code to be correct.

cpp/src/groupby/sort/sort_helper.cu
```diff
@@ -195,17 +189,16 @@ sort_groupby_helper::index_vector const& sort_groupby_helper::group_labels(
   if (_group_labels) return *_group_labels;
 
   // Get group labels for future use in segmented sorting
-  _group_labels = std::make_unique<index_vector>(num_keys(stream), stream);
+  auto group_labels = std::make_unique<index_vector>(num_keys(stream), stream);
```
ttnghia, line 192: Sorry I forgot to mention about commenting for this too.
```suggestion
  // Create a temporary variable and only set _group_labels right before the return.
  // This way, a 2nd (parallel) call to this will not be given a partially created object.
  auto group_labels = std::make_unique<index_vector>(num_keys(stream), stream);
```

cpp/src/groupby/sort/sort_helper.cu
```diff
@@ -242,14 +235,12 @@ column_view sort_groupby_helper::keys_bitmask_column(rmm::cuda_stream_view strea
...
 
-  _keys_bitmask_column = make_numeric_column(
-    data_type(type_id::INT8), _keys.num_rows(), std::move(row_bitmask), null_count, stream);
-
-  auto keys_bitmask_view = _keys_bitmask_column->mutable_view();
-  using T                = id_to_type<type_id::INT8>;
-  thrust::fill(
-    rmm::exec_policy(stream), keys_bitmask_view.begin<T>(), keys_bitmask_view.end<T>(), 0);
+  auto const zero          = numeric_scalar<int8_t>(0);
+  auto keys_bitmask_column = cudf::detail::sequence(
```
ttnghia, line 239: Nit: Similar comment may be needed for this.

ttnghia (approved):

PR #12922, opened 2023-03-10
cuDF: Build CUDA 12 packages

bdice:
## Description
This PR builds `libcudf` and `cudf` conda packages using CUDA 12. Resolves #12771.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13528, opened 2023-06-07
Add benchmark (Google Benchmark) dependency to conda packages.

bdice:
## Description
While working with @DaceT on benchmarks, we found that the benchmark executables in the `libcudf-tests` conda package are not usable because the `libcudf-tests` package is missing a dependency on `benchmark` (Google Benchmark). This PR adds that dependency.

Currently Google Benchmark is fetched by rapids-cmake CPM, and is pinned to version `1.8.0`. This PR matches that pinning, and ensures that a consistent package is used at build time and run time.

This PR also adds a minimal benchmark to CI tests to ensure that benchmarks are runnable.

## Checklist


fetch_rapids.cmake
```diff
@@ -12,6 +12,9 @@
 # the License.
 # =============================================================================
 if(NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/CUDF_RAPIDS.cmake)
+  # TODO: Revert this before merging.
+  set(rapids-cmake-repo vyasr/rapids-cmake)
+  set(rapids-cmake-branch fix/benchmark_name)
```
bdice, line 17: TODO: apply this suggestion after CI logs have been verified.
```suggestion
```

ajschmidt8 (approved):

ci/test_cpp.sh
```diff
@@ -37,5 +37,21 @@ if (( ${SUITEERROR} == 0 )); then
...
+    pushd $CONDA_PREFIX/bin/benchmarks/libcudf
+    ./MERGE_BENCH --benchmark_filter=/2/
+    SUITEERROR=$?
+    popd
+fi
+
+if (( ${SUITEERROR} == 0 )); then
+    # Run a small nvbench benchmark to test that benchmarks are runnable
+    pushd $CONDA_PREFIX/bin/benchmarks/libcudf
+    ./STRINGS_NVBENCH --run-once --benchmark 0
```
bdice, line 51: Does anyone know how nvbench / Google benchmark handle exit codes? It's my hope that they always return `0` for success if the benchmark completes. These are very minimal benchmarks just to test that the executables are valid. I don't want bad performance or noise on the CI system to cause random failures.

cpp/benchmarks/string/case.cpp
```diff
@@ -26,7 +26,7 @@
 void bench_case(nvbench::state& state)
 {
   auto const n_rows    = static_cast<cudf::size_type>(state.get_int64("num_rows"));
-  auto const max_width = static_cast<int32_t>(state.get_int64("width"));
+  auto const max_width = static_cast<int32_t>(state.get_int64("row_width"));
```
bdice, line 29: @davidwendt I slipped in this small fix and wanted to let you know. I saw the benchmarks were failing here, because of a mismatch in axis / parameter names.

bdice:
@GregoryKimball If you have a moment, please review the lines where I'm running a small benchmark in CI. I'd like your thoughts on that. Thanks!

PointKernel (approved):

conda/recipes/libcudf/build.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
-# Copyright (c) 2018-2022, NVIDIA CORPORATION.
+# Copyright (c) 2018-2023, NVIDIA CORPORATION.
 
 export cudf_ROOT="$(realpath ./cpp/build)"
-./build.sh -n -v libcudf libcudf_kafka benchmarks tests --build_metrics --incl_cache_stats --cmake-args=\"-DCMAKE_INSTALL_LIBDIR=lib -DCUDF_ENABLE_ARROW_S3=ON\"
+./build.sh -n -v \
+    libcudf libcudf_kafka benchmarks tests \
+    --build_metrics --incl_cache_stats \
+    --cmake-args=\"-DCMAKE_INSTALL_LIBDIR=lib -DCUDF_ENABLE_ARROW_S3=ON -DNVBench_ENABLE_CUPTI=OFF\"
```
bdice, line 8: We must disable the NVBench option for cupti in our conda packages of `libcudf-tests` because cupti breaks minor version compatibility. If we build with CUDA 11.8 and cupti enabled, the benchmarks can only be run with CUDA 11.8. I decided that being able to run benchmarks with other CUDA runtimes (of the same major version) is important enough to disable the additional metrics provided by cupti. If those metrics are needed, libcudf benchmarks can be built from source or cupti support can be enabled for a specific build of the `libcudf-tests` package by changing this flag.

bdice: Disabling this option also simplifies the CUDA 12 conda packaging because it means we don't have to depend on cupti. Using cupti was a pain point during early tests of CUDA 12 conda builds: https://github.com/rapidsai/cudf/pull/12922#issuecomment-1478359438

vyasr (changes requested):

ci/test_cpp.sh
robertmaynard: Yes the benchmarks will always return `0` for success

ci/test_cpp.sh
```diff
@@ -8,20 +8,38 @@ trap "EXITCODE=1" ERR
...
+
+if (( ${SUITEERROR} == 0 )); then
+    # Run a small Google benchmark
+    ./MERGE_BENCH --benchmark_filter=/2/
+    SUITEERROR=$?
+fi
+
+if (( ${SUITEERROR} == 0 )); then
+    # Run a small nvbench benchmark
+    ./STRINGS_NVBENCH --run-once --benchmark 0
```
davidwendt, line 39:
```suggestion
    ./STRINGS_NVBENCH --run-once --benchmark 0 -d 0
```
I would recommend specifying a single device to minimize the number of runs in case multiple GPUs are available.

bdice: Good idea. I'll use the long form of the flag for consistency.
```suggestion
    ./STRINGS_NVBENCH --run-once --benchmark 0 --devices 0
```

vyasr (approved):

PR #13530, opened 2023-06-07
Refactor `hostdevice_vector` `hostdevice_2dvector` and `hostdevice_span`

ttnghia:
The classes `hostdevice_vector` `hostdevice_2dvector` and `hostdevice_span` have the following APIs:
 * `host_to_device(..., bool synchronize = false)`
 * `device_to_host(..., bool synchronize = false)`

Using them is a bit awkward. For example: `buffer.device_to_host(stream)` or `buffer.device_to_host(stream, true)` which do not explicitly say about stream synchronization.

This PR modify them, removing the `bool synchronize` parameter. Now each of them becomes two separate functions instead:
 * `host_to_device_async(...)` and `host_to_device_sync(...)`.
 * `device_to_host_async(...)` and `device_to_host_sync(...)`.

Their usage (such as `buffer.device_to_host_async(stream)`) will be very straightforward and obvious about the stream synchronization. 

In addition, this modification also make the related APIs become aligned/consistent with other APIs such as `make_device_uvector_async`/`make_device_uvector_sync`.

### DIFF
--- a/cpp/src/io/avro/reader_impl.cu
+++ b/cpp/src/io/avro/reader_impl.cu
@@ -218,12 +218,12 @@ rmm::device_buffer decompress_data(datasource& source,
       meta.block_list[i].size   = static_cast<uint32_t>(inflate_out[i].size());
       dst_pos += meta.block_list[i].size;
     }
-    inflate_in.host_to_device(stream);
+    inflate_in.host_to_device_async(stream);
 
     for (int loop_cnt = 0; loop_cnt < 2; loop_cnt++) {
-      inflate_out.host_to_device(stream);
+      inflate_out.host_to_device_async(stream);
       gpuinflate(inflate_in, inflate_out, inflate_stats, gzip_header_included::NO, stream);
-      inflate_stats.device_to_host(stream, true);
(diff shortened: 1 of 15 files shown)


bdice (approved):
This is a nice change, thank you!

bdice:
@ttnghia Can we rename this PR to: "Make synchronization explicit in the names of `hostdevice_*` copying APIs"? That might mean more to readers in the changelog.

bdice:
@ttnghia I noticed that some of https://github.com/ttnghia/cudf/blob/hostdevice_vector/cpp/src/io/utilities/hostdevice_vector.hpp is not covered by the `namespace cudf::detail`. Can you make sure these names are in `cudf::detail` if that's the right place -- and if they're meant to be public, this PR must be marked as `breaking`.

ttnghia:
> Can you make sure these names are in cudf::detail if that's the right place

Yeah that's fine. I'm going to add that too.

The related classes should only be used internally in libcudf. They were just forgotten to put into `cudf::detail::`.

hyperbolic2346 (approved):
I like this change. It makes it super clear the intent. Thanks!

PR #13531, opened 2023-06-07
Switch to scikit-build-core

vyasr:
## Description

Contributes to rapidsai/build-planning#2

## Checklist


dependencies.yaml
```diff
@@ -257,15 +252,20 @@ dependencies:
     common:
       - output_types: [conda, requirements, pyproject]
         packages:
+          - *cmake_ver
```
bdice, line 255: What's the benefit of moving cmake/ninja from `build_all`?

Alternatively we could make them even more specific since those packages are only needed for building `cudf` and `cudf-kafka` but not `dask-cudf` or `custreamz`.

bdice (approved):
Approving with one question.

ajschmidt8 (approved):

dependencies.yaml
vyasr: I made the change to add back cmake/ninja as per our discussion, but I didn't do so in a way aimed at minimizing the final PR diff, I just did what made sense when looking at the state of dependencies.yaml. With this approach we reduce the number of dependency lists used in `py_build_cudf` and `py_build_cudf_kafka` and make them more consistent. Also, this change is already more specific in the way you mean because `build_python_common` (this list) is only used by cudf and cudf-kafka, not the other two.

rjzamora (approved):
dask-cudf changes seem fine to me. Thanks!

dependencies.yaml
bdice: Thanks, I see now. LGTM!

PR #13534, opened 2023-06-08
Implement iloc-getitem using parse-don't-validate approach

### LINKED ISSUE #13013: [BUG] iloc indexing with a Categorical not possible

### LINKED ISSUE #13267: [BUG] Series and DataFrame iloc indexing does not handle `Ellipsis`

### LINKED ISSUE #13515: [BUG] multiindex-iloc sometimes treats keys as label lookups
**Describe the bug**

```python
@pytest.mark.parametrize(
    "indexer",
    [
        0,
        (slice(None), 0),
        pytest.param(
            ([0, 2], 1),
            marks=pytest.mark.xfail(
                reason="https://github.com/rapidsai/cudf/issues"
            ),
        ),
        (slice(None), slice(None)),
        (slice(None), [1, 0]),
        (0, 0),
        (1, [1, 0]),
        pytest.param(
            ([1, 0], 0),
            marks=pytest.mark.xfail(
                reason="https://github.com/rapidsai/cudf/issues"
            ),
        ),
        pytest.param(
            ([1, 2], [0, 1]),
            marks=pytest.mark.xfail(
                reason="https://github.com/rapidsai/cudf/issues"
            ),
        ),
    ],
)
def test_iloc_multiindex(indexer):
    df = pd.DataFrame(
        {"a": [1, 1, 3], "b": [2, 3, 4], "c": [1, 6, 7], "d": [1, 8, 9]}
    ).set_index(["a", "b"])
    cdf = cudf.from_pandas(df)

    expect = df.iloc[indexer]
    actual = cdf.iloc[indexer]
    assert_eq(expect, actual)
```

wence-:
## Description
To simplify the low-level implementation of iloc-based getitem on both
Series and DataFrames, change the dispatching approach to parse the
user-provided "unstructured" key into structured data (an appropriate
tagged union using new dataclasses). At the libcudf level, there are four
styles of indexing we can do:

1. index by slice
2. index by mask
3. index by map
4. index by scalar

iloc keys are parsed into information that tags them by type and
normalises the key to an appropriate column or other low-level object.

This centralises the business logic for index parsing in a
single place, and ensures that downstream consumers of the validated
and normalised indexer don't need to inspect it again to determine
what to do. Note that we treat index by scalar as composition of index
by map with get_element (since that simplifies the logic when
extracting the single row of a dataframe: we want to keep it on
device), but the scalar "type tag" allows us to determine this
unambiguously without reinspecting the key.

The major benefits will come when updating loc-based getitem (where
the parsing rules are more complicated, but eventually turn into one
of the above four cases). In this latter case, we will no longer
attempt to turn a loc-based key into a "user-facing" key for iloc, but
rather will call directly into the pre-parsed interface.

That said, we already provide some performance improvements since we
only do inspection once.

- Closes #13013
- Closes #13267
- Closes #13515

## Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -401,57 +396,80 @@ class _DataFrameIlocIndexer(_DataFrameIndexer):
...
+            # appropriate indices.
+            (name,) = column_names
+            s = Series._from_data(
+                ca.__class__(
+                    {name: ca[name]},
+                    multiindex=ca.multiindex,
+                    level_names=ca.level_names,
+                ),
+                index=index,
+            )
```
wence-, line 424: I guess I can try and solve this TODO in this PR (or do some prep-PRs?).

python/cudf/cudf/core/dataframe.py
```diff
@@ -401,57 +396,80 @@ class _DataFrameIlocIndexer(_DataFrameIndexer):
...
+            return s._get(row_tag, row_key)
```
wence-, line 425: Call into the `Series` pre-parsed "I know what I need" interface.

python/cudf/cudf/core/dataframe.py
```diff
@@ -401,57 +396,80 @@ class _DataFrameIlocIndexer(_DataFrameIndexer):
...
+                nullify=False,
+                normalize_and_check=False,
+            )
+            # Attempt to turn into series.
+            try:
+                # Behaviour difference from pandas, which will merrily
+                # turn any heterogeneous set of columns into a series if
+                # you only ask for one row.
+                new_name = result.index[0]
+                result = Series._concat(
```
wence-, line 463: No need to check if we can downcast to a series, we know we must try here because the indexing row tag was `SCALAR`.

wence- (commented):
Some signposts

python/cudf/cudf/core/series.py
```diff
@@ -181,22 +182,16 @@ class _SeriesIlocIndexer(_FrameIndexer):
...
-        if (
-            isinstance(data, (dict, list))
-            or _is_scalar_or_zero_d_array(data)
-            or _is_null_host_scalar(data)
-        ):
-            return data
-        return self._frame._from_data(
-            {self._frame.name: data},
-            index=cudf.Index(self._frame.index[arg]),
+        tag, key = indexing_utils.normalize_row_iloc_indexer(
```
shwina, line 189: Would it be better to make a named tuple for `(tag, key)`?

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,224 @@
...
+        Tuple of row and (for dataframes) column keys
+
+    Raises
+    ------
+    IndexError
+        If provided a structurally invalid key
+    """
+    # This is more consistent than pandas, using a fixed point
+    # iteration to remove all callables.
+    # See https://github.com/pandas-dev/pandas/issues/53533
```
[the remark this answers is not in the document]
wence-: The pandas position is that a callable given to `iloc` cannot return a tuple. So I can simplify this function.

python/cudf/cudf/core/series.py
wence-: Probably, then I can change the `_get` signature to just take a single object.

So probably

```
class ParsedIndexer(NamedTuple):
    tag: IndexTag
    key: slice | NumericalColumn # Probably that will cause too many issues.
```

python/cudf/cudf/core/series.py
```diff
@@ -1306,6 +1301,53 @@ def map(self, arg, na_action=None) -> "Series":
             result = self.apply(arg)
         return result
 
+    def _get(
```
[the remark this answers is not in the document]
wence-: How about `_normalized_getitem(self, key: ParsedIndexer)` ?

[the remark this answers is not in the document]
shwina: That's fine by me (with a slight preference for `_getitem_normalized`).

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,225 @@
...
+    is_bool_dtype,
+    is_integer,
+    is_integer_dtype,
+)
+
+if TYPE_CHECKING:
+    from cudf.core.column import ColumnBase
+
+
+class IndexTag(enum.IntEnum):
```
shwina, line 25: How about:

```suggestion
class IndexerTag(enum.IntEnum):
```

`Index` is a bit overloaded in Pandas, and I believe indexer is the more appropriate term here anyway.

shwina: Ditto below for `IndexSpec`.

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+IndexSpec: TypeAlias = Tuple[IndexTag, Union[slice, "ColumnBase"]]
+ColumnLabels: TypeAlias = List[str]
+
+
+def unpack_iloc_key(
+    key: Any, frame: cudf.DataFrame | cudf.Series, n: int
+) -> List[Any]:
+    """Unpack a user-level key to iloc.__getitem__
```
shwina, line 39: What does it mean to "unpack"? Even just a quick example as part of the description here would greatly help readers of this function.

shwina: Ditto below for unpack_dataframe_iloc_indexer

python/cudf/cudf/api/types.py
```diff
@@ -154,10 +154,8 @@ def _is_scalar_or_zero_d_array(val):
         Return True if given object is scalar.
     """
     return (
-        (isinstance(val, (np.ndarray, cp.ndarray)) and val.ndim == 0)
-        or (isinstance(val, pd.Categorical) and len(val) == 1)
-        or is_scalar(val)
-    )
+        isinstance(val, (np.ndarray, cp.ndarray)) and val.ndim == 0
```
shwina, line 157: How/why are we able to discard the `Categorical` check?

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,225 @@
...
+    TypeError
+        If the indexing key is otherwise invalid.
+    """
+    if key is Ellipsis:
+        key = slice(None)
+    if isinstance(key, slice):
+        return (IndexTag.SLICE, key)
+    else:
+        if _is_scalar_or_zero_d_array(key):
+            key = np.asarray(key)
```
shwina, line 185: If `key` happens to be a scalar/array on the device then we're copying it to host and back again to device later. Can we avoid that?

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,225 @@
...
+            if len(key):
+                if check_bounds:
+                    mi, ma = libcudf.reduce.minmax(key)
+                    # Faster than going through cudf.Scalar binop machinery.
+                    if mi.value < -n or ma.value >= n:
+                        raise IndexError("Gather map index is out of bounds.")
+            else:
+                key = cudf.core.column.column_empty(
+                    0, dtype=libcudf.types.size_type_dtype
+                )
```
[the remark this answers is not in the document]
shwina: I like this idea

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -2898,22 +2977,32 @@ def _drop_na_rows(self, how="any", subset=None, thresh=None):
...
         Rows corresponding to `False` is dropped.
+
+        If keep_index is False, the index is not preserved.
         """
-        boolean_mask = cudf.core.column.as_column(boolean_mask)
-        if not is_bool_dtype(boolean_mask.dtype):
-            raise ValueError("boolean_mask is not boolean type.")
-        if (bn := len(boolean_mask)) != (n := len(self)):
-            raise IndexError(f"Boolean mask has wrong length: {bn} not {n}")
+        if normalize_and_check:
```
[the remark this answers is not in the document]
shwina: I think this would be helpful. I especially think the term `normalized` is being overloaded quite a bit and we need to either hash out terminology or use more typing to be able to distinguish between:

* the kind of normalization being done by `unpack_iloc_key`
* the kind of normalization being done by `normalize_row_iloc_indexer`
* other kind of normalization?

shwina:
Just a few general comments for now. This is shaping up to look very nice!

python/cudf/cudf/api/types.py
wence-: Why? Because a length-1 categorical is not a scalar and never has been (from a pandas point of view).

I'm pretty sure this special-case arrived in error at some point in the past.

python/cudf/cudf/core/indexed_frame.py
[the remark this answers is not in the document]
wence-: Let me introduce types for gather maps and boolean masks separately (then the `check_bounds` arguments can disappear here entirely).

python/cudf/cudf/core/indexing_utils.py
wence-: I think this is perhaps now a better implementation:
```python


def destructure_iloc_key(
    key: Any, frame: cudf.Series | cudf.DataFrame
) -> tuple[Any, ...]:
    """
    Destructure a potentially tuple-typed key into row and column indexers

    Tuple arguments to iloc indexing are treated specially. They are
    picked apart into indexers for the row and column. If the number
    of entries is less than the number of modes of the frame, missing
    entries are slice-expanded.

    If the user-provided key is not a tuple, it is treated as if it
    were a singleton tuple, and then slice-expanded.

    Once this destructuring has occurred, any entries that are
    callables are then called with the indexed frame. This should
    return a valid indexing object for the rows (respectively
    columns), namely one of:

    - A boolean mask of the same length as the frame in the given
      dimension
    - A scalar integer that indexes the frame
    - An array-like of integers that index the frame
    - A slice that indexes the frame

    Integer and slice-based indexing follows usual Python conventions.

    Parameters
    ----------
    key
        The key to destructure
    frame
        DataFrame or Series to provide context

    Returns
    -------
    tuple of length equal to the dimension of the frame of indexers

    Raises
    ------
    IndexError
        If there are too many indexers, or any individual indexer is a tuple.
    """
    n = len(frame)
    if isinstance(key, tuple):
        indexers = tuple(
            itertools.chain(key, itertools.repeat(slice(None), n - len(key)))
        )
        if (ni := len(indexers)) > n:
            raise IndexError(f"Too many indexers: got {ni} expected {n}")
    else:
        indexers = (key, *itertools.repeat(slice(None), n - 1))
    indexers = tuple(k(frame) if callable(k) else k for k in indexers)
    if any(isinstance(k, tuple) for k in indexers):
        raise IndexError(
            "Too many indexers: can't have nested tuples in iloc indexing"
        )
    return indexers
```

wence-: > Ditto below for unpack_dataframe_iloc_indexer

This is doing a little bit more. It destructures into row and column indexers (using the pattern above), and then transforms the column indexer into column names.

python/cudf/cudf/core/dataframe.py
```diff
@@ -401,57 +396,80 @@ class _DataFrameIlocIndexer(_DataFrameIndexer):
...
+                keep_index=True,
+                nullify=False,
+                normalize_and_check=False,
+            )
+        elif row_tag is indexing_utils.IndexTag.MASK:
+            return frame._apply_boolean_mask(
+                row_key, keep_index=True, normalize_and_check=False
+            )
+        elif row_tag is indexing_utils.IndexTag.SLICE:
+            return frame._slice(cast(slice, row_key))
```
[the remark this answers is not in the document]
shwina: Hmm, are there other types we could use than a `tuple` that would _perhaps_ help the type checker see things better? Admittedly, I can't think of one.

python/cudf/cudf/core/indexing_utils.py
wence-: Likely yes...

python/cudf/cudf/core/dataframe.py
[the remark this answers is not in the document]
wence-: well, I can just make dataclasses and do `isinstance` dispatch. I guess that's not so bad actually. The cost of making a `NamedTuple` with two int fields is the same as that of making a dataclass (200ns); as opposed to a plain tuple which is 5ns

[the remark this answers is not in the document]
wence-: I the scheme of things that's probably small.

python/cudf/cudf/core/indexing_utils.py
wence-: Used that suggestion.

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,254 @@
...
+@dataclass
+class ScalarIndexer:
+    """An indexer for a scalar value"""
+
+    gather_map: ct.GatherMap
+
+
+IndexingSpec: TypeAlias = (
+    EmptyIndexer | MapIndexer | MaskIndexer | ScalarIndexer | SliceIndexer
+)
```
wence-, line 58: Replaced the "untyped" tuple by dataclasses.

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,254 @@
...
+    check_bounds
+        If True, perform bounds checking of the key if it is a gather
+        map.
+
+    Returns
+    -------
+    IndexingSpec
+        Structured data for indexing. The first entry is a
+        :class:`IndexingTag` tag, the second entry is normalized
+        arguments to the tag-specific indexing routine.
```
wence-, line 218: Update docstring in light of new return type.

python/cudf/cudf/core/copy_types.py
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
```
wence-, line 1: New code after discussion that introduces "validated" types for columns suitable as gather maps and boolean masks. Used by the new indexing routine.

python/cudf/cudf/core/join/join.py
```diff
@@ -190,17 +191,25 @@ def perform_merge(self) -> cudf.DataFrame:
...
             "keep_index": self._using_left_index or self._using_right_index,
         }
         left_result = (
-            self.lhs._gather(gather_map=left_rows, **gather_kwargs)
+            self.lhs._gather(
+                ct.as_gather_map(
+                    left_rows, len(self.lhs), nullify=True, check_bounds=False
+                ),
+                **gather_kwargs,
+            )
```
wence-, line 202: `_gather` now takes a `GatherMap`, so we must update callers everywhere. A bigger refactoring would slowly push these types into more places, but here I have just done the "closest" possible change.

python/cudf/cudf/core/series.py
```diff
@@ -1306,6 +1301,41 @@ def map(self, arg, na_action=None) -> "Series":
...
+            return self._apply_boolean_mask(spec.mask, keep_index=True)
+        elif isinstance(spec, iu.SliceIndexer):
+            return self._slice(spec.slice)
+        elif isinstance(spec, iu.ScalarIndexer):
+            return self._gather(
+                spec.gather_map, keep_index=False
+            )._column.element_indexing(0)
+        elif isinstance(spec, iu.EmptyIndexer):
+            return self._empty_like(keep_index=True)
+        assert_never(spec)
```
wence-, line 1337: Will be tidier once we drop 3.9 and can use `match`.

python/cudf/cudf/core/indexing_utils.py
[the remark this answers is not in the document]
wence-: Done.

python/cudf/cudf/core/series.py
wence-: Used dataclasses so I can add type annotations the members.

python/cudf/cudf/core/indexing_utils.py
wence-: Went for `Indexing` but can be persuaded to use `Indexer` as well.

python/cudf/cudf/core/indexing_utils.py
wence-: I now just directly turn it into a column and then inspect the dtype there.

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,254 @@
...
+
+    Raises
+    ------
+    IndexError
+        If there are too many indexers, or any individual indexer is a tuple.
+    """
+    n = len(frame.shape)
+    if isinstance(key, tuple):
+        indexers = tuple(
+            itertools.chain(key, itertools.repeat(slice(None), n - len(key)))
```
vyasr, line 109: It's probably negligible in something of this complexity, but I have on occasion noted itertools as a significant perf bottleneck so it may be worth double-checking that it's not noticeable here.

vyasr (commented):
Just did a first pass through. This is great! Much cleaner and more consistent treatment of how indexers are set up. I've verified the general approach, but I'll need to take another pass to check that all the actual conditions in each case are valid. Will do that ASAP.

python/cudf/cudf/core/_base_index.py
```diff
@@ -71,6 +71,10 @@ class BaseIndex(Serializable):
     _accessors: Set[Any] = set()
     _data: ColumnAccessor
 
+    @property
+    def _columns(self) -> Tuple[Any, ...]:
+        raise NotImplementedError
```
wence-, line 76: Needed to pacify the type-checker.

python/cudf/cudf/core/indexing_utils.py
wence-: Did the "simple" thing since it's actually easier to read that way.

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,250 @@
...
+
+@dataclass
+class ScalarIndexer:
+    """An indexer for a scalar value"""
+
+    gather_map: ct.GatherMap
+
+
+IndexingSpec: TypeAlias = (
+    EmptyIndexer | MapIndexer | MaskIndexer | ScalarIndexer | SliceIndexer
```
wence-, line 56: TODO: Use `Union` for Py3.9 support.

python/cudf/cudf/core/dataframe.py
```diff
@@ -401,57 +405,55 @@ class _DataFrameIlocIndexer(_DataFrameIndexer):
...
-        # Iloc Step 4:
-        # Downcast
-        if self._can_downcast_to_series(df, arg):
-            return self._downcast_to_series(df, arg)
-
-        if df.shape[0] == 0 and df.shape[1] == 0 and isinstance(arg[0], slice):
-            df._index = as_index(self._frame.index[arg[0]])
-        return df
+            frame = self._frame
+        if isinstance(row_spec, iu.MapIndexer):
```
vyasr, line 431: With the typing that you've added it could be helpful to extract this logic into a `functools.singledispatch` helper function.

vyasr (approved):
This is a major improvement IMO. I don't have much to suggest here; the changeset is quite large, and generally everything looks good, so I'm inclined to merge sooner rather than later and seek incremental improvements.

python/cudf/cudf/core/dataframe.py
wence-: I thought about that, but it has the slight downside that the typecheckers can't see through the dispatch registration to determine that the base registration's `assert_never(spec)` is indeed unreachable:

https://mypy-play.net/?mypy=latest&python=3.11&gist=1a7dcd4f626de70aae853467666dc4b2

You might notice that this dispatch is _almost_ identical to `Series._getitem_preprocessed` apart from the attempt to downcast to a 1xN result to a series.... Should really propose a deprecation in pandas.

python/cudf/cudf/core/copy_types.py
```diff
@@ -0,0 +1,140 @@
...
+import cudf._lib as libcudf
+from cudf._lib.types import size_type_dtype
+
+if TYPE_CHECKING:
+    from cudf.core.column import NumericalColumn
+
+
+@dataclass
+class GatherMap:
+    """A witness to the validity of a given column as a gather map.
```
shwina, line 16: This is mostly a question about terminology and semantics. Admittedly I haven't come across "witness" types like this before but perhaps they are used in other languages/contexts:

What does it mean for a `GatherMap` to be a _witness_, versus a type _composed_ of three things: an integer column of indexes, the number of rows in a target column, and a nullify parameter?

wence-: The tuple type `(column_indices, rows_in_target, nullify)` doesn't carry the extra semantic information that `column_indices` are valid for gathering (with given `nullify` parameter) from a column with `rows_in_target` rows.

The extra piece of information you need is a proof that the indices are valid (computed by doing a reduction to check that they are in bounds). So we need to add some extra information to the tuple type to carry around this "proof" information. The `GatherMap` type is that information. It is a witness to the statement that "this tuple of indices, rows, nullify" is valid. You can think of this type as a new type or a pair of the original objects and a proof of this statement.

The terminology comes from [logic](https://en.m.wikipedia.org/wiki/Witness_(mathematics)), formally if you have some existentially qualified proposition: $\exists x \in X . \phi(x)$ then a _witness_ is some (specific) $t \in X . \phi(t)$. AKA, "you claim there is such a thing, show me it.". In this case you can think of the set $X$ as the set of all indices that are valid for gathering `n` rows (with specified `nullify` flag), and the `GatherMap` type is giving you one such $x \in X$.

The use here is I suppose a little colloquial but reasonably common in the proof assistant literature.

vyasr: https://wiki.haskell.org/Type_witness

wence-: That is not quite this usage. Here I'm using witness in the sense of a value + proof of some restriction. e.g., in Haskell-ese:

```haskell

data NonEmptyList a = NonEmpty a [a]

head :: NonEmptyList a -> a
head (NonEmpty x _) = x

-- rather than the partial function
head' :: [a] -> Maybe a
head' (x:_) = Just x
head' _ = Nothing
```
So here `NonEmptyList` is a pair of a list, plus a proof that the list is non-empty. So the type witnesses the statement that the list is non-empty.

shwina: Would it be equivalent "proof" if the `GatherMap` constructor did this validation? Then, the fact that you were able to construct a `GatherMap` instance is proof enough that it is valid. If that's way off, sorry! Still trying to wrap my head around this

wence-: That's what the `as_gather_map` function does

bdice: Performing validation in the constructor would be opposed to “parse don’t validate”. If I am understanding correctly, you should be able to construct a GatherMap without triggering validation if you know (by some prior knowledge) that the inputs are valid. That’s where the validating factory function comes in. Is that right?

wence-: I guess it's a bit of a philosophical distinction in python (In Haskell data type constructors don't _do_ anything, so you need to use a function to do validation plus production of a datatype, in Python your init method can do arbitrary validation).

There are two things here: what do you know when you have a `GatherMap` object and how are you allowed to construct one? What you know (or at least, what the API contract promises) is that if you hold in your hand a `GatherMap(column, nrows, nullify)` object, then it is suitable for gathering a table with `nrows` rows. The idea being that by providing this object to an API that asks for a suitable gather map, you promise "yes, it is fine" (and, moreover, can distinguish that explicit promise from the implicit one that the low-level gather APIs require that you've provided valid input).

As to construction. There's nothing stopping you from just calling the constructor (I can't hide it from public view) and hence (in effect) asserting by appeal to authority that the `GatherMap` you build is valid. So my idea was to make the construction explicit in the sense that the person constructing the gather map should think about that step. So the module offers an `as_gather_map` function to do this. The type is not very different from the gather map constructor:

```python
def as_gather_map(column, nrows, nullify, check_bounds) -> GatherMap:
    ...

GatherMap(column, nrows, nullify) -> GatherMap
```

The small (possible) advantage of the separate function is that I think it makes it slightly clearer that the bounds checking (if any) is performed in construction (if I have a `GatherMap` with a `check_bounds=True` slot, does that mean that I check the bounds every time it is used in a new context, or is it as before?).

TBH it might mostly be a personal style preference here: I tend to prefer cheap-to-construct dataclass-style objects and functional interfaces to expensive object construction.

python/cudf/cudf/core/algorithms.py
```diff
@@ -4,7 +4,7 @@
 import cupy as cp
 import numpy as np
 
-from cudf.core.column import as_column
+from cudf.core import copy_types as ct
```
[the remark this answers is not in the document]
wence-: Everywhere, and also with `indexing_utils`?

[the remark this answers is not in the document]
bdice: I think so? Depends on the scope of the change. It's fine to separate this into another PR. Honestly I hadn't seen this pattern before so I don't know how big "everywhere" is. If it's prevalent in cudf code then maybe it's fine to keep, I just dislike using abbreviations for internal names.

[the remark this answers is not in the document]
wence-: I meant the ones I've introduced. I do `from cudf.core import indexing_utils as iu` and `from cudf.core import copy_types as ct`. But I can back out to qualified unabbreviated names easily

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,250 @@
...
+from cudf.api.types import (
+    _is_scalar_or_zero_d_array,
+    is_bool_dtype,
+    is_integer,
+    is_integer_dtype,
+)
+from cudf.core import copy_types as ct
+
+
+# Poor man's algebraic data types
```
bdice, line 20: Let's use another phrase or just drop the phrase to avoid being too colloquial. Also, what does it mean to have "algebraic data types"? I am familiar with the terms "sum types" or "product types" but I don't see the connection here.
```suggestion
# Algebraic data types
```

bdice: Is the "algebraic" part that `IndexingSpec` is a sum type (variant of index-like types)?

python/cudf/cudf/core/series.py
```diff
@@ -1306,6 +1301,41 @@ def map(self, arg, na_action=None) -> "Series":
             result = self.apply(arg)
         return result
 
+    def _getitem_preprocessed(
+        self,
+        spec: iu.IndexingSpec,
+    ) -> Union[Self, ScalarLike]:
+        """Get subset of entries given structured data
```
bdice, line 1308: ...and similar in other docstrings. Always end the summary line with a period.
```suggestion
        """Get subset of entries given structured data.
```

python/cudf/cudf/core/series.py
```diff
@@ -1306,6 +1301,41 @@ def map(self, arg, na_action=None) -> "Series":
...
+
+        Parameters
+        ----------
+        spec:
```
bdice, line 1312: Use numpydoc style (and similarly in other docstrings). https://numpydoc.readthedocs.io/en/latest/format.html#parameters
```suggestion
        spec : IndexingSpec
```

python/cudf/cudf/tests/test_indexing.py
```diff
@@ -1898,24 +1897,25 @@ def test_iloc_column_boolean_mask_issue_13265():
...
     # https://github.com/rapidsai/cudf/issues/13273
     df = pd.DataFrame(np.arange(4).reshape(2, 2))
     cdf = cudf.from_pandas(df)
 
     expect = df.iloc[:, [0, 1, 0]]
-    actual = cdf.iloc[:, [0, 1, 0]]
-    assert_eq(expect, actual)
+    with pytest.raises(NotImplementedError):
+        actual = cdf.iloc[:, [0, 1, 0]]
+        assert_eq(expect, actual)
```
bdice, line 1909: Does the `assert_eq` raise, or does the `iloc` raise? The `pytest.raises` should be narrowly scoped to the line that raises. If we're solely testing for an error in cudf (due to repeated columns being unsupported), it doesn't make a lot of sense to compute the `expect` value and try to assert things about it.

bdice (changes requested):
Thanks @wence-! I have some comments, as well as the offline discussion about how to use a constructor/factory here for validation.

python/cudf/cudf/tests/test_indexing.py
wence-: it's the `iloc`. I will remove the other line.

python/cudf/cudf/core/series.py
wence-: I can never get this right. TBH, I would prefer to not type the parameters in the docstring (since they can get out of date with the types in the function) and our API docs show the types.

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1725,37 +1745,113 @@ def hash_values(self, method="murmur3", seed=None):
...
+                    start=lo, stop=hi, step=step, name=index.name
+                )
+            return result
+        if start < 0:
+            start = start + num_rows
+
+        # Decreasing slices that terminates at -1, such as slice(4, -1, -1),
+        # has end index of 0, The check below makes sure -1 is not wrapped
+        # to `-1 + num_rows`.
+        if stop < 0 and not (stride < 0 and stop == -1):
```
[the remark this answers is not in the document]
wence-: This is just migrating code. The logic here is right, but the comment is misleading.

Suppose we start with `x = slice(4, None, -1)`. Then when we do `start, stop, step = x.indices(10)` we get `(4, -1, -1)`. it is this triple that the comment is referring to. i.e. by the time we are here, one level of wrapping has _already_ been applied by python, so we have to ensure we don't wrap again.

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1725,37 +1745,113 @@ def hash_values(self, method="murmur3", seed=None):
...
         )
 
+    def _slice(self, arg: slice, keep_index=True) -> Self:
+        """Slice a frame
+
+        Parameters
+        ----------
+        arg
+            The slice
+        keep_index
```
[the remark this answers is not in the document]
wence-: Would prefer to just add the `bool` type to the keep_index argument and avoid the duplication of information.

python/cudf/cudf/core/indexed_frame.py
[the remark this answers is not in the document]
wence-: Done.

python/cudf/cudf/core/dataframe.py
```diff
@@ -57,7 +57,13 @@
     is_string_dtype,
     is_struct_dtype,
 )
-from cudf.core import column, df_protocol, reshape
+from cudf.core import (
+    column,
+    copy_types as ct,
+    df_protocol,
+    indexing_utils as iu,
```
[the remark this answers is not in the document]
wence-: Done (I now import the `GatherMap` and so forth names directly).

python/cudf/cudf/core/copy_types.py
wence-: After discussion we've gone with:

1. `GatherMap.__init__(Any, nrows: int, nullify: bool)` _constructor_ takes anything, converts to column and does validation (no opt out)
2. Class method `GatherMap.from_column_unchecked(Column, nrows: int, nullify: bool)` believes you and doesn't check anything.

shwina (approved):
This is great! Thanks @wence- !

python/cudf/cudf/core/copy_types.py
```diff
@@ -13,128 +14,161 @@
...
+        IndexError
+            If the column is of unsuitable dtype. Or does not have the
+            requested number of rows.
+        """
+        self.column = cudf.core.column.as_column(column)
+        self.nrows = nrows
+        if self.column.dtype.kind != "b":
+            raise IndexError("Boolean mask must have bool dtype")
+        if len(column) != nrows:
+            raise IndexError(
```
bdice, line 144: I vote for these to be `TypeError` and `ValueError`, respectively. Similarly for other classes in this file.

The docs for `IndexError` say that if an index is not an integer, `TypeError` is raised.

My preference for a `ValueError` is that this occurs during the construction of a gather map with an invalid value, whereas typically I only see `IndexError` raised when an out-of-bounds access is being performed. Here, the access never actually occurs because of the validation. If the gather could attempt a disallowed access, then perhaps `IndexError` would be suitable for that case.

https://docs.python.org/3/library/exceptions.html#IndexError

python/cudf/cudf/core/copy_types.py
```diff
@@ -13,128 +14,161 @@
...
+        """A representation of a column as a boolean mask.
+
+        This augments the column with information that it is valid as a
+        boolean mask for columns with a given number of rows
+
+        Parameters
+        ----------
+        column
+            The data to turn into a column to then verify
+        nrows
```
bdice, line 123: Why does this class need the `nrows` parameter? We know the number of rows must be `len(column)` or else it's invalid. It doesn't seem useful to require that input, we should just use `len(column)`. This is especially true for the unchecked form.

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1031,11 +1033,10 @@ def interpolate(
             if isinstance(data._index, cudf.RangeIndex)
             # TODO: This should be a scatter, avoiding an argsort.
```
bdice, line 1034: Is it easy to just make that fix to use a scatter here? Or are there new APIs/bindings that would be needed to do so?

python/cudf/cudf/core/join/join.py
```diff
@@ -313,11 +313,10 @@ def _sort_result(self, result: cudf.DataFrame) -> cudf.DataFrame:
             by.extend([result._data[col.name] for col in self._right_keys])
         if by:
             to_sort = cudf.DataFrame._from_data(dict(enumerate(by)))
-            sort_order = ct.as_gather_map(
-                to_sort.argsort(),
+            sort_order = GatherMap.from_column_unchecked(
+                cudf.core.column.as_column(to_sort.argsort()),
```
bdice, line 317: Does this need the same TODO about scattering as above? (Or make the change to use a scatter if that's correct.)

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -0,0 +1,243 @@
...
+        raise TypeError(
+            "Column indices must be integers, slices, "
+            "or list-like of integers"
+        )
+    if scalar:
+        assert (
+            len(column_names) == 1
+        ), "Scalar column indexer should not produce more than one column"
+
+    return (rows, (scalar, column_names))
```
bdice, line 174: Parens are extraneous here, I think.
```suggestion
    return rows, (scalar, column_names)
```

python/cudf/cudf/core/indexed_frame.py
wence-: scatter doesn't exist at the frame level yet, so this would be a column-by-column scatter and then `_from_data`. I am inclined to do that separately.

python/cudf/cudf/core/join/join.py
wence-: This one is a sort-by-key for which there is an open issue, #13557

python/cudf/cudf/core/copy_types.py
wence-: I guess I don't need to store it. The question of requiring it as input is to do with whether you find out early or late that your mask is the wrong length. If you don't require the length of the target (to-be-masked) frame as input, then you get a libcudf `RuntimeError` when `apply_boolean_mask` is called.

wence-: Note as well, because we don't have dependent types, that the consumers of `BooleanMask` and `GatherMap` do some extra (cheap) runtime validation:

e.g. in `IndexedFrame._gather`
```
    def _gather(
        self,
        gather_map: GatherMap,
        keep_index=True,
    ):
        """Gather rows of frame specified by indices in `gather_map`.

        Maintain the index if keep_index is True.

        This function does no expensive bounds checking, but does
        check that the number of rows of self matches the validated
        number of rows.
        """
        if not gather_map.nullify and len(self) != gather_map.nrows:
            raise IndexError("Gather map is out of bounds")
```

python/cudf/cudf/core/copy_types.py
wence-: I think I picked `IndexError` because otherwise I need to catch `ValueError` and raise `IndexError` in (for example) `DataFrame.take` and `xxx.iloc[out-of-bounds-index]`. I can do that, but potentially it hides other problems

python/cudf/cudf/core/copy_types.py
```diff
@@ -13,128 +14,161 @@
...
-
     #: The gather map
     column: "NumericalColumn"
     #: The number of rows the gather map has been validated for
     nrows: int
     #: Was the validation for nullify=True?
     nullify: bool
 
+    def __init__(self, column: Any, nrows: int, *, nullify: bool):
+        """A representation of a column as a gather map.
```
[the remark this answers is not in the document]
wence-: Done.

python/cudf/cudf/core/copy_types.py
wence-: removed storing `nrows`.

python/cudf/cudf/core/copy_types.py
wence-: But changed the first to TypeError.

bdice: @wence- also mentioned offline that this aligns with pandas. Please align with pandas, I hadn't considered the impact there.

bdice (approved):
Approving to unblock. Thanks for your work on this @wence-, the result is quite nice.

wence-:
Thanks everyone!

PR #13538, opened 2023-06-08
Update to CMake 3.26.4

vyasr:
Updates minimum required CMake version to 3.26.4. This will allow us to take advantage of a variety of new features in the last three minor releases, particularly with respect to better CUDA and Python support. See https://github.com/rapidsai/rapids-cmake/issues/315 for more information.

### DIFF
--- a/conda/recipes/cudf_kafka/conda_build_config.yaml
+++ b/conda/recipes/cudf_kafka/conda_build_config.yaml
@@ -6,3 +6,6 @@ cxx_compiler_version:
 
 sysroot_version:
   - "2.17"
+
+cmake_version:
+  - ">=3.26.4"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,7 +73,7 @@ Compilers:
 
 * `gcc` version 9.3+
 * `nvcc` version 11.5+
-* `cmake` version 3.23.1+
+* `cmake` version 3.26.4+
 
 CUDA/GPU:
 
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -13,7 +13,7 @@ dependencies:
 - botocore>=1.24.21
 - c-compiler
 - cachetools
-- cmake>=3.23.1,!=3.25.0
+- cmake>=3.26.4
 - cubinlinker
 - cuda-python>=11.7.1,<12.0
 - cuda-sanitizer-api=11.8.86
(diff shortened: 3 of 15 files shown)


ttnghia (approved):
Is there any specific reason why this upgrade is needed?

vyasr:
> Is there any specific reason why this upgrade is needed?

Good question. We want to upgrade beyond 3.23 because there are features in newer versions that we could take advantage of mainly with improvements to FindCUDAToolkit and FindPython as well as better CUDA support in other areas (e.g. the new built-in `CMAKE_CUDA_ARCHITECTURES=native` option).

ttnghia:
Then can you update that into the PR description, please?

ajschmidt8 (approved):

vyasr:
Sure! Done.

ttnghia:
Although this is not merged, upstream changes have been propagated:
```
[INFO]      [exec] CMake Error at build/_deps/kvikio-src/cpp/CMakeLists.txt:15 (cmake_minimum_required):
[INFO]      [exec]   CMake 3.26.4 or higher is required.  You are running version 3.23.3
```

vyasr:
> Although this is not merged, upstream changes have been propagated:
> 
> ```
> [INFO]      [exec] CMake Error at build/_deps/kvikio-src/cpp/CMakeLists.txt:15 (cmake_minimum_required):
> [INFO]      [exec]   CMake 3.26.4 or higher is required.  You are running version 3.23.3
> ```

Where are you seeing this? Locally? It shouldn't be an issue in CI anywhere because the conda environments and pip builds will just install the latest CMake from conda-forge/PyPI. If it's local you need to update your local env.

ttnghia:
Yes I see it locally. But it doesn't matter to me since I've updated my local CMake.

conda/recipes/cudf_kafka/conda_build_config.yaml
```diff
@@ -6,3 +6,6 @@ cxx_compiler_version:
 
 sysroot_version:
   - "2.17"
+
+cmake_version:
```
bdice, line 10: I tend to avoid using `conda_build_config.yaml` for defining a variable when the version is only specified once in a recipe. I see the value of being consistent between `cudf_kafka` and other recipes, though.

bdice (approved):
Thanks! Can we merge this, or is there a list of downstream PRs that we need to merge first?

conda/recipes/cudf_kafka/conda_build_config.yaml
vyasr: Yeah I made this change since it then made it easily to do a batch upgrade across RAPIDS when everything is consistent. Could always add one more file to sed though.

vyasr:
Nope this is mergeable now.

PR #11613, opened 2022-08-27
Generic type casting to support the new nested JSON reader

elstehle:
## Description
This PR unifies the type casting logic across the existing JSON lines reader, the CSV reader, and the new nested JSON reader. 

_I'm taking over this excellent piece of work from @vuule, who did all the heavy lifting!_ 🙏  I just adapted the parsing of string values for things like escape sequences, optionally stripping off quotation characters, etc., to meet all the requirements for the new nested JSON parser. 

## Supported escape sequences:
We decided to follow [ECMA-404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf#page=12) and [RFC 7159](https://www.ietf.org/rfc/rfc7159.txt). 
 
```
\" represents the quotation mark character (U+0022).
\\ represents the reverse solidus character (U+005C).
\/ represents the solidus character (U+002F).
\b represents the backspace character (U+0008).
\f represents the form feed character (U+000C).
\n represents the line feed character (U+000A).
\r represents the carriage return character (U+000D).
\t represents the character tabulation character (U+0009).
\uDDDD, where `D` is a hex digit 0-9, a-f, A-F, for code points on the MBP
\uDDDD\uDDDD, where `D` is a hex digit 0-9, a-f, A-F, representing UTF-16 surrogate pairs for remaining unicode code points
```

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13539, opened 2023-06-08
Add unicode encoding for string columns in JSON writer 

karthikeyann:
## Description
Bug: In the JSON benchmarks, output string  column size differ from input string column size for json reader benchmarks because it uses JSON writer to create benchmark input. This bug happens because JSON writer does not encode unicode characters to unicode hex format.

Reference: JSON reader https://github.com/rapidsai/cudf/pull/11613 follows [ECMA-404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf#page=12). The same rules are used while doing unicode encoding of string columns.

Unicode encoding for utf-8 characters are added for string columns in JSON writer. 
UTF-8 char array is converted to utf8 char, then converted to unicode codepoint. 
This codepoint is converted to utf-8 for Basic Multilingual Plane (U+0000 through U+FFFF), and to  UTF-16 surrogate pair for Supplementary Plane (U+10000 through U+10FFFF).

- [x] UTF-8, UTF-16 Encoding
- [x] Unit tests

CSV writer also could use this encoding.

## Checklist


cpp/src/io/json/write_json.cu
```diff
@@ -86,6 +86,52 @@ struct escape_strings_fn {
...
+
+  __device__ void write_utf8_codepoint(uint32_t codepoint, char*& d_buffer, offset_type& bytes)
+  {
+    if (d_buffer) {
+      d_buffer[0] = '\\';
+      d_buffer[1] = 'u';
+      d_buffer[2] = nibble_to_hex((codepoint >> 12) & 0x0F);
+      d_buffer[3] = nibble_to_hex((codepoint >> 8) & 0x0F);
+      d_buffer[4] = nibble_to_hex((codepoint >> 4) & 0x0F);
+      d_buffer[5] = nibble_to_hex((codepoint)&0x0F);
```
karthikeyann, line 102: Can we use intrinsics?
```cuda
  void __device__ bytes_to_hex(uint16_t codepoint, char* d_buffer)
  {
    auto nibble_in_bytes =  __byte_perm(codepoint, (codepoint >> 4), 0x0415) & 0x0f0f'0f0f;
    auto nibbles_to_ascii = [](auto bytes) {
      // int + '0' + (int > 9) * ('a' - '0' - 10)
      return bytes + 0x3030'3030 + (__vcmpgtu4(bytes, 0x0909'0909) & 0x2727'2727);
    };
    auto nibble_as_ascii = nibbles_to_ascii(nibble_in_bytes);
    memcpy(d_buffer, &nibble_as_ascii, 4);
  }
```

```suggestion
      bytes_to_hex(codepoint, d_buffer+2);
```

karthikeyann: It barely makes any difference. Improved only 3us on avg, 5us on median. Besides, this contributes to only 0.1% of total runtime. so, skipping this idea.
```
INTRINSICS:
 Time (%)  Total Time (ns)  Instances    Avg (ns)      Med (ns)     Min (ns)    Max (ns)   StdDev (ns)   Style                        Range                      
      0.1         11638826        110      105807.5       81410.5       68185      156432      33467.6  PushPop  libcudf:get_escaped_strings::get_escaped_strings
      0.1          6358774         40      158969.4      158641.0      150164      169667       4462.8  PushPop  libcudf:operator()::string_view                 

BEFORE:
 Time (%)  Total Time (ns)  Instances    Avg (ns)      Med (ns)     Min (ns)    Max (ns)   StdDev (ns)   Style                        Range                      
      0.2         11905546        110      108232.2       86092.0       69701      163563      33932.9  PushPop  libcudf:get_escaped_strings::get_escaped_strings
      0.1          6441591         40      161039.8      159258.0      153868      177243       5139.8  PushPop  libcudf:operator()::string_view                 
```

davidwendt (commented):
LGTM

cpp/src/io/json/write_json.cu
karthikeyann: @davidwendt any opinion on using intrinsics.

davidwendt: Since the intrinsics are not significantly faster, I'd prefer the code be easier to read, maintain. 
Very cool though.

cpp/src/io/json/write_json.cu
```diff
@@ -86,6 +86,52 @@ struct escape_strings_fn {
...
+      d_buffer[6]   = '\\';
+      d_buffer[7]   = 'u';
+      d_buffer[8]   = nibble_to_hex((hex_low >> 12) & 0x0F);
+      d_buffer[9]   = nibble_to_hex((hex_low >> 8) & 0x0F);
+      d_buffer[10]  = nibble_to_hex((hex_low >> 4) & 0x0F);
+      d_buffer[11]  = nibble_to_hex((hex_low)&0x0F);
+      d_buffer += 12;
+    } else {
+      bytes += 12;
+    }
```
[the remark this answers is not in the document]
karthikeyann: updated.

cpp/tests/io/json_writer.cpp
```diff
@@ -517,4 +517,40 @@ TEST_F(JsonWriterTest, StructAllNullCombinations)
...
+TEST_F(JsonWriterTest, Unicode)
+{
+  //                                       UTF-8,                      UTF-16
+  cudf::test::strings_column_wrapper col1{"\"\\/\b\f\n\r\t", "ராபிட்ஸ்", "$€𐐷𤭢", "C𝞵𝓓𝒻"};
+  // Unicode
+  // 0000-FFFF     Basic Multilingual Plane
+  // 10000-10FFFF  Supplementary Plane
+  cudf::test::strings_column_wrapper col2{
+    "CႮ≪ㇳ䍏凹沦王辿龸ꁗ믜스폶ﴠ",  //  0000-FFFF
+    "𐀀𑿪𒐦𓃰𔙆 𖦆𗿿𘳕𚿾[↳] 𜽆𝓚𞤁🄰",                            // 10000-1FFFF
```
davidwendt, line 529: Is that a box kite?

bdice: 🤯

cpp/src/io/json/write_json.cu
```diff
@@ -86,6 +86,37 @@ struct escape_strings_fn {
       bytes += cudf::strings::detail::bytes_in_char_utf8(chr);
   }
 
+  __device__ inline char nibble_to_hex(uint8_t nibble) const
```
bdice, line 89: We reinvent functions similar to this one several times. Is reuse possible?

https://github.com/rapidsai/cudf/blob/deec3f8f981fd89f1aa46c6aea3714fd7c7355b9/cpp/src/strings/convert/convert_urls.cu#L60

https://github.com/rapidsai/cudf/blob/deec3f8f981fd89f1aa46c6aea3714fd7c7355b9/cpp/src/strings/convert/convert_hex.cu#L135

https://github.com/rapidsai/cudf/blob/deec3f8f981fd89f1aa46c6aea3714fd7c7355b9/cpp/include/cudf/detail/utilities/hash_functions.cuh#L170

bdice (approved):
Changes look good. We should investigate what level of reuse is possible for functions converting between string representations of hexadecimal values and integer representations.

cpp/src/io/json/write_json.cu
karthikeyann: `uint32ToLowercaseHexString` is interesting.  It's similar algorithm to instrinsics suggestion but without comparison (comparison achieved adding 6 & ANDing higher order bit). But uint32 is not required here. Only uint16 is required here.
Reusing `byte_to_hex` is possible. IMO, while working directly with nibbles and bytes, I prefer hex usage instead of numbers `16`, division and subtractions.

PR #13543, opened 2023-06-09
Use rapids_init_cuda_runtime.

bdice:
## Description
This PR updates libcudf to use `rapids_init_cuda_runtime`.

## Checklist

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -141,6 +141,10 @@ rapids_cmake_support_conda_env(conda_env MODIFY_PREFIX_PATH)
 
 # ##################################################################################################
 # * compiler options ------------------------------------------------------------------------------
+
+# CUDA runtime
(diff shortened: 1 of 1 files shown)


cpp/CMakeLists.txt
```diff
@@ -710,18 +714,6 @@ if(TARGET conda_env)
   target_link_libraries(cudf PRIVATE conda_env)
 endif()
 
-if(CUDA_STATIC_RUNTIME)
```
robertmaynard, line 713: We will still need
```cmake
if(CUDA_STATIC_RUNTIME AND NOT BUILD_SHARED_LIBS )
  # Make sure to export to consumers what runtime we used
  target_link_libraries(cudf PUBLIC CUDA::cudart_static)
endif()

since when doing static linking libcudf won't have linked to the cuda runtime and so we need to tell the consumer that is the intent.

robertmaynard (changes requested):

vyasr:
Closing in favor of #14833

PR #13545, opened 2023-06-09
Enable nested types for `lists::concatenate_list_elements`

ttnghia:
When `lists::concatenate_list_elements` was implemented, it could not handle nested structs in the related APIs. Now nested structs can be properly processed thus this PR enables fully nested type input for that API.

### DIFF
--- a/cpp/src/lists/combine/concatenate_list_elements.cu
+++ b/cpp/src/lists/combine/concatenate_list_elements.cu
@@ -256,24 +256,18 @@ std::unique_ptr<column> concatenate_list_elements(column_view const& input,
                                                   rmm::cuda_stream_view stream,
                                                   rmm::mr::device_memory_resource* mr)
 {
-  auto type = input.type();  // Column that is lists of lists.
-  CUDF_EXPECTS(
-    type.id() == type_id::LIST, "Input column must be a lists column.", std::invalid_argument);
+  CUDF_EXPECTS(input.type().id() == type_id::LIST,
+               "Input column must be a lists column.",
+               std::invalid_argument);
 
-  auto col = lists_column_view(input).child();  // Rows, which are lists.
-  type     = col.type();
-  CUDF_EXPECTS(
-    type.id() == type_id::LIST, "Rows of the input column must be lists.", std::invalid_argument);
-
-  col  = lists_column_view(col).child();  // The last level entries what we need to check.
-  type = col.type();
-  CUDF_EXPECTS(type.id() == type_id::LIST || !cudf::is_nested(type),
-               "Entry of the input lists column must be of list or non-nested types.");
+  auto const child = lists_column_view(input).child();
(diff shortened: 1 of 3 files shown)


cpp/src/lists/combine/concatenate_list_elements.cu
```diff
@@ -256,24 +256,18 @@ std::unique_ptr<column> concatenate_list_elements(column_view const& input,
...
-  CUDF_EXPECTS(
-    type.id() == type_id::LIST, "Rows of the input column must be lists.", std::invalid_argument);
-
-  col  = lists_column_view(col).child();  // The last level entries what we need to check.
-  type = col.type();
-  CUDF_EXPECTS(type.id() == type_id::LIST || !cudf::is_nested(type),
-               "Entry of the input lists column must be of list or non-nested types.");
+  auto const child = lists_column_view(input).child();
+  CUDF_EXPECTS(child.type().id() == type_id::LIST,
+               "Rows of the input column must be lists of lists.",
```
hyperbolic2346, line 265: This is really verifying the original message, right? "Rows of the input column must be lists". It expects lists of list, or does it require list of list of list?

hyperbolic2346 (approved):
Small nit on a CUDF_EXPECTS, but otherwise looks good.

cpp/src/lists/combine/concatenate_list_elements.cu
ttnghia: The original message was incorrect. "Rows of the input column must be lists of lists" is the most accurate one. They must be at least 2 level lists: `LIST<LIST<ANY_TYPE>>` (so they can be lists of lists of plain types, or lists of lists of any nested types). The first expect above checks for the first level and this expect checks for the second level.

hyperbolic2346: Yeah, but the input column row is a row inside a list column already, so saying it that way means `list<list<list<something>>>`. I think the original way was correct because you are already one level down inside the list if you are looking at a row of the input column. Maybe this indicates the original wording was awkward to begin with and we should rethink it.

ttnghia: Okay I see the problem. This error message should not be "rows" anymore. Thanks.

ttnghia, line 265:
```suggestion
               "Child of the input lists column must also be a lists column.",
```

ttnghia: Updated :)

nvdbaranec (commented):
A test with `List<Struct<List<X>>>` might be useful here.

ttnghia:
> A test with `List<Struct<List<X>>>` might be useful here.

Done.

nvdbaranec (approved):

PR #13548, opened 2023-06-09
Fix `dask_cudf` pytest failures for `pandas-2.0` upgrade

galipremsagar:
## Description
This PR fixes all `dask_cudf` side failures that happen due to `pandas-2.0` upgrade. The fixes are trivial to be broken down into separate PRs. 

- [x] `check_less_precise` is removed in `pandas-2.0`, since it is a parameter that we never supported and just had it for the sake of matching signature I removed it from all the methods.
- [x] Due to the removal of `StringIndex`, we had to perform some re-ordering of `if/elif` logic in `_nonempty_index`.
- [x] `dask_cudf.DataFrame.var` got `numeric_only` support.
- [x] `Series.count` doesn't have `skipna` support. Hence removed it from the call.

This PR fixes 56 pytest failures:

```
== 1100 passed, 13 skipped, 8 xfailed, 5 xpassed, 114 warnings in 57.10s ==
```

```
== 56 failed, 1044 passed, 13 skipped, 8 xfailed, 5 xpassed, 114 warnings in 73.73s (0:01:13) ==
```

## Checklist

### DIFF
--- a/python/dask_cudf/dask_cudf/core.py
+++ b/python/dask_cudf/dask_cudf/core.py
@@ -269,9 +269,12 @@ class DataFrame(_Frame, dd.core.DataFrame):
         dtype=None,
         out=None,
         naive=False,
+        numeric_only=False,
     ):
         axis = self._validate_axis(axis)
-        meta = self._meta_nonempty.var(axis=axis, skipna=skipna)
+        meta = self._meta_nonempty.var(
+            axis=axis, skipna=skipna, numeric_only=numeric_only
+        )
         if axis == 1:
             result = map_partitions(
                 M.var,
@@ -281,6 +284,7 @@ class DataFrame(_Frame, dd.core.DataFrame):
                 axis=axis,
                 skipna=skipna,
                 ddof=ddof,
+                numeric_only=numeric_only,
             )
             return handle_out(out, result)
         elif naive:
@@ -421,7 +425,7 @@ def _naive_var(ddf, meta, skipna, ddof, split_every, out):
 def _parallel_var(ddf, meta, skipna, split_every, out):
     def _local_var(x, skipna):
         if skipna:
-            n = x.count(skipna=skipna)
+            n = x.count()
             avg = x.mean(skipna=skipna)
         else:
             # Not skipping nulls, so might as well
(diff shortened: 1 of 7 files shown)


mroeschke (approved):

python/dask_cudf/dask_cudf/core.py
```diff
@@ -421,7 +425,7 @@ def _naive_var(ddf, meta, skipna, ddof, split_every, out):
 def _parallel_var(ddf, meta, skipna, split_every, out):
     def _local_var(x, skipna):
         if skipna:
-            n = x.count(skipna=skipna)
```
rjzamora, line 424: Hmm, did `count` ever accept `skipna`?

python/dask_cudf/dask_cudf/tests/test_reductions.py
```diff
@@ -71,10 +71,15 @@ def test_rowwise_reductions(data, op):
...
-        expected = getattr(pddf, op)(axis=1)
-        got = getattr(pddf, op)(axis=1)
-
-    dd.assert_eq(expected.compute(), got.compute(), check_exact=False)
+        expected = getattr(pddf, op)(numeric_only=True, axis=1)
+        got = getattr(pddf, op)(numeric_only=True, axis=1)
+
+    dd.assert_eq(
+        expected.compute(),
+        got.compute(),
```
rjzamora, line 82: Side Note: I don't love seeing so many `compute` calls within `assert_eq`, but they were already present, so no need to change anything here.

rjzamora (approved):
Dask changes seem fine (at least nothing is jumping out at me). Thanks for pushing on this @galipremsagar !

python/dask_cudf/dask_cudf/core.py
galipremsagar: Yup, `count` never had `skipna` and `count` only returns the count of non-NA values.

python/dask_cudf/dask_cudf/tests/test_reductions.py
galipremsagar: No worries, dropped it 👍

rjzamora: >dropped it 

Do tests pass with `compute` being dropped? My larger concern is that we often use compute when something isn't quite right (and the test doesn't pass without it). In that case, it is best to leave a comment explaining "why" it doesn't work (which is probably outside the scope of this PR).

Issue #12236, opened 2022-11-23
[FEA] Add a new cuDF option stable_sort that provides ordering guarantees for otherwise nondeterministic APIs


vyasr:
**Is your feature request related to a problem? Please describe.**
Currently various cuDF APIs (examples include `merge`, `groupby`, and `drop_duplicates`) rely on libcudf APIs that do not promise stable ordering. libcudf will not (and should not) ever be forced to make such guarantees since requiring ordering hamstrings potential optimizations and can always be achieved by calling code with the addition of suitable index columns and sorts. However, the requirements of cuDF Python are different. In particular, for pandas users transitioning to cuDF the lack of stability may be confusing at best and workflow-breaking at worst. Since there are multiple APIs that may exhibit this API divergence, and these APIs may be invoked many times in any particular piece of code, the responsibility should not fall to users to wrap those calls in appropriate sorting logic every time.

**Describe the solution you'd like**
cuDF should leverage the options framework introduced in #11193 to add a new option `stable_sort` that, when True, will change the behavior of all cuDF APIs to guarantee that the input order will be preserve. The default value should be `False` to force users to opt in to performance-inhibiting changes. The implementation of this option will depend on the API; for instance, with `merge` it will require consistent ordering in both tables, whereas for `drop_duplicates` it only affects one. I would recommend implementing the option as `_stable_sort` at first to indicate that it is internal, then adding support to one algorithm at a time, then exposing the option publicly as `stable_sort` once we have sufficient algorithmic coverage.

**Describe alternatives you've considered**
We have rejected similar requests in the past, in part because of the lack of clarity on how this should be handled in cuDF Python vs. libcudf, as well as because we did not wish to slow down all calls to these APIs by default. The use of `cudf.options` at the Python level provides an elegant and well-scoped solution to the problem that avoids these problems.

**Additional context**
Addressing this issue will close #1781 and #5286.

wence-:
FWIW, I think we should also consider the (ongoing) maintenance cost of implementing this scheme. We would double the test-time run in a naive setup since we'd need to do everything twice (once with, once without order preservation).

I read through a bunch of the past discussions, I didn't see really any compelling use-cases for order-preserving API that would not be better served by using algorithmic approaches at the user-level that don't rely on order-preservation. I agree that sometimes you have to think a bit harder, but ...

vyasr:
[...]

PR #13551, opened 2023-06-12
Introduce parity with pandas when `sort=False` in `Groupby`

galipremsagar:
## Description
This PR introduces preserving of the grouping key columns order in `groupby` when `sort=False`. This is not free, hence it is only achieved when pandas compatibility mode is enabled. `mode.pandas_compatible=True`.

## Checklist


bdice:
@galipremsagar Please feel free to comment/document as much as possible about the algorithmic side of this change. I have been thinking about this problem in the back of my mind since working on #11656 and I think we can achieve a relatively low-cost implementation of the algorithm for retaining the sort order if we’re smart about the usage of libcudf, perhaps with a slight bit of cuCollections magic with the row keys and a segmented sort. 🪄 If you can outline routes for improvement or feature requests for libcudf, please share!

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -568,6 +570,31 @@ def agg(self, func):
 
         if self._sort:
             result = result.sort_index()
+        else:
+            if cudf.get_option(
+                "mode.pandas_compatible"
+            ) and not libgroupby._is_all_scan_aggregate(normalized_aggs):
+                # Even with `sort=False`, pandas clearly documents and
+                # ensures that the ordering inside the group by columns
+                # is preserved(i.e., it is not non-deterministic).
```
shwina, line 579:
```suggestion
                # is preserved (i.e., it is deterministic).
```

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -568,6 +570,31 @@ def agg(self, func):
...
+                if not isinstance(result_index, cudf.MultiIndex):
+                    lcol, rcol = _match_join_keys(
+                        self.grouping.keys.drop_duplicates()._column,
+                        result_index._column,
+                        "inner",
```
shwina, line 584: Do we need an inner join, or a left join?

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -568,6 +570,31 @@ def agg(self, func):
...
+                    left_cols = list(
+                        self.grouping.keys.drop_duplicates()._data.columns
+                    )
+                    right_cols = list(result_index._data.columns)
+                    _, indices = libcudf.join.join(
+                        left_cols, right_cols, how="inner"
+                    )
+
+                new_index = result_index[indices]
+                result = result.reindex(new_index)
```
shwina, line 597: Rather than `reindex`, are we not able to reorder `result` using `indices`? Or does that not do the right thing?

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -568,6 +570,31 @@ def agg(self, func):
...
+                    _, indices = libcudf.join.join([lcol], [rcol], how="inner")
+                else:
+                    left_cols = list(
+                        self.grouping.keys.drop_duplicates()._data.columns
+                    )
+                    right_cols = list(result_index._data.columns)
+                    _, indices = libcudf.join.join(
+                        left_cols, right_cols, how="inner"
+                    )
+
```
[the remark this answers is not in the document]
galipremsagar: Unified this code, I added a todo to replace this repeating pattern. The same pattern is in the pandas 2.0 feature branch.

python/cudf/cudf/core/groupby/groupby.py
galipremsagar: We can directly use `take`, changed it.

python/cudf/cudf/core/groupby/groupby.py
galipremsagar: Switched to left join, yes left join is sufficient in this case.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -568,6 +570,29 @@ def agg(self, func):
 
         if self._sort:
             result = result.sort_index()
+        else:
+            if cudf.get_option(
+                "mode.pandas_compatible"
+            ) and not libgroupby._is_all_scan_aggregate(normalized_aggs):
+                # Even with `sort=False`, pandas clearly documents and
+                # ensures that the ordering inside the group by columns
+                # is preserved(i.e., it is deterministic).
```
shwina, line 579: Let's just use the same language Pandas does:


```suggestion
                # Even with `sort=False`, pandas guarantees that
                # groupby preserves the order of rows within each group.
```

shwina (approved):

galipremsagar:
> @galipremsagar Please feel free to comment/document as much as possible about the algorithmic side of this change. I have been thinking about this problem in the back of my mind since working on #11656 and I think we can achieve a relatively low-cost implementation of the algorithm for retaining the sort order while using a hash-based approach if we’re smart about the usage of libcudf, perhaps with a slight bit of cuCollections magic with the row keys and a segmented sort at the end before gathering. 🪄 I believe this is possible using only “fast” libcudf routines and never needing a join or other expensive calls in Python. If you can outline routes for improvement or feature requests for libcudf, please share in #12236!

Thanks @bdice ! I've added documentation for this change in cudf docs and posted in the issue #12236. Since I'm not sure about how trivial the change is on the libcudf side I'd merge this change and remove this workaround once libcudf side support is enabled.

PR #13553, opened 2023-06-12
Get rid of `cuco::pair_type` aliases

PointKernel:
## Description
This PR gets rid of the use of the deprecated `cuco::pair_type` alias. `cuco::pair` is all what we need.

## Checklist

### DIFF
--- a/cpp/src/join/join_common_utils.hpp
+++ b/cpp/src/join/join_common_utils.hpp
@@ -39,7 +39,7 @@ constexpr int DEFAULT_JOIN_BLOCK_SIZE = 128;
 constexpr int DEFAULT_JOIN_CACHE_SIZE = 128;
 constexpr size_type JoinNoneValue     = std::numeric_limits<size_type>::min();
 
-using pair_type = cuco::pair_type<hash_value_type, size_type>;
+using pair_type = cuco::pair<hash_value_type, size_type>;
 
 using hash_type = cuco::murmurhash3_32<hash_value_type>;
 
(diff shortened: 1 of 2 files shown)


bdice (approved):

cpp/src/join/join_common_utils.hpp
```diff
@@ -39,7 +39,7 @@ constexpr int DEFAULT_JOIN_BLOCK_SIZE = 128;
 constexpr int DEFAULT_JOIN_CACHE_SIZE = 128;
 constexpr size_type JoinNoneValue     = std::numeric_limits<size_type>::min();
 
-using pair_type = cuco::pair_type<hash_value_type, size_type>;
+using pair_type = cuco::pair<hash_value_type, size_type>;
```
ttnghia, line 42: I wonder why we ever need to use `cuco::pair`? Why not `std::pair` (or `thrust::pair`)? By doing so we will be less dependent on `cuco` and less be affected by future changes.

PointKernel: `thrust::pair` or `cuda::std::pair` doesn't guarantee alignment as we do for `cuco::pair`: https://github.com/NVIDIA/cuCollections/blob/291b81880b2d1e2ea1f3f3d8dd44a5f35420eb98/include/cuco/detail/pair.cuh#L181

ttnghia (approved):

cpp/src/join/join_common_utils.hpp
ttnghia: I really want to recommend changing its name into `cuco::aligned_pair` :smile:

PR #13562, opened 2023-06-13
Add pylibcudf subpackage with gather implementation

vyasr:
## Description

This PR is the first step in creating pylibcudf, a new Cython library composed of minimal wrappers around libcudf. The new library implements its own data structures analogous to libcudf data structures rather than wrapping libcudf data structures for direct use. These data structures are necessary in order to support data interop with other Python libraries (libcudf data structures own their memory). pylibcudf algorithms, on the other hand, are simple wrappers around libcudf algorithms that simply translate pylibcudf data types into libcudf data types then call the underlying libcudf code. This PR implements `gather` as a POC. Future PRs will use the data structures introduced here to add more algorithms.

## Checklist


wence-:
> The new library implements its own data structures analogous to libcudf data structures rather than wrapping libcudf data structures for direct use. These data structures are necessary in order to support data interop with other Python libraries (libcudf data structures own their memory)

Can you explain a bit more why this is needed? In an ideal world we would not have to main two sources of truth between what libcudf and pylibcudf think a column is (and what its state is).

Is cython a pragmatic choice here due to existing infrastructure and knowledge, or would nanobind/pybind be a preferred choice in isolation?

python/cudf/cudf/_lib/pylibcudf/column.pyx
```diff
@@ -0,0 +1,95 @@
...
+                for child in self.children:
+                    c_children.push_back(dereference(child.get_underlying()))
+
+            self._underlying.reset(
+                new column_view(
+                    self.data_type.c_obj, self.size, data, null_mask,
+                    self.null_count, self.offset, c_children
+                )
+            )
+        return self._underlying.get()
```
wence-, line 57: Is there a reason you store a `unique_ptr[column_view]` and then hand out the raw pointer, rather than the C++ version which just hands out a new `column_view` (no pointers anywhere) each time?

python/cudf/cudf/_lib/pylibcudf/types.pxd
```diff
@@ -0,0 +1,63 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+from libc.stdint cimport int32_t
+from libcpp cimport bool as cbool
+
+from cudf._lib.cpp.types cimport data_type, interpolation, null_policy, type_id
+
+ctypedef int32_t underlying_type_t_type_id
+cpdef enum TypeId:
+    EMPTY = <underlying_type_t_type_id> type_id.EMPTY
```
wence-, line 10: It does seem like we're fighting cython here rather.

python/cudf/cudf/_lib/types.pyx
```diff
@@ -94,6 +96,29 @@ SUPPORTED_NUMPY_TO_LIBCUDF_TYPES = {
     np.dtype("timedelta64[ns]"): TypeId.DURATION_NANOSECONDS,
 }
 
+SUPPORTED_NUMPY_TO_PYLIBCUDF_TYPES = {
```
wence-, line 99: Could we construct this table from the previous one?

wence- (commented):
A first cursory pass

vyasr:
> > The new library implements its own data structures analogous to libcudf data structures rather than wrapping libcudf data structures for direct use. These data structures are necessary in order to support data interop with other Python libraries (libcudf data structures own their memory)
> 
> Can you explain a bit more why this is needed? In an ideal world we would not have to main two sources of truth between what libcudf and pylibcudf think a column is (and what its state is).

Yup, this is definitely worth expanding upon. libcudf algorithms all operate on `cudf::[mutable_](table|column)_view` objects. The reason is that algorithms only need views on data, they don't need to take owning types like `cudf::column`. In practice, this also means that libcudf algorithms can operate on data owned by something other than cudf's owning types. For example, you could allocated some buffers manually using cudaMalloc APIs, then create a `cudf::column_view` from those buffers [by passing in the pointers to the constructor](https://github.com/rapidsai/cudf/blob/branch-23.08/cpp/include/cudf/column/column_view.hpp#L279), and then run algorithms on that view.

That explains why we _can_ use a different set of data structures than libcudf's. As for why we _should_, the reason is that libcudf's data structures all must own their data. For instance, `cudf::column` contains, [an `rmm::device_buffer` of data](https://github.com/rapidsai/cudf/blob/branch-23.08/cpp/include/cudf/column/column.hpp#L325), which in turn is an RAII-based object [whose underlying data pointer is private](https://github.com/rapidsai/rmm/blob/branch-23.08/include/rmm/device_buffer.hpp#L397) and is [deleted upon destruction](https://github.com/rapidsai/rmm/blob/branch-23.08/include/rmm/device_buffer.hpp#L232). There is no way to create a column that refers to data owned by another object. That is problematic for any Python API because we want to be able to work with any object that exposes the CUDA Array Interface, such as a cupy array. Not only should it be possible, but the API should ideally be largely transparent to whether data was originally allocated by libcudf or some other array-like object. While we have the ability to release memory allocated by a libcudf column into a set of raw buffers that we control, similar functionality is not guaranteed to exist for other objects like cupy arrays. Therefore, the best option we have for placing all of these on an equal footing is to wrap any allocated memory into a `memoryview`-like object (the new `gpumemoryview`) that manages ownership via standard Python ref-counting (allowing the original object to continue being the true owner, but having it be kept alive by the `gpumemoryview`), then create new Table/Column classes in Cython/Python that own these as their buffers. That way shared memory ownership between our types and other objects is no longer a problem.

Does that make sense?

> Is cython a pragmatic choice here due to existing infrastructure and knowledge, or would nanobind/pybind be a preferred choice in isolation?

That's a good question. I think there are tradeoffs here. There's definitely a pragmatic benefit to using cython, not only because of existing infra/knowledge but because if we use Cython the bulk of the code can be extracted from the bindings that we already have. In a vacuum, nanobind/pybind would probably be a better choice to due better intrinsic C++ support and the ability to write more idiomatic bindings using those APIs. We may also get slightly more performant bindings or more compact bindings (with respect to the size of the compiled binary), at least using nanobind. OTOH using Cython does have the advantage of allowing users to also accelerate code by actually writing in Cython. All the types will be `cdef` classes, so in theory Python overhead could be almost completely removed even at the level of user code, which could be beneficial if people are constructing a bunch of pylibcudf objects in loops etc. 

I think the "we already mostly have the Cython bindings" is the strongest argument to use Cython at this stage though. If we were to try to switch to pybind I think we'd have to think a bit more carefully about how it interacts with rmm's types (which we normally use in Cython too since they're `cdef` classes).

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pyx
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+
+cdef class gpumemoryview:
+    """Minimal representation of a memory buffer."""
+    def __init__(self, object obj):
+        try:
+            cai = obj.__cuda_array_interface__
```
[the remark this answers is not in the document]
vyasr: Indeed. For that purpose we may have to special case these functions to check the type and call the getter method instead to avoid exposing.

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pyx
```diff
@@ -0,0 +1,20 @@
...
+                "gpumemoryview must be constructed from an object supporting "
+                "the CUDA array interface"
+            )
+        self.obj = obj
+        # TODO: Need to respect readonly
+        self.ptr = cai["data"][0]
+
+    @property
+    def __cuda_array_interface__(self):
+        return self.obj.__cuda_array_interface__
```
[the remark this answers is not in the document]
vyasr: If you're effectively asking whether we could cache this, I think the answer is yes.

python/cudf/cudf/_lib/pylibcudf/table.pxd
```diff
@@ -0,0 +1,18 @@
...
+
+from libcpp.memory cimport unique_ptr
+
+from cudf._lib.cpp.table.table cimport table
+from cudf._lib.cpp.table.table_view cimport table_view
+
+
+cdef class Table:
+    # List[pylibcudf.Column]
+    cdef object columns
```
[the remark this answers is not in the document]
vyasr: You can't put Cython `cdef` classes into C++ containers unfortunately.

python/cudf/cudf/_lib/pylibcudf/table.pyx
```diff
@@ -0,0 +1,42 @@
...
+                c_columns.push_back(dereference(col.get_underlying()))
+
+            self._underlying.reset(new table_view(c_columns))
+
+        return self._underlying.get()
+
+    @staticmethod
+    cdef Table from_libcudf(unique_ptr[table] libcudf_tbl):
+        cdef vector[unique_ptr[column]] c_columns = move(
+            dereference(libcudf_tbl).release()
```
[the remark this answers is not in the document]
vyasr: Correct. All algorithms returning libcudf objects are wrapped in logic that immediately takes ownership of the data from libcudf.

python/cudf/cudf/_lib/pylibcudf/table.pyx
```diff
@@ -0,0 +1,42 @@
...
+        )
+
+        cdef vector[unique_ptr[column]].size_type i
+        return Table([
+            Column.from_libcudf(move(c_columns[i]))
```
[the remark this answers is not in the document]
vyasr: Yes.

python/cudf/cudf/_lib/pylibcudf/column.pyx
```diff
@@ -0,0 +1,95 @@
...
+        cdef DataType dtype = DataType.from_data_type(libcudf_col.get().type())
+        cdef size_type size = libcudf_col.get().size()
+        cdef size_type null_count = libcudf_col.get().null_count()
+
+        cdef column_contents contents = move(libcudf_col.get().release())
+
+        # Note that when converting to cudf Column objects we'll need to pull
+        # out the base object.
+        cdef gpumemoryview data = gpumemoryview(
+            DeviceBuffer.c_from_unique_ptr(move(contents.data))
```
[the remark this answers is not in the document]
vyasr: That is almost right. The order will be: 1) `Column` ref count goes to 0; 2) `Column` deleted, `Column.data` (`gpumemoryview`) ref count goes to 0; 3) `Column.data` deleted, Column.data.obj (`DeviceBuffer`) ref count goes to 0; 4) `Column.data.obj` is deleted, `Column.data.obj.c_obj` ref count goes to 0; 5) the c_obj is deleted.

[`DeviceBuffer.c_obj` is a `std::unique_ptr<rmm::device_buffer>`](https://github.com/rapidsai/rmm/blob/branch-23.08/python/rmm/_lib/device_buffer.pxd#L40), so that's the deallocation function that is eventually called to delete the data.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
```diff
@@ -0,0 +1,48 @@
...
+
+from .column cimport Column
+from .table cimport Table
+
+
+# Cython doesn't support scoped enumerations. It assumes that enums correspond
+# to their underlying value types and will thus attempt operations that are
+# invalid. This code will ensure that these values are explicitly cast to the
+# underlying type before casting to the final type.
+cdef cpp_copying.out_of_bounds_policy py_policy_to_c_policy(
```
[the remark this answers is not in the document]
vyasr: Yes we could inline this. Just would have to move it to the pxd.

python/cudf/cudf/_lib/pylibcudf/column.pyx
vyasr: Yeah, there are a couple of reasons (not necessarily great ones):
1. The way that Cython transpiles to C++ usually adds a bunch of extra intermediate assignments. In C++ if we have a method `cudf::foo(column_view x)` and do something like `column x = ...; cudf::foo(x->view())`, the compiler would typically elide the copy i.e. the temporary created by `x->view()` wouldn't be copied _again_ when passed as an argument to `cudf::foo` because the compiler would see that the temporary was being immediately consumed so it would reuse the object. However, with Cython it's likely that something like `cudf.foo(x.view())` will actually translate to something like `column_view y; ...; y = x->view(); ...; cudf::foo(y)`, and as a result copy elision will probably not occur. I chose the current approach because it guarantees not copying column views all over the place.
2. There are restrictions on what types of code can be run in `nogil` blocks. The `get_underlying` method needs to access the list (a Python object) of children in order to construct the `cudf::column_view`. That means that this function cannot be marked nogil (not quite true, you could mark it `nogil` and then use a `with gil` block inside it, but I didn't want to do anything that fancy yet without good reason). That in turn means that we can't call `get_underlying` directly inside the `with nogil` blocks that we typically use in the algorithms calling to libcudf code. That pretty much guarantees that we can't get copy elision working here, so using a pointer is pretty much the only way to guarantee reducing copies.

I say these aren't great reasons because this certainly gets back to the point about whether we are doing things to work around Cython limitations. That said, I'm not sure pybind11/nanobind would necessary fix this altogether, you might just trade them in for different problems (I haven't worked pybind in a couple years now, and I've never used nanobind, only seen its code).

python/cudf/cudf/_lib/pylibcudf/types.pxd
vyasr: Yes, although I believe this is fixed in Cython 3.

python/cudf/cudf/_lib/types.pyx
vyasr: Hmm maybe I didn't consider that. I'm not sure offhand how we'd go from libcudf types to pylibcudf types but it should be possible by accessing the underlying enum values.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -497,6 +548,26 @@ cdef class Column:
...
+            dtype=dtype,
+            size=col.size,
+            mask=as_buffer(col.mask.obj) if col.mask is not None else None,
+            offset=col.offset,
+            null_count=col.null_count,
+            children=tuple([
+                Column.from_pylibcudf_column(child)
+                for child in col.children
+            ])
+        )
```
wence-, line 569: What is your longer-term vision for how this "should" work? To my mind it seems like what we _want_ is for cudf python-exposed columns to own a pylibcudf column and just layer syntax sugar and (perhaps) pandas-specific behaviour on top.

I guess the challenge is how to do this migration piecemeal.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -428,6 +438,47 @@ cdef class Column:
...
+            # Where is the mask stored for categoricals?
+            mask = pylibcudf.gpumemoryview(self.base_mask)
+
+        cdef Column child_column
+        children = []
+        if col.base_children:
+            for child_column in col.base_children:
+                children.append(child_column.to_pylibcudf())
+
+        return pylibcudf.Column(
```
wence-, line 468: Presume this method is cheap enough that we don't really care about memoizing.

python/cudf/cudf/_lib/pylibcudf/column.pyx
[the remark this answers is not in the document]
wence-: Thanks.

python/cudf/cudf/_lib/pylibcudf/column.pxd
```diff
@@ -0,0 +1,32 @@
...
+    cdef:
+        # Core data
+        DataType data_type
+        size_type size
+        gpumemoryview data
+        gpumemoryview mask
+        size_type null_count
+        offset_type offset
+        # TODO: Is there a more efficient container I could use than a list
+        # here? For now defaulting to List[gpumemoryview]
```
shwina, line 23:
```suggestion
        # here? For now defaulting to List[Column]
```

shwina: A list is probably fine. We can't use a tuple as I'm pretty sure there are places in cuDF where we have to mutate the `children` attribute of a column.

python/cudf/cudf/_lib/pylibcudf/column.pxd
```diff
@@ -0,0 +1,32 @@
...
+        object children
+
+        # Internals
+        unique_ptr[column_view] _underlying
```
shwina, line 27: Does it really matter if this is a `unique_ptr[column_view]` v/s just `column_view`?

shwina: OK, I see your comment below (https://github.com/rapidsai/cudf/pull/13562/files#r1230200771)

shwina: We should probably document somewhere why we're doing it this way

python/cudf/cudf/_lib/pylibcudf/column.pxd
```diff
@@ -0,0 +1,32 @@
...
+
+    cdef column_view* get_underlying(self)
+
+    @staticmethod
+    cdef Column from_libcudf(unique_ptr[column] libcudf_col)
```
shwina, line 32: Do we want to consistently name these factories `from_libcudf`? In `DataType` for example the equivalent constructor is named `from_data_type`.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
```diff
@@ -0,0 +1,48 @@
...
+from cudf._lib.cpp.table.table_view cimport table_view
+
+from .column cimport Column
+from .table cimport Table
+
+
+# Cython doesn't support scoped enumerations. It assumes that enums correspond
+# to their underlying value types and will thus attempt operations that are
+# invalid. This code will ensure that these values are explicitly cast to the
+# underlying type before casting to the final type.
```
shwina, line 23: Likely there's going to be tons of places we have to do this. Perhaps we document this pattern in the dev guide instead?

python/cudf/cudf/_lib/column.pyx
```diff
@@ -497,6 +548,26 @@ cdef class Column:
             children=tuple(children)
         )
 
+    @staticmethod
+    def from_pylibcudf_column(
```
shwina, line 552:
```suggestion
    def from_pylibcudf(
```

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pyx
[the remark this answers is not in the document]
vyasr: Now using `functools.cached_property`.

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pxd
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+
+cdef class gpumemoryview:
```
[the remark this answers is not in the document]
vyasr: Is there any harm in using this name? It signals a clear intent, and since it's within pylibcudf it's unlikely that anyone will think of it as a generally usable utility. Keeping the same name will save some refactoring work if/when we do get to having a complete proposal.

python/cudf/cudf/_lib/pylibcudf/column.pxd
vyasr: Agreed. I'll stick to this for now.

python/cudf/cudf/_lib/pylibcudf/column.pyx
vyasr: I've added documentation of this behavior so that it's clear enough in case we want to reverse this decision later.

python/cudf/cudf/_lib/pylibcudf/column.pxd
vyasr: Added docs.

python/cudf/cudf/_lib/pylibcudf/column.pxd
vyasr: Good call. I think I went back and forth during development and didn't make them consistent. I will do so now.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -428,6 +438,47 @@ cdef class Column:
...
+
+        return pylibcudf.Column(
+            dtype,
+            self.size,
+            data,
+            mask,
+            # TODO: We may need to do some extra work to ensure that
+            # self.null_count is not None. However, that would be out of scope
+            # for this PR and would probably need to be resolved as part of the
+            # UNKNOWN_NULL_COUNT removal.
```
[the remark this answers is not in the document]
vyasr: Honestly I think we can just delete this comment without adding an assert. Python doesn't optimize away asserts unless it's run with `-O`, and we can make this assumption fairly safely at this point (otherwise we would be seeing test failures all over the place after the `UNKNOWN_NULL_COUNT` removal. Unless that makes you particularly uneasy I'd rather not add the assert (but if you do want to add, there's probably a better place to put it than `to_pylibcudf` since we'd want to assert this property more generally anywhere the null count is queried I assume).

python/cudf/cudf/_lib/pylibcudf/column.pxd
vyasr: I renamed the DataType method. Do you see any others? I think that was the only other one.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
vyasr: Added a note about this in the pylibcudf dev docs PR (#13639) and removed all instances of this comment from cudf.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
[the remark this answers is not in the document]
vyasr: Never mind, no move needed. Was thinking of a more specific edge case. inline here works fine.

python/cudf/cudf/_lib/types.pyx
vyasr: Done.

python/cudf/cudf/_lib/column.pyx
vyasr: I'd say let's put off memoizing for now since as you brought up in your comment above the long term vision is for a cudf's columns to own pylibcudf columns, and as such there won't be "conversion" in the long run so much as just accessing the owned object.

python/cudf/cudf/_lib/column.pyx
vyasr: Yes, that is exactly how I've been envisioning it. There's some more thought that will have to go into it since there's also cases where we might want to cache the pylibcudf Table representing a cudf Frame for calling functions on a libcudf::table, but we need a single source of truth for the ownership pathway (i.e. it makes more sense for each cudf  column to own a pylibcudf column than for each cudf Frame to own a list of pylibcudf columns due to the fact that we need to store additional metadata like dtypes per column because there isn't a 1-1 mapping between Python and C++ dtypes). We'll definitely have to spend some time getting that mapping right once we've done enough the migration. For now, though, I'm OK eating a bit of overhead to keep the migration simple and as minimally intrusive as possible as demonstrated in this PR.

python/cudf/cudf/_lib/pylibcudf/types.pyx
```diff
@@ -0,0 +1,32 @@
...
+    cpdef TypeId id(self):
+        return TypeId(self.c_obj.id())
+
+    cpdef int32_t scale(self):
+        return self.c_obj.scale()
+
+    @staticmethod
+    cdef DataType from_data_type(data_type dt):
+        # Spoof an empty data type then swap in the real one.
+        cdef DataType ret = DataType.__new__(DataType, TypeId.EMPTY)
```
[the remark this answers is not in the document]
vyasr: We have a `__cinit__` defined on this class, which will be called by `__new__`. And I think we want `__cinit__` rather than `__init__` here because we're initialize a C++ member (`self.c_obj`).

python/cudf/cudf/_lib/column.pyx
```diff
@@ -428,6 +438,47 @@ cdef class Column:
             offset,
             children)
 
+    cpdef pylibcudf.Column to_pylibcudf(self):
+        # TODO: Categoricals will need to be treated differently eventually.
+        # There is no 1-1 correspondence between cudf and libcudf for
+        # categoricals due to the ordering question.
```
wence-, line 444: Maybe reference some comment on the "ordering question" for those of us (like me) how don't know what it is.

wence- (commented):
A few more small comments, but stopping here because reading between the lines you've made some further updates that are not yet pushed.

python/cudf/cudf/_lib/pylibcudf/column.pyx
```diff
@@ -0,0 +1,95 @@
...
+                )
+
+        return Column(
+            dtype,
+            size,
+            data,
+            mask,
+            null_count,
+            # Initial offset when capturing a C++ column is always 0.
+            0,
```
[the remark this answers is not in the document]
vyasr: Yes, exactly. This difference is a consequence of the ownership semantics in pylibcudf. pylibcudf.Column is always a view into a set of `gpumemoryview`s of data. Each `gpumemoryview` is itself also a view into memory owned by some other object. Two `gpumemoryview`s could view the same data, and therefore two pylibcudf.Column objects could also view into the same data with different objects. For example slicing would produce a new Column that has an offset relative the original Column.

python/cudf/cudf/_lib/column.pyx
vyasr: Added more of a note.

python/cudf/cudf/_lib/pylibcudf/column.pxd
```diff
@@ -0,0 +1,32 @@
...
+        size_type null_count
+        offset_type offset
+        # TODO: Is there a more efficient container I could use than a list
+        # here? For now defaulting to List[gpumemoryview]
+        object children
+
+        # Internals
+        unique_ptr[column_view] _underlying
+
+    cdef column_view* get_underlying(self)
```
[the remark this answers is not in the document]
vyasr: We will eventually need it. I chose to only implement a minimal subset of methods in this PR to get gather working.

vyasr:
> A few more small comments, but stopping here because reading between the lines you've made some further updates that are not yet pushed.

Yup, just finished! Ready for another pass now.

wence- (approved):
Minor comments, but I think this looks in a good shape.

python/cudf/cudf/_lib/pylibcudf/column.pyx
vyasr: OK, I've actually decided to walk this back. Adding a little extra complexity to how the conversion works was enough to make it possible to return a new `column_view` by value each time this function is created, and to ensure that we could write Cython code in a way that favors copy elision.

python/cudf/cudf/_lib/column.pxd
```diff
@@ -22,10 +23,12 @@ cdef class Column:
         cdef object _data
         cdef object _mask
         cdef object _null_count
+        cdef object _distinct_count
```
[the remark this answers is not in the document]
vyasr: Yeah, somehow Cython hasn't errored on this before but it is necessary.

python/cudf/cudf/_lib/pylibcudf/column.pyx
```diff
@@ -0,0 +1,93 @@
...
+from .gpumemoryview cimport gpumemoryview
+from .types cimport DataType
+from .utils cimport int_to_bitmask_ptr, int_to_void_ptr
+
+
+cdef class Column:
+    """A container of nullable device data as a column of elements."""
+    def __init__(
+        self, DataType data_type not None, size_type size, gpumemoryview data,
+        gpumemoryview mask, size_type null_count, offset_type offset,
```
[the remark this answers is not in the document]
vyasr: I'd like to avoid putting any attributes into a `gpumemoryview` that don't exist in a Python [`memoryview`](https://docs.python.org/3/library/stdtypes.html#memoryview) so that if we do eventually try to align/merge them we won't have too much difficulty.

Regarding the offset applying to the mask and the data, yes it applies to both. That's the same behavior as in libcudf, so I don't know if there's any additional documentation to do here, but happy to add if you have thoughts.

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pyx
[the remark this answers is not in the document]
vyasr: This is addressed in https://github.com/rapidsai/cudf/pull/13562/commits/64eb29bfe5e615215e0cd1e538f946a525ec9b22. @wence- @shwina PTAL and let me know if my approach seems right here.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -438,7 +439,32 @@ cdef class Column:
...
+    # TODO: Consider whether this function should support some sort of `copy`
+    # parameter. Not urgent until this functionality is moved up to the Frame
+    # layer and made public. This function will also need to mark the
+    # underlying buffers as exposed before this function can itself be exposed
+    # publicly.  User requests to convert to pylibcudf must assume that the
+    # data may be modified afterwards.
+    cpdef pylibcudf.Column to_pylibcudf(self, mode: Literal["read", "write"]):
+        """Convert this Column to a pylibcudf.Column.
+
+        This function will generate a pylibcudf Column pointing to the same
```
wence-, line 451: Just to note that this pattern produces, I think, badly formatted docstrings in ipython (say), since the indent is not stripped from subsequent lines.
```suggestion
        """
        Convert this Column to a pylibcudf.Column.

        This function will generate a pylibcudf Column pointing to the same
```

python/cudf/cudf/_lib/column.pyx
```diff
@@ -546,12 +582,32 @@ cdef class Column:
             children=tuple(children)
         )
 
+    #  TODO: Actually support exposed data pointers.
     @staticmethod
     def from_pylibcudf(
         pylibcudf.Column col, bint data_ptr_exposed=False
     ):
+        """Create a Column from a pylibcudf.Column.
```
wence-, line 590:
```suggestion
        """
        Create a Column from a pylibcudf.Column.
```

python/cudf/cudf/_lib/pylibcudf/column.pyx
```diff
@@ -1,22 +1,45 @@
...
+from cudf._lib.cpp.types cimport offset_type, size_type
 
 from .gpumemoryview cimport gpumemoryview
 from .types cimport DataType
 from .utils cimport int_to_bitmask_ptr, int_to_void_ptr
 
 
 cdef class Column:
-    """A container of nullable device data as a column of elements."""
+    """A container of nullable device data as a column of elements.
```
wence-, line 17:
```suggestion
    """
    A container of nullable device data as a column of elements.
```

python/cudf/cudf/_lib/pylibcudf/column.pyx
```diff
@@ -30,32 +53,50 @@ cdef class Column:
         self.offset = offset
         self.children = children
 
-    cdef column_view* view(self):
+    cdef column_view view(self) nogil:
+        """Generate a libcudf column_view to pass to libcudf algorithms.
```
wence-, line 57:
```suggestion
        """
        Generate a libcudf column_view to pass to libcudf algorithms.
```

python/cudf/cudf/_lib/pylibcudf/column.pyx
```diff
@@ -30,32 +53,50 @@ cdef class Column:
...
-            )
-        return self._view.get()
+                    # Need to cast to Column here so that Cython knows that
+                    # `view` returns a typed object, not a Python object. We
+                    # cannot use a typed variable for `child` because cdef
+                    # declarations cannot be inside nested blocks (`if` or
+                    # `with` blocks) but without that we cannot declare it
+                    # inside the `with gil` block and it is erroneous to
+                    # declare a variable of a cdef class type in a `nogil`
+                    # context (which this whole function is).
```
wence-, line 84: Good explanation, also, ugh.

python/cudf/cudf/_lib/pylibcudf/column.pyx
```diff
@@ -30,32 +53,50 @@ cdef class Column:
...
+                    c_children.push_back((<Column> child).view())
+
+        return column_view(
+            self.data_type.c_obj, self.size, data, null_mask,
+            self.null_count, self.offset, c_children
+        )
 
     @staticmethod
     cdef Column from_libcudf(unique_ptr[column] libcudf_col):
+        """Create a Column from a libcudf column.
```
wence-, line 94:
```suggestion
        """
        Create a Column from a libcudf column.
```

et seq....

python/cudf/cudf/_lib/pylibcudf/copying.pyx
```diff
@@ -20,24 +17,44 @@ from .table cimport Table
...
     OutOfBoundsPolicy py_policy
 ) nogil:
+    """Convert a Cython policy the corresponding libcudf policy type."""
     return <cpp_copying.out_of_bounds_policy> (
         <underlying_type_t_out_of_bounds_policy> py_policy
     )
 
 
+# TODO: Is it OK to reference the corresponding libcudf algorithm in the
+# documentation? Otherwise there's a lot of room for duplication.
```
wence-, line 27: I think yes. We should figure out a way to cross-link in the generated docs...

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pyx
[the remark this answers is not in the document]
wence-: Looks correct to me. As noted there, we will need to think harder about ownership once pylibcudf objects become the default inside cudf.

python/cudf/cudf/_lib/pylibcudf/table.pyx
```diff
@@ -13,24 +13,41 @@ from .column cimport Column
 
 
 cdef class Table:
-    """A set of columns of the same size."""
-    def __init__(self, object columns):
+    """A set of columns of the same size.
```
wence-, line 16:
```suggestion
    """
    A list of columns of the same size.
```
Or collection (the columns need not be unique, and are ordered...)

wence- (approved):
Some minor docstring fixes, but otherwise still looks good.

python/cudf/cudf/_lib/column.pxd
```diff
@@ -22,10 +25,12 @@ cdef class Column:
...
 
     cdef column_view _view(self, size_type null_count) except *
     cdef column_view view(self) except *
     cdef mutable_column_view mutable_view(self) except *
+    cpdef pylibcudf.Column to_pylibcudf(self, mode: Literal["read", "write"])
```
shwina, line 33: What concepts in libcudf do `"read"` and `"write"` translate to? Also, is a boolean more appropriate than a literal?

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pyx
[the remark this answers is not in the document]
shwina: Can this bite us if the pointer gets changed from underneath us? I'm thinking situations like when `obj` is a `Series` and the column beneath that Series is changed.

python/cudf/cudf/_lib/pylibcudf/column.pyx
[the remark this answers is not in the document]
shwina: > So I guess one way to do it would be to construct the memoryview from ptr + offset, rather than ptr. Then one would remove the need (I think) for columns to have an offset at all

That would only work for fixed width data types, no?

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pxd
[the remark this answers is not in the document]
shwina: Nope - I think it's fine

shwina (approved):
This is looking good. Couple of minor questions but shouldn't block merging IMO

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pyx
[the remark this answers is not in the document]
vyasr: IIUC that sequence would require creating a new `gpumemoryview` from the new `Column` before any pylibcudf algorithms would be called, right?

python/cudf/cudf/_lib/pylibcudf/column.pyx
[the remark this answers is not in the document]
vyasr: Right. I think you could make it work for nested types but the math would become nontrivial.

python/cudf/cudf/_lib/column.pyx
vyasr: Hmm could you show me a situation where it causes problems? I think the first line of docstrings are handled as a special case to support this everywhere:
```
In [1]: def f(x):
   ...:     """Hello
   ...:     
   ...:     What does this look like.
   ...:     """
   ...:     pass
   ...: 

In [3]: f?
Signature: f(x)
Docstring:
Hello

What does this look like.
File:      ~/<ipython-input-1-2dc0b815adaf>
Type:      function

In [4]: def g(x):
   ...:     """
   ...:     Hello
   ...:     
   ...:     What about this?
   ...:     """
   ...:     pass
   ...: 

In [5]: g?
Signature: g(x)
Docstring:
Hello

What about this?
File:      ~/<ipython-input-4-3369882f9e04>
Type:      function
```

Also `help(f)`:
```
Help on function f in module __main__:

f(x)
    Hello
    
    What does this look like.
```
and `help(g)`:
```
Help on function g in module __main__:

g(x)
    Hello
    
    What about this?
```

python/cudf/cudf/_lib/pylibcudf/column.pyx
vyasr: I know :(

python/cudf/cudf/_lib/column.pxd
vyasr: I don't like that we need this parameter. It has nothing to do with libcudf, and everything to do with cudf Python. It maps directly to the `Buffer.get_ptr` function and is necessary so that we access the pointers in the correct way to manage exposure. I don't see a good way around this, but even if there is a better approach I think it makes sense to wait on implementing more complexity until the COW/spilling unification is completed.

To be clear, I don't foresee exposing this functionality at all in a public `to_pylibcudf` function. If a user calls e.g. `Series.to_pylibcudf` we should always assume that it's exposed. This functionality is purely internal to simplify the mapping of the existing cudf Cython internal logic.

python/cudf/cudf/_lib/column.pyx
vyasr: Discussed offline with @wence- and confirmed that we don't need to make this change. I'll resolve all the associated threads as well.

python/cudf/cudf/_lib/pylibcudf/table.pyx
vyasr: For now it's typed to require a list. We can see about relaxing that in the future.

python/cudf/cudf/_lib/pylibcudf/gpumemoryview.pyx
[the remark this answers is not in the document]
vyasr: We discussed this offline and came to two conclusions:
1. Caching this property should be safe. The case that would be problematic would be if the `gpumemoryview` was holding on to a reference to a `Series` or other higher-level object instead of the underlying `Buffer` since the data underneath any cudf `Column` could be swapped out by some operations. The current code always converts from a `Buffer` though, so that conversion is safe.
2. `cached_property` is [actually broken in Cython code](https://github.com/cython/cython/issues/5523).

Therefore, while it may be safe to cache this value, it isn't actually possible with the current Cython, so I've removed this.

PR #13564, opened 2023-06-13
Preemptive fix for hive-partitioning change in dask

rjzamora:
## Description
Preemptive fix for incoming dask PR: https://github.com/dask/dask/pull/10353

Adds handling for the case that the array of known hive-partitioning keys is empty. This will happen when we **don't** want to convert a hive-partitioned column into a categorical column.

### DIFF
--- a/python/dask_cudf/dask_cudf/io/parquet.py
+++ b/python/dask_cudf/dask_cudf/io/parquet.py
@@ -167,6 +167,12 @@ class CudfEngine(ArrowDatasetEngine):
 
             for i, (name, index2) in enumerate(partition_keys):
 
+                if not len(partitions[i].keys):
+                    # Non-categorical case
+                    if name not in df.columns:
+                        df[name] = as_column(index2, length=len(df))
+                    continue
+
                 # Build the column from `codes` directly
                 # (since the category is often a larger dtype)
                 codes = as_column(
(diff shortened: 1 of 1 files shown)


wence- (approved):
One (optional) stylistic change. LGTM! Does this have to be synchronised with the dask side of things?

wence-:
Ah, and can we add a test?

rjzamora:
>Does this have to be synchronised with the dask side of things?... Ah, can we add a test?

We can't really test this until https://github.com/dask/dask/pull/10353 is merged (the test would probably be similar to `test_null_partition`, but with a `filters` argument added to make sure dask is tracking hive-partition info). That may seem like motivation to wait for the dask PR to get in. However, 10353 will break `test_null_partition` without **this** PR, so I was hoping the `dask_cudf` changes would get in first.

wence-:
> We can't really test this until [dask/dask#10353](https://github.com/dask/dask/pull/10353) is merged (the test would probably be similar to `test_null_partition`, but with a `filters` argument added to make sure dask is tracking hive-partition info). That may seem like motivation to wait for the dask PR to get in. However, 10353 will break `test_null_partition` without **this** PR, so I was hoping the `dask_cudf` changes would get in first.

You could write the test you want and `xfail` it (and then the followup PR just removes the xfail).

rjzamora:
>You could write the test you want and xfail it (and then the followup PR just removes the xfail).

Happy to do this. However, if we are strict about xfail in cudf (I think we are?), CI will still be blocked between the time that the dask PR is merged and the xfail is removed.

wence-:
> However, if we are strict about xfail in cudf (I think we are?)

I think we aren't at the moment in dask_cudf actually. But good point, let's get this in and then bring the test in when dask is updated.

PR #13574, opened 2023-06-14
Fix memcheck errors found in REDUCTION_TEST

davidwendt:
## Description
The nightly tests found memcheck errors in `REDUCTION_TEST` for **any/all** aggregation types on **dictionary** columns.
The error is caused by the internal `atomicOr()` function which reads the target value as 32-bit even if the type is bool (8-bit).
The fix works around this limitation by using `atomicOr()` on a 32-bit device scalar and then converting that to a bool scalar on return.

All other tests in `REDUCTION_TEST` now pass memcheck successfully.

Closes #13572 

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13578, opened 2023-06-14
Fix memcheck error found in STRINGS_TEST 

davidwendt:
## Description
Fixes a memcheck error found in `STRINGS_TEST` where an `atomicOr` was used on a boolean device scalar. The workaround uses a `cub::WarpReduce` to compute the result in the warp-per-string kernel.

Reference #13574 

## Checklist

### DIFF
--- a/cpp/src/strings/search/find.cu
+++ b/cpp/src/strings/search/find.cu
@@ -274,30 +274,40 @@ namespace {
 /**
  * @brief Check if `d_target` appears in a row in `d_strings`.
  *
- * This executes as a warp per string/row.
+ * This executes as a warp per string/row and performs well for longer strings.
+ * @see AVG_CHAR_BYTES_THRESHOLD
+ *
+ * @param d_strings Column of input strings
+ * @param d_target String to search for in each row of `d_strings`
+ * @param d_results Indicates which rows contain `d_target`
  */
-struct contains_warp_fn {
-  column_device_view const d_strings;
-  string_view const d_target;
-  bool* d_results;
+__global__ void contains_warp_parallel_fn(column_device_view const d_strings,
+                                          string_view const d_target,
+                                          bool* d_results)
+{
+  size_type const idx = static_cast<size_type>(threadIdx.x + blockIdx.x * blockDim.x);
+  using warp_reduce   = cub::WarpReduce<bool>;
+  __shared__ typename warp_reduce::TempStorage temp_storage;
 
-  __device__ void operator()(std::size_t idx)
-  {
-    auto const str_idx = static_cast<size_type>(idx / cudf::detail::warp_size);
-    if (d_strings.is_null(str_idx)) { return; }
-    // get the string for this warp
-    auto const d_str = d_strings.element<string_view>(str_idx);
-    // each thread of the warp will check just part of the string
-    auto found = false;
-    for (auto i = static_cast<size_type>(idx % cudf::detail::warp_size);
-         !found && (i + d_target.size_bytes()) < d_str.size_bytes();
-         i += cudf::detail::warp_size) {
-      // check the target matches this part of the d_str data
-      if (d_target.compare(d_str.data() + i, d_target.size_bytes()) == 0) { found = true; }
-    }
-    if (found) { atomicOr(d_results + str_idx, true); }
+  if (idx >= (d_strings.size() * cudf::detail::warp_size)) { return; }
+
+  auto const str_idx  = idx / cudf::detail::warp_size;
+  auto const lane_idx = idx % cudf::detail::warp_size;
+  if (d_strings.is_null(str_idx)) { return; }
+  // get the string for this warp
+  auto const d_str = d_strings.element<string_view>(str_idx);
+  // each thread of the warp will check just part of the string
+  auto found = false;
+  for (auto i = static_cast<size_type>(idx % cudf::detail::warp_size);
+       !found && (i + d_target.size_bytes()) < d_str.size_bytes();
+       i += cudf::detail::warp_size) {
+    // check the target matches this part of the d_str data
+    if (d_target.compare(d_str.data() + i, d_target.size_bytes()) == 0) { found = true; }
   }
-};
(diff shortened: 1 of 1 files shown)


cpp/src/strings/search/find.cu
```diff
@@ -274,30 +274,40 @@ namespace {
...
+  // each thread of the warp will check just part of the string
+  auto found = false;
+  for (auto i = static_cast<size_type>(idx % cudf::detail::warp_size);
+       !found && (i + d_target.size_bytes()) < d_str.size_bytes();
+       i += cudf::detail::warp_size) {
+    // check the target matches this part of the d_str data
+    if (d_target.compare(d_str.data() + i, d_target.size_bytes()) == 0) { found = true; }
   }
-};
+  // if (found) { atomicOr(d_results + str_idx, true); }
```
bdice, line 307: Delete the previous code's comment.
```suggestion
```

bdice:
Does this change have performance implications?

davidwendt:
> Does this change have performance implications?

Nothing noticeable. I just thought the `atomicOr` code looked cleaner with the `thrust::for_each` taking care of all the kernel launch details. Maybe @PointKernel could recommend the `cuda::atomic_ref` equivalent? I had this working before he mentioned replacing `atomicOr`.

cpp/src/strings/search/find.cu
```diff
@@ -324,11 +333,11 @@ std::unique_ptr<column> contains_warp_parallel(strings_column_view const& input,
...
   if (!d_target.empty()) {
     // launch warp per string
-    auto d_strings = column_device_view::create(input.parent(), stream);
-    thrust::for_each_n(rmm::exec_policy(stream),
-                       thrust::make_counting_iterator<std::size_t>(0),
-                       static_cast<std::size_t>(input.size()) * cudf::detail::warp_size,
-                       contains_warp_fn{*d_strings, d_target, results_view.data<bool>()});
+    auto const d_strings     = column_device_view::create(input.parent(), stream);
+    constexpr int block_size = 256;
+    cudf::detail::grid_1d grid{input.size() * cudf::detail::warp_size, block_size};
```
ttnghia, line 338: What if `input.size() * cudf::detail::warp_size` overflow? `grid_1d` only has `int` members.

ttnghia: Oh I just found the similar code in other files (`attributes.cu` and `find.cu`). So this may be a new potential issue.

davidwendt: You missed this line of code perhaps?
https://github.com/rapidsai/cudf/pull/13578/files#diff-048f86c21559b14f64f86aaeaa57776d366c3a4948a5aba7c0ab1a3801be87bcR292
```
 if (idx >= (d_strings.size() * cudf::detail::warp_size)) { return; }
```

cpp/src/strings/search/find.cu
```diff
@@ -274,30 +274,39 @@ namespace {
...
-    // each thread of the warp will check just part of the string
-    auto found = false;
-    for (auto i = static_cast<size_type>(idx % cudf::detail::warp_size);
-         !found && (i + d_target.size_bytes()) < d_str.size_bytes();
-         i += cudf::detail::warp_size) {
-      // check the target matches this part of the d_str data
-      if (d_target.compare(d_str.data() + i, d_target.size_bytes()) == 0) { found = true; }
-    }
-    if (found) { atomicOr(d_results + str_idx, true); }
+  if (idx >= (d_strings.size() * cudf::detail::warp_size)) { return; }
```
ttnghia, line 292: Again I suspect that this mul can overflow, as all the operands are of type `int`. So maybe we should cast into `int64_t`?
```suggestion
  if (static_cast<int64_t>(idx) >= static_cast<...>(d_strings.size()> * static_cast<...>(cudf::detail::warp_size)) { return; }
```

PointKernel:
I was inclined to use `atomic_ref` for this issue and now realize the warp-reduce workaround is a better solution: `atomic_ref` supports exclusively 4-byte or 8-byte types since only 32-bit and 64-bit atomic CAS are supported at the hardware level (plus 128-bit on hopper). Technically, we should never do atomic operations over bools.

cpp/src/strings/search/find.cu
ttnghia: Wait, that line is inside the kernel, while this line is before kernel. If we have overflow here, we may still launch a kernel with some (large?) input. We should avoid launching the kernel from here.

davidwendt: An overflow cannot technically occur here since this code path is only for long strings which is always much greater than 32 bytes on average. This means the number of rows * 32 will never overflow under these conditions.

cpp/src/strings/search/find.cu
davidwendt: An overflow cannot technically occur here since this code path is only for long strings which is always much greater than 32 bytes on average. This means the (number of rows * 32) will never overflow under these conditions.

cpp/src/strings/search/find.cu
```diff
@@ -274,30 +274,39 @@ namespace {
...
+  // each thread of the warp will check just part of the string
+  auto found = false;
+  for (auto i = static_cast<size_type>(idx % cudf::detail::warp_size);
+       !found && (i + d_target.size_bytes()) < d_str.size_bytes();
+       i += cudf::detail::warp_size) {
+    // check the target matches this part of the d_str data
+    if (d_target.compare(d_str.data() + i, d_target.size_bytes()) == 0) { found = true; }
   }
-};
+  auto const result = warp_reduce(temp_storage).Reduce(found, cub::Max());
```
bdice, line 307: Wondering if we could get an early-exit benefit by checking the warp-reduced result before reading the full string. (Discussed offline with @davidwendt.) I don't have a good expectation for the synchronization cost of a single warp sync. It'll probably be slower, but I'd like to learn by how much.

bdice (approved):

ttnghia (approved):

Issue #13326, opened 2023-05-10
[BUG] ORC compression of nested strings is worse than CPU and GPU parquet


revans2:
**Describe the bug**
When a customer was trying to write out data that is similar to [this](https://github.com/rapidsai/cudf/files/11443156/data.zip) they saw that the output size of the ORC data written with CUDF(Spark) was more than 2x that of the same data written on the CPU. 

For this particular customer we are talking about TiB of difference. Not only was it more expensive to store the data, the size difference was enough to slow down later jobs that read the data enough that they could not win against the CPU in performance.

Looking at the footers for the files in question it looks like the the GPU is not doing dictionary encoding where as the CPU is. Looking at the GPU code it is clear that we don't try to do dictionary encoding for ORC if there are more rows than would fit in a `uint16_t`

https://github.com/rapidsai/cudf/blob/ac158da6c3ff3cb3d0b7743d65b738c0d422a84d/cpp/src/io/orc/writer_impl.cu#L2112

The default stripe size (at the top level) is 100,000 rows, which should allow dictionary encoding for all columns, but if a nested column with a LIST or a MAP in it has on average more than 7 entries it will not even be considered for dictionary encoding. I think it has something to do with wanting to compute the dictionary chunks in a single kernel call, saving memory while doing it, and not needing to compute 32-bit indexes as temporary values when trying to get to 16-bit indexes.

Whatever the reason it results in much larger files. I would be willing to have slightly slower run time for ORC compression and slightly more memory if it would allow us to compress some of these columns in these different cases.

Perhaps we could have an alternative path for string columns that have too many entries in them instead of just skipping them all together.

**Steps/Code to reproduce bug**
Take the attached file and rewrite the data using the GPU.  For spark it is just a normal transcode, but for CUDF and the python API I am not 100% sure. I will try to make it happen, but I wanted to file this sooner than later.

**Expected behavior**
The ORC files produced would be much closer to the size of the files produced by the CPU. I know that we might not be smaller, but much closer in size would be good. It would be nice if we could get them close to the size of the same files written with Parquet on the GPU. Which is much better than the CPU/Spark implementation so I know we should be able to do something.

| | GPU | CPU(Spark) |
|-|-------|--------|
|*ORC*| 46 MiB | 12 MiB |
|*Parquet*| 15 MiB | 42 MiB |

GregoryKimball:
Thank you @revans2 for sharing this issue. I believe the size signals here are coming from compression differences, rather than encoding.

[...]

Issue #10495, opened 2022-03-23
[FEA] Refactor ORC's dictionary encoding using cuCo


devavret:
ORC writer uses dictionary encoding on string columns and uses a custom hash map like structure. This is not needed anymore as we can use cuCo's hash maps.

Using #8476 as a template, we should be able to refactor and clean up the dictionary encoding code that currently resides in `src/io/orc/dict_enc.cu`

PointKernel:
Still relevant

GregoryKimball:
Closed by https://github.com/rapidsai/cudf/pull/13580

PR #13580, opened 2023-06-14
Use `cuco::static_map` to build string dictionaries in ORC writer

vuule:
## Description
Issue #13326, #10495

This PR reimplements creation of stripe dictionaries in ORC writer to eliminate row group size limitations.
New implementation uses `cuco::static_map` in a way that's very similar to the Parquet writer.

PR brings large performance gains because per-column X per-stripe sorting that invoked hundreds of thrust calls is now removed.
Also verified that the original row group size limit (2^16) for dictionary encoding is removed, allowing dictionaries to be applicable to large lists of strings.

## Checklist


cpp/src/io/orc/writer_impl.cu
```diff
@@ -72,6 +72,17 @@ namespace orc {
 using namespace cudf::io::orc;
 using namespace cudf::io;
 
+template <typename T>
+[[nodiscard]] constexpr int varint_size(T val)
```
vuule, line 76: Could not eliminate the duplicated implementation without creating a new header file. Still, moved one of the implementations from a header to a source file.
This function should be in a utility header and reused in ORC and Parquet.

ttnghia:
You mentioned "large performance gain", so can you provide the benchmark results please?

vuule:
Extracted from a spreadsheet:
```
cardinality	run_length	23.08 BPS	CUCO BPS	Speedup
0		1		884191465	2346393107	2.653716078
100000		1		761144814	1925509704	2.529754744
10000		1		670813573	2980105169	4.442523659
1000		1		648093761	5048907003	7.790395938
0		4		878086121	2344256447	2.669734085
100000		4		755215739	3595923315	4.761451767
10000		4		710512029	3944427488	5.551528091
1000		4		648151737	4569376914	7.049856775
0		16		877626971	2341761710	2.668288222
100000		16		814321137	7689395387	9.442706369
10000		16		788388948	7827880669	9.928957894
1000		16		752374125	8268318027	10.98963634
```

cpp/src/io/orc/orc.cpp
```diff
@@ -24,9 +24,19 @@
 
 #include <string>
 
-namespace cudf {
-namespace io {
-namespace orc {
+namespace cudf::io::orc {
+
+namespace {
+[[nodiscard]] constexpr uint32_t varint_size(uint64_t val)
```
ttnghia, line 30: Unused?

cpp/src/io/orc/orc_gpu.hpp
```diff
@@ -39,6 +40,19 @@ namespace gpu {
...
+
+using map_type = cuco::static_map<size_type, size_type>;
+
+/**
+ * @brief The alias of `map_type::pair_atomic_type` class.
+ *
+ * Declare this struct by trivial subclassing instead of type aliasing so we can have forward
+ * declaration of this struct somewhere else.
+ */
+struct slot_type : public map_type::pair_atomic_type {};
```
PointKernel, line 54:
```suggestion
struct slot_type : public map_type::slot_type {};
```
Can we use [slot_type](https://github.com/NVIDIA/cuCollections/blob/303f134573afa315cf14fca3f7a0b730438497c3/include/cuco/static_map.cuh#L596) directly?

cpp/src/io/orc/orc_gpu.hpp
```diff
@@ -172,36 +186,63 @@ struct StripeStream {
...
-  orc_column_device_view const* leaf_column;  //!< Pointer to string column
-};
+void initialize_dictionary_hash_maps(device_2dspan<stripe_dictionary> dictionaries,
+                                     rmm::cuda_stream_view stream);
+
+/**
+ * @brief Populates the hash maps with unique values from the stripe.
+ *
+ * @param dictionaries Dictionary descriptors
+ * @param orc_columns Pre-order flattened device array of ORC column views
```
PointKernel, line 220:
```suggestion
 * @param columns  Pre-order flattened device array of ORC column views
```

cpp/src/io/orc/orc_gpu.hpp
```diff
@@ -317,14 +358,16 @@ void EncodeOrcColumnData(device_2dspan<EncChunk const> chunks,
 /**
  * @brief Launches kernel for encoding column dictionaries
  *
- * @param[in] stripes Stripe dictionaries device array [stripe][string_column]
+ * @param[in] stripes Stripe dictionaries device array
+ * @param[in] orc_columns Pre-order flattened device array of ORC column views
```
PointKernel, line 362:
```suggestion
 * @param[in] columns Pre-order flattened device array of ORC column views
```

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -996,14 +996,15 @@ __global__ void __launch_bounds__(block_size)
 /**
  * @brief Encode column dictionaries
  *
- * @param[in] stripes Stripe dictionaries device array [stripe][string_column]
+ * @param[in] stripes Stripe dictionaries device array
```
PointKernel, line 999:
```suggestion
 * @param[in] stripes Stripe dictionaries device array
 * @param[in] columns Pre-order flattened device array of ORC column views
```

cpp/src/io/orc/dict_enc.cu
```diff
@@ -16,471 +16,265 @@
...
-      if (ck_row_ref != ck_row) {
-        string_view reference_string =
-          s->chunk.leaf_column->element<string_view>(ck_row_ref + start_row);
-        is_dupe = (string_value == reference_string);
-        dict_char_count += (is_dupe) ? 0 : string_length;
-      }
+    if (t == 0) {
+      auto dict_g            = &dictionaries[col_idx][stripe_idx];
+      total_num_dict_entries = atomicAdd(&dict_g->entry_count, num_unique);
+      total_num_dict_entries += num_unique;
```
PointKernel, line 156: Is this `total_num_dict_entries` used anywhere?

cpp/src/io/orc/dict_enc.cu
```diff
@@ -16,471 +16,265 @@
...
-  using block_reduce = cub::BlockReduce<uint32_t, block_size>;
-  using block_scan   = cub::BlockScan<uint32_t, block_size, cub::BLOCK_SCAN_WARP_SCANS>;
+  auto const start_row = dict.start_row;
+  auto const end_row   = dict.start_row + dict.num_rows;
 
-  __shared__ union {
-    typename block_reduce::TempStorage reduce_storage;
-    typename block_scan::TempStorage scan_storage;
-  } temp_storage;
+  if (t == 0) { auto const& offsets = col.child(strings_column_view::offsets_column_index); }
```
PointKernel, line 121: `offsets` seems unused

cpp/src/io/orc/dict_enc.cu
```diff
@@ -16,471 +16,265 @@
...
   }
 }
 
-/**
- * @brief Gather all non-NULL string rows and compute total character data size
- */
-// blockDim {block_size,1,1}
+struct equality_functor {
+  column_device_view const& col;
+  __device__ bool operator()(size_type lhs_idx, size_type rhs_idx)
```
PointKernel, line 91:
```suggestion
  __device__ bool operator()(size_type lhs_idx, size_type rhs_idx) const
```

cpp/src/io/orc/orc_gpu.hpp
vuule: Would you be opposed to removing the slot_type "alias"? Once I apply your suggestion, the benefit from the "alias" shrinks and it really only saves a few words.

cpp/src/io/orc/dict_enc.cu
vuule: good catch! Simplified the kernel logic around this.

abellina:
@vuule, this patch greatly improved the size of nested string columns in a customer case we have. Before, the example we had locally was 2x larger, but now it's around 30% larger than the CPU, which is a big improvement. We still want to figure out what else could be improved but it seems we should proceed and get this change in and open a subsequent issue if we find more areas of improvement.

cpp/src/io/orc/orc_gpu.hpp
```diff
@@ -29,6 +29,7 @@
 #include <cudf/types.hpp>
 #include <cudf/utilities/span.hpp>
 
+#include <cuco/static_map.cuh>
```
PointKernel, line 32: Theoratically, cuco is a different header group from rmm and is considered "further" so this should be placed after `#include <rmm/cuda_stream_view.hpp>` with a blank line in between.

PointKernel (approved):
One minor suggestion and LGTM

cpp/src/io/orc/orc_gpu.hpp
vuule: sorted :+1:

cpp/src/io/json/json_tree.cu
```diff
@@ -28,15 +28,15 @@
...
-#include <cub/device/device_radix_sort.cuh>
-
 #include <rmm/cuda_stream_view.hpp>
 #include <rmm/device_uvector.hpp>
 #include <rmm/exec_policy.hpp>
 #include <rmm/mr/device/polymorphic_allocator.hpp>
 
+#include <cub/device/device_radix_sort.cuh>
+
+#include <cuco/static_map.cuh>
```
ttnghia, line 38: I believe that `cub` and `thurst` headers should stay closer together than with `cuco`.

cpp/src/io/orc/dict_enc.cu
```diff
@@ -16,471 +16,261 @@
...
-    return 0;
-  } else {
-    char const* ptr = val.data();
-    uint32_t len    = val.size_bytes();
-    return (ptr[0] + (ptr[len - 1] << 5) + (len << 10)) & ((1 << init_hash_bits) - 1);
-  }
-}
+namespace cudf::io::orc::gpu {
+
+constexpr int DEFAULT_BLOCK_SIZE = 256;
```
ttnghia, line 27: What block? Data block or thread block? Can you add a comment for it please?

cpp/src/io/orc/dict_enc.cu
```diff
@@ -16,471 +16,261 @@
...
-    __syncthreads();
-    if (!t) { s->nnz += tmp_nnz; }
-    if (is_valid) { s->dict[nz_pos] = i + t; }
-    __syncthreads();
-  }
+  // Index of the column in the `str_col_indexes` array
+  uint32_t const str_col_idx = blockIdx.y;
+  // Index of the column in the `orc_columns` array
+  auto const col_idx     = str_col_indexes[str_col_idx];
+  uint32_t row_group_idx = blockIdx.x * blockDim.x + threadIdx.x;
```
ttnghia, line 41: `auto const`?

cpp/src/io/orc/dict_enc.cu
```diff
@@ -16,471 +16,261 @@
...
-      t_dict_data[i + t] = start_row + ck_row;
+template <int block_size>
+__global__ void __launch_bounds__(block_size)
+  initialize_dictionary_hash_maps_kernel(device_span<stripe_dictionary> dictionaries)
+{
+  auto const dict_map = dictionaries[blockIdx.x].map_slots;
+  auto const t        = threadIdx.x;
+  for (size_type i = 0; i < dict_map.size(); i += block_size) {
+    if (t + i < dict_map.size()) {
+      new (&dict_map[t + i].first) map_type::atomic_key_type{KEY_SENTINEL};
```
ttnghia, line 83: Can we use just this?
```
auto& dict_map = ...;
dict_map[t + i] = {..., ...};
```

cpp/src/io/orc/dict_enc.cu
```diff
@@ -16,471 +16,261 @@
...
+template <int block_size>
+__global__ void __launch_bounds__(block_size)
+  populate_dictionary_hash_maps_kernel(device_2dspan<stripe_dictionary> dictionaries,
+                                       device_span<orc_column_device_view const> columns)
+{
+  auto const col_idx    = blockIdx.x;
+  auto const stripe_idx = blockIdx.y;
+  auto const t          = threadIdx.x;
+  auto& dict            = dictionaries[col_idx][stripe_idx];
+  auto const col        = columns[dict.column_idx];
```
ttnghia, line 116: `auto const&`?

cpp/src/io/orc/dict_enc.cu
```diff
@@ -16,471 +16,261 @@
...
+  auto const t = threadIdx.x;
+  auto map     = map_type::device_view(dict.map_slots.data(),
+                                   dict.map_slots.size(),
+                                   cuco::empty_key{KEY_SENTINEL},
+                                   cuco::empty_value{VALUE_SENTINEL});
+
+  __shared__ cuda::atomic<size_type, cuda::thread_scope_block> counter;
+
+  using cuda::std::memory_order_relaxed;
+  if (t == 0) { new (&counter) cuda::atomic<size_type, cuda::thread_scope_block>{0}; }
```
ttnghia, line 181: Similarly, why placement new instead of simple assignment?

cpp/src/io/orc/dict_enc.cu
vuule: I don't know why placement new is used here.
This is a question for @PointKernel, took this from his Parquet code :grin:

cpp/src/io/json/json_tree.cu
vuule: My understanding of the include order:

1. local dir
2. cuIO
3. libcudf
4. rmm
5. other
6. std C++ lib

Is there a whole hierarchy in "other" that I'm missing? :D

ttnghia: I meant, the order should be thrust=>cub (or cub=>thrust) => cuco, instead of cub => cuco => thrust like this.

vuule: could you please explain why?
I assumed these are all equal so I sorted alphabetically.

ttnghia: Because `thrust` internally calls `cub` so they are more related so I think they should stay together (in theory).
In reality that doesn't matter at all :smile:

vuule: AFAIK the ordering in general aims to be such that the lowest dependencies are listed last, so we don't cover up missing dependencies by accident.
I don't think we need to be as cautious with the order as you suggest, but I'll make the change if you think it's warranted.

ttnghia: That makes sense to me. It doesn't matter much---very minor nit.

cpp/src/io/orc/dict_enc.cu
vuule: I tried to assign, getting this error
` error: function "cuda::__4::atomic<_Tp, _Sco>::operator=(const cuda::__4::atomic<cudf::size_type, cuda::std::__4::__detail::thread_scope_device> &) [with _Tp=cudf::size_type, _Sco=cuda::std::__4::__detail::thread_scope_device]" (declared implicitly) cannot be referenced -- it is a deleted function`
I assume it's because of atomic types used here.

ttnghia (approved):

PR #13583, opened 2023-06-14
Clean up cudf device atomic with `cuda::atomic_ref`

### LINKED ISSUE #13575: [BUG] The libcudf device atomic operations cause memcheck errors on types smaller than 4 bytes
The [`device_atomics`](https://github.com/rapidsai/cudf/blob/branch-23.08/cpp/include/cudf/detail/utilities/device_atomics.cuh) implementation in libcudf are failing memchecks when used on types smaller than 4 bytes.

https://github.com/rapidsai/cudf/blob/c929a84719baf169d05e4d0ee87a0df6e77f5b45/cpp/include/cudf/detail/utilities/device_atomics.cuh#L65-L69

This was found while investigating some memcheck errors that occurred on `cudf::reduce` where the aggregation return type is `bool` (1 byte) and an `atomicOr` was used to produce the output. Reference: https://github.com/rapidsai/cudf/pull/13574

The solution is to use the libcudaxx `cuda::atomic` and `cuda::atomic_ref` in place of these implementations and potentially remove the `device_atomics.cuh` altogether.

PointKernel:
## Description

Contributes to #13575

Depends on #13574, #13578

This PR cleans up custom atomic implementations in libcudf by using `cuda::atomic_ref` when possible. It removes atomic bitwise operations like `and`, `or` and `xor` since libcudac++ already provides proper replacements.

## Checklist


cpp/tests/hash_map/map_test.cu
```diff
@@ -68,10 +68,7 @@ struct InsertTest : public cudf::test::BaseFixture {
 
 using TestTypes = ::testing::Types<key_value_types<int32_t, int32_t>,
                                    key_value_types<int64_t, int64_t>,
-                                   key_value_types<int8_t, int8_t>,
                                    key_value_types<int16_t, int16_t>,
-                                   key_value_types<int8_t, float>,
-                                   key_value_types<int16_t, double>,
```
PointKernel, line 74: Removed types not supported by `atomic_ref`. Not really an issue since those data structures will be replaced by cuco ones soon.

cpp/src/text/minhash.cu
```diff
@@ -92,7 +93,9 @@ struct minhash_fn {
       for (std::size_t seed_idx = 0; seed_idx < seeds.size(); ++seed_idx) {
         auto const hasher = cudf::detail::MurmurHash3_32<cudf::string_view>{seeds[seed_idx]};
         auto const hvalue = hasher(hash_str);
-        atomicMin(d_output + seed_idx, hvalue);
+        cuda::atomic_ref<cudf::hash_value_type, cuda::thread_scope_block> ref{
```
PointKernel, line 96: IIUC, this is a per warp function thus using `thread_scope_block` is safe.

cpp/src/strings/case.cu
```diff
@@ -167,7 +168,10 @@ struct count_bytes_fn {
       size += converter.process_character(u8);
     }
     // this is every so slightly faster than using the cub::warp_reduce
-    if (size > 0) atomicAdd(d_offsets + str_idx, size);
+    if (size > 0) {
+      cuda::atomic_ref<size_type, cuda::thread_scope_block> ref{*(d_offsets + str_idx)};
```
PointKernel, line 172: Is `thread_scope_block` correct?

cpp/src/strings/search/find.cu
```diff
@@ -154,7 +155,9 @@ __global__ void finder_warp_parallel_fn(column_device_view const d_strings,
 
   // find stores the minimum position while rfind stores the maximum position
   // note that this was slightly faster than using cub::WarpReduce
-  forward ? atomicMin(d_results + str_idx, position) : atomicMax(d_results + str_idx, position);
+  cuda::atomic_ref<size_type, cuda::thread_scope_block> ref{*(d_results + str_idx)};
```
PointKernel, line 158: Is `thread_scope_block` correct?`

cpp/src/strings/case.cu
bdice: I think that might be correct, but it'd be good to double check with @davidwendt.

In my understanding this is safe because only one block (really one warp, but that doesn't seem to be an atomic scope in libcudacxx) is accessing the data at `d_offsets + str_idx`. If there is a way to express atomicity at the warp level in libcudacxx I'd like to know how. Warps are not mentioned in https://nvidia.github.io/libcudacxx/extended_api/memory_model.html.

Similar reasoning applies to the other places where you left this comment.

bdice (approved):
Minor question about `#include`s. I think your assessment of `thread_scope_block` is correct, so I'll approve. Thanks for doing this!

cpp/src/strings/case.cu
davidwendt: This is correct. No two warps will update the same memory location so thread-scope should be sufficient here.

PointKernel:
> Is something in cpp/include/cudf/detail/utilities/device_atomics.cuh still being used?

Yes, we cannot close the issue for now since some types are not supported by `atomic_ref` yet like durations, timestamps and bools. Opened https://github.com/NVIDIA/libcudacxx/issues/496 to track this.

davidwendt (approved):

cpp/src/join/mixed_join_size_kernels_semi.cu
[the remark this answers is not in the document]
bdice: Okay. No worries!

PR #13584, opened 2023-06-14
Fix Parquet multi-file reading 

### LINKED ISSUE #13559: [BUG] Unexpected malformed parquet page data error when reading multiple parquet files using libcudf
**Describe the bug**

I have two parquet files. These files can be read successfully with no issue when I read them one at a time. However, when I read them together using `cudf::io::read_parquet`, it fails with an error below.

```
CUDF failure at: /path/to/libcudf/cpp/src/io/parquet/reader_impl_preprocess.cu:1391: Encountered malformed parquet page data (row count mismatch in page data)
```

IIUC, this error means that some column has a different number of rows than other columns in the given files. However, I don't think that is the case because each file can be read individually with no issue.

This error occurs very rarely. I can reproduce it only when certain conditions are met:

- When you read certain files together
- When you read certain columns in a certain order. The column order should be different than the column definition order in the parquet file schema.

This error can be reproduced only with certain input files. Please find the files I've been using in https://drive.google.com/file/d/16VWhJu8q6eoUAM1h20eSqDkK4MZDJ_wP/view?usp=sharing (it's about 50 MB). In the data set shared, there are two parquet files of the TPC-H customer table. Once you download them, you can reproduce the error using the code below. You can copy-paste this test into `cudf/cpp/tests/io/parquet_test.cpp`.

Note that the column order matters to reproduce the error. The same files can be read successfully if you reverse the order to `{"c_phone", "c_acctbal"}`.

```cpp
#include <arrow/io/api.h>
#include <arrow/filesystem/localfs.h>

TEST_F(ParquetReaderTest, MalformedParquetPageDataError)
{
  std::vector<std::string> files = {"/path/to/customer/customer.8619.parquet", "/path/to/customer/customer.8881.parquet"};
  std::vector<std::string> cols = {"c_acctbal", "c_phone"};

  auto fs = std::make_shared<arrow::fs::LocalFileSystem>();
  std::vector<std::shared_ptr<arrow::io::RandomAccessFile>> arrow_files;
  for (const auto& f : files) {
    arrow_files.emplace_back(fs->OpenInputFile(f).ValueOrDie());
  }

  std::vector<cudf::io::arrow_io_source> arrow_sources;
  std::vector<cudf::io::datasource*> datasources;
  for (auto& file : arrow_files) {
    arrow_sources.emplace_back(cudf::io::arrow_io_source{std::move(file)});
  }
  for (auto& arrow_source : arrow_sources) {
    datasources.emplace_back(&arrow_source);
  }
  cudf::io::source_info source{datasources};

  cudf::io::parquet_reader_options read_opts =
      cudf::io::parquet_reader_options::builder(source)
        .use_pandas_metadata(false)
        .convert_strings_to_categories(false)
        .columns(cols);
  auto result = cudf::io::read_parquet(read_opts);
  fprintf(stderr, "%d\n", result.tbl->num_rows());
}
```

**Expected behavior**

Since those files are not malformed, `cudf::io::read_parquet` should not fail with this malformed page data error.

 - Method of cuDF install: conda and from source

**Libcudf version tested**

Only 23.04.0 is tested.

etseidl:
## Description
Fixes #13559 

When reading parquet files, an attempt is made to read multiple column chunks in a single read if they are contiguous. In the case where there are multiple files with equal sized column chunks, reading the files with adjacent column swapped can result in reading data from the wrong file. This PR fixes this by testing that otherwise contiguous seeming chunks are actually in the same file.

## Checklist


jihoonson:
@etseidl thanks for working on this issue! I tested your change, and it does fix https://github.com/rapidsai/cudf/issues/13559 in my testing setup!

etseidl:
> This means, the number of read tasks depends on the input column ordering. Is my understanding correct?

Yes, this is correct.  In a pathological case you could wind up with one read per column rather than one read per row group (assuming the goal is to read all columns, but in reverse order).  I would view this PR as a short-term fix. It would probably be better to first read the requested columns in order, and then do a fast reordering of the columns before returning the table to the user.

> If so, will this have performance impact? I'm wondering if it would be almost always better to match the input column ordering to that in the file schema when reading adjacent columns.

I think the performance hit will vary.  Column pruning can also lead to disjoint reads even if done in order, but reduces the amount of data read. With fast storage, it's probably worth doing the extra I/O calls, but if data is in S3, for instance, it might be better to read columns you don't want to do fewer gets.

etseidl:
@jihoonson I did a quick experiment with a 39 column file I've got.  The top is reading columns 21-39 in order, the bottom is reading from 39-21. The I/O times are nearly identical, although the access patterns are a little more even when the file is read in order. Perhaps someone who knows kvikio better than me can chime in here.

Oddly, the decompression times are much longer reversed than in order.  I think this is because the 38th column is very wide, so there might be a utilization problem here with too many wide pages starving the GPU of work.

Screenshot from 2023-06-16 13-20-20

vuule (approved):
good catch! Thank you for the fix @etseidl

davidwendt (approved):
LGTM

ttnghia (approved):

PR #13585, opened 2023-06-15
Fix memcheck error found in PARQUET_TEST

etseidl:
## Description
Memcheck found invalid memory reads in the Parquet reader. The read overruns were caused by reading beyond the end of misaligned buffers during copy operations such as [this](https://github.com/rapidsai/cudf/blob/67efaf62c7314fbda56d39ea5f60bb893c406084/cpp/src/io/parquet/page_data.cu#L654). This PR adds padding to the end of buffers used to store column chunk data to allow these copies to succeed without reading unallocated memory.

Closes #13571

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

Issue #13571, opened 2023-06-14
[BUG] Invalid memory read in cudf::io::parquet::gpu::gpuDecodePageData


davidwendt:
The nightly memcheck tests found an invalid global device memory read in `cudf::io::parquet::gpu::gpuDecodePageData`

```
========= Invalid __global__ read of size 4 bytes
=========     at 0x4dd0 in void cudf::io::parquet::gpu::<unnamed>::gpuDecodePageData<(int)256, unsigned char>(cudf::io::parquet::gpu::PageInfo *, cudf::device_span<const cudf::io::parquet::gpu::ColumnChunkDesc, (unsigned long)18446744073709551615>, unsigned long, unsigned long)
=========     by thread (81,0,0) in block (0,0,0)
=========     Address 0x7f0511e00ad8 is out of bounds
=========     and is inside the nearest allocation at 0x7f0511e00a00 of size 218 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x304e32]
=========                in /usr/lib/x86_64-linux-gnu/libcuda.so.1
=========     Host Frame: [0x1488c]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudart.so.11.0
=========     Host Frame:cudaLaunchKernel [0x6c318]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudart.so.11.0
=========     Host Frame:cudf::io::parquet::gpu::DecodePageData(cudf::detail::hostdevice_vector<cudf::io::parquet::gpu::PageInfo>&, cudf::detail::hostdevice_vector<cudf::io::parquet::gpu::ColumnChunkDesc> const&, unsigned long, unsigned long, int, rmm::cuda_stream_view) [0x155aaef]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudf.so
=========     Host Frame:cudf::io::detail::parquet::reader::impl::decode_page_data(unsigned long, unsigned long) [0x1564ba9]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudf.so
=========     Host Frame:cudf::io::detail::parquet::reader::impl::read_chunk_internal(bool) [0x1566b2d]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudf.so
[...]

Issue #13570, opened 2023-06-14
[BUG] Invalid memory read in cudf::io::orc::gpu::gpuDecodeOrcColumnData


davidwendt:
The nightly memcheck tests found an invalid global device memory read in `cudf::io::orc::gpu::gpuDecodeOrcColumnData`

```
========= Invalid __global__ read of size 1 bytes
=========     at 0x15e0 in void cudf::io::orc::gpu::gpuDecodeOrcColumnData<(int)1024>(cudf::io::orc::gpu::ColumnDesc *, cudf::io::orc::gpu::DictionaryEntry *, cudf::table_device_view, cudf::detail::base_2dspan<cudf::io::orc::gpu::RowGroup, cudf::device_span>, unsigned long, unsigned int, unsigned long, int *)
=========     by thread (13,0,0) in block (0,0,0)
=========     Address 0x7f25a0200c7a is out of bounds
=========     and is 1 bytes after the nearest allocation at 0x7f25a0200c00 of size 122 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x304e32]
=========                in /usr/lib/x86_64-linux-gnu/libcuda.so.1
=========     Host Frame: [0x1488c]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudart.so.11.0
=========     Host Frame:cudaLaunchKernel [0x6c318]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudart.so.11.0
=========     Host Frame:cudf::io::orc::gpu::DecodeOrcColumnData(cudf::io::orc::gpu::ColumnDesc*, cudf::io::orc::gpu::DictionaryEntry*, cudf::detail::base_2dspan<cudf::io::orc::gpu::RowGroup, cudf::device_span>, unsigned int, unsigned int, unsigned long, cudf::table_device_view, unsigned int, unsigned int, unsigned long, int*, rmm::cuda_stream_view) [0x1521efd]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudf.so
[...]

Issue #13567, opened 2023-06-14
[BUG] Invalid memory read in cudf::io::inflate_kernel


davidwendt:
The nightly memcheck tests found an invalid global device memory read in `cudf::io::inflate_kernel`

```
========= Invalid __global__ read of size 4 bytes
=========     at 0x11fa0 in void cudf::io::inflate_kernel<(int)128>(cudf::device_span<const cudf::device_span<const unsigned char, (unsigned long)18446744073709551615>, (unsigned long)18446744073709551615>, cudf::device_span<const cudf::device_span<unsigned char, (unsigned long)18446744073709551615>, (unsigned long)18446744073709551615>, cudf::device_span<cudf::io::compression_result, (unsigned long)18446744073709551615>, cudf::io::gzip_header_included)
=========     by thread (0,0,0) in block (2605,0,0)
=========     Address 0x7fba9f00654c is out of bounds
=========     and is inside the nearest allocation at 0x7fba9a000000 of size 83912015 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x304e32]
=========                in /usr/lib/x86_64-linux-gnu/libcuda.so.1
=========     Host Frame: [0x1488c]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudart.so.11.0
=========     Host Frame:cudaLaunchKernel [0x6c318]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudart.so.11.0
=========     Host Frame:cudf::io::gpuinflate(cudf::device_span<cudf::device_span<unsigned char const, 18446744073709551615ul> const, 18446744073709551615ul>, cudf::device_span<cudf::device_span<unsigned char, 18446744073709551615ul> const, 18446744073709551615ul>, cudf::device_span<cudf::io::compression_result, 18446744073709551615ul>, cudf::io::gzip_header_included, rmm::cuda_stream_view) [0x1429446]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudf.so
=========     Host Frame:cudf::io::text::(anonymous namespace)::bgzip_data_chunk_reader::decompression_blocks::decompress(rmm::cuda_stream_view) [0x15b3ad1]
=========                in /opt/conda/envs/test/bin/gtests/libcudf/../../../lib/libcudf.so
[...]

PR #13586, opened 2023-06-16
Fix memory issues in cuIO due to removal of memory padding

ttnghia:
After `rmm` removed memory padding (https://github.com/rapidsai/rmm/pull/1278), some of cuIO code started to have out-of-bound access issues because many of its compute kernels shift the input pointers back and forth to satisfy some alignment.

This adds back padding to various memory buffers so the buffers now will have some extra space enough for such shifting.

With this fix, the reported issues (https://github.com/rapidsai/cudf/issues/13567,  https://github.com/rapidsai/cudf/issues/13571, https://github.com/rapidsai/cudf/issues/13570) no longer show up.

Closes:
 * https://github.com/rapidsai/cudf/issues/13567
 * https://github.com/rapidsai/cudf/issues/13571
 * https://github.com/rapidsai/cudf/issues/13570


ttnghia:
@vuule Please also check the entire codebase more carefully. I tried to find as many related buffers as possible but I may still miss some.

ttnghia:
I just realize that there is a duplicate fix for Parquet (https://github.com/rapidsai/cudf/pull/13585), which implements a different approach than this: adding `8` bytes of padding to the buffers all the times, instead of `align_up` to the nearest multiple of `8`.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1831,9 +1832,12 @@ auto convert_table_to_parquet_data(table_input_metadata& table_meta,
...
     (stats_granularity != statistics_freq::STATISTICS_NONE) ? num_pages + num_chunks : 0;
-  rmm::device_buffer uncomp_bfr(max_uncomp_bfr_size, stream);
-  rmm::device_buffer comp_bfr(max_comp_bfr_size, stream);
-  rmm::device_buffer col_idx_bfr(column_index_bfr_size, stream);
+
+  // Buffers need to be padded as the compute kernels require aligned data pointers.
+  rmm::device_buffer uncomp_bfr(cudf::detail::align_up(max_uncomp_bfr_size, 8 /*alignment*/),
+                                stream);
+  rmm::device_buffer comp_bfr(cudf::detail::align_up(max_comp_bfr_size, 8 /*alignment*/), stream);
+  rmm::device_buffer col_idx_bfr(column_index_bfr_size * 2, stream);
```
etseidl, line 1840: Was this causing issues? `column_index_bfr_size` should already have alignment padding built in.

ttnghia: Oh sorry, I tested with this variable, didn't find issue with it but forgot to change it back :smile:

etseidl:
> I just realize that there is a duplicate fix for Parquet (#13585), which implements a different approach than this: adding `8` bytes of padding to the buffers all the times, instead of `align_up` to the nearest multiple of `8`.

The approach here is superior so I closed #13585.

cpp/include/cudf/detail/utilities/alignment.hpp
```diff
@@ -42,5 +42,17 @@ T* align_ptr_for_type(void* destination)
...
 
+/**
+ * @brief Align a value up to the nearest multiple of the given alignment.
+ *
+ * @param value The input value to align
+ * @param alignment The amount of bytes to align, must be a power of 2
+ * @return The aligned value
+ */
+inline std::size_t align_up(std::size_t value, std::size_t alignment)
+{
```
karthikeyann, line 53: nit:
If `alignment` is going to be compile-time constant, it could be a template parameter and a `static_assert` could be used to enforce it as power of 2 during compilation.
If not, an assert statement to check for power of 2 would be helpful here.

harrism (commented):
Most of the use cases here are constructing a `device_buffer` with padding (not alignment, as the comments say).  I suggest creating a factory function `make_padded_device_buffer()` and replace calls to 
```
rmm::device_buffer buf(cudf::detail::align_up(size, 8 /* alignment */), stream);
```
with

```
auto buf = make_padded_device_buffer(size, padding, stream);
```

You may alternatively want to create a globally available `memory_resource` adaptor `padding_adaptor` which wraps the default memory resource, and that `make_padded_device_buffer` uses by default (and hence the `padding` argument can be optional.)

ttnghia:
@harrism I finished implementing `padding_memory_resource_adaptor` and the `rmm::device_buffer make_padded_device_buffer` factory function. However, something coming up in my mind. The output `rmm::device_buffer` can be output from some utility function and then `resize` at the caller. In such cases, we have little idea if that buffer is already padded or not, and won't know if it will automatically have its size padded during `resize`.

I think the better solution for this is to explicitly say "padding" in the buffer type. That means, we should have a new buffer type like `rmm::padded_device_buffer` that can be constructed similar to `rmm::device_buffer` but it will have its own private padding mechanism. When using it, we can directly infer from its type that any of its memory allocation will involve padding.

cpp/include/cudf/detail/utilities/alignment.hpp
```diff
@@ -42,5 +42,17 @@ T* align_ptr_for_type(void* destination)
...
 }
 
+/**
+ * @brief Align a value up to the nearest multiple of the given alignment.
+ *
+ * @param value The input value to align
+ * @param alignment The amount of bytes to align, must be a power of 2
+ * @return The aligned value
+ */
+inline std::size_t align_up(std::size_t value, std::size_t alignment)
```
bdice, line 52: Is this the same as `round_up_safe`? https://github.com/rapidsai/cudf/blob/54b81df4e189cf36aca306a73b93dd68a717a80c/cpp/include/cudf/detail/utilities/integer_utils.hpp#L47

cpp/src/io/avro/reader_impl.cu
```diff
@@ -206,7 +206,9 @@ rmm::device_buffer decompress_data(datasource& source,
     uint32_t const initial_blk_len = meta.max_block_size * 2 + (meta.max_block_size * 2) % 4096;
     size_t const uncomp_size       = initial_blk_len * meta.block_list.size();
 
-    rmm::device_buffer decomp_block_data(uncomp_size, stream);
+    // Buffer needs to be padded as the compute kernels require aligned data pointers.
+    rmm::device_buffer decomp_block_data(cudf::detail::align_up(uncomp_size, 8 /*alignment*/),
```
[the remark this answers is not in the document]
bdice: How did we know that 8 is the right amount of padding? Also, can we use a constant for this rather than hardcoding the value 8 many times?

davidwendt:
I would like to consider the device-buffer-padding work in a follow on PR. 
I think the current implementation is fine along with Bradley's comments to get the nightly builds to work again.

cpp/include/cudf/detail/utilities/alignment.hpp
vuule: I think this function could be useful if it took the number of alignment bits. Removes the power of 2 requirement and differentiates it from `round_up_safe`.

ttnghia: Sorry I'm confused. What is the conclusion here? Should `round_up_safe` be used instead of adding `align_up`?

davidwendt: I would prefer just using `round_up_safe` here and using a `constexpr` named variable for the round-up value instead of the  hardcoded 8 value.

ttnghia: Indeed, `round_up_safe` in this use case produces exactly the same output as `align_up`. So I'm going to switch to use it, as suggested by @davidwendt to get the bugs fixed ASAP. `padded_device_buffer` will be the follow up work.

harrism: Could have also used RMM's function?

ttnghia: Finished adoption of `round_up_safe` but still strugging to find a good place to put in the `constexpr` that @davidwendt recommended.

cpp/src/io/avro/reader_impl.cu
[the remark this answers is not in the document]
ttnghia: I believe that @vuule has some idea about why cuIO kernels require 8-bytes aligned pointers?

cpp/include/cudf/detail/utilities/alignment.hpp
davidwendt: The changes seem to correspond to temporary compression/decompression buffers. Perhaps `gpuinflate.hpp`? 
I'm ok defining it in each .cu file too if there is a chance the padding may be different for the different implementations.

ttnghia: > Could have also used RMM's function?

I  think we can. However, that is a detail function (`rmm::detail::align_up`) so I was reluctant to use it.

cpp/src/io/avro/reader_impl.cu
```diff
@@ -309,8 +310,10 @@ rmm::device_buffer decompress_data(datasource& source,
     CUDF_EXPECTS(status == nvcompStatus_t::nvcompSuccess,
                  "Unable to get scratch size for snappy decompression");
 
-    rmm::device_buffer scratch(temp_size, stream);
-    rmm::device_buffer decomp_block_data(uncompressed_data_size, stream);
+    // Buffer needs to be padded.
```
ttnghia, line 313:
```suggestion
    // Buffers need to be padded.
```

davidwendt (approved):

cpp/src/io/comp/gpuinflate.hpp
```diff
@@ -47,6 +47,16 @@ struct compression_result {
...
 
+/**
+ * @brief The value used for padding a data buffer such that its size will be multiple of this
+ * modulus.
+ *
+ * Padding is necessary for input/output buffers of several compression/decompression kernels.
+ * Such kernels operate on aligned data pointers, which require padding to the buffers so that the
+ * pointers can shift along theirs address space to satisfy their alignment requirement.
+ */
+constexpr std::size_t PADDING_MODULUS{8};
```
bdice, line 58: I weakly prefer a different name, I like “multiple” more than “modulus”:
```suggestion
constexpr std::size_t BUFFER_PADDING_MULTIPLE{8};
```

bdice (approved):
A few small comments. I am happy with this design overall, feel free to address as you see fit.

cpp/src/io/comp/gpuinflate.hpp
```diff
@@ -47,6 +47,16 @@ struct compression_result {
 
 enum class gzip_header_included { NO, YES };
 
+/**
+ * @brief The value used for padding a data buffer such that its size will be multiple of this
+ * modulus.
+ *
+ * Padding is necessary for input/output buffers of several compression/decompression kernels.
+ * Such kernels operate on aligned data pointers, which require padding to the buffers so that the
```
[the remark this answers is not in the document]
ttnghia: @vuule do you have any comment on this?

cpp/src/io/comp/gpuinflate.hpp
ttnghia: Sounds good, I've updated this. Though I was trying to shorten the name instead.

cpp/src/io/avro/reader_impl.cu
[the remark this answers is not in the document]
vuule: No idea. For me there's a gap between the reported issue and this solution, where I don't see why these buffers need the padding, and why this much padding. I don't see this information in the PR.

cpp/src/io/avro/reader_impl.cu
```diff
@@ -206,7 +205,9 @@ rmm::device_buffer decompress_data(datasource& source,
     uint32_t const initial_blk_len = meta.max_block_size * 2 + (meta.max_block_size * 2) % 4096;
     size_t const uncomp_size       = initial_blk_len * meta.block_list.size();
 
-    rmm::device_buffer decomp_block_data(uncomp_size, stream);
+    // Buffer needs to be padded.
+    rmm::device_buffer decomp_block_data(
+      cudf::util::round_up_safe(uncomp_size, BUFFER_PADDING_MULTIPLE), stream);
```
vuule, line 210: padding because of inflate_kernel (this is the output of this kernel)

cpp/src/io/avro/reader_impl.cu
```diff
@@ -309,8 +310,11 @@ rmm::device_buffer decompress_data(datasource& source,
     CUDF_EXPECTS(status == nvcompStatus_t::nvcompSuccess,
                  "Unable to get scratch size for snappy decompression");
 
-    rmm::device_buffer scratch(temp_size, stream);
-    rmm::device_buffer decomp_block_data(uncompressed_data_size, stream);
+    // Buffer needs to be padded.
+    rmm::device_buffer scratch(cudf::util::round_up_safe(temp_size, BUFFER_PADDING_MULTIPLE),
+                               stream);
+    rmm::device_buffer decomp_block_data(
+      cudf::util::round_up_safe(uncompressed_data_size, BUFFER_PADDING_MULTIPLE), stream);
```
vuule, line 317: this one is odd, there should be no padding requirement from nvcomp. Have you verified that padding is required in the snappy case?

cpp/src/io/avro/reader_impl.cu
```diff
@@ -524,6 +528,10 @@ table_with_metadata read_avro(std::unique_ptr<cudf::io::datasource>&& source,
         block_data        = rmm::device_buffer{buffer->data(), buffer->size(), stream};
       }
 
+      // Buffer needs to be padded.
+      block_data.resize(cudf::util::round_up_safe(block_data.size(), BUFFER_PADDING_MULTIPLE),
+                        stream);
+
```
vuule, line 534: padding because of inflate_kernel? (this is an input of this kernel)

cpp/src/io/orc/reader_impl.cu
```diff
@@ -1067,7 +1069,9 @@ table_with_metadata reader::impl::read(int64_t skip_rows,
           CUDF_EXPECTS(not is_stripe_data_empty or stripe_info->indexLength == 0,
                        "Invalid index rowgroup stream data");
 
-          stripe_data.emplace_back(total_data_size, stream);
+          // Buffer needs to be padded.
+          stripe_data.emplace_back(
+            cudf::util::round_up_safe(total_data_size, BUFFER_PADDING_MULTIPLE), stream);
```
vuule, line 1074: based on the failing test, this buffer is the input to the nvcomp snappy decompression.

cpp/src/io/orc/reader_impl.cu
```diff
@@ -319,7 +319,9 @@ rmm::device_buffer reader::impl::decompress_stripe_data(
   }
   CUDF_EXPECTS(total_decomp_size > 0, "No decompressible data found");
 
-  rmm::device_buffer decomp_data(total_decomp_size, stream);
+  // Buffer needs to be padded.
+  rmm::device_buffer decomp_data(
+    cudf::util::round_up_safe(total_decomp_size, BUFFER_PADDING_MULTIPLE), stream);
```
vuule, line 324: output of (presumed) nvcomp snappy decompression.

cpp/src/io/parquet/reader_impl_preprocess.cu
vuule, line 1: same as ORC, based on the test, padded buffers are in/out of the nvcomp snappy decompression

cpp/src/io/text/bgzip_data_chunk_source.cu
vuule, line 1: this one is about inflate_kernel

vuule (commented):
Left some notes to help track where the padding requirements originate
Looks like there are two sources - inflate_kernel and nvcomp snappy
the nvcomp one is surprising, we should follow up on it with the nvcomp team.

vyasr (approved):
Fixes look correct to me. Would also like to discuss the adaptor/device_buffer changes in a separate follow-up PR.

cpp/src/io/comp/gpuinflate.hpp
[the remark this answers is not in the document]
vuule: I think the current situation is that we don't know why this padding is needed. Maybe we can document as such to make it clear that a follow-up is warranted.
I like the current PR as it makes it easy to find places with padding and investigate after this fix is merged.

cpp/src/io/avro/reader_impl.cu
ttnghia: Why not? It is input to `nvcompBatchedSnappyDecompressAsync`. 
Indeed, we don't have any unit test for AVRO in libcudf thus I can't verify it.

cpp/src/io/comp/gpuinflate.hpp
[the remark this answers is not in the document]
ttnghia: I've added kernel names to this comment as mentioned by https://github.com/rapidsai/cudf/pull/13586#pullrequestreview-1493829239.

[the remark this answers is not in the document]
ttnghia: I've filed an issue for tracking the follow up work: https://github.com/rapidsai/cudf/issues/13605

cpp/src/io/avro/reader_impl.cu
vuule: Just checked, there are no alignment/padding requirements from nvcomp here. IMO this part of the change can be reverted.

Are cuDF tests useless in this context?

ttnghia: Then I'll revert this.

> Are cuDF tests useless in this context?
There is 0 (zero) C++ test for avro.

vuule: I was referring to Python tests, we have some limited coverage there.

cpp/src/io/orc/reader_impl.cu
vuule: Assuming the issue has nothing to do with snappy, maybe this padding is needed in the uncompressed case. @ttnghia do you know which test triggered on this one?

ttnghia: Reversing this will cause memory access issue at:
```
 at 0x430 in ...cpp/src/io/comp/gpuinflate.cu:1197:cudf::io::copy_uncompressed_kernel
```

cpp/src/io/avro/reader_impl.cu
ttnghia: We have `avro_reader_test` in Python but I don't know how to check it with `compute-sanitizer`.

cpp/src/io/orc/reader_impl.cu
vuule: Based on the repro (without combing through `copy_uncompressed_kernel`), my understanding is that it requires the input buffers to be padded. I would suggest to add this to places where buffers are padded for this reason, and also add a note to `gpu_copy_uncompressed_blocks` docs. I really want changes in this PR to be traceable to the root(ish) cause.

cpp/src/io/orc/reader_impl.cu
ttnghia: Issue from `gpuDecodeOrcColumnData`.

cpp/src/io/orc/reader_impl.cu
ttnghia: This buffer is related to issue from `copy_uncompressed_kernel`.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -440,8 +446,10 @@ int decode_page_headers(cudf::detail::hostdevice_vector<gpu::ColumnChunkDesc>& c
     }
   }
 
-  // Dispatch batches of pages to decompress for each codec
-  rmm::device_buffer decomp_pages(total_decomp_size, stream);
+  // Dispatch batches of pages to decompress for each codec.
+  // Buffer needs to be padded.
+  rmm::device_buffer decomp_pages(
```
ttnghia, line 451: All changed buffers in this file is due to `gpuDecodePageData`.

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -1831,8 +1832,13 @@ auto convert_table_to_parquet_data(table_input_metadata& table_meta,
   // Initialize data pointers in batch
   uint32_t const num_stats_bfr =
     (stats_granularity != statistics_freq::STATISTICS_NONE) ? num_pages + num_chunks : 0;
-  rmm::device_buffer uncomp_bfr(max_uncomp_bfr_size, stream);
-  rmm::device_buffer comp_bfr(max_comp_bfr_size, stream);
+
+  // Buffers need to be padded.
```
ttnghia, line 1836: Due to `gpuGatherPages`.

cpp/src/io/orc/reader_impl.cu
ttnghia: Done.

vuule (approved):

PR #13590, opened 2023-06-20
Handle nullptr return value from bitmask_or in distinct_count

### LINKED ISSUE #13576: [BUG] cudaErrorIllegalAddress on distinct_count and nunique for reduce
**Describe the bug**
I recently needed to get an approximate distinct count for a dynamic optimization I was doing in Spark.  As a part of this I tried to use nunique as a reduce aggregation and ran into cudaErrorIllegalAddress on anything more complicated than some very basic types. I then tried to use distinct_count for the same thing, because I wanted the result on the CPU and this made it very simple to do, but I hit the same problem.
 
I was doing a test with TPCDS data and the keys to various hash aggregates, but I also saw it in a few unit tests.  Generally the rule that I found, but don't really trust is that anything with more than one column in it or any value that is larger than 64-bits. So strings cause this, structs cause this, tables with more than one column has this show up.  Decimal_128 causes this.  Lots of things, but here is one example test that I wrote that also causes it to crash.

```
TEST_F(DistinctCount, SimpleStruct)
{
//            .column(5, 3, 3, 1, 1)
//            .column(5, 3, null, null, 5)
//            .column(1, 3, 5, 7, 9)

    auto a  = cudf::test::fixed_width_column_wrapper<int32_t>{5, 3, 3, 1, 1};
    auto b  = cudf::test::fixed_width_column_wrapper<int32_t>({5, 3, XXX, XXX, 5}, nulls_at({2, 3}));
    auto c  = cudf::test::fixed_width_column_wrapper<int32_t>{1, 3, 5, 7, 9};

    cudf::table_view input_table({a, b, c});
    cudf::distinct_count(input_table);
}
```

I am happy to provide more data that causes crashes if people want me to. In the short term I have worked around the issue by hashing the row before sending it to distinct_count, because all I really care about is an approximate count, not an exact one. But I don't want to put anything into production until this gets fixed.

I did run compute sanitizer on the failing test and got back a lot of issues, but this is the first one

```
========= Invalid __global__ read of size 4 bytes
=========     at 0x170 in void cuco::experimental::detail::insert_if_n<(int)1, (int)128, thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, cudf::detail::row_validity, cuda::__4::atomic<unsigned long, (cuda::std::__4::__detail::thread_scope)1>, cuco::experimental::static_set_ref<int, (cuda::std::__4::__detail::thread_scope)1, cudf::experimental::row::equality::device_row_comparator<(bool)0, cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>, cuco::experimental::linear_probing<(int)1, const cudf::detail::experimental::compaction_hash<cudf::experimental::row::hash::device_row_hasher<cudf::detail::default_hash, cudf::nullate::DYNAMIC>>>, cuco::experimental::detail::aow_storage_ref<(int)1, int, cuco::experimental::extent<unsigned long, (unsigned long)18446744073709551615>>, cuco::experimental::op::insert_tag>>(T3, long, T4, T5, T6 *, T7)
=========     by thread (0,0,0) in block (0,0,0)
=========     Address 0x0 is out of bounds
=========     and is 140591841148928 bytes before the nearest allocation at 0x7fde16c00000 of size 20 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x23adbc]
=========                in /usr/lib/x86_64-linux-gnu/libcuda.so.1
=========     Host Frame:__cudart1071 [0x85215b]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:cudaLaunchKernel_ptsz [0x891da8]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:unsigned long cuco::experimental::static_set<int, cuco::experimental::extent<unsigned long, 18446744073709551615ul>, (cuda::std::__4::__detail::thread_scope)1, cudf::experimental::row::equality::device_row_comparator<false, cudf::nullate::DYNAMIC, cudf::experimental::row::equality::nan_equal_physical_equality_comparator>, cuco::experimental::linear_probing<1, cudf::detail::experimental::compaction_hash<cudf::experimental::row::hash::device_row_hasher<cudf::detail::MurmurHash3_32, cudf::nullate::DYNAMIC> > const>, rmm::mr::stream_allocator_adaptor<default_allocator<char> >, cuco::experimental::aow_storage<1> >::insert_if<thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, cudf::detail::row_validity>(thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, thrust::counting_iterator<int, thrust::use_default, thrust::use_default, thrust::use_default>, cudf::detail::row_validity, cuco::experimental::cuda_stream_ref) [0x40476a]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:cudf::detail::distinct_count(cudf::table_view const&, cudf::null_equality, rmm::cuda_stream_view) [0x400cf6]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:cudf::distinct_count(cudf::table_view const&, cudf::null_equality) [0x400de5]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:DistinctCount_SimpleStruct_Test::TestBody() [0x20b045]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) [0x84c8ed]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:testing::Test::Run() [clone .part.0] [0x83bd76]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:testing::TestInfo::Run() [0x83c4aa]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:testing::TestSuite::Run() [clone .part.0] [0x83ccbb]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:testing::internal::UnitTestImpl::RunAllTests() [0x842e7a]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:testing::UnitTest::Run() [0x83c641]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:main [0x1de237]
=========                in gtests/DISTINCT_COUNT
=========     Host Frame:__libc_start_main [0x240b3]
=========                in /lib/x86_64-linux-gnu/libc.so.6
=========     Host Frame: [0x20636e]
=========                in gtests/DISTINCT_COUNT
```

This was on the latest 23.08 code.

**Expected behavior**
No crashes.

wence-:
## Description

If nulls should compare equal then we want to filter out rows for which all column entries are invalid (bitmask not set). If any column is not nullable, then bitmask_or returns an empty bitmask buffer (and a null count of zero) indicating that the returned bitmask is fully valid. When passed as a predicate to insert_if we get a null pointer dereference.

To avoid this, only run predicated insertion if the null count returned from bitmask_or is positive (which guarantees that the validity bitmask exists). This also avoids running predicated insertion when the predicate is always true.

Closes #13576.

## Checklist

### DIFF
--- a/cpp/src/stream_compaction/distinct_count.cu
+++ b/cpp/src/stream_compaction/distinct_count.cu
@@ -150,15 +150,23 @@ cudf::size_type distinct_count(table_view const& keys,
       stream.value()};
 
     auto const iter = thrust::counting_iterator<cudf::size_type>(0);
-    // when nulls are equal, insert non-null rows only to improve efficiency
+    // when nulls are equal, only insert those rows that are not all null to improve efficiency
     if (nulls_equal == null_equality::EQUAL and has_nulls) {
       thrust::counting_iterator<size_type> stencil(0);
+      // We must consider a row if any of its column entries is valid,
+      // hence OR together the validities of the columns.
       auto const [row_bitmask, null_count] =
         cudf::detail::bitmask_or(keys, stream, rmm::mr::get_current_device_resource());
-      row_validity pred{static_cast<bitmask_type const*>(row_bitmask.data())};
 
-      return key_set.insert_if(iter, iter + num_rows, stencil, pred, stream.value()) +
-             static_cast<cudf::size_type>(null_count > 0);
+      // Unless all columns have a null mask, row_bitmask will be
+      // null, and null_count will be zero. Equally, unless there is
+      // some row which is null in all columns, null_count will be
+      // zero. So, it is only when null_count is not zero that we need
+      // to do a filtered insertion.
+      if (null_count > 0) {
+        row_validity pred{static_cast<bitmask_type const*>(row_bitmask.data())};
+        return key_set.insert_if(iter, iter + num_rows, stencil, pred, stream.value()) + 1;
+      }
     }
     // otherwise, insert all
     return key_set.insert(iter, iter + num_rows, stream.value());
--- a/cpp/src/stream_compaction/distinct_count.cu
+++ b/cpp/src/stream_compaction/distinct_count.cu
@@ -150,7 +150,8 @@ cudf::size_type distinct_count(table_view const& keys,
       stream.value()};
 
     auto const iter = thrust::counting_iterator<cudf::size_type>(0);
-    // when nulls are equal, only insert those rows that are not all null to improve efficiency
+    // when nulls are equal, we skip hashing any row that has a null
+    // in every column to improve efficiency.
     if (nulls_equal == null_equality::EQUAL and has_nulls) {
       thrust::counting_iterator<size_type> stencil(0);
       // We must consider a row if any of its column entries is valid,
--- a/cpp/tests/stream_compaction/distinct_count_tests.cpp
+++ b/cpp/tests/stream_compaction/distinct_count_tests.cpp
@@ -274,6 +274,16 @@ TEST_F(DistinctCount, TableWithNull)
   EXPECT_EQ(10, cudf::distinct_count(input, null_equality::UNEQUAL));
 }
 
+TEST_F(DistinctCount, TableWithSomeNull)
+{
+  cudf::test::fixed_width_column_wrapper<int32_t> col1{{1, 2, 3, 4, 5, 6}, {1, 0, 1, 0, 1, 0}};
+  cudf::test::fixed_width_column_wrapper<int32_t> col2{{1, 1, 1, 1, 1, 1}};
+  cudf::table_view input{{col1, col2}};
+
+  EXPECT_EQ(4, cudf::distinct_count(input, null_equality::EQUAL));
+  EXPECT_EQ(6, cudf::distinct_count(input, null_equality::UNEQUAL));
+}
+
 TEST_F(DistinctCount, EmptyColumnedTable)
 {
   std::vector<cudf::column_view> cols{};


cpp/src/stream_compaction/distinct_count.cu
```diff
@@ -150,15 +150,23 @@ cudf::size_type distinct_count(table_view const& keys,
       stream.value()};
 
     auto const iter = thrust::counting_iterator<cudf::size_type>(0);
-    // when nulls are equal, insert non-null rows only to improve efficiency
+    // when nulls are equal, only insert those rows that are not all null to improve efficiency
```
davidwendt, line 153:
```suggestion
    // when nulls are equal, only insert those rows that are not all valid to improve efficiency
```
Did you mean `not all valid` here? An null bitmask means all rows are valid and not all null. The wording is a bit confusing here.

ttnghia: We have a table here, so "not all valid" here means all elements at the same row in all columns are valid.

revans2:
I ran my tests with this patch and I saw no crashes for any of TPCDS, so I am happy with it.

wence-:
> I ran my tests with this patch and I saw no crashes for any of TPCDS, so I am happy with it.

Thanks for confirming!

cpp/src/stream_compaction/distinct_count.cu
wence-: Updated the comment to clarify.

PointKernel (approved):
@wence- thanks for helping fix the issue! 

LGTM

PR #13595, opened 2023-06-20
Java support: Floating point order-by columns for RANGE window functions

mythrocks:
## Description
This is a JNI follow-up to #13512. This adds support in the Java CUDF API to use floating point types (`FLOAT32`/`FLOAT64`) as the order-by column for range-based window functions.

There are no API changes; only the implementation was modified to permit floating-point types for the operation. A test was added to ratify behaviour.

## Checklist


java/src/test/java/ai/rapids/cudf/TableTest.java
```diff
@@ -6264,6 +6264,108 @@ void testRangeWindowsWithDecimalOrderBy() {
     }
   }
 
+  /**
+   * Helper to get scalar for preceding == Decimal(value),
+   * with data width depending upon the order-by
+   * column index:
```
ttnghia, line 6270:
```suggestion
   * with data width depending upon the order-by column index:
```

java/src/test/java/ai/rapids/cudf/TableTest.java
```diff
@@ -6264,6 +6264,108 @@ void testRangeWindowsWithDecimalOrderBy() {
...
+      case 3: return Scalar.fromDouble(Double.valueOf(value));
+      default:
+        throw new IllegalStateException("Unexpected order by column index: "
+                + orderby_col_idx);
+    }
+  }
+
+  @Test
+  void testRangeWindowsWithFloatOrderBy() {
+
```
ttnghia, line 6288:
```suggestion
```

java/src/test/java/ai/rapids/cudf/TableTest.java
```diff
@@ -6264,6 +6264,108 @@ void testRangeWindowsWithDecimalOrderBy() {
...
+      // Columns 2-3 are order-by columns of type FLOAT32 and FLOAT64 respectively, with similarly ordered values.
+      // In the following loop, each float type is tested as the order-by column,
+      // producing the same results with similar range bounds.
+      for (int float_oby_col_idx = 2; float_oby_col_idx <= 3; ++float_oby_col_idx) {
+        try (Table sorted = unsorted.orderBy(OrderByArg.asc(0),
+                OrderByArg.asc(1),
+                OrderByArg.asc(float_oby_col_idx));
+             ColumnVector expectSortedAggColumn = ColumnVector.fromBoxedInts(7, 5, 1, 9, 7, 9, 8, 2, 8, 0, 6, 6)) {
+          ColumnVector sortedAggColumn = sorted.getColumn(4);
```
ttnghia, line 6309: Does `ColumnVector` needs to be wrapped in `try` to prevent memory leak?

mythrocks: Good catch. It does look like a leak. Testing the fix now.

mythrocks: Nope, my bad. This is correct already. Closing this leads to a double-close error.

This vector is closed when the `sorted` table is closed.

ttnghia (approved):

revans2 (approved):

mythrocks:
Thank you for the reviews, @ttnghia, @revans2. I've merged this change.

PR #13597, opened 2023-06-21
Fixed a memory leak caused by Exception thrown while constructing a ColumnView

razajafri:
## Description
Close unclosed buffers in case of any Exception that is thrown in the ColumnView constructor. The catch all Exception clause then caused a double-free corruption which was also fixed

## Checklist


java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -54,14 +54,17 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
...
-      viewHandle = 0;
-      throw ae;
+      success = true;
+    } finally {
+      if (!success) {
+        // offHeap state is null, so there is nothing to clean in offHeap
+        // delete ColumnView to avoid memory leak
+        deleteColumnView(viewHandle);
+        viewHandle = 0;
+      }
```
jlowe, line 67: Using a boolean seems unnecessary here, we simply need to cleanup if anything is thrown.  For example:
```suggestion
    } catch (Throwable t) {
      // delete ColumnView to avoid memory leak
      deleteColumnView(viewHandle);
      viewHandle = 0;
      throw t;
    }
```

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -79,13 +82,16 @@ protected ColumnView(ColumnVector.OffHeapState state) {
     type = DType.fromNative(ColumnView.getNativeTypeId(viewHandle), ColumnView.getNativeTypeScale(viewHandle));
     rows = ColumnView.getNativeRowCount(viewHandle);
     nullCount = ColumnView.getNativeNullCount(viewHandle);
+    boolean success = false;
```
jlowe, line 85: Same unnecessary boolean comment here.

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -3631,7 +3631,12 @@ public static Table fromPackedTable(ByteBuffer metadata, DeviceMemoryBuffer data
       }
       result = new Table(columns);
     } catch (Throwable t) {
-      ColumnView.cleanupColumnViews(columnViewAddresses, columns);
+      // NOTE: Don't clean the columnViewAddresses as they were already cleaned by ColumnView
```
jlowe, line 3634: This still can leak.  If we don't successfully complete converting all of the raw column view addresses into ColumnVectors then we leak all the views that weren't attempted for conversion.  Can probably fix this by leaving this as-is but instead of nulling the address _after trying to convert, we null it _before_ converting, that way we leave the cleanup of the one trying to be converted to `ColumnVecotr.fromViewWithContiguousAllocation` but all remaining ones can be cleaned up by `ColumnView.cleanupColumnViews` as before.

java/src/main/java/ai/rapids/cudf/ColumnVector.java
```diff
@@ -1728,4 +1728,34 @@ public static ColumnVector empty(HostColumnVector.DataType colType) {
...
+          }
+        }
+        for (long nativeHandle : nativeHandles) {
+          if (nativeHandle != 0) {
+            deleteCudfColumn(nativeHandle);
+          }
+        }
+      } catch (Throwable s) {
+        t.addSuppressed(s);
+      }
```
jlowe, line 1757: This try block needs to be wrapping each close() or deleteCudfColumn() call, otherwise the first close to throw causes all subsequent columns/views to leak.

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -3462,16 +3460,7 @@ public static GatherMap mixedLeftAntiJoinGatherMap(Table leftKeys, Table rightKe
...
-      for (int i = 0; i < ptrs.length; i++) {
-        ret[i] = new ColumnVector(ptrs[i]);
-        ptrs[i] = 0;
-      }
-    } catch (Throwable t) {
-      ColumnView.cleanupColumnViews(ptrs, ret);
-      throw t;
-    }
+    ColumnVector[] ret = ColumnVector.getColumnVectorsFromPointers(ptrs);
     return ret;
```
jlowe, line 3464:
```suggestion
    return ColumnVector.getColumnVectorsFromPointers(ptrs);
```

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -3551,16 +3540,7 @@ public ColumnVector[] convertToRows() {
...
```
jlowe, line 3544:
```suggestion
    return ColumnVector.getColumnVectorsFromPointers(ptrs);
```

java/src/main/java/ai/rapids/cudf/Scalar.java
```diff
@@ -677,25 +677,7 @@ public ColumnView[] getChildrenFromStructScalar() {
...
-        try {
-          child.close();
-        } catch (Exception suppressed) {
-          ex.addSuppressed(suppressed);
-        }
-      }
-      throw ex;
-    }
+    ColumnView[] children = ColumnView.getColumnViewsFromPointers(childHandles);
     return children;
```
jlowe, line 681:
```suggestion
    return ColumnView.getColumnViewsFromPointers(childHandles);
```

java/src/main/java/ai/rapids/cudf/Table.java
```diff
@@ -83,20 +83,18 @@ public Table(ColumnVector... columns) {
...
       long[] views = new long[columns.length];
       for (int i = 0; i < columns.length; i++) {
         views[i] = columns[i].getNativeView();
       }
       nativeHandle = createCudfTableView(views);
       this.rows = columns[0].getRowCount();
     } catch (Throwable t) {
-      ColumnView.cleanupColumnViews(cudfColumns, this.columns);
+      for (ColumnVector column: columns) {
+        column.close();
```
revans2, line 96: Don't you need all of the suppressed try/catch logic here too?

razajafri: Thanks. Updated!

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -818,23 +822,47 @@ public ColumnView[] splitAsViews(int... indices) {
...
+        t.addSuppressed(s);
+      }
       throw t;
     }
     return columnViews;
   }
 
-  static void cleanupColumnViews(long[] nativeHandles, ColumnView[] columnViews) {
-    for (ColumnView columnView: columnViews) {
+  static void cleanupColumnViews(long[] nativeHandles, ColumnView[] columnViews) throws Throwable {
```
jlowe, line 835: This method is only called in practice when cleaning up from an exception, so rather than adding the complexity of first-vs-subsequent exceptions, this could simply always take a Throwable parameter and simply add a suppressed exception for every error caught during cleanup.

razajafri: Good point. My assumption was that we were calling it from outside of the catch block.

razajafri:
@jlowe PTAL. I think I have addressed your concerns

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -674,11 +674,10 @@ public final ColumnVector[] slice(int... indices) {
       }
     } catch (Throwable t) {
       try {
-        cleanupColumnViews(nativeHandles, columnVectors);
-      } catch (Throwable s) {
-        t.addSuppressed(s);
+        cleanupColumnViews(nativeHandles, columnVectors, t);
+      } finally {
+        throw t;
```
jlowe, line 679: This should still have the catch instead of finally, otherwise if `cleanupColumnViews` throws (it shouldn't, but if it somehow did) then this will smash any exception thrown from `cleanupColumnViews`.

Comment applies to the other places this was done.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -673,8 +673,13 @@ public final ColumnVector[] slice(int... indices) {
...
     } catch (Throwable t) {
-      cleanupColumnViews(nativeHandles, columnVectors);
-      throw t;
+      try {
+        cleanupColumnViews(nativeHandles, columnVectors, t);
+      } catch (Throwable s) {
+        t.addSuppressed(s);
+      } finally {
+        throw t;
+      }
```
jlowe, line 682: Nit: There's no need for a finally block here, it's already catching everything.  Throwing from a `finally` which is normally not desired, since it has the smashing effect I mentioned earlier, so could change to this here and elsewhere this pattern appears.
```suggestion
      } catch (Throwable s) {
        t.addSuppressed(s);
      }
      throw t;
```

jlowe (approved):
Minor nit about catching everything yet still using finally, but otherwise lgtm.

PR #13603, opened 2023-06-21
Fix `localize(None)` for timezone-naive columns

shwina:
Closes https://github.com/rapidsai/cudf/issues/13601

### DIFF
--- a/python/cudf/cudf/core/_internals/timezones.py
+++ b/python/cudf/cudf/core/_internals/timezones.py
@@ -209,9 +209,8 @@ def delocalize(data: DatetimeColumn) -> DatetimeColumn:
     """
     if isinstance(data, DatetimeTZColumn):
         return data._local_time
-    else:
-        # already timezone-naive:
-        return data
+    # already timezone-naive:
+    return data
 
 
 def convert(data: DatetimeTZColumn, zone_name: str) -> DatetimeTZColumn:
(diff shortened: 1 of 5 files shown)


python/cudf/cudf/core/_internals/timezones.py
```diff
@@ -202,6 +202,14 @@ def localize(
     )
 
 
+def delocalize(data: DatetimeColumn) -> DatetimeColumn:
```
shwina, line 205:
```suggestion
def delocalize(data: DatetimeColumn) -> DatetimeColumn:
    """
    Convert a timezone-aware datetime column to a timezone-naive one.
    If the column is already timezone-naive, return it as is.
    """
```

galipremsagar (approved):

python/cudf/cudf/core/_internals/timezones.py
```diff
@@ -202,6 +202,18 @@ def localize(
...
+    """
+    Convert a timezone-aware datetime column to a timezone-naive one.
+    If the column is already timezone-naive, return it as is.
+    """
+    if isinstance(data, DatetimeTZColumn):
+        return data._local_time
+    else:
+        # already timezone-naive:
+        return data
```
bdice, line 214: I have a weak preference for avoiding unnecessary `else:` clauses, so that you always have a `return` at the indentation level of the function body. This is also [preferred by pylint](https://pylint.readthedocs.io/en/stable/user_guide/messages/refactor/no-else-return.html).
```suggestion
    # already timezone-naive:
    return data
```

bdice (approved):
One small suggestion, otherwise LGTM.

Issue #13608, opened 2023-06-22
[BUG] Exception while reading ORC file in cudf


ttnghia:
When reading the problematic file in cudf, an exception is thrown:
```
CUDF failure at: ../../src/io/orc/reader_impl.cu:326: No decompressible data found
```
However, it can be read just fine in other applications (such as Spark). So there should be something wrong in the ORC reader.

The issue shows up with the input file having this column:
```
cudf::list_view<cudf::struct_view<cudf::string_view,cudf::string_view,cudf::string_view,cudf::list_view<cudf::string_view>,>>:
Length : 757
[...]

PR #13609, opened 2023-06-22
Fix issue with no decompressed data in ORC reader

vuule:
## Description
Currently, the ORC reader assumes that if data in the stripes of the current level are not empty, the decompressed data will also not be empty.
However, there is a corner case where the stripe is empty, but data blocks are still compressed so they contain the compression header. In this case, decompressed data is empty even with non-empty compressed blocks. 

This PR removes the assertion in the reader to allow for this corner case. Also adds a short-circuit to `decompress_stripe_data` to return early if the decompressed data is empty.

Issue https://github.com/rapidsai/cudf/issues/13608

## Checklist
- [x] Verify to fix the issue.
- [X] Check with compute-sanitizer.
- [X] Test with spark-rapids integration tests.


vuule:
Need to add a test, but still not sure what corner case triggers the issue.

ttnghia:
Confirm that this fixes the issue.

cpp/src/io/orc/reader_impl.cu
```diff
@@ -317,9 +317,12 @@ rmm::device_buffer reader::impl::decompress_stripe_data(
     num_uncompressed_blocks += compinfo[i].num_uncompressed_blocks;
     total_decomp_size += compinfo[i].max_uncompressed_size;
   }
-  CUDF_EXPECTS(total_decomp_size > 0, "No decompressible data found");
+  CUDF_EXPECTS(
+    not((num_compressed_blocks + num_compressed_blocks > 0) and (total_decomp_size == 0)),
```
mythrocks, line 321: The cleverness of this bit is eluding me:
```c++
(num_compressed_blocks + num_compressed_blocks > 0) and ...
```
Surely `num_compressed_blocks > 0` suffices, no?

ttnghia, line 321: Should be this:
```suggestion
    not((num_compressed_blocks + num_uncompressed_blocks > 0) and (total_decomp_size == 0)),
```

ttnghia: Good catch. I also didn't realize that :D

mythrocks: Ok, that makes sense now.

vuule: the names are too similar at first glance :)
Sorry for the unintentional puzzle :D

ttnghia (approved):

bdice (approved):

PointKernel (approved):

PR #13612, opened 2023-06-23
Add XXHash_64 hash function to cudf 

davidwendt:
## Description
Add XXHash_64 hash function to libcudf

```
std::unique_ptr<column> xxhash_64(
  table_view const& input,  uint64_t seed,
  rmm::cuda_stream_view stream,  rmm::mr::device_memory_resource* mr);
```

## Checklist


cpp/src/hash/xxhash64.cu
```diff
@@ -0,0 +1,323 @@
...
+    return h;
+  }
+
+ private:
+  hash_value_type m_seed{};
+  static constexpr uint64_t prime1 = 11400714785074694791ul;
+  static constexpr uint64_t prime2 = 14029467366897019727ul;
+  static constexpr uint64_t prime3 = 1609587929392839161ul;
+  static constexpr uint64_t prime4 = 9650029242287828579ul;
+  static constexpr uint64_t prime5 = 2870177450012600261ul;
```
bdice, line 184: Use hexadecimal.

bdice:
@davidwendt Note that a bug was found in the cuCollections implementation here: https://github.com/NVIDIA/cuCollections/pull/326

davidwendt:
> @davidwendt Note that a bug was found in the cuCollections implementation here: [NVIDIA/cuCollections#326](https://github.com/NVIDIA/cuCollections/pull/326)

Already fixed in https://github.com/rapidsai/cudf/pull/13612/commits/a5a0d4bf49f82936737238276c9669b0e1155f53

cpp/src/hash/xxhash64.cu
```diff
@@ -0,0 +1,336 @@
...
+      return hasher(col.element<T>(row_index));
+    }
+
+    template <typename T, CUDF_ENABLE_IF(not column_device_view::has_element_accessor<T>())>
+    __device__ hash_value_type operator()(column_device_view const&,
+                                          size_type,
+                                          Nullate const,
+                                          hash_value_type const) const noexcept
+    {
+      CUDF_UNREACHABLE("Unsupported type for MurmurHash64");
```
bdice, line 286: Fix name.
```suggestion
      CUDF_UNREACHABLE("Unsupported type for XXHash_64");
```

cpp/tests/hashing/xxhash64_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <cudf_test/column_utilities.hpp>
+#include <cudf_test/column_wrapper.hpp>
+#include <cudf_test/iterator_utilities.hpp>
+#include <cudf_test/type_lists.hpp>
+
+using NumericTypesNoBools =
+  cudf::test::Concat<cudf::test::IntegralTypesNotBool, cudf::test::FloatingPointTypes>;
+
+template <typename T>
+class HashXX64TestTyped : public cudf::test::BaseFixture {};
```
bdice, line 31: Name the tests consistently with the function name.

cpp/tests/hashing/xxhash64_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+                                                     0,
+                                                     200000,
+                                                     128,
+                                                     std::numeric_limits<int32_t>::max(),
+                                                     std::numeric_limits<int32_t>::min(),
+                                                     std::numeric_limits<int32_t>::lowest()}};
+
+  auto const output = cudf::hashing::xxhash64(cudf::table_view({col1}));
+
+  // these were generated using the CPU compiled version of the cuco xxhash64 source
```
bdice, line 70: Can you independently verify these results with a 3rd party implementation of xxhash, maybe the Python library https://pypi.org/project/xxhash/? Just want to be sure we cover our bases since there are always weird bugs in hashing functions...

cpp/tests/hashing/xxhash64_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+TEST_F(HashXX64Test, StringType)
+{
+  auto col1 = cudf::test::strings_column_wrapper(
+    {"The",
+     "quick",
+     "brown fox",
+     "jumps over the lazy dog.",
+     "I am Jack's complete lack of null value",
+     "A very long (greater than 128 bytes/char string) to test a a very long string",
```
bdice, line 124: This is only 77 (or 75 if you remove the duplicate "a") chars long. Is that right?
```suggestion
     "A very long (greater than 128 bytes/char string) to test a very long string",
```

cpp/tests/hashing/xxhash64_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+TEST_F(HashXX64Test, TestFixedPoint)
+{
+  auto const col1 = cudf::test::fixed_point_column_wrapper<int32_t>(
+    {0, 100, -100, -999999999, 999999999}, numeric::scale_type{-3});
+  auto const output = cudf::hashing::xxhash64(cudf::table_view({col1}));
+
+  // these were generated using the CPU compiled version of the cuco xxhash64 source
+  // https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/detail/hash_functions/xxhash.cuh
+  // and passing the 'value' of each input (without the scale) as the decimal-type
```
bdice, line 163: I don't like that cudf hashes decimals the same way regardless of their scale. To me, this is a violation of "different values hash differently". Should we file an issue for discussion here? I think this is a bug but would like a second opinion. My proposal would be to combine the column scale into the hash of the value.

cpp/include/cudf/hashing.hpp
```diff
@@ -125,6 +126,24 @@ std::unique_ptr<column> md5(
...
+ * This function takes a 64-bit seed value and returns a column of type UINT64.
+ *
+ * @param input The table of columns to hash
+ * @param seed Optional seed value to use for the hash function
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ *
+ * @returns A column where each row is the hash of a row from the input
+ */
+std::unique_ptr<column> xxhash64(
```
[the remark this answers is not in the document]
davidwendt: I based this code on https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/detail/hash_functions/xxhash.cuh
in case that matters. I really could not follow the source code for the link you provided.

cpp/src/hash/xxhash64.cu
```diff
@@ -0,0 +1,336 @@
...
+        uint64_t k1 = getblock64(data, offset) * prime2;
+
+        k1 = cudf::detail::rotate_bits_left(k1, 31) * prime1;
+        h64 ^= k1;
+        h64 = cudf::detail::rotate_bits_left(h64, 27) * prime1 + prime4;
+      }
+    }
+
+    // remaining data can be processed in 4-byte chunks
+    if (((nbytes % 32) % 8) >= 4) {
```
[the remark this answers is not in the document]
davidwendt: I think the optimizer will figure this out. I'd rather the code much the source where it was taken from.
https://github.com/NVIDIA/cuCollections/blob/51f68cae3a957fe83de055f76666db18e46cc89d/include/cuco/detail/hash_functions/xxhash.cuh#L342

cpp/tests/hashing/xxhash64_test.cpp
davidwendt: > I don't like that cudf hashes decimals the same way regardless of their scale. To me, this is a violation of "different values hash differently". Should we file an issue for discussion here? I think this is a bug but would like a second opinion. My proposal would be to combine the column scale into the hash of the value.

I remember this was discussed when first added and this was the agreed up on implementation. I wasn't part of the discussion so I don't know if it is recorded anywhere. It may be worth creating a separate issue to reopen the discussion.

cpp/tests/hashing/xxhash64_test.cpp
davidwendt: > Can you independently verify these results with a 3rd party implementation of xxhash, maybe the Python library https://pypi.org/project/xxhash/? Just want to be sure we cover our bases since there are always weird bugs in hashing functions...

I was able to verify the results match the results from this Python library

cpp/include/cudf/hashing.hpp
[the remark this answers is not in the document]
bdice: @sleeepyjack Do you know which algorithm the cuCo implementation is based on?

cpp/src/hash/xxhash64.cu
```diff
@@ -0,0 +1,336 @@
...
+using hash_value_type = uint64_t;
+
+template <typename Key>
+struct XXHash_64 {
+  using result_type = hash_value_type;
+
+  constexpr XXHash_64() = default;
+  constexpr XXHash_64(hash_value_type seed) : m_seed(seed) {}
+
+  __device__ inline uint32_t getblock32(std::byte const* data, cudf::size_type offset) const
```
bdice, line 43: We might want to consider putting functions like this into the common hashing utilities. We repeat this in a few hashing functions now, I think.

cpp/tests/hashing/xxhash64_test.cpp
bdice, line 1: Rename to `xxhash_64_test.cpp`.

cpp/tests/hashing/xxhash64_test.cpp
bdice: Filed an issue for discussion. I don't recall the original conversation about this, it might have preceded my time. https://github.com/rapidsai/cudf/issues/13705

cpp/include/cudf/hashing.hpp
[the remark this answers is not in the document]
davidwendt: The code matches more to the cuCollections implementation and I foresee any updates would come from there as well. 
Regardless, I have no preference on the names we use here.

bdice (commented):
Still some open consideration about `XXH64` vs. `XXHash_64`. I am happy with either outcome so long as we're consistent, but weakly lean in favor of `XXH64` because it matches the upstream repo and the name chosen by other implementations. e.g. the [Python implementation uses `xxhash.xxh64`](https://github.com/ifduyue/python-xxhash).

cpp/src/hash/xxhash64.cu
bdice: https://github.com/rapidsai/cudf/issues/13706

cpp/include/cudf/hashing.hpp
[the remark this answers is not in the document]
bdice: Sounds fine to me.

cpp/src/hash/xxhash64.cu
```diff
@@ -0,0 +1,336 @@
...
+namespace cudf {
+namespace hashing {
+namespace detail {
+
+namespace {
+
+using hash_value_type = uint64_t;
+
+template <typename Key>
+struct XXHash_64 {
```
bdice, line 37: We're still mixing `xxhash64` and `XXHash_64` (with and without underscores). Let's align one way or the other.
```suggestion
struct XXHash_64 {
```

bdice (commented):
Pick one of the following and then I can approve:
- `xxhash64` (cudf API) and `XXHash64` (functor)
- `xxhash_64` (cudf API) and `XXHash_64` (functor)

cpp/include/cudf/hashing.hpp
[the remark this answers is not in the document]
sleeepyjack: > @sleeepyjack Do you know which algorithm the cuCo implementation is based on?

Oh wow, I found this comment by chance. Somehow I don't get notifications from the RAPIDS org since we moved to GH enterprise -.-

> I'm wondering why we chose to implement XXHash_64 rather than XXH3, if that's what is being used in cuCo.

My understanding is that XXH3 is the SSE2 version of XXH64. Since we don't use explicit vectorization, I think XXHash_64 or XXH64 is the right term.

cpp/include/cudf/hashing.hpp
```diff
@@ -144,6 +145,24 @@ std::unique_ptr<column> md5(
   rmm::cuda_stream_view stream        = cudf::get_default_stream(),
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
+/**
+ * @brief Computes the hash value of each row in the input set of columns
```
karthikeyann, line 149: please add detail about the hash algorithm used.

cpp/tests/hashing/xxhash_64_test.cpp
```diff
@@ -0,0 +1,175 @@
...
+  auto col2 = cudf::test::fixed_width_column_wrapper<T, int32_t>{
+    {-1, -1, 0, 2, 22, 1, 11, 12, 116, 32, 0, 42, 7, 62, 1, -22, 1, -22},
+    {1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0}};
+
+  auto output1 = cudf::hashing::xxhash_64(cudf::table_view({col1}));
+  auto output2 = cudf::hashing::xxhash_64(cudf::table_view({col2}));
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(output1->view(), output2->view());
+
+  output1 = cudf::hashing::xxhash_64(cudf::table_view({col1}), 7);
+  output2 = cudf::hashing::xxhash_64(cudf::table_view({col2}), 7);
```
karthikeyann, line 50: add a constexpr `7` as seed.

cpp/src/hash/xxhash_64.cu
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  /**
+   * @brief Computes the hash value of an element in the given column.
+   */
+  class element_hasher_adapter {
+   public:
+    template <typename T, CUDF_ENABLE_IF(column_device_view::has_element_accessor<T>())>
+    __device__ hash_value_type operator()(column_device_view const& col,
+                                          size_type row_index,
```
karthikeyann, line 269: question: why is `row_index` alone is not const?

cpp/include/cudf/hashing.hpp
davidwendt: Like what?

bdice: I think he means that the brief does not mention the name of the algorithm. However this is true for all the hash functions in libcudf currently. I would defer this change to #13706 where it can be done uniformly for all hash functions.

bdice (approved):

cpp/include/cudf/hashing.hpp
bdice: I noted this in #13706. I think this can be resolved.

karthikeyann (approved):

cpp/include/cudf/hashing.hpp
karthikeyann: Sounds good.

cpp/src/hash/xxhash_64.cu
```diff
@@ -0,0 +1,336 @@
...
+
+  constexpr XXHash_64() = default;
+  constexpr XXHash_64(hash_value_type seed) : m_seed(seed) {}
+
+  __device__ inline uint32_t getblock32(std::byte const* data, cudf::size_type offset) const
+  {
+    // Read a 4-byte value from the data pointer as individual bytes for safe
+    // unaligned access (very likely for string types).
+    auto block = reinterpret_cast<uint8_t const*>(data + offset);
+    return block[0] | (block[1] << 8) | (block[2] << 16) | (block[3] << 24);
```
sleeepyjack, line 48: This will always emit 4x pipelined `LDG.E.U8`. I wonder if we should add an extra path that performs a single `LDG.E.32` in case the pointer is aligned correctly.
Dumb question: When can the start of a string be not aligned to 4 bytes?

Instead of loading and shifting the result, a common pattern is to use a `memcpy` for this:
```cpp
uint32_t ret;
memcpy(&ret, block, sizeof(uint32_t));
return ret;
```

cpp/src/hash/xxhash_64.cu
```diff
@@ -0,0 +1,336 @@
...
+      for (; offset <= nbytes - 4; offset += 4) {
+        h64 ^= (getblock32(data, offset) & 0xfffffffful) * prime1;
+        h64 = rotate_bits_left(h64, 23) * prime2 + prime3;
+      }
+    }
+
+    // and the rest
+    if (nbytes % 4) {
+      while (offset < nbytes) {
+        h64 ^= (static_cast<uint8_t>(data[offset]) & 0xff) * prime5;
```
sleeepyjack, line 93: The canonical way of casting a `std::byte` to an arithmetic integer type is to use `std::to_integer` ([ref](https://en.cppreference.com/w/cpp/types/byte#Non-member_functions)). But that's just a personal preference.

```suggestion
        h64 ^= (std::to_integer<uint8_t>(data[offset]) & 0xff) * prime5;
```

cpp/src/hash/xxhash_64.cu
```diff
@@ -0,0 +1,336 @@
...
+        h64 = rotate_bits_left(h64, 11) * prime1;
+        ++offset;
+      }
+    }
+    return h64;
+  }
+
+  result_type __device__ compute_bytes(std::byte const* data, cudf::size_type const nbytes) const
```
sleeepyjack, line 101: Once we switch to libcu++ v2.1, we should use `cuda::std::span` instead of passing a raw pointer + size. This would also allow us to pass a static extent when possible, i.e., the size of the key is known at compile time, which allows for certain optimizations.

cpp/src/hash/xxhash64.cu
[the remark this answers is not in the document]
sleeepyjack: > I think the optimizer will figure this out. 
I thought so too, but looking at [this minimal reproducer](https://godbolt.org/z/6Yz7fabs6), the emitted SASS is differen.

cpp/src/hash/xxhash_64.cu
```diff
@@ -0,0 +1,336 @@
...
+namespace hashing {
+namespace detail {
+
+namespace {
+
+using hash_value_type = uint64_t;
+
+template <typename Key>
+struct XXHash_64 {
+  using result_type = hash_value_type;
```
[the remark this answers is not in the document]
davidwendt: It basically is in this context. I want `hash_value_type` and `result_type` to be the same type.
It just makes it easier to refer to it later in the code rather than re-coding every current usage of `hash_value_type` as `XXHash_64<T>::result_type`

cpp/src/hash/xxhash_64.cu
davidwendt: We have a `device_span` class I can use here for now.

cpp/src/hash/xxhash_64.cu
sleeepyjack: > A string is rarely allocated individually but usually part of a larger contiguous block of memory.

Good point! Let's leave this as-is for now then.

sleeepyjack (approved):

PR #13623, opened 2023-06-26
Add convert_dtypes API

shwina:
Adds [convert_dtypes](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.convert_dtypes.html) which in cuDF should do nothing more than return a copy.

## Checklist


python/cudf/cudf/core/indexed_frame.py
```diff
@@ -4898,6 +4898,23 @@ def rank(
...
+        convert_floating=True,
+        dtype_backend=None,
+    ):
+        """
+        Returns a copy of the Series/DataFrame.
+
+        Unlike Pandas, this method just returns a copy. All data
+        types in cuDF are nullable, so there is no need to convert.
+        """
+        return self.copy()
```
galipremsagar, line 4916: We might need a little bit of special handling for float columns that have `nan` values. 
```
>>> df
   a  b      c    d     e      f
0  1  x   True    h  10.0    NaN
1  2  y  False    i   NaN  100.5
2  3  z    NaN  NaN  20.0  200.0

>>> dfn = df.convert_dtypes()
>>> dfn
   a  b      c     d     e      f
0  1  x   True     h    10   <NA>
1  2  y  False     i  <NA>  100.5
2  3  z   <NA>  <NA>    20  200.0
```

In column `e`, it appears pandas is converting `nan` to `NA` - which in our case means we'll need to call `.nans_to_nulls()`

shwina: Nice catch!

shwina: I also notice that Pandas has the data-dependent behaviour of casting column `e` to `Int64` type :(

galipremsagar: Do you think it is a bug that `e` is converted to `Int64` and `f` is converted to `Float64`?

mroeschke: Looking at `convert_dtypes` in pandas, it looks intentional that the dtype "downcasting" occurs. And from the docstring

```
    convert_floating : bool, defaults True
        Whether, if possible, conversion can be done to floating extension types.
        If `convert_integer` is also True, preference will be give to integer
        dtypes if the floats can be faithfully casted to integers.
```

shwina: No, I think it's intentional in Pandas. Just annoying that we have to introspect the data if we want to match the behaviour.

galipremsagar: > Looking at `convert_dtypes` in pandas, it looks intentional that the dtype "downcasting" occurs. And from the docstring
> 
> ```
>     convert_floating : bool, defaults True
>         Whether, if possible, conversion can be done to floating extension types.
>         If `convert_integer` is also True, preference will be give to integer
>         dtypes if the floats can be faithfully casted to integers.
> ```

I see, @shwina then we might have to mimic this behavior - maybe have this behavior enabled only when `mode.pandas_compatible` is enabled? I'm for having it either way, no strong opinions.

shwina: I've gone ahead and tried to match Pandas behaviour more closely here.

galipremsagar (approved):

Issue #13357, opened 2023-05-16
[ENH] Indices are immutable by definition and should take advantage of that


wence-:
**Is your feature request related to a problem? Please describe.**

Indices advertise a bunch of properties (for example `is_monotonic_increasing`). These are implemented by inheriting from `Frame` or `SingleColumnFrame`. But those latter classes are _mutable_, and so their implementation of the property must recompute every time.

For an index, this is not the case, and so we lose some performance (for example, every time we do a slice `.loc` we'll run a libcudf kernel to check if the index is sorted).

**Describe the solution you'd like**

Index properties (especially where they return scalars, and therefore the memory footprint is negligible) should use `functools.cached_property` to provide their (immutable) properties.

**Describe alternatives you've considered**

None

vyasr:
Related to https://github.com/rapidsai/cudf/issues/9593 (the immutability is also mentioned there).

vyasr:
@mroeschke I think you've already done a good chunk of this work (ex. https://github.com/rapidsai/cudf/blob/c867802d8c640e91756309fe0419212955259077/python/cudf/cudf/core/index.py#L472). I think all that's left is verifying that we don't have multiple overloads colliding in the MRO, and also auditing for any other outstanding functions that need the change. WDYT?

PR #13625, opened 2023-06-27
Refactor `Index` search to simplify code and increase correctness

### LINKED ISSUE #8693: [BUG] `loc` behavior differs from pandas when a duplicated index is requested
**Describe the bug**
When we have a series with a duplicated index such as 

```python
x = cudf.Series([1,2,3,4], index=[0,1,1,2])
```

And we try and fetch the items corresponding to that duplicated index, we get just the first element rather than a series of all the elements corresponding to that index:

```python
x.loc[1]
# 2
```

See above

**Expected behavior**
We should get the same thing as pandas, basically all the rows corresponding to that index gathered. 
```python
x.to_pandas().loc[1]
# 1    2
# 1    3
# dtype: int64
```

 - Environment location: [Bare-metal]
 - Method of cuDF install: [source]

**Environment details**

**Additional context**

### LINKED ISSUE #12833: [BUG] `loc`-based indexing with slice ranges inconsistent with pandas
**Describe the bug**

When indexing a frame with `.loc` one can provide a slice range where the endpoints of the slice are not contained in the index. If the index is not sorted then both pandas and cudf agree that you receive an error.

```python
import cudf

df = cudf.DataFrame({"a": [1, 2, 3]}, index=[7, 0, 4])

df.loc[1:5] # => ValueError
df.to_pandas().loc[1:5] # => KeyError
```

Probably we should fix the distinction in the exception that is raised.

However, when the index _is_ sorted, behaviour starts to differ:

When the index is monotone increasing (repeated values are allowed), and the slice has a positive step, then things work
```python
import cudf
df = cudf.DataFrame({"a": [1, 2, 3]}, index=[0, 4, 7])

df.loc[1:5] == df.to_pandas().loc[1:5]
```

If the slice has negative stride, then cudf produces the wrong answer:

```python
df = cudf.DataFrame({"a": [1, 2, 3]}, index=[0, 4, 7])
df.loc[5:0:-1] 
#   a
# 7  3
df.to_pandas().loc[5:0:-1]
#    a
# 4  2
# 0  1
```

If the index is monotone decreasing, and the slice has positive step, then the behaviour matches if both slice endpoints are in the index, and the behaviour differs if _at least_ one of the slice end points is not in the index, in these circumstances, pandas always returns an empty dataframe, and cudf always produces a `ValueError`.
```python
df = cudf.DataFrame({"a": [1, 2, 3]}, index=[7, 4, 0])
df.loc[0:4] == df.to_pandas().loc[0:4] # empty dataframe
df.loc[1:4] # => ValueError
df.to_pandas().loc[1:4] # empty data frame
```

If the slice has negative step, then the behaviour also differs, but differently. If both slice endpoints are in the index, then cudf returns an empty dataframe, whereas pandas returns the (reordered) dataframe sliced by the endpoints
```python
df = cudf.DataFrame({"a": [1, 2, 3]}, index=[7, 4, 0])
df.loc[0:4:-1] # empty dataframe
df.to_pandas().loc[0:4:-1]
#    a
# 0  3
# 4  2
```
If _at least_ one of the slice endpoints is not in the index, then cudf produces a `ValueError` and pandas behaves as for the monotone increasing case.
```python
df = cudf.DataFrame({"a": [1, 2, 3]}, index=[7, 4, 0])
df.loc[0:5:-1] # => ValueError
df.to_pandas().loc[0:5:-1]
#    a
# 0  3
# 4  2
```

The above issues apply, _mutatis mutandis_ to almost all index types in almost all cases (although I have not exhaustively checked). Amusingly the monotone increasing case with positive step slice with endpoint not in the index _works_ for `FloatXXXIndex`.

**Expected behavior**

We should be consistent, both internally, and probably matching pandas where possible (raising errors otherwise).

wence-:
## Description

Much of the index-specific search code for label-based lookup only worked in the case where the index was sorted in ascending order and/or the requested slice had the same step sign as the index. To fix this, handle ascending and descending sorted indices specially, as well as refactoring to remove unused codepaths.

The core idea is to push `find_first` and `find_last` to being generic column methods (in doing so we remove the `closest` argument, but that was only used to produce pandas-incompatible index behaviour).

Lookup in indices then uses `get_slice_bound` (that can be specialised for index types) that uses first (if applicable) `searchsorted` and then `find_first/last`.

While we're here, since we now check the sortedness properties of Index objects, turn them into `@cached_property` (partially addressing the request of #13357).

- Closes #8693
- Closes #12833

## Checklist


wence-:
This is partially prep work for the followon to #13534 that will implement `loc`-getitem with the same approach.

python/cudf/cudf/core/_base_index.py
```diff
@@ -1432,6 +1432,61 @@ def rename(self, name, inplace=False):
             out.name = name
             return out
 
+    def indices_of(self, value) -> cudf.core.column.NumericalColumn:
```
shwina, line 1435: Do these need to be private (`_`) methods?

wence-: I suppose the `Column.indices_of` method can be "public", since I suppose that is not a stable API. But yes, probably the user-facing `Index` methods should be private.

python/cudf/cudf/core/index.py
```diff
@@ -1462,14 +1404,32 @@ def notna(self):
...
                 "removed in a future version.",
                 FutureWarning,
             )
-        return self._values.get_slice_bound(label, side, kind)
+        if side not in {"left", "right"}:
+            raise ValueError(f"Invalid side argument {side}")
+        if self.is_monotonic_increasing:
+            return self.searchsorted(label, side=side, ascending=True)
+        elif self.is_monotonic_decreasing:
+            return self.searchsorted(label, side=side, ascending=False)
```
wence-, line 1419: If `RangeIndex` offered a `searchsorted` method, then the `get_slice_bound` implementation could be promoted to the base class (a range index is always monotonic increasing or decreasing, so the `else` branch will never be hit).

shwina: Were you thinking of adding a `searchsorted` method to `RangeIndex` here?

wence-: I realised re-reading the code that I could do that, so I guess yes..

python/cudf/cudf/core/_base_index.py
wence-: Done.

python/cudf/cudf/core/index.py
wence-: Done.

shwina (approved):
Nice work!

PR #13626, opened 2023-06-27
Create separate libcudf hash APIs for each supported hash function 

davidwendt:
## Description
Deprecates `cudf::hash` in favor of individual hash functions in the new `cudf::hashing` namespace. This will help with adding new hash algorithms.

## Checklist


cpp/include/cudf/hashing.hpp
```diff
@@ -22,16 +22,23 @@
...
  * @file
  */
 
 /**
- *  @brief Identifies the hash function to be used
+ * @brief Type of hash value
+ *
+ * @deprecated Since 23.08
+ */
+using hash_value_type = uint32_t;
```
PointKernel, line 36: This deprecation would affect a number of places. Any suggestion on how we adapt related use cases?

davidwendt: That is true. I think this probably should be moved to a different header instead of existing here. 
I will look into this. The same consideration for the enum `hash_id` as well -- I found the enum is used by [`hash_partition()`](https://docs.rapids.ai/api/libcudf/stable/group__reorder__partition.html#ga65fe4ae8d1cbe9db31dbda43ebea0e7a) so it can't go away completely.

ttnghia: Why is it deprecated? Are we going to use another more "standard" type?

davidwendt: After further thought, I'm going to un-deprecate the `hash_value_type` and `hash_id` since they are used in other places besides the `cudf::hash` function.

cpp/include/cudf/hashing.hpp
```diff
@@ -63,5 +70,62 @@ std::unique_ptr<column> hash(
...
+  table_view const& input,
+  uint32_t seed                       = DEFAULT_HASH_SEED,
+  rmm::cuda_stream_view stream        = cudf::get_default_stream(),
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Computes the MurmurHash3_32 of each row in the given table
+ *
+ * This function computes the hash similar to MurmurHash3_32 with special processing
+ * to match Sparks implementation results.
```
bdice, line 99:
```suggestion
 * to match Spark's implementation results.
```

cpp/include/cudf/hashing.hpp
```diff
@@ -63,5 +70,62 @@ std::unique_ptr<column> hash(
...
+/**
+ * @brief Computes the MD5 hash of each row in the given table
+ *
+ * @param input The table of columns to hash
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ *
+ * @returns A column where each row is the hash of a row from the input
+ */
+std::unique_ptr<column> md5_hash(
```
bdice, line 123: Do we need the word “hash” in the function name if it’s in the hashing namespace?

cpp/src/hash/md5_hash.cu
```diff
@@ -153,7 +153,7 @@ struct MD5Hasher {
   };
 
   char* result_location;
-  hash_circular_buffer<message_chunk_size, md5_hash_step> buffer;
+  cudf::detail::hash_circular_buffer<message_chunk_size, md5_hash_step> buffer;
```
bdice, line 156: Some of these should move to the hashing namespace.

cpp/include/cudf/hashing.hpp
bdice: We’re using `hash_value_type` to mean “the return type of libcudf’s default hash function.” We already hardcode the result type as a member type of the hash function implementation, maybe we can reuse that information?

bdice: Like `using hash_value_type = cudf::detail::default_hash<int>::result_type`?

bdice: For hash_partition we still run into some of the same issues with API awkwardness and seeds — we can’t hash-partition by the string result of an MD5 hash, for example. And even if we could, it wouldn’t use a seed. Same for HASH_IDENTITY. We should assess the use cases and determine if we need to accept a hash_id in hash_partition or if that should be deprecated.

bdice: Another alternative would be to rename/restrict hash_id to be something that defines a subset of hash algorithms usable for this purpose.

cpp/include/cudf/hashing.hpp
ttnghia: Good point. I also think `md5` is enough to tell what it is.

bdice: This would align with Python’s `hashlib.md5` and PySpark’s `pyspark.sql.functions.md5`. The word hash is in the proper name of some hash functions (like XXHash or MurmurHash) but not all.

PointKernel (approved):
LGTM

cpp/include/cudf/hashing.hpp
```diff
@@ -63,5 +70,62 @@ std::unique_ptr<column> hash(
   rmm::cuda_stream_view stream        = cudf::get_default_stream(),
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
+namespace hashing {
+
+/**
+ * @brief Computes the MurmurHash3_32 of each row in the given table
```
karthikeyann, line 76: nit: `MurmurHash3_32` (MurmurHash3 32-bit)

ttnghia (approved):
LGTM, although I'm not 100% satisfied with the way we add the `detail::` namespace. But it is not just for this specific PR so I'm fine with it.

PR #13629, opened 2023-06-28
Expose streams in all public copying APIs

vyasr:
## Description

Contributes to #925. 

## Checklist


cpp/include/cudf/copying.hpp
```diff
@@ -378,6 +396,7 @@ std::unique_ptr<column> copy_range(
  * @param input      Column to be shifted
  * @param offset     The offset by which to shift the input
  * @param fill_value Fill value for indeterminable outputs
+ * @param stream CUDA stream used for device memory operations and kernel launches.
```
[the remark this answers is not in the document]
davidwendt: Sorry, but will you also be removing periods from all the other `stream` descriptions in this file for consistency?

cpp/tests/streams/copying_test.cpp
bdice, line 1: Is the intent that we're creating two copies of the same test suite, one with defaulted streams and one with explicit streams? I wasn't aware of this, so we ought to be sure that developers know that all tests need to be duplicated if that's the case.

cpp/src/transform/one_hot_encode.cu
```diff
@@ -62,12 +62,12 @@ std::pair<std::unique_ptr<column>, table_view> one_hot_encode(column_view const&
 {
   CUDF_EXPECTS(input.type() == categories.type(), "Mismatch type between input and categories.");
 
-  if (categories.is_empty()) { return std::pair(make_empty_column(type_id::BOOL8), table_view{}); }
+  if (categories.is_empty()) { return {make_empty_column(type_id::BOOL8), table_view{}}; }
```
bdice, line 65: For future PRs, let's keep the unrelated refactoring separate from stream changes.

cpp/include/cudf_test/column_wrapper.hpp
```diff
@@ -1595,7 +1596,12 @@ class lists_column_wrapper : public detail::column_wrapper {
     thrust::copy_if(
       std::cbegin(cols), std::cend(cols), valids, std::back_inserter(children), thrust::identity{});
 
-    auto data = children.empty() ? cudf::empty_like(expected_hierarchy) : concatenate(children);
+    // TODO: Once the public concatenate API exposes streams, use that instead.
```
bdice, line 1599: Can we track this TODO in an issue for stream APIs? I fear this will be lost.

cpp/include/cudf_test/column_wrapper.hpp
```diff
@@ -1595,7 +1596,12 @@ class lists_column_wrapper : public detail::column_wrapper {
...
+    auto data =
+      children.empty()
+        ? cudf::empty_like(expected_hierarchy)
+        : cudf::detail::concatenate(
+            children, cudf::test::get_default_stream(), rmm::mr::get_current_device_resource());
```
bdice, line 1604: We should be aware that the changes in this PR that call `detail` APIs rather than public APIs will no longer create NVTX ranges. If we intend to keep those NVTX ranges, we should track the detail/public changes in an issue or TODO and change them back to use public methods once the public methods accept streams.

cpp/include/cudf/copying.hpp
```diff
@@ -77,13 +77,15 @@ enum class out_of_bounds_policy : bool {
  * use `DONT_CHECK` when they are certain that the gather_map contains only valid indices for
  * better performance. If `policy` is set to `DONT_CHECK` and there are out-of-bounds indices
  * in the gather map, the behavior is undefined. Defaults to `DONT_CHECK`.
+ * @param stream CUDA stream used for device memory operations and kernel launches.
```
bdice, line 80: Echoing @davidwendt's request for consistency in the removal of periods (I saw that thread earlier but didn't notice it again when scrolling through the PR diff).

bdice (commented):
Mostly minor comments/requests. Happy to approve once we've discussed.

cpp/include/cudf/copying.hpp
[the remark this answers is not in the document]
vyasr: :facepalm: I didn't think and assumed this was a one-time fix. Sorry. Will fix the rest.

cpp/tests/streams/copying_test.cpp
vyasr: Not two full copies. The streams test is stripped down, just the bare minimum required to verify that each API call properly forwards the stream down. I already updated the libcudf developer documentation for this ([see the bottom of the latest nightly dev docs regarding testing](https://docs.rapids.ai/api/libcudf/nightly/md_developer_guide_testing)).

cpp/include/cudf/copying.hpp
vyasr: Fixed.

cpp/src/transform/one_hot_encode.cu
vyasr: Ah yeah sorry about that. I don't recall why I did this in this PR, I think it might have been an autofix from my LSP.

cpp/include/cudf_test/column_wrapper.hpp
vyasr: https://github.com/rapidsai/cudf/issues/13687

cpp/include/cudf_test/column_wrapper.hpp
vyasr: Indeed, however I think this is the only such change. All other places where I added a `detail::` are just to disambiguate calls that were previously unambiguous but the compiler now cannot distinguish automatically because the public and detail copying APIs now share the same signature except for the default stream value, which is not sufficient to prefer one over the other (i.e. now `allocate_like` could be `cudf::detail::allocate_like` or `cudf::allocate_like` if used inside a `namespace cudf::detail {` block).

vyasr:
@davidwendt apologies for the confusion. I've properly addressed your requests now I believe.

cpp/include/cudf_test/column_wrapper.hpp
bdice: Cool, I was hoping it was just disambiguation but wasn't 100% sure. I will resolve this.

bdice (approved):
Approving. Thanks!

cpp/src/transform/one_hot_encode.cu
mythrocks: I was about to applaud fixing this as a matter of course. :] Noted.

mythrocks (approved):

cpp/tests/streams/copying_test.cpp
```diff
@@ -0,0 +1,346 @@
...
+{
+  constexpr cudf::size_type source_size{1000};
+
+  auto data = cudf::detail::make_counting_transform_iterator(0, [](auto i) { return i; });
+  cudf::test::fixed_width_column_wrapper<int32_t> source_column(data, data + source_size);
+  cudf::test::fixed_width_column_wrapper<int32_t> gather_map(data, data + source_size);
+
+  cudf::table_view source_table({source_column});
+
+  std::unique_ptr<cudf::table> result = cudf::gather(source_table,
```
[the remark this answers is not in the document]
vyasr: Good question! There are no assertions in these tests because we aren't asserting any invariants. What we are doing is calling an API (e.g. `cudf::gather`) and then relying on some hidden machinery to verify that the same stream is actually used in every CUDA API call that is performed inside `cudf::gather`. See https://github.com/rapidsai/cudf/pull/13629#discussion_r1260335119 for more info (and a useful link).

Regarding the unused outputs, that's a good call. It's probably not causing any issues because I'm guessing we don't compile tests with warnings as errors, only libcudf itself. That said, it would be good to remove these. I'll fix.

cpp/tests/streams/copying_test.cpp
```diff
@@ -0,0 +1,342 @@
...
+#include "cudf/utilities/default_stream.hpp"
+#include "cudf/utilities/type_dispatcher.hpp"
+#include "cudf_test/iterator_utilities.hpp"
+#include <cudf/copying.hpp>
+#include <cudf/detail/null_mask.hpp>
+
+#include <cudf_test/base_fixture.hpp>
+#include <cudf_test/column_wrapper.hpp>
+#include <cudf_test/default_stream.hpp>
+#include <cudf_test/type_lists.hpp>
```
davidwendt, line 26:
```suggestion
#include <cudf_test/base_fixture.hpp>
#include <cudf_test/column_wrapper.hpp>
#include <cudf_test/default_stream.hpp>
#include <cudf_test/iterator_utilities.hpp>
#include <cudf_test/type_lists.hpp>

#include <cudf/copying.hpp>
#include <cudf/detail/null_mask.hpp>
#include <cudf/utilities/default_stream.hpp>
```
Prefer to use `<>` instead of `""` for these headers. 
Also, I don't think the `type_dispatcher.hpp` is needed.

cpp/include/cudf/copying.hpp
```diff
@@ -234,12 +244,14 @@ std::unique_ptr<column> empty_like(scalar const& input);
  * @param[in] input Immutable view of input column to emulate
  * @param[in] mask_alloc Optional, Policy for allocating null mask. Defaults to RETAIN
  * @param[in] mr Device memory resource used to allocate the returned column's device memory
+ * @param stream CUDA stream used for device memory operations and kernel launches
```
davidwendt, line 247: Could you remove the `[in]`s from the other `@param`s here (and a few other places where `stream` was added)?
I suppose the alternative is to add `[in]` to the `@param stream` tag but would prefer to remove the `[in]`s instead since it is clear these are all inputs.

davidwendt (changes requested):

cpp/tests/streams/copying_test.cpp
vyasr: Was able to clean the includes up even a bit further.

cpp/include/cudf/copying.hpp
vyasr: Done.

davidwendt (approved):

cpp/src/transform/one_hot_encode.cu
bdice: @mythrocks I like the change, just not in these API changing PRs. The public stream PRs are going to be large and touch lots of files, so it’s best to keep their scope tight.

Issue #13557, opened 2023-06-13
[ENH] Audit `argsort + gather/scatter` patterns for missing performance


wence-:
During review of #13419 we noted a few places where there is a pattern like:

```python
some_frame = ...
indices = some_frame.some_column.argsort()
new_frame = some_frame.take(indices)
```

As well as the unnecessary bounds-check (see #13456), this is a pattern that is captured by libcudf's [`sort_by_key`](https://docs.rapids.ai/api/libcudf/stable/group__column__sort.html#ga6db0403a43150b3bca0fbb9b2fbd68a3) and [`stable_sort_by_key`](https://docs.rapids.ai/api/libcudf/stable/group__column__sort.html#gaea04f441fe246b5a7e4f6420864024d4) functions (we would want to use the latter in pandas-compat mode).

At present, libcudf implements this as a `argsort` of the key columns followed by a gather. But that's an implementation detail (there may in the future be updates to that implementation). In the Python layer we should "say what we mean" and call into the appropriate libcudf API.

A cursory search shows:

```
/usr/bin/rg -U -e argsort.\*'     
'.\*take

core/_base_index.py
1305:        indices = self.argsort(ascending=ascending, na_position=na_position)
1306:        index_sorted = self.take(indices)

core/indexed_frame.py
2464:                # double-argsort to map back from sorted to unsorted positions
2465:                df = df.take(index.argsort(ascending=True).argsort())

core/column/column.py
1382:        order = order.take(left_gather_map, check_bounds=False).argsort()
1383:        codes = codes.take(order)

core/groupby/groupby.py
686:            gather_map = ordering.take(to_take).argsort()
687:            return result.take(gather_map)
```

Of these, the calls in `_base_index.py`, `_column.py`, and `groupby.py` can definitely be replaced by `sort_by_key`. (Note also that none of these calls pass `check_bounds=False` to `take` so incur an unnecessary kernel launch to check in-boundsness for something that is guaranteed in bounds.

The `take(argsort().argsort())` pattern is not a `sort_by_key`, however, we can elide one of the argsorts by noticing that `take` is a gather operation and for a permutation, the dual to gather is scatter. So this should be implemented as `df.scatter(index.argsort())` instead...

These are just the cases where an argsort is _immediately_ followed by a take, probably more diligent searching would find more.

bdice:
[...]

PR #13634, opened 2023-06-28
Expose stable versions of libcudf sort routines

wence-:
## Description

As preparation for addressing parts of #13557, this just exposes the existing sort-by and stable-sort-by routines from libcudf.

## Checklist

### DIFF
--- a/python/cudf/cudf/_lib/sort.pyx
+++ b/python/cudf/cudf/_lib/sort.pyx
@@ -161,10 +161,10 @@ def order_by(
         Columns from the table which will be sorted
     ascending : sequence[bool]
          Sequence of boolean values which correspond to each column
-         in the to-be-sorted table signifying order of each column
+         in the table to be sorted signifying the order of each column
          True - Ascending and False - Descending
     na_position : str
-        whether null value should show up at the "first" or "last"
+        Whether null values should show up at the "first" or "last"
         position of **all** sorted column.
     stable : bool
         Should the sort be stable? (no default)
@@ -252,10 +252,10 @@ def sort_by_key(
     keys : list[Column]
         Columns making up the sort key
     ascending : list[bool]
-        Sequence of boolean values which correspond to each column in
-        keys providing the sort order (default all True).
-        With True <=> ascending; False <=> descending.
-    null_precedence : list[str]
+        Sequence of boolean values which correspond to each column
+        in the table to be sorted signifying the order of each column
+        True - Ascending and False - Descending
+    na_position : list[str]
         Sequence of "first" or "last" values (default "first")
         indicating the position of null values when sorting the keys.
     stable : bool
--- a/python/cudf/cudf/_lib/sort.pyx
+++ b/python/cudf/cudf/_lib/sort.pyx
@@ -157,13 +157,17 @@ def order_by(
 
     Parameters
     ----------
-    columns_from_table : columns from the table which will be sorted
-    ascending : sequence of boolean values which correspond to each column
-                in source_table signifying order of each column
-                True - Ascending and False - Descending
-    na_position : whether null value should show up at the "first" or "last"
-                position of **all** sorted column.
-    stable : should the sort be stable? (no default)
+    columns_from_table : list[Column]
+        Columns from the table which will be sorted
+    ascending : sequence[bool]
+         Sequence of boolean values which correspond to each column
+         in the to-be-sorted table signifying order of each column
+         True - Ascending and False - Descending
+    na_position : str
+        whether null value should show up at the "first" or "last"
+        position of **all** sorted column.
+    stable : bool
+        Should the sort be stable? (no default)
 
     Returns
     -------
@@ -255,7 +259,7 @@ def sort_by_key(
         Sequence of "first" or "last" values (default "first")
         indicating the position of null values when sorting the keys.
     stable : bool
-        Should the sort be stable?
+        Should the sort be stable? (no default)
 
     Returns
     -------
@@ -312,9 +316,8 @@ def segmented_sort_by_key(
     null_precedence : list[str], optional
         Sequence of "first" or "last" values (default "first")
         indicating the position of null values when sorting the keys.
-
     stable : bool
-        Should the sort be stable?
+        Should the sort be stable? (no default)
 
     Returns
     -------
--- a/python/cudf/cudf/_lib/sort.pyx
+++ b/python/cudf/cudf/_lib/sort.pyx
@@ -149,7 +149,8 @@ def order_by(
     list columns_from_table,
     object ascending,
     str na_position,
-    bool stable=False
+    *,
+    bool stable
 ):
     """
     Get index to sort the table in ascending/descending order.
@@ -162,7 +163,7 @@ def order_by(
                 True - Ascending and False - Descending
     na_position : whether null value should show up at the "first" or "last"
                 position of **all** sorted column.
-    stable : should the sort be stable?
+    stable : should the sort be stable? (no default)
 
     Returns
     -------
@@ -234,7 +235,8 @@ def sort_by_key(
     list keys,
     object ascending,
     object na_position,
-    bool stable=False,
+    *,
+    bool stable,
 ):
     """
     Sort a table by given keys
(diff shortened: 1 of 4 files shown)


python/cudf/cudf/_lib/sort.pyx
```diff
@@ -153,6 +162,11 @@ def order_by(list columns_from_table, object ascending, str na_position):
                 True - Ascending and False - Descending
     na_position : whether null value should show up at the "first" or "last"
                 position of **all** sorted column.
+    stable : should the sort be stable?
```
bdice, line 165: We're not using proper numpydoc parameter formatting here... https://numpydoc.readthedocs.io/en/latest/format.html#parameters

But it's okay with me if that change is considered out-of-scope for this PR. 😕

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -208,13 +228,69 @@ def sort(
...
+    keys : list[Column]
+        Columns making up the sort key
+    ascending : list[bool], optional
+        Sequence of boolean values which correspond to each column in
+        keys providing the sort order (default all True).
+        With True <=> ascending; False <=> descending.
+    null_precedence : list[str], optional
+        Sequence of "first" or "last" values (default "first")
+        indicating the position of null values when sorting the keys.
+
```
bdice, line 255:
```suggestion
```

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -141,7 +145,12 @@ cdef pair[vector[order], vector[null_order]] ordering(
 
 
 @acquire_spill_lock()
-def order_by(list columns_from_table, object ascending, str na_position):
+def order_by(
```
bdice, line 148: I would weakly prefer two separate functions, one stable and one unstable. The function name explicitly carries this information in C++ and mirroring that in Cython would be ideal in my view. Otherwise you get confusing questions about what `True` or `False` means in the signature, and there's an undesirable "default" choice that will be inevitably misused.

Another option that gives it more verbosity would be to have a _required_ parameter with a string like `stable="stable"` / `stable="unstable"` but that's gross...

(Same for `[stable_]sort_by_key` below.)

bdice (changes requested):

python/cudf/cudf/_lib/sort.pyx
wence-: I can use required keyword args:

```python
def order_by(..., *, stable):
    ...
```
Then `order_by(..., stable=True)` or `order_by(...., stable=False)` are the only calls that are offered and are reasonably self-documenting at the call site. WDYT?

bdice: I’m fine with a required kwarg with no default.

wence-: Used kwargs, and made the callers default to `stable=True` if we are in pandas-compatible mode, and `stable=False` otherwise.

wence-:
As part of the `stable` keyword argument case, I realise that there is zero consistency in the handling of the `column_order`/`null_precedence` arguments. Some functions allow defaults, others don't. And the defaults in the cython wrapper match those of libcudf, but the null precedence argument defaults are opposite to the defaults in the higher-level cudf pandas wrappers.

In the interest of keeping this small, I'll open an issue.

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -141,18 +145,33 @@ cdef pair[vector[order], vector[null_order]] ordering(
...
-    na_position : whether null value should show up at the "first" or "last"
-                position of **all** sorted column.
+    columns_from_table : list[Column]
+        Columns from the table which will be sorted
+    ascending : sequence[bool]
+         Sequence of boolean values which correspond to each column
+         in the to-be-sorted table signifying order of each column
+         True - Ascending and False - Descending
+    na_position : str
+        whether null value should show up at the "first" or "last"
```
vyasr, line 167:
```suggestion
        Whether null values should show up at the "first" or "last"
```

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -141,18 +145,33 @@ cdef pair[vector[order], vector[null_order]] ordering(
...
-    ascending : sequence of boolean values which correspond to each column
-                in source_table signifying order of each column
-                True - Ascending and False - Descending
-    na_position : whether null value should show up at the "first" or "last"
-                position of **all** sorted column.
+    columns_from_table : list[Column]
+        Columns from the table which will be sorted
+    ascending : sequence[bool]
+         Sequence of boolean values which correspond to each column
+         in the to-be-sorted table signifying order of each column
```
vyasr, line 164:
```suggestion
         Sequence of boolean values which correspond to each column
         in the table to be sorted signifying the order of each column
```

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -208,13 +233,70 @@ def sort(
...
+    ----------
+    values : list[Column]
+        Columns of the table which will be sorted
+    keys : list[Column]
+        Columns making up the sort key
+    ascending : list[bool]
+        Sequence of boolean values which correspond to each column in
+        keys providing the sort order (default all True).
+        With True <=> ascending; False <=> descending.
+    null_precedence : list[str]
```
vyasr, line 258: This parameter is called `na_position` (although it is `null_precedence` in the C++ layer). Perhaps something we can align better in the future, for now I'm happy with any resolution that doesn't leave it the code/docs in an internally inconsistent state.

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -208,13 +233,70 @@ def sort(
...
+    Parameters
+    ----------
+    values : list[Column]
+        Columns of the table which will be sorted
+    keys : list[Column]
+        Columns making up the sort key
+    ascending : list[bool]
+        Sequence of boolean values which correspond to each column in
+        keys providing the sort order (default all True).
+        With True <=> ascending; False <=> descending.
```
vyasr, line 257: Maybe match the docstring from `order_by`? This one has very slightly different text.

vyasr (approved):
Looks good aside from some minor doc improvements so approving with a few suggestions.

python/cudf/cudf/_lib/sort.pyx
wence-: Done, thanks

python/cudf/cudf/_lib/sort.pyx
wence-: Done.

python/cudf/cudf/_lib/sort.pyx
wence-: Done.

python/cudf/cudf/_lib/sort.pyx
wence-: Done.

wence-:
Thanks for the proof-reading @vyasr. Any further issues @bdice?

python/cudf/cudf/core/frame.py
```diff
@@ -1580,7 +1580,13 @@ def _get_sorted_inds(self, by=None, ascending=True, na_position="last"):
         if np.isscalar(ascending):
             ascending = [ascending] * len(to_sort)
 
-        return libcudf.sort.order_by(to_sort, ascending, na_position)
+        # Only use a stable sort in pandas-compat mode
```
bdice, line 1583: I’d like to propose using stable sorts always. If I recall correctly there is no real performance difference in Thrust for these two calls. If there is no performance loss, I would prefer to ensure that we always match pandas regardless of compatibility mode. The compatibility mode flag is best used when it incurs a significant loss of performance in order to establish identical behavior. For cases with similar perf or only a small loss of perf, I would always match pandas. It’s one fewer thing for our users to worry about.

bdice (dismissed):
Proposing always using stable sort. Otherwise LGTM!

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -22,7 +22,11 @@ from cudf._lib.cpp.sorting cimport (
     rank,
     segmented_sort_by_key as cpp_segmented_sort_by_key,
     sort as cpp_sort,
+    sort_by_key as cpp_sort_by_key,
     sorted_order,
```
bdice, line 26: Do we need to make naming consistent with cpp_…?

wence-: For some reason the cython wrapper for `sorted_order` is called `order_by` (hence the lack of need to change).

PR #13635, opened 2023-06-28
Fix inf/NaN comparisons for FLOAT orderby in window functions

mythrocks:
## Description

This is a follow-up to #13512 (which added support for floating point order-by columns in window functions), and #13606 (which fixed how negative values are handled for floating point order-by).

This commit fixes how `NaN` and `+/- Infinity` values are handled for floating point.

Prior to this commit, the calculations for range window extents depended on the behaviour of `thrust::less<float>` and `thrust::greater<float>`, as well as addition/subtraction on `+/- Infinity`. This produced some unexpected results:
1. `thrust::less`/`greater` on `NaN` does not produce strict ordering.
2. Addition/Subtraction on the numerical values of `Infinity` could produce finite values that interfere with window extent calculations. Ideally, the results should have remained infinite.

This commit adds custom comparators with `NaN` awareness, to better handle columns that contain `NaN`s. It also fixes range calculations where `Infinity` is involved.

Tests have been added to cover ASC/DESC order sorting on `FLOAT`, with `NaN` and `Infinity` values.

## Checklist


cpp/src/rolling/grouped_rolling.cu
```diff
@@ -228,6 +228,9 @@ namespace {
 template <typename T, CUDF_ENABLE_IF(cuda::std::numeric_limits<T>::is_signed)>
 __device__ T add_safe(T const& value, T const& delta)
 {
+  if constexpr (cuda::std::numeric_limits<T>::has_infinity) {
```
vuule, line 231: Should this condition be `is_floating_point`? The branch also calls isnan, which is not explicitly tied to the condition.

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -388,6 +394,24 @@ std::unique_ptr<column> expand_to_column(Calculator const& calc,
...
+  __device__ bool operator()(T const& lhs, T const& rhs) const
+  {
+    return thrust::less<T>{}(lhs, rhs);
+  }
+
+  template <typename T, CUDF_ENABLE_IF(cudf::is_floating_point<T>())>
+  __device__ bool operator()(T const& lhs, T const& rhs) const
+  {
+    if (std::isnan(lhs)) { return !std::isnan(rhs); }
+    return std::isnan(rhs) ? true : lhs < rhs;
```
vuule, line 411: Returns true if one if the inputs (but not both) is a nan? That's surprising to me and I don't see how it allows strict ordering. Can you please add a comment? to doxygen above?

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -665,6 +697,24 @@ std::unique_ptr<column> range_window_ASC(column_view const& input,
...
+  __device__ bool operator()(T const& lhs, T const& rhs) const
+  {
+    return thrust::greater<T>{}(lhs, rhs);
+  }
+
+  template <typename T, CUDF_ENABLE_IF(cudf::is_floating_point<T>())>
+  __device__ bool operator()(T const& lhs, T const& rhs) const
+  {
+    if (std::isnan(lhs)) { return !std::isnan(rhs); }
+    return std::isnan(rhs) ? false : lhs > rhs;
```
vuule, line 714: Should we also call thrust::greater<T>{}(lhs, rhs) here for consistency with the other implementation?

vuule (commented):
few small comments. Mostly confused about ordering when one element is NaN.

cpp/src/rolling/grouped_rolling.cu
mythrocks: This is a fair point. I've been torn about this myself.

One option is to check for `has_infinity` for `infinity`, and `has_quiet_nan()` for `nan`. Another would be to check `std::numeric_limits<T>::is_iec559 == true`.

I think your suggestion might be most readable.

cpp/src/rolling/grouped_rolling.cu
mythrocks: Yes, I agree. Fixed.

cpp/src/rolling/grouped_rolling.cu
mythrocks: I've added documentation to explain this. `NaN` values should compare equal. But `NaN` values are deemed larger than `numeric_limits::infinity()`.

The [problem](https://godbolt.org/z/9fMM5Kr44) with using `thrust::less` is:
1. `thrust::less(NaN, 10)` returns `false`
2. `thrust::less(10, NaN)` also returns `false`

This seriously messes up the binary search (`lower_bound()`, `upper_bound()`). This comparator fixes comparisons involving `NaN`.

vuule: I still don't get it. If NaNs are supposed to be at the end of sorted arrays, why would `my_new_less(NaN, 10)` be true?

cpp/src/rolling/grouped_rolling.cu
mythrocks: Changed now.

cpp/src/rolling/grouped_rolling.cu
mythrocks: Please let me know if the added comment clarifies the requirement.

mythrocks: > why would `my_new_less(NaN, 10)` be true?

It isn't. `my_new_less(NaN, 10) == false`, just like `thrust::less<float>{}(NaN, 10)`. This is the way it should be, because `NaN < 10` is logically false.

Try the reverse:
1. Although `10` should be less than NaN, `10 < NaN` returns `false`. [Godbolt illustration](https://godbolt.org/z/Gnqq7Y1zM).
2. #1 above causes `thrust::less<float>{}(10, NaN)` to return `false`. This messes up binary search.
3. `my_new_less(10, NaN)` returns `true`.

mythrocks: (I wonder if I'm not grokking your question.)

vuule: I understand why (10, NaN) should and does return true (actual diff from thrust). The problem I see is the comparison in the other direction:
https://godbolt.org/z/YKaGcseWd
The operator returns true for(NaN, 10), but it should return false.

mythrocks: Oh, wow. Sorry I missed your point. It's a good one.

I'll examine the other side as well.

mythrocks: It appears the definition should be:
```c++
    if (std::isnan(lhs)) { return false; }
    return std::isnan(rhs) ? true : thrust::less<T>{}(lhs, rhs);
```
Does this look right, @vuule? It's a little confusing in that it's not symmetric.

I'm checking that this doesn't break a larger test.

mythrocks: All good. I have added tests for this specifically. The CUDF `ROLLING_TEST` and the Spark tests seem to pass.

mythrocks:
I've added tests for the `add/subtract_safe()` functions as well. Just in case.

cpp/src/rolling/detail/range_comparator_utils.cuh
```diff
@@ -0,0 +1,143 @@
...
+  __host__ __device__ bool operator()(T const& lhs, T const& rhs) const
+  {
+    return thrust::less<T>{}(lhs, rhs);
+  }
+
+  template <typename T, CUDF_ENABLE_IF(cudf::is_floating_point<T>())>
+  __host__ __device__ bool operator()(T const& lhs, T const& rhs) const
+  {
+    if (std::isnan(lhs)) { return false; }
+    return std::isnan(rhs) ? true : thrust::less<T>{}(lhs, rhs);
```
vuule, line 119: could be arguably simpler as 
```suggestion
    return std::isnan(rhs) or thrust::less<T>{}(lhs, rhs);
```

cpp/src/rolling/detail/range_comparator_utils.cuh
```diff
@@ -0,0 +1,143 @@
...
+           ? (value - delta)
+           : cuda::std::numeric_limits<T>::lowest();
+}
+
+/**
+ * @brief Comparator for numeric order-by columns, handling floating point NaN values.
+ *
+ * This is required for binary search through sorted vectors that contain NaN values.
+ * With ascending sort, NaN values are stored at the end of the sequence, even
+ * greater than infinity.
```
vuule, line 100: :+1:

vuule (approved):
Love the test coverage, thank you for taking the time to add the unit tests.

cpp/src/rolling/detail/range_comparator_utils.cuh
mythrocks: Right you are.
Plus, the condition in `greater` does not need change, since it's implemented in terms of `less`.

cpp/src/rolling/detail/range_comparator_utils.cuh
```diff
@@ -0,0 +1,143 @@
...
+  template <typename T, CUDF_ENABLE_IF(cudf::is_floating_point<T>())>
+  __host__ __device__ bool operator()(T const& lhs, T const& rhs) const
+  {
+    if (std::isnan(lhs)) { return false; }
+    return std::isnan(rhs) or thrust::less<T>{}(lhs, rhs);
+  }
+};
+
+/**
+ * @brief Comparator for numeric order-by columns, handling floating point NaN values. *
```
nvdbaranec, line 124:
```suggestion
 * @brief Comparator for numeric order-by columns, handling floating point NaN values.
```

cpp/tests/rolling/range_comparator_test.cu
```diff
@@ -0,0 +1,145 @@
...
+  using T = TypeParam;
+  EXPECT_EQ(cudf::detail::add_safe(T{3}, T{4}), T{7});
+
+  if constexpr (cuda::std::numeric_limits<T>::is_signed) {
+    EXPECT_EQ(cudf::detail::add_safe(T{-3}, T{4}), T{1});
+  }
+
+  auto constexpr max = cuda::std::numeric_limits<T>::max();
+  EXPECT_EQ(cudf::detail::add_safe(T{max - 5}, T{4}), max - 1);
+  EXPECT_EQ(cudf::detail::add_safe(T{max - 3}, T{4}), max);
```
hyperbolic2346, line 114: I'd like to see an on the nose test here too:
```
EXPECT_EQ(cudf::detail::add_safe(T{max - 4}, T{4}, max);
```

cpp/tests/rolling/range_comparator_test.cu
```diff
@@ -0,0 +1,145 @@
...
+  using T = TypeParam;
+  EXPECT_EQ(cudf::detail::subtract_safe(T{4}, T{3}), T{1});
+
+  if constexpr (cuda::std::numeric_limits<T>::is_signed) {
+    EXPECT_EQ(cudf::detail::subtract_safe(T{3}, T{4}), T{-1});
+  }
+
+  auto constexpr min = cuda::std::numeric_limits<T>::lowest();
+  EXPECT_EQ(cudf::detail::subtract_safe(T{min + 5}, T{4}), min + 1);
+  EXPECT_EQ(cudf::detail::subtract_safe(T{min + 3}, T{4}), min);
```
hyperbolic2346, line 136: Same on the nose test.

cpp/src/rolling/detail/range_comparator_utils.cuh
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * This is required for binary search through sorted vectors that contain NaN values.
+ * With descending sort, NaN values are stored at the beginning of the sequence, even
+ * greater than infinity.
+ * But thrust::greater would have trouble locating it because:
+ * 1. thrust::greater(NaN, 10) returns false
+ * 2. thrust::greater(10, NaN) also returns false
+ *
+ * This comparator honours the position of NaN values vis-à-vis non-NaN values.
```
hyperbolic2346, line 133: Too close to independence day to allow such language...
```suggestion
 * This comparator honors the position of NaN values vis-à-vis non-NaN values.
```

hyperbolic2346 (approved):
This is a nice change and the tests added are very welcome.

cpp/tests/rolling/range_comparator_test.cu
mythrocks: Good point. Fixed.

cpp/tests/rolling/range_comparator_test.cu
mythrocks: Likewise. Fixed.

cpp/src/rolling/detail/range_comparator_utils.cuh
mythrocks: Let the record show that I make this change under protest.

nvdbaranec (approved):
lgtm.  One small typo in comment block.

cpp/src/rolling/detail/range_comparator_utils.cuh
mythrocks: Oh, thanks for catching that.

mythrocks:
Thank you for the reviews, all. I have merged this change.

PR #13643, opened 2023-06-29
Fix memcheck error in ORC reader call to cudf::io::copy_uncompressed_kernel

davidwendt:
## Description
Fixes memcheck regression in ORC reader uncompressed logic. The regression was introduced in #13396.
The original fix is copied from #13586 

The error was caught by the nightly build here: https://github.com/rapidsai/cudf/actions/runs/5409203449/jobs/9829059618

## Checklist

### DIFF
--- a/cpp/src/io/orc/reader_impl.cu
+++ b/cpp/src/io/orc/reader_impl.cu
@@ -1122,6 +1122,8 @@ table_with_metadata reader::impl::read(uint64_t skip_rows,
         CUDF_EXPECTS(not is_stripe_data_empty or stripe_info->indexLength == 0,
                      "Invalid index rowgroup stream data");
 
+        // Buffer needs to be padded.
+        // Required by `copy_uncompressed_kernel`.
         stripe_data.emplace_back(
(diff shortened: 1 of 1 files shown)


cpp/src/io/orc/reader_impl.cu
```diff
@@ -1122,7 +1122,8 @@ table_with_metadata reader::impl::read(uint64_t skip_rows,
         CUDF_EXPECTS(not is_stripe_data_empty or stripe_info->indexLength == 0,
                      "Invalid index rowgroup stream data");
 
-        stripe_data.emplace_back(total_data_size, _stream);
+        stripe_data.emplace_back(
```
ttnghia, line 1125: Sorry, this should be fixed in https://github.com/rapidsai/cudf/pull/13586 but git decided to omit it somehow when I merged into https://github.com/rapidsai/cudf/pull/13396.
```suggestion
        // Buffer needs to be padded.
        // Required by `copy_uncompressed_kernel`.
        stripe_data.emplace_back(
```

ttnghia (approved):

bdice (approved):

PR #13645, opened 2023-06-29
[Java] Avoid double-free corruption in case of an Exception while creating a ColumnView

razajafri:
## Description
This PR addresses a case where an AssertionError will cause a double-free corruption. This was missed as part of https://github.com/rapidsai/cudf/pull/13597

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/ColumnView.java
+++ b/java/src/main/java/ai/rapids/cudf/ColumnView.java
@@ -50,11 +50,11 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
    */
   ColumnView(long address) {
     this.viewHandle = address;
-    this.type = DType.fromNative(ColumnView.getNativeTypeId(viewHandle), ColumnView.getNativeTypeScale(viewHandle));
-    this.rows = ColumnView.getNativeRowCount(viewHandle);
-    this.nullCount = ColumnView.getNativeNullCount(viewHandle);
-    this.offHeap = null;
     try {
+      this.type = DType.fromNative(ColumnView.getNativeTypeId(viewHandle), ColumnView.getNativeTypeScale(viewHandle));
+      this.rows = ColumnView.getNativeRowCount(viewHandle);
+      this.nullCount = ColumnView.getNativeNullCount(viewHandle);
+      this.offHeap = null;
       AssertEmptyNulls.assertNullsAreEmpty(this);
     } catch (Throwable t) {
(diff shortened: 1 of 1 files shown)


java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -669,8 +669,11 @@ public final ColumnVector[] slice(int... indices) {
     ColumnVector[] columnVectors = new ColumnVector[nativeHandles.length];
     try {
       for (int i = 0; i < nativeHandles.length; i++) {
-        columnVectors[i] = new ColumnVector(nativeHandles[i]);
+        long nativeHandle = nativeHandles[i];
+        // setting address to zero, so we don't clean it in case of an exception as it
+        // will be cleaned up by the constructor
         nativeHandles[i] = 0;
+        columnVectors[i] = new ColumnVector(nativeHandle);
```
revans2, line 676: So are we guaranteed that when ColumnVector/ColumnView gets the nativeHandle that it will close it on any exception? It looks like there are a few corner cases like OOM in java but also some native calls that could cause exceptions where this is not guaranteed. I'm not sure how much we need to worry about most of them, but it might be nice to fix that while we are here.  Specifically I am thinking that the ColumnView construction should move the try from line 82 up to line 78, right after we set the offHeap state.  The creation of the offHeapState I think we are okay with that failing and we leak a native handle.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -819,8 +822,11 @@ public ColumnView[] splitAsViews(int... indices) {
     ColumnView[] columnViews = new ColumnView[nativeHandles.length];
     try {
       for (int i = 0; i < nativeHandles.length; i++) {
-        columnViews[i] = new ColumnView(nativeHandles[i]);
+        long nativeHandle = nativeHandles[i];
+        // setting address to zero, so we don't clean it in case of an exception as it
+        // will be cleaned up by the constructor
         nativeHandles[i] = 0;
+        columnViews[i] = new ColumnView(nativeHandle);
```
revans2, line 829: Same comment here. I think we want to move the try up to line 53 for this constructor.

java/src/main/java/ai/rapids/cudf/ColumnView.java
razajafri: good point. I have updated

revans2 (approved):

PR #13648, opened 2023-06-30
Avoid storing metadata in pointers in ORC and Parquet writers

### LINKED ISSUE #12552: Replace pointers with std::optional in ORC, Parquet writer options
> Keeping pointers is very risky because of life time issues. I'd like to move away from this approach in the existing code as well.

_Originally posted by @vuule in https://github.com/rapidsai/cudf/pull/12474#discussion_r1070202706_
            

Replace usage of optional pointers with std::optional in ORC, Parquet writer options, and its builder.

https://github.com/rapidsai/cudf/blob/36b07f13480d8f7ba18274cb3a63f06855d848ff/cpp/include/cudf/io/orc.hpp#L811
https://github.com/rapidsai/cudf/blob/36b07f13480d8f7ba18274cb3a63f06855d848ff/cpp/include/cudf/io/parquet.hpp#L474

Use `std::optional` at all places

vuule:
## Description
Closes #12552

Replaced usage of pointers with `std::optional` in ORC and Parquet writer options. Avoids potential dangling pointer issues.

Breaking change because `set_metadata` and `metadata` setters now take an object, instead of a pointer.

## Checklist


cpp/tests/io/parquet_test.cpp
```diff
@@ -2870,7 +2870,7 @@ TEST_F(ParquetReaderTest, ReorderedColumns)
     md.column_metadata[0].set_name("a");
     md.column_metadata[1].set_name("b");
     cudf::io::parquet_writer_options opts =
-      cudf::io::parquet_writer_options::builder(cudf::io::sink_info{filepath}, tbl).metadata(&md);
+      cudf::io::parquet_writer_options::builder(cudf::io::sink_info{filepath}, tbl).metadata(md);
```
hyperbolic2346, line 2873: Some of these changed to `std::move(md)` and some are just `md`. Is there reasoning behind this? Should they all be moved? If so, maybe we should enforce that by taking an rvalue reference.

hyperbolic2346 (changes requested):
I like this change. It avoids a potential dangling pointer issue. My only concern is that it isn't abundantly clear to callers that you are attempting to take ownership of their object.

cpp/include/cudf/io/orc.hpp
```diff
@@ -649,7 +649,7 @@ class orc_writer_options {
    *
    * @param meta Associated metadata
    */
-  void set_metadata(table_input_metadata const* meta) { _metadata = meta; }
+  void set_metadata(table_input_metadata meta) { _metadata = std::move(meta); }
```
ttnghia, line 652: Maybe using `&&` is better since we can avoid copying if not needed.
```suggestion
  void set_metadata(table_input_metadata&& meta) { _metadata = std::move(meta); }
```

cpp/include/cudf/io/orc.hpp
```diff
@@ -779,9 +779,9 @@ class orc_writer_options_builder {
    * @param meta Associated metadata
    * @return this for chaining
    */
-  orc_writer_options_builder& metadata(table_input_metadata const* meta)
+  orc_writer_options_builder& metadata(table_input_metadata meta)
```
ttnghia, line 782: Similarly, use universal ref `&&`.

vuule:
Here's godbolt code that illustrates how this kind of API behaves with respect to the passed object https://godbolt.org/z/E4dbdc5sa
The interesting impact of passing by value is that we do:
- A copy + move with an lvalue object.
- Two moves with an rvalue object.
Hopefully this helps explain some details of the behavior:
- We don't move from a passed object if it has been passed as an lvalue (move semantics only enabled for rvalue, users don't experience any surprising behavior).
- We don't incur a copy when the object is passed as an rvalue (no need to explicitly take an rvalue ref, as suggested in the comments).

This is the approach we use to pass other objects into the cuIO APIs. The only downside is that it incurs an extra move ctor. It does allow (almost) optimal ingest with a single function (rather than having lvalue ref and rvalue ref overloads).

hyperbolic2346 (approved):

ttnghia (approved):

cpp/tests/io/orc_test.cpp
```diff
@@ -1462,7 +1462,7 @@ TEST_F(OrcReaderTest, NestedColumnSelection)
   auto filepath = temp_env->get_temp_filepath("OrcNestedSelection.orc");
   cudf::io::orc_writer_options out_opts =
     cudf::io::orc_writer_options::builder(cudf::io::sink_info{filepath}, expected)
-      .metadata(&expected_metadata);
+      .metadata(std::move(expected_metadata));
```
wence-, line 1465: And then `std::move` from here on?

python/cudf/cudf/_lib/cpp/io/orc.pxd
```diff
@@ -69,7 +69,7 @@ cdef extern from "cudf/io/orc.hpp" \
         size_type get_stripe_size_rows() except +
         size_type get_row_index_stride() except +
         cudf_table_view.table_view get_table() except +
-        const cudf_io_types.table_input_metadata *get_metadata() except +
+        const cudf_io_types.table_input_metadata& get_metadata() except +
```
wence-, line 72: AIUI, the C++ interface now returns a `std::optional<table_input_metadata>`, how does this work?


I _suspect_ it is because these Cython forward-declarations are only used for typing the transpilation and so this will (if it is used) cause some C++ to be generated that will then fail to compile.

As it happens, I think `get_metadata` is never called anywhere in the Cython codebase, so therefore things "work".

So maybe
```cython
from libcpp.optional cimport optional

...

const optional[cudf_io_types.table_input_metadata]& get_metadata() except +
```
?

python/cudf/cudf/_lib/cpp/io/parquet.pxd
```diff
@@ -76,7 +76,7 @@ cdef extern from "cudf/io/parquet.hpp" namespace "cudf::io" nogil:
             vector[cudf_io_types.partition_info] partitions
```
wence-, line 76: Line 68 (which provide the forward-decl for `parquet_writer_options.get_metadata()` also wants changed I think.

python/cudf/cudf/_lib/cpp/io/parquet.pxd
```diff
@@ -153,15 +153,15 @@ cdef extern from "cudf/io/parquet.hpp" namespace "cudf::io" nogil:
         cudf_io_types.sink_info get_sink() except +
         cudf_io_types.compression_type get_compression() except +
         cudf_io_types.statistics_freq get_stats_level() except +
-        cudf_io_types.table_input_metadata* get_metadata(
+        cudf_io_types.table_input_metadata& get_metadata(
```
wence-, line 156: Same comment about std::optional here.

python/cudf/cudf/_lib/cpp/io/orc.pxd
```diff
@@ -119,7 +119,7 @@ cdef extern from "cudf/io/orc.hpp" \
         size_type stripe_size_rows() except +
         size_type row_index_stride() except +
         cudf_table_view.table_view get_table() except +
-        const cudf_io_types.table_input_metadata *get_metadata(
+        const cudf_io_types.table_input_metadata& get_metadata(
```
wence-, line 122: std::optional here too

cpp/tests/io/orc_test.cpp
```diff
@@ -1432,7 +1432,7 @@ TEST_F(OrcWriterTest, TestMap)
   auto filepath = temp_env->get_temp_filepath("MapColumn.orc");
   cudf::io::orc_writer_options out_opts =
     cudf::io::orc_writer_options::builder(cudf::io::sink_info{filepath}, expected)
-      .metadata(&expected_metadata);
+      .metadata(expected_metadata);
```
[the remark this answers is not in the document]
vuule: Ah, I forgot to reply to @hyperbolic2346 's question about this :(
Since the `metadata` API now takes the object by value, we can avoid a copy if we pass a temporary (rvalue). This can be done even with a variable (lvalue) with the use of `std::move`, which casts the object into rvalue so it can be moved from. 
This move (move ctor, not `std::move`) puts the passed object into an invalid state, so we only use this when the variable is not used afterwards. In some tests we use the metadata object for comparison after round-tripping the table, so I could not use `std::move` there. In tests where the metadata object is not used after it's passed to `metadata()`, I used `std::move` to avoid making a copy.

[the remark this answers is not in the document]
wence-: Aha, thanks!

python/cudf/cudf/_lib/cpp/io/orc.pxd
vuule: nice catch! Fixed all four getters.

wence- (approved):
Thanks!

Issue #12822, opened 2023-02-22
[FEA] CEC for CUDA 12: using nvjitlink (instead of ptxcompiler / cubinlinker)


sevagh:
cuDF uses ptxcompiler and cubinlinker for CEC. Will this be changing for CUDA 12? We should discuss.

@bdice @gmarkall @vyasr

bdice:
In my understanding, we have (at least) two options:

1. Create ptxcompiler/cubinlinker packages for CUDA 12 that do "nothing" (or at least less work than the current packages by leveraging nvJitLink?) -- then we don't have to make CUDA-version-dependent lists of dependencies for cuDF.
2. Alter cuDF in a way that doesn't depend on ptxcompiler/cubinlinker packages when building for CUDA 12 (but this may require some kind of nvJitLink support built into cuDF?). This requires CUDA version knowledge when determining cuDF dependencies.

Happy to be corrected if anything above is inaccurate.

vyasr:
My understanding was that this wouldn't be a build-time issue at all, but that numba would "just work" when using CUDA 12 JITted kernels across different minor CUDA versions (probably by using nvJitLink under the hood). I thought that all cudf would need to do is to make the cubinlinker and ptxcompiler requirements conditional on the CUDA version.

sevagh:
The only issue I encountered when test-driving CUDA 12 pip wheel builds for cuDF (i.e. cudf-cu12, dask-cudf-cu12) was in the test step, where the install of cudf-cu12 couldn't find ptxcompiler-cu12, cubinlinker-cu12 to install from its requirements.

So, if the only change for CUDA 12 is to delete those lines (perhaps using `ci/release/apply_wheel_modifications.sh`, Vyas' new mechanism for modifying setuptools/python package files), that's fine too.

vyasr:
If simply removing the requirements makes everything work then yeah we could use that script (although it wouldn't be enough by itself, we would also need to do something for conda packages.

bdice:
Conda packages are easier because we can template the `meta.yaml` on the `RAPIDS_CUDA_VERSION` environment variable. That will already be necessary to make CUDA 11/12 compatible sets of CUDA dependencies.

vyasr:
Yup I agree, it's not hard. Just one more thing to do in addition to what we'd do for wheels.

jakirkham:
cc @gmarkall

jakirkham:
(bump)

vyasr:
[...]

PR #13650, opened 2023-06-30
Use `pynvjitlink` for CUDA 12+ MVC

brandon-b-miller:
Fixes https://github.com/rapidsai/cudf/issues/12822

This PR provides minor version compatibility in the CUDA 12.x range through `nvjitlink` via the preliminary [nvjiitlink python binding](https://github.com/gmarkall/nvjitlink). Thus far this PR merely leverages a local installation of the library and should not be merged until `nvjitlink` is hosted on `conda-forge` and cuDF's dependencies are adjusted accordingly, likely as part of this PR.


jakirkham:
Added a "Fixes" note in the OP. @brandon-b-miller please double check if that is correct

python/cudf/cudf/utils/_numba.py
```diff
@@ -7,6 +7,19 @@
 
 from numba import config as numba_config
 
+try:
+    from pynvjitlink.patch import (
+        patch_numba_linker as patch_numba_linker_cuda_12,
+    )
+except ModuleNotFoundError:
```
bdice, line 14: I think you can use the more generic `ImportError` (parent class of `ModuleNotFoundError`).
```suggestion
except ImportError:
```

python/cudf/cudf/utils/_numba.py
```diff
@@ -171,6 +174,7 @@ def _get_cuda_version_from_ptx_file(path):
         "7.8": (11, 8),
         "8.0": (12, 0),
         "8.1": (12, 1),
+        "8.2": (12, 2),
```
bdice, line 177: Can we add 12.3 here, too?

python/cudf/cudf/tests/test_numba_import.py
```diff
@@ -19,10 +19,10 @@
 TEST_NUMBA_MVC_ENABLED = """
 import numba.cuda
 import cudf
-from cudf.utils._numba import _CUDFNumbaConfig, _patch_numba_mvc
+from cudf.utils._numba import _CUDFNumbaConfig, patch_numba_linker_cuda_11
 
 
-_patch_numba_mvc()
+patch_numba_linker_cuda_11()
```
bdice, line 25: This seems to be unconditionally using CUDA 11 patching logic in tests. We should use CUDA 12 patching logic when on CUDA 12, right?

bdice (changes requested):

python/cudf/cudf/tests/test_numba_import.py
brandon-b-miller: I think this whole test skips in a CUDA 12 environment based on the below
```
@pytest.mark.skipif(
    not IS_CUDA_11, reason="Minor Version Compatibility test for CUDA 11"
)
```

bdice: The MVC tests should be enabled, even if we don't have an MVC-triggering environment in CI with the drivers/runtimes we use. We want to be able to run `pytest` on a system requiring MVC and know that it's being covered by cudf.

bdice: It's okay to have them be skipped if on CUDA 12 `and` MVC is needed `and` pynvjitlink isn't available.

brandon-b-miller: Ok, that makes sense. I'm thinking I'll rename this file `test_mvc` and have a separate test for cuda 11 and cuda 12. updates to follow.

brandon-b-miller: did some refactoring in https://github.com/rapidsai/cudf/pull/13650/commits/35b7e59e002c7c1bf07eb61063cb86a7c61c6577

python/cudf/cudf/tests/test_mvc.py
```diff
@@ -0,0 +1,97 @@
...
+    from ptxcompiler.patch import safe_get_versions
+except ModuleNotFoundError:
+    from cudf.utils._ptxcompiler import safe_get_versions
+
+# do not test cuda 12 if pynvjitlink isn't present
+HAVE_PYNVJITLINK = False
+try:
+    import pynvjitlink  # noqa: F401
+
+    HAVE_PYNVJITLINK
```
bdice, line 19:
```suggestion
    HAVE_PYNVJITLINK = True
```

python/cudf/cudf/utils/_numba.py
```diff
@@ -7,6 +7,19 @@
 
 from numba import config as numba_config
 
+try:
+    from pynvjitlink.patch import (
+        patch_numba_linker as patch_numba_linker_cuda_12,
+    )
+except ImportError:
+
+    def patch_numba_linker_cuda_12():
```
bdice, line 16: Do we expect that future major versions of CUDA will also use pynvjitlink? Do we want the naming of these functions to reflect "nvjitlink" rather than "cuda_12"?
```suggestion
    def patch_numba_linker_pynvjitlink():
```

python/cudf/cudf/utils/_numba.py
```diff
@@ -65,7 +78,7 @@ def _get_ptx_file(path, prefix):
         return regular_result[1]
 
 
-def _patch_numba_mvc():
+def patch_numba_linker_cuda_11():
```
bdice, line 81: Maybe this should be named by the package being used rather than the major version? But I'm okay with this one saying CUDA 11 since it really is specific to that major version.
```suggestion
def patch_numba_linker_cubinlinker():
```

bdice (commented):
A few cleanups -- but this is looking close to right. Have you verified that the tests run locally as expected with pynvjitlink?

brandon-b-miller:
> A few cleanups -- but this is looking close to right. Have you verified that the tests run locally as expected with pynvjitlink?

For me at least tests pass locally building from source with a 12.0 driver and 12.3 CTK. I'll also do a round of follow up testing on the tip of 23.12 after the merge.

python/cudf/cudf/utils/_numba.py
```diff
@@ -7,6 +7,19 @@
...
+try:
+    from pynvjitlink.patch import (
+        patch_numba_linker as patch_numba_linker_pynvjitlink,
+    )
+except ImportError:
+
+    def patch_numba_linker_pynvjitlink():
+        warnings.warn(
+            "Minor version compatibility requires pynvjitlink. "
+            "Please pip install pynvjitlink to proceed using cuDF."
```
bdice, line 19: Depending on the timing of pynvjitlink being shipped, we'll want to do one of the following:
- If pynvjitlink isn't shipped by the time we release, we should remove this note and say something like before (explain the MVC problem and direct users to match their driver and toolkit)
- If pynvjitlink _is_ available in time, we should give the full install command, like `pip install pynvjitlink --extra-index-url=https://pypi.nvidia.com`

bdice (approved):
Approving with one request for a final change based on the status/timeline of pynvjitlink wheels being published: https://github.com/rapidsai/cudf/pull/13650#discussion_r1398487365

python/cudf/cudf/utils/_numba.py
brandon-b-miller: How do you feel about keeping the original warning to be safe then. The changes will still allow `pynvjitlink` to work once it's out, but we won't document it or direct users to it in any official way until then.

bdice: Yes, do that. We can change the warning during code freeze if we ship pynvjitlink in time.

shwina (approved):

PR #13654, opened 2023-07-03
Add nvtext hash_character_ngrams function

davidwendt:
## Description
Adds a new nvtext function to return a hash column of generated character ngrams. The result can help speed up minhash and jaccard calculations which can use the hash values as unique tokens in place of the actual character ngrams themselves. There should be very few hash collisions especially when the ngram value is small.

Also, since the character ngrams are not actually generated, the cudf column size limit becomes less of an issue since the number of hashes becomes the limit instead of the total number of generated characters. So larger batch sizes may be possible with this approach.

The code for the `ngram` function from https://github.com/rapidsai/rapids-deduplication/issues/36 can be modified as follows
```python
def ngram(ds):
    return ds.str.hash_character_ngrams(5, True).list.unique()
```
The rest of the code can remain the same.
[except perhaps this additional improvement](https://github.com/rapidsai/rapids-deduplication/issues/36#issuecomment-1618545514)

This change reduced the runtime for the reference code above from 3.9s to 2.4s on my local machine.

## Checklist


wence- (approved):
Python changes look good, thanks!

cpp/include/nvtext/generate_ngrams.hpp
```diff
@@ -90,5 +90,36 @@ std::unique_ptr<cudf::column> generate_character_ngrams(
   cudf::size_type ngrams              = 2,
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
+/**
+ * @brief Hashes ngrams of characters within each string
```
karthikeyann, line 94: Should the description include what hashing algorithm is used in the implementation?

python/cudf/cudf/core/column/string.py
```diff
@@ -4850,6 +4850,48 @@ def character_ngrams(
             return result.explode()
         return result
 
+    def hash_character_ngrams(
+        self, n: int = 5, as_list: bool = False
+    ) -> SeriesOrIndex:
+        """
+        Generate hashes of n-grams from characters in a column of strings.
+
```
karthikeyann, line 4858: nit: same here. should we mention the hashing function used in the doc?

karthikeyann (approved):
Looks good to me. 
minor suggestion on the doc.

PointKernel (approved):

PR #13657, opened 2023-07-03
[REVIEW] Introduce parity with pandas for `MultiIndex.loc` ordering & fix a bug in `Groupby` with `as_index`

galipremsagar:
## Description
This PR:

- [x] Adds ordering in `MultiIndex.loc` when pandas compatibility mode is enabled.
- [x] Fixes a bug in `Groupby` when `as_index` is passed and is not being respected when a getitem call is performed on the groupby.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13660, opened 2023-07-05
Fix multiindex loc ordering in pandas-compat mode

wence-:
## Description

#13657 attempted to provide pandas-compatible ordering to the return value from multiindex-based `loc`-indexing. However, one must sort by keys corresponding to both the left and right gather maps of the join, rather than just the right gather map, to ensure that  correct ordering is maintained.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/multiindex.py
+++ b/python/cudf/cudf/core/multiindex.py
@@ -763,20 +763,6 @@ class MultiIndex(Frame, BaseIndex, NotIterable):
             if isinstance(row, slice) and row == slice(None):
                 continue
             lookup[i] = cudf.Series(row)
-        # Sort indices in pandas compatible mode
-        # because we want the indices to be fetched
-        # in a deterministic order.
-        # TODO: Remove this after merge/join
-        # obtain deterministic ordering.
-        should_sort = cudf.get_option("mode.pandas_compatible")
-        if should_sort:
-            lookup_order = "_" + "_".join(map(str, lookup.columns))
-            lookup[lookup_order] = column.arange(len(lookup))
-            postprocess = operator.methodcaller(
-                "sort_values", by=[lookup_order, "idx"]
-            )
-        else:
-            postprocess = lambda r: r  # noqa: E731
         frame = cudf.DataFrame(dict(enumerate(index._data.columns)))
         data_table = cudf.concat(
             [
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/multiindex.py
```diff
@@ -772,7 +787,7 @@ def _compute_validity_mask(self, index, row_tuple, max_length):
             ],
             axis=1,
         )
-        result = lookup.merge(data_table)["idx"]
+        result = postprocess(lookup.merge(data_table))["idx"]
```
brandon-b-miller, line 790: very minor nit: I think `postprocess` should be set just before it's used without the concat in between

python/cudf/cudf/core/multiindex.py
```diff
@@ -762,6 +763,20 @@ def _compute_validity_mask(self, index, row_tuple, max_length):
             if isinstance(row, slice) and row == slice(None):
                 continue
             lookup[i] = cudf.Series(row)
```
brandon-b-miller, line 765: out of scope of this PR but it would probably improve perf here if we didn't go through the whole song and dance of constructing a series for each column we're setting in and furthermore going through `DataFrame.__setitem__` each time

brandon-b-miller (approved):
lgtm

python/cudf/cudf/core/multiindex.py
wence-: All of this dataframe/series level code everywhere should be rewritten one level down operating directly on columns, but ...

galipremsagar (approved):
Thank you for fixing this @wence-!

python/cudf/cudf/core/multiindex.py
```diff
@@ -772,7 +773,20 @@ def _compute_validity_mask(self, index, row_tuple, max_length):
...
             axis=1,
         )
-        result = lookup.merge(data_table)["idx"]
+        # Sort indices in pandas compatible mode
+        # because we want the indices to be fetched
+        # in a deterministic order.
+        # TODO: Remove this after merge/join
+        # obtain deterministic ordering.
+        if cudf.get_option("mode.pandas_compatible"):
+            lookup_order = "_" + "_".join(map(str, lookup.columns))
```
galipremsagar, line 782: We will need to use `lookup._data.names` here instead of `lookup.columns`

wence-: Ah, thanks.

wence-, line 782:
```suggestion
            lookup_order = "_" + "_".join(map(str, lookup._data.names))
```

Issue #11675, opened 2022-09-08
[FEA] `read_orc_metadata` and `read_parquet_metadata` in libcudf


vuule:
libcudf does not have functions to read metadata of a file without reading (a portion) of the data as well. 
Exposing an efficient way to get information like column names and the number of rowgroups/stripes would make some uses of `read_orc` and `read_parquet` much simpler.

Exact information that these function should return is to be determined.

- [x] `read_orc_metadata`
- [ ] `read_parquet_metadata`

Issue #11214, opened 2022-07-07
[FEA] Add more functionality to `cudf.io.read_parquet_metadata` API


galipremsagar:
**Is your feature request related to a problem? Please describe.**
It would be nicer to have the row-group wise metadata returned instead of returning just the number of row-groups. That way users can identify how many rows & columns are stored in each row-group.

galipremsagar:
@vuule Did you intended to the same pyarrow parquet schema? Like:
```python
(Pdb) x = pq.ParquetFile(fname)
(Pdb) x.metadata
<pyarrow._parquet.FileMetaData object at 0x7fbe0a172040>
  created_by: parquet-cpp-arrow version 8.0.0
  num_columns: 15
  num_rows: 0
  num_row_groups: 1
  format_version: 1.0
  serialized_size: 7481
(Pdb) x.metadata.row_group
<built-in method row_group of pyarrow._parquet.FileMetaData object at 0x7fbe0a172040>
(Pdb) x.metadata.row_group(0)
<pyarrow._parquet.RowGroupMetaData object at 0x7fbe0a172270>
  num_columns: 15
  num_rows: 0
  total_byte_size: 196
```

or do we want to extract only the necessary bit of details and return those?

vuule:
I like the option to follow the pyarrow metada structure here, if it's not a huge overhead to gather.

galipremsagar:
Yea, should not be an issue since we already tap into this API anyways: https://github.com/rapidsai/cudf/blob/branch-22.08/python/cudf/cudf/io/parquet.py#L199

GregoryKimball:
> That way users can identify how many rows & columns are stored in each row-group.

Now that we have `read_parquet_metadata` from #13663, could we re-scope this issue to specify the changes we would like to see in the `parquet_metadata` class?

As I understand, the number of columns will be identical for all row groups. We could add row counts for each row group as a new vector, or perhaps to `metadata`. Where are the row group min/max statistics stored?

mhaseeb123:
Closed by #15398

PR #13663, opened 2023-07-05
Add read_parquet_metadata libcudf API

karthikeyann:
## Description
Closes https://github.com/rapidsai/cudf/issues/11675
Adds `read_parquet_metadata` to libcudf.
The metadata has following information
- schema - (type, name, children)
- num_rows
- num_rowgroups
- key-value string metadata in file footer

To Reviewers: Request for adding more information in metadata. Refer https://github.com/rapidsai/cudf/issues/11214

## Checklist


GregoryKimball:
Hello @etseidl would you please share your thoughts on this first PQ metadata function?

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -440,4 +440,26 @@ bool reader::impl::has_next()
...
+
+parquet_metadata read_parquet_metadata(std::vector<std::unique_ptr<datasource>>&& sources)
+{
+  // Open and parse the source dataset metadata
+  auto metadata = std::make_unique<aggregate_reader_metadata>(sources);
+
+  return parquet_metadata{parquet_schema{walk_schema(metadata.get(), 0)},
+                          metadata->get_num_rows(),
+                          metadata->get_num_row_groups(),
+                          metadata->get_key_value_metadata()[0]};
```
vuule, line 462: FWIW, we could avoid a copy here with an r-value specific overload for `get_key_value_metadata` and a bit more verbose code in this function. I don't know what the average size of this metadata is, so this might be irrelevant.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -440,4 +440,26 @@ bool reader::impl::has_next()
...
+    children.push_back(walk_schema(mt, child_idx));
+  }
+  return parquet_column_schema{
+    sch.name, static_cast<parquet::TypeKind>(sch.type), std::move(children)};
+}
+
+parquet_metadata read_parquet_metadata(std::vector<std::unique_ptr<datasource>>&& sources)
+{
+  // Open and parse the source dataset metadata
+  auto metadata = std::make_unique<aggregate_reader_metadata>(sources);
```
vuule, line 457: why not make an object here?
```suggestion
  auto metadata = aggregate_reader_metadata(sources);
```

cpp/include/cudf/io/parquet_metadata.hpp
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  /**
+   * @brief Returns the parquet schema.
+   *
+   * @return parquet schema
+   */
+  [[nodiscard]] auto const& schema() const { return _schema; }
+
+  ///< Number of rows in the root column; can vary for nested columns
```
vuule, line 189: looks like an accidentally kept line
```suggestion
```

cpp/include/cudf/io/detail/parquet.hpp
```diff
@@ -211,5 +212,14 @@ class writer {
...
 
+/**
+ * @brief Reads metadata of parquet dataset.
+ *
+ * @param sources Dataset sources to read from
+ *
+ * @return parquet_metadata with parquet schema, number of rows, number of row groups and key-value
+ * metadata.
+ */
+parquet_metadata read_parquet_metadata(std::vector<std::unique_ptr<datasource>>&& sources);
```
vuule, line 223: Do we want to make `read_parquet_metadata` uncallable with l-values (as it is now)? We don't really have to move from the sources, IMO const& is fine here.

vuule (changes requested):
Looks pretty good! I really like the  extent in which  `aggregate_reader_metadata` was used.
Some minor suggestions.

etseidl:
> Hello @etseidl would you please share your thoughts on this first PQ metadata function?

It's good for what it is, but for my purposes I'd need something more akin to returning the entire `aggregate_reader_metadata`. That's a bit much to expect, though, esp since I can just generate my own with thrift. 😄

karthikeyann:
> akin to returning the entire `aggregate_reader_metadata`.

do you need `aggregate_reader_metadata`'s public methods or access to  `FileMetaData` of each source?

vuule:
Looking at https://github.com/rapidsai/cudf/issues/11214, sounds like we'd want to add information about each row group. I'm fine with this being in a separate PR.

cpp/src/io/parquet/reader_impl.cpp
karthikeyann: Great suggestion!

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5601,147 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+    for (auto const& child : schema.children()) {
+      child_str += print(child, depth + 1);
+    }
+    return std::string(depth, ' ') + schema.name() + "\n" + child_str;
+  }
+};
+
+TEST_F(ParquetMetadataReaderTest, TestBasic)
+{
+  auto const num_rows = 1'200'000;
```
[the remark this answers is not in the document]
karthikeyann: reduced.

cpp/tests/io/parquet_test.cpp
```diff
@@ -5600,4 +5601,147 @@ TEST_F(ParquetReaderTest, ReorderedReadMultipleFiles)
...
+  EXPECT_EQ(meta.schema().root().type_kind(), cudf::io::parquet::TypeKind::UNDEFINED_TYPE);
+  ASSERT_EQ(meta.schema().root().num_children(), 2);
+
+  EXPECT_EQ(meta.schema().root().child(0).name(), "int_col");
+  EXPECT_EQ(meta.schema().root().child(1).name(), "float_col");
+}
+
+TEST_F(ParquetMetadataReaderTest, TestNested)
+{
+  auto const num_rows       = 1'200'000;
```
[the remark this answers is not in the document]
karthikeyann: reduced to 1200

cpp/include/cudf/io/detail/parquet.hpp
karthikeyann: Replaced with const-ref.

cpp/include/cudf/io/detail/parquet.hpp
```diff
@@ -211,5 +212,14 @@ class writer {
...
 
+/**
+ * @brief Reads metadata of parquet dataset.
+ *
+ * @param sources Dataset sources to read from
+ *
+ * @return parquet_metadata with parquet schema, number of rows, number of row groups and key-value
+ * metadata.
+ */
+parquet_metadata read_parquet_metadata(std::vector<std::unique_ptr<datasource>> const& sources);
```
vuule, line 223: nit: could be a `host_span<std::unique_ptr<datasource> const>`

cpp/src/io/parquet/reader_impl_helpers.hpp
```diff
@@ -115,6 +115,8 @@ class aggregate_reader_metadata {
 
   [[nodiscard]] auto const& get_key_value_metadata() const { return keyval_maps; }
 
+  [[nodiscard]] auto&& get_key_value_metadata() { return std::move(keyval_maps); }
```
vuule, line 118: Looks like there's a bit missing here; I believe this is the right overload set:
```suggestion
  [[nodiscard]] auto const& get_key_value_metadata() const& { return keyval_maps; }

  [[nodiscard]] auto&& get_key_value_metadata() && { return std::move(keyval_maps); }
```
First overload works on any lvalue ref (const or not), while the second one only works for rvalue, as it rips out the keyval_maps object out of the aggregate_reader_metadata object.

vuule (changes requested):
Looks great, just need to iron out the `get_key_value_metadata` overload set.

Thank you for addressing previous comments!

cpp/include/cudf/io/detail/parquet.hpp
karthikeyann: updated, and a few more places.

cpp/src/io/parquet/reader_impl_helpers.hpp
karthikeyann: Great suggestion. Learning to use ref-qualifiers here.

vuule (approved):

galipremsagar (approved):

cpp/include/cudf/io/parquet_metadata.hpp
```diff
@@ -0,0 +1,231 @@
...
+   *
+   * @return Children schemas
+   */
+  [[nodiscard]] auto const& children() const& { return _children; }
+
+  /** @copydoc children
+   * Children array is moved out of the object (rvalues only).
+   *
+   */
+  [[nodiscard]] auto children() && { return std::move(_children); }
```
divyegala, line 95: What is the necessity of this?

cpp/include/cudf/io/parquet_metadata.hpp
```diff
@@ -0,0 +1,231 @@
...
+   *
+   * @return Child schema
+   */
+  [[nodiscard]] auto const& child(int idx) const& { return children().at(idx); }
+
+  /** @copydoc child
+   * Child is moved out of the object (rvalues only).
+   *
+   */
+  [[nodiscard]] auto child(int idx) && { return std::move(children().at(idx)); }
```
divyegala, line 110: And this as well? Seems like it will invalidate `num_children()` once a child has been moved

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -440,4 +440,26 @@ bool reader::impl::has_next()
   return _current_read_chunk < _chunk_read_info.size();
 }
 
+static parquet_column_schema walk_schema(aggregate_reader_metadata const* mt, int idx)
```
divyegala, line 443: Can we use an anonymous namespace instead?

cpp/include/cudf/io/parquet_metadata.hpp
karthikeyann: @vuule 's suggestion here https://github.com/rapidsai/cudf/pull/13663#discussion_r1270909433

karthikeyann: For rvalues objects, `.children()` can return rvalue. It will be useful at places where only this member of rvalue objects are accessed.

cpp/include/cudf/io/parquet_metadata.hpp
karthikeyann: same as above. Allows child to be moved out of rvalue object.

cpp/include/cudf/io/parquet_metadata.hpp
divyegala: Oh okay, if you expect the pattern to be of the kind `parquet_column_schema().children()` then it makes sense to have these overloads

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -440,4 +440,28 @@ bool reader::impl::has_next()
   return _current_read_chunk < _chunk_read_info.size();
 }
 
+namespace {
+static parquet_column_schema walk_schema(aggregate_reader_metadata const* mt, int idx)
```
divyegala, line 444:
```suggestion
parquet_column_schema walk_schema(aggregate_reader_metadata const* mt, int idx)
```

karthikeyann: why not `static`?

karthikeyann: got it. Thanks
https://stackoverflow.com/questions/4422507/superiority-of-unnamed-namespace-over-static

divyegala: As far as I can tell, it's not a member function of a class, so there is no static internal state you are relying on.

static on free functions prevents leakage across multiple TUs, but using an anonymous namespace provides the same benefit and is more in line with other libcudf code.

What were your reasons for using static, if it's something different?

divyegala (approved):

raydouglass (approved):

PR #13666, opened 2023-07-06
Fix combined filtering and column projection in `read_parquet`

rjzamora:
## Description
Follow-up to https://github.com/rapidsai/cudf/pull/13334 for the special case that the `filters` argument includes column names that are **not** included in the current column projection (i.e. the `columns` argument). Although this pattern is not a common case at the moment, it is perfectly valid, and will become more common when cudf is used as a dask-expr backend (since the predicate-pushdown optimizations in dask-expr are significantly more advanced than those in dask-dataframe).

**Note**:
Prior to #13334, the special case targeted by this PR would not have produced any run-time errors, but it also wouldn't have produced proper filtering in many cases. Now that `cudf.read_parquet` **does** produce proper row-wise filtering, it turns out that we now need to sacrifice a bit of IO in cases like this. Although this is unfortunate, I personally feel that it is still worthwhile to enforce row-wise filtering.

## Checklist

### DIFF
--- a/python/cudf/cudf/io/parquet.py
+++ b/python/cudf/cudf/io/parquet.py
@@ -558,6 +558,17 @@ def read_parquet(
                 "for full CPU-based filtering functionality."
             )
 
+    # Make sure we read in the "filtered" columns needed for
+    # row-wise filtering after IO. This means that one or
+    # more columns will be dropped almost immediately after IO.
+    # However, we do NEED these columns for accurate filtering.
+    projection = None
+    if columns and filters:
+        filtered_columns = _filtered_columns(filters)
+        if filtered_columns - set(columns):
+            projection = columns
+            columns = sorted(filtered_columns | set(columns))
+
     # Convert parquet data to a cudf.DataFrame
     df = _parquet_to_frame(
         filepaths_or_buffers,
@@ -574,7 +585,10 @@ def read_parquet(
     )
 
     # Apply filters row-wise (if any are defined), and return
-    return _apply_post_filters(df, filters)
+    df = _apply_post_filters(df, filters)
+    if projection:
+        return df[projection]
+    return df
 
 
 def _normalize_filters(filters: list | None) -> List[List[tuple]] | None:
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/io/parquet.py
```diff
@@ -558,6 +558,17 @@ def read_parquet(
                 "for full CPU-based filtering functionality."
             )
 
+    # Make sure we read in the columns needed for row-wise
+    # filtering after IO. This means that one or more columns
+    # will be dropped almost immediately after IO. However,
+    # we do NEED these columns for accurate filtering.
+    projection = None
+    if columns and filters:
+        filtered_columns = _filtered_columns(filters)
```
wence-, line 567: At this point you've normalized the filters into `list[list[tuple]]` so `_filtered_columns` doesn't need to recurse to flatten things, and you can just do:
```
filtered_columns = set(v[0] for v in itertools.chain.from_iterable(filters))
```

Perhaps then:
```python
projected_columns = columns # this could be None so it's fine
if columns and filters:
    projected_columns = columns
    columns = sorted(set(v[0] for v in itertools.chain.from_iterable(filters)) | set(columns))
```
WDYT?

wence- (approved):
I think the inclusion of the filtering columns can be made a bit simpler, but otherwise logic looks good to me, thanks.

python/cudf/cudf/io/parquet.py
rjzamora: Ah, good point about `filters` being normalized here. In that case, `itertools.chain` makes a lot of sense.

> `projected_columns = columns # this could be None so it's fine`

Interestingly, this line caused failures, and exposed a bug in my logic. `columns` can technically include fields that are used to set the index, and so we need to make sure `projected_columns` only includes current column names before the `getitem` operation at the end of this function.

Since we need extra logic to check the elements of  `projected_columns`, I decided it probably makes more sense for the `projected_columns`  default to be `None` (rather than `columns`).

wence-:
Thanks Rick.

PR #13668, opened 2023-07-06
Add rmm_mode parameter to nvbench base fixture

davidwendt:
## Description
Adds a command-line parameter to libcudf nvbench benchmarks to specify an RMM memory resource when executing the benchmark.
```
Example:
benchmarks/STRINGS_NVBENCH -d 0 -b lengths --rmm_mode managed
```
The default is unspecified or specified incorrectly is `pool`.

This requires a minor update the `nvbench_global_setup.diff` patch to pass `argc/argv` to the `NVBENCHMARK_ENVIRONMENT` instance.
Also, remove the unneeded `rmm_pool_raii.hpp` header file and added `nvbench_fixture.hpp` to manage the RMM resource.

## Checklist


davidwendt:
34 out of the 39 files changed here are only removing the include for a header that is deleted in this PR.

cpp/benchmarks/fixture/nvbench_fixture.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <rmm/mr/device/managed_memory_resource.hpp>
+#include <rmm/mr/device/owning_wrapper.hpp>
+#include <rmm/mr/device/per_device_resource.hpp>
+#include <rmm/mr/device/pool_memory_resource.hpp>
+
+#include <string>
+
+namespace cudf {
+namespace detail {
+static std::string rmm_mode_parm{"--rmm_mode"};  ///< RMM mode command-line parameter name
```
bdice, line 30: I typically see parameter abbreviated as `param` and prefer that.
```suggestion
static std::string rmm_mode_param{"--rmm_mode"};  ///< RMM mode command-line parameter name
```

cpp/benchmarks/fixture/nvbench_fixture.hpp
```diff
@@ -0,0 +1,84 @@
...
+  inline auto make_cuda() { return std::make_shared<rmm::mr::cuda_memory_resource>(); }
+
+  inline auto make_pool()
+  {
+    return rmm::mr::make_owning_wrapper<rmm::mr::pool_memory_resource>(make_cuda());
+  }
+
+  inline auto make_async() { return std::make_shared<rmm::mr::cuda_async_memory_resource>(); }
+
+  inline auto make_managed() { return std::make_shared<rmm::mr::managed_memory_resource>(); }
```
bdice, line 48: I think @GregoryKimball may have an interest in supporting "managed pool" as well?

cpp/benchmarks/fixture/nvbench_fixture.hpp
```diff
@@ -0,0 +1,84 @@
...
+  {
+    return rmm::mr::make_owning_wrapper<rmm::mr::arena_memory_resource>(make_cuda());
+  }
+
+  inline std::shared_ptr<rmm::mr::device_memory_resource> create_memory_resource(std::string& mode)
+  {
+    if (mode == "cuda") return make_cuda();
+    if (mode == "async") return make_async();
+    if (mode == "arena") return make_arena();
+    if (mode == "managed") return make_managed();
```
bdice, line 60: Can this output an error if a match is not found?

cpp/cmake/thirdparty/patches/nvbench_global_setup.diff
```diff
@@ -1,27 +1,29 @@
...
  #define NVBENCH_MAIN_PARSE(argc, argv)                                                             \
    nvbench::option_parser parser;                                                                   \
    parser.parse(argc, argv)
-@@ -77,6 +85,7 @@
+@@ -77,6 +87,7 @@
      printer.set_total_state_count(total_states);                                                   \
                                                                                                     \
      printer.set_completed_state_count(0);                                                          \
-+    [[maybe_unused]] auto env_state = NVBENCH_ENVIRONMENT();                                       \
++    [[maybe_unused]] auto env_state = NVBENCH_ENVIRONMENT(argc,argv);                              \
```
bdice, line 26:
```suggestion
+    [[maybe_unused]] auto env_state = NVBENCH_ENVIRONMENT(argc, argv);                             \
```

bdice (commented):
This is a nice feature! A few small requests.

robertmaynard (approved):

cpp/benchmarks/fixture/nvbench_fixture.hpp
davidwendt: @bdice 
Would you have a snippet on how to create a managed-pool memory resource in this context?

GregoryKimball: > I think @GregoryKimball may have an interest in supporting "managed pool" as well?

Yes - very great interest!

bdice: @davidwendt Try this:


```cpp
  inline auto make_managed_pool()
  {
    return rmm::mr::make_owning_wrapper<rmm::mr::pool_memory_resource>(make_managed());
  }
```

The pool resource has a template parameter for its upstream. I think it’ll deduce the right upstream template type.

https://github.com/rapidsai/rmm/blob/005992676c9412b724278a91555536635698e3aa/include/rmm/mr/device/pool_memory_resource.hpp#L60-L61

cpp/benchmarks/fixture/nvbench_fixture.hpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  nvbench_base_fixture(int argc, char const* const* argv)
+  {
+    for (int i = 1; i < argc - 1; ++i) {
+      std::string arg = argv[i];
+      if (arg == detail::rmm_mode_param) {
+        i++;
+        rmm_mode = argv[i];
+      }
```
hyperbolic2346, line 75: There is no checking here for things like `bench --rmm_mode cuda --rmm_mode pool`. I don't know if this is necessary to check though.

davidwendt: I don't think this is worth checking for.

robertmaynard: The current behavior would select the last value provided when multiple `--rmm_mode` values exist. That matches the behavior of linux command line programs ( e.g. gcc, ls, ... )

Now what we are missing ( if we care ) is error handling around usage of `--rmm_mode` with no following value.

davidwendt: It defaults to 'pool' just fine in this case. That was my intention in this situation.

robertmaynard: Ahh, you are correct. I missed that you are iterating to `argc-1` and therefore ignore a trailing `--rmm_mode` without a value. :+1:

cpp/benchmarks/fixture/nvbench_fixture.hpp
bdice: Resolved in 0d1a2de.

bdice (approved):
Nice job!

PR #13669, opened 2023-07-06
Add nvtext::jaccard_index API for strings columns

davidwendt:
## Description
Adds the `nvtext::jaccard_index()` API to cudf.

```
std::unique_ptr<cudf::column> jaccard_index(
  cudf::strings_column_view const& input1,  cudf::strings_column_view const& input2,  cudf::size_type width,
```

The Jaccard Index is described here for computing the distance between two arrays of elements:  https://en.wikipedia.org/wiki/Jaccard_index

```
 Formula is J = |A ∩ B| 
                -------
                |A ∪ B|

 where |A ∩ B| is number of common values between A and B
 and |x| is the number of unique values in x.
```

The computation here compares strings columns by treating each string as text (i.e. sentences, paragraphs, articles) instead of individual words or tokens to be compared directly. The algorithm applies a sliding window (size specified by the `width` parameter) to each string to form the set of tokens to compare within each row of the two input columns.

These substrings are essentially character ngrams and each substring is part of the union and intersect calculations for that row. For efficiency, the substrings are hashed using the default MurmurHash32 to identify uniqueness within each row.
Once the union and intersect sizes for the row are resolved, the Jaccard index is computed using the above formula and returned as a float32 value.

The two input columns must be the same size and will match the output column size.

## Checklist


VibhuJawa:
We are seeing a 6x+speedup with this PR on our internal workflows. Thanks for the great work on this @davidwendt . 

Let's try to get this one in soon. 
CC: @ayushdg , @GregoryKimball

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,304 @@
...
+
+    // TODO: investigate cuCollections device-side static-map to match row values
+
+    cudf::size_type count = 0;
+    for (auto itr = needles.begin() + lane_idx; itr < needles.end() && begin < end;
+         itr += cudf::detail::warp_size) {
+      if (itr != needles.begin() && *itr == *(itr - 1)) { continue; }  // skip duplicates
+      // search haystack for this needle (*itr)
+      auto const found = thrust::lower_bound(thrust::seq, begin, end, *itr);
+      count += found != end && *found == *itr;  // increment if found;
```
bdice, line 138: Parens might be useful for readability.
```suggestion
      count += (found != end && *found == *itr);  // increment if found;
```

python/cudf/cudf/_lib/cpp/nvtext/jaccard.pxd
```diff
@@ -0,0 +1,16 @@
...
+from cudf._lib.cpp.column.column_view cimport column_view
+from cudf._lib.cpp.types cimport size_type
+
+
+cdef extern from "nvtext/jaccard.hpp" namespace "nvtext" nogil:
+
+    cdef unique_ptr[column] jaccard_index(
+        const column_view &input1,
+        const column_view &input2,
+        size_type ngrams
```
bdice, line 15: I'd like to standardize the name between Python and C++, either `ngrams` or `width`.

python/cudf/cudf/core/column/string.py
```diff
@@ -5337,6 +5337,35 @@ def minhash(
             libstrings.minhash(self._column, seeds_column, n, method)
         )
 
+    def jaccard_index(self, input: cudf.Series, n: int = 5) -> SeriesOrIndex:
+        """
+        Compute the jaccard index with this column against the given
```
bdice, line 5342:
```suggestion
        Compute the Jaccard index between this column and the given
```

bdice (commented):
Implementation looks good. My primary concern is around the default width (I'd prefer no default) and the requirement for a minimum of width 5.

cpp/include/nvtext/jaccard.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * @endcode
+ *
+ * If either input column's row is null, the output for that row will also be null.
+ *
+ * @throw std::invalid_argument if the `width < 5` or `input1.size() != input2.size()`
+ *
+ * @param input1 Strings column to compare with `input2`
+ * @param input2 Strings column to compare with `input1`
+ * @param width The character width used for apply substrings;
+ *              Default is 5 characters.
```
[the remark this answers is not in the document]
davidwendt: I suppose this a Data Science question so I don't have a definitive answer. Anything less than this would be very impracticable since it would likely blow up memory and not yield good results. I picked this because that is the default used by the code I was given. I can certainly see using with greater than this and so wanted to make it a parameter. 
Do you think solving this should hold up this PR?

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,304 @@
...
+                                              rmm::cuda_stream_view stream)
+{
+  auto hashes = hash_character_ngrams(col, width, stream, rmm::mr::get_current_device_resource());
+  auto const input   = cudf::lists_column_view(hashes->view());
+  auto const offsets = input.offsets_begin();
+  auto const data    = input.child().data<uint32_t>();
+
+  rmm::device_uvector<uint32_t> sorted(input.child().size(), stream);
+
+  // this is wicked fast and much faster than using cudf::lists::detail::sort_list
```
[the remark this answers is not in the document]
davidwendt: So sort-list does a sort-order followed by a gather and includes extra code for handling different types (type-dispatcher) and nulls. None of those are needed here. Also, all of the internal functions call SortPairs and we only need SortKeys here.

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,304 @@
...
+std::unique_ptr<cudf::column> jaccard_index(cudf::strings_column_view const& input1,
+                                            cudf::strings_column_view const& input2,
+                                            cudf::size_type width,
+                                            rmm::cuda_stream_view stream,
+                                            rmm::mr::device_memory_resource* mr)
+{
+  CUDF_EXPECTS(
+    input1.size() == input2.size(), "input columns must be the same size", std::invalid_argument);
+  CUDF_EXPECTS(width >= 5,
+               "Parameter width should be an integer value of 5 or greater",
```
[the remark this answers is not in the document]
davidwendt: Mainly because smaller values will increase memory size and likely give bad results.

python/cudf/cudf/_lib/cpp/nvtext/jaccard.pxd
davidwendt: Oops. That's a copy error. It should be width.

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,304 @@
...
+               std::invalid_argument);
+
+  auto const output_type = cudf::data_type{cudf::type_id::FLOAT32};
```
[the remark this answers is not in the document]
davidwendt: Why is this better?

cpp/src/text/jaccard.cu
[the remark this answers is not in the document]
bdice: Interesting. I see the [sort-order and gather here](https://github.com/rapidsai/cudf/blob/84578d793e4bf5a0924766d39ff717f5d9d71820/cpp/src/sort/segmented_sort_impl.cuh#L302C24-L302C52). I noticed recently that we're doing a _lot_ of "sort indices and gather" in the Python layer too (#13557). I suspect there's a ton of perf to be uncovered by specializing sorting paths like those for the case of no-nulls and simple/primitive types.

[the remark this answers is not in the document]
bdice: We can call this resolved.

cpp/include/nvtext/jaccard.hpp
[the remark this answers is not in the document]
bdice: I'd vote in favor of providing an unopinionated algorithm. n=2 seems just as natural as n=5, and the memory usage shouldn't be greatly affected. For a string of 100 characters, there are 96 sliding windows of size 5 and 99 sliding windows of size 2, so the number of windows doesn't grow much (input length - width + 1). And there are going to be more unique 5-character strings than 2-character strings after hash deduplication, so the postprocessing step should be even faster.

That said, this performance analysis could change a lot if we adopted a hashmap from cuCollections.

cpp/src/text/jaccard.cu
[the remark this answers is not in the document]
bdice: Resolving in favor of another thread. https://github.com/rapidsai/cudf/pull/13669#discussion_r1268297795

cpp/include/nvtext/jaccard.hpp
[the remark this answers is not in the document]
bdice: To be clear, I'd vote for:
- no default width (user must provide it)
- require width >= 2 (not 5)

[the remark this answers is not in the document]
davidwendt: Requesting input from @VibhuJawa and @ayushdg on the width parameter value range.

python/cudf/cudf/_lib/cpp/nvtext/jaccard.pxd
```diff
@@ -0,0 +1,16 @@
...
+
+from cudf._lib.cpp.column.column cimport column
+from cudf._lib.cpp.column.column_view cimport column_view
+from cudf._lib.cpp.types cimport size_type
+
+
+cdef extern from "nvtext/jaccard.hpp" namespace "nvtext" nogil:
+
+    cdef unique_ptr[column] jaccard_index(
+        const column_view &input1,
```
bdice, line 13: Does Cython support east const? :) (We don't have to change it in this PR, Cython const alignment isn't important to me, just wondering out loud whether it matters for Cython syntax...)

python/cudf/cudf/core/column/string.py
```diff
@@ -5337,6 +5337,35 @@ def minhash(
             libstrings.minhash(self._column, seeds_column, n, method)
         )
 
+    def jaccard_index(self, input: cudf.Series, n: int = 5) -> SeriesOrIndex:
```
bdice, line 5340:
```suggestion
    def jaccard_index(self, input: cudf.Series, width: int = 5) -> SeriesOrIndex:
```

python/cudf/cudf/core/column/string.py
```diff
@@ -5337,6 +5337,35 @@ def minhash(
...
+        """
+        Compute the Jaccard index between this column and the given
+        input strings column.
+
+        Parameters
+        ----------
+        input : Series
+            The input strings column to compute the jaccard index against.
+            Must have the same number of strings as this column.
+        n : int
```
bdice, line 5350:
```suggestion
        width : int
```

python/cudf/cudf/core/column/string.py
```diff
@@ -5337,6 +5337,35 @@ def minhash(
...
 
+    def jaccard_index(self, input: cudf.Series, n: int = 5) -> SeriesOrIndex:
+        """
+        Compute the Jaccard index between this column and the given
+        input strings column.
+
+        Parameters
+        ----------
+        input : Series
+            The input strings column to compute the jaccard index against.
```
bdice, line 5348:
```suggestion
            The input strings column to compute the Jaccard index against.
```

python/cudf/cudf/core/column/string.py
```diff
@@ -5337,6 +5337,35 @@ def minhash(
...
+        >>> str1 = cudf.Series(["the brown dog", "jumped about"])
+        >>> str2 = cudf.Series(["the black cat", "jumped around"])
+        >>> str1.str.jaccard_index(str2, 5)
+        0    0.058824
+        1    0.307692
+        dtype: float32
+        """
+
+        return self._return_or_inplace(
+            libstrings.jaccard_index(self._column, input._column, n),
```
bdice, line 5366:
```suggestion
            libstrings.jaccard_index(self._column, input._column, width),
```

bdice (approved):
A couple small suggestions, otherwise LGTM!

cpp/include/nvtext/jaccard.hpp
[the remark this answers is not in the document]
VibhuJawa: 1. I agree on `no default width `.

I have seen papers with `4`, `5` and `6` parameter ranges so far in papers . 

2. require width >= 2 , I think this seems reasonable.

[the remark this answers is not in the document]
davidwendt: Ok, thanks. I've removed the default at set the minimum to 2.

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,305 @@
...
+                                     sorted.size(),
+                                     input.size(),
+                                     offsets,
+                                     offsets + 1,
+                                     stream.value());
+
+  auto contents = hashes->release();
+  return cudf::make_lists_column(
+    col.size(),
+    std::move(contents.children.front()),
```
vyasr, line 236: Is this taking advantage of the fact the hashes has the same offsets as the input column so you can avoid allocating a new set of offsets? If so, that would be helpful to add a comment for since it is confusing to see a part of the hashes going into the output.

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,305 @@
...
+                                            rmm::cuda_stream_view stream,
+                                            rmm::mr::device_memory_resource* mr)
+{
+  CUDF_EXPECTS(
+    input1.size() == input2.size(), "input columns must be the same size", std::invalid_argument);
+  CUDF_EXPECTS(width >= 2,
+               "Parameter width should be an integer value of 2 or greater",
+               std::invalid_argument);
+
+  auto constexpr output_type = cudf::data_type{cudf::type_id::FLOAT32};
```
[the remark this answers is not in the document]
vyasr: Found it, [here's an example](https://godbolt.org/z/o4YaEhrsE) (admittedly not immediately relevant here since it's for pointers, but it's a good reason to maintain a consistent rule that has the expected behavior everywhere).

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,305 @@
...
+    d_input.child(cudf::lists_column_view::child_column_index).data<uint32_t>() + offset;
+  return cudf::device_span<uint32_t const>(begin, size);
+}
+
+/**
+ * @brief Count the unique values within each row of the input column
+ *
+ * This is called with a warp per row
+ */
+struct jaccard_unique_fn {
```
vyasr, line 68: Maybe more appropriate to call this something like `sorted_unique_fn` since it's not Jaccard-specific but it does assume that the data is presorted.

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,305 @@
...
+    auto const row_idx  = idx / cudf::detail::warp_size;
+    auto const lane_idx = idx % cudf::detail::warp_size;
+
+    auto const needles  = get_row(d_input1, row_idx);
+    auto const haystack = get_row(d_input2, row_idx);
+
+    auto begin     = haystack.begin();
+    auto const end = haystack.end();
+
+    // TODO: investigate cuCollections device-side static-map to match row values
```
vyasr, line 130: When investigating this, it would also be worth considering whether it would be optimal to select which column is the needles and which is the haystack based on the size of the row (I would guess you'd want to build the hash table on the larger row and search with the smaller row, but it may not really matter that much).

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,305 @@
...
+                     fn);
+  return d_results;
+}
+
+/**
+ * @brief Count the number of common values within each row of the 2 input columns
+ *
+ * This is called with a warp per row
+ */
+struct jaccard_intersect_fn {
```
vyasr, line 110: Same note about removing jaccard from the name potentially.

cpp/src/text/jaccard.cu
```diff
@@ -0,0 +1,305 @@
...
+  __device__ float operator()(cudf::size_type idx) const
+  {
+    auto const count1     = d_uniques1[idx];
+    auto const count2     = d_uniques2[idx];
+    auto const intersects = d_intersects[idx];
+    // the intersect values are in both sets so a union count
+    // would need to subtract the intersect count from one set
+    // (see formula in comment above)
+    auto const unions = count1 + count2 - intersects;
+    return unions ? ((float)intersects / (float)unions) : 0.f;
```
vyasr, line 185: Prefer `static_cast`? Although this is device code so I can see the argument to maintain C-style casts too.
```suggestion
    return unions ? (static_cast<float>(intersects) / static_cast<float>(unions)) : 0.f;
```

cpp/tests/text/jaccard_tests.cpp
vyasr, line 1: These tests look fine, but they're also fairly simple.
1. Does the Jaccard index have any invariants we could verify? I don't see any meaningful property other than the trivial verification that 0 < J < 1, but perhaps there is something I'm missing?
2. Would it be possible to implement a suboptimal version using public libcudf APIs? I believe it should be possible to do it by combining `generate_ngrams`, `cudf::split`, and one of `cudf::unique` or `cudf::distinct` (forget which one). If it's a lot of work I don't want to hold up this PR, but it would be nice to validate a nontrivial example against an alternative implementation that uses more extensively tested components of libcudf.

cpp/include/nvtext/jaccard.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * such that `output[row] = J(input1[row],input2[row])`.
+ *
+ * The Jaccard index formula is https://en.wikipedia.org/wiki/Jaccard_index
+ * ```
+ *  J = |A ∩ B| / |A ∪ B|
+ *  where |A ∩ B| is number of common values between A and B
+ *  and |x| is the number of unique values in x.
+ * ```
+ *
+ * The computation here compares strings columns by treating each string as text (i.e. sentences,
```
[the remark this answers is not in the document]
vyasr: From the perspective of an ngram it's probably not useful to have an ngram of n=1, but perhaps it does for Jaccard?

cpp/tests/text/jaccard_tests.cpp
vyasr: It would probably be much easier to write the more complex test in Python by using some host library to generate the inputs, so that would be a faster route.

cpp/benchmarks/text/jaccard.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+  if (static_cast<std::size_t>(num_rows) * static_cast<std::size_t>(row_width) >=
+      static_cast<std::size_t>(std::numeric_limits<cudf::size_type>::max())) {
+    state.skip("Skip benchmarks greater than size_type limit");
+  }
+
+  data_profile const strings_profile =
+    data_profile_builder()
+      .distribution(cudf::type_id::STRING, distribution_id::NORMAL, 0, row_width)
+      .no_validity();
```
vyasr, line 41: I don't know if it matters for the immediate applications at hand, but do we want to see what happens to perf when there is a long-tailed distribution? I would expect "real" documents to have a small number of extremely large strings, which I'd expect to perform substantially worse (not something we have to address in this PR, I'm asking for benchmarks so that we can be aware of it).

vyasr (changes requested):
This is a slick implementation, nice work! I have some queries, but most are either not terribly substantial or could be punted on if needed.

Do we care about the potential for hash collisions, or is the approximation we get from the hashed values good enough?

cpp/benchmarks/text/jaccard.cpp
vyasr: https://github.com/rapidsai/cudf/issues/13726

cpp/include/nvtext/jaccard.hpp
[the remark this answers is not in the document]
vyasr: Discussed offline, I'm fine leaving this documentation as is.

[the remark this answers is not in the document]
vyasr: @VibhuJawa is n=1 important for any of your use cases, or do you anticipate it ever being so? If not, I'll resolve this with no action. Otherwise, we should open an issue to track that generalization.

cpp/src/text/jaccard.cu
davidwendt: The offsets are the same because the segmented-sort does not change them.

cpp/src/text/jaccard.cu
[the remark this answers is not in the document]
davidwendt: Wow. That just looks strange to me.

cpp/tests/text/jaccard_tests.cpp
davidwendt: The solution here started from a Python implementation from https://github.com/rapidsai/rapids-deduplication/issues/36

cpp/include/nvtext/jaccard.hpp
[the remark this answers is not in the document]
VibhuJawa: Nope, not important to any use case i am aware of.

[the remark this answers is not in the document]
vyasr: OK cool thanks.

cpp/include/nvtext/jaccard.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * @code{.pseudo}
+ * input1 = ["the fuzzy dog", "little piggy", "funny bunny", "chatty parrot"]
+ * input2 = ["the fuzzy cat", "bitty piggy", "funny bunny", "silent partner"]
+ * r = jaccard_index(input1, input2)
+ * r is now [0.5, 0.15384616, 1.0, 0]
+ * @endcode
+ *
+ * If either input column's row is null, the output for that row will also be null.
+ *
+ * @throw std::invalid_argument if the `width < 2` or `input1.size() != input2.size()`
```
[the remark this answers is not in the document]
vyasr: Discussed [here](https://github.com/rapidsai/cudf/pull/13669#discussion_r1268748767), not important. Also, would require more substantial code changes to support, so definitely not worth the effort.

cpp/tests/text/jaccard_tests.cpp
davidwendt: Coded https://github.com/rapidsai/rapids-deduplication/issues/36 as a pytest. Hopefully that is ok.

python/cudf/cudf/tests/text/test_text_methods.py
```diff
@@ -855,3 +855,47 @@ def test_minhash():
...
+    db = db.rename(columns={0: "token"})
+    db["match"] = 1
+    inter = da.merge(db, on=["index", "token"], how="left")
+    inter = inter.groupby("index")["match"].sum()
+    union = da.merge(db, on=["index", "token"], how="outer")
+    union = union.groupby("index").size()
+    res = inter / union
+    res = res.sort_index()
+    res = res.values.astype("float32")
+    expected = cudf.Series(res)
```
vyasr, line 892: Thanks! This is helpful. However, my suggestion would be to run this in a separate pytest on some collections of random and longer strings rather than the same strings that you've already validated explicitly in the same test.

vyasr (approved):
Thanks for the fixes! One request, but totally fine to address it later.

python/cudf/cudf/tests/text/test_text_methods.py
vyasr: Quick suggestion for how this could be generated:
```
import random, string, cudf

word_length = 5
num_words = 10
s = cudf.Series([''.join(random.choice(string.ascii_lowercase) for _ in range(word_length)) for _ in range(num_words)])
```

davidwendt: Sorry, I've run out of time now. I'm not going to be able to do this before the end of burndown.
If you want, you can move this PR to 23.10 when that gets created later this week.
Thanks for reviewing!

vyasr: I added this test myself and verified that it works locally.

PR #13673, opened 2023-07-07
Add `cuda-nvcc-impl` to `cudf` for `numba` CUDA 12

jakirkham:
## Description

Ensure `cuda-nvcc-impl` is pulled in by `cudf` CUDA 12 packages alongside `numba` for CUDA support.

## Checklist

### DIFF
--- a/conda/recipes/cudf/meta.yaml
+++ b/conda/recipes/cudf/meta.yaml
@@ -90,6 +90,8 @@ requirements:
     - cubinlinker  # CUDA enhanced compatibility.
     - cuda-python >=11.7.1,<12.0a0
     {% else %}
+    # Needed by Numba for CUDA support
+    - cuda-nvcc-impl
     # TODO: Add nvjitlink here
     # xref: https://github.com/rapidsai/cudf/issues/12822
     - cuda-nvrtc


raydouglass (approved):

conda/recipes/cudf/meta.yaml
```diff
@@ -90,6 +90,8 @@ requirements:
     - cubinlinker  # CUDA enhanced compatibility.
     - cuda-python >=11.7.1,<12.0a0
     {% else %}
+    # Needed by Numba for CUDA support
```
bdice, line 93: Can we move this package and the comment so it’s clear that the nvrtc package is also needed for Numba? I’m not sure if the nvrtc package is needed for anything else.

jakirkham: It looks like NVRTC was added in commit ( https://github.com/rapidsai/cudf/pull/12922/commits/b5487524d34ac1b0bf501e70fb32371fba8e4cbb ). Do you recall why it was added then?

<hr>

Looking at the code, it appears NVRTC is used in other cases as well

https://github.com/rapidsai/cudf/blob/ecdaa91dc977ca58f94266d40a47ec8dffffbebd/cpp/include/cudf/utilities/bit.hpp#L31

https://github.com/rapidsai/cudf/blob/ecdaa91dc977ca58f94266d40a47ec8dffffbebd/cpp/tests/binaryop/util/runtime_support.h#L23-L24

<hr>

AIUI Jitify, [which cuDF also uses]( https://github.com/search?q=repo%3Arapidsai%2Fcudf%20jitify&type=code ), [uses NVRTC]( https://github.com/NVIDIA/jitify#jitify )

jakirkham: Anyways ok adding some kind of comment on `cuda-nvrtc` about `numba` needing. It just seems that it is used for more than `numba`, which would be good to capture somehow

robertmaynard: libcudf uses nvrtc via jitify. During the c++ build process we compile the jitify program, and use it to generate pre-processed input for nvrtc to be used at runtime.

bdice (approved):

conda/recipes/cudf/meta.yaml
bdice: I believe that CI is currently passing because Numba is finding the system CTK. Our `rapidsai/ci` images are based on `nvidia/cuda` and that ships the pieces that Numba needs. We should have seen failures from our tests that relied on `cuda-nvcc-impl` but instead Numba located those from the system. We'd need to address https://github.com/rapidsai/ops/issues/2565 to prevent the problem of mixing system and conda versions of the CUDA Toolkit.

conda/recipes/cudf/meta.yaml
```diff
@@ -90,6 +90,8 @@ requirements:
...
+    - cuda-nvcc-impl
```
bdice, line 94: I believe that CI is currently passing because Numba is finding the system CTK. Our `rapidsai/ci` images are based on `nvidia/cuda` and that ships the pieces that Numba needs. We should have seen failures from our tests that relied on `cuda-nvcc-impl` but instead Numba located those from the system. We'd need to address https://github.com/rapidsai/ops/issues/2565 to prevent the problem of mixing system and conda versions of the CUDA Toolkit.

jakirkham: Filed issue ( https://github.com/rapidsai/cudf/issues/13683 ) to follow up

Issue #13010, opened 2023-03-25
[BUG] Parquet writer unable to set nullability to false on columns with no nulls


vuule:
Parquet writer seems too stringent with its checks to write columns as not nullable. 
The following cases fail to write when `set_nullability(false)` is called on the input metadata:

- Column with no nulls;
- Sliced column that only has nulls outside of the slice.

This behaviors limits the use of the `nullable` option.

Repro code:
```
  column_wrapper<int32_t> ages_col{{1, 2, 3}};
  auto valids = cudf::detail::make_counting_transform_iterator(0, [](auto i) { return i > 1; });
  cudf::test::structs_column_wrapper col5{{ages_col}, valids};

  table_view expected({col5});

  cudf::io::table_input_metadata expected_metadata(expected);
  expected_metadata.column_metadata[0].set_nullability(false);

  auto expected_slice = cudf::slice(expected, {1, 3});

  auto filepath = temp_env->get_temp_filepath("SlicedTable.parquet");
  cudf::io::parquet_writer_options out_opts =
    cudf::io::parquet_writer_options::builder(cudf::io::sink_info{filepath}, expected_slice)
      .metadata(&expected_metadata);
  cudf::io::write_parquet(out_opts);
```

To test the case where the column has no nulls, modify `valids` to always return `true`.

vuule:
Issue partially blocks a Python feature https://github.com/rapidsai/cudf/pull/12952 (part of the PR got reverted in https://github.com/rapidsai/cudf/pull/12996)

GregoryKimball:
I believe this is closed by #13675

vuule:
@GregoryKimball  #13675 does not address the sliced column case

Issue #7654, opened 2021-03-19
[BUG] new parquet writer code checks for `nullable` not `has_nulls`


revans2:
**Describe the bug**
New parquet code was added to support writing nested types.  This is great, but it broke the java build.  As a part of fixing the java build I found that the new code checks for `nullable` on all of the columns to see if it matches what was set when the writer was initially configured.  But Spark can tell that validity is not needed in some cases where cudf apparently cannot, and cudf will add in a validity column in some cases when it is not needed.  because `nullable` only checks to see if there is a column, and not if there are actually any nulls we can run into a situation where spark tells us that there will be no nulls, but cudf blows up because it thinks that there might be.

razajafri:
This is still a valid issue

harrism:
@vuule is this on your radar?

vuule:
CC @devavret

devavret:
Do you mean this: https://github.com/rapidsai/cudf/blob/3c050bbcf369c7d6b56e1fd561e1e4efe1b7b64b/cpp/src/io/parquet/writer_impl.cu#L388-L406

If so, I'm not sure this is really a bug. We have to check whether there can be nulls otherwise we'd break downstream in the reader. So it either has to be `has_nulls()` or `nullable()`. The former incurs a kernel call for each column. That's why the latter was chosen. With the understanding, that if the user knew there wouldn't be nulls, they'd drop the mask before passing it in.

revans2:
Sorry about the delay in the response. I understand why nullable was chosen, but the issue I run into is that it is very easy to have something that is nullable, but has no nulls. This happens quite often in Spark. We can analyze the query we are able to tell that something will never have a null in it, but the last cudf call was not able to make the same decision and ended up including a validity column. For example we could filter out all of the nulls from a column. In spark we know that all of the nulls are gone, but cudf just called `filter` on it. It has no knowledge that the boolean mask is correlated to the validity column in any way. I just knows that the input has nulls so there is the possibility that the output will have nulls, and it needs to allocate a validity mask. If you don't do this then I have to figure out a way to walk through all of the columns/child columns and strip out the validity mask if it exists for things that I know don't need it, which then I would also want to run a kernel on it to be sure that there was no bug in the code.

vuule:
IMO we can think of this as a perf vs. memory trade off. For ORC/Parquet ATM we generally lean on the side of reducing the memory use, so it would make sense to have a slight perf overhead to avoid allocating the null masks.
[...]

PR #13675, opened 2023-07-07
Allow ORC and Parquet writers to write nullable columns without nulls as non-nullable

vuule:
## Description
Issue #7654, #13010

Writers have a strict check for nullability when applying the user metadata's nullability options, because checking the actual number of nulls is (was) not cheap.
Since we now create all columns with a known number of nulls, the `null_count` check became cheap and we have no reason to prevent columns without nulls to be written as non-nullable.
This PR changes the condition to allow this case. 
The PR does not address the issue with sliced columns, where it's not possible to write sliced column as non-nullable, even if the slice has no nulls. That check is still not cheap :)

## Checklist


cpp/src/io/orc/writer_impl.cu
```diff
@@ -687,9 +687,9 @@ orc_streams create_streams(host_span<orc_column_view> columns,
         // For chunked write, when not provided nullability, we assume the worst case scenario
         // that all columns are nullable.
         auto const chunked_nullable = column.user_defined_nullable().value_or(true);
-        CUDF_EXPECTS(chunked_nullable or !column.nullable(),
-                     "Mismatch in metadata prescribed nullability and input column nullability. "
-                     "Metadata for nullable input column cannot prescribe nullability = false");
+        CUDF_EXPECTS(chunked_nullable or column.null_count() == 0,
```
ttnghia, line 690: Probably you also need to update the comment above it.

vuule: The comment refers to the unchanged line above, tries to explain the default `true` value.

cpp/tests/io/orc_test.cpp
```diff
@@ -1806,4 +1806,21 @@ TEST_F(OrcWriterTest, EmptyRowGroup)
...
+
+  cudf::io::table_input_metadata expected_metadata(expected);
+  expected_metadata.column_metadata[0].set_nullability(false);
+
+  auto filepath = temp_env->get_temp_filepath("NonNullable.orc");
+  cudf::io::orc_writer_options out_opts =
+    cudf::io::orc_writer_options::builder(cudf::io::sink_info{filepath}, expected)
+      .metadata(&expected_metadata);
+  // Writer should be able to write a column without nulls as non-nullable
+  EXPECT_NO_THROW(cudf::io::write_orc(out_opts));
```
harrism, line 1823: Just a nit, either way is fine, but I don't think EXPECT_NO_THROW is required. If it throws and the exception is not caught, the test fails anyway.

harrism (approved):

cpp/tests/io/orc_test.cpp
vuule: I use it just to make it easier to see what we're checking. I prefer to have the macro for this reason.

ttnghia (approved):

Issue #13667, opened 2023-07-06
[FEA] support min/max for LIST group by aggregations.


revans2:
**Is your feature request related to a problem? Please describe.**
This is specifically related to https://github.com/NVIDIA/spark-rapids/issues/8668

It looks like we got CUDF support for most types, but LIST at the top level still appears to return an error.

```
ai.rapids.cudf.CudfException: CUDF failure at:/home/jenkins/agent/workspace/jenkins-spark-rapids-jni_nightly-dev-476-cuda11/thirdparty/cudf/cpp/src/groupby/sort/group_single_pass_reduction_util.cuh:108: Unsupported groupby reduction type-agg combination.
	at ai.rapids.cudf.Table.groupByAggregate(Native Method)
```

Oddly if I wrap the LIST in a struct and do the aggregation it all works correctly.

**Describe the solution you'd like**
LISTs work for the value in a min or max of a group by aggregation.

**Describe alternatives you've considered**
Do something really ugly, like wrap the LIST in a struct.

GregoryKimball:
Thank you @revans2 for noticing this. @divyegala would you please check out this scenario with LIST value types in aggregations? I believe @PointKernel tackled most of the refactoring in aggregations for experimental comparators. Hopefully this will be as simple as adjusting the overloads in `group_single_pass_reduction_util.cuh`. Thank you for your help!

GregoryKimball:
@divyegala did a quick evaluation here. To add LIST value support for min/max aggregations, we'll need to introduce a new utility module (e.g. `list_minmax_util.cuh`) to complement the existing `struct_minmax_util.cuh`. 

From where we are today, I sadly must encourage you to `Do something really ugly, like wrap the LIST in a struct.`, until we invest in a better solution here.

ttnghia:
I suspect that `struct_minmax_util` can be used for lists as well. Previously I implemented it to support structs while couldn't support lists but now we have the experimental comparator working for any type. 

I'm testing that idea and will confirm.

ttnghia:
Confirm that my idea work. Need more tests and cleanup. A PR should be up by next week.

PR #13676, opened 2023-07-07
Support `min` and `max` aggregations for list type in groupby and reduction

ttnghia:
This adds support for list type in `min` and `max` aggregations in groupby and reduction contexts.

Closes https://github.com/rapidsai/cudf/issues/13667 and closes https://github.com/rapidsai/cudf/issues/10408.

Status:
 * [X] Implementation.
 * [X] Unit tests.
 * [X] Run `compute-sanitizer`.
 * [X] Test with spark-rapids (https://github.com/NVIDIA/spark-rapids/pull/8689).


cpp/src/reductions/nested_type_minmax_util.cuh
```diff
@@ -141,6 +159,8 @@ class comparison_binop_generator {
   template <typename BinOp>
   static auto create(column_view const& input, rmm::cuda_stream_view stream)
   {
+    CUDF_EXPECTS(cudf::is_nested(input.type()),
+                 "This utility class only is designed exclusively for nested input types.");
```
ttnghia, line 163:
```suggestion
                 "This utility class is designed exclusively for nested input types.");
```

cpp/src/reductions/nested_type_minmax_util.cuh
```diff
@@ -149,8 +169,9 @@ class comparison_binop_generator {
 
   template <cudf::aggregation::Kind K>
   static auto create(column_view const& input, rmm::cuda_stream_view stream)
-
   {
+    CUDF_EXPECTS(cudf::is_nested(input.type()),
+                 "This utility class only is designed exclusively for nested input types.");
```
ttnghia, line 174:
```suggestion
                 "This utility class is designed exclusively for nested input types.");
```

cpp/src/groupby/hash/groupby.cu
```diff
@@ -646,18 +646,18 @@ std::unique_ptr<table> groupby(table_view const& keys,
...
                           ? cudf::dictionary_column_view(r.values).keys().type()
                           : r.values.type();
 
-    return not(r.values.type().id() == type_id::STRUCT) and
-           std::all_of(r.aggregations.begin(), r.aggregations.end(), [v_type](auto const& a) {
-             return cudf::has_atomic_support(cudf::detail::target_type(v_type, a->kind)) and
-                    is_hash_aggregation(a->kind);
-           });
+    // Currently, structs and lists values are not supported in any of hash-based aggregations.
+    // For those situations, we must fallback to sort-based aggregations.
```
divyegala, line 654: This is not true anymore, is it?

cpp/src/reductions/nested_type_minmax_util.cuh
```diff
@@ -68,29 +68,34 @@ auto static constexpr DEFAULT_NULL_ORDER = cudf::null_order::BEFORE;
...
- * ARGMIN, or smaller than all other non-null STRUCT elements - if finding for ARGMAX. This helps
- * achieve the results of finding the min or max STRUCT element when nulls are excluded from the
- * operations, returning null only when all the input elements are nulls.
+ * The binary operator provided by this class has an explicit non-inline `operator()` method to
+ * prevent excessive compile time when working with `thrust::reduce_by_key`.
+ *
+ * When it is a structs column, nulls STRUCT are compared as larger than all other non-null STRUCT
+ * elements - if finding for ARGMIN, or smaller than all other non-null STRUCT elements - if finding
+ * for ARGMAX. This helps achieve the results of finding the min or max STRUCT element when nulls
+ * are excluded from the operations, returning null only when all the input elements are nulls.
```
divyegala, line 79:
```suggestion
 * When it is a structs or a lists column, top-level NULLs are compared as larger than all other non-null
 * elements - if finding for ARGMIN, or smaller than all other non-null elements - if finding
 * for ARGMAX. This helps achieve the results of finding the min or max element when nulls
 * are excluded from the operations, returning null only when all the input elements are nulls.
```

cpp/src/reductions/nested_type_minmax_util.cuh
```diff
@@ -68,29 +68,34 @@ auto static constexpr DEFAULT_NULL_ORDER = cudf::null_order::BEFORE;
...
  */
 class comparison_binop_generator {
  private:
   cudf::table_view const input_tview;
+  bool const has_nulls;
   bool const is_min_op;
   rmm::cuda_stream_view stream;
+
+  // Contain data used in `row_comparator` below, thus needs to be kept alive as a member variable.
```
divyegala, line 88:
```suggestion
  // Contains data used in `row_comparator` below, thus needs to be kept alive as a member variable.
```

cpp/src/reductions/nested_type_minmax_util.cuh
```diff
@@ -68,29 +68,34 @@ auto static constexpr DEFAULT_NULL_ORDER = cudf::null_order::BEFORE;
...
   std::unique_ptr<cudf::structs::detail::flattened_table> const flattened_input;
-  bool const has_nulls;
-  std::vector<null_order> null_orders;
 
+  // Contain data used in the returned binop, thus needs to be kept alive as a member variable.
```
divyegala, line 91:
```suggestion
  // Contains data used in the returned binop, thus needs to be kept alive as a member variable.
```

cpp/src/groupby/hash/groupby.cu
ttnghia: This is true. Technically we can implement code to support hash-based aggregates for list type. However, we didn't implement any yet.

divyegala: I'm confused, how do we run algorithms like `distinct` which hash key based columns? `distinct` supports list type as key

ttnghia: Yeah but here we are checking for lists as **values**. We support lists as keys, which is a different story.

divyegala (approved):

cpp/src/groupby/hash/groupby.cu
mythrocks: I was confused by that as well, on the first read. I skipped over the "*values*" part.
Would there be value in including your clarification in the comment?
> lists as values. We support lists as keys, which is a different story.

cpp/src/reductions/nested_type_minmax_util.cuh
```diff
@@ -42,7 +42,7 @@ struct row_arg_minmax_fn {
   }
 
   // This function is explicitly prevented from inlining, because it calls to
-  // `row_lexicographic_comparator::operator()` which is inlined and very heavy-weight. As a result,
+  // `DeviceComparator::operator()` which is inlined and very heavy-weight. Otherwise,
   // instantiating this functor will result in huge code, and objects of this functor used with
   // `thrust::reduce_by_key` or `thrust::scan_by_key` will result in significant compile time.
```
mythrocks, line 47: The wording is a little stilted, and not much better than `As a result`. It isn't obvious what `Otherwise` is referring to.
```suggestion
  // `DeviceComparator::operator()` which is inlined and very heavy-weight. Inlining
  // this functor would result in huge code when instantiated, and significantly increase
  // compile time, if used with `thrust::reduce_by_key` or `thrust::scan_by_key`.
```

cpp/src/groupby/hash/groupby.cu
ttnghia: Thanks. I've updated the comments.

cpp/tests/reductions/reduction_tests.cpp
```diff
@@ -2548,6 +2548,105 @@ TEST_F(ListReductionTest, NonValidListReductionNthElement)
...
+TEST_F(ListReductionTest, ReductionMinMaxSlicedInput)
+{
+  using INTS_CW          = cudf::test::fixed_width_column_wrapper<int>;
+  using LISTS_CW         = cudf::test::lists_column_wrapper<int>;
+  using STRINGS_CW       = cudf::test::strings_column_wrapper;
+  using LISTS_STRINGS_CW = cudf::test::lists_column_wrapper<cudf::string_view>;
+
+  {
+    auto const input_original = LISTS_CW{{9, 9} /*dont care*/,
+                                         {0, 0} /*dont care*/,
```
mythrocks, line 2587: Nitpick:
```suggestion
    auto const input_original = LISTS_CW{{9, 9} /*don't care*/,
                                         {0, 0} /*don't care*/,
```

mythrocks (approved):
A couple of nitpicks. I appreciate the clarity in the code comments, and the test coverage.

cpp/tests/reductions/reduction_tests.cpp
```diff
@@ -2548,6 +2548,105 @@ TEST_F(ListReductionTest, NonValidListReductionNthElement)
     *cudf::make_nth_element_aggregation<reduce_aggregation>(0, cudf::null_policy::INCLUDE));
 }
 
+TEST_F(ListReductionTest, ReductionMinMaxNoNull)
+{
+  using INTS_CW          = cudf::test::fixed_width_column_wrapper<int>;
+  using LISTS_CW         = cudf::test::lists_column_wrapper<int>;
+  using STRINGS_CW       = cudf::test::strings_column_wrapper;
+  using LISTS_STRINGS_CW = cudf::test::lists_column_wrapper<cudf::string_view>;
+
```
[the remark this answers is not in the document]
ttnghia: Yeah I thought about that too. However, those aliases are used throughout this file so I want to enforce consistency.

PR #13681, opened 2023-07-10
Separate MurmurHash32 from hash_functions.cuh 

davidwendt:
## Description
Moves the `MurmurHash32` class definition from `hash_functions.cuh` to a new `murmur32.cuh` file.
Also moves the new file and the `hash_functions.cuh` from `cpp/include/cudf/detail/utilities/` to `cpp/include/cudf/hashing/detail`
The hash functions were redeclared from the `cudf::detail` namespace to the `cudf::hashing::detail` namespace.

The remaining changes are side-effects of making the above changes.
This PR is a follow on to PR #13626 
No new logic or functions have changed. Only internal detail headers have been refactored.

## Checklist


cpp/include/cudf/hashing/detail/hash_functions.cuh
```diff
@@ -0,0 +1,103 @@
+/*
```
davidwendt, line 1: This file was moved from `cpp/include/cudf/detail/utilities/hash_functions.cuh` and also split up
The `MurmurHash3_32` was moved to `murmurh_hash32.cuh`
Some utilities were only used by md5 and were moved into `md5_hash.cu`

cpp/include/cudf/hashing/detail/murmur_hash32.cuh
```diff
@@ -0,0 +1,194 @@
+/*
```
davidwendt, line 1: This was moved from `cpp/include/cudf/detail/utilities/hash_functions.cuh`

ajschmidt8 (approved):

cpp/src/hash/md5_hash.cu
```diff
@@ -41,6 +42,95 @@ namespace detail {
 
 namespace {
 
+template <int capacity, typename hash_step_callable>
```
davidwendt, line 45: All of this was moved from `cpp/include/cudf/detail/utilities/hash_functions.cuh`

cpp/include/cudf/hashing/detail/hash_functions.cuh
```diff
@@ -0,0 +1,100 @@
...
+
+#pragma once
+
+#include <cudf/utilities/traits.hpp>
+
+#include <limits>
+
+namespace cudf::hashing::detail {
+
+template <typename K>
```
bdice, line 25:
```suggestion
template <typename Key>
```

cpp/include/cudf/hashing/detail/murmur_hash32.cuh
bdice, line 1: Let’s name this identically to the hash function, murmurhash3_32.cuh.

cpp/include/cudf/hashing/detail/murmur_hash32.cuh
```diff
@@ -0,0 +1,194 @@
...
+#include <cstddef>
+
+namespace cudf::hashing::detail {
+
+// MurmurHash3_32 implementation from
+// https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp
+//-----------------------------------------------------------------------------
+// MurmurHash3 was written by Austin Appleby, and is placed in the public
+// domain. The author hereby disclaims copyright to this source code.
+// Note - The x86 and x64 versions do _not_ produce the same results, as the
```
bdice, line 36: Should we note whether this is the x64 or x86 version? Should this be named MurmurHash3_x64_32 or similar? Clarity here is crucial since we are adding new variants in a separate PR.

cpp/include/cudf/join.hpp
```diff
@@ -34,10 +34,11 @@
 namespace cudf {
 
 // forward declaration
-namespace detail {
+namespace hashing::detail {
 template <typename T>
 class MurmurHash3_32;
```
bdice, line 39: Should internal uses of hash functions like this use the MurmurHash3 explicitly or should they use the “default” hash function?

cpp/src/hash/murmur_hash.cu
```diff
@@ -51,11 +52,21 @@ std::unique_ptr<column> murmur_hash3_32(table_view const& input,
...
                    output_view.end<hash_value_type>(),
-                   row_hasher.device_hasher<cudf::detail::MurmurHash3_32>(nullable, seed));
+                   row_hasher.device_hasher<MurmurHash3_32>(nullable, seed));
 
   return output;
 }
 
 }  // namespace detail
+
+std::unique_ptr<column> murmur_hash3_32(table_view const& input,
```
bdice, line 62: As above, let’s pick a consistent name and choice of where underscores go. I understand we may have case differences between a class and function but the name and placement of underscores should be matched.

cpp/include/cudf/hashing/detail/hash_functions.cuh
```diff
@@ -0,0 +1,100 @@
...
+struct MurmurHash3_32;
+
+template <typename Key>
+using default_hash = MurmurHash3_32<Key>;
```
bdice, line 29: It’s awkward that this forward declaration forces users to include the MurmurHash3 header just to use the “default.” It injects a requirement of knowing the default so that the header can be included. I wonder if we can reorganize this so that there’s a utility header separate from the header defining the default hash functor. Then we can change the includes so that including the default hash header will include the MurmurHash3 header, which includes the utility header.

cpp/include/cudf/hashing/detail/murmur_hash32.cuh
davidwendt: There is no x64 version if that helps.
https://github.com/aappleby/smhasher/blob/61a0530f28277f2e850bfc39600ce61d02b518de/src/MurmurHash3.cpp#L94

cpp/include/cudf/hashing/detail/default_hash.cuh
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+#pragma once
+
+#include <cudf/hashing/detail/murmurhash3_x86_32.cuh>
+
+namespace cudf::hashing::detail {
+
+template <typename Key>
+struct MurmurHash3_x86_32;
```
bdice, line 24: Is this forward declaration still meaningful now that the header defining `MurmurHash3_x86_32<Key>` is included?

cpp/include/cudf/hashing/detail/hashing.hpp
bdice, line 1: I could see an argument to move the `hash_combine` utility functions like from this file into `hashing/detail/hash_functions.cuh`. What do you think?

cpp/src/hash/spark_murmurhash3_x86_32.cu
```diff
@@ -35,11 +36,11 @@ namespace {
 using spark_hash_value_type = int32_t;
 
 template <typename Key, CUDF_ENABLE_IF(not cudf::is_nested<Key>())>
-struct SparkMurmurHash3_32 {
+struct SparkMurmurHash3_x86_32 {
```
bdice, line 39: Probably need an underscore after "Spark" to match the function name `spark_murmurhash3_x86_32`.
```suggestion
struct Spark_MurmurHash3_x86_32 {
```

cpp/src/partitioning/partitioning.cu
```diff
@@ -724,6 +724,35 @@ struct dispatch_map_type {
 
 namespace detail {
 namespace {
+
+/**
+ * @brief  This hash function simply returns the value that is asked to be hash
+ * reinterpreted as the result_type of the functor.
```
bdice, line 730: This sentence is missing something to be grammatically correct. Maybe we can simplify it like:
```suggestion
 * @brief  This hash function simply returns the input value cast to the
 * result_type of the functor.
```

bdice (approved):
Approving with minor questions/suggestions.

cpp/include/cudf/hashing/detail/hashing.hpp
davidwendt: Some of these are used by .cpp files so I did not want them to include `hash_functions.cuh` which has some CUDA intrinsics in it.

cpp/include/cudf/join.hpp
davidwendt: This turned out to be more involved so I will fix this in a follow-up PR.

cpp/src/hash/md5_hash.cu
karthikeyann: nit: consider renaming `md5_hash.cu` to `md5.cu` or `md5hash.cu` to be consistent with other files. ignore if present name itself is okay.

karthikeyann (approved):
Looks good to me.
Only couple of minor questions to be clarified.

cpp/src/partitioning/partitioning.cu
```diff
@@ -724,6 +724,35 @@ struct dispatch_map_type {
...
 namespace {
+
+/**
+ * @brief This hash function simply returns the input value cast to the
+ * result_type of the functor.
+ */
+template <typename Key>
+struct IdentityHash {
+  using result_type = uint32_t;
+  IdentityHash()    = default;
```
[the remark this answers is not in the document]
davidwendt: Probably because of the `m_seed` but since that has an initial value it should be possible. 
Is there some benefit to declare the default constructor is `constexpr` ?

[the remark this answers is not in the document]
karthikeyann: This question came up because other constructor and member functions are constexpr, and the default constructor is not.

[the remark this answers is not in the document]
karthikeyann: `m_seed` may not be needed, since it's not used. can `m_seed` be removed?

cpp/include/cudf/hashing/detail/hashing.hpp
bdice: FWIW Those CUDA intrinsics for funnel shifts don't actually compile to fewer instructions. Originally I thought it would, but I checked this a while back and didn't see any change in the PTX/SASS. I just hadn't gone back to replace those. If it helps, we can replace the intrinsics with the equivalent shifting code.

davidwendt: We can address this in a follow-up PR.

ttnghia:
Re-label as breaking as it causes breaking in spark-rapids-jni.

PR #13692, opened 2023-07-12
Enable deterministic tokenization for cudf objects in dask

rjzamora:
## Description
As discussed in https://github.com/dask/dask/issues/6718, https://github.com/dask-contrib/dask-expr/issues/80, and https://github.com/dask/dask/pull/10398, cudf-backed objects do not produce a deterministic hash in `dask.base.tokenize`.

This PR registers `normalize_token` logic to manually hash rows and move cudf-based data to pandas before the final "tokenization" is performed. The result is that `tokenize(<cudf-data>)` is deterministic.

**Important Detail**:
Since this solution requires a device-to-host transfer, the specialized logic is only triggered when the `"tokenize.ensure-deterministic"` configuration option is explicitly set to `True` (in which case the default tokenization path would fail for cudf-backed data anyway).

## Checklist

### DIFF
--- a/python/dask_cudf/dask_cudf/backends.py
+++ b/python/dask_cudf/dask_cudf/backends.py
@@ -1,5 +1,6 @@
 # Copyright (c) 2020-2023, NVIDIA CORPORATION.
 
+import uuid
 import warnings
 from collections.abc import Iterator
 
@@ -12,6 +13,7 @@ from pandas.core.tools.datetimes import is_datetime64tz_dtype
 
 import dask.dataframe as dd
 from dask import config
+from dask.base import normalize_token
 from dask.dataframe.backends import (
     DataFrameBackendEntrypoint,
     PandasBackendEntrypoint,
@@ -471,6 +473,25 @@ def sizeof_cudf_dataframe(df):
 @_dask_cudf_nvtx_annotate
 def sizeof_cudf_series_index(obj):
     return obj.memory_usage()
+
+
+@normalize_token.register(cudf.core.frame.Frame)
+def _normalize_cudf_frame(obj):
+    # Only move data to pandas if we "need" to
+    if config.get("tokenize.ensure-deterministic", False):
+        return [
+            type(obj),
+            obj._dtypes,
+            (
+                obj.hash_values() if hasattr(obj, "hash_values") else obj
+            ).to_pandas(),
+        ]
+    return uuid.uuid4().hex
+
+
+@normalize_token.register(cudf.core.index.RangeIndex)
+def _normalize_cudf_rangeindex(index):
+    return index.to_pandas()
 
 
 def _default_backend(func, *args, **kwargs):
--- a/python/dask_cudf/dask_cudf/backends.py
+++ b/python/dask_cudf/dask_cudf/backends.py
@@ -492,7 +492,7 @@ def _normalize_cudf_frame(obj):
 
 @normalize_token.register(cudf.core.index.RangeIndex)
 def _normalize_cudf_rangeindex(index):
-    return index.to_pandas()
+    return (type(index), index.start, index.stop, index.step)
 
 
 def _default_backend(func, *args, **kwargs):
--- a/python/dask_cudf/dask_cudf/backends.py
+++ b/python/dask_cudf/dask_cudf/backends.py
@@ -486,6 +486,7 @@ def _normalize_cudf_frame(obj):
                 obj.hash_values() if hasattr(obj, "hash_values") else obj
             ).to_pandas(),
         ]
+    # If deterministic tokens are not needed, return a random UUID4 value.
     return uuid.uuid4().hex
 
 
--- a/python/dask_cudf/dask_cudf/tests/test_dispatch.py
+++ b/python/dask_cudf/dask_cudf/tests/test_dispatch.py
@@ -6,6 +6,7 @@ import pytest
 from packaging import version
 
 import dask
+from dask.base import tokenize
 from dask.dataframe import assert_eq
 from dask.dataframe.methods import is_categorical_dtype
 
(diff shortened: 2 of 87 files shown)


rjzamora:
cc @wence-

python/dask_cudf/dask_cudf/backends.py
```diff
@@ -473,6 +475,25 @@ def sizeof_cudf_series_index(obj):
...
+    # Only move data to pandas if we "need" to
+    if config.get("tokenize.ensure-deterministic", False):
+        return [
+            type(obj),
+            obj._dtypes,
+            (
+                obj.hash_values() if hasattr(obj, "hash_values") else obj
+            ).to_pandas(),
+        ]
+    return uuid.uuid4().hex
```
bdice, line 489: Let's clarify that this is a random value.
```suggestion
    # If deterministic tokens are not needed, return a random UUID4 value.
    return uuid.uuid4().hex
```

python/dask_cudf/dask_cudf/tests/test_dispatch.py
```diff
@@ -38,3 +39,16 @@ def test_pyarrow_conversion_dispatch():
...
+
+def test_deterministic_tokenize():
+    # Checks that `dask.base.normalize_token` correctly
+    # dispatches to the logic defined in `backends.py`
+    # when "tokenize.ensure-deterministic" == `True`
+    # (making `tokenize(<cudf-data>)` deterministic).
+    df = cudf.DataFrame({"A": list(range(10)), "B": ["dog", "cat"] * 5})
+
+    with dask.config.set({"tokenize.ensure-deterministic": True}):
+        assert tokenize(df) == tokenize(df)
```
bdice, line 52: Should we test that a deep copy of the dataframe also returns the same token?

rjzamora: Yeah, good idea.

python/dask_cudf/dask_cudf/backends.py
```diff
@@ -473,6 +475,26 @@ def sizeof_cudf_series_index(obj):
     return obj.memory_usage()
 
 
+@normalize_token.register(cudf.core.frame.Frame)
+def _normalize_cudf_frame(obj):
+    # Only move data to pandas if we "need" to
+    if config.get("tokenize.ensure-deterministic", False):
```
wence-, line 481: ensure-deterministic is not the default in dask, right?

python/dask_cudf/dask_cudf/backends.py
```diff
@@ -473,6 +475,26 @@ def sizeof_cudf_series_index(obj):
...
+                obj.hash_values() if hasattr(obj, "hash_values") else obj
+            ).to_pandas(),
+        ]
+    # If deterministic tokens are not needed, return a random UUID4 value.
+    return uuid.uuid4().hex
+
+
+@normalize_token.register(cudf.core.index.RangeIndex)
+def _normalize_cudf_rangeindex(index):
+    return index.to_pandas()
```
wence-, line 495: I guess this is cheap enough that it doesn't matter, but `(type(index), index.start, index.stop, index.step)` or similar would work too, right?

wence- (commented):
I think this looks OK, until we have some fast GPU hashing. But do we want determinism by default even if it's not requested in the dask config?

python/dask_cudf/dask_cudf/tests/test_dispatch.py
```diff
@@ -38,3 +39,19 @@ def test_pyarrow_conversion_dispatch():
...
+        assert tokenize(df.A) == tokenize(df.A)
+        assert tokenize(df.index) == tokenize(df.index)
+        assert tokenize(df) == tokenize(df.copy(deep=True))
+        assert tokenize(df.A) == tokenize(df.A.copy(deep=True))
+        assert tokenize(df.index) == tokenize(df.index.copy(deep=True))
```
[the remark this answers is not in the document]
rjzamora: Good call.

python/dask_cudf/dask_cudf/backends.py
rjzamora: Right, this is `False` by default in `dask.dask`, but [is now `True` in dask-expr](https://github.com/dask-contrib/dask-expr/pull/223). I'm slightly hesitant to add a new D->H transfer to dask-cudf workflows having `from_pandas`/`cudf`. Although I doubt the overhead will be significant, I saw this config as a pathway to punt on that decision for now.

python/dask_cudf/dask_cudf/backends.py
rjzamora: Yeah, that works

rjzamora:
>But do we want determinism by default even if it's not requested in the dask config?

Yes, we **may** want to use the `hash_values` approach more generally, but I'd prefer to make that change in a follow-up. After I've had time to test performance implications a bit.

rjzamora:
>but I'd prefer to make that change in a follow-up

On second thought, I am now thinking it may be best to avoid this "waiting period".

I'm also thinking this may be a great opportunity to solve the `cudf` vs `dask_cudf` dispatching problem (the problem being that dispatching logic for `cudf` is implemented in `dask_cudf`, and not all users understand if/when they need `dask_cudf`). That is, we can use the [`__dask_tokenize__` approach within `cudf`](https://docs.dask.org/en/latest/custom-collections.html#implementing-deterministic-hashing), instead of registering a `normalize_token` dispatch.  The advantage to doing this is that we can also import `dask_cudf` (when available) within the `__dask_tokenize__` definition to make sure we are registering the necessary dispatch functions for dask-dataframe/dask-expr to properly operate with a `cudf` backend. 

Note that I am currently experimenting with this idea in https://github.com/rjzamora/cudf/tree/normalize-token-cudf-2

Any thoughts or opinions on this @wence- ?

rjzamora:
Closing in favor of #13695

PR #13695, opened 2023-07-13
Add `__dask_tokenize__` definitions to cudf classes

rjzamora:
## Description
Possible alternative to https://github.com/rapidsai/cudf/pull/13692. See [this comment](https://github.com/rapidsai/cudf/pull/13692#issuecomment-1634741196)

General motivation: We need deterministic tokenization in dask-expr. These changes provide what we need. We also need dask-cudf to be imported whenever we are using a `cudf` backend in `dask.dataframe` or `dask_expr`. These changes ensure that `dask_cudf` is imported (when available).

## Checklist


python/cudf/cudf/core/index.py
```diff
@@ -1543,6 +1558,23 @@ def _indices_of(self, value):
...
+            warnings.warn(
+                f"Using dask to tokenize a {type(self)} object, "
+                f"but `dask_cudf` is not installed. Please install "
+                f"`dask_cudf` for proper dispatching."
+            )
+
+        # TODO: Avoid `to_pandas` once gpu hashing can
+        # produce a single (deterministic) token
+        return (type(self), self.to_pandas())
+
```
wence-, line 1577:
```suggestion
```

Rationale: `GernericIndex` inherits from `SingleColumnFrame` which already implements this.

python/dask_cudf/dask_cudf/tests/test_dispatch.py
```diff
@@ -38,3 +39,27 @@ def test_pyarrow_conversion_dispatch():
...
+    assert tokenize(df.A) == tokenize(df.A)
+    assert tokenize(df.index) == tokenize(df.index)
+    assert tokenize(df) == tokenize(df.copy(deep=True))
+    assert tokenize(df.A) == tokenize(df.A.copy(deep=True))
+    assert tokenize(df.index) == tokenize(df.index.copy(deep=True))
+
+    # Modifying an element should change the token
+    original_token = tokenize(df.A)
+    df.A.iloc[2] = 10
+    assert original_token != tokenize(df.A)
```
wence-, line 65: Can you test tokenizing an index as well as a dataframe?

python/cudf/cudf/core/index.py
wence-: That said, an index is immutable so that probably means there's a better way of doing this.

wence- (commented):
Not sure the specialisation for `genericindex` is necessary.

python/dask_cudf/dask_cudf/tests/test_dispatch.py
rjzamora: Ah, good call. It actually turns out that `hash_values` does **not** include the index in the hash, therefore we actually "need" `GenericIndex.__dask_tokenize__` (thought we can use caching for the index since it is immutable).

python/cudf/cudf/core/frame.py
```diff
@@ -2839,6 +2839,34 @@ def _repeat(
...
+                type(self),
+                self._dtypes,
+                self.index,
+                self.hash_values().values_host,
+            ]
+        return [
+            type(self),
+            self._dtypes,
+            self.to_pandas(),
+        ]
```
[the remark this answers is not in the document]
rjzamora: Yeah, I agree that this was bad. Thanks for the suggestion - changed.

python/cudf/cudf/core/index.py
```diff
@@ -915,6 +915,21 @@ def __pos__(self):
     def __abs__(self):
         return abs(self._as_int_index())
 
+    def __dask_tokenize__(self):
```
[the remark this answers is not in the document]
rjzamora: Right, I don't think there is a `cached`, but `lru_cache` should make sense here. (Note that I originally did this manually because I have been getting mypy warnings for `lru_cache` recently, but this doesn't seem to be an issue here)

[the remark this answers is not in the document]
wence-: Ah sorry, it's `functools.cache`.

python/cudf/cudf/core/index.py
```diff
@@ -5,7 +5,7 @@
 import math
 import pickle
 import warnings
-from functools import cached_property
+from functools import cached_property, lru_cache
```
wence-, line 8:
```suggestion
from functools import cached_property, cache
```

wence- (approved):
Minor change if you like, but looks good now I think.

python/cudf/cudf/core/index.py
rjzamora: Ah, right.

PR #13698, opened 2023-07-14
Remove Arrow dependency from the `datasource.hpp` public header

vuule:
## Description
Remove arrow dependency from `datasource.hpp`.

Breaking only because users of `arrow_io_source` now need to include the new `arrow_io_source.hpp` header instead on `datasource.hpp`

## Checklist


vuule:
Will re-target for 23.10 once changes are done; this one is not worth the risk to merge late into the release cycle.

galipremsagar (approved):

galipremsagar:
Thanks @vuule ! LGTM

ajschmidt8 (approved):

cpp/include/cudf/io/arrow_io_source.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+// We disable warning 2810 to workaround the compile issue (warning treated as error):
+// result.h(263): error #2810-D: ignoring return value type with "nodiscard" attribute
+#ifdef __CUDACC__
+#pragma nv_diag_suppress 2810
+#endif
+#include <arrow/result.h>
+#ifdef __CUDACC__
+#pragma nv_diag_default 2810
+#endif
```
ttnghia, line 43: We've recently upgraded to arrow 12. Can you try removing these suppression please?

ttnghia: Testing locally your latest changes in https://github.com/rapidsai/cudf/pull/13698/commits/031ca64f5fab857befc650ed61e0d12aa2d1bf73.

cpp/include/cudf/io/arrow_io_source.hpp
```diff
@@ -0,0 +1,145 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "datasource.hpp"
+
+#include <arrow/buffer.h>
+
```
ttnghia, line 22:
```suggestion
```

ttnghia: In addition, my clang-format changes this file a lot. So I wonder why CI clang-format doesn't complain?

cpp/include/cudf/io/arrow_io_source.hpp
ttnghia: The test build was successful. So we may be safe to remove such suppressions.

vuule: Let's see what CI says :)

python/cudf/cudf/_lib/cpp/io/arrow_io_source.pxd
```diff
@@ -11,5 +11,5 @@ cdef extern from "cudf/io/arrow_io_source.hpp" \
         namespace "cudf::io" nogil:
 
     cdef cppclass arrow_io_source(cudf_io_datasource.datasource):
-        arrow_io_source(string arrow_uri) except +
+        arrow_io_source(const string& arrow_uri) except +
```
ttnghia, line 14: east const.

vuule: For some reason I only see "const string" in pxd files so I kept this consistent with other instances. I'm not sure if our east const guideline reached Cython.

ttnghia (approved):

cpp/include/cudf/io/arrow_io_source.hpp
```diff
@@ -0,0 +1,137 @@
...
+    [[nodiscard]] uint8_t const* data() const override { return arrow_buffer->data(); }
+  };
+
+ public:
+  /**
+   * @brief Constructs an object from an Apache Arrow Filesystem URI
+   *
+   * @param arrow_uri Apache Arrow Filesystem URI
+   */
+  explicit arrow_io_source(std::string const& arrow_uri)
```
PointKernel, line 62: Same to all other member functions. Maybe considering separating class declaration from definition

PointKernel (approved):
Just one small software engineering comment. LGTM

cpp/include/cudf/io/arrow_io_source.hpp
vuule: I want to keep this class separate from libcudf, but we don't have to take that route.
@davidwendt thoughts on keeping the implementation in libcudf?

davidwendt: I'm not sure I understand what you mean by keeping this separate from libcudf.
I think Yunsong means moves the implementation to `cpp/src/io/arrow_io_source.cpp` and keep only the declarations in this header file. I agree with that approach.

python/cudf/cudf/_lib/cpp/io/arrow_io_source.pxd
vyasr: Cython doesn't support east const.

cpp/include/cudf/io/arrow_io_source.hpp
vuule: I meant that I will need to compile cpp/src/io/arrow_io_source.cpp as part of libcudf. Not an issue, just different from what I imagined in the beginning.

cpp/src/io/utilities/arrow_io_source.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace cudf::io {
+
+/**
+ * @brief Implementation for an owning buffer where `arrow::Buffer` holds the data.
+ */
+class arrow_io_buffer : public datasource::buffer {
+  std::shared_ptr<arrow::Buffer> arrow_buffer;
+
+ public:
```
robertmaynard, line 34: since you have public inheritance this `public:` is a no-op

robertmaynard (approved):

PR #13699, opened 2023-07-14
Update cython-lint and replace flake8 with ruff

vyasr:
## Description

This PR makes three changes to linting
- Updates the cython-lint version and fixes associated errors
- Adds the [ruff](https://beta.ruff.rs/docs/) linter
- Removes flake8

[cython-lint has achieved parity with flake8 for Cython files](https://github.com/MarcoGorelli/cython-lint/issues/45), so we no longer need flake8 for that. Meanwhile [ruff advertises itself as a suitable replacement for flake8 under three conditions](https://beta.ruff.rs/docs/faq/#how-does-ruff-compare-to-flake8): 1) few plugins, 2) using with black, and 3) Python 3 only. All of those apply to cudf code. From a quick analysis, there are a small handful of error codes supported by flake8 (via pyflakes or pycodestyle) that are not yet supported by ruff (pyflakes: F723, F842; pycodstyle: W391,W503, W504, W505, W606). On the flip side, ruff autofixes a range of issues that flake8 can't, it runs more than an order of magnitude faster, it supports more flexible configuration via pyproject.toml, and it promises to replace multiple tools over time. Therefore, I think this is a good switch for us to make.

Note that ruff could in theory also replace isort for Python files, but it does not support Cython yet, so I have left our isort hook in place. We could revisit in the future if that changes.

## Checklist


.pre-commit-config.yaml
```diff
@@ -165,6 +156,12 @@ repos:
         hooks:
             - id: rapids-dependency-file-generator
               args: ["--clean"]
+      - repo: https://github.com/astral-sh/ruff-pre-commit
+        rev: v0.0.278
+        hooks:
+          - id: ruff
+            files: python/.*$
```
bdice, line 163: I’d be open to running ruff on the whole repo if that makes sense. It could clean up various helper scripts, for example. Maybe in a follow-up.

python/cudf/cudf/_lib/sort.pyx
```diff
@@ -129,17 +131,19 @@ cdef pair[vector[order], vector[null_order]] ordering(
     -------
     pair of vectors (order, and null_order)
     """
-    cdef vector[order] c_column_order
+    cdef vector[cpp_order] c_column_order
     cdef vector[null_order] c_null_precedence
     for asc, null in zip(column_order, null_precedence):
-        c_column_order.push_back(order.ASCENDING if asc else order.DESCENDING)
+        c_column_order.push_back(
```
bdice, line 137: I dislike that we use c_ and cpp_ in different places here…

bdice (approved):
Awesome. I’d been wanting to explore this change, thanks for doing so.

.pre-commit-config.yaml
vyasr: Agreed, I would like to do that but for this PR kept the scope limited.

ajschmidt8 (approved):

python/cudf/cudf/_lib/sort.pyx
vyasr: I agree, it's not great, but I don't think it's worth addressing here. All of these bits of code will get rewritten in pylibcudf so we should aim to do a better job from the get-go with those versions instead.

PR #13701, opened 2023-07-16
Fix a corner case of list lexicographic comparator

ttnghia:
This fixes a bug showing up in spark-rapids (https://github.com/NVIDIA/spark-rapids/issues/8702) when the lexicographic comparator encounters a corner case with an input table having:
 * At least two structs-of-lists columns followed by at least one more column, and
 * All elements of the first structs-of-lists column are all null.

This bug is due to a short circuit in the lexicographic comparator that forgets to keep track of the number of processed lists columns during early exit.

### DIFF
--- a/cpp/include/cudf/table/experimental/row_operators.cuh
+++ b/cpp/include/cudf/table/experimental/row_operators.cuh
@@ -558,14 +558,12 @@ class device_row_comparator {
       // TODO: At what point do we verify that the columns of lhs and rhs are
       // all of the same types? I assume that it's already happened before
       // here, otherwise the current code would be failing.
-      auto [l_dremel_i, r_dremel_i] = [&]() {
-        if (_lhs.column(i).type().id() == type_id::LIST) {
-          return std::make_tuple(optional_dremel_view(_l_dremel[list_column_index]),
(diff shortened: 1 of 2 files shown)


cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -542,8 +542,10 @@ class device_row_comparator {
                                                 size_type const rhs_index) const noexcept
   {
     int last_null_depth = std::numeric_limits<int>::max();
-    size_type list_column_index{0};
+    size_type list_column_index{-1};
     for (size_type i = 0; i < _lhs.num_columns(); ++i) {
+      if (_lhs.column(i).type().id() == type_id::LIST) { ++list_column_index; }
```
ttnghia, line 547: Previously `list_column_index` is not increased at line 550 below. Thus, if early exit is called at line 550, the value of `list_column_index` is not updated properly.

bdice (approved):
This change seems fine to me. Thanks for adding a test.

divyegala (approved):

ttnghia:
Thanks everybody to approve it on Sunday :smile:

PR #13702, opened 2023-07-17
Alleviates JSON parser's need for multi-file sources to end with a newline

elstehle:
## Description

When reading a batch of `n` JSON files (i.e., line-delimited JSON files), we used to run into an issue if any of the first `n-1` do not contain a trailing newline character. Because the files are just concatenated, without joining them with an extra newline character:
```
{"a":"file_0_line_0"}
{"a":"file_0_line_1"}
{"a":"file_0_line_2"}{"a":"file_1_line_0"} <== invalid line, two records on one line
{"a":"file_1_line_1"}
{"a":"file_1_line_2"}
```

This PR adds support for these multi-file sources by "glueing" together these files with an extra `\n`. Introducing the extra newline, even if a file already is terminated with an `\n`, is safe, because our JSON lines format will ignore consecutive newlines as whitespace.

_Thanks to @wence- for suggesting this approach._

## Checklist


cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -59,16 +70,35 @@ rmm::device_uvector<char> ingest_raw_input(host_span<std::unique_ptr<datasource>
...
       }
     }
 
+    // If this is a multi-file source, we scatter the JSON line delimiters between files
+    if (sources.size() > 1) {
+      static_assert(num_delimiter_chars == 1,
+                    "Currently only single-character delimiters are supported");
+      auto const delimiter_source = thrust::make_constant_iterator('\n');
+      auto const d_delimiter_map  = cudf::detail::make_device_uvector_async(
+        host_span<size_type const>{delimiter_map.data(), delimiter_map.size() - 1},
```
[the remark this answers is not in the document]
elstehle: > Or is this deliberate because you want to interleave newlines between the sources, but not add a newline to the very last source?

Exactly, that's the idea. 

> (In which case I think the scatter end iterator wants to be .size() - 1?

Yup. Please note, though, that `scatter` is using the `d_delimiter_map.size()`, which already has the proper size (i.e., it already excludes the offset for the last file).

cpp/src/io/json/experimental/read_json.cpp
```diff
@@ -59,16 +70,35 @@ rmm::device_uvector<char> ingest_raw_input(host_span<std::unique_ptr<datasource>
...
+          bytes_read += num_delimiter_chars;
         } else {
           h_buffers.emplace_back(source->host_read(range_offset, data_size));
           auto const& h_buffer = h_buffers.back();
           CUDF_CUDA_TRY(cudaMemcpyAsync(
             destination, h_buffer->data(), h_buffer->size(), cudaMemcpyDefault, stream.value()));
           bytes_read += h_buffer->size();
+          delimiter_map.push_back(bytes_read);
+          bytes_read += num_delimiter_chars;
         }
```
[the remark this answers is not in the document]
elstehle: That's much better. Thank you 👍

cpp/src/io/json/experimental/read_json.cpp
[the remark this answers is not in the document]
wence-: > Yup. Please note, though, that `scatter` is using the `d_delimiter_map.size()`, which already has the proper size (i.e., it already excludes the offset for the last file).

Ah missed that, thanks.

wence- (approved):
I don't believe this has any power here...

cpp/src/io/json/experimental/read_json.cu
```diff
@@ -46,10 +51,19 @@ rmm::device_uvector<char> ingest_raw_input(host_span<std::unique_ptr<datasource>
                                            rmm::cuda_stream_view stream)
 {
   CUDF_FUNC_RANGE();
+  // We append a line delimiter between two files to make sure the last line of file i and the first
+  // line of file i+1 don't end up on the same JSON line, if file i does not already end with a line
+  // delimiter.
+  auto const num_delimiter_chars  = 1;
```
hyperbolic2346, line 57:
```suggestion
  auto constexpr num_delimiter_chars  = 1;
```

hyperbolic2346 (approved):
One tiny nit, but otherwise this seems a solid improvement.

cpp/tests/io/json_test.cpp
```diff
@@ -1235,6 +1247,49 @@ TEST_P(JsonReaderParamTest, JsonLinesMultipleFileInputs)
                                  float64_wrapper{{1.1, 2.2, 3.3, 4.4}, validity});
 }
 
+TEST_P(JsonReaderNoLegacy, JsonLinesMultipleFileInputsNoNL)
+{
+  auto const test_opt = GetParam();
+  std::vector<std::string> row_orient{"[11, 1.1]\n[22, 2.2]", "[33, 3.3]\n[44, 4.4]"};
+  std::vector<std::string> record_orient{
+    to_records_orient({{{"0", "11"}, {"1", "1.1"}}, {{"0", "22"}, {"1", "2.2"}}}, "\n"),
+    to_records_orient({{{"0", "33"}, {"1", "3.3"}}, {{"0", "44"}, {"1", "4.4"}}}, "\n")};
```
vuule, line 1256: The special property of the test data is that there are no newlines at the end of each file?

elstehle: Correct

vuule: can you please add a comment? The data looks "regular" at first glance :)

elstehle: Sure! Done 👍

cpp/src/io/json/experimental/read_json.cu
elstehle: Thanks! Adopted your suggestion 👍

vuule (approved):

robertmaynard (approved):
Approving cmake changes

cpp/src/io/json/experimental/read_json.cu
```diff
@@ -66,9 +80,27 @@ rmm::device_uvector<char> ingest_raw_input(host_span<std::unique_ptr<datasource>
...
+        delimiter_map.push_back(bytes_read);
+        bytes_read += num_delimiter_chars;
       }
     }
 
+    // If this is a multi-file source, we scatter the JSON line delimiters between files
+    if (sources.size() > 1) {
+      static_assert(num_delimiter_chars == 1,
+                    "Currently only single-character delimiters are supported");
+      auto const delimiter_source = thrust::make_constant_iterator('\n');
```
robertmaynard, line 92: I think we need to also look for `\r\n` in addition to `\n`

elstehle: Thanks, this shouldn't be an issue here. We just need to make sure that the last line of file `i` and the first line of file `i+1` are parsed as different JSON lines. Inserting a `\n` is sufficient for this.

PR #13704, opened 2023-07-17
Test using a centos7 image for java tests

raydouglass:
## Description
Just testing out using a CentOS 7 image for the java tests.

This is just a test for now, so do not merge!


vyasr:
@raydouglass is this still something that we want?

PR #13707, opened 2023-07-17
Fix reading of RLE encoded boolean data from parquet files with V2 page headers

### LINKED ISSUE #13655: [BUG] RLE encoded booleans are not loaded properly from Parquet
**Describe the bug**
Using libcudf to load the [rle_boolean_encoding.parquet file from apache/parquet-testing](https://github.com/apache/parquet-testing/blob/master/data/rle_boolean_encoding.parquet) results in a table with no nulls, but the parquet-cli tools shows that this file contains a number of null booleans.

Load the rle_boolean_encoding and note the column null count is 0 when it should be 6.

**Expected behavior**
  For reference, here's the values that are expected based on the output of the [parquet-cli tool](https://github.com/apache/parquet-mr/tree/master/parquet-cli):
```
$ java -cp '/home/jlowe/src/parquet-mr/parquet-cli/target/parquet-cli-1.12.2.jar:/home/jlowe/src/parquet-mr/parquet-cli/target/dependency/*' org.apache.parquet.cli.Main cat parquet-testing/data/rle_boolean_encoding.parquet
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": null}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": null}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": null}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": null}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": null}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": null}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": false}
{"datatype_boolean": false}
{"datatype_boolean": true}
{"datatype_boolean": true}
{"datatype_boolean": true}
```

etseidl:
## Description
The current parquet reader assumes that repetition or definition level data with a bit length of 0 will have no data encoded in the header.  In the case of V2 headers, this assumption is false. This PR checks the V2 page header data to see if level data needs to be accounted for. Also fixes an error that was present in the RLE data decoder where the encoded length of the RLE data was not skipped properly.

Fixes #13655

## Checklist


cpp/src/io/parquet/page_decode.cuh
```diff
@@ -871,45 +871,64 @@ inline __device__ uint32_t InitLevelSection(page_state_s* s,
...
-          cur++;
-        }
-        s->initial_rle_value[lvl] = v;
-      }
-      s->lvl_start[lvl] = cur;
-    }
-
-    if (cur > end) { s->error = 2; }
+    // subtract 4 because we've already incremented cur by 4
+    if (s->error == 0) { init_rle(cur, cur + len - 4); }
```
vuule, line 931: Could these lines be moved to the if branch above? I can't tell if `s->abs_lvl_start[lvl] = cur;` is really needed for the error branch.

vuule (commented):
Looks good, appreciate that the code duplication was avoided.
A few mildly confused comments below :)

cpp/src/io/parquet/page_hdr.cu
```diff
@@ -405,7 +405,7 @@ __global__ void __launch_bounds__(128)
             case PageType::DATA_PAGE_V2:
               index_out = num_dict_pages + data_page_count;
               data_page_count++;
-              bs->page.flags = 0;
+              bs->page.flags = PAGEINFO_FLAGS_V2;
```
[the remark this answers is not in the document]
etseidl: They don't _have_ to be OR'd because the 3 states we have are orthogonal...the page is either a dictionary page (flags == 1), regular data page (flags == 0), or V2 data page (flags == 2). But given that they sort of comprise a mask, they probably should be OR'd just in case other flags are added in the future (can't really think of any imminent ones right now). We'd just have to move the initialization of `flags` outside the switch.

I can make that change if you'd like.

cpp/src/io/parquet/page_decode.cuh
etseidl: yep, they sure can. good catch.

vuule (approved):

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -871,45 +871,65 @@ inline __device__ uint32_t InitLevelSection(page_state_s* s,
...
+    // V2 only uses RLE encoding so no need to check encoding
+    len = lvl == level_type::DEFINITION ? s->page.def_lvl_bytes : s->page.rep_lvl_bytes;
+    s->abs_lvl_start[lvl] = cur;
+    if (len == 0) {
+      s->initial_rle_run[lvl]   = s->page.num_input_values * 2;  // repeated value
+      s->initial_rle_value[lvl] = 0;
+      s->lvl_start[lvl]         = cur;
+    } else {
+      init_rle(cur, cur + len);
+    }
```
nvdbaranec, line 914: Is it always going to be safe to do this?  That is, is it always the case that for V2 headers, there will be a valid varint, regardless of whether `level_bits` is 0 or not?

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -1247,7 +1267,13 @@ inline __device__ bool setupLocalPageInfo(page_state_s* const s,
           s->dict_val  = 0;
           if ((s->col.data_type & 7) == BOOLEAN) { s->dict_run = s->dict_size * 2 + 1; }
           break;
-        case Encoding::RLE: s->dict_run = 0; break;
+        case Encoding::RLE: {
+          // first 4 bytes are length of RLE data
+          int const len = (cur[0]) + (cur[1] << 8) + (cur[2] << 16) + (cur[3] << 24);
+          cur += 4;
+          if (cur + len > end) { s->error = 2; }
+          s->dict_run = 0;
```
nvdbaranec, line 1275: File this under "how did this ever work?".  What's the story here - how did RLE decoding work if we were starting 4 bytes off?

etseidl: > File this under "how did this ever work?". What's the story here - how did RLE decoding work if we were starting 4 bytes off?

I'm guessing you never ran into RLE encoded bool data before :wink:

cpp/src/io/parquet/page_decode.cuh
etseidl: init_rle is only called if the number of bytes for the level is non-zero (`len != 0`). So at this point there should be _some_ bytes to read. The problem file actually only encodes the RLE length, but not the RLE value (which is then assumed to be 0). That's why the test for `cur < end` in `init_rle`

cpp/src/io/parquet/page_decode.cuh
etseidl: [Some light summer reading](https://issues.apache.org/jira/browse/PARQUET-2222) :wink: Near as I can tell booleans will only be encoded with RLE with V2 writers (although it seems arrow-rs might allow RLE w/ V1, but not by default). So it likely is true that this code has never been exercised.

nvdbaranec:
Currently running this through the spark plugin integration tests.  Should be done in ~20 minutes.

nvdbaranec (approved):
Tests pass. LGTM.

cpp/src/io/parquet/page_decode.cuh
etseidl: Looks like the encoder has RLE boolean support but it's disabled. If I enable it, it writes RLE encoded boolean data, but lacks the 4 bytes of length required by the Parquet spec. So I guess that answers the question of how did this ever work.

Once this PR and #13751 are merged I'll submit a PR to enable RLE encoding for booleans when writing V2 files.

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -2538,6 +2538,15 @@ def test_parquet_reader_binary_decimal(datadir):
     assert_eq(expect, got)
 
 
+def test_parquet_reader_rle_boolean(datadir):
+    fname = datadir / "rle_boolean_encoding.parquet"
+
+    expect = pd.read_parquet(fname)
+    got = cudf.read_parquet(fname).to_pandas()
+
+    assert_eq(expect, got)
```
galipremsagar, line 2547:
```suggestion
    got = cudf.read_parquet(fname)

    assert_eq(expect, got)
```

galipremsagar (approved):

python/cudf/cudf/tests/test_parquet.py
galipremsagar: roud-trip to pandas is handled by `assert_eq`

PointKernel (approved):

PR #13708, opened 2023-07-17
Optimize ORC reader performance for list data 

### LINKED ISSUE #13674: [BUG] Improve ORC reader performance with list columns and high row counts
We observe degenerate performance with the ORC reader with list columns when the row count is high. The Parquet reader does not have a similar issue.

Here is a cuDF-python code snippet for reproducing and profiling this issue:
```
    for num_rows in [1_000, 50_000_000]:
        df = cudf.DataFrame({'a': [[1]]*num_rows})

        df.to_parquet('/raid/tmp/tmp.pq')
        df.to_orc('/raid/tmp/tmp.orc')

        with nvtx.annotate(f'pq {num_rows}', color="blue"):
            t0 = time.time()
            df2 = cudf.read_parquet('/raid/tmp/tmp.pq')
            t1 = time.time()        

        print(f'parquet (numrows {num_rows})', t1-t0, 'seconds')

        with nvtx.annotate(f'orc {num_rows}', color="green"):
            t0 = time.time()
            df2 = cudf.read_orc('/raid/tmp/tmp.orc')
            t1 = time.time()

        print(f'orc (numrows {num_rows})', t1-t0, 'seconds')
```
With cuDF 23.06 and on my Epyc-A100 workstation, this script yields the following execution times:
```
parquet (numrows 1000) 0.0058746337890625 seconds
orc (numrows 1000) 0.003462076187133789 seconds
parquet (numrows 50000000) 0.012792587280273438 seconds
orc (numrows 50000000) 1.6445508003234863 seconds
```
Note how the Parquet and ORC times are similar for 1000 rows, but ORC is 100x slower for the 50M row case.

Nsys profiling shows that the bulk of the extra time is spent in a call to `generate_offsets_for_list`.
image

We missed this degenerate performance case previously because our benchmarks focus on wide tables with relatively large lists, rather than tall tables with short lists.

vyasr:
## Description

For list types the ORC reader needs to generate offsets from the sizes of nested lists. This process was previously being parallelized over columns. In practice even with wide tables we have enough rows that parallelizing over rows always makes more sense, so this PR swaps the parallelization strategy. 

I also removed what appears to be an unnecessary stream synchronization. That likely won't affect performance in any microbenchmarks but is worthwhile in case it helps improve asynchronous execution overall.

There are still noticeable bottlenecks for deeply nested lists, but those are in the decode kernels so optimizing them is a separate task for future work.

Resolves #13674 

## Checklist


cpp/src/io/orc/reader_impl.cu
```diff
@@ -725,12 +725,16 @@ struct list_buffer_data {
 void generate_offsets_for_list(device_span<list_buffer_data const> buff_data,
                                rmm::cuda_stream_view stream)
 {
-  auto const transformer = [] __device__(list_buffer_data const list_data) {
+  std::vector<list_buffer_data> buff_data_vec(buff_data.size());
+  CUDF_CUDA_TRY(cudaMemcpyAsync(buff_data_vec.data(),
+                                buff_data.data(),
+                                sizeof(list_buffer_data) * buff_data.size(),
+                                cudaMemcpyDeviceToHost,
```
vuule, line 732:
```suggestion
                                cudaMemcpyDefault,
```

cpp/src/io/orc/reader_impl.cu
```diff
@@ -725,12 +725,16 @@ struct list_buffer_data {
 void generate_offsets_for_list(device_span<list_buffer_data const> buff_data,
```
vuule, line 725: The call site of `generate_offsets_for_list` copies buffer data from host to device. You can just skip the round trip and pass a host span instead.

vuule (changes requested):
Great stuff!
Requesting a change for a suspected simplification.

cpp/src/io/orc/reader_impl.cu
vyasr: Great call, I missed that.

cpp/src/io/orc/reader_impl.cu
vyasr: I tried that initially and got some strange results. Perhaps there was something wrong with the intermediate case when I tested it though, let me see.

vuule: I guess you don't need it if you avoid the H2D2H round-trip :)

vyasr:
Moving benchmarks to a separate thread to avoid polluting the Git log when the PR merges:


<details><summary>Test script from #13674</summary>
<p>

```
# Old implementation
parquet (numrows 1000) 0.0062103271484375 seconds                                                                                                                                                          
orc (numrows 1000) 0.0033025741577148438 seconds                                                                                                                                                           
parquet (numrows 50000000) 0.016933441162109375 seconds                                                                                                                                                    
orc (numrows 50000000) 1.1314363479614258 seconds                      

# New implementation
parquet (numrows 1000) 0.0062923431396484375 seconds                                                                                                                                                       
orc (numrows 1000) 0.003666400909423828 seconds                                                                                                                                                            
parquet (numrows 50000000) 0.016930341720581055 seconds                                                                                                                                                    
orc (numrows 50000000) 0.045706748962402344 seconds    
```
</p>
</details>
The changes clearly close the gap for the above case.

<details><summary>libcudf ORC reader benchmarks</summary>
<p>

```
# Old implementation
  |            LIST | DEVICE_BUFFER |           0 |          1 |     10x |  52.919 ms | 0.06% |  52.913 ms | 0.06% |      10146288118 |         1.005 GiB |       488.481 MiB |                            
  |            LIST | DEVICE_BUFFER |        1000 |          1 |      6x |  87.514 ms | 0.45% |  87.508 ms | 0.45% |       6135114527 |         1.014 GiB |       363.095 MiB |                            
  |            LIST | DEVICE_BUFFER |           0 |         32 |      8x |  67.902 ms | 0.12% |  67.897 ms | 0.12% |       7907179514 |       574.083 MiB |        34.179 MiB |                            
  |            LIST | DEVICE_BUFFER |        1000 |         32 |      8x |  70.422 ms | 0.09% |  70.416 ms | 0.09% |       7624239955 |       572.610 MiB |        23.229 MiB |    


# New implementation
  |            LIST | DEVICE_BUFFER |           0 |          1 |     12x |  45.033 ms | 0.09% |  45.028 ms | 0.09% |      11923067760 |         1.005 GiB |       488.481 MiB |                            
  |            LIST | DEVICE_BUFFER |        1000 |          1 |      7x |  79.599 ms | 0.49% |  79.593 ms | 0.49% |       6745162704 |         1.014 GiB |       363.095 MiB |                            
  |            LIST | DEVICE_BUFFER |           0 |         32 |      9x |  60.006 ms | 0.12% |  60.000 ms | 0.12% |       8947826725 |       574.083 MiB |        34.179 MiB |                            
  |            LIST | DEVICE_BUFFER |        1000 |         32 |      8x |  62.578 ms | 0.18% |  62.573 ms | 0.18% |       8579974589 |       572.610 MiB |        23.229 MiB |   

```
</p>
</details>
The ORC reader benchmarks above show a marked improvement for the benchmarked list data. All other benchmarks in that file remain unchanged.

To also test the behavior of the new code on wide tables, I reran the test script from #13674, except instead of using tall tables I used wide tables containing varying number of columns.

<details><summary>Wide table</summary>
<p>

```
# Old code
orc (numrows 1, numcols 10) 0.00503087043762207 seconds
orc (numrows 1, numcols 100) 0.012778520584106445 seconds
orc (numrows 1, numcols 1000) 0.1071467399597168 seconds
orc (numrows 1, numcols 2000) 0.21944570541381836 seconds

orc (numrows 1000, numcols 10) 0.015061140060424805 seconds
orc (numrows 1000, numcols 100) 0.012740135192871094 seconds
orc (numrows 1000, numcols 1000) 0.10806846618652344 seconds
orc (numrows 1000, numcols 2000) 0.2716226577758789 seconds

orc (numrows 10000, numcols 10) 0.016198396682739258 seconds
orc (numrows 10000, numcols 100) 0.013859033584594727 seconds
orc (numrows 10000, numcols 1000) 0.11646103858947754 seconds
orc (numrows 10000, numcols 2000) 0.23884344100952148 seconds

orc (numrows 100000, numcols 10) 0.02451324462890625 seconds
orc (numrows 100000, numcols 100) 0.02280116081237793 seconds
orc (numrows 100000, numcols 1000) 0.19908523559570312 seconds
orc (numrows 100000, numcols 2000) 0.43782830238342285 seconds

# New code
orc (numrows 1, numcols 10) 0.004733562469482422 seconds
orc (numrows 1, numcols 100) 0.014241456985473633 seconds
orc (numrows 1, numcols 1000) 0.12213611602783203 seconds
orc (numrows 1, numcols 2000) 0.25175976753234863 seconds

orc (numrows 1000, numcols 10) 0.014937639236450195 seconds
orc (numrows 1000, numcols 100) 0.01426076889038086 seconds
orc (numrows 1000, numcols 1000) 0.12375545501708984 seconds
orc (numrows 1000, numcols 2000) 0.30136775970458984 seconds

orc (numrows 10000, numcols 10) 0.016460657119750977 seconds
orc (numrows 10000, numcols 100) 0.015311002731323242 seconds
orc (numrows 10000, numcols 1000) 0.13204264640808105 seconds
orc (numrows 10000, numcols 2000) 0.2677123546600342 seconds

orc (numrows 100000, numcols 10) 0.0243988037109375 seconds
orc (numrows 100000, numcols 100) 0.02396106719970703 seconds
orc (numrows 100000, numcols 1000) 0.214888334274292 seconds
orc (numrows 100000, numcols 2000) 0.4545779228210449 seconds
```
</p>
</details>

In this case it looks like the new code is a bit slower as expected. Given that the slowdown is on the order of a factor of 2 at worst whereas the improvements in the other cases are a couple orders of magnitude, the tradeoff seems more than worth it. In the future we could explore dispatching to different algorithms based on table width, but we'd want to come up with reliable heuristics first.

cpp/src/io/orc/reader_impl.cu
vyasr: Yeah haha this code is gone altogether now.

cpp/src/io/orc/reader_impl.cu
vyasr: Done.

cpp/src/io/orc/reader_impl.cu
```diff
@@ -722,15 +722,13 @@ struct list_buffer_data {
 };
 
 // Generates offsets for list buffer from number of elements in a row.
-void generate_offsets_for_list(device_span<list_buffer_data const> buff_data,
+void generate_offsets_for_list(host_span<list_buffer_data> const& buff_data,
```
ttnghia, line 725: A span can be copied by value.
```suggestion
void generate_offsets_for_list(host_span<list_buffer_data const> buff_data,
```

ttnghia (approved):

vuule (commented):
Looks good. Just the few posted comments.

cpp/src/io/orc/reader_impl.cu
```diff
@@ -1334,11 +1332,7 @@ table_with_metadata reader::impl::read(uint64_t skip_rows,
           }
         });
 
-      if (buff_data.size()) {
-        auto const dev_buff_data = cudf::detail::make_device_uvector_async(
-          buff_data, _stream, rmm::mr::get_current_device_resource());
-        generate_offsets_for_list(dev_buff_data, _stream);
-      }
+      if (buff_data.size()) { generate_offsets_for_list(buff_data, _stream); }
```
karthikeyann, line 1335:
```suggestion
      if (buff_data.empty()) { generate_offsets_for_list(buff_data, _stream); }
```

karthikeyann (dismissed):

cpp/src/io/orc/reader_impl.cu
```diff
@@ -722,15 +722,13 @@ struct list_buffer_data {
...
                                rmm::cuda_stream_view stream)
 {
-  auto const transformer = [] __device__(list_buffer_data const list_data) {
+  for (auto& list_data : buff_data) {
     thrust::exclusive_scan(
-      thrust::seq, list_data.data, list_data.data + list_data.size, list_data.data);
-  };
-  thrust::for_each(rmm::exec_policy(stream), buff_data.begin(), buff_data.end(), transformer);
-  stream.synchronize();
+      rmm::exec_policy(stream), list_data.data, list_data.data + list_data.size, list_data.data);
```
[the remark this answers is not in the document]
vyasr: I was considering making this change but wasn't sure if we were trying to make the change in a more consistent manner across RAPIDS. Happy to stick it in this PR.

cpp/src/io/orc/reader_impl.cu
vyasr: This suggestion inverts the condition. We could change it to:
```suggestion
      if (buff_data.size() > 0) { generate_offsets_for_list(buff_data, _stream); }
```
or 
```suggestion
      if (!buff_data.empty()) { generate_offsets_for_list(buff_data, _stream); }
```
Do you have a preference?

cpp/src/io/orc/reader_impl.cu
vyasr: Good point, however the change to the type qualifier here isn't quite right. The underlying list buffers will be changed by the algorithm, it's only the top-level reference that is const. I can accept the whole thing by value though and that should work fine.

cpp/src/io/orc/reader_impl.cu
vuule: My vote is for `not buff_data.empty()` :)

vyasr: I thought about including that as another option too :D I can do that.

cpp/src/io/orc/reader_impl.cu
vyasr: Changed.

cpp/src/io/orc/reader_impl.cu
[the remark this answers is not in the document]
vyasr: Done.

vuule (approved):

vyasr:
In the interest of wrapping things up in time for code freeze, I'm going to merge this since there are two approvals and @karthikeyann's requests have been addressed. Happy to do a follow-up if you have more requests afterwards though @karthikeyann.

Issue #13664, opened 2023-07-05
[BUG] Parquet list schema interpretation bug.


nvdbaranec:
The following file loads incorrectly using the cudf parquet reader:

https://github.com/apache/parquet-testing/blob/master/data/repeated_no_annotation.parquet

Internally there are 2 columns:
- `int`
- `list<struct<int64, string>>`

The reader appears to interpret the latter as `struct<struct<int64, string>`.   Ultimately this results in unpredictable/broken decoding.

The actual schema in the file is below ("phoneNumbers" is the column in question)

```
message spark_schema {
  required int32 id;
  optional group phoneNumbers {
    repeated group phone {
      required int64 number;
      optional binary kind (STRING);
    }
  }
}
```

GregoryKimball:
Thank you @nvdbaranec for raising this issue. It seems like another strange file case, where if we rewrite the data by arrow or cudf, cudf produces the correct table on read.

```
df = pd.read_parquet('repeated_no_annotation.parquet')
df.to_parquet('pd_write.pq')
df2 = cudf.read_parquet('pd_write.pq')
assert cudf.DataFrame.from_pandas(df).to_pandas().equals(df2.to_pandas())
# rewrite with pandas - OK!

df = pd.read_parquet('repeated_no_annotation.parquet')
df2 = cudf.DataFrame.from_pandas(df)
df2.to_parquet('cudf_write.pq')
df3 = cudf.read_parquet('cudf_write.pq')
assert df2.to_pandas().equals(df3.to_pandas())
# rewrite with cudf - OK!
```
Note, cuDF-python doesn't support `equals` on dataframes with complex columns

hyperbolic2346:
Changing this to properly read this as a list reveals another layer to this onion. The header shows the number of rows as 0, but the row groups inside have a valid row count. Need to determine how to properly reconcile when they disagree.

nvdbaranec:
If this is a list column, the number of rows in the page data is just the number of values in the data stream - the number of rows (in the cudf sense) can only be determined by examining the repetition levels.  What code is getting deceived here? Some sort of early out condition?

jlowe:
Note that Apache Spark thinks the schema of this file is not an INT, LIST<STRUCT<INT64,STRING>> but rather an INT, STRUCT<LIST<STRUCT<INT64,STRING>>>.

hyperbolic2346:
Can you show the table that spark creates? Does it have one row?

PR #13712, opened 2023-07-17
Parquet uses row group row count if missing from header

hyperbolic2346:
## Description
When investigating [this issue](https://github.com/rapidsai/cudf/issues/13664) I noticed that the file provided has 0 rows in the header. This caused cudf's parquet reader to fail at reading the file, but other tools such as `parq` and `parquet-tools` had no issues reading the file. This change counts up the number of rows in the row groups of the file and will complain loudly if the number differ, but not if the main header is 0. This allows us to properly read the data inside this file. Note that it will not properly parse it as a list of structs yet, that will be fixed in another PR. I didn't add a test since this is the only file I have seen with this issue and we can't read it yet in cudf. A test will be added for reading this file, which will test this change as well, with the PR for that issue.

## Checklist


ttnghia (approved):

karthikeyann (approved):
LGTM 👍

PR #13715, opened 2023-07-18
Fixing parquet list of struct interpretation

### LINKED ISSUE #13664: [BUG] Parquet list schema interpretation bug.

hyperbolic2346:
## Description
This change alters how we interpret non-annotated data in a parquet file. Most modern parquet writers would produce something like:
```
message spark_schema {
  required int32 id;
  optional group phoneNumbers (LIST) {
    repeated group phone {
      required int64 number;
      optional binary kind (STRING);
    }
  }
}
```

But the list annotation isn't required. If it didn't exist, we would incorrectly interpret this schema as a struct of struct and not a list of struct. This change alters the code to look at the child and see if it is repeated. If it is, this indicates a list.

closes #13664

## Checklist


harrism (approved):

jlowe:
I tried testing this PR with the Spark plugin, but Apache Spark thinks the schema of the repeated_no_annotation.parquet file is:
```
root                                                                            
 |-- id: integer (nullable = true)
 |-- phoneNumbers: struct (nullable = true)
 |    |-- phone: array (nullable = true)
 |    |    |-- element: struct (containsNull = true)
 |    |    |    |-- number: long (nullable = true)
 |    |    |    |-- kind: string (nullable = true)
```

I think the issue is that there's no logical type annotation on `phoneNumbers` so we can't assume it's a list even though the type pattern underneath it looks list-like.

cpp/tests/io/parquet_test.cpp
```diff
@@ -5618,4 +5618,77 @@ TEST_F(ParquetWriterTest, NoNullsAsNonNullable)
   EXPECT_NO_THROW(cudf::io::write_parquet(out_opts));
 }
 
+TEST_F(ParquetReaderTest, RepeatedNoAnnotations)
+{
+  unsigned char repeated_bytes[] = {
```
ttnghia, line 5623: Can this be `constexpr` or `const`?

ttnghia (approved):

hyperbolic2346:
This is really close now, but not 100%.

expected:
```
cudf::struct_view<cudf::list_view<cudf::struct_view<int64_t,cudf::string_view,>>,>:
Length : 6:
Null count: 2
111100
   cudf::list_view<cudf::struct_view<int64_t,cudf::string_view,>>:
   Length : 6
   Offsets(0x7f253c000600) : 0, 0, 0, 0, 1, 2, 5
   Null count: 2
   111100
      cudf::struct_view<int64_t,cudf::string_view,>:
      Length : 5:
         5555555555, 1111111111, 1111111111, 2222222222, 3333333333
         10110
         NULL, home, home, NULL, mobile
```
produced:
```
cudf::struct_view<cudf::list_view<cudf::struct_view<int64_t,cudf::string_view,>>,>:
Length : 6:
Null count: 2
111100
   cudf::list_view<cudf::struct_view<int64_t,cudf::string_view,>>:
   Length : 6
   Offsets(0x7f253c007000) : 0, 0, 0, 0, 1, 2, 5
   Null count: 3
   111000
      cudf::struct_view<int64_t,cudf::string_view,>:
      Length : 5:
         5555555555, 1111111111, 1111111111, 2222222222, 3333333333
         10110
         NULL, home, home, NULL, mobile
```

Note the null difference on the inner list.

hyperbolic2346:
Spark integration tests show a few errors like `java.lang.IllegalStateException: Logical error: no valid casts are found STRUCT to ByteType`, `Type conversion is not allowed from Table{columns=[ColumnVector{rows=1, type=LIST, nullCount=Optional.empty, offHeap=(ID: 435757 7f52640e7fa0)}], cudfTable=139991842717840, rows=1} to [MapType(StructType(StructField(first,StringType,true),StructField(middle,StringType,true),StructField(last,StringType,true)),StringType,true)] columns 0 to 1`, and `java.lang.IllegalStateException: Logical error: no valid casts are found STRUCT to BooleanType` which make me think there may be some changes to how parquet normal files are being interpreted by cudf and not just the very old format.

More investigation is warranted here.

hyperbolic2346:
Integration tests for spark are passing now and this is ready for review.

cpp/src/io/parquet/parquet.hpp
```diff
@@ -204,9 +204,10 @@ struct SchemaElement {
   // can be a struct or a list
   [[nodiscard]] bool is_struct() const
   {
+    // this assumption might be a little weak.
     return type == UNDEFINED_TYPE &&
            // this assumption might be a little weak.
```
vuule, line 209: are both assumptions little weak?

cpp/src/io/parquet/reader_impl_helpers.cpp
```diff
@@ -175,6 +175,81 @@ type_id to_type_id(SchemaElement const& schema,
...
+        struct_elem.name            = "struct_node";
+        struct_elem.repetition_type = REQUIRED;
+        struct_elem.num_children    = schema_elem.num_children;
+        struct_elem.type            = UNDEFINED_TYPE;
+        struct_elem.converted_type  = UNKNOWN;
+
+        // swap children
+        struct_elem.children_idx = std::move(schema_elem.children_idx);
+        schema_elem.children_idx = {struct_node_idx};
+        schema_elem.num_children = 1;
```
vuule, line 228: can `num_children` be different from `children_idx.size()`?

cpp/src/io/parquet/reader_impl_helpers.cpp
```diff
@@ -175,6 +175,81 @@ type_id to_type_id(SchemaElement const& schema,
   return type_id::EMPTY;
 }
 
+void metadata::sanitize_schema()
```
vuule, line 178: neat approach, localizes the madness :+1:

vuule (commented):
Looks good; few questions.

cpp/src/io/parquet/parquet.hpp
hyperbolic2346: super extra weak.

cpp/src/io/parquet/reader_impl_helpers.cpp
hyperbolic2346: I think this could ultimately fix things like single-level list as well. I like this approach too.

cpp/src/io/parquet/reader_impl_helpers.cpp
hyperbolic2346: I don't know of a way where that would be legal. I'm not sure about the history of this schema and why it was a discrete value instead of using the vector. I'd be happy to file a follow-up PR to remove that if you'd like.

vuule: Looked again into this, seems like `num_children` is a part of the compact protocol type.

vuule (approved):

PR #13717, opened 2023-07-18
Parse (non-MultiIndex) label-based keys to structured data

### LINKED ISSUE #13014: [BUG] loc indexing with length-1 Categorical not possible
**Describe the bug**
```python
import cudf
import pandas as pd
f = pd.Series([1, 2, 3], index=pd.CategoricalIndex([0, 1, 2]))
f.loc[pd.Categorical([1])] # => second row.
cf = cudf.from_pandas(f)
cf.loc[pd.Categorical([1])] # => KeyErrror
```

I think this is similar to #13013 in that length-one categoricals are treated as scalars in some cases.

**Expected behavior**

Probably this lookup should work.

### LINKED ISSUE #13268: [BUG] Series and DataFrame loc indexing does not handle `Ellipsis`

### LINKED ISSUE #13379: [BUG] `loc`-based indexing of DataFrames silently discards missing keys if at least one key is present in indexer
**Describe the bug**

When performing `loc` indexing of a `DataFrame`, if one asks for a missing key, pandas raises a `KeyError`. In constrast, cudf only does so if none of the requested keys are in the index. If at least one requested key is present then the subsetted data frame with that key is returned and the missing keys are silently dropped. 

Series `loc` indexing does the right thing here and raises `KeyError` if any keys are missing.

```python
import pandas as pd
import cudf

# same failure with rangeindex too.
df = pd.DataFrame({"A": range(5)}, index=list(range(5)))
cdf = cudf.from_pandas(df)

df.loc[[0, 5]] # 5 is missing, raises KeyError

cdf.loc[[0, 5]]
#    A
# 0  0
```

**Expected behavior**

Should match pandas behaviour and raise `KeyError`.

### LINKED ISSUE #13652: [BUG] loc-based indexing fails when looking up array like with ordered categorical index
**Describe the bug**

When a frame's index is a `CategoricalIndex` that is _ordered_, looking up an array-like list of indices fails with `TypeError: Merging on categorical variables with mismatched ordering is ambiguous`.

This occurs because to perform the merge between the index and the to-be-looked-up values (to find matches), the values are turned into a categorical column, but the default category dtype is un-ordered.

```python
import cudf
s = cudf.Series([1, 2, 3, 4, 5], index=cudf.CategoricalIndex([1, 2, 3, 4, 5], categories=[1, 2, 3, 4, 5], ordered=True))
s.loc[[1, 4]] # => KeyError
```

**Expected behavior**

```python
s.to_pandas().loc[[1, 4]]
# 1    1
# 4    4
# dtype: int64
```

I note actually that pandas doesn't care at all about ordered-ness of categoricals in lookup. Which kind of makes sense because you're just looking up values.

So probably the solution is to merge the decategorized index column with the asked-for labels.

### LINKED ISSUE #13653: [BUG] Scalar loc-based lookup in integer categorical indices is incorrect
**Describe the bug**

`loc`-based lookup does (I think incorrect) fallback to positional indexing rather than label-based lookup when the index is a categorical one with integer values.

```python
import cudf

s = cudf.Series([1, 2], index=cudf.CategoricalIndex([3, 4], categories=[3, 4]))
s.loc[3] # IndexError: single positional indexer is out-of-bounds
```

**Expected behavior**

```python
s.to_pandas().loc[3]
# 1
```

~Annoyingly, one can't just stop doing positional indexing fallback in all cases because if the index is (say) a string index then integer indexing _does_ fall back to positional.~ This is for `Series.__getitem__` and the behaviour is deprecated in pandas 2.

### LINKED ISSUE #13658: [BUG] loc-based lookup where index has repeated labels does not (always) match pandas order

wence-:
## Description

Following on from #13534, this extends the scheme to handle label-based lookups as long as the index is not a multiindex.

As is the case for positional indexing, all of the different ways one can index a frame with labels eventually boil down to indexing by slice, boolean mask, map, or scalar in libcudf. `loc`-based keys are parsed into information that tags them by type. Since this information is the same as is used for `iloc` indexing, we can then dispatch to the same "internal" calls that don't do further bounds-checking or normalisation: rather than converting a label-based lookup to an argument we can pass to `iloc`-getitem (which must reinterpret it), we just take the decision straight away.

The next stage (which will help to remove a bunch of code) will be to handle multiindex keys, but that will be sufficiently complicated that I'd rather do it separately.

- Closes #13014
- Closes #13268 
- Closes #13379 
- Closes #13652
- Closes #13653
- Closes #13658 (or at least, maintains the correctness)

## Checklist

### DIFF
--- a/python/cudf/cudf/core/indexing_utils.py
+++ b/python/cudf/cudf/core/indexing_utils.py
@@ -3,9 +3,12 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from functools import partial, reduce
 from typing import TYPE_CHECKING, Any, Tuple, Union
 
+import cupy as cp
 import numpy as np
+import pandas as pd
 from typing_extensions import TypeAlias
 
 import cudf
@@ -462,6 +465,42 @@ def ordered_find(needles: "ColumnBase", haystack: "ColumnBase") -> GatherMap:
     )
 
 
+def find_label_range_or_mask(
+    key: slice, index: cudf.BaseIndex
+) -> EmptyIndexer | MapIndexer | MaskIndexer | SliceIndexer:
+    # TODO: datetime index must only be handled specially until pandas 2
+    if (
+        not (key.start is None and key.stop is None)
+        and isinstance(index, cudf.core.index.DatetimeIndex)
+        and not index.is_monotonic_increasing
+    ):
+        start = pd.to_datetime(key.start)
+        stop = pd.to_datetime(key.stop)
+        mask = []
+        if start is not None:
+            mask.append(index >= start)
+        if stop is not None:
+            mask.append(index <= stop)
+        bool_mask = reduce(partial(cp.logical_and, out=mask[0]), mask)
+        if key.step is None or key.step == 1:
+            return MaskIndexer(BooleanMask(bool_mask, len(index)))
+        else:
+            (map_,) = bool_mask.nonzero()
+            return MapIndexer(
+                GatherMap.from_column_unchecked(
+                    cudf.core.column.as_column(map_[:: key.step]),
+                    len(index),
+                    nullify=False,
+                )
+            )
+    else:
+        parsed_key = index.find_label_range(key)
+        if len(range(len(index))[parsed_key]) == 0:
+            return EmptyIndexer()
+        else:
+            return SliceIndexer(parsed_key)
+
+
 def parse_single_row_loc_key(
     key: Any,
     index: cudf.BaseIndex,
@@ -499,14 +538,7 @@ def parse_single_row_loc_key(
     """
     n = len(index)
     if isinstance(key, slice):
-        # Convert label slice to index slice
-        # TODO: datetime index must be handled specially (unless we go for
-        # pandas 2 compatibility)
-        parsed_key = index.find_label_range(key)
-        if len(range(n)[parsed_key]) == 0:
-            return EmptyIndexer()
-        else:
-            return SliceIndexer(parsed_key)
+        return find_label_range_or_mask(key, index)
     else:
         is_scalar = _is_scalar_or_zero_d_array(key)
         if is_scalar and isinstance(key, np.ndarray):
--- a/python/cudf/cudf/core/indexing_utils.py
+++ b/python/cudf/cudf/core/indexing_utils.py
@@ -3,19 +3,26 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any, Tuple, Union
+from typing import TYPE_CHECKING, Any, Tuple, Union
 
+import numpy as np
 from typing_extensions import TypeAlias
 
 import cudf
+import cudf._lib as libcudf
+from cudf._lib.types import size_type_dtype
 from cudf.api.types import (
     _is_scalar_or_zero_d_array,
     is_bool_dtype,
     is_integer,
     is_integer_dtype,
+    is_scalar,
 )
 from cudf.core.column_accessor import ColumnAccessor
 from cudf.core.copy_types import BooleanMask, GatherMap
+
+if TYPE_CHECKING:
+    from cudf.core.column import ColumnBase
 
 
 class EmptyIndexer:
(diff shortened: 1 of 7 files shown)


wence-:
Explicitly requested the same set of reviewers as #13534.

On the one hand, I would have liked to handle multiindex lookups as well in one change, but unfortunately I think the rules are too complicated and I'm still waiting for clarification on a bunch of ambiguities in the way pandas handles things. As a consequence this doesn't remove as much code as one would have hoped.

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -241,3 +267,356 @@ def parse_row_iloc_indexer(key: Any, n: int) -> IndexingSpec:
...
+
+    Returns
+    -------
+    Single key that will index the rows
+    """
+    (rows,) = destructure_loc_key(key, frame)
+    return rows
+
+
+def ordered_find(needles: "ColumnBase", haystack: "ColumnBase") -> GatherMap:
```
wence-, line 378: This is basically just `DataFrame.merge(sort=True)` but I don't _really_ want to pay the price of making dataframes and all of the merge inspection machinery.

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -241,3 +267,356 @@ def parse_row_iloc_indexer(key: Any, n: int) -> IndexingSpec:
...
+    identically to other index types and fail with a KeyError for
+    an unsorted index if either of the slice endpoints are not unique
+    in the index or are not in the index at all.
+    """
+    if (
+        not (key.start is None and key.stop is None)
+        and isinstance(index, cudf.core.index.DatetimeIndex)
+        and not index.is_monotonic_increasing
+    ):
+        # TODO: datetime index must only be handled specially until pandas 2
```
wence-, line 487: In pandas 2, slicing unsorted datetime indexes works the same as all other indexes (although the warnings were not applied consistently, see pandas-dev/pandas#53983). So should we warning here as well? (Requires extra checking to only warn when the start or stop are not in the index).

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -241,3 +267,356 @@ def parse_row_iloc_indexer(key: Any, n: int) -> IndexingSpec:
...
+                    len(index),
+                    nullify=False,
+                )
+            )
+    else:
+        parsed_key = index.find_label_range(key)
+        if len(range(len(index))[parsed_key]) == 0:
+            return EmptyIndexer()
+        else:
+            return SliceIndexer(parsed_key)
```
wence-, line 512: This is the only bit that is relevant for pandas-2 compat.

python/cudf/cudf/core/indexing_utils.py
```diff
@@ -241,3 +267,356 @@ def parse_row_iloc_indexer(key: Any, n: int) -> IndexingSpec:
...
+    KeyError
+        If a valid type of indexer is provided, but not all keys are
+        found
+    TypeError
+        If the indexing key is otherwise invalid.
+    """
+    if isinstance(index, cudf.MultiIndex):
+        raise NotImplementedError(
+            "This code path is not designed for multiindices"
+        )
```
wence-, line 615: There are two things that need to happen in this case:
1. individual keys can have more cases (e.g. `(a, b):(c, d)` is a valid slice for a multiindex if the multiindex is ordered), a list-like can be not just single-level labels, but tuples of labels)
2. If we get a tuple of per-level indexers we must combine them to a single output indexer and the semantics of this combination in pandas are ambiguous in many cases.

wence- (commented):
Some signposts

vyasr:
@wence- did this stall out just waiting for reviews? If so, I'm very sorry I lost track of it. Please bring it up to date and I'm happy to review this ASAP (probably for 24.08 not 24.06 at this point though).

wence-:
> @wence- did this stall out just waiting for reviews? If so, I'm very sorry I lost track of it. Please bring it up to date and I'm happy to review this ASAP (probably for 24.08 not 24.06 at this point though).

Sorry, it got rather struck on my end, I will try and resurrect for 24.10

vyasr:
I am loathe to close this PR outright given how many issues that it fixes, but I also fear that it is woefully out of sync with the current state of the code base given how much @mroeschke has refactored the underlying code. Also, I doubt that @wence- will have time to return to this work any time soon given the various other major pushes that he is currently involved with.

@mroeschke and @wence- do you think it might make sense for there to be some knowledge transfer here (in the form of pair coding or simply discussion) to try and get this work over the finish line, perhaps in the form of multiple smaller and more easily digestible PRs?

wence-:
Happy to try and do that, yes

mroeschke:
Yeah I can try to digest @wence-'s change here and reimplement it amidst all the refactors. I can reach out to @wence- if I need any clarification.

wence-:
> Yeah I can try to digest @wence-'s change here and reimplement it amidst all the refactors. I can reach out to @wence- if I need any clarification.

Note I also flamed out a bit on the multiindex stuff because I couldn't understand what pandas was doing (and at least when I was looking, was ambiguous in some cases).

See, for reference:

- pandas-dev/pandas#53535
- pandas-dev/pandas#53613
- pandas-dev/pandas#53995
- pandas-dev/pandas#53996
- pandas-dev/pandas#54009
- pandas-dev/pandas#54019

mroeschke:
I've opened https://github.com/rapidsai/cudf/pull/19103 to supersede this PR (sorry I didn't maintain the git history).

Going to close this PR then

Issue #13450, opened 2023-05-25
[FEA] Move `cudf::test::print` out of test into a public API


ttnghia:
With the growth of libcudf development, there is also an increasing number of debugging tasks. Printing a column content is one of the useful debugging approaches. In cudf, we can easily call `cudf::test::print` for that purpose, but the downstream applications don't have access to such functionality in the C++ test module.

I propose to change the function `cudf::test::print` into a public API `cudf::print(column_view const&)` or `cudf::to_string(column_view const&)`.

davidwendt:
I would rather consider a cmake compile option that links the `libcudftestutil.a` with `libcudf.so` such that you can use `cudf::test::print` within libcudf functions temporarily for development.
There is a bunch of code in `cudf::test::print` and I fear moving it to `cudf::print` would generate too many distracting requirements including scalability and performance. I feel it is a perfectly suited for debugging small test cases and would prefer it be maintained in `cudf::test`.

Another option is to use the CSV writer with a [`data_sink` that takes a `std::vector<char>`](https://github.com/rapidsai/cudf/blob/e01e497fe0d1f884ad2966497d1c5f3d6b5fc90b/cpp/include/cudf/io/types.hpp#L508). The CSV writer contains alot of specialized code to convert columns to strings and I would not want to duplicate it unnecessarily.

ttnghia:
I have another idea: continue to keep `cudf::test::print` in the test framework but make that utility function independently compilable in a separate TU. When somebody needs to use it, they can manually change the cmake file to include such TU directly into libcudf.

PR #13720, opened 2023-07-18
Extract `debug_utilities.hpp/cu` from `column_utilities.hpp/cu`

ttnghia:
This PR extracts the implementation of the debug utility function `cudf::test::print()` from `column_utilities.hpp/cu` into its separate header/source files (`debug_utilities.hpp/cu`) for better organizing the relevant code. The new header name is also more expressive and more relevant to its purpose.

The changes in this PR are only moving code around. Not any new functionality or implementation was added.
Closes https://github.com/rapidsai/cudf/issues/13450 (although this is not to address that issue).


davidwendt:
This looks good. Could you describe the steps to add this to libcudf.so for temporary debugging?
Maybe include the steps in the `debug_utilities.hpp` comments.

cpp/include/cudf_test/debug_utilities.hpp
```diff
@@ -20,8 +20,9 @@
 // Do NOT include any header that directly or indirectly depends on gtest.
 // It is to make sure this file can be used and compiled outside of the test framework.
 //
-// In order to use the APIs in this file outside of the test framework, simply include the source
-// file `test/utilities/debug_utilities.cu` into the list of source files of the target build.
+// In order to use the APIs listed below outside of the test framework, simply include this header
+// and add the corresponding source file `test/utilities/debug_utilities.cu` into the CMake
+// target build.
```
davidwendt, line 25: Would it be possible to include the file as a ifdef in the CMakeLists.txt so it could be compiled in with a cmake parameter?

robertmaynard: The source file ` tests/utilities/debug_utilities.cu` is already part of the cudftestutil. We shouldn't be building it part of two different DSO as we could run into duplicate symbol issues from calling tests.

All the libcudf tests and benchmarks have access to this function since they link to to `cudftestutil`. If you are debugging something external, you can gain access to this feature by adding `cudftestutil` to your link line and not be required to recompile libcudf.

ttnghia: Oh I only see your comment after pushed my cmake changes :smile: . Anyway, I tried to avoid duplicate symbols using `if/else`. Please take a look.

ttnghia: In addition, I don't know if it is possible to add `cudftestutil` into `libcudf.so`, not the exec application, since I need to output `cudf::test::print` into `libcudf.so` to be loaded in Spark.

robertmaynard: spark builds libcudf statically right and embeds some extra code into a new `libcudf.so` no reason that can't also embed `cudftestuil`. 

This approach is the wrong way to accomplish our goals.

ttnghia: Okay, so I reverted the approach with cmake changes.
Personally I had (and still have) so many trouble with linking `cudftestutil` so I still prefer manually add the source file. In addition, extracting the `print` module should also make the code base cleaner.

ajschmidt8 (approved):

ajschmidt8 (approved):

cpp/tests/groupby/structs_tests.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include <cudf_test/base_fixture.hpp>
 #include <cudf_test/column_wrapper.hpp>
+#include <cudf_test/debug_utilities.hpp>
```
davidwendt, line 21: Is this include needed?

ttnghia: Yes. This file left a debug function:
```
void print_agg_results(cudf::column_view const& keys, cudf::column_view const& vals)
{
  if constexpr (print_enabled) {
    auto requests = std::vector<cudf::groupby::aggregation_request>{};
    requests.push_back(cudf::groupby::aggregation_request{});
    requests.back().values = vals;
    requests.back().aggregations.push_back(cudf::make_sum_aggregation<cudf::groupby_aggregation>());
    requests.back().aggregations.push_back(
      cudf::make_nth_element_aggregation<cudf::groupby_aggregation>(0));
    auto gby = cudf::groupby::groupby{
      cudf::table_view({keys}), cudf::null_policy::INCLUDE, cudf::sorted::NO, {}, {}};
    auto result = gby.aggregate(requests);
    std::cout << "Results: Keys: " << std::endl;
    cudf::test::print(result.first->get_column(0).view());
    std::cout << "Results: Values: " << std::endl;
    cudf::test::print(result.second.front().results[0]->view());
  }
}
```

cpp/tests/utilities/debug_utilities.cu
```diff
@@ -0,0 +1,485 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//
+// Do NOT include any header that directly or indirectly depends on gtest.
```
bdice, line 18: Does GTest define any macros that we can check for to raise an error?

ttnghia: Good idea. There is a macro `GTEST_INCLUDE_GTEST_GTEST_H_`  that is currently used in `cudf_gtest.hpp` to detect gtest.

cpp/include/cudf_test/debug_utilities.hpp
ttnghia: In fact, adding `cudftestutil` is not very straightforward, and I don't know how to fix this:
```
[INFO]      [exec] /opt/rh/devtoolset-11/root/usr/libexec/gcc/x86_64-redhat-linux/11/ld: /home/nghiat/Devel/jni/target/libcudf-install/lib64/libgtest.a(gtest-all.cc.o): relocation R_X86_64_32S against symbol `_ZTVN7testing8internal17TestEventRepeaterE' can not be used when making a shared object; recompile with -fPIC
```

robertmaynard: You need to enable position independent code ( the -fPIC flag )for the `libgtest ` to resolve those link error. Can you please open this as an issue on `rapids-cmake` and I can fix the issue.

robertmaynard (approved):

cpp/tests/utilities/debug_utilities.cu
davidwendt: The macro check was undone. Is this no longer needed then?

ttnghia: This PR now falls back to a normal refactor PR that extracts `cudf::test::print` into a separate compilation unit.

davidwendt (approved):

PR #13727, opened 2023-07-20
Switch fully unbounded window functions to use aggregations

mythrocks:
## Description

A fully unbounded window function (i.e. [unbounded_preceding, unbounded_following]) need not go through the window function machinery for execution. E.g. Consider the following:

```c++
auto grps = { 0, 0, 0, 0, 1, 1, 1, 1, 2, 2 };
auto vals = { 3, 1, 4, 2, 6, 7, 8, 5, 9, 0 };
```

Running the `MIN` window function on the groups, over an `[UNBOUNDED, UNBOUNDED]` window should produce:

```c++
auto res = { 1, 1, 1, 1, 5, 5, 5, 5, 0, 0 };
```

This result could more easily be achieved using a grouped `MIN` aggregation, and replicating each group's result for every entry in the group.

This commit adds logic to detect fully unbounded windows, and use `groupby::aggregate()` (when one or more grouping keys are specified), or `reduce()` (when there are no grouping keys).

Tangentially, this change also adds the following:
1. A new overload of `cudf::groupby::groupby::aggregate()` that takes a `stream` parameter.
2. A `detail` header to declare the (pre-existing) `cudf::reduction::detail::reduce()` function.

## Checklist


mythrocks:
A couple of tangential changes were necessitated for this one:
1. The "no grouping keys" case requires access to `cudf::reduction::detail::reduce()` for the `stream` version of the function. A declaration was added in a `detail` header.
2. The grouped case requires `cudf::groupby::groupby::aggregate()` to take a stream. An overload was added, per @vyasr's advice.

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -115,6 +128,15 @@ std::unique_ptr<column> grouped_rolling_window(table_view const& group_keys,
   CUDF_EXPECTS((default_outputs.is_empty() || default_outputs.size() == input.size()),
                "Defaults column must be either empty or have as many rows as the input column.");
 
+  // Detect and bypass fully UNBOUNDED windows.
+  if (can_optimize_unbounded_window(preceding_window_bounds.is_unbounded,
+                                    following_window_bounds.is_unbounded,
+                                    min_periods,
+                                    aggr)) {
+    std::cout << "CALEB: Short circuit via aggs." << std::endl;
```
mythrocks, line 136: Argh! I'll remove this debug message.

mythrocks: Removed.

mythrocks:
```
=================================== FAILURES ===================================
______________________ test_jaccard_index_random_strings _______________________
...
>       assert_eq(expected, actual)
...
AssertionError: Series are different
```

mythrocks:
It turns out that the JNI side of unbounded row-based window functions wasn't wired properly: Unbounded windows were treated as "very large, finite windows".

This was corrected, and an appropriate test was added.

mythrocks:
Running correctness tests in Spark right now, but it looks like there's reasonable speedup. 

A simulation of the user query (`MIN` aggregation on unbounded windows for 10K groups of about 150K rows each) went from 730 seconds to 53 seconds on an RTX 6000.

revans2 (approved):
Java code looks good.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
```diff
@@ -0,0 +1,155 @@
...
+  auto is_supported = [](auto const& agg) {
+    switch (agg.kind) {
+      case cudf::aggregation::Kind::COUNT_ALL:
+      case cudf::aggregation::Kind::COUNT_VALID:
+      case cudf::aggregation::Kind::SUM:
+      case cudf::aggregation::Kind::MIN:
+      case cudf::aggregation::Kind::MAX: return true;
+      default:
+        // COLLECT_LIST and COLLECT_SET can be added at a later date.
+        // Other aggregations do not fit into the [UNBOUNDED, UNBOUNDED]
```
bdice, line 45: I'm not super familiar with the "unbounded window" concept. If the window is unbounded in one direction, does it just become a scan / all-prefix-sums operation? It seems like the case of "unbounded in only one direction" should be possible to accelerate, too. Not asking for changes here, just conversation about the current and future scope.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
```diff
@@ -0,0 +1,155 @@
...
+      // Note: cudf::reduce() does not support COUNT_VALID.
+      return cudf::make_count_aggregation<Base>(null_policy::EXCLUDE);
+    } else if constexpr (k == aggregation::Kind::SUM) {
+      return cudf::make_sum_aggregation<Base>();
+    } else if constexpr (k == aggregation::Kind::MIN) {
+      return cudf::make_min_aggregation<Base>();
+    } else if constexpr (k == aggregation::Kind::MAX) {
+      return cudf::make_max_aggregation<Base>();
+    } else {
+      CUDF_FAIL("Unsupported aggregation kind: " + std::to_string(static_cast<int>(k)));
```
bdice, line 78: The other errors I can see of this type do not report the aggregation. I don't think we get much from having an integer here, anyway, and I don't know if there's an easy string conversion.
```suggestion
      CUDF_FAIL("Unsupported aggregation kind.");
```

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -93,6 +93,19 @@ std::unique_ptr<column> grouped_rolling_window(table_view const& group_keys,
 
 namespace detail {
 
+/// Checks if it is possible to optimize fully UNBOUNDED window function.
+bool can_optimize_unbounded_window(bool unbounded_preceding,
```
bdice, line 97: It would be better to use a header for a repeated declaration from another source file.

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -115,6 +128,14 @@ std::unique_ptr<column> grouped_rolling_window(table_view const& group_keys,
   CUDF_EXPECTS((default_outputs.is_empty() || default_outputs.size() == input.size()),
                "Defaults column must be either empty or have as many rows as the input column.");
 
+  // Detect and bypass fully UNBOUNDED windows.
+  if (can_optimize_unbounded_window(preceding_window_bounds.is_unbounded,
```
bdice, line 132: It seems a bit awkward that `window_bounds` has a member `is_unbounded`, while `range_window_bounds` has a method `is_unbounded()`. Should we file an issue to rectify that? The method seems better.

cpp/tests/groupby/count_tests.cpp
```diff
@@ -139,6 +139,40 @@ TYPED_TEST(groupby_count_test, null_keys_and_values)
...
+                    use_sort,
+                    cudf::null_policy::EXCLUDE,
+                    cudf::sorted::NO,
+                    {},
+                    {},
+                    cudf::sorted::NO,
+                    cudf::test::get_default_stream());
+  };
+
+  test_with_stream(cudf::make_count_aggregation<cudf::groupby_aggregation>());
```
bdice, line 169: I think we're using a different directory for stream tests, see https://github.com/rapidsai/cudf/tree/branch-23.08/cpp/tests/streams.

cpp/tests/groupby/groupby_test_util.hpp
```diff
@@ -33,7 +34,8 @@ void test_single_agg(cudf::column_view const& keys,
                      cudf::sorted keys_are_sorted                 = cudf::sorted::NO,
                      std::vector<cudf::order> const& column_order = {},
                      std::vector<cudf::null_order> const& null_precedence = {},
-                     cudf::sorted reference_keys_are_sorted               = cudf::sorted::NO);
+                     cudf::sorted reference_keys_are_sorted               = cudf::sorted::NO,
+                     rmm::cuda_stream_view test_stream = cudf::get_default_stream());
```
bdice, line 38: Note that the other tests in https://github.com/rapidsai/cudf/tree/branch-23.08/cpp/tests/streams are using `cudf::test::get_default_stream()`.

bdice (changes requested):
This is a nice optimization. I have a few comments on its implementation, some of which might require a bit of separate work to prepare the pieces used by this PR.

cpp/include/cudf/groupby.hpp
```diff
@@ -186,6 +186,15 @@ class groupby {
     host_span<aggregation_request const> requests,
     rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
+  /**
+   * @copydoc aggregate(host_span<aggregation_request const>, rmm::mr::device_memory_resource*)
+   *
+   * @param stream CUDA stream used for device memory operations and kernel launches.
+   */
+  std::pair<std::unique_ptr<table>, std::vector<aggregation_result>> aggregate(
```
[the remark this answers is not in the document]
mythrocks: Thanks for reviewing, @bdice.
I might have misunderstood @vyasr's advice on the matter when adding the overload. I figured we would do the larger change to `aggregate()` after this change, i.e. in the next release.

Would you rather we pulled this into this release?

cpp/src/rolling/detail/optimized_unbounded_window.cpp
mythrocks: I think you're right. We might be able to do `UNBOUNDED PRECEDING` as a prefix sums thing. I'm not sure how `UNBOUNDED FOLLOWING` might work, though.

bdice: Suppose you have a column with an `UNBOUNDED FOLLOWING` sum. Then:
- element 0 contains `a_0 + a_1 + ... + a_n`
- element 1 contains `a_1 + a_2 + ... + a_n`
- element n contains `a_n`

Compute a prefix sum over the reversed column, which will contain `a_n, a_n + a_(n-1), ..., a_n + ... + a_0`. Then reverse the elements of this prefix sum, and you should have the desired result. (Caution with commutativity, so that left/right operands are handled appropriately if needed by the operation.)

mythrocks: > the reversed column...

I'm guessing there's a way to do this without materializing the reversed column?

cpp/include/cudf/groupby.hpp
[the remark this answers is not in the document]
bdice: https://github.com/rapidsai/cudf/issues/13737

cpp/src/rolling/detail/optimized_unbounded_window.cpp
bdice: Thrust has `reverse_iterator`, but I haven't investigated specifically whether the algorithms you would use here are iterator-based.

mythrocks: I will bear this in mind for the next round of optimization here.

raydouglass (approved):

cpp/src/rolling/grouped_rolling.cu
mythrocks: This is mostly used internally. I'll fix it here.

cpp/tests/groupby/groupby_test_util.hpp
mythrocks: Line#166 does use `cudf::test::get_default_stream()`.

That said, I have moved the streams test to a separate file under `tests/streams`. This test uses `cudf::test::get_default_stream()`.

cpp/tests/groupby/count_tests.cpp
mythrocks: Thanks for catching that, @bdice. I have moved the test to `test/streams/groupby_test.cpp`.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
```diff
@@ -0,0 +1,155 @@
...
+      // Note: cudf::reduce() does not support COUNT_VALID.
+      return cudf::make_count_aggregation<Base>(null_policy::EXCLUDE);
+    } else if constexpr (k == aggregation::Kind::SUM) {
+      return cudf::make_sum_aggregation<Base>();
+    } else if constexpr (k == aggregation::Kind::MIN) {
+      return cudf::make_min_aggregation<Base>();
+    } else if constexpr (k == aggregation::Kind::MAX) {
+      return cudf::make_max_aggregation<Base>();
+    } else {
+      CUDF_FAIL("Unsupported aggregation kind for optimized unbounded windows.");
```
vyasr, line 78: `COUNT_ALL` and `COUNT_VALID` include checks for the type of the aggregation (`groupby_aggregation`). Does that mean that in some situations `can_optimize_unbounded_window` will return true but this converter would actually fail (if the type of aggregation was unsupported even if a particular kind is supported)? Or are those checks simply there to prevent converting to an unsupported type that would result in a compile-time error (because as your comments point out COUNT_ALL and COUNT_VALID are not supported reduce_aggregations)?

cpp/src/rolling/detail/optimized_unbounded_window.cpp
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+};
+
+template <typename Base>
+std::unique_ptr<Base> convert_to(cudf::rolling_aggregation const& aggr)
```
vyasr, line 84: I don't have a great suggestion here, but I have to note that I find the name `convert_to` confusing to read when I just see the signature. When I see how it's used in practice it's much more readable (`convert_to<X>`) so it's probably good enough.

cpp/src/rolling/detail/optimized_unbounded_window.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace rmm::mr {
+class device_memory_resource;
+}
+
+namespace cudf {
+
+class rolling_aggregation;
+class table_view;
+struct window_bounds;
```
vyasr, line 30: Can this be deleted? Or do consumers of these APIs need this struct to exist for some reason?
```suggestion
```

cpp/src/rolling/detail/optimized_unbounded_window.cpp
```diff
@@ -0,0 +1,155 @@
...
+                                             group_labels,
+                                             cudf::out_of_bounds_policy::DONT_CHECK,
+                                             cudf::detail::negative_index_policy::NOT_ALLOWED,
+                                             stream,
+                                             mr)
+                          ->release();
+  return std::move(result_columns.front());
+}
+
+std::unique_ptr<column> reduction_based_rolling_window(column_view const& input,
```
vyasr, line 121: It would help to add a note that this code path is taking when there is no grouping so the rolling window is over the entire dataset. Right now neither this function nor the one above have any docs that I can see (including in the header since they're not declared there).

cpp/include/cudf/reduction/detail/reduction.hpp
vyasr, line 1: I'm happy seeing this header added. Based on our discussions earlier I do think it's worth having a broader discussion about how much we want to enforce having these detail headers available for every API, but for this PR nothing needs to change IMO.

cpp/tests/CMakeLists.txt
```diff
@@ -622,6 +622,7 @@ ConfigureTest(
 
 ConfigureTest(STREAM_HASHING_TEST streams/hash_test.cpp STREAM_MODE testing)
 ConfigureTest(STREAM_COPYING_TEST streams/copying_test.cpp STREAM_MODE testing)
+ConfigureTest(STREAM_GROUPBY_TEST streams/groupby_test.cpp STREAM_MODE testing)
```
vyasr, line 625: For context, the `STREAM_MODE testing` here is what triggers running the stream validation tests. It is also why we need a separate TU, otherwise the stream validation preload library would be used for all of the tests in the existing groupby tests (and they would fail without additional source code modifications).

cpp/tests/groupby/groupby_test_util.hpp
vyasr: Essentially the requirement for streams testing is that in every code path it must be _possible_ to pass `cudf::test::get_default_stream()` through. https://github.com/rapidsai/cudf/pull/13506 contains a number of changes to existing cudf test utilities where there is no input stream so the code had to be changed to always use `cudf::test::get_default_stream`. Since this util is now supporting passing a stream it's fine for this to be `cudf::get_default_stream()`.

That said, it would probably be easier to remove the stream parameter here and just always use `cudf::test::get_default_stream()` internally in this util. [`cudf::test::get_default_stream()` is just an alias for `cudf::get_default_stream()` under normal circumstances](https://github.com/rapidsai/cudf/blob/branch-23.08/cpp/tests/utilities/default_stream.cpp#L24). It exists to provide a symbol that can be overridden to use a different stream when `STREAM_MODE testing` is set in the CMake configuration for a particular test. Using `cudf::test::get_default_stream()` will therefore "just work" in all cases.

cpp/tests/rolling/grouped_rolling_range_test.cpp
```diff
@@ -207,8 +207,7 @@ struct GroupedRollingRangeOrderByNumericTest : public BaseGroupedRollingRangeOrd
     auto const following = make_unbounded_range_bounds();
     auto const results   = get_grouped_range_rolling_sum_result(preceding, following, *order_by);
 
-    auto const expected_results = bigints_column{{6, 6, 6, 6, 6, 6, 8, 8, 8, 8, 12, 12, 12, 12},
-                                                 cudf::test::iterators::no_nulls()};
```
vyasr, line 211: Why were all the `no_nulls` removed? I guess because it's the default?

cpp/tests/streams/groupby_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+      requests.front().values = vals;
+      if (use_sort == force_use_sort_impl::YES) {
+        requests.front().aggregations.push_back(cudf::make_nth_element_aggregation<cudf::groupby_aggregation>(0));
+      }
+      requests.front().aggregations.push_back(std::move(agg));
+      return requests;
+    }();
+
+    auto gby = cudf::groupby::groupby{cudf::table_view{{keys}}, include_null_keys, keys_are_sorted, {}, {}};
+    auto result = gby.aggregate(requests, test_stream);
```
vyasr, line 53: Should be able to drop the result.
```suggestion
    gby.aggregate(requests, test_stream);
```

cpp/tests/streams/groupby_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+#include <tests/groupby/groupby_test_util.hpp>
+
+#include <cudf_test/base_fixture.hpp>
+#include <cudf_test/column_wrapper.hpp>
+#include <cudf_test/iterator_utilities.hpp>
+#include <cudf_test/type_lists.hpp>
+
+#include <cudf/detail/aggregation/aggregation.hpp>
```
vyasr, line 24: Can we remove this header? If not, is whatever we need available via a public header instead?
```suggestion
```

vyasr (changes requested):
Looks like a good optimization here. Would love to see more perf numbers if you have them (I see one comment indicating that one particular case went from 730 to 53 seconds?). A few minor suggestions for improvement, but in general the stream piece has been handled correctly now (looks like some changes were made after our last discussion). Regarding the detail reduction header I don't think that's critical, but it's a nice-to-have now that it's there.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
mythrocks: The name of the function used to be `to_groupby_aggregation()`, before I added support for `reduce_aggregation`. Converting this to a template yielded: 
```c++
to_groupby_aggregation<groupby_aggregation>;
```
`convert_to<groupby_aggregation>` reads better. But I should probably leave an example there to head off any further confusion.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
mythrocks: The key is in the implementation of `reduction_based_rolling_window()`. You'll notice it doesn't simply call `convert_to<reduce_aggregation>(aggr)`; it needs special handling for `COUNT_ALL`/`COUNT_VALID`. This is because `cudf::reduce()` does not support `COUNT_*`, so the code cannot be uniform.

The check in `convert_to` is to ensure that the `rolling_window` code does not accidentally call `convert_to` for `COUNT_*` as a reduction.

Perhaps I should change that to a `static_assert()`.

mythrocks: I'll change the comment to clarify that `COUNT_*` cannot be a reduce aggregation.

cpp/src/rolling/detail/optimized_unbounded_window.hpp
mythrocks: Good catch. This is an artifact from a previous refactor.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
mythrocks: Good point. Even if the functions are private, it would be best to document the intention.

cpp/include/cudf/reduction/detail/reduction.hpp
mythrocks: I set out to add a detail header for groupby, only to realize that `aggregate()` is a member function, and can't be abstracted similarly. (Not without some obfuscation.)

cpp/tests/CMakeLists.txt
mythrocks: Gotcha. Thank you for the clarification, @vyasr.

cpp/tests/rolling/grouped_rolling_range_test.cpp
mythrocks: This is a slight change in behaviour, in that rolling_window didn't always remove the validity mask, while `groupby::aggregate()` and `cudf::reduce()` do.

cpp/src/rolling/detail/optimized_unbounded_window.hpp
mythrocks: Removed.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
mythrocks: Documented.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
mythrocks: Changed the comments for `COUNT_*`.

cpp/tests/streams/groupby_test.cpp
mythrocks: Yes, looks like this can be removed. Good catch.

vyasr (approved):

mythrocks:
Thanks for the review, @vyasr. I just ran an updated test with `spark-rapids`, to confirm that things are in shape.

@bdice, I should now have addressed the concerns you'd raised in your review. I'd very much appreciate it if you'd have another look.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+/// Converts rolling_aggregation to corresponding reduce/groupby_aggregation.
+template <typename Base>
+struct aggregation_converter {
+  template <aggregation::Kind k>
+  std::unique_ptr<Base> operator()() const
+  {
+    if constexpr (std::is_same_v<Base, cudf::groupby_aggregation> and
+                  k == aggregation::Kind::COUNT_ALL) {
+      // Note: COUNT_ALL cannot be used as a cudf::reduce_aggregation; cudf::reduce does not support
```
bdice, line 65: I think it could be in-scope (for the future) to support reductions for `COUNT_ALL` and `COUNT_VALID`, even if it just means that this conversion/dispatch occurs in a different place (if that movement would enable count reductions, it would be a net benefit imo). That would make this code much less awkward. If you agree, let's file an issue.

bdice (approved):
Thanks @mythrocks. Two comments for forward-looking changes, otherwise approving.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
mythrocks: I agree wholeheartedly. In fact, I intended to spring this on you chaps in 23.10. :]

It makes sense to make this uniform.

cpp/include/cudf/rolling.hpp
```diff
@@ -114,19 +114,34 @@ struct window_bounds {
...
+  [[nodiscard]] size_type value() const { return _value; }
 
  private:
   explicit window_bounds(bool is_unbounded_, size_type value_ = 0)
-    : is_unbounded{is_unbounded_}, value{value_}
+    : _is_unbounded{is_unbounded_}, _value{value_}
   {
   }
+
+  // TODO: In the future, add units for bounds.
```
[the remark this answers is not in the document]
mythrocks: This is an artifact from a time preceding the introduction of `range_window_bounds`. I don't think this applies any more.

Permission to remove this TODO?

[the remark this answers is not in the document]
mythrocks: I have removed this `TODO`, to avoid further confusion.

cpp/src/rolling/detail/optimized_unbounded_window.cpp
mythrocks: https://github.com/rapidsai/cudf/issues/13756

mythrocks:
This change has been merged. Thank you all for your reviews and advice.

PR #13729, opened 2023-07-21
Support more numeric types in `Groupby.apply` with `engine='jit'`

brandon-b-miller:
draft

This PR adds additional numeric dtypes to `GroupBy.apply` with `engine='jit'`.

### DIFF
--- a/python/cudf/cudf/core/udf/groupby_typing.py
+++ b/python/cudf/cudf/core/udf/groupby_typing.py
@@ -17,9 +17,9 @@ from numba.np import numpy_support
 
 index_default_type = types.int64
 group_size_type = types.int64
-SUPPORTED_GROUPBY_NUMBA_TYPES = [types.int64, types.float64]
+SUPPORTED_GROUPBY_NUMBA_TYPES = [types.int32, types.int64, types.float64]
 SUPPORTED_GROUPBY_NUMPY_TYPES = [
-    numpy_support.as_dtype(dt) for dt in [types.int64, types.float64]
+    numpy_support.as_dtype(dt) for dt in SUPPORTED_GROUPBY_NUMBA_TYPES
 ]
 
 
@@ -133,6 +133,23 @@ def _register_cuda_idx_reduction_caller(funcname, inputty):
     call_cuda_functions[funcname.lower()][type_key] = caller
 
 
+class GroupSum(AbstractTemplate):
+    key = "GroupType.sum"
+
+    def generic(self, args, kws):
+        if self.this.group_scalar_type in types.integer_domain:
+            retty = types.int64
+        else:
+            retty = self.this.group_scalar_type
+        return nb_signature(retty, recvr=self.this)
+
+
+def _group_sum_attr(self, mod):
+    return types.BoundFunction(
+        GroupSum, GroupType(mod.group_scalar_type, mod.index_type)
+    )
+
+
 def _create_reduction_attr(name, retty=None):
     class Attr(AbstractTemplate):
         key = name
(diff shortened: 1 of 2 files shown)


brandon-b-miller:
I ran into some issues compiling the shim library with `int16` and `int8` dtypes here. It looks like our implementations of the underlying block level reduction functions rely on [libcu++ atomic_ref](https://nvidia.github.io/libcudacxx/extended_api/synchronization_primitives/atomic_ref.html) which per the documentation `may only be instantiated with a T that are either 4 or 8 bytes`. This seems consistent with the error I'm getting:

```
atomic_cuda.h(111): error: static assertion failed with "atomic_ref does not support 1 or 2 byte types
```

This means that to support shorter ints we will need to do a little more involved development on the c++ side that I think should be separated out into another PR. I'm updating the PR title to reflect that this PR itself only goes as far as rounding us out with `int32` and `float32`.

brandon-b-miller:
raised https://github.com/rapidsai/cudf/issues/13736

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -201,13 +261,30 @@ def resolve_idxmin(self, mod):
...
     _register_cuda_reduction_caller("Max", ty, ty)
     _register_cuda_reduction_caller("Min", ty, ty)
-    _register_cuda_reduction_caller("Sum", ty, ty)
-    _register_cuda_reduction_caller("Mean", ty, types.float64)
-    _register_cuda_reduction_caller("Std", ty, types.float64)
-    _register_cuda_reduction_caller("Var", ty, types.float64)
     _register_cuda_idx_reduction_caller("IdxMax", ty)
     _register_cuda_idx_reduction_caller("IdxMin", ty)
 
+_register_cuda_reduction_caller("Sum", types.int32, types.int64)
```
bdice, line 267: It feels like the typecasting rules were hard-coded twice. Once here, and once in `GroupSum`. Is there a way to reuse a single mapping / type promotion function?

bdice (commented):
I'm not a huge fan of repeating the typecasting logic twice in Python and once in C++... we'll need to find a better architecture for this before we get too deep into adding more types.

python/cudf/cudf/core/udf/groupby_typing.py
brandon-b-miller: Yeah, this is a good point. There should only be one type mapping. Let me see what I can do here.

brandon-b-miller: In the latest commit, I've refactored towards an attribute generating function which checks the existing registry of cuda functions for a match in order to return a signature. I think that makes it so that this is the only place in python where the signatures are written by hand. Let me know what you think.

brandon-b-miller:
@bdice thanks for your review. I've thought a bit about making it so that there's only one centralized place to keep all of the function signatures that could be read directly from c++ or python. Since the shim function library is built before the python is imported, I suppose it should be the source of truth as well. This leads my mind towards some kind of solution which inspects the shim file at import and generates signatures dynamically from it. @gmarkall can `cuobjdump` or `nvdisasm` help us here?

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -133,6 +138,24 @@ def caller(data, index, size):
     call_cuda_functions[funcname.lower()][type_key] = caller
 
 
+def _make_unary_attr(funcname):
+    class GroupUnaryReductionAttrTyping(AbstractTemplate):
+        key = f"GroupType.{funcname}"
+
+        def generic(self, args, kws):
+            for retty, inputty in call_cuda_functions[funcname.lower()].keys():
```
bdice, line 146: Is this function supposed to return `None` or raise an error if no match is found?

bdice (approved):
Much better to only define these once in Python. I'm okay with redefining in C++ and Python if that is needed. It is probably difficult to do this dynamically. I have one question, otherwise approving.

python/cudf/cudf/core/udf/groupby_typing.py
brandon-b-miller: It should return `None`. This signals to the numba typing system that a match was not found.

bdice: I'd like to have an explicit `return None` rather than an implicit no-return.

PR #13750, opened 2023-07-25
Use pinned bounce buffer in the CSV reader

vuule:
## Description
Decompressed input data in the CSV reader is copied to the device in chunks.
This PR modifies the process to copy each chunk to a pinned bounce buffer, from which the data is then copied to the device. This should lead to higher throughput of the H2D copies.

## Checklist

### DIFF
--- a/cpp/src/io/csv/reader_impl.cu
+++ b/cpp/src/io/csv/reader_impl.cu
@@ -251,15 +251,20 @@ std::pair<rmm::device_uvector<char>, selected_rows_offsets> load_data_and_gather
   rmm::device_uvector<char> d_data{
     (load_whole_file) ? data.size() : std::min(buffer_size * 2, data.size()), stream};
   d_data.resize(0, stream);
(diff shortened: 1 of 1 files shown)


vuule:
No perf numbers because my dev system has oddly low throughput that does not change with pinned memory.
Profiles from a lab system show big improvement.

GregoryKimball:
Thank you @vuule for preparing this option for switching `read_csv` to use a pinned buffer for mmap file reads. Here are my quick benchmarking results read files containing LLM-inspired data. I compared recent 23.08 commit `d4c2d1ccee0` with `fc7b81690b18` on this branch. Solid points use the pageable mmap baseline and open circles use the pinned mmap proposal. 

image

The CSV format is the only one expected to change from this diff. We see perhaps a 15% runtime increase from the largest file sizes. I will update this issue after we discuss the merits internally.

vuule:
Closing based on results in the comment above

Issue #11982, opened 2022-10-25
[FEA] Make `experimental/read_json.hpp` the default JSON reading engine


GregoryKimball:
**Is your feature request related to a problem? Please describe.**
Let's make `engine='cudf_experimental'` the default JSON reader. 

**Describe the solution you'd like**

1. refactor cuDF's `read_json` to use the nested JSON reader by default
2. refactor libcudf to promote the experimental JSON reader
3. refactor libcudf to begin deprecating the legacy JSON reader

**Describe alternatives you've considered**
n/a

**Additional context**
Please review the [nested JSON reader](https://github.com/rapidsai/cudf/milestone/13) milestone

vuule:
What is left to do here? IIRC, just internal code reorg to match the new default (namespace shuffle). is that correct?

karthikeyann:
Correct.

PR #13757, opened 2023-07-25
Restructure JSON code to correctly reflect legacy/experimental status

vuule:
## Description

Closes https://github.com/rapidsai/cudf/issues/11982

Moved files to correct directories.
Updated namespaces: experimental -> base, base -> legacy
Improved namespace nesting to reduce the need for fully qualified names (`json::detail` instead of `detail::json`).
Use `host_span` instead of `std::vector &` in all `read_json` variants.

## Checklist


cpp/src/io/json/legacy/json_gpu.cu
```diff
@@ -547,7 +543,7 @@ void convert_json_to_columns(parse_options_view const& opts,
 }
 
 /**
- * @copydoc cudf::io::gpu::detect_data_types
+ * @copydoc cudf::io::json::detail::legacy::detect_data_types
```
vuule, line 546: the namespace was incorrect here AFAICT

cpp/include/cudf/io/detail/json.hpp
```diff
@@ -33,7 +33,7 @@ namespace cudf::io::json::detail {
  *
  * @return cudf::table object that contains the array of cudf::column.
  */
-table_with_metadata read_json(std::vector<std::unique_ptr<cudf::io::datasource>>& sources,
+table_with_metadata read_json(host_span<std::unique_ptr<datasource>> sources,
```
vyasr, line 36: Out of scope for this PR, but a `host_span<unique_ptr>` feels off when it comes to ownership semantics. This pattern seems prevalent in cuIO though. Also I don't know if there's a nice idiomatic way to write this otherwise though since most implicit conversions won't quite work and there's limited interop between implicit conversions and template deduction.

cpp/src/io/json/legacy/read_json.hpp
vyasr, line 1: Why do we need this new header?

vyasr (approved):

cpp/src/io/json/legacy/read_json.hpp
vuule: Because we now call `legacy::read_json` from `read_json.cu`. It does make me wonder if we still need the other `read_json.hpp`.

cpp/include/cudf/io/detail/json.hpp
vuule: Yeah, really sends a mixed message :D
Surely we can lose the `unique_ptr`s somewhere along the call chain...

ttnghia (approved):

cpp/src/io/json/legacy/read_json.hpp
vuule: Just checked, the other header is needed for tests; keeping the files as they are.

PR #13767, opened 2023-07-26
Support `corr` in `GroupBy.apply` through the jit engine

shwina:
## Description
This PR enables computing the pearson correlation between two columns of a group within a UDF. Concretely, syntax such as the following will be allowed and produce the same result as pandas:

```python
ans = df.groupby('key').apply(lambda group_df: group_df['x'].corr(group_df['y']))
```

## Checklist


python/cudf/udf_cpp/shim.cu
```diff
@@ -662,3 +662,47 @@ make_definition_idx(BlockIdxMax, int64, int64_t);
...
+  sum_sq_ls[threadIdx.x]  = 0.0;
+  sum_sq_rs[threadIdx.x]  = 0.0;
+
+  auto block = cooperative_groups::this_thread_block();
+
+  for (int64_t idx = block.thread_rank(); idx < size; idx += block.size()) {
+    double delta_l = lhs_ptr[idx] - lhs_mean;
+    double delta_r = rhs_ptr[idx] - rhs_mean;
+
+    numerators[idx] = delta_l * delta_r;
```
bdice, line 692: We should reimplement BlockVar to be a special case of a new function for computing Covariance (where the variance is the covariance of the data with itself). Then I think you can use the covariance function to compute the numerators and the variance function for the terms in the denominator.

brandon-b-miller:
This should now be ready for review.

bdice:
@brandon-b-miller Before requesting review, can we get a title and description for the PR?

brandon-b-miller:
> @brandon-b-miller Before requesting review, can we get a title and description for the PR?

Updated! :)

brandon-b-miller:
I wondered while working on this PR to what extent `corr` should be supported as a one-off vs writing general machinery for any similar function that maps two vectors to one scalar (the dot product comes to mind as something we already support, I'm sure there's other things). Would be interested in others opinions on this.

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -104,7 +104,22 @@ def __init__(self, dmm, fe_type):
...
+        f"Block{funcname}_{lty}_{rty}",
+        retty(types.CPointer(lty), types.CPointer(rty), group_size_type),
+    )
+
+    def caller(lhs, rhs, size):
+        return cuda_func(lhs, rhs, size)
+
+    call_cuda_functions.setdefault(funcname.lower(), {})
+
+    type_key = (retty, lty, rty)
```
bdice, line 118: I think these parentheses are extraneous.
```suggestion
    type_key = retty, lty, rty
```

python/cudf/udf_cpp/shim.cu
```diff
@@ -437,37 +437,49 @@ __device__ double BlockMean(T const* data, int64_t size)
...
   }
 
-  cuda::atomic_ref<double, cuda::thread_scope_block> ref{block_var};
-  ref.fetch_add(local_var, cuda::std::memory_order_relaxed);
+  cuda::atomic_ref<double, cuda::thread_scope_block> ref{block_covar};
+  ref.fetch_add(local_covar, cuda::std::memory_order_relaxed);
   block.sync();
 
-  if (block.thread_rank() == 0) { block_var = block_var / static_cast<double>(size - 1); }
+  if (block.thread_rank() == 0) { block_covar = block_covar / static_cast<double>(size - 1); }
```
bdice, line 473:
```suggestion
  if (block.thread_rank() == 0) { block_covar /= static_cast<double>(size - 1); }
```

python/cudf/udf_cpp/shim.cu
```diff
@@ -437,37 +437,55 @@ __device__ double BlockMean(T const* data, int64_t size)
...
   block.sync();
 
-  T local_sum      = 0;
-  double local_var = 0;
-
-  device_sum<T>(block, data, size, &block_sum);
+  device_sum<T>(block, lhs, size, &block_sum_lhs);
+  auto const mu_l = static_cast<double>(block_sum_lhs) / static_cast<double>(size);
+  auto const mu_r = [=]() {
+    if (lhs == rhs) {
```
bdice, line 459:
```suggestion
    if (lhs == rhs) {
      // If the lhs and rhs are the same, this is calculating variance.
      // Thus we can reuse the value of mu_l.
```

python/cudf/udf_cpp/shim.cu
```diff
@@ -437,37 +437,55 @@ __device__ double BlockMean(T const* data, int64_t size)
...
 
-  auto const mean = static_cast<double>(block_sum) / static_cast<double>(size);
+  double local_covar = 0;
 
   for (int64_t idx = block.thread_rank(); idx < size; idx += block.size()) {
-    auto const delta = static_cast<double>(data[idx]) - mean;
-    local_var += delta * delta;
+    auto const delta =
+      (static_cast<double>(lhs[idx]) - mu_l) * (static_cast<double>(rhs[idx]) - mu_r);
+    local_covar += delta;
```
bdice, line 472:
```suggestion
    local_covar +=
      (static_cast<double>(lhs[idx]) - mu_l) * (static_cast<double>(rhs[idx]) - mu_r);
```

python/cudf/cudf/tests/test_groupby.py
```diff
@@ -433,6 +435,20 @@ def func(df):
     run_groupby_apply_jit_test(groupby_jit_data, func, ["key1"])
 
 
+@pytest.mark.parametrize("dtype", SUPPORTED_GROUPBY_NUMPY_TYPES)
+def test_groupby_apply_jit_correlation(groupby_jit_data, dtype):
```
bdice, line 439: Do we need to test data with NaNs? Infinity? Empty groups? Negative numbers? etc.

I'd like to see stronger test coverage for much more of our JIT code paths, not just `corr`...

bdice (commented):
Just needs stronger test cases, and a couple minor tweaks.

python/cudf/cudf/tests/test_groupby.py
brandon-b-miller: Closing the loop on this conversation, after some discussion offline it was found that significant changes are needed to robustly support special values for this reduction which we'll tackle in a separate pull request.

bdice: Please file an issue for this -- and we also need to test the behavior of existing functions like variance and standard deviation for NaN support.

bdice (approved):
Some follow-up work but this looks good to me.

PR #13768, opened 2023-07-26
Raise error on constructing an array from mixed type inputs

galipremsagar:
## Description
We currently are allowing construction of mixed-dtype by type-casting them into a common type as below:
```python

In [1]: import cudf

In [2]: import pandas as pd

In [3]: s = pd.Series([1, 2, 3], dtype='datetime64[ns]')


In [5]: p = pd.Series([10, 11])

In [6]: new_s = pd.concat([s, p])

In [7]: new_s
Out[7]: 
0    1970-01-01 00:00:00.000000001
1    1970-01-01 00:00:00.000000002
2    1970-01-01 00:00:00.000000003
0                               10
1                               11
dtype: object

In [8]: cudf.Series(new_s)
Out[8]: 
0   1970-01-01 00:00:00.000000
1   1970-01-01 00:00:00.000000
2   1970-01-01 00:00:00.000000
0   1970-01-01 00:00:00.000010
1   1970-01-01 00:00:00.000011
dtype: datetime64[us]
```
This behavior is incorrect and we are getting this from `pa.array` constructor. This PR ensures we do proper handling around such cases and raise an error.

## Checklist


python/cudf/cudf/tests/test_index.py
```diff
@@ -2897,3 +2897,16 @@ def test_scalar_getitem(self, index_values, i):
         assert not isinstance(index[i], cudf.Index)
         assert index[i] == index_values[i]
         assert_eq(index, index.to_pandas())
+
+
+def test_index_mixed_dtype_error():
+    pi = pd.Index(
+        [
+            pd.Timestamp("1970-01-01 00:00:00.000000001"),
+            pd.Timestamp("1970-01-01 00:00:00.000000002"),
```
mroeschke, line 2906: Could you test with `pd.Timedelta` too? (We had a similar issue with Timedelta not so long ago)

galipremsagar: Done 👍

mroeschke (approved):

python/cudf/cudf/core/column/column.py
```diff
@@ -2335,18 +2340,29 @@ def as_column(
...
+                )
+
+                if (
+                    isinstance(arbitrary, pd.Index)
+                    and arbitrary.dtype == cudf.dtype("object")
+                    and isinstance(
+                        pyarrow_array, (pa.DurationArray, pa.TimestampArray)
+                    )
+                ):
+                    retry = False
```
bdice, line 2356: The need for this `retry` logic highlights the existing complexity of the `as_column` function... we should really try to refactor this at some point, because a single try-except no longer suffices.

I suppose you could raise a different error class than `TypeError` and re-raise a TypeError from that message... like this:

```python
class MixedTypeError(TypeError):
    pass

try:
    raise MixedTypeError("Cannot create column with mixed types")
except TypeError as e:
    if isinstance(e, MixedTypeError):
        raise TypeError(str(e))
```

I think that solution might be clearer than "retry" because it's not totally clear what is being retried. I typically think of "retry" as a `while True` loop.

bdice (commented):
One suggestion for improvement.

python/cudf/cudf/core/column/column.py
galipremsagar: Done 👍

bdice (approved):

PR #13770, opened 2023-07-26
Relax protobuf pinnings.

bdice:
## Description
This PR relaxes cudf's protobuf pinnings to help with compatibility issues. `cudf` uses `protobuf` in two places.

The first place `protobuf` is used is at build time, to generate a Python module from a `.proto` file in `python/cudf/cmake/Modules/ProtobufHelpers.cmake`: https://github.com/rapidsai/cudf/blob/f8e5a89e983065e1202f1151dd499bea3102a537/python/cudf/cmake/Modules/ProtobufHelpers.cmake#L16-L17

The second place `protobuf` is used is in the generated file `python/cudf/cudf/utils/metadata/orc_column_statistics_pb2.py` which is [imported here](https://github.com/rapidsai/cudf/blob/f8e5a89e983065e1202f1151dd499bea3102a537/python/cudf/cudf/io/orc.py#L14-L16).

The generated Python module used at runtime should be compatible with newer versions of `protobuf` than the version used to build the Python module, from my understanding of https://protobuf.dev/support/cross-version-runtime-guarantee/. Therefore, we only require that the runtime pinning of `protobuf` is of the same major version and an equal-or-greater minor version. That allows us to relax this pinning.

Follow-up to #12864, see that PR for more context.

## Checklist

### DIFF
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -64,7 +64,8 @@ dependencies:
 - pandoc
 - pip
 - pre-commit
-- protobuf>=4.21.6,<4.22
+- protobuf==4.21
+- protobuf>=4.21,<5
 - ptxcompiler
 - pyarrow==11.0.0.*
 - pydata-sphinx-theme
--- a/conda/environments/all_cuda-120_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-120_arch-x86_64.yaml
@@ -62,7 +62,8 @@ dependencies:
 - pandoc
 - pip
 - pre-commit
-- protobuf>=4.21.6,<4.22
+- protobuf==4.21
+- protobuf>=4.21,<5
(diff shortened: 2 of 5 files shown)


conda/environments/all_cuda-118_arch-x86_64.yaml
```diff
@@ -64,7 +64,8 @@ dependencies:
 - pandoc
 - pip
 - pre-commit
-- protobuf>=4.21.6,<4.22
+- protobuf==4.21
+- protobuf>=4.21,<5
```
shwina, line 68: What does it mean when we list a package twice like this?

bdice: Fixed.

galipremsagar (approved):

raydouglass (approved):
Should technically be `<5.0.0a0` (I think the minor & patch might be required) to prevent pre-release versions, but it is unlikely that `conda-forge` will upload a pre-release, so I will approve.

Issue #13741, opened 2023-07-24
[BUG] Constructing a `DataFrame` with a multi-level column does not work


shwina:
```python
columns = cudf.MultiIndex.from_tuples([('A', 'one'), ('A', 'two')])
df = cudf.DataFrame(np.random.randn(2, 2), columns=columns)
df
```

Output:

```
  (A, one)  (A, two)
0 -1.004870  1.253556
1  0.176658  0.856762
```

Expected output:

```python
          A
        one       two
0 -0.926630  0.718061
1 -0.920203 -0.074759
```

wence-:
Maybe this?
```diff
diff --git a/python/cudf/cudf/core/dataframe.py b/python/cudf/cudf/core/dataframe.py
index 0fe8949090..cc57091539 100644
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -722,6 +722,8 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
 
         if dtype:
             self._data = self.astype(dtype)._data
+        # Fix up
+        self._data.multiindex = isinstance(columns, pd.MultiIndex)
 
     @_cudf_nvtx_annotate
     def _init_from_series_list(self, data, columns, index):
```

Issue #13740, opened 2023-07-24
[BUG] Column level names lost when constructing `DataFrame`


shwina:
When passing a `MultIiIndex` with level names as the `columns=` argument to a `DataFrame`, the level names are lost:

```python
import numpy as np
import cudf

columns = cudf.MultiIndex.from_tuples(
    [
        ("A", "cat"),
        ("B", "dog"),
        ("B", "cat"),
        ("A", "dog"),
    ],
    names=["exp", "animal"],
)
df = cudf.DataFrame(np.random.randn(8, 4), columns=columns)
print(columns.names)
print(df.columns.names)
```

Output:

```
['exp', 'animal']
[None]
```

wence-:
Possibly like so:
```diff
diff --git a/python/cudf/cudf/core/dataframe.py b/python/cudf/cudf/core/dataframe.py
index 0fe8949090..318f66b62c 100644
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -722,6 +722,10 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
 
         if dtype:
             self._data = self.astype(dtype)._data
+        # Fix up
+        if isinstance(columns, pd.MultiIndex):
+            self._data.multiindex = True
+            self._data._level_names = tuple(columns.names)
 
     @_cudf_nvtx_annotate
     def _init_from_series_list(self, data, columns, index):
```

This is inappropriate intimacy, so is not a good fix. The issue with the `DataFrame` constructor is that it has ad-hoc handling of the columns argument that depends very much on the type of the `data` argument. Some code-paths look at the columns properly and produce the appropriate `ColumnAccessor`, others don't.

galipremsagar:
Something to note, this is an issue with regular Index in various code-flows. I have a PR coming up soon to cover them

galipremsagar:
Fixed by #13772

PR #13772, opened 2023-07-26
Preserve names of column object in various APIs

galipremsagar:
## Description
This PR preserves column names in various APIs by retaining `self._data._level_names` and also calculating when to preserve the column names.
Fixes: #13741, #13740

## Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -1820,19 +1824,29 @@ def _make_operands_and_index_for_binop(
...
 
         if _is_scalar_or_zero_d_array(other):
             rhs = {name: other for name in self._data}
+            equal_columns = True
         elif isinstance(other, Series):
             rhs = dict(zip(other.index.values_host, other.values_host))
             # For keys in right but not left, perform binops between NaN (not
             # NULL!) and the right value (result is NaN).
             left_default = as_column(np.nan, length=len(self))
+            equal_columns = other.index.to_pandas().equals(
```
mroeschke, line 1844: Just noting that this will still be `True` if the `name`s are different

```
In [4]: pd.Index([1], name="foo").equals(pd.Index([1], name="bar"))
Out[4]: True
```

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -2483,8 +2484,12 @@ def test_bitwise_binops_series(pdf, gdf, binop):
 
 
 @pytest.mark.parametrize("unaryop", [operator.neg, operator.inv, operator.abs])
-def test_unaryops_df(pdf, gdf, unaryop):
-    d = unaryop(pdf - 5)
+@pytest.mark.parametrize("col_name", [None, "abc"])
```
shwina, line 2487: Instead, can we not assign `pdf.columns.name` as part of the fixture?

python/cudf/cudf/core/dataframe.py
galipremsagar: Yup

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -3465,14 +3469,23 @@ def _binaryop(
...
             other, op, fill_value, reflect, can_reindex
         )
         if operands is NotImplemented:
             return NotImplemented
 
         return self._from_data(
-            ColumnAccessor(type(self)._colwise_binop(operands, op)),
+            ColumnAccessor(
+                type(self)._colwise_binop(operands, op),
+                level_names=None
```
bdice, line 3485: I'd prefer to define `level_names = ...` just above the `return` statement since the `return` is already fairly complicated.

python/cudf/cudf/tests/test_dataframe.py
bdice: Or you can write something like
```python
@pytest.mark.parametrize("col_name,assign_col_name", [(None, False), (None, True), ("abc", True)])
```

and then
```
if assign_col_name:
    pd_df.columns.name = col_name
```

Not a perfect solution but avoids duplicating the entire test.

bdice: Or use some other sentinel value like `False` or `...` (`Ellipsis`) rather than a boolean flag `assign_col_name`... just trying to be creative. 😄

bdice (approved):
A few small comments but nothing blocking. LGTM!

python/cudf/cudf/core/indexed_frame.py
galipremsagar: Done 👍

python/cudf/cudf/tests/test_dataframe.py
galipremsagar: > Or you can write something like
> 
> ```python
> @pytest.mark.parametrize("col_name,assign_col_name", [(None, False), (None, True), ("abc", True)])
> ```
> 
> and then
> 
> ```
> if assign_col_name:
>     pd_df.columns.name = col_name
> ```
> 
> Not a perfect solution but avoids duplicating the entire test.

Used this approach 👍

python/cudf/cudf/core/dataframe.py
```diff
@@ -1820,19 +1824,29 @@ def _make_operands_and_index_for_binop(
...
+                self._data.to_pandas_index()
+            )
+            can_use_self_column_name = equal_columns or (
+                list(other._index._data.names) == self._data._level_names
+            )
```
wence-, line 1849: Ah, I saw this in #13778, can we only make `equal_columns` if the cheap check is False?

wence- (approved):
One small comment, but I think this makes sense.

Although, the index alignment is pretty hairy...

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -3465,14 +3469,23 @@ def _binaryop(
...
+                if not can_use_self_column_name
+                else self._data._level_names,
```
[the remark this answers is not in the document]
galipremsagar: Yup, will address this in https://github.com/rapidsai/cudf/pull/13778

python/cudf/cudf/core/dataframe.py
galipremsagar: Addressed in #13778

python/cudf/cudf/core/indexed_frame.py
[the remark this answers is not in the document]
galipremsagar: Addressed in #13778

PR #13775, opened 2023-07-26
Refactor Parquet reader handling of V2 page header info

etseidl:
## Description
This PR replaces the `def_lvl_bytes` and `rep_lvl_bytes` fields of the `gpu::PageInfo` struct with an array indexed by `gpu::level_type` (as is done with the `lvl_decode_buf` array). This allows for some streamlining in `InitLevelSection()`, removing some redundant code and improving readability.

See this [comment](https://github.com/rapidsai/cudf/pull/13707#discussion_r1274118429) for context.

## Checklist


cpp/src/io/parquet/page_decode.cuh
```diff
@@ -901,17 +901,10 @@ inline __device__ uint32_t InitLevelSection(page_state_s* s,
...
-    len = lvl == level_type::DEFINITION ? s->page.def_lvl_bytes : s->page.rep_lvl_bytes;
     s->abs_lvl_start[lvl] = cur;
-    if (len == 0) {
-      s->initial_rle_run[lvl]   = s->page.num_input_values * 2;  // repeated value
-      s->initial_rle_value[lvl] = 0;
-      s->lvl_start[lvl]         = cur;
-    } else {
-      init_rle(cur, cur + len);
-    }
+    init_rle(cur, cur + len);
```
ttnghia, line 907: Why do we have such big change here? Will this a breaking change?

etseidl: it's because now I set and test `len` in the `if`, so `len` is guaranteed non-zero and I can get rid of the redundant branching for the `len==0` case.

vuule: Oh, do we now hit the else branch at 915 instead? I see similar code with the branch that's removed here.

etseidl: Yes, exactly.  The original fix from #13707 created a duplicate path to get to 0 bits for the level RLE encoding. This refactor gets rid of the duplicate.

vuule (approved):

ttnghia (approved):

PR #13778, opened 2023-07-27
Fix binary operation column ordering and missing column issues

galipremsagar:
## Description
This PR fixes various cases in binary operations where columns are of certain dtypes and the binary operations on those dataframes and series don't yield correct results, correct resulting column types, or have missing columns altogether. 
This PR also introduces ensuring column ordering to match pandas binary ops column ordering when pandas compatibility mode is enabled.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -1845,7 +1845,8 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
                 self._data.to_pandas_index()
             )
             can_use_self_column_name = (
-                equal_columns or can_use_self_column_name
+                equal_columns
+                or list(other._index._data.names) == self._data._level_names
             )
         elif isinstance(other, DataFrame):
             if (
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -1841,6 +1841,22 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
             # For keys in right but not left, perform binops between NaN (not
             # NULL!) and the right value (result is NaN).
             left_default = as_column(np.nan, length=len(self))
+            can_use_self_column_name = (
+                list(other._index._data.names) == self._data._level_names
+            )
+            if cudf.get_option("mode.pandas_compatible"):
+                equal_columns = other.index.to_pandas().equals(
+                    self._data.to_pandas_index()
+                )
+                can_use_self_column_name = (
+                    equal_columns or can_use_self_column_name
+                )
+            else:
+
+                if not can_use_self_column_name:
+                    can_use_self_column_name = other.index.to_pandas().equals(
+                        self._data.to_pandas_index()
+                    )
             equal_columns = other.index.to_pandas().equals(
                 self._data.to_pandas_index()
             )
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -1852,7 +1852,6 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
                     equal_columns or can_use_self_column_name
                 )
             else:
-
                 if not can_use_self_column_name:
                     can_use_self_column_name = other.index.to_pandas().equals(
                         self._data.to_pandas_index()
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -1856,12 +1856,6 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
                     can_use_self_column_name = other.index.to_pandas().equals(
                         self._data.to_pandas_index()
                     )
-            equal_columns = other.index.to_pandas().equals(
-                self._data.to_pandas_index()
-            )
-            can_use_self_column_name = equal_columns or (
-                list(other._index._data.names) == self._data._level_names
-            )
         elif isinstance(other, DataFrame):
             if (
                 not can_reindex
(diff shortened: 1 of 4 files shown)


galipremsagar:
Note that there is some intersection with changes in https://github.com/rapidsai/cudf/pull/13772/

python/cudf/cudf/core/dataframe.py
```diff
@@ -1876,7 +1891,22 @@ def _make_operands_and_index_for_binop(
...
+
+        if cudf.get_option("mode.pandas_compatible") and not equal_columns:
+            if isinstance(other, DataFrame):
+                column_names_list = self._data.to_pandas_index().join(
+                    other._data.to_pandas_index(), how="outer"
+                )
+            elif isinstance(other, Series):
+                column_names_list = self._data.to_pandas_index().join(
+                    other.index.to_pandas(), how="outer"
+                )
```
bdice, line 1903: Does this need an "else raise"? I don't want to run into an unbound variable error.
```suggestion
                )
            else:
                raise ValueError("other must be a DataFrame or Series.")
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -1876,7 +1891,22 @@ def _make_operands_and_index_for_binop(
...
+
+            sorted_dict = {}
+            for key in column_names_list:
+                sorted_dict[key] = operands[key]
```
bdice, line 1907:
```suggestion
            sorted_dict = {key: operands[key] for key in column_names_list}
```

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -3465,14 +3465,23 @@ def _binaryop(
...
             other, op, fill_value, reflect, can_reindex
         )
         if operands is NotImplemented:
             return NotImplemented
 
         return self._from_data(
-            ColumnAccessor(type(self)._colwise_binop(operands, op)),
+            ColumnAccessor(
+                type(self)._colwise_binop(operands, op),
+                level_names=None
```
bdice, line 3481: Same as in the other PR -- would prefer to see `level_names` defined before the `return` statement.

galipremsagar: Yup, I've made this change in the other PR. I'll merge #13772 before this one and bring those changes into this PR.

python/cudf/cudf/core/dataframe.py
```diff
@@ -1820,19 +1820,29 @@ def _make_operands_and_index_for_binop(
...
+            rhs = dict(zip(other.index.to_pandas(), other.values_host))
             # For keys in right but not left, perform binops between NaN (not
             # NULL!) and the right value (result is NaN).
             left_default = as_column(np.nan, length=len(self))
+            equal_columns = other.index.to_pandas().equals(
+                self._data.to_pandas_index()
+            )
+            can_use_self_column_name = equal_columns or (
+                list(other._index._data.names) == self._data._level_names
+            )
```
wence-, line 1845: Can we do the cheap thing first So that we don't convert to pandas and check equality if we don't need to?

wence- (approved):
I think this looks fine with one comment that we can potentially avoid one "expensive" check?

python/cudf/cudf/core/dataframe.py
galipremsagar: We could change this to something like this:
```python
if cudf.get_option("mode.pandas_compatible"):
    equal_columns = other.index.to_pandas().equals(self._data.to_pandas_index())
else:
    can_use_self_column_name = (list(other._index._data.names) == self._data._level_names)
    if not can_use_self_column_name:
        can_use_self_column_name = other.index.to_pandas().equals(self._data.to_pandas_index())
```

galipremsagar: Switched to a cheaper condition first.

vyasr (approved):
LGTM

python/cudf/cudf/core/dataframe.py
```diff
@@ -1837,10 +1837,25 @@ def _make_operands_and_index_for_binop(
...
+                    equal_columns or can_use_self_column_name
+                )
+            else:
+                if not can_use_self_column_name:
+                    can_use_self_column_name = other.index.to_pandas().equals(
+                        self._data.to_pandas_index()
+                    )
             equal_columns = other.index.to_pandas().equals(
                 self._data.to_pandas_index()
             )
```
bdice, line 1861: This is really repetitive, the `equal_columns` is being re-evaluated too many different times. In this `if`, but also in the `else` and also below the `if/else`!

galipremsagar: Oh wait, it was messy merge.

galipremsagar: Okay now removed the repetitive parts

python/cudf/cudf/core/dataframe.py
```diff
@@ -1837,16 +1837,25 @@ def _make_operands_and_index_for_binop(
...
+                    self._data.to_pandas_index()
+                )
+                can_use_self_column_name = (
+                    equal_columns or can_use_self_column_name
+                )
+            else:
+                if not can_use_self_column_name:
+                    can_use_self_column_name = other.index.to_pandas().equals(
+                        self._data.to_pandas_index()
+                    )
```
bdice, line 1858: This should be logically equivalent:
```suggestion
            if not can_use_self_column_name:
                can_use_self_column_name = other.index.to_pandas().equals(
                    self._data.to_pandas_index()
                )
```

galipremsagar: Discussed offline and decided to remove pandas compatibility mode logic from here.

bdice (approved):

PR #13779, opened 2023-07-27
Fix contains(`in`) method for `Series`

galipremsagar:
## Description
Checking for boolean values in a range results in incorrect behavior:
```python
In [1]: True in range(0, 0)
Out[1]: False


In [3]: True in range(0, 2)
Out[3]: True
```

This results in the following bug:
```python
In [23]: s = cudf.Series([True, False])

In [24]: s[0]
Out[24]: True

In [25]: type(s[0])
Out[25]: numpy.bool_

In [26]: True in s
Out[26]: True

In [26]: True in s.to_pandas()
Out[26]: False
```

This PR fixes this issue by properly checking if an integer is passed to the `RangeIndex. __contains__`

## Checklist

### DIFF
--- a/python/cudf/cudf/core/index.py
+++ b/python/cudf/cudf/core/index.py
@@ -311,6 +311,10 @@ class RangeIndex(BaseIndex, BinaryOperand):
             item, tuple(np.sctypes["int"] + np.sctypes["float"] + [int, float])
         ):
             return False
+        try:
+            item = pd.core.dtypes.common.ensure_python_int(item)
+        except TypeError:
+            return False
         if not item % 1 == 0:
             return False
         return item in range(self._start, self._stop, self._step)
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/index.py
```diff
@@ -311,6 +311,10 @@ def __contains__(self, item):
             item, tuple(np.sctypes["int"] + np.sctypes["float"] + [int, float])
         ):
             return False
+        try:
+            item = pd.core.dtypes.common.ensure_python_int(item)
```
mroeschke, line 315: `ensure_python_int` only really calls `int(item)` in essence. Maybe to avoid the private pandas import `int` can just be used and catch the `TypeError` and `ValueError`?

galipremsagar: I was under an assumption this was a public method, I'm trying to avoid the following:
```python
In [1]: int(True)
Out[1]: 1
```

mroeschke: Ah I see. The `pandas.core` namespace is considered "private" (things could get moved around without warning), but I suppose this is okay to use. `dtypes/common.py` doesn't tend to change much

bdice (approved):

mroeschke (approved):

PR #13780, opened 2023-07-27
Fix negative unary operation for boolean type

galipremsagar:
## Description
Negative unary op on boolean series is resulting in conversion to `int` type:
```python
In [1]: import cudf

In [2]: s = cudf.Series([True, False])

In [3]: s
Out[3]: 
0     True
1    False
dtype: bool

In [4]: -s
Out[4]: 
0   -1
1    0
dtype: int8

In [5]: -s.to_pandas()
Out[5]: 
0    False
1     True
dtype: bool
```
The PR fixes the above issue by returning inversion of the boolean column instead of multiplying with `-1`.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/frame.py
+++ b/python/cudf/cudf/core/frame.py
@@ -97,6 +97,7 @@ class Frame(BinaryOperand, Scannable):
     def _has_nulls(self):
         return any(col.has_nulls() for col in self._data.values())
 
+    @_cudf_nvtx_annotate
     def serialize(self):
         header = {
             "type-serialized": pickle.dumps(type(self)),
@@ -106,6 +107,7 @@ class Frame(BinaryOperand, Scannable):
         return header, frames
 
     @classmethod
+    @_cudf_nvtx_annotate
     def deserialize(cls, header, frames):
         cls_deserialize = pickle.loads(header["type-serialized"])
         column_names = pickle.loads(header["column_names"])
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/tests/test_unaops.py
```diff
@@ -123,3 +123,9 @@ def test_scalar_no_negative_bools():
         ),
     ):
         -x
+
+
+def test_series_bool_neg():
+    sr = Series([True, False, True, False, True, True])
```
mroeschke, line 129: Maybe good to test with NA too?

python/cudf/cudf/core/frame.py
```diff
@@ -1891,8 +1891,17 @@ def __rmatmul__(self, other):
         return self.dot(other, reflect=True)
 
     # Unary logical operators
+    @_cudf_nvtx_annotate
```
bdice, line 1894: Can we annotate everything in this file, or drop the annotation here? Just want consistency. Fine to do that in a separate PR.

python/cudf/cudf/tests/test_unaops.py
galipremsagar: Done 👍

python/cudf/cudf/core/frame.py
galipremsagar: Annotated in this PR 👍

bdice (approved):

wence- (approved):

Issue #13600, opened 2023-06-21
[BUG] cudf_kafka Consumer not able to read data


chinmaychandak:
I am using cuStreamz with cudf 23.06. 

When using the Confluent Kafka engine, things work as expected. When using the accelerated `cudf_kafka` datasource, I see the below error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_1704/3564192403.py in <module>
----> 1 stream_cudf_kafka.start()

/opt/conda/envs/rapids/lib/python3.9/site-packages/streamz/core.py in start(self)
    340         """ Start any upstream sources """
    341         for upstream in self.upstreams:
--> 342             upstream.start()
    343 
    344     def stop(self):

/opt/conda/envs/rapids/lib/python3.9/site-packages/streamz/sources.py in start(self)
    595             # blocks for consumer thread to come up and invoke poll to establish
    596             # connection with broker to fetch oauth token for kafka
--> 597             self.consumer.poll(timeout=1)
    598             self.consumer.get_watermark_offsets(tp)
    599             self.loop.add_callback(self.poll_kafka)

AttributeError: 'Consumer' object has no attribute 'poll'
```

@jdye64 / @randerzander - Can you please help take a look?

GregoryKimball:
Thank you @chinmaychandak for reporting this. Would you please share a bit more background? Have you been relying on the `cudf_kafka` datasource for some time, or are you trying to use it for the first time? Are there any old versions of cuDF that work correctly for you?

chinmaychandak:
Hey @GregoryKimball! 

We're resuming dev on `custreamz` after 1.5+ years. cuStreamz uses `cudf_kafka` to fetch batches of data from Kafka in an accelerated manner (data gets passed from `librdkafka` to `libcudf` via the JSON reader in the C/C++ layer bypassing the Python layer and effectively the GIL). `cudf_kafka` has the same API as `confluent_kafka`.

This worked well for versions around `0.20`/ `21.06` (long time ago, ikr!). 

Here's a smaller MRE:
```
import confluent_kafka as ck
from custreamz import kafka

kafka_brokers = "somekafkabroker:9092"
kafka_consumer_configs = {
    "bootstrap.servers": kafka_brokers,
    "group.id": "mre-checkpoint",
    "auto.offset.reset": "earliest"
}
topic = "topic"

# initializing consumer works
[...]

PR #13782, opened 2023-07-28
Add 'poll' function to custreamz kafka consumer

jdye64:
## Description
Streamz has updated their codebase to include a call to the Confluent Kafka Consumer library function 'poll'. Currently custreamz does not include this method. This PR adds the 'poll' function to custreamz to simply proxy the call to the underlying confluent kafka library so that streamz is no longer broken for end users. Without this function end users are no longer able to use custreamz with newer versions of the streamz library.

This closes: #13600


bdice:
Does this need to target 23.08 or 23.10? Breakage with new versions of streamz sounds like a candidate for 23.08 despite code freeze.

jdye64:
23.10 is fine. Caught it early and the version of streamz with the breaking change hasn’t even been released yet

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jdye64:
@bdice I don't have the permissions to alter the labels, hence `Label Checker` CI failure. Can you add those for me?

vyasr:
Added labels. I assumed that this could be treated as a new feature even though technically we would be broken by the new version of the library.

chinmaychandak:
Hey guys, would it be possible to merge this PR some time this week? This is blocking our cuStreamz dev work.

python/custreamz/custreamz/kafka.py
```diff
@@ -270,3 +271,21 @@ def commit(self, offsets=None, asynchronous=True):
...
+        Consumes a single message, calls callbacks and returns events.
+
+        The application must check the returned Message object's
+        Message.error() method to distinguish between proper messages
+        (error() returns None), or an event or error
+        (see error().code() for specifics).
+
+        Parameters
+        ----------
+        timeout : float,
```
bdice, line 286:
```suggestion
        timeout : float
```

bdice (commented):
@jdye64 Are there dependency changes needed to support this? Are these pinnings still accurate or do they need to be updated? https://github.com/rapidsai/cudf/blob/989c4116a3b054f46a7f1152f234b2601d2583b0/dependencies.yaml#L496-L508

jdye64:
@bdice the 

> @jdye64 Are there dependency changes needed to support this? Are these pinnings still accurate or do they need to be updated?
> 
> https://github.com/rapidsai/cudf/blob/989c4116a3b054f46a7f1152f234b2601d2583b0/dependencies.yaml#L496-L508
> 
> https://github.com/rapidsai/cudf/blob/989c4116a3b054f46a7f1152f234b2601d2583b0/conda/recipes/custreamz/meta.yaml#L51

No, the dependencies are still fine as is. We didn't add a new function but rather just added a hook to call one that had always existed.

bdice (approved):
Looks fine to me. @jdye64 Do we need to add tests for this?

Do you have merge privileges? Do you want me to merge after CI passes?

PR #13787, opened 2023-07-31
Fix default behavior of index metaclass instance and subclass checks

vyasr:
## Description

The current implementation of IndexMeta gives the wrong result for subclasses of Index: for instance, `DatetimeIndex.__instancecheck__(DatetimeIndex(), DatetimeIndex)` or `DatetimeIndex.__subclasscheck__(DatetimeIndex, DatetimeIndex)` would return False. In the case of `isinstance`, however, we have been saved by https://bugs.python.org/issue35083, wherein Python is silently injecting a `if type(instance) is self: return True` short-circuit. `issubclass` currently does have the wrong behavior, though. The fix is to fall back to the original behavior of `isinstance`/`issubclass` rather than hardcoding a boolean result. That will also ensure that we get the correct behavior if e.g. a subclass of `Index` is itself subclassed, e.g. if we introduced a `class SpecialDatetimeIndex(DatetimeIndex)`.

## Checklist


galipremsagar (approved):
Verified locally, Thanks @vyasr for fixing this!

PR #13789, opened 2023-07-31
Fix all warnings in Python docs

vyasr:
## Description

The Sphinx documentation has historically been warning-filled, which makes it difficult to identify when there are real issues like missing APIs. This PR fixes all the current issues and converts warnings to errors during the build, ensuring that doc builds are reliable indicators of issues in the future.

I will say that there are a few changes that may not be exactly what we want, particularly in cases of including APIs that may not be documented in exactly the same way in pandas. However, I think we'd be better off merging this PR so that we can get to a 0 warnings state and then work through further improvements in follow-ups where the build will be more robust.

Here is an inexhaustive list of the most significant changes:
- Adds all missing BaseIndex APIs. The goal of BaseIndex is to provide an abstract interface defining all functions that should match pandas.Index, but up until now some methods were missing. This PR does not implement any new ones, it either lifts existing implementations up from subclasses (where those implementations are generic for all Index types) or it simply defines them as returning NotImplemented. The result is that all methods at least exist so that docs don't complain.
- Cleans up the listed APIs in rst files so that all existing APIs are included somewhere and no nonexistent APIs are listed anywhere. APIs that don't have an exact equivalent in the pandas docs are given a new home in these docs. That includes pieces like extension dtypes, which were previously documented in the user guide and therefore weren't part of any summary list (causing warnings).
- Fixed missing dependencies for doc notebooks.
- Fixed various formatting issues with docstrings, especially around bulleted lists that were missing the requisite spacing to be rendered correctly.
- Fixing header ordering (going from level 1 to level 3 headings is a warning in Sphinx) and links in notebooks. 

## Checklist


ci/build_docs.sh
ajschmidt8, line 1: is there a reason the `sphinx-build` commands were switched to using `make`?

when we rolled out the docs build scripts for GHAs across RAPIDS, we intentionally avoided using wrapper scripts like `make` and `build.sh` so that it was overtly clear what commands were being used to build the docs and to avoid requiring additional dependencies.

if there's a good reason for this change, we should open an issue to update all of RAPIDS for consistency.

bdice: I would also vote to avoid make here.

ajschmidt8: I can see the benefit of consolidating/single-sourcing the sphinx build arguments in the Makefile, but we can also do that with a bash variable.

Unless the `make` command is gonna be reused by other scripts like `build.sh`, I would vote against adding `make`.

If we want to incorporate using `make` with `build.sh` in the future, I'd be open to that but I think that should be its own effort.

Vyas, let us know your thoughts here.

vyasr: In my experience the `make` interface is the standard way of building Sphinx docs. I have never seen developer docs for a package recommend building documentation by invoking `sphinx-build` directly. I just randomly picked three large projects from [Sphinx's official list of projects](https://www.sphinx-doc.org/en/master/examples.html) and all three recommend `make html` in their dev docs ([flask](https://github.com/pallets/flask/blob/main/CONTRIBUTING.rst#building-the-docs), [click](https://github.com/pallets/click/blob/main/CONTRIBUTING.rst#building-the-docs), and [pytest](https://docs.pytest.org/en/7.4.x/)). Prior to working on RAPIDS I have only ever run sphinx-build manually when I wanted to test out options. 

> I can see the benefit of consolidating/single-sourcing the sphinx build arguments in the Makefile

This is the main reason to make this change in this PR. For better or worse I think developers who choose to build the docs will default to `make html`, and they will then see different behaviors locally than in CI. Yes, they could do the full approach of downloading the CI container and rerunning the build_docs CI script, but ideally we'd make the behavior as close as possible even in local development environments. For example, [rapids-compose uses the Makefile](https://github.com/trxcllnt/rapids-compose/blob/main/etc/bash-utils.sh#L1043), not sphinx-build (I would assume that devcontainers would as well, although AFAICT that's not implemented yet).

> If we want to incorporate using make with build.sh in the future, I'd be open to that but I think that should be its own effort.

FWIW, right now:
- rmm and cudf don't support building docs in build.sh
- [cuml](https://github.com/rapidsai/cuml/blob/branch-23.08/build.sh#L294) and [cugraph](https://github.com/rapidsai/cugraph/blob/branch-23.08/build.sh#L396) both use `make html`
- [raft](https://github.com/rapidsai/raft/blob/branch-23.08/build.sh#L496) uses `sphinx-build`

So there is a slight preference for `make html` in our existing build.sh out of the main libraries at least. There's certainly no consistent approach atm.

> we intentionally avoided using wrapper scripts like make and build.sh so that it was overtly clear what commands were being used to build the docs and to avoid requiring additional dependencies.

In sum, I'd argue the `make html` is really closer to the de facto standard for building sphinx than directly invoking `sphinx-build` anyway. [The Sphinx docs themselves introduce it immediately in their quickstart guides](https://www.sphinx-doc.org/en/master/usage/quickstart.html#running-the-build). I do see the argument for avoiding additional dependencies, but in this case I think there's good reason to do so.

ajschmidt8: Ok, I'm fine with it in that case.

I think we should strive for the repositories to be consistent.

Vyas, can you open an Ops issue to track this effort with a list of repositories that will need to be updated?

vyasr: Agreed. Yes, I can open an ops issue.

vyasr: https://github.com/rapidsai/ops/issues/2772

vyasr: Closing as resolved.

docs/cudf/source/user_guide/data-types.md
```diff
@@ -16,11 +16,11 @@ All data types in cuDF are [nullable](missing-data).
 | Floating-point       | `'float32'`, `'float64'`                                                                          |
 | Datetime             | `'datetime64[s]'`, `'datetime64[ms]'`, `'datetime64['us']`, `'datetime64[ns]'`                    |
 | Timedelta (duration) | `'timedelta[s]'`, `'timedelta[ms]'`, `'timedelta['us']`, `'timedelta[ns]'`                        |
-| Category             | {py:func}`cudf.CategoricalDtype`                                                                  |
+| Category             | `cudf.CategoricalDtype`                                                                           |
```
galipremsagar, line 19: Looks like there are now dedicated pages for `cudf.CategoricalDtype`, `cudf.ListDtype`, `cudf.StructDtype`, `cudf.Decimal...Dtype`, is it possible to link them here? Right now they seem to be appearing as text only here.

ajschmidt8 (approved):

ajschmidt8 (changes requested):
blocking while I investigate some build path changes resulting from this PR

ajschmidt8:
It looks like the build paths are different as a result of the changes in this PR.

For example, compare the preview environments below:

- this PR - https://downloads.rapids.ai/ci/cudf/pull-request/13789/4f3435f/docs/cudf/html/
- #13813 https://downloads.rapids.ai/ci/cudf/pull-request/13813/2579c64/docs/cudf/html/

I see the same problem on the `rmm` PR.

docs/dask_cudf/Makefile
```diff
@@ -3,8 +3,9 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
+SPHINXOPTS    ?= -n -v
```
ajschmidt8, line 6: Is there a reason why these flags don't match the `cudf` Makefile flags?

Will warnings be treated as errors for `dask-cudf` in a follow-up PR?

docs/cudf/Makefile
ajschmidt8, line 1: I think the directory discrepancy comes from the Makefile's using `-M` as the builder option instead of `-b`.

That's the only argument I can see that's different from the previous invocation and may also affect the output directory.

See the `sphinx-build` docs here for reference: https://www.sphinx-doc.org/en/master/man/sphinx-build.html.

According to the docs, it looks like `-M` is the newer builder argument, so we should probably keep that and just update the `build_docs.sh` scripts to adjust their paths accordingly.

docs/dask_cudf/Makefile
vyasr: Yes, I didn't want to also try and fix dask-cudf issues in the same PR. This one is large enough already. I also don't know dask-cudf as well so may want to get someone more familiar to work through those issues.

docs/cudf/Makefile
vyasr: Sounds good, I'll update the file.

docs/cudf/source/user_guide/data-types.md
vyasr: Hmm I wonder why I'm not seeing warnings. Let me see what's going on here.

vyasr: I needed to add in the roles again so that they would be found. Everything seems to be working again now. For some reason it requires fully qualified names for the dtypes (`cudf.CategoricalDtype` doesn't work because it's in the inventory as `cudf.core.dtypes.CategoricalDtype`), which is unfortunate but something that I think we should address in a follow-up.

ajschmidt8 (approved):

galipremsagar (approved):

PR #13795, opened 2023-08-01
Example code for blog on new row comparators

divyegala:
## Description
Example code using a few libcudf APIs to demonstrate nested-type usage.

## Checklist


karthikeyann:
@divyegala which is a good location for adding code to round tripping data using JSON reader and writer?

karthikeyann:
Added metadata round trip for json reader, writer usages.

divyegala:
@karthikeyann can you rebase and remove your commits? I removed `example.cpp` and only have `deduplication.cpp` in my working branch locally. When I try to merge your commits and run the examples, I get `std::bad_alloc()` and I'm probably doing a bad merge somewhere. It'll be easier for me to push my commits and for you to apply yours on top of mine

karthikeyann:
The cmake compilation didn’t work right away.
Added `-DCMAKE_CUDA_ARCHITECTURES=native`  to all cmake commands in cpp/examples/build.sh to compile.

GregoryKimball:
@divye Would you please update this example to use an RMM pool? @davidwendt added this in `examples/strings/common.hpp`

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,173 @@
...
+int main(int argc, char const** argv)
+{
+  std::string input_filepath;
+  std::string output_filepath;
+  std::string mr_name;
+  if (argc < 3) {
+    input_filepath  = "example.json";
+    output_filepath = "output.json";
+    mr_name         = "pool";
+  } else if (argc == 3) {
```
GregoryKimball, line 139:
```suggestion
  if (argc < 4) {
    input_filepath  = "example.json";
    output_filepath = "output.json";
    mr_name         = "pool";
  } else if (argc == 4) {
```

GregoryKimball:
@vyasr and @davidwendt would you please share your feedback on the way this example demonstrates the libcudf public API?

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,173 @@
...
+ */
+auto make_pool_mr()
+{
+  return rmm::mr::make_owning_wrapper<rmm::mr::pool_memory_resource>(make_cuda_mr());
+}
+
+/**
+ * @brief Create memory resource for libcudf functions
+ */
+std::shared_ptr<rmm::mr::device_memory_resource> create_memory_resource(std::string const& name)
```
PointKernel, line 52:
```suggestion
std::shared_ptr<rmm::mr::device_memory_resource> create_memory_resource(std::string_view const name)
```
nit: `string_view` instead of `string &`

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+#include <rmm/mr/device/cuda_memory_resource.hpp>
+#include <rmm/mr/device/device_memory_resource.hpp>
+#include <rmm/mr/device/owning_wrapper.hpp>
+#include <rmm/mr/device/pool_memory_resource.hpp>
+
+#include <chrono>
+#include <iostream>
+#include <string>
+
```
PointKernel, line 35: I find it useful to have one or two paragraphs summarizing what we want to demonstrate in the current example, e.g. https://github.com/NVIDIA/cuCollections/blob/dev/examples/static_map/device_view_example.cu#L31-L42

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,173 @@
...
+  return cudf::io::read_json(options);
+}
+
+void write_json(cudf::table_view tbl, cudf::io::table_metadata metadata, std::string filepath)
+{
+  // write the data for inspection
+  auto sink_info = cudf::io::sink_info(filepath);
+  auto builder2  = cudf::io::json_writer_options::builder(sink_info, tbl).lines(true);
+  builder2.metadata(metadata);
+  auto options2 = builder2.build();
```
PointKernel, line 72:
```suggestion
  auto builder  = cudf::io::json_writer_options::builder(sink_info, tbl).lines(true);
  builder.metadata(metadata);
  auto options = builder.build();
```
Not a hard requirement: there are no naming conflicts between this and the above `read_json` function thus `builder` and `options` are good enough IMO. If we want more descriptive names: could use `writer_builder` and `writer_options` instead.

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+cudf::io::table_with_metadata read_json(std::string filepath)
+{
+  auto source_info = cudf::io::source_info(filepath);
+  auto builder     = cudf::io::json_reader_options::builder(source_info).lines(true);
+  auto options     = builder.build();
+  return cudf::io::read_json(options);
+}
+
+void write_json(cudf::table_view tbl, cudf::io::table_metadata metadata, std::string filepath)
```
PointKernel, line 66:
```suggestion
void write_json(cudf::table_view input, cudf::io::table_metadata input_metadata, std::string filepath)
```
Not saying the suggestions are better. Just throwing ideas to use more descriptive names and avoid abbreviations.

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,173 @@
...
+    return 1;
+  }
+
+  auto resource = create_memory_resource(mr_name);
+  rmm::mr::set_current_device_resource(resource.get());
+
+  // read input file
+  auto [tbl, metadata] = read_json(input_filepath);
+
+  auto st = std::chrono::steady_clock::now();
```
PointKernel, line 154:
```suggestion
  auto const start_time = std::chrono::steady_clock::now();
```

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+  auto count                               = count_aggregate(tbl->view());
+  std::chrono::duration<double> count_time = std::chrono::steady_clock::now() - st;
+  std::cout << "Wall time: " << count_time.count() << " seconds\n";
```
PointKernel, line 158: The example output would look like this:
```bash
Wall time: 1.0 seconds
Wall time: 2.0 seconds
Wall time: 3.0 seconds
Wall time: 4.0 seconds
```
Could we print out more specific timing info? something like (just the rough idea, you see what I mean):
```bash
Count aggregation time: 1.0 seconds
Join time: 2.0 seconds
Sort time: ...
```

PointKernel (commented):
Some minor suggestions.

Probably worth adding doxygen docs for all local functions as well.

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,173 @@
...
+  requests.emplace_back(cudf::groupby::aggregation_request());
+  auto agg = cudf::make_count_aggregation<cudf::groupby_aggregation>();
+  requests[0].aggregations.push_back(std::move(agg));
+  requests[0].values = *val;
+  auto agg_results   = grpby_obj.aggregate(requests);
+  auto result_key    = std::move(agg_results.first);
+  auto result_val    = std::move(agg_results.second[0].results[0]);
+
+  auto left_cols = result_key->release();
+  left_cols.push_back(std::move(result_val));
```
davidwendt, line 94: I know this may be out of scope for this specific example but I've always found the groupby API logic very difficult to follow. I understand the current API pattern may provide the most flexibility but I don't see how someone could come up with this logic just from the documentation. Since the documentation would not be able to cover every possible aggregation, I think it would serve libcudf users better by developing a good set of groupby examples.

divyegala: Thanks @davidwendt, I entirely agree with your sentiment here. Even after writing the example I still could not write a groupby aggregation from memory. We went through iterations of different aggravations while developing this example and every time I would copy-paste groupby code.

davidwendt: groupby-aggregations == aggravations 😆

divyegala: My phone autocorrect knows what's up!

cpp/examples/nested_types/deduplication.cpp
GregoryKimball: One comment from the console parsing side, using the pattern: 
`{func_name} time: X seconds` is easier to parse on a `.split(' ')`

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+  auto st = std::chrono::steady_clock::now();
+
+  auto count_st                            = std::chrono::steady_clock::now();
+  auto count                               = count_aggregate(tbl->view());
+  std::chrono::duration<double> count_time = std::chrono::steady_clock::now() - count_st;
+  std::cout << "count_aggregate time: " << count_time.count() << " seconds\n";
+
+  auto count_w_copy_st = std::chrono::steady_clock::now();
+  auto count_w_copy    = count_aggregate(tbl->view());
```
GregoryKimball, line 185:
```suggestion
  auto count_w_copy    = count_aggregate_with_copy(tbl->view());
```

cpp/examples/nested_types/deduplication.cpp
divyegala: As we are using `tbl` abbrevation everywhere for `table_view` type, I'd like to keep it as that

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * @brief Create memory resource for libcudf functions
+ */
+std::shared_ptr<rmm::mr::device_memory_resource> create_memory_resource(std::string_view name)
+{
+  if (name == "pool") { return make_pool_mr(); }
+  return make_cuda_mr();
+}
+
+cudf::io::table_with_metadata read_json(std::string filepath)
```
bdice, line 76: Use a `string_view` instead of a `string`.
```suggestion
cudf::io::table_with_metadata read_json(std::string_view filepath)
```

cpp/examples/nested_types/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+set(CPM_DOWNLOAD_VERSION v0.35.3)
+file(
+  DOWNLOAD
+  https://github.com/cpm-cmake/CPM.cmake/releases/download/${CPM_DOWNLOAD_VERSION}/get_cpm.cmake
+  ${CMAKE_BINARY_DIR}/cmake/get_cpm.cmake
+)
+include(${CMAKE_BINARY_DIR}/cmake/get_cpm.cmake)
+
+set(CUDF_TAG branch-23.10)
```
bdice, line 19:
```suggestion
set(CUDF_TAG branch-23.12)
```

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * @file deduplication.cpp
+ * @brief Demonstrates usage of the libcudf APIs to perform operations on nested-type tables.
+ *
+ * The algorithms chosen to be demonstrated are to showcase nested-type row operators of three
+ * kinds:
+ * 1. hashing: Used by functions `count_aggregate` and `join_count` to hash inputs of any type
+ * 2. equality: Used by functions `count_aggregate` and `join_count` in conjunction with hashing
+ *              to determine equality for nested types
```
bdice, line 45: Don't indent this. Match item 3.
```suggestion
 * 2. equality: Used by functions `count_aggregate` and `join_count` in conjunction with hashing
 * to determine equality for nested types
```

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+ * 3. lexicographic: Used by function `sort_keys` to create a lexicographical order for nested-types
+ * so as to enable sorting
+ *
+ * @note This example is for demonstration purposes only. It is not intended to show the most
+ * performant way to do the example algorithm.
```
bdice, line 50: Should we annotate any code with comments explaining how to do it fast? It can be confusing when examples don't show best practices, especially for a performance-focused library.

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * @brief Join each row with its duplicate counts
+ */
+std::unique_ptr<cudf::table> join_count(cudf::table_view left, cudf::table_view right)
+{
+  auto [left_indices, right_indices] =
+    cudf::inner_join(cudf::table_view{{left.column(0)}}, cudf::table_view{{right.column(0)}});
+  auto new_left  = cudf::gather(left, cudf::device_span<int const>{*left_indices});
+  auto new_right = cudf::gather(right, cudf::device_span<int const>{*right_indices});
```
bdice, line 127:
```suggestion
  auto new_left  = cudf::gather(left, cudf::device_span<cudf::size_type const>{*left_indices});
  auto new_right = cudf::gather(right, cudf::device_span<cudf::size_type const>{*right_indices});
```

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+}
+
+/**
+ * @brief Aggregate count of duplicate rows in nested-type column
+ */
+std::unique_ptr<cudf::table> count_aggregate(cudf::table_view tbl)
+{
+  // Get count for each key
+  auto keys = cudf::table_view{{tbl.column(0)}};
+  auto val  = cudf::make_numeric_column(cudf::data_type{cudf::type_id::INT32}, keys.num_rows());
```
bdice, line 101: Do count aggregations return `int32_t` or `size_type`? I think it's supposed to be `size_type` but not 100% sure.
```suggestion
  auto val  = cudf::make_numeric_column(cudf::data_type{cudf::type_to_id<cudf::size_type>()}, keys.num_rows());
```

cpp/examples/nested_types/deduplication.cpp
bdice: I'm not sure what you mean by "everywhere" but we're pretty consistent about naming `table_view` objects `input` in libcudf. I would adopt @PointKernel's suggestion throughout.

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+  // read input file
+  auto [tbl, metadata] = read_json(input_filepath);
+
+  auto const st = std::chrono::steady_clock::now();
+
+  auto const count_st                            = std::chrono::steady_clock::now();
+  auto count                                     = count_aggregate(tbl->view());
+  std::chrono::duration<double> const count_time = std::chrono::steady_clock::now() - count_st;
+  std::cout << "count_aggregate time: " << count_time.count() << " seconds\n";
```
bdice, line 184: Is showing timing results helpful when the example input is so small?

cpp/include/cudf/detail/gather.cuh
```diff
@@ -483,7 +483,7 @@ struct column_gatherer_impl<struct_view> {
     auto const nullable =
       nullify_out_of_bounds || std::any_of(sliced_children.begin(),
                                            sliced_children.end(),
-                                           [](auto const& col) { return col.nullable(); });
+                                           [](auto const& col) { return col.has_nulls(); });
```
bdice, line 486: Quite a few of the changes in this PR (here and below) don't belong in a PR aimed at introducing new example code. They should be fixed in a separate PR.

cpp/examples/nested_types/deduplication.cpp
divyegala: The example input is just an example, after all. You could use your own JSON input file with this.

cpp/examples/nested_types/deduplication.cpp
divyegala: Everywhere in this example. Will change it.

cpp/examples/nested_types/deduplication.cpp
divyegala: I don't think any of the code in this example is not the most-performant way to run that algorithm in isolation. However, a user could theoretically find other ways to pipeline this example better based on their input and requirements, and a disclaimer helps cover that scenario.

cpp/examples/nested_types/deduplication.cpp
divyegala: It's just a dummy column to force usage of hash aggregation instead of sort aggregation, it does not hold the return values.

cpp/examples/nested_types/deduplication.cpp
ttnghia: Our (implicit) convention in cudf is to use `input` "everywhere". My convention/recommendation is to use `_tview` or `_tv` suffix for `table_view` variables, but only for local variables. APIs' input parameters should (almost) always be `input`.

cpp/include/cudf/detail/gather.cuh
ttnghia: So if this PR requires such fix to work, please post a fix PR and mark as dependency of this one.

cpp/examples/nested_types/deduplication.cpp
bdice: I would just drop this disclaimer. Otherwise it implies that there's a _known_ better way, which leaves me confused as a reader.

cpp/examples/nested_types/deduplication.cpp
bdice: I'm not sure if I understand. We usually want to rely on hash aggregation anytime we can, since it tends to be faster than sort aggregations. Can you point to the behavior that you're avoiding, and why this is the right solution (as opposed to fixing the aggregation dispatch to use hashing instead of sorting)?

cpp/examples/nested_types/deduplication.cpp
bdice: I agree, but I think it's unlikely that a user will want to use this exact pipeline. In other words, it's not helpful to show timing if the provided input example is so small. I think you have two choices:
1. just execute the pipeline without showing timing, since you're mostly just measuring launch latency.
2. write a script that generates a sufficiently large input file (where the timing becomes a meaningful measure of speed) and ship it with this example.

cpp/examples/nested_types/deduplication.cpp
divyegala: https://github.com/rapidsai/cudf/blob/abc0d41d1d9033d581948ae19384e0aa0f33da77/cpp/src/groupby/hash/groupby.cu#L654-L656

divyegala: This isn't the right solution. It should definitely be fixed at the core of the problem.

cpp/examples/nested_types/deduplication.cpp
divyegala: @GregoryKimball should we ship your script that generates large inputs?

cpp/examples/nested_types/deduplication.cpp
bdice: Is this something we now have the machinery to fix? I'm not sure what stops this from working. We should have full nested type support now.

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+  return std::make_unique<cudf::table>(std::move(left_cols));
+}
+
+/**
+ * @brief Sort nested-type column
+ */
+std::unique_ptr<cudf::table> sort_keys(cudf::table_view tbl)
+{
+  auto sort_order = cudf::sorted_order(cudf::table_view{{tbl.column(0)}});
```
vyasr, line 141: suggestion (non-blocking): Perhaps it's worth adding a comment that if stability is desired you could use `stable_sorted_order` here. Someone is always bound to want that and including information about it in more places seems helpful.

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+  return cudf::gather(tbl, *sort_order);
+}
+
+/**
+ * @brief Main for nested_types examples
+ *
+ * Command line parameters:
+ * 1. JSON input file name/path (default: "example.json")
+ * 2. JSON output file name/path (default: "output.json")
+ * 3. Memory resource (optional): "pool" or "cuda" (default: "pool")
```
vyasr, line 151: suggestion (blocking): let's include a sample command-line invocation to help users out.

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+    output_filepath = "output.json";
+    mr_name         = "pool";
+  } else if (argc == 4) {
+    input_filepath  = argv[1];
+    output_filepath = argv[2];
+    mr_name         = argv[3];
+  } else {
+    std::cout << "Either provide all command-line arguments, or none to use defaults" << std::endl;
+    return 1;
+  }
```
vyasr, line 171: suggestion (non-blocking): this error isn't quite doing what you want. It's probably better to be explicit here.

```suggestion
  std::string input_filepath;
  std::string output_filepath;
  std::string mr_name;
  if (argc != 4 && argc != 1) {
    std::cout << "Either provide all command-line arguments, or none to use defaults" << std::endl;
    return 1;
  }
  if (argc == 1) {
    input_filepath  = "example.json";
    output_filepath = "output.json";
    mr_name         = "pool";
  } else {
    input_filepath  = argv[1];
    output_filepath = argv[2];
    mr_name         = argv[3];
  }
```

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+  auto resource = create_memory_resource(mr_name);
+  rmm::mr::set_current_device_resource(resource.get());
+
+  // read input file
+  auto [tbl, metadata] = read_json(input_filepath);
+
+  auto const st = std::chrono::steady_clock::now();
+
+  auto const count_st                            = std::chrono::steady_clock::now();
```
vyasr, line 181: nitpick: This seems redundant. Can we just use `count_st` as the overall start when computing the total elapsed time at the end?

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+  auto count                                     = count_aggregate(tbl->view());
+  std::chrono::duration<double> const count_time = std::chrono::steady_clock::now() - count_st;
```
vyasr, line 183: nitpick: Can this be auto-deduced, or will the operator result in lower precision by default or something like that?

cpp/examples/nested_types/deduplication.cpp
vyasr: thought: I'm inclined to say we should drop the timings. The performance of this example is not itself going to be valuable.

cpp/include/cudf/detail/gather.cuh
vyasr: suggestion (non-blocking): Ideally I'd also split the nvtx range changes from the nullable->has_nulls changes since they're fairly different classes of issues.

vyasr (changes requested):

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+  auto builder   = cudf::io::json_writer_options::builder(sink_info, tbl).lines(true);
+  builder.metadata(metadata);
+  auto options = builder.build();
+  cudf::io::write_json(options);
+}
+
+/**
+ * @brief Aggregate count of duplicate rows in nested-type column
+ */
+std::unique_ptr<cudf::table> count_aggregate(cudf::table_view tbl)
```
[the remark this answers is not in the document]
divyegala: I think of these functions as analogous to libcudf public APIs, which don't use `auto` return type. It's more readable for users reading this example to know the return type of the function.

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,205 @@
...
+ */
+auto make_pool_mr()
+{
+  return rmm::mr::make_owning_wrapper<rmm::mr::pool_memory_resource>(make_cuda_mr());
+}
+
+/**
+ * @brief Create memory resource for libcudf functions
+ */
+std::shared_ptr<rmm::mr::device_memory_resource> create_memory_resource(std::string_view name)
```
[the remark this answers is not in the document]
vyasr: Ah got it. Then maybe a boolean flag `pool=false` would convey that more simply than a string, which is very flexible. Up to you though.

cpp/examples/build.sh
```diff
@@ -29,3 +29,11 @@ STRINGS_EXAMPLE_BUILD_DIR=${STRINGS_EXAMPLE_DIR}/build
...
 # Build
 cmake --build ${STRINGS_EXAMPLE_BUILD_DIR} -j${PARALLEL_LEVEL}
+
+# Nested Types example
+NESTED_TYPES_EXAMPLE_DIR=${EXAMPLES_DIR}/nested_types
+NESTED_TYPES_EXAMPLE_BUILD_DIR=${NESTED_TYPES_EXAMPLE_DIR}/build
+# Configure
+cmake -S ${NESTED_TYPES_EXAMPLE_DIR} -B ${NESTED_TYPES_EXAMPLE_BUILD_DIR} -Dcudf_ROOT="${LIB_BUILD_DIR}"
+# Build
+cmake --build ${NESTED_TYPES_EXAMPLE_BUILD_DIR} -j${PARALLEL_LEVEL}
```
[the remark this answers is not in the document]
vyasr: Wow I totally just wrote this like it was a CMake file and not a bash script... Sorry about that.

cpp/examples/nested_types/deduplication.cpp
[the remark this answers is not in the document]
vyasr: Good point, agreed.

cpp/examples/build.sh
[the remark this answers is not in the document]
vyasr: I'm going to implement this now.

cpp/examples/nested_types/deduplication.cpp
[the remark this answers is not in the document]
vyasr: Looks like this was done in https://github.com/rapidsai/cudf/pull/13795/commits/f7e96a99716a383f43d6066653f1558af247cdd1

cpp/include/cudf/detail/gather.cuh
vyasr: Looks like this was done in a sequence of commits (e.g. https://github.com/rapidsai/cudf/pull/13795/commits/f2c42be2c8868a23375892921f7f9aae0544568b for range).

cpp/examples/nested_types/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+CPMFindPackage(
+  NAME cudf GIT_REPOSITORY https://github.com/rapidsai/cudf
+  GIT_TAG ${CUDF_TAG}
+  GIT_SHALLOW
+    TRUE
+    SOURCE_SUBDIR
+    cpp
+)
```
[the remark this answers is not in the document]
vyasr: Done.

cpp/examples/nested_types/deduplication.cpp
vyasr: If we don't force hash aggregations do we still have at least some? As long as the example is demonstrating the all the comparator functionality that we want I think it's totally fine if it isn't the most performant right now (and we've removed timings anyway). If there are improvements to be made to libcudf, we should open a separate issue for that.

vyasr (approved):

cpp/examples/nested_types/deduplication.cpp
divyegala: I don't get it, what do you mean by "at least have some"?

bdice: 1. We need to file an issue that makes this use hash-based aggregations without any "forcing"
2. I think Vyas is asking, if you don't add this column to "force" hash aggregations, are _some_ of the aggregations hash-based and others sort-based? My understanding of libcudf's behavior is that if any aggregation is sort-based, all the aggregations fall back to using sort-based implementations. Is that true?

divyegala: Yes, that's right. If an aggregation only has a sort-based implementation then it will fall back to a sort aggregation.

In this particular case though, a hash-based aggregation is available and it needs to be "forced" due to a limitation.

Here's the issue: https://github.com/rapidsai/cudf/issues/14412

cpp/examples/nested_types/deduplication.cpp
divyegala: https://github.com/rapidsai/cudf/issues/14413

cpp/examples/nested_types/deduplication.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+  auto count = count_aggregate(input->view());
+
+  auto combined = join_count(input->view(), count->view());
+
+  auto sorted = sort_keys(combined->view());
+
+  metadata.schema_info.emplace_back("count");
+
+  write_json(sorted->view(), metadata, output_filepath);
```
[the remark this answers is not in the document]
divyegala: So `std::cout` or `CUDF_LOG_INFO`?

cpp/examples/nested_types/deduplication.cpp
vyasr: @bdice are you OK with marking this thread as resolved for now? The issue looks like it has a reasonable plan for how to move forward, but I don't think we should block this PR on getting that enabled.

cpp/examples/nested_types/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+  VERSION 0.0.1
+  LANGUAGES CXX CUDA
+)
+
+include(../fetch_dependencies.cmake)
+
+# Configure your project here
+add_executable(deduplication deduplication.cpp)
+target_link_libraries(deduplication PRIVATE cudf::cudf)
+target_compile_features(deduplication PRIVATE cxx_std_17)
```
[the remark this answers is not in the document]
vyasr: `target_compile_features` is the newer option. As for why to prefer it, in this case I don't think it matters too much although there are subtle differences in how it interacts with other CMake variables like `CMAKE_CXX_STANDARD`. It would be more helpful if you wanted it to not be `PRIVATE` but instead `PUBLIC`/`INTERFACE` because compile features will propagate to consuming targets while properties won't.

cpp/examples/nested_types/deduplication.cpp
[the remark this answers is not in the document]
bdice: Your call. I vote `std::cout` but either is fine.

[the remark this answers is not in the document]
vyasr: I think that since none of the other examples currently do this we should skip for now and come back to add in a follow-up if we want. I do think some sort of logging would be nice, but we should take the opportunity to establish a consistent standard of some sort while we still have a small number of examples and it's easy to make them consistent.

cpp/examples/nested_types/CMakeLists.txt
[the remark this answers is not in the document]
robertmaynard: The `CXX_STANDARD` property is a request for a compiler level, but can decay to the highest supported level. You need to also set `CXX_STANDARD_REQUIRED` to stop the decaying.

`target_compile_features(deduplication PRIVATE cxx_std_17)` is a hard requirement and errors out instead of decaying.

cpp/examples/nested_types/deduplication.cpp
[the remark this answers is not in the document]
vyasr: Ah looks like Divye added some. All good then.

cpp/examples/nested_types/CMakeLists.txt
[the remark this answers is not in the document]
bdice: Thanks, both of you.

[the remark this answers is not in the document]
bdice: I struggle with knowing what "the newer/better option" is most of the time. 😕

bdice (approved):

karthikeyann (approved):
LGTM 👍

PointKernel (approved):

ttnghia (approved):

raydouglass (approved):

PR #13796, opened 2023-08-01
Add minhash support for MurmurHash3_x64_128 

davidwendt:
## Description
Adds `nvtext::minhash64` to libcudf and the Cython/Python changes to call it.
The `MurmurHash3_x64_128` is called and only the first `uint64` value is used.

The libcudf API was changed to remove the `hash_id` parameter since it was incompatible with the seed types.

## Checklist


cpp/benchmarks/text/minhash.cpp
```diff
@@ -30,6 +30,7 @@ static void bench_minhash(nvbench::state& state)
   auto const row_width  = static_cast<cudf::size_type>(state.get_int64("row_width"));
   auto const hash_width = static_cast<cudf::size_type>(state.get_int64("hash_width"));
   auto const seed_count = static_cast<cudf::size_type>(state.get_int64("seed_count"));
+  auto const b64        = state.get_int64("htype") == 64;
```
bdice, line 33: Can we rename `b64` (base 64?) and `htype` (hash type)? We have pretty readable names for all the other variables in this file.
```suggestion
  auto const b64        = state.get_int64("htype") == 64;
```

cpp/include/nvtext/minhash.hpp
```diff
@@ -72,20 +73,75 @@ std::unique_ptr<cudf::column> minhash(
...
+ *
+ * Hash values are computed from substrings of each string and the
+ * minimum hash value is returned for each string.
+ *
+ * Any null row entries result in corresponding null output rows.
+ *
+ * This function uses MurmurHash3_x64_128 for the hash algorithm.
+ *
+ * @throw std::invalid_argument if the width < 2
+ * @throw std::invalid_argument if hash_function is not HASH_MURMUR3
```
bdice, line 100:
```suggestion
```

cpp/include/nvtext/minhash.hpp
```diff
@@ -72,20 +73,75 @@ std::unique_ptr<cudf::column> minhash(
...
+
+/**
+ * @brief Returns the minhash value for each string
+ *
+ * Hash values are computed from substrings of each string and the
+ * minimum hash value is returned for each string.
+ *
+ * Any null row entries result in corresponding null output rows.
+ *
+ * This function uses MurmurHash3_x64_128 for the hash algorithm.
```
bdice, line 97: Should this part of the docstring clarify that the beginning/end of the 128 bit hash value is dropped, leaving a 64 bit result?

cpp/src/text/minhash.cu
```diff
@@ -51,79 +50,98 @@ namespace {
...
+  if (lane_idx == 0) {
+    auto const init = d_str.empty() ? 0 : std::numeric_limits<hash_value_type>::max();
+    thrust::fill(thrust::seq, d_output, d_output + seeds.size(), init);
+  }
+  __syncwarp();
+
+  auto const begin = d_str.data() + lane_idx;
+  auto const end   = d_str.data() + d_str.size_bytes();
+
+  // each lane hashes 'width'  substrings of d_str
```
bdice, line 95:
```suggestion
  // each lane hashes 'width' substrings of d_str
```

cpp/src/text/minhash.cu
```diff
@@ -51,79 +50,98 @@ namespace {
...
+      auto const hasher = HashFunction(seeds[seed_idx]);
+      // hash substring and store the min value
+      if constexpr (std::is_same_v<hash_value_type, uint32_t>) {
         auto const hvalue = hasher(hash_str);
-        cuda::atomic_ref<cudf::hash_value_type, cuda::thread_scope_block> ref{
-          *(d_output + seed_idx)};
+        cuda::atomic_ref<hash_value_type, cuda::thread_scope_block> ref{*(d_output + seed_idx)};
+        ref.fetch_min(hvalue, cuda::std::memory_order_relaxed);
+      } else {
+        auto const hvalue = thrust::get<0>(hasher(hash_str));  // just use the first uint64_t
```
bdice, line 114: Can we add comments (or statically assert?) that this assumes `HashFunction` is `MurmurHash_x64_128`? The conditional checks the `hash_value_type` but that assumes behavior that wouldn't apply if we used XXHash (also comes in a 64 bit flavor but doesn't return a two-part value like MurmurHash_x64_128's 128 bit result).

cpp/src/text/minhash.cu
```diff
@@ -51,79 +50,98 @@ namespace {
...
-  CUDF_EXPECTS(hash_function == cudf::hash_id::HASH_MURMUR3,
-               "Only murmur3 hash algorithm supported",
-               std::invalid_argument);
   CUDF_EXPECTS((static_cast<std::size_t>(input.size()) * seeds.size()) <
                  static_cast<std::size_t>(std::numeric_limits<cudf::size_type>::max()),
                "The number of seeds times the number of input rows exceeds the column size limit",
                std::overflow_error);
 
-  auto output_type = cudf::data_type{cudf::type_to_id<cudf::hash_value_type>()};
+  auto output_type = cudf::data_type{cudf::type_to_id<hash_value_type>()};
```
bdice, line 144:
```suggestion
  auto const output_type = cudf::data_type{cudf::type_to_id<hash_value_type>()};
```

cpp/src/text/minhash.cu
davidwendt: You would get a compile error either way. Adding another hash algorithm would be a big change so I don't think it would be accidentally dropped in here. Maybe I'm missing what you are trying to protect against.

bdice: I agree an error would occur either way. I just want to make it explicit that this 64 bit minhash function is assuming the result layout of the 128 bit hash function.

davidwendt: Would a comment suffice or are you looking for something else?

bdice: Yup! Comments are fine.

cpp/src/text/minhash.cu
```diff
@@ -51,79 +50,100 @@ namespace {
...
     }
   }
-};
+}
 
 }  // namespace
 
+template <typename HashFunction,
+          typename hash_value_type =
+            std::conditional_t<std::is_same_v<decltype(HashFunction::result_type), uint32_t>,
```
bdice, line 128: Same question as above.
```suggestion
            std::conditional_t<std::is_same_v<typename HashFunction::result_type, uint32_t>,
```

cpp/src/text/minhash.cu
```diff
@@ -174,24 +193,45 @@ std::unique_ptr<cudf::column> minhash(cudf::strings_column_view const& input,
 }  // namespace detail
 
 std::unique_ptr<cudf::column> minhash(cudf::strings_column_view const& input,
-                                      cudf::numeric_scalar<cudf::hash_value_type> seed,
+                                      cudf::numeric_scalar<uint32_t> seed,
                                       cudf::size_type width,
-                                      cudf::hash_id hash_function,
                                       rmm::mr::device_memory_resource* mr)
```
bdice, line 198: Is this API-breaking change a good time to introduce public streams, too? (Totally fine if the answer is no / not yet.)

cpp/src/text/minhash.cu
```diff
@@ -133,13 +153,12 @@ std::unique_ptr<cudf::column> minhash(cudf::strings_column_view const& input,
...
-    rmm::exec_policy(stream),
-    thrust::counting_iterator(std::size_t{0}),
-    static_cast<std::size_t>(input.size()) * static_cast<std::size_t>(cudf::detail::warp_size),
-    minhash_fn{*d_strings, seeds, width, d_hashes});
+  constexpr int block_size = 256;
+  cudf::detail::grid_1d grid{input.size() * cudf::detail::warp_size, block_size};
+  minhash_fn<HashFunction><<<grid.num_blocks, grid.num_threads_per_block, 0, stream.value()>>>(
+    *d_strings, seeds, width, d_hashes);
 
   if (seeds.size() == 1) {
```
bdice, line 163: I'm not sure how I feel about this behavior. From the "list of seeds" overload, I expect that a list of seeds with length 1 should give back a list column of hash values with length 1 -- not a "flattened" column. The documentation mentions this is expected behavior, but flattening this on behalf of the user is not what my intuition would expect when we provide both a scalar and a list overload. If we have precedent in libcudf for doing automatic flattening (especially when scalar/list overloads exist), I'd like to know. It's fine with me to defer this to follow-up work (by filing an issue, perhaps).

python/cudf/cudf/core/column/string.py
```diff
@@ -5322,18 +5322,37 @@ def minhash(
...
-            seeds_column = column.as_column(0, dtype=np.uint32, length=1)
-        elif isinstance(seeds, cudf.Series) and seeds.dtype == np.uint32:
-            seeds_column = seeds._column
-        else:
-            raise ValueError(
-                f"Expecting a Series with dtype uint32, got {type(seeds)}"
+        if method == "MurmurHash3_x86_32":
+            if seeds is None:
+                seeds_column = column.as_column(0, dtype=np.uint32, length=1)
+            elif isinstance(seeds, cudf.Series) and seeds.dtype == np.uint32:
```
bdice, line 5334: It's a bit awkward to force users to supply a `Series` here. I think normally we would allow any `ColumnLike` object, such as a list or a NumPy array. That makes it easier to call this API. Maybe we can call `as_column` on the user input?

python/cudf/cudf/core/column/string.py
```diff
@@ -5322,18 +5322,37 @@ def minhash(
...
+        dtype: list
         """
-        if seeds is None:
-            seeds_column = column.as_column(0, dtype=np.uint32, length=1)
-        elif isinstance(seeds, cudf.Series) and seeds.dtype == np.uint32:
-            seeds_column = seeds._column
-        else:
-            raise ValueError(
-                f"Expecting a Series with dtype uint32, got {type(seeds)}"
+        if method == "MurmurHash3_x86_32":
```
bdice, line 5331: Can we rewrite this as a single path for both hash functions, and predefine the type variables (and other parts that change) like this?
```python
seed_type = np.uint32 if method == "MurmurHash3_x86_32" else np.uint64

...

if seeds is None:
    seeds_column = column.as_column(0, dtype=seed_type, length=1)
```

bdice (commented):
A few questions / suggestions. I think this will be good after this round of feedback!

cpp/src/text/minhash.cu
davidwendt: Changed this to always return lists column even if that returns only a single element per row.

python/cudf/cudf/core/column/string.py
davidwendt: Split the code into `minhash()` and `minhash64()` as discussed offline.

python/cudf/cudf/core/column/string.py
```diff
@@ -5287,35 +5287,29 @@ def edit_distance_matrix(self) -> SeriesOrIndex:
...
         --------
         >>> import cudf
         >>> str_series = cudf.Series(['this is my', 'favorite book'])
         >>> seeds = cudf.Series([0], dtype=np.uint32)
         >>> str_series.str.minhash(seeds)
-        0     21141582
-        1    962346254
+        0     [21141582]
+        1    [962346254]
         dtype: uint32
```
bdice, line 5313: This dtype doesn't match the list data. The example for minhash64 shows this:
```suggestion
        dtype: list
```

python/cudf/cudf/core/column/string.py
```diff
@@ -5325,14 +5319,54 @@ def minhash(
...
+        width : int
+            The width of the substring to hash.
+            Default is 4 characters.
+
+        Examples
+        --------
+        >>> import cudf
+        >>> str_series = cudf.Series(['this is my', 'favorite book'])
+        >>> seeds = cudf.Series([0, 1, 2], dtype=np.uint64)
+        >>> str_series.str.minhash(seeds, method='MurmurHash3_x64_128')
```
bdice, line 5355:
```suggestion
        >>> str_series.str.minhash64(seeds)
```

python/cudf/cudf/core/column/string.py
```diff
@@ -5287,35 +5287,29 @@ def edit_distance_matrix(self) -> SeriesOrIndex:
...
+        self, seeds: Optional[ColumnLike] = None, width: int = 4
     ) -> SeriesOrIndex:
         """
         Compute the minhash of a strings column.
+        This uses the MurmurHash3_x86_32 algorithm for the hash function.
 
         Parameters
         ----------
-        seeds : Series
+        seeds : ColumnLike
```
bdice, line 5298: Is there a reason we're not including the scalar seed API in the Python interface?

bdice (approved):

cpp/include/nvtext/minhash.hpp
```diff
@@ -36,24 +36,24 @@ namespace nvtext {
...
  * @param mr Device memory resource used to allocate the returned column's device memory
  * @return Minhash values for each string in input
  */
 std::unique_ptr<cudf::column> minhash(
   cudf::strings_column_view const& input,
-  cudf::numeric_scalar<cudf::hash_value_type> seed = cudf::numeric_scalar(cudf::DEFAULT_HASH_SEED),
-  cudf::size_type width                            = 4,
-  cudf::hash_id hash_function                      = cudf::hash_id::HASH_MURMUR3,
-  rmm::mr::device_memory_resource* mr              = rmm::mr::get_current_device_resource());
+  cudf::numeric_scalar<uint32_t> seed = 0,
```
vyasr, line 53: Why switch to hardcoding the value here instead of using the constant?

cpp/include/nvtext/minhash.hpp
```diff
@@ -64,28 +64,88 @@ std::unique_ptr<cudf::column> minhash(
...
+ * @param seed  Seed value used for the hash algorithm
+ * @param width The character width used for apply substrings;
+ *              Default is 4 characters.
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ * @return Minhash values as UINT64 for each string in input
+ */
+std::unique_ptr<cudf::column> minhash64(
+  cudf::strings_column_view const& input,
+  cudf::numeric_scalar<uint64_t> seed = 0,
```
vyasr, line 114: Same question as before, why the different seed choice? Is it because of the potential for unsafe casts depending on the type of `DEFAULT_HASH_SEED` (currently OK since it's `uint32_t`)?

cpp/src/text/minhash.cu
```diff
@@ -51,79 +50,96 @@ namespace {
...
+  typename hash_value_type = std::
+    conditional_t<std::is_same_v<typename HashFunction::result_type, uint32_t>, uint32_t, uint64_t>>
+__global__ void minhash_kernel(cudf::column_device_view const d_strings,
+                               cudf::device_span<hash_value_type const> seeds,
+                               cudf::size_type width,
+                               hash_value_type* d_hashes)
+{
+  auto const idx = static_cast<std::size_t>(threadIdx.x + blockIdx.x * blockDim.x);
+  if (idx >= (static_cast<std::size_t>(d_strings.size() *
+                                       static_cast<std::size_t>(cudf::detail::warp_size)))) {
```
vyasr, line 72: This seems like one cast too many. Isn't at least one of them redundant?

cpp/src/text/minhash.cu
```diff
@@ -51,79 +50,96 @@ namespace {
...
+    return;
+  }
+
+  auto const str_idx  = static_cast<cudf::size_type>(idx / cudf::detail::warp_size);
```
vyasr, line 76: In light of the recent discussions about #10368, do we need to be worried about this ever overflowing? Specifically since it's coming from `idx`, which is a `std::size_t`?

cpp/src/text/minhash.cu
```diff
@@ -133,27 +149,26 @@ std::unique_ptr<cudf::column> minhash(cudf::strings_column_view const& input,
...
+{
   // build the offsets for the output lists column
-  auto offsets = cudf::detail::sequence(
-    input.size() + 1,
-    cudf::numeric_scalar<cudf::size_type>(0),
-    cudf::numeric_scalar<cudf::size_type>(static_cast<cudf::size_type>(seeds.size())),
-    stream,
-    mr);
+  auto const zero = cudf::numeric_scalar<cudf::size_type>(0);
+  auto const size = cudf::numeric_scalar<cudf::size_type>(static_cast<cudf::size_type>(seeds_size));
```
vyasr, line 170: I'm realizing now that most of the code is not changed from what it looked like before this PR, you just converted thrust+functor into a kernel, so no need to change but just leaving another note: isn't this also an unnecessary cast?

vyasr (approved):
Couple of questions, but looks good.

cpp/include/nvtext/minhash.hpp
davidwendt: Because the type may be different. I'd rather be clear that the default seed is actually 0 and would not want to change that if the rest of libcudf decided on a different default. Hopefully that is ok.

cpp/include/nvtext/minhash.hpp
davidwendt: Yes, the type is different here and I'd rather the 2 functions be consistent over any need to use the constant def.

cpp/src/text/minhash.cu
davidwendt: Argh, my parentheses are not right here.

cpp/src/text/minhash.cu
davidwendt: Good question. No, the `str_idx` should always be within `size_type` after this division otherwise we would have other issues. This is consistent with the recent discussions and changes as well.

cpp/src/text/minhash.cu
davidwendt: You are correct. This mistake was a side-effect of refactoring and copying code around. The `seed_size` had previously come from a device-span which has sizes of type `size_t` and required casting here. That is done before calling this code so the extra cast is definitely not needed.

PR #13801, opened 2023-08-02
Refactoring of Buffers (last step towards unifying COW and Spilling) 

madsbk:
This PR de-couples buffer slices/views from owning buffers. As it is now, all buffer classes (`ExposureTrackedBuffer`, `BufferSlice`, `SpillableBuffer`, `SpillableBufferSlice`) inherent from `Buffer`, however they are not Liskov substitutable as pointed by @wence- and @vyasr ([here](https://github.com/rapidsai/cudf/pull/13307#discussion_r1189748484) and [here](https://github.com/rapidsai/cudf/pull/13307#discussion_r1239014158)). 

To fix this, we now have a `Buffer` and a `BufferOwner` class. We still use the `Buffer` throughout cuDF but it now points to an `BufferOwner`.  

We have the following class hierarchy: 
``` 
ExposureTrackedBufferOwner -> BufferOwner 
SpillableBufferOwner -> BufferOwner 
ExposureTrackedBuffer -> Buffer 
SpillableBuffer -> Buffer 
``` 

With the following relationship: 
``` 
Buffer -> BufferOwner 
ExposureTrackedBuffer -> ExposureTrackedBufferOwner 
SpillableBuffer -> SpillableBufferOwner 
``` 

#### Unify COW and Spilling 

In a follow-up PR, the spilling buffer classes will inherent from the exposure tracked buffer classes so we get the following hierarchy: 
``` 
SpillableBufferOwner -> ExposureTrackedBufferOwner -> BufferOwner 
SpillableBuffer -> ExposureTrackedBuffer -> Buffer 
```


python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -90,37 +114,42 @@ def cuda_array_interface_wrapper(
...
         No data is being copied.
 
         Parameters
         ----------
         data : device-buffer-like
             An object implementing the CUDA Array Interface.
+        exposed : bool
+            Mark the buffer as permanently exposed. This is used by
+            ExposureTrackedBuffer to determine when a deep copy is required
+            and by SpillableBuffer to mark the buffer unspillable.
```
wence-, line 147: So this is ignored in this class and only makes sense in the exposuretrackedbuffer object, right?

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
```diff
@@ -139,103 +41,50 @@ def mark_exposed(self) -> None:
...
-
-        Returns
-        -------
-        ExposureTrackedBuffer
-            Buffer representing the same device memory as `data`
-        """
-        ret = super()._from_device_memory(data)
+    def _from_device_memory(cls, data: Any, exposed: bool) -> Self:
+        ret = super()._from_device_memory(data, exposed=exposed)
         ret._exposed = exposed
```
wence-, line 46: **issue:** We pass `exposed` to the superclass constructor but are none-the-less on the hook to remember it on the returned object. This feels wrong to me. Either the superclass doesn't know about exposed at all (in which case `_from_device_memory` can't be a shared thing with differing signatures). Or else we should store `exposed` in the superclass and implement the `exposed` property and `mark_exposed` with overrides here.

WDYT?

As it is, I think an `ExposureTrackedBufferOwner` can sit in for any `BufferOwner`, but not vice-versa (perhaps that is intended).

In some senses to be "generic", the return value of `_from_device_memory` is only ever allowed to claim that it supports the interface that `BufferOwner` provides (not that of a subclass)

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -90,37 +114,42 @@ def cuda_array_interface_wrapper(
...
 
     _ptr: int
     _size: int
     _owner: object
 
-    def __init__(self):
-        raise ValueError(
-            f"do not create a {self.__class__} directly, please "
-            "use the factory function `cudf.core.buffer.as_buffer`"
-        )
```
wence-, line 107: **question:** We've now lost this "don't call init" path. Is that deliberate?

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
```diff
@@ -139,103 +41,50 @@ def mark_exposed(self) -> None:
...
         ret._slices = weakref.WeakSet()
         return ret
 
-    def _getitem(self, offset: int, size: int) -> BufferSlice:
-        return BufferSlice(base=self, offset=offset, size=size)
-
     @property
     def __cuda_array_interface__(self) -> Mapping:
         self.mark_exposed()
```
wence-, line 52: If a `BufferOwner` offered `exposed` and `mark_exposed` (that were respectively `False` and a no-op) then there wouldn't be the need for a separate implementation of `__cuda_array_interface__` here, and I think the only "special" code would have to be `_from_device_memory` and the implemenation of `mark_exposed`.

Like this?

```python
class Owner:

    mem: DeviceBuffer
    exposed: bool
    _slices: weakref.WeakSet

    def mark_exposed(self):
        pass

    @classmethod
    def _from_device_memory(cls, mem, exposed):
        self = cls.__new__(cls)
        self.mem = mem
        self.exposed = False
        return self

        
    @property
    def __cuda_array_interface__(self):
        self.mark_exposed()
        return ...


class ExposedTracker(Owner):
    def mark_exposed(self):
        self.exposed = True

    @classmethod
    def _from_device_memory(cls, mem, exposed):
        self = super()._from_device_memory(mem, exposed)
        self.exposed = exposed
        self._slices = weakref.WeakSet()
        return self
```

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
```diff
@@ -139,103 +41,50 @@ def mark_exposed(self) -> None:
...
-        base._slices.add(self)
-
-    def __getattr__(self, name):
-        if name in self._passthrough_attributes:
-            return getattr(self._base, name)
-        raise AttributeError(
-            f"{self.__class__.__name__} object has no attribute {name}"
-        )
+        super().__init__(owner=owner, offset=offset, size=size)
+        self._owner._slices.add(self)
```
wence-, line 78: So this asserts that `owner` is an `ExposureTrackedBuffer`.

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -30,7 +43,18 @@ def as_buffer(
 
     If `data` is an integer, it is assumed to point to device memory.
 
-    Raises ValueError if data isn't C-contiguous.
+    Raises ValueError if `data` isn't C-contiguous.
+
+    If copy-on-write is enabled, a ExposureTrackedBuffer that refers to a
+    ExposureTrackedBufferOwner is returned.
+
+    If spilling is enabled, a SpillableBuffer that refers ot a
```
wence-, line 51:
```suggestion
    If spilling is enabled, a SpillableBuffer that refers to a
```

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -30,7 +43,18 @@ def as_buffer(
...
+    SpillableBufferOwner is returned. If `data` is owned by a spillable buffer,
+    it must either be "exposed" or spilled locked (called within an
```
wence-, line 53:
```suggestion
    it must either be "exposed" or spill locked (called within an
```

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -73,13 +97,49 @@ def as_buffer(
...
+        buffer_class = SpillableBuffer
+    elif get_option("copy_on_write"):
+        owner_class = ExposureTrackedBufferOwner
+        buffer_class = ExposureTrackedBuffer
+    else:
+        owner_class = BufferOwner
+        buffer_class = Buffer
+
+    # Handle host memory,
+    if not hasattr(data, "__cuda_array_interface__"):
```
wence-, line 114: **question:** If `data` is an `ExposureTrackedBuffer` or `SpillableBuffer`, this line will ensure that it is marked as permanently exposed. Is that what you want?

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -73,13 +97,49 @@ def as_buffer(
...
+    # should be the same class as specified by the current config (see above)
+    assert owner.__class__ is owner_class
+    if (
+        isinstance(owner, SpillableBufferOwner)
+        and not owner.exposed
+        and get_spill_lock() is None
+    ):
+        raise ValueError(
+            "A owning spillable buffer must "
+            "either be exposed or spilled locked."
```
wence-, line 136:
```suggestion
            "An owning spillable buffer must "
            "either be exposed or spill locked."
```

wence- (changes requested):
Mostly requesting changes for discussion points

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -277,20 +259,147 @@ def memoryview(
         )
         return memoryview(host_buf).toreadonly()
 
+    def __repr__(self) -> str:
+        return (
+            f"<{self.__class__.__name__} size={format_bytes(self._size)} "
+            f"ptr={hex(self._ptr)} owner={repr(self._owner)}>"
```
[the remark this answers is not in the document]
madsbk: I think this is a bit too verbose, I actually prefer `hex()`.

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -90,37 +114,42 @@ def cuda_array_interface_wrapper(
...
+            ExposureTrackedBuffer to determine when a deep copy is required
+            and by SpillableBuffer to mark the buffer unspillable.
 
         Returns
         -------
-        Buffer
-            Buffer representing the same device memory as `data`
+        BufferOwner
+            BufferOwner wrapping `data`
         """
```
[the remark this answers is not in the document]
madsbk: > When can `ret.size < 0` be true?

Only by mistake

python/cudf/cudf/core/buffer/buffer.py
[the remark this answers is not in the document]
madsbk: but no strong opinion, if you guys prefer  `0x0000_0000_ABCD_1234`, it is fine with me

python/cudf/cudf/core/buffer/buffer.py
madsbk: yes

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -277,20 +259,147 @@ def memoryview(
...
+        """Size of the buffer in bytes."""
+        return self._size
+
+    @property
+    def nbytes(self) -> int:
+        """Size of the buffer in bytes."""
+        return self._size
+
+    @property
+    def owner(self) -> Any:
```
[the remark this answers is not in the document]
madsbk: fixed

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -277,20 +259,147 @@ def memoryview(
...
+            return self.__class__(
+                owner=self._owner, offset=self._offset, size=self._size
+            )
+
+        # Otherwise, we create a new copy of the memory
+        owner = self._owner._from_device_memory(
+            rmm.DeviceBuffer(
+                ptr=self._owner.get_ptr(mode="read") + self._offset,
+                size=self.size,
+            ),
```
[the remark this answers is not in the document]
madsbk: yes, and to handle the offset

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
madsbk: good idea, done

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
madsbk: Yes, this is enforced by `as_buffer()`

python/cudf/cudf/core/buffer/utils.py
madsbk: Notice, we have the guard above:
```python
    if isinstance(data, Buffer):
        return data
```

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -277,20 +259,147 @@ def memoryview(
...
+            exposed=False,
+        )
+        return self.__class__(owner=owner, offset=0, size=owner.size)
+
+    @property
+    def __cuda_array_interface__(self) -> Mapping:
+        """Implementation of the CUDA Array Interface."""
+        return {
+            "data": (self.get_ptr(mode="write"), False),
```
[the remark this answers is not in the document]
madsbk: > Will we potentially want the concept of read-only buffer slices

I don't think so

> I don't think it will actually change the mode argument we pass to get_ptr because that's more to do with "do we have control over the lifetime of the usage".

Agree

python/cudf/cudf/core/buffer/utils.py
wence-: Ah, missed that in the diff, thanks.

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -17,6 +17,30 @@
...
+    Return
+    ------
+    BufferOwner or None
+        The owner of `data` if found otherwise None.
+    """
+
+    if isinstance(data, BufferOwner):
+        return data
+    if hasattr(data, "owner"):
+        return get_buffer_owner(data.owner)
```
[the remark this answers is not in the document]
vyasr: The public/private interface of objects at the Column/Buffer layer of cudf Python is pretty rough. Personally I've been punting on sorting out such questions until pylibcudf is ready since we'll have to rearchitect a lot of the internals around pylibcudf at that time anyway.

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
vyasr: I'm somewhat second-guessing the benefit of even separating exposure tracking from the parent BufferOwner class. If the parent needs to define the exposed property already for Liskov substitutability, then perhaps the two classes should just be merged (I made the same point above but reiterating in this thread). What if we did that, and it was only the usage of non-owning ExposureTrackedBuffer vs the Buffer that differentiated COW mode from non-COW mode?

vyasr (commented):
First pass was pretty high-level to reacquaint myself with the concepts here. My main open question is around the separation between ExposureTrackedBufferOwner and BufferOwner and whether they should maybe be unified if we have to maintain the exposure properties at the BufferOwner level for the two to satisfy the LSP. Maybe we only want to distinguish at the Buffer, not BufferOwner, level (I think the SpillableBufferOwner should stay separate though).

madsbk:
> First pass was pretty high-level to reacquaint myself with the concepts here. My main open question is around the separation between ExposureTrackedBufferOwner and BufferOwner and whether they should maybe be unified if we have to maintain the exposure properties at the BufferOwner level for the two to satisfy the LSP. Maybe we only want to distinguish at the Buffer, not BufferOwner, level (I think the SpillableBufferOwner should stay separate though).

Agree, I think merging `ExposureTrackedBufferOwner` and `BufferOwner` is a good idea. 
@wence-, I think you had a similar idea?

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -17,6 +17,30 @@
...
+def get_buffer_owner(data) -> Optional[BufferOwner]:
+    """Get the owner of `data`, if one exists
+
+    Search through the stack of data owners in order to find an
+    owner BufferOwner (incl. subclasses).
+
+    Parameters
+    ----------
+    data
+        The data object
```
shwina, line 29: Would be good to clarify what types are acceptable for `data` (ideally with typing, otherwise here)

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -277,20 +288,151 @@ def memoryview(
...
         return memoryview(host_buf).toreadonly()
 
+    def __str__(self) -> str:
+        return (
+            f"<{self.__class__.__name__} size={format_bytes(self._size)} "
+            f"ptr={hex(self._ptr)} owner={self._owner!r}>"
+        )
+
+
+class Buffer(Serializable):
```
shwina, line 298: Suggesting a hopefully non-controversial color for the bike shed:

`BufferOwner` -> `Buffer`
`Buffer` -> `BufferView`

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -241,6 +232,26 @@ def __cuda_array_interface__(self) -> Mapping:
...
+    @property
+    def exposed(self) -> bool:
+        """The current exposure status of the buffer
+
+        This is used by ExposureTrackedBuffer to determine when a deep copy
+        is required and by SpillableBuffer to mark the buffer unspillable.
+        """
+        return self._exposed
+
+    def mark_exposed(self) -> None:
```
shwina, line 244: Any reason not to use `exposed.setter` here instead?

python/cudf/cudf/core/buffer/buffer.py
madsbk: No strong opinion, the important thing is to make it clear that  `BufferOwner` is never to be used directly. 
 `Buffer` (current name) is the one that we should use throughout cudf.

python/cudf/cudf/core/buffer/buffer.py
madsbk: To prevent `.exposed = False`, the expose mark is permanent.

python/cudf/cudf/core/buffer/buffer.py
madsbk: Added `data: Any` and some more doc

python/cudf/cudf/core/buffer/buffer.py
madsbk: Fixed

python/cudf/cudf/core/buffer/buffer.py
madsbk: `ExposureTrackedBufferOwner` has been merged with `BufferOwner`

python/cudf/cudf/core/buffer/exposure_tracked_buffer.py
madsbk: `ExposureTrackedBufferOwner` has been merged with `BufferOwner`

python/cudf/cudf/core/abc.py
```diff
@@ -89,7 +89,14 @@ def device_serialize(self):
         """
         header, frames = self.serialize()
         assert all(
-            isinstance(f, (cudf.core.buffer.Buffer, memoryview))
+            isinstance(
+                f,
+                (
+                    cudf.core.buffer.BufferOwner,
+                    cudf.core.buffer.Buffer,
```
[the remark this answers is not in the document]
madsbk: Good point, we use `cudf.core.buffer.Buffer` exclusively now.

python/cudf/cudf/core/buffer/buffer.py
shwina: OK - that makes sense

python/cudf/cudf/core/buffer/spillable_buffer.py
```diff
@@ -465,148 +360,114 @@ def memoryview(
...
+        also mean that `self` becomes exposed permanently if the frames are
+        later accessed through `__cuda_array_interface__`, which is exactly
+        what libraries like Dask+UCX would do when communicating!
+
+        The sound solution is to modify Dask et al. so that they access the
+        frames through `.get_ptr()` and holds on to the `spill_lock` until
+        the frame has been transferred. However, until this adaptation we
+        use a hack where the frame is a `Buffer` with a `spill_lock` as the
+        owner, which makes `self` unspillable while the frame is alive but
+        doesn't expose `self` when `__cuda_array_interface__` is accessed.
```
vyasr, line 432: We should reassess this evaluation after the next PR unifying COW and spilling. I don't think I agree with this statement, it implies leaking knowledge of cudf buffer internals to dask. Once we've finished the unification we should revisit whether there's a more API-friendly way of doing this. If not, we need to think about the appropriate generalization of our exposure semantics to generic CAI usage.

python/cudf/cudf/core/buffer/utils.py
```diff
@@ -30,7 +40,17 @@ def as_buffer(
 
     If `data` is an integer, it is assumed to point to device memory.
 
-    Raises ValueError if data isn't C-contiguous.
+    Raises ValueError if `data` isn't C-contiguous.
+
+    If copy-on-write is enabled, a ExposureTrackedBuffer is returned.
```
vyasr, line 45:
```suggestion
    If copy-on-write is enabled, an ExposureTrackedBuffer is returned.
```

vyasr (changes requested):
This is very close! Thanks for all the iterations.

python/cudf/cudf/core/buffer/buffer.py
wence-: Since we are going to go round again in a final step of unification, let us punt a repeated renaming to there. But agreed, in that change we should have:

- `Buffer` (the "owning" object that contains the concrete devicebuffer/cupy array)
- `BufferView` (the "user-visible" object)

python/cudf/cudf/core/buffer/spillable_buffer.py
wence-: FWIW, this approach (as the code in this PR uses) is (morally) the way pytorch ensures lifetime of a `CAI`-supporting object being turned into a torch tensor

wence- (approved):
I think I am happy with this! :tada:

python/cudf/cudf/core/buffer/buffer.py
vyasr: We discussed offline and are in agreement that the renaming can happen in a follow-up.

python/cudf/cudf/core/buffer/buffer.py
```diff
@@ -17,6 +18,30 @@
 from cudf.utils.string import format_bytes
 
 
+def get_buffer_owner(data: Any) -> Optional[BufferOwner]:
```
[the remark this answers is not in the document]
vyasr: Recursive function works for me.

vyasr (approved):

madsbk:
Thanks for the reviews. All tests marked `spilling` passes for me locally so let's merge this PR
```
CUDF_SPILL=on CUDF_SPILL_DEVICE_LIMIT=1 py.test -m spilling python/cudf/cudf/tests/
```

PR #13803, opened 2023-08-02
Long string optimization for string column parsing in JSON reader

### LINKED ISSUE #13724: [FEA] Improve JSON reader performance with "document" strings columns
In the "Common Crawl" document dataset, the size of "documents" (i.e. large strings) varies from 200 characters to 1M characters and typically each file contains a few thousand "documents". In libcudf JSON reading, the `get_token_stream` step is data-parallel and efficiently processes the strings data of these documents. However in the `device_json_column_to_cudf_column` and `parse_data` steps, we observe warp divergence and longer runtimes for variable document sizes. Depending on the row count, this divergence results in 20% to 6x slower end-to-end read time for variable length strings compared to constant length strings.

Here is a code snippet to demonstrate the impact of variable document sizes.
```
import time
from io import BytesIO

import cudf
import cupy
import nvtx

cupy.random.seed(0)

num_rows = 6000
df = cudf.DataFrame({'a': ['a']*num_rows})
df['b'] = df['a'].str.repeat(cupy.random.lognormal(7.6, 1.2, num_rows).astype(int))
average_length = df['b'].str.len().sum() // num_rows
variable_lengths_buffer = BytesIO()
df.to_json(variable_lengths_buffer, engine='cudf', lines=True)

df = cudf.DataFrame({'a': ['a']*num_rows})
df['b'] = df['a'].str.repeat(average_length)
constant_lengths_buffer = BytesIO()
df.to_json(constant_lengths_buffer, engine='cudf', lines=True)

with nvtx.annotate('variable lengths', color="yellow"):
    t0 = time.time()
    df = cudf.read_json(variable_lengths_buffer, lines=True)
    t1 = time.time()
print('variable lengths', t1-t0)

with nvtx.annotate('constant lengths', color="yellow"):
    t0 = time.time()
    df = cudf.read_json(constant_lengths_buffer, lines=True)
    t1 = time.time()
print('constant lengths', t1-t0)
```

```
variable lengths 0.12197709083557129
constant lengths 0.02486705780029297
```

The end-to-end impact of the warp divergence seems to be reduced when reading large numbers of documents (100K+), which explains why multi-source reads of document data have shown better performance.

Here's a profile of 100K `num_rows`, which shows the difference in runtime and warp occupancy in `parse_data`. Also this higher number of rows shows less difference in end-to-end runtime.
image

Here's a profile of 6K `num_rows`, which shows a 6x increase in end-to-end reading time for variable length strings versus constant length strings.
image

Also, here is a snapshot of "Common Crawl" document lengths and the `lognormal(7.6, 1.2)` distribution used to model it.
image

karthikeyann:
## Description
closes #13724

In old code, 1 thread per string is allocated for parsing a string column.
For longer strings (>1024), the runtime of 1-thread-per-string to decode is taking too long even for few strings.

In this change, 1 warp per string is used for parsing for strings length <=1024 and 1 block per string for string length >1024. If max string length < 128, 1 thread per string is used as usual.

256 threads_per_block is used for both kernels.
Code for 1-warp-per-string and 1-block-per-string is similar, but only varies with warp-wide and block-wide primitives for reduction and scan operations. shared memory usage will differ slightly too.

## Checklist


cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -295,6 +305,470 @@ process_string(in_iterator_t in_begin,
...
+      __shared__ typename BlockScan::TempStorage temp_slash;
+      state_table curr{is_within_bounds && c == '\\', false};  // state transition vector.
+      auto composite_op = [](state_table op1, state_table op2) {
+        return state_table{op2.state[op1.state[0]], op2.state[op1.state[1]]};
+      };
+      state_table scanned;
+      // inclusive scan. TODO both inclusive and exclusive available in cub.
+      BlockScan(temp_slash).InclusiveScan(curr, scanned, composite_op);
+      auto is_escaping_backslash = scanned.state[init_state];
+      if (threadIdx.x == BLOCK_SIZE - 1) init_state = is_escaping_backslash;
```
wence-, line 711:
```suggestion
      __syncthreads();
      if (threadIdx.x == BLOCK_SIZE - 1) init_state = is_escaping_backslash;
```

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -295,6 +305,470 @@ process_string(in_iterator_t in_begin,
...
+//        // if [c-7]=='\' & [c-6]=='u' & end_idx-curr_idx >= UNICODE_HEX_DIGIT_COUNT &&
+//        hex,DECODE [c-5:4]=prev_hex_val prev_hex_val, curr_hex_val, next_hex_val
+//        // if prev_hex_val in high, curr_hex_val in low, skip.
+//        // if curr_hex_val in high, next_hex_val in low, write u16.
+// if curr_hex_val not in high, write u8.
+// before writing, find size, then intra-warp scan for out_idx
+// propagate offset from 32nd thread to others in warp to carry forward.
+// 1 warp per string.
+template <typename str_tuple_it>
+__global__ void parse_fn_warp_parallel(str_tuple_it str_tuples,
```
vuule, line 343: All implementations should be in a .cu file.

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -366,20 +841,102 @@ std::unique_ptr<column> parse_data(str_tuple_it str_tuples,
...
+        col_size,
+        str_counter.data(),
+        static_cast<bitmask_type*>(null_mask.data()),
+        null_count_data,
+        options,
+        d_offsets,
+        d_chars2);
+      str_counter.set_value(0, stream);
+
+      // if (max_length > WARP_THRESHOLD)
```
vuule, line 921:
```suggestion
      // for strings longer than WARP_THRESHOLD
```

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -366,20 +841,102 @@ std::unique_ptr<column> parse_data(str_tuple_it str_tuples,
...
+        str_tuples,
+        col_size,
+        str_counter.data(),
+        static_cast<bitmask_type*>(null_mask.data()),
+        null_count_data,
+        options,
+        d_offsets,
+        nullptr);
+      str_counter.set_value(0, stream);
+      // if (max_length > WARP_THRESHOLD)
```
vuule, line 890:
```suggestion
      // for strings longer than WARP_THRESHOLD
```

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -366,20 +841,102 @@ std::unique_ptr<column> parse_data(str_tuple_it str_tuples,
...
+      auto offsets2 = cudf::make_numeric_column(
+        data_type{cudf::type_id::INT32}, col_size + 1, cudf::mask_state::UNALLOCATED, stream, mr);
+      auto d_offsets = offsets2->mutable_view().data<size_type>();
+
+      constexpr auto warps_per_block  = 8;
+      constexpr int threads_per_block = cudf::detail::warp_size * warps_per_block;
+      auto num_blocks                 = min(65535, col_size / warps_per_block + 1);
+      auto str_counter                = cudf::numeric_scalar(size_type{0}, true, stream);
+
+      parse_fn_warp_parallel<<<num_blocks, threads_per_block, 0, stream.value()>>>(
```
vuule, line 880: `parse_fn_warp_parallel` and `parse_fn_block_parallel` look independent. Might improve performance to run them in different stream. I think having a static stream pool with two streams would be sufficient here to help overlap some of the work.

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -366,20 +841,102 @@ std::unique_ptr<column> parse_data(str_tuple_it str_tuples,
...
+          null_count_data,
+          options,
+          d_offsets,
+          nullptr);
+      auto const bytes =
+        cudf::detail::sizes_to_offsets(d_offsets, d_offsets + col_size + 1, d_offsets, stream);
+      str_counter.set_value(0, stream);
+
+      // CHARS column
+      std::unique_ptr<column> chars2 =
```
vuule, line 906: Let's use a descriptive name, something like
```suggestion
      std::unique_ptr<column> parsed_chars =
```

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -295,6 +305,470 @@ process_string(in_iterator_t in_begin,
...
+                // if u8
+                write_char   = strings::detail::codepoint_to_utf8(hex_val);
+                this_num_out = strings::detail::bytes_in_char_utf8(write_char);
+              }
+            }
+          }
+        }
+      }  // !skip end.
+      {
+        using BlockScan2 = cub::BlockScan<size_type, BLOCK_SIZE>;
```
vuule, line 756: same as char2, please use a more descriptive name.

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -295,6 +305,470 @@ process_string(in_iterator_t in_begin,
...
+      auto escaped_char = get_escape_char(c);
+      bool error        = false;
+      if (is_within_bounds) {
+        error |= (c == '\\' && char_index == (in_end - in_begin) - 1);
+        error |= (prev_c == '\\' && escaped_char == NON_ESCAPE_CHAR);
+        error |= (prev_c == '\\' && c == 'u' &&
+                  ((in_begin + char_index + UNICODE_HEX_DIGIT_COUNT >= in_end) |
+                   !is_hex(in_begin[char_index + 1]) | !is_hex(in_begin[char_index + 2]) |
+                   !is_hex(in_begin[char_index + 3]) | !is_hex(in_begin[char_index + 4])));
+      }
```
vuule, line 653: is this the part that you mention had to be duplicated because of warp vs block?

vuule (commented):
partial review, got a bit hung up on higher level details

cpp/src/io/json/write_json.cu
```diff
@@ -61,6 +61,21 @@
 #include <vector>
 
 namespace cudf::io::json::detail {
+__device__ __forceinline__ thrust::pair<char, char> get_escaped_char(char escaped_char)
```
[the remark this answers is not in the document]
karthikeyann: Yes. This is the only function in this file depending on data_casting.cuh, but this writer code file does not need to be recompiled when I change the reader. So, I duplicated the code here.

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -366,20 +841,102 @@ std::unique_ptr<column> parse_data(str_tuple_it str_tuples,
...
+      str_tuples,
+      str_tuples + col_size,
+      [] __device__(auto t) { return t.second; },
+      size_type{0},
+      thrust::maximum<size_type>{});
+
+    if (max_length < SINGLE_THREAD_THRESHOLD) {
+      // this utility calls the functor to build the offsets and chars columns;
+      // the bitmask and null count may be updated by parse failures
+      nvtxRangePush("make_strings_children");
```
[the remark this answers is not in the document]
karthikeyann: This is a debug nvtx range. This will be removed.

cpp/include/cudf/io/detail/data_casting.cuh
karthikeyann: updated in the merged code.

cpp/include/cudf/io/detail/data_casting.cuh
karthikeyann: merged both warp and block kernel to single kernel with template parameter.

cpp/include/cudf/io/detail/data_casting.cuh
karthikeyann: updated.

cpp/include/cudf/io/detail/data_casting.cuh
karthikeyann: done

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -295,6 +305,470 @@ process_string(in_iterator_t in_begin,
   return {bytes, data_casting_result::PARSING_SUCCESS};
 }
 
+// 1 warp per string.
+// algorithm
+
+// \uXXXX         6->2/3/4
+// \uXXXX\uXXXX  12->2/3/4
+// \"             2->1
+// _              1->1
```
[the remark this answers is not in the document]
karthikeyann: updated. Thanks!

cpp/include/cudf/io/detail/data_casting.cuh
karthikeyann: do we already use static stream pool in io code?

robertmaynard (approved):
CMake changes LGTM

cpp/include/cudf/io/detail/data_casting.cuh
vuule: not yet. I hope it will be merged soon, in https://github.com/rapidsai/cudf/pull/13922.
We can leave a TODO until the global pool lands.

cpp/src/io/json/write_json.cu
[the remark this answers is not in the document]
vuule: Should we move this to `parsing_utils.cuh`? `get_escaped_char` and `get_escape_char` seem generic.

cpp/src/io/utilities/string_parsing.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * @param data JSON string input
+ * @param offset_length_begin The beginning of an offset-length tuple sequence
+ * @param size Size of the string input
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @return The inferred data type
+ */
+cudf::data_type infer_data_type(
+  cudf::io::json_inference_options_view const& options,
+  device_span<char const> data,
+  thrust::zip_iterator<thrust::tuple<const size_type*, const size_type*>> offset_length_begin,
```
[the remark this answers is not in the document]
karthikeyann: both are templated functions. The templates seems to useful only for unit test. So, I replaced the templates. There is no specific reason for zip iterator. Old code has zip_iterator, so, I used the same. it could be separate iterators too.

cpp/tests/io/json_test.cpp
```diff
@@ -1370,6 +1370,39 @@ TEST_F(JsonReaderTest, JsonExperimentalLines)
...
+TEST_F(JsonReaderTest, JsonLongString)
+{
+  std::string json_string =
+    R"({"a":"a0"}
+    {"a":"a1"}
+    {"a":"a2", "b":"b2"}
+    {"a":"a3", "c":"c3"})";
+  json_string += std::string("\n") + R"({"a":"a4\u20ac)" + std::string(103, 'a') + "\"}";
+  json_string += std::string("\n") + R"({"a":"a5\u20ac)" + std::string(1034, 'a') + "\"}";
+  json_string += std::string("\n") + R"({"a":"a6\u20ac)" + std::string(10340, 'a') + "\"}";
```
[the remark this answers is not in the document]
karthikeyann: No. You are right. It needs to be more varied, mix of unicodes too.

cpp/src/io/json/write_json.cu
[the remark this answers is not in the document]
karthikeyann: moved.

cpp/tests/io/json_test.cpp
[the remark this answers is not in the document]
karthikeyann: Added a unit test with varied and mixed unicodes of different lengths.

harrism:
Since this is an optimization, do you have benchmark results?

cpp/tests/io/json_test.cpp
```diff
@@ -1370,6 +1371,66 @@ TEST_F(JsonReaderTest, JsonExperimentalLines)
...
+    {"\"\\/\b\f\n\r\t",
+     "\"",
+     "\\",
+     "/",
+     "\b",
+     "\f\n",
+     "\r\t",
+     "$€",
+     "ராபிட்ஸ்",
+     "C𝞵𝓓𝒻",
```
vuule, line 1389: ❤️

cpp/tests/io/json_test.cpp
```diff
@@ -1370,6 +1371,66 @@ TEST_F(JsonReaderTest, JsonExperimentalLines)
...
+     "",                                                // null
+     "கார்த்தி",
+     "CႮ≪ㇳ䍏凹沦王辿龸ꁗ믜스폶ﴠ",  //  0000-FFFF
+     "𐀀𑿪𒐦𓃰𔙆 𖦆𗿿𘳕𚿾[↳] 𜽆𝓚𞤁🄰",                            // 10000-1FFFF
+     "𠘨𡥌𢗉𣇊𤊩𥅽𦉱𧴱𨁲𩁹𪐢𫇭𬬭𭺷𮊦屮",                // 20000-2FFFF
+     "𰾑𱔈𲍉"},                                         // 30000-3FFFF
+    cudf::test::iterators::nulls_at({10, 11})};
+
+  cudf::test::fixed_width_column_wrapper<int16_t> repeat_times{
+    1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 20, 40, 80, 160, 320, 640, 1280};
```
vuule, line 1400: Just in case we have some issues when the length is not aligned with warps. Feel free to reject, it does look odd.
```suggestion
    1, 2, 3, 4, 5, 6, 7, 8, 9, 13, 19, 37, 81, 161, 323, 631, 1279};
```

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,812 @@
...
+    while (in_begin != in_end) {
+      if (d_buffer) *d_buffer++ = *in_begin;
+      ++in_begin;
+      ++bytes;
+    }
+    return {bytes, data_casting_result::PARSING_SUCCESS};
+  }
+  // Whether in the original JSON this was a string value enclosed in quotes
+  // ({"a":"foo"} vs. {"a":1.23})
+  char const backslash_char = '\\';
```
vuule, line 191: the comment does not seem to match the variable (meant for is_string_value?)

vuule (commented):
Another partial review
Thank you for expanding the test!

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,832 @@
...
+// 1 warp per string or 1 block per string
+template <bool is_warp, size_type num_warps, typename str_tuple_it>
+__global__ void parse_fn_string_parallel(str_tuple_it str_tuples,
+                                         size_type total_out_strings,
+                                         size_type* str_counter,
+                                         bitmask_type* null_mask,
+                                         size_type* null_count_data,
+                                         cudf::io::parse_options_view const options,
+                                         size_type* d_offsets,
+                                         char* d_chars)
```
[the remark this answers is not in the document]
karthikeyann: added

cpp/src/io/utilities/data_casting.cu
karthikeyann: Yes. removed.

karthikeyann:
Resource usage on GV100.
Block per string kernel:
 Function void cudf::io::json::detail::parse_fn_string_parallel<(bool)0, (int)8, thrust::transform_iterator<cudf::io::json::detail::to_string_view_pair<char>, thrust::zip_iterator<thrust::tuple<const int *, const int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, thrust::use_default, thrust::use_default>>(T3, int, int *, unsigned int *, int *, cudf::io::parse_options_view, int *, char *):
  REG:63 STACK:8 SHARED:2112 LOCAL:0 CONSTANT[0]:496 CONSTANT[2]:88 TEXTURE:0 SURFACE:0 SAMPLER:0

Warp per string kernel:
 Function void cudf::io::json::detail::parse_fn_string_parallel<(bool)1, (int)8, thrust::transform_iterator<cudf::io::json::detail::to_string_view_pair<char>, thrust::zip_iterator<thrust::tuple<const int *, const int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, thrust::use_default, thrust::use_default>>(T3, int, int *, unsigned int *, int *, cudf::io::parse_options_view, int *, char *):
  REG:52 STACK:8 SHARED:296 LOCAL:0 CONSTANT[0]:496 CONSTANT[2]:88 TEXTURE:0 SURFACE:0 SAMPLER:0
  
Instead of bool array, we could reduce the shared memory usage by ~200 bytes in shared memory, by using 1 bit instead of 1 byte.

karthikeyann:
For code snippet in #13724, (in GV100 32GB)
branch-23.10:
```
variable lengths 0.06959819793701172
constant lengths 0.024032115936279297
```
This PR:
```
variable lengths 0.03034520149230957
constant lengths 0.018741130828857422
```

### Bechmark comparison:
['../../branch-23.10/release/branch-23.10.json', './512_4K.json']
# json_read_string_column

## [0] Quadro GV100

|  max_length  |  is_fixed_length  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |             Diff |   %Diff |  Status  |
|--------------|-------------------|------------|-------------|------------|-------------|------------------|---------|----------|
|      16      |         1         | 374.700 ms |       0.27% | 375.921 ms |       0.22% |         1.221 ms |   0.33% |   FAIL   |
|      32      |         1         | 312.954 ms |       0.10% | 315.346 ms |       0.28% |         2.392 ms |   0.76% |   FAIL   |
|     128      |         1         | 129.315 ms |       0.12% | 131.216 ms |       0.44% |         1.901 ms |   1.47% |   FAIL   |
|     256      |         1         | 136.168 ms |       0.54% | 137.339 ms |       0.50% |         1.171 ms |   0.86% |   FAIL   |
|     512      |         1         | 169.409 ms |       0.04% | 188.759 ms |       0.04% |        19.350 ms |  11.42% |   FAIL   |
|     1024     |         1         | 267.165 ms |       0.03% | 176.773 ms |       0.06% |    -90392.752 us | -33.83% |   FAIL   |
|   1048576    |         1         |   35.039 s |        inf% |   18.166 s |        inf% | -16872318.359 us | -48.15% |   PASS   |
|      16      |         0         | 446.740 ms |       0.21% | 449.951 ms |       0.44% |         3.211 ms |   0.72% |   FAIL   |
|      32      |         0         | 309.995 ms |       0.26% | 313.321 ms |       0.18% |         3.326 ms |   1.07% |   FAIL   |
|     128      |         0         | 141.225 ms |       0.36% | 143.295 ms |       0.44% |         2.069 ms |   1.47% |   FAIL   |
|     256      |         0         | 125.929 ms |       0.28% | 128.335 ms |       0.50% |         2.406 ms |   1.91% |   FAIL   |
|     512      |         0         | 143.098 ms |       0.50% | 160.784 ms |       0.44% |        17.685 ms |  12.36% |   FAIL   |
|     1024     |         0         | 210.757 ms |       0.30% | 206.471 ms |       0.10% |     -4285.916 us |  -2.03% |   FAIL   |
|   1048576    |         0         |   32.716 s |        inf% |   20.133 s |        inf% | -12582935.547 us | -38.46% |   PASS   |

For size>=512 warp kernel is also used. For size>=4K, block kernel is also used.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+    char* d_buffer = d_chars ? d_chars + d_offsets[istring] : nullptr;
+
+    // Copy literal/numeric value
+    if (not is_string_value) {
+      if (!d_chars) {
+        if (lane == 0) { d_offsets[istring] = in_end - in_begin; }
+      } else {
+        for (size_type char_index = lane; char_index < (in_end - in_begin);
+             char_index += BLOCK_SIZE) {
+          d_buffer[char_index] = in_begin[char_index];
```
vuule, line 455: How common is this path? We could probably make this faster with something like `memcpy_block`, but I'm not sure how important it is overall.

vuule (commented):
Another batch of comments, nothing big.
Almost done!

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+        __syncwarp();
+        is_slash.set_bits(warp_id, is_escaping_backslash);
+        __syncwarp();
+        is_prev_escaping_backslash = is_slash.get_bit(warp_id, lane - 1);
+      } else {
+        using SlashScan = cub::BlockScan<state_table, BLOCK_SIZE>;
+        __shared__ typename SlashScan::TempStorage temp_slash;
+        SlashScan(temp_slash).InclusiveScan(curr, scanned, composite_op);
+        is_escaping_backslash = scanned.state[init_state];
+        if (threadIdx.x == BLOCK_SIZE - 1) init_state = is_escaping_backslash;
```
wence-, line 541:
```suggestion
        __syncthreads();
        if (threadIdx.x == BLOCK_SIZE - 1) init_state = is_escaping_backslash;
```

To avoid WAR conflict with previous line.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+      // TODO both inclusive and exclusive available in cub.
+      if constexpr (is_warp) {
+        using SlashScan = cub::WarpScan<state_table>;
+        __shared__ typename SlashScan::TempStorage temp_slash[num_warps];
+        SlashScan(temp_slash[warp_id]).InclusiveScan(curr, scanned, composite_op);
+        is_escaping_backslash = scanned.state[init_state];
+        init_state            = __shfl_sync(MASK, is_escaping_backslash, BLOCK_SIZE - 1);
+        __syncwarp();
+        is_slash.shift(warp_id);
+        __syncwarp();
```
wence-, line 532: Is it the case that one must always sync after `shift`/`set_bits`? Perhaps these syncs into the definition of `shift`/`set_bits`?

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+        __syncthreads();
+        error = error_reduced;
+      }
+      if (error) {
+        if (lane == 0) {
+          if (null_mask != nullptr) {
+            clear_bit(null_mask, istring);
+            atomicAdd(null_count_data, 1);
+          }
+          last_offset = 0;
```
wence-, line 580: I didn't quite manage to keep track but I _think_ that in the `!is_warp` case, this is a `__shared__` variable, so does one need to `__syncthreads()` here?

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+          OffsetScan(temp_storage).ExclusiveSum(this_num_out, offset);
+          __syncthreads();
+        }
+        offset += last_offset;
+        if (d_chars && !skip) { strings::detail::from_char_utf8(write_char, d_buffer + offset); }
+        offset += this_num_out;
+        if constexpr (is_warp) {
+          last_offset = __shfl_sync(0xffffffff, offset, BLOCK_SIZE - 1);
+        } else {
+          if (threadIdx.x == BLOCK_SIZE - 1) last_offset = offset;
```
wence-, line 661:
```suggestion
          __syncthreads();
          if (threadIdx.x == BLOCK_SIZE - 1) last_offset = offset;
```

Rationale: potential WAR conflict with line 655?

wence- (dismissed):
I found it rather hard to follow the logic in `parse_string_parallel_fn`, but I _think_ there are some missing syncs in the `!is_warp` case, could you check please?

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+                                         cudf::io::parse_options_view const options,
+                                         size_type* d_offsets,
+                                         char* d_chars)
+{
+  constexpr auto BLOCK_SIZE =
+    is_warp ? cudf::detail::warp_size : cudf::detail::warp_size * num_warps;
+  size_type lane = is_warp ? (threadIdx.x % BLOCK_SIZE) : threadIdx.x;
+
+  // get 1-string index per warp
+  auto get_next_string = [&]() {
```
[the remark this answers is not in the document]
karthikeyann: This mechanism is used for 1-string per warp, & block because it does not cause divergence. (entire warp works on a string, so no intra-warp divergence, once a warp is done, it can pick up next string from this global counter. similar case for block-per-string too).
Also, when a string column has Large strings, the number of rows will be less. so atomicAdd will have less congestion. 
If we use same atomicAdd for smaller strings, congestion will happen for large number of rows. That's why this pattern is avoided in 1-thread-per-string.

in current implementation, long strings are deferred for warp, block kernels.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+  // String values are indicated by keeping the quote character
+  bool const is_string_value =
+    num_in_chars >= 2LL &&
+    (options.quotechar == '\0' ||
+     (*in_begin == options.quotechar) && (*thrust::prev(in_end) == options.quotechar));
+
+  // Copy literal/numeric value
+  if (not is_string_value) {
+    while (in_begin != in_end) {
+      if (d_buffer) *d_buffer++ = *in_begin;
```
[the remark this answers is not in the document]
karthikeyann: no. we will have strings in contiguous memory only.
I now changed to `thrust::copy`. 
is there a warp collaborative memcpy function available?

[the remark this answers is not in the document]
vuule: > is there a warp collaborative memcpy function available?

I'm convinced I saw something like that earlier, but can't find it now.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+    }
+  }
+
+  // The last character of the input is a backslash -> "fail"/null for this item
+  if (escape) { return {bytes, data_casting_result::PARSING_FAILURE}; }
+  return {bytes, data_casting_result::PARSING_SUCCESS};
+}
+
+template <unsigned num_warps>
+struct bitfield_warp {
```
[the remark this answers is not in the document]
karthikeyann: added.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+  // 5+32 for each warp.
```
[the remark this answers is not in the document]
karthikeyann: added comment about `5`

cpp/src/io/utilities/data_casting.cu
karthikeyann: This is not a very common path. This path is take only if string and non-string are mixed together in same column, or some string elements does not have quotes.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,925 @@
...
+  // get 1-string index per warp
+  auto get_next_string = [&]() {
+    if constexpr (is_warp) {
+      size_type istring;
+      if (lane == 0) { istring = atomicAdd(str_counter, 1); }
+      __syncwarp();
+      return __shfl_sync(0xffffffff, istring, 0);
+    } else {
+      __shared__ size_type istring;
+      if (lane == 0) { istring = atomicAdd(str_counter, 1); }
```
elstehle, line 427: important:
```suggestion
      // Ensure lane 0 doesn't update istring before all threads have read the previous iteration's istring value
      __syncthreads();
      __shared__ size_type istring;
      if (lane == 0) { istring = atomicAdd(str_counter, 1); }
```

elstehle (commented):
Flushing a few comments, mostly minor ones.
There are three potential data races that we want to fix. I recommend giving `compute-sanitizer` a try, e.g., `compute-sanitizer --tool racecheck ./cpp/build/release/gtests/JSON_TEST` to check for potential data races. I generally find it really helpful.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,925 @@
...
+ * @tparam num_warps Number of warps per block
+ * @tparam str_tuple_it Iterator type for tuple with string pointer and its length
+ * @param str_tuples iterator of tuple with string pointer and its length
+ * @param total_out_strings Number of string rows to be processed
+ * @param str_counter Counter to keep track of processed number of strings
+ * @param null_mask Null mask
+ * @param null_count_data pointer to store null count
+ * @param options Settings for controlling string processing behavior
+ * @param d_offsets Offsets to identify where to store the results for each string
+ * @param d_chars Character array to store the characters of strings
```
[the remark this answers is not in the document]
karthikeyann: This is typically how string algorithms in cudf is coded. We use the same kernel or functor for computing sizes if `d_chars` is null, and write output if `d_chars` is not null. So, I assumed it to be a well established convention.
We don't make it compile time because that will create a lot of kernels and inturn lead to bigger library size. so, we keep it as runtime parameter. 
Besides, the sizes can't be computed without actually decoding the string in this case. so, we need most of the code anyway in both kernels.

cpp/include/cudf/io/detail/data_casting.cuh
```diff
@@ -366,20 +841,102 @@ std::unique_ptr<column> parse_data(str_tuple_it str_tuples,
...
+                                 std::move(chars),
+                                 d_null_count.value(stream),
+                                 std::move(null_mask));
+    } else {
+      nvtxRangePush("string_parallel");
+      auto offsets2 = cudf::make_numeric_column(
+        data_type{cudf::type_id::INT32}, col_size + 1, cudf::mask_state::UNALLOCATED, stream, mr);
+      auto d_offsets = offsets2->mutable_view().data<size_type>();
+
+      constexpr auto warps_per_block  = 8;
```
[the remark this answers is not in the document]
karthikeyann: I preferred to keep a fixed block size because it's easy to manage the shared memory size.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+      if (d_buffer) *d_buffer++ = *in_begin;
+      ++in_begin;
+      ++bytes;
+    }
+    return {bytes, data_casting_result::PARSING_SUCCESS};
+  }
+  char const backslash_char = '\\';
+
+  // Escape-flag, set after encountering a backslash character
+  bool escape = false;
```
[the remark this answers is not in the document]
karthikeyann: updated.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+
+  // The last character of the input is a backslash -> "fail"/null for this item
+  if (escape) { return {bytes, data_casting_result::PARSING_FAILURE}; }
+  return {bytes, data_casting_result::PARSING_SUCCESS};
+}
+
+template <unsigned num_warps>
+struct bitfield_warp {
+  // 5+32 for each warp.
+  bool is_slash[num_warps][5 + 32];
```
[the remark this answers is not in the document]
karthikeyann: done.

cpp/src/io/utilities/data_casting.cu
karthikeyann: Thanks. moved it inside.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,925 @@
...
+  auto get_next_string = [&]() {
+    if constexpr (is_warp) {
+      size_type istring;
+      if (lane == 0) { istring = atomicAdd(str_counter, 1); }
+      __syncwarp();
+      return __shfl_sync(0xffffffff, istring, 0);
+    } else {
+      __shared__ size_type istring;
+      if (lane == 0) { istring = atomicAdd(str_counter, 1); }
+      __syncthreads();  // memory fence?
```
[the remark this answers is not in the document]
karthikeyann: question: Can I replace `__syncthreads()` with`__threadfence_block()`?
Why or why not?

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,925 @@
...
+
+    // Check if the value corresponds to the null literal
+    auto const is_null_literal =
+      (!d_chars) &&
+      serialized_trie_contains(options.trie_na, {in_begin, static_cast<std::size_t>(num_in_chars)});
+    if (is_null_literal && null_mask != nullptr) {
+      if (lane == 0) {
+        clear_bit(null_mask, istring);
+        atomicAdd(null_count_data, 1);
+        if (!d_chars) d_offsets[istring] = 0;
```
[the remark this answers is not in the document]
karthikeyann: `is_null_literal` is not supposed to have `(!d_chars)`. but i understand why it was added before. it's added so that null literal parsing happens during size estimation itself and does not get repeated during output writing.

wrapping whole block inside if(!d_chars) is good choice.

cpp/include/cudf/io/detail/data_casting.cuh
karthikeyann: updated. Thanks for catching this.

cpp/src/io/utilities/data_casting.cu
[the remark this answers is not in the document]
karthikeyann: is it [cooperative groups `memcpy_async`](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#memcpy-async)?

cpp/src/io/utilities/data_casting.cu
karthikeyann: Thanks.

cpp/src/io/utilities/data_casting.cu
karthikeyann: Added. Thanks.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,925 @@
...
+  constexpr auto BLOCK_SIZE =
+    is_warp ? cudf::detail::warp_size : cudf::detail::warp_size * num_warps;
+  size_type lane = is_warp ? (threadIdx.x % BLOCK_SIZE) : threadIdx.x;
+
+  // get 1-string index per warp
+  auto get_next_string = [&]() {
+    if constexpr (is_warp) {
+      size_type istring;
+      if (lane == 0) { istring = atomicAdd(str_counter, 1); }
+      __syncwarp();
```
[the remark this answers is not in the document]
karthikeyann: remove it.

cpp/src/io/utilities/data_casting.cu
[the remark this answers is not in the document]
karthikeyann: wrapped whole block inside if(!d_chars)

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,925 @@
...
+    // entire warp executes but with mask.
+    for (size_type char_index = lane;
+         char_index < cudf::util::round_up_safe(in_end - in_begin, static_cast<long>(BLOCK_SIZE));
+         char_index += BLOCK_SIZE) {
+      auto MASK =
+        is_warp ? __ballot_sync(0xffffffff, char_index < (in_end - in_begin)) : 0xffffffff;
+      bool is_within_bounds = char_index < (in_end - in_begin);
+      auto c                = is_within_bounds ? in_begin[char_index] : '\0';
+      auto prev_c       = (char_index > 0 and is_within_bounds) ? in_begin[char_index - 1] : '\0';
+      auto escaped_char = get_escape_char(c);
```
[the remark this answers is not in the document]
karthikeyann: updated.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,925 @@
...
+      __syncthreads();  // memory fence?
+      return istring;
+    }
+  };
+  // grid-stride loop.
+  for (size_type istring = get_next_string(); istring < total_out_strings;
+       istring           = get_next_string()) {
+    // skip nulls
+    if (null_mask != nullptr && not bit_is_set(null_mask, istring)) {
+      if (!d_chars) d_offsets[istring] = 0;
```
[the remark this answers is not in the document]
karthikeyann: Thanks. good suggestion.

cpp/src/io/utilities/data_casting.cu
karthikeyann: resolved a data hazard! Thanks.

cpp/src/io/utilities/data_casting.cu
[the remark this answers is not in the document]
elstehle: I see. Thanks for clarifying.

cpp/src/io/utilities/data_casting.cu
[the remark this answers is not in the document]
elstehle: No, `__syncthreads()` is right here, because you also need to make sure that threads other than lane 0 will not progress to reading `istring ` before lane 0 has updated its value via the atomicAdd. But you can remove the comment.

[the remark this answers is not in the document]
wence-: A `threadfence` only provides "happens-before" semantics for reads and writes. If you had a `threadfence` you would know that "if I observed the updated value of `istring`, then I must also be able to observe prior writes to unconnected data". Note the crucial "if". There's no synchronisation of the threads, so as @elstehle says, threads can happily walk past the `threadfence` and continue to read an old value of `istring`.

The [cppreference page](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence) for `atomic_thread_fence` is also instructive here. Consider this example:

```c++
std::atomic<int> flag = -1;
int data;

void thread1() {
    data = 2;
    std::atomic_thread_fence(std::memory_order::seq_cst); // this is `__threadfence_block()` memory order
    flag.fetch_add(1, std::memory_order::relaxed); // this is the memory order of `atomicXXX`
}

void thread2() {
    int const v0 = flag.load(std::memory_order::relaxed);
    std::atomic_thread_fence(std::memory_order::seq_cst);
    if (v0 != -1) {
       // _If_ we see the completion of the `fetch_add`, then, because of the fence,
       // we _must_ observe the write to data.
       // _But_, without synchronisation between the threads, we are not guaranteed to observe the write
       // thread2 can execute the load _before_ thread1 runs fetch_add.
        assert(data == 2);
    }
}
```

cpp/src/io/utilities/data_casting.cu
wence-: @karthikeyann I think this potential hazard still needs to be checked (this is the only outstanding one from my requested changes...)

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,928 @@
...
+ */
+template <unsigned num_warps>
+struct bitfield_warp {
+  static constexpr auto UNICODE_LOOK_BACK{5};
+  // 5 because for skipping unicode hex chars, look back upto 5 chars are needed.
+  // 5+32 for each warp.
+  bool is_slash[num_warps][UNICODE_LOOK_BACK + cudf::detail::warp_size];
+  __device__ void reset(unsigned warp_id)
+  {
+    if (threadIdx.x < UNICODE_LOOK_BACK) {
```
elstehle, line 298: important: I think every warp needs to reset the `UNICODE_LOOK_BACK` extra flags?
```suggestion
    if (threadIdx.x % cudf::detail::warp_size < UNICODE_LOOK_BACK) {
```

elstehle (commented):
Great work! Flushing a few more comments.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,928 @@
...
+      // After inclusive scan, all current S states translate to escaping backslash.
+      // All escaping backslash should be skipped.
+
+      struct state_table {
+        bool state[2];
+      };
+      state_table curr{is_within_bounds && c == '\\', false};  // state transition vector.
+      auto composite_op = [](state_table op1, state_table op2) {
+        return state_table{op2.state[op1.state[0]], op2.state[op1.state[1]]};
+      };
```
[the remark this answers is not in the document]
karthikeyann: `cuobjdump --dump-resource-usage` shows this.
 Function void cudf::io::json::detail::parse_fn_string_parallel<(bool)0, (int)8, thrust::transform_iterator<cudf::io::json::detail::to_string_view_pair<char>, thrust::zip_iterator<thrust::tuple<const int *, const int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, thrust::use_default, thrust::use_default>>(T3, int, int *, unsigned int *, int *, cudf::io::parse_options_view, int *, char *):
  REG:63 STACK:8 SHARED:2112 LOCAL:0 CONSTANT[0]:496 CONSTANT[2]:88 TEXTURE:0 SURFACE:0 SAMPLER:0

 Function void cudf::io::json::detail::parse_fn_string_parallel<(bool)1, (int)8, thrust::transform_iterator<cudf::io::json::detail::to_string_view_pair<char>, thrust::zip_iterator<thrust::tuple<const int *, const int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, thrust::use_default, thrust::use_default>>(T3, int, int *, unsigned int *, int *, cudf::io::parse_options_view, int *, char *):
  REG:52 STACK:8 SHARED:296 LOCAL:0 CONSTANT[0]:496 CONSTANT[2]:88 TEXTURE:0 SURFACE:0 SAMPLER:0

The gmem-backed local memory is LOCAL right? it's zero here.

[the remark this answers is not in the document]
elstehle: Possibly, I can't find the documentation on `cuobjdump's` `LOCAL`. Wondering how it would index into the array if it doesn't use local memory? Does it use branching?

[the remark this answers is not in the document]
elstehle: Were you able to replicate the performance gains from using existing utilities?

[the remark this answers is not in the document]
karthikeyann: Another suggestion without arrays. (using bitfields in struct)
```cpp
struct state_table {
        bool state0 : 1;
        bool state1 : 1;
        // bool state[2];
      };
      state_table curr{is_within_bounds && c == '\\', false};  // state transition vector.
      auto composite_op = [](state_table op1, state_table op2) {
        // return state_table{op2.state[op1.state[0]], op2.state[op1.state[1]]};
        return state_table{ op1.state0 ? op2.state1 : op2.state0,  op1.state1 ? op2.state1 : op2.state0};
      };
```
This change reduces register and shared memory usage. Also, **eliminates STACK to zero!** 

 Function void cudf::io::json::detail::parse_fn_string_parallel<(bool)0, (int)8, thrust::transform_iterator<cudf::io::json::detail::to_string_view_pair<char>, thrust::zip_iterator<thrust::tuple<const int *, const int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, thrust::use_default, thrust::use_default>>(T3, int, int *, unsigned int *, int *, cudf::io::parse_options_view, int *, char *):
  REG:55 STACK:0 SHARED:1824 LOCAL:0 CONSTANT[0]:496 CONSTANT[2]:88 TEXTURE:0 SURFACE:0 SAMPLER:0
 Function void cudf::io::json::detail::parse_fn_string_parallel<(bool)1, (int)8, thrust::transform_iterator<cudf::io::json::detail::to_string_view_pair<char>, thrust::zip_iterator<thrust::tuple<const int *, const int *, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>>, thrust::use_default, thrust::use_default>>(T3, int, int *, unsigned int *, int *, cudf::io::parse_options_view, int *, char *):
  REG:50 STACK:0 SHARED:296 LOCAL:0 CONSTANT[0]:496 CONSTANT[2]:88 TEXTURE:0 SURFACE:0 SAMPLER:0

Also, this boosts speedup for block kernel, and overall speedup of -37.69% in the benchmarks! 
Using `fst::detail::MultiFragmentInRegArray` got same runtime improvements (-35.12%), but shared memory didn't reduce.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,928 @@
...
+        error_reduced = ErrorReduce(temp_storage_error).Sum(error);  // TODO use cub::LogicalOR.
+        // only valid in thread0, so shared memory is used for broadcast.
+        __syncthreads();
+        error = error_reduced;
+      }
+      if (error) {
+        if (lane == 0) {
+          if (null_mask != nullptr) {
+            clear_bit(null_mask, istring);
+            atomicAdd(null_count_data, 1);
```
elstehle, line 600: question: Do we need to make sure that `clear_bit` and especially `atomicAdd` is only executed if `!d_chars`? Wouldn't it we count the null twice otherwise?

elstehle (commented):
One last comment, otherwise LGTM. I like the approach. Thanks for working on this 👍

cpp/src/io/utilities/data_casting.cu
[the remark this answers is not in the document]
elstehle: Yup, sure, that's fine too. As long as we reap the performance gains 🙂

cpp/src/io/utilities/data_casting.cu
karthikeyann: oh. that didn't happen because nulls are skipped in line 442-444
size computation kernel call sets the null, writing output kernel call skips it early.

cpp/src/io/utilities/data_casting.cu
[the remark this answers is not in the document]
vuule: I think that's the one!

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+  // Iterate over hex digits right-to-left
+  size_type index = UNICODE_HEX_DIGIT_COUNT;
+  while (index-- > 0) {
+    char const ch = str[index];
+    if (ch >= '0' && ch <= '9') {
+      result += static_cast<int32_t>((ch - '0') + 0) * base;
+      base *= hex_radix;
+    } else if (ch >= 'A' && ch <= 'F') {
+      result += static_cast<int32_t>((ch - 'A') + 10) * base;
+      base *= hex_radix;
```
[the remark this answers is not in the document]
karthikeyann: Perfect hash is great! Would like to learn about it more. Great suggestion.
When I tested with it, improvement in benchmarks was not significant. Also, unicode will be less, the most common case will be ascii and escape characters. So, I am skipping this for now.

cpp/src/io/utilities/data_casting.cu
karthikeyann: updated. Thanks!

cpp/src/io/utilities/data_casting.cu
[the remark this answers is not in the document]
wence-: > Perfect hash is great! Would like to learn about it more. Great suggestion

This is quite a nice exposition about the idea https://orlp.net/blog/worlds-smallest-hash-table/

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,904 @@
...
+
+#include <cub/cub.cuh>
+
+#include <memory>
+#include <type_traits>
+
+namespace cudf::io::json::detail {
+
+constexpr auto SINGLE_THREAD_THRESHOLD = 512;
+constexpr auto WARP_THRESHOLD          = 1024 * 4;
```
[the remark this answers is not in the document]
karthikeyann: The shared memory usage is not much(1840 bytes). So, as long as each SM has 2048 threads, it's enough to keep it occupied. So, I went for fixed block size (8*32=256).
Here, I got SINGLE_THREAD_THRESHOLD an WARP_THRESHOLD by some experimentation on V100.
I ran all of them again and adjusted it.
`SINGLE_THREAD_THRESHOLD = 128`
`WARP_THRESHOLD = 128 * 128`

vuule (approved):
Just a few small comments, looks great otherwise.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,954 @@
...
+    auto single_thread_fn =
+      string_parse<decltype(str_tuples)>{str_tuples,
+                                         static_cast<bitmask_type*>(null_mask.data()),
+                                         null_count_data,
+                                         options,
+                                         d_offsets};
+    thrust::for_each_n(rmm::exec_policy(stream),
+                       thrust::make_counting_iterator<size_type>(0),
+                       col_size,
+                       single_thread_fn);
```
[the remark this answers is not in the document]
karthikeyann: Yes. This writes sizes to d_offsets vector.

cpp/tests/io/json_test.cpp
```diff
@@ -1370,6 +1371,123 @@ TEST_F(JsonReaderTest, JsonExperimentalLines)
...
+  // Initialize parsing options (reading json lines)
+  cudf::io::json_reader_options json_lines_options =
+    cudf::io::json_reader_options::builder(
+      cudf::io::source_info{out_buffer.data(), out_buffer.size()})
+      .lines(true)
+      .dtypes(types);
+
+  // Read test data via nested JSON reader
+  auto const table = cudf::io::read_json(json_lines_options);
+  CUDF_TEST_EXPECT_TABLES_EQUIVALENT(expected, table.tbl->view());
```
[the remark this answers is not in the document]
karthikeyann: because of nullability.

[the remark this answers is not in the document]
vuule: My understanding was that both input and output have nulls. What am I missing?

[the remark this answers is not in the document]
karthikeyann: It's for the `repeat_times` int16 column. input doesn't have nullmask, output has nullmask.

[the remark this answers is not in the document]
karthikeyann: but we don't need that to be json column anyway. we can get rid of it.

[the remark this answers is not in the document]
karthikeyann: Fixed it

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+    if (thrust::distance(in_begin, in_end) >= NUM_UNICODE_ESC_SEQ_CHARS &&
+        *in_begin == backslash_char && *thrust::next(in_begin) == 'u') {
+      // Try to parse hex value following the '\' and 'u' characters from what may be a UTF16 low
+      // surrogate
+      hex_low_val = parse_unicode_hex(thrust::next(in_begin, 2));
+    }
+
+    // This is indeed a UTF16 surrogate pair
+    if (hex_val >= UTF16_HIGH_SURROGATE_BEGIN && hex_val < UTF16_HIGH_SURROGATE_END &&
+        hex_low_val >= UTF16_LOW_SURROGATE_BEGIN && hex_low_val < UTF16_LOW_SURROGATE_END) {
```
wence-, line 261: **question:** ‏Is it valid to have two successive UTF-16 codepoints where the first is contained in `[HIGH_SURROGATE_BEGIN, HIGH_SURROGATE_END)` and the second is _not contained_ in `[LOW_SURROGATE_BEGIN, LOW_SURROGATE_END)`? If not, then I think we are missing a parse error here for this case.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+  // 5 because for skipping unicode hex chars, look back up to 5 chars are needed.
+  // 5+32 for each warp.
+  bool is_slash[num_warps][UNICODE_LOOK_BACK + cudf::detail::warp_size];
+  __device__ void reset(unsigned warp_id)
+  {
+    if (threadIdx.x % cudf::detail::warp_size < UNICODE_LOOK_BACK) {
+      is_slash[warp_id][threadIdx.x % cudf::detail::warp_size] = 0;
+    }
+    is_slash[warp_id][threadIdx.x % cudf::detail::warp_size + UNICODE_LOOK_BACK] = 0;
+  }
```
wence-, line 302: **question:** ‏Does `reset` need a `__syncwarp()` like `shift` and `set_bits` do?

No, `reset` is called immediately before a `__syncthreads()`

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+  __device__ void shift(unsigned warp_id)
+  {
+    if (threadIdx.x % cudf::detail::warp_size < UNICODE_LOOK_BACK)
+      is_slash[warp_id][threadIdx.x % cudf::detail::warp_size] =
+        is_slash[warp_id][cudf::detail::warp_size + threadIdx.x % cudf::detail::warp_size];
```
wence-, line 307: So this shifts `UNICODE_LOOK_BACK` elements from the end of `is_slash` to the beginning.

**suggestion:** ‏Perhaps a comment to explain why this is the "right" way to shift the data?

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+    size_type& last_offset(is_warp ? last_offset_reg : last_offset_shared);
+    if (is_warp || lane == 0) {
+      init_state  = false;
+      last_offset = 0;
+    }
+    using bitfield =
+      std::conditional_t<is_warp, bitfield_warp<num_warps>, bitfield_block<num_warps>>;
+    __shared__ bitfield is_slash;
+    is_slash.reset(warp_id);
+    __syncthreads();
```
wence-, line 516: Comment: (To answer my query above about `reset` needing `__syncwarp()`). No, it does not, because of this `__syncthreads()` call.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+    // 0-31, 32-63, ... i*32-n.
+    // entire warp executes but with mask.
+    for (size_type char_index = lane;
+         char_index < cudf::util::round_up_safe(in_end - in_begin, static_cast<long>(BLOCK_SIZE));
+         char_index += BLOCK_SIZE) {
```
wence-, line 521: **question:** ‏Can this increment overflow? (cross-ref https://github.com/rapidsai/cudf/issues/10368)

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+                                            ((hex_val - UTF16_HIGH_SURROGATE_BEGIN) << 10) +
+                                            (hex_low_val - UTF16_LOW_SURROGATE_BEGIN);
+              write_char   = strings::detail::codepoint_to_utf8(unicode_code_point);
+              this_num_out = strings::detail::bytes_in_char_utf8(write_char);
+            } else {
+              if (hex_val >= UTF16_LOW_SURROGATE_BEGIN && hex_val < UTF16_LOW_SURROGATE_END) {
+                skip         = true;
+                this_num_out = 0;
+                write_char   = 0;
+              } else {
```
wence-, line 672: **issue/question:** ‏This logic looks a _little_ bit different to the UTF-16 codepoint handling in `process_string` above. Is that deliberate? If not, which one is correct?

Specifically, in `process_string` there's no handling of `hex_val` being a "low" surrogate in the `else` case.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+    auto const max_length = thrust::transform_reduce(
+      rmm::exec_policy(stream),
+      str_tuples,
+      str_tuples + col_size,
+      [] __device__(auto t) { return t.second; },
+      size_type{0},
+      thrust::maximum<size_type>{});
+
+    auto offsets = cudf::make_numeric_column(
+      data_type{cudf::type_id::INT32}, col_size + 1, cudf::mask_state::UNALLOCATED, stream, mr);
```
wence-, line 806:
```suggestion
      data_type{type_to_id<size_type>()}, col_size + 1, cudf::mask_state::UNALLOCATED, stream, mr);
```
To make any switch to a different `size_type` easier.

wence- (changes requested):
Some minor nitpicks:

two, I think, want addressed before merge (even if only to convince me that there is "nothing to see here"). Specifically:

- possible integer overflow in the grid stride loop
- logic difference in the two places UTF16 surrogate codepoints are handled

cpp/src/io/utilities/data_casting.cu
wence-: OK, thanks

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+    } else if (ch >= 'A' && ch <= 'F') {
+      result += static_cast<int32_t>((ch - 'A') + 10) * base;
+      base *= hex_radix;
+    } else if (ch >= 'a' && ch <= 'f') {
+      result += static_cast<int32_t>((ch - 'a') + 10) * base;
+      base *= hex_radix;
+    } else {
+      return -1;
+    }
+  }
```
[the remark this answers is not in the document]
karthikeyann: Good point. Updated teh code.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+    thrust::advance(in_begin, UNICODE_HEX_DIGIT_COUNT);
+
+    // If this may be a UTF-16 encoded surrogate pair:
+    // we expect another \uXXXX sequence
+    int32_t hex_low_val = 0;
+    if (thrust::distance(in_begin, in_end) >= NUM_UNICODE_ESC_SEQ_CHARS &&
+        *in_begin == backslash_char && *thrust::next(in_begin) == 'u') {
+      // Try to parse hex value following the '\' and 'u' characters from what may be a UTF16 low
+      // surrogate
+      hex_low_val = parse_unicode_hex(thrust::next(in_begin, 2));
```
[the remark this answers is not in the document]
karthikeyann: Good point. I have added a condition to skip parsing if hex_val is not in `HIGH_SURROGATE_BEGIN/END` range.

cpp/src/io/utilities/data_casting.cu
```diff
@@ -0,0 +1,952 @@
...
+
+      // Compute UTF16-encoded code point
+      uint32_t unicode_code_point = 0x10000 + ((hex_val - UTF16_HIGH_SURROGATE_BEGIN) << 10) +
+                                    (hex_low_val - UTF16_LOW_SURROGATE_BEGIN);
+      auto utf8_chars = strings::detail::codepoint_to_utf8(unicode_code_point);
+      bytes += write_utf8_char(utf8_chars, d_buffer);
+    }
+
+    // Just a single \uXXXX sequence
+    else {
```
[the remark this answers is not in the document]
karthikeyann: done.

cpp/src/io/utilities/data_casting.cu
karthikeyann: Thank you!

cpp/src/io/utilities/data_casting.cu
karthikeyann: `in_end - in_begin` is expected to be in range of `size_type`. so, overflow may happen for last `in_end - in_begin > std::numerical_limits<size_type>::max() - BLOCK_SIZE + 1`
I converted char_index to `size_t`

cpp/src/io/utilities/data_casting.cu
karthikeyann: added a comment

cpp/src/io/utilities/data_casting.cu
wence-: Thanks, although unsigned integer types normally generate _slightly_ worse code (the compiler has to work harder to prove to itself that the loop terminates, and so often ends up adding a little more checking).

Should this change have been to `thread_index_type`?

cpp/src/io/utilities/data_casting.cu
wence-: OK, so if I understand correctly, we want logic that is something like:

```
hex_val = parse_unicode_hex(stream);
if (is_high_surrogate(hex_val)) {
    auto hex_val_low = parse_unicode_hex(stream);
    if (is_low_surrogate(hex_val_low)) {
        // valid surrogate pair, decode
    } else {
        // invalid surrogate pair, parse error
        // I think this case is missing
        return {bytes, parse_error};
    }
} else {
   // non-surrogate pair, decode as normal
}
```

But I think the parse error in the case that we have a pair where the first is a surrogate pair and the second is not handled. Except that maybe this is deliberate per:

> The result is not valid UTF-16, but the majority of UTF-16 encoder and decoder implementations do this when translating between encodings.

So you're just "carrying on" and letting an invalid character appear in the output stream. Is that right?

If so, maybe

**suggestion:** Add a comment explaining this parsing behaviour.

wence- (approved):

cpp/src/io/utilities/data_casting.cu
karthikeyann: That's right.
added a comment at the place where this is not skipped.

cpp/src/io/utilities/data_casting.cu
karthikeyann: changed to `thread_index_type`

cpp/include/cudf/io/detail/data_casting.cuh
karthikeyann: Left a TODO.
stream pool did not achieve signficant improvement. So, will experiment more for another PR. Not including it in this PR.
`./_deps/nvbench-src/scripts/nvbench_compare.py ./c2.json c4.json `
['./c2.json', 'c4.json']
# json_read_string_column

## [0] Quadro GV100

|  max_length  |  is_fixed_length  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |         Diff |   %Diff |  Status  |
|--------------|-------------------|------------|-------------|------------|-------------|--------------|---------|----------|
|      16      |         1         | 244.442 ms |       0.21% | 245.402 ms |       0.96% |   959.972 us |   0.39% |   FAIL   |
|      32      |         1         | 160.267 ms |       0.18% | 160.262 ms |       0.09% |    -5.881 us |  -0.00% |   PASS   |
|      64      |         1         | 161.867 ms |       0.44% | 161.403 ms |       0.06% |  -463.632 us |  -0.29% |   FAIL   |
|     128      |         1         | 162.441 ms |       0.36% | 161.767 ms |       0.10% |  -673.526 us |  -0.41% |   FAIL   |
|     256      |         1         | 144.881 ms |       0.47% | 144.819 ms |       0.26% |   -62.016 us |  -0.04% |   PASS   |
|     512      |         1         | 132.372 ms |       0.48% | 131.672 ms |       0.07% |  -700.235 us |  -0.53% |   FAIL   |
|     1024     |         1         | 125.344 ms |       0.37% | 125.133 ms |       0.05% |  -211.272 us |  -0.17% |   FAIL   |
|     4096     |         1         | 121.457 ms |       0.50% | 120.974 ms |       0.08% |  -483.184 us |  -0.40% |   FAIL   |
|    16384     |         1         | 135.223 ms |       0.49% | 134.880 ms |       0.47% |  -343.044 us |  -0.25% |   PASS   |
|    65536     |         1         | 142.435 ms |       0.27% | 142.837 ms |       0.49% |   402.222 us |   0.28% |   FAIL   |
|    262144    |         1         | 150.038 ms |       0.41% | 150.162 ms |       0.43% |   123.804 us |   0.08% |   PASS   |
|   1048576    |         1         | 347.749 ms |       0.29% | 347.054 ms |       0.12% |  -695.074 us |  -0.20% |   FAIL   |
|   10485760   |         1         |    2.737 s |       0.08% |    2.735 s |       0.13% | -1876.563 us |  -0.07% |   PASS   |
|  104857600   |         1         |   18.052 s |        inf% |   18.056 s |        inf% |     4.113 ms |   0.02% |   PASS   |
|      16      |         0         | 338.887 ms |       0.18% | 337.814 ms |       0.07% | -1073.114 us |  -0.32% |   FAIL   |
|      32      |         0         | 202.162 ms |       0.26% | 203.290 ms |       0.48% |     1.128 ms |   0.56% |   FAIL   |
|      64      |         0         | 148.158 ms |       0.46% | 148.238 ms |       0.45% |    80.219 us |   0.05% |   PASS   |
|     128      |         0         | 188.805 ms |       0.14% | 188.934 ms |       0.14% |   129.172 us |   0.07% |   PASS   |
|     256      |         0         | 153.771 ms |       0.50% | 153.039 ms |       0.04% |  -731.565 us |  -0.48% |   FAIL   |
|     512      |         0         | 141.374 ms |       0.10% | 141.641 ms |       0.29% |   266.837 us |   0.19% |   FAIL   |
|     1024     |         0         | 132.822 ms |       0.49% | 132.785 ms |       0.48% |   -36.966 us |  -0.03% |   PASS   |
|     4096     |         0         | 126.245 ms |       0.47% | 126.147 ms |       0.18% |   -98.391 us |  -0.08% |   PASS   |
|    16384     |         0         | 128.630 ms |       0.30% | 126.918 ms |       0.49% | -1711.989 us |  -1.33% |   FAIL   |
|    65536     |         0         | 145.012 ms |       0.17% | 146.203 ms |       0.37% |     1.191 ms |   0.82% |   FAIL   |
|    262144    |         0         | 145.651 ms |       0.10% | 145.060 ms |       0.50% |  -590.480 us |  -0.41% |   FAIL   |
|   1048576    |         0         | 292.832 ms |       0.11% | 293.013 ms |       0.34% |   180.487 us |   0.06% |   PASS   |
|   10485760   |         0         |    2.323 s |       0.02% |    2.323 s |       0.05% |    -8.594 us |  -0.00% |   PASS   |
|  104857600   |         0         |   10.164 s |        inf% |   10.164 s |        inf% |   130.859 us |   0.00% |   PASS   |

karthikeyann:
## Final benchmark:

~70-80% runtime reduction for long strings. (overall range 30%-89%)
### json_read_string_column final comparion (branch-23.10 vs this PR)
single column with max length string.

#### [0] Quadro GV100

|  max_length  |  is_fixed_length  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |             Diff |   %Diff |
|--------------|-------------------|------------|-------------|------------|-------------|------------------|---------|
|      16      |         1         | 247.072 ms |       0.67% | 244.442 ms |       0.21% |     -2629.677 us |  -1.06% |
|      32      |         1         | 161.946 ms |       0.48% | 160.267 ms |       0.18% |     -1678.671 us |  -1.04% |
|      64      |         1         | 160.508 ms |       0.07% | 161.867 ms |       0.44% |         1.358 ms |   0.85% |
|     128      |         1         | 225.849 ms |       0.43% | 162.441 ms |       0.36% |    -63408.243 us | -28.08% |
|     256      |         1         | 291.920 ms |       0.33% | 144.881 ms |       0.47% |   -147039.083 us | -50.37% |
|     512      |         1         | 317.194 ms |       0.20% | 132.372 ms |       0.48% |   -184822.043 us | -58.27% |
|     1024     |         1         | 353.383 ms |       0.23% | 125.344 ms |       0.37% |   -228039.301 us | -64.53% |
|     4096     |         1         | 235.406 ms |       0.16% | 121.457 ms |       0.50% |   -113949.635 us | -48.41% |
|    16384     |         1         | 123.327 ms |       0.03% | 135.223 ms |       0.49% |        11.896 ms |   9.65% |
|    65536     |         1         | 272.949 ms |       0.04% | 142.435 ms |       0.27% |   -130513.632 us | -47.82% |
|    262144    |         1         | 951.804 ms |       0.03% | 150.038 ms |       0.41% |   -801766.458 us | -84.24% |
|   1048576    |         1         |    3.243 s |       0.03% | 347.749 ms |       0.29% |  -2895555.682 us | -89.28% |
|   10485760   |         1         |   15.121 s |        inf% |    2.737 s |       0.08% | -12384781.445 us | -81.90% |
|  104857600   |         1         |   59.740 s |        inf% |   18.052 s |        inf% | -41687603.516 us | -69.78% |
|      16      |         0         | 341.065 ms |       0.20% | 338.887 ms |       0.18% |     -2177.289 us |  -0.64% |
|      32      |         0         | 203.926 ms |       0.41% | 202.162 ms |       0.26% |     -1764.044 us |  -0.87% |
|      64      |         0         | 148.662 ms |       0.06% | 148.158 ms |       0.46% |      -504.291 us |  -0.34% |
|     128      |         0         | 179.354 ms |       0.80% | 188.805 ms |       0.14% |         9.450 ms |   5.27% |
|     256      |         0         | 243.734 ms |       0.44% | 153.771 ms |       0.50% |    -89962.469 us | -36.91% |
|     512      |         0         | 287.616 ms |       0.26% | 141.374 ms |       0.10% |   -146241.812 us | -50.85% |
|     1024     |         0         | 315.399 ms |       0.07% | 132.822 ms |       0.49% |   -182577.268 us | -57.89% |
|     4096     |         0         | 296.523 ms |       0.20% | 126.245 ms |       0.47% |   -170277.805 us | -57.42% |
|    16384     |         0         | 114.834 ms |       0.12% | 128.630 ms |       0.30% |        13.796 ms |  12.01% |
|    65536     |         0         | 210.012 ms |       0.34% | 145.012 ms |       0.17% |    -64999.411 us | -30.95% |
|    262144    |         0         | 697.045 ms |       0.04% | 145.651 ms |       0.10% |   -551394.577 us | -79.10% |
|   1048576    |         0         |    2.579 s |       0.01% | 292.832 ms |       0.11% |  -2285863.312 us | -88.64% |
|   10485760   |         0         |   11.416 s |        inf% |    2.323 s |       0.02% |  -9093176.855 us | -79.65% |
|  104857600   |         0         |   41.310 s |        inf% |   10.164 s |        inf% | -31145317.383 us | -75.39% |

karthikeyann:
Thanks to all the reviewers for the great inputs! Many suggestions lead to great performance improvements 🚀

cpp/include/cudf/io/detail/data_casting.cuh
vuule: Darn. It would be good to take a look at a profile with/without streams.

PR #13806, opened 2023-08-02
Fix for Parquet writer when requested pages per row is smaller than fragment size

etseidl:
## Description
#12685 introduced a bug in page calculation. If the `max_page_size_rows` parameter is set smaller than the page fragment size, the writer will produce a spurious empty page. This PR fixes this by only checking the fragment size if there are already rows in the page, and then returns the old check for number of rows exceeding the page limit.

Interestingly, libcudf can read these files with empty pages just fine, but parquet-mr cannot.

## Checklist

### DIFF
--- a/cpp/src/io/parquet/page_enc.cu
+++ b/cpp/src/io/parquet/page_enc.cu
@@ -433,8 +433,9 @@ __global__ void __launch_bounds__(128)
                                   max_RLE_page_size(col_g.num_rep_level_bits(), num_vals));
 
       if (num_rows >= ck_g.num_rows ||
-          (values_in_page > 0 && (page_size + fragment_data_size > this_max_page_size)) ||
-          rows_in_page + frag_g.num_rows > max_page_size_rows) {
+          (values_in_page > 0 && (page_size + fragment_data_size > this_max_page_size ||
+                                  rows_in_page + frag_g.num_rows > max_page_size_rows)) ||
+          rows_in_page >= max_page_size_rows) {
         if (ck_g.use_dictionary) {
           // Additional byte to store entry bit width
           page_size = 1 + max_RLE_page_size(ck_g.dict_rle_bits, values_in_page);
--- a/cpp/src/io/parquet/page_enc.cu
+++ b/cpp/src/io/parquet/page_enc.cu
@@ -432,10 +432,15 @@ __global__ void __launch_bounds__(128)
                                 max_RLE_page_size(col_g.num_def_level_bits(), num_vals) +
                                   max_RLE_page_size(col_g.num_rep_level_bits(), num_vals));
 
-      if (num_rows >= ck_g.num_rows ||
-          (values_in_page > 0 && (page_size + fragment_data_size > this_max_page_size ||
-                                  rows_in_page + frag_g.num_rows > max_page_size_rows)) ||
-          rows_in_page >= max_page_size_rows) {
+      // checks to see when we need to close the current page and start a new one
+      auto const is_last_chunk          = num_rows >= ck_g.num_rows;
+      auto const is_page_bytes_exceeded = page_size + fragment_data_size > this_max_page_size;
+      auto const is_page_rows_exceeded  = rows_in_page + frag_g.num_rows > max_page_size_rows;
+      // only check for limit overflow if there's already at least one fragment for this page
+      auto const is_page_too_big =
+        values_in_page > 0 && (is_page_bytes_exceeded || is_page_rows_exceeded);
+
+      if (is_last_chunk || is_page_too_big) {
         if (ck_g.use_dictionary) {
           // Additional byte to store entry bit width
           page_size = 1 + max_RLE_page_size(ck_g.dict_rle_bits, values_in_page);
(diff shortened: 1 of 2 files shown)


cpp/src/io/parquet/page_enc.cu
```diff
@@ -433,8 +433,9 @@ __global__ void __launch_bounds__(128)
                                   max_RLE_page_size(col_g.num_rep_level_bits(), num_vals));
 
       if (num_rows >= ck_g.num_rows ||
-          (values_in_page > 0 && (page_size + fragment_data_size > this_max_page_size)) ||
-          rows_in_page + frag_g.num_rows > max_page_size_rows) {
+          (values_in_page > 0 && (page_size + fragment_data_size > this_max_page_size ||
```
vuule, line 436: why is `values_in_page` used here instead of `rows_in_page`?

cpp/src/io/parquet/page_enc.cu
```diff
@@ -433,8 +433,9 @@ __global__ void __launch_bounds__(128)
...
+                                  rows_in_page + frag_g.num_rows > max_page_size_rows)) ||
+          rows_in_page >= max_page_size_rows) {
```
vuule, line 438: When is this check relevant? As I understand this, if `rows_in_page >= max_page_size_rows`, then `values_in_page > 0` is also true and the `rows_in_page + frag_g.num_rows > max_page_size_rows` condition will be checked anyway.

vuule (commented):
Should we refactor this condition to use a few local variables? It's getting pretty hefty.

cpp/src/io/parquet/page_enc.cu
etseidl: History? On the reader side there's the problem of pages with 0 rows but many values, so maybe that's why `values_in_page` was originally used?

cpp/src/io/parquet/page_enc.cu
ttnghia, line 438: To be honest, I feel the entire of this `if` condition is too complex and error-prone. Can we break it into multiple conditions? For example:
```
auto const name1 = num_rows >= ck_g.num_rows;
auto const name2 = ....;
....

if(name1 && name2 || ....) {
...
}
```

cpp/src/io/parquet/page_enc.cu
vuule: omg, I forgot about 0 row pages :face_exhaling: 
Need to stare at this some more now

cpp/src/io/parquet/page_enc.cu
etseidl: I'm not sure...put it back in just to be safe. I'll see if I can dream up some weird edge case where this is necessary :sweat_smile:

cpp/src/io/parquet/page_enc.cu
```diff
@@ -432,9 +432,15 @@ __global__ void __launch_bounds__(128)
...
-          rows_in_page + frag_g.num_rows > max_page_size_rows) {
+      // checks to see when we need to close the current page and start a new one
+      auto const is_last_chunk          = num_rows >= ck_g.num_rows;
+      auto const is_page_bytes_exceeded = page_size + fragment_data_size > this_max_page_size;
+      auto const is_page_rows_exceeded  = rows_in_page + frag_g.num_rows > max_page_size_rows;
+      // only check for limit overflow if there's already at least one fragment for this page
+      auto const is_page_too_big =
+        values_in_page > 0 && (is_page_bytes_exceeded || is_page_rows_exceeded);
+
+      if (is_last_chunk || is_page_too_big) {
```
ttnghia, line 443: Fantastic! This is much cleaner and better to understand what's going on.

ttnghia (approved):

cpp/tests/io/parquet_test.cpp
```diff
@@ -3709,6 +3709,44 @@ TEST_F(ParquetWriterTest, CheckPageRowsAdjusted)
...
+      .max_page_fragment_size(fragment_size)
+      .max_page_size_rows(rows_per_page);
+  cudf::io::write_parquet(out_opts);
+
+  // check that file is written correctly when rows/page < fragment size
+  auto const source = cudf::io::datasource::create(filepath);
+  cudf::io::parquet::FileMetaData fmd;
+
+  read_footer(source, &fmd);
+  CUDF_EXPECTS(fmd.row_groups.size() > 0, "No row groups found");
```
vuule, line 3736: AFAIK we should use `ASSERT_TRUE` instead of `CUDF_EXPECTS` in tests.
I'll get rid of the existing instances.

vuule (changes requested):
Looks great, thanks for cleaning up the condition!
Just one change request :)

cpp/tests/io/parquet_test.cpp
etseidl: Ah, I wasn't aware of the whole `ASSERT_X` vs `EXPECT_X` in gtest. So sorry, I use `CUDF_EXPECTS` a lot in the test code :(

vuule: Can't blame you, we missed it in code reviews as well.

vuule (approved):

PR #13807, opened 2023-08-02
Fix Byte-Pair-Encoding usage of cuco static-map for storing merge-pairs 

davidwendt:
## Description

Switching to use `cuco::experimental::static_map` for storing the unique merge-pair strings that can be looked up by `string_view`. This takes advantage of a feature of the `static_map` that allows storing with one key (index to a string entry) and lookup with a different type (string). The map uses a hash on the string for storing the index but allows lookup by string since the hash of string can resolve the entry and duplicates can be resolved by comparing the string with row entries.

## Checklist


cpp/src/text/subword/load_merges_file.cu
```diff
@@ -103,24 +96,22 @@ std::unique_ptr<cudf::column> load_file_to_column(std::string const& filename_me
 std::unique_ptr<detail::merge_pairs_map_type> initialize_merge_pairs_map(
   cudf::strings_column_view const& input, rmm::cuda_stream_view stream)
 {
+  auto d_strings = cudf::column_device_view::create(input.parent(), stream);
   // Ensure capacity is at least (size/0.7) as documented here:
   // https://github.com/NVIDIA/cuCollections/blob/6ec8b6dcdeceea07ab4456d32461a05c18864411/include/cuco/static_map.cuh#L179-L182
   auto merge_pairs_map = std::make_unique<merge_pairs_map_type>(
     static_cast<size_t>(input.size() * 2),  // capacity is 2x;
     cuco::empty_key{std::numeric_limits<cudf::hash_value_type>::max()},
```
PointKernel, line 104:
```suggestion
    cuco::empty_key{-1},
```
map key is no longer a hash value thus we need to change the empty key sentinel accordingly.

cpp/src/text/subword/bpe_tokenizer.cuh
```diff
@@ -21,27 +21,67 @@
...
+  }
+};
+
 using hash_table_allocator_type = rmm::mr::stream_allocator_adaptor<default_allocator<char>>;
 
-using merge_pairs_map_type = cuco::static_map<cudf::hash_value_type,
-                                              cudf::size_type,
-                                              cuda::thread_scope_device,
-                                              hash_table_allocator_type>;
+using probe_scheme = cuco::experimental::double_hashing<1, bpe_hasher>;
```
PointKernel, line 71:
```suggestion
using probe_scheme = cuco::experimental::linear_probing<1, bpe_hasher>;
```

PointKernel (commented):
The code should build properly with the proposed changes. The problem with `double_hashing` is that we expect users to provide a hasher that can be constructed from an integer (or a `seed`). In that way, we can minimize the chance of secondary collisions during probing but now I realize it's a too-restrictive requirement. By all means, for a single map with 50% occupancy, `linear_probing` would probably outperform `double_hashing` all the time.

PointKernel:
I'm looking into other issues in the meantime.

cpp/include/nvtext/bpe_tokenize.hpp
```diff
@@ -68,12 +68,6 @@ struct bpe_merge_pairs {
    * @return The number of merge pairs in the table
    */
   cudf::size_type get_size();
```
PointKernel, line 70: This seems no longer used.

cpp/src/text/subword/load_merges_file.cu
```diff
@@ -36,22 +36,14 @@
...
-
-  string_hasher_type const _hasher;
-  cudf::column_device_view const d_strings;
-};
+// struct make_pair_function {
+//   __device__ cuco::pair<cudf::size_type, cudf::size_type> operator()(cudf::size_type idx)
+//   {
+//     return cuco::make_pair(idx, idx);
+//   }
+// };
```
PointKernel, line 46:
```suggestion
```

cpp/src/text/subword/load_merges_file.cu
```diff
@@ -101,36 +93,34 @@ std::unique_ptr<cudf::column> load_file_to_column(std::string const& filename_me
...
-  auto iter = cudf::detail::make_counting_transform_iterator(0, pair_func);
+  auto iter = cudf::detail::make_counting_transform_iterator(
+    0, [] __device__(cudf::size_type idx) { return cuco::make_pair(idx, idx); });
 
-  merge_pairs_map->insert(iter,
-                          iter + input.size(),
-                          thrust::identity<cudf::hash_value_type>{},
-                          thrust::equal_to<cudf::hash_value_type>{},
-                          stream.value());
+  merge_pairs_map->insert(iter, iter + input.size(), stream.value());
```
PointKernel, line 112:
```suggestion
  merge_pairs_map->insert_async(iter, iter + input.size(), stream.value());
```
If the number of successful insertions is not needed, `insert_async` is less expensive and can get rid of one extra stream sync.

PointKernel (commented):
CMake changes to be reverted once https://github.com/rapidsai/rapids-cmake/pull/452 is merged.

cpp/src/text/subword/bpe_tokenizer.cuh
```diff
@@ -21,39 +21,96 @@
...
 #include <cudf/column/column.hpp>
+#include <cudf/column/column_device_view.cuh>
 #include <cudf/hashing/detail/murmurhash3_x86_32.cuh>
+#include <cudf/strings/string_view.cuh>
 
 #include <rmm/cuda_stream_view.hpp>
 #include <rmm/device_uvector.hpp>
 #include <rmm/mr/device/polymorphic_allocator.hpp>
 
+#include <cuco/operator.hpp>
```
PointKernel, line 32: Is this inclusion necessary? If so, I probably need to add it to the top-level `static_map.cuh` header as well.

PointKernel (approved):
Looks great! :fire: 

Heterogeneous lookup in libcudf :fire:

cpp/src/text/subword/bpe_tokenizer.cu
```diff
@@ -136,17 +137,13 @@ struct byte_pair_encoding_fn {
   }
 
   /**
-   * @brief Compute the hash over the input strings.
+   * @brief Lookup the pair of strings in the d_map/d_merges
```
bdice, line 140:
```suggestion
   * @brief Look up the pair of strings in the d_map/d_merges
```

cpp/src/text/subword/bpe_tokenizer.cu
```diff
@@ -155,7 +152,7 @@ struct byte_pair_encoding_fn {
     // Edge case check.
     // Empirically found only two merge pair strings that were greater than 70 bytes
     // and they both looked like ignorable errors. Double check this analysis with Vibhu.
```
bdice, line 154: Is "Double check this with Vibhu" an action item needed here, or is this comment outdated?

cpp/src/text/subword/bpe_tokenizer.cuh
```diff
@@ -30,30 +32,84 @@
...
 namespace detail {
 
+using hash_value_type    = uint32_t;
+using string_hasher_type = cudf::hashing::detail::MurmurHash3_x86_32<cudf::string_view>;
+
+/**
+ * @brief Hasher function used for building and using the cuco static-map
+ *
+ * This takes advantage of heterogeneous lookup feature in cuco static-map which
+ * allows inserting with one type (index) and looking up with a different type (string).
```
bdice, line 47: "index" seems like the wrong word here (and in the function signature). I'd propose "hash value" or similar. Indices are ordered, but hash values are just arbitrary integers.

bdice (commented):
This is really nice. Heterogeneous lookup is cool. Can you comment on the performance impact of this change?

cpp/src/text/subword/bpe_tokenizer.cuh
bdice: Ah, okay. My mistake.

bdice (approved):

PR #13808, opened 2023-08-02
Add cudf::strings::find function with target per row 

### LINKED ISSUE #12013: [FEA] Support a strings_column_view as the target parameter in cudf::string::find
Currently there is a limitation in RAPIDS plugin to support some string functions such as instr and locate `instr(column, column)`.
The plugin can do partial support (column, scalar)
To fully support the string operations, ColumnView.stringLocate needs to support two columns as arguments.

**Additional context**

- https://github.com/NVIDIA/spark-rapids/issues/6884
- https://github.com/NVIDIA/spark-rapids/issues/6930

davidwendt:
## Description

Adds new strings find function
```
std::unique_ptr<column> find(
  strings_column_view const& input,
  strings_column_view const& target,
  size_type start,
```
The output of row `i` is the character position of the target string for row `i`  within input string of row `i` starting at the character position `start`.  If the target is not found within the input string, -1 is returned for that row entry in the output column.

Closes #12013 

## Checklist


cpp/include/cudf/strings/find.hpp
```diff
@@ -87,6 +87,31 @@ std::unique_ptr<column> rfind(
...
+ * @param target Strings to search for in `input`
+ * @param start First character position to include in the search
+ * @param mr Device memory resource used to allocate the returned column's device memory
+ * @return New integer column with character position values
+ */
+std::unique_ptr<column> find(
+  strings_column_view const& input,
+  strings_column_view const& target,
+  size_type start                     = 0,
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
```
bdice, line 113: Should new public APIs include a stream parameter?

cpp/tests/strings/find_tests.cpp
```diff
@@ -63,6 +63,19 @@ TEST_F(StringsFindTest, Find)
     auto results = cudf::strings::rfind(strings_view, cudf::string_scalar(""));
     CUDF_TEST_EXPECT_COLUMNS_EQUAL(*results, expected);
   }
+  {
+    auto const targets = cudf::test::strings_column_wrapper({"l", "t", "", "x", "é", "o"});
```
bdice, line 67: Can we test:
- multi-character targets
- a target string that is longer than the corresponding input string
- a null target string

cpp/src/strings/search/find.cu
```diff
@@ -271,6 +321,15 @@ std::unique_ptr<column> rfind(strings_column_view const& strings,
   return detail::rfind(strings, target, start, stop, cudf::get_default_stream(), mr);
 }
 
+std::unique_ptr<column> find(strings_column_view const& input,
```
bdice, line 324: Do we need an `rfind` variation for target-per-row?

davidwendt: This was not requested. I don't want to add this feature without a reason.

bdice: Can you comment on the original issue to indicate that this can be done (the machinery appears to be in place) but we're not including in this PR because it was not requested? Just want to communicate that so that we aren't skipping over something that is easy and might be expected by the requesters.

bdice (approved):

cpp/src/strings/search/find.cu
bdice: Ehhh maybe not. I thought some more, and I think this can be skipped.

ttnghia: > This was not requested. I don't want to add this feature without a reason.

Typically, I always advocate for better coverage of the library feature.

PR #13810, opened 2023-08-02
Update `lists::contains` to experimental row comparator

divyegala:
## Description
Part of #11844

## Checklist

### DIFF
--- a/cpp/src/lists/contains.cu
+++ b/cpp/src/lists/contains.cu
@@ -216,8 +216,7 @@ std::unique_ptr<column> dispatch_index_of(lists_column_view const& lists,
   auto const has_nulls   = has_nested_nulls(child_tview) || has_nested_nulls(keys_tview);
   auto const comparator =
     cudf::experimental::row::equality::two_table_comparator(child_tview, keys_tview, stream);
-  if (search_keys.type().id() == cudf::type_id::STRUCT or
-      search_keys.type().id() == cudf::type_id::LIST) {
+  if (cudf::is_nested(search_keys.type())) {
     auto const d_comp = comparator.equal_to<true>(nullate::DYNAMIC{has_nulls});
     index_of(input_it, num_rows, output_it, child, search_keys, find_option, d_comp, stream);
   } else {
--- a/cpp/src/lists/contains.cu
+++ b/cpp/src/lists/contains.cu
@@ -18,6 +18,7 @@
 #include <cudf/detail/iterator.cuh>
 #include <cudf/detail/valid_if.cuh>
 #include <cudf/lists/detail/contains.hpp>
+#include <cudf/lists/detail/lists_column_factories.hpp>
 #include <cudf/lists/list_device_view.cuh>
 #include <cudf/lists/lists_column_device_view.cuh>
 #include <cudf/lists/lists_column_view.hpp>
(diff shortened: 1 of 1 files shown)


cpp/src/lists/contains.cu
```diff
@@ -172,119 +127,66 @@ struct search_list_nested_types_fn {
...
-    if constexpr (search_key_is_scalar) {
-      auto tmp_column = make_column_from_scalar(search_keys, 1, stream);
-      return {table_view{{tmp_column->view()}}, std::move(tmp_column)};
-    } else {
-      return {table_view{{search_keys}}, nullptr};
-    }
-  }();
-
+  auto const keys_tview  = cudf::table_view{{search_keys}};
   auto const child_tview = table_view{{child}};
```
bdice, line 166: Let's pick whether to use the `cudf::` namespace or not, and be consistent within this function.

cpp/src/lists/contains.cu
```diff
@@ -172,119 +127,66 @@ struct search_list_nested_types_fn {
...
+    thrust::transform(rmm::exec_policy(stream),
+                      input_it,
+                      input_it + num_rows,
+                      output_it,
+                      search_list{find_option, key_validity_iter, d_comp});
+  };
+
+  if (cudf::detail::has_nested_columns(child_tview) or
+      cudf::detail::has_nested_columns(keys_tview)) {
+    auto const comparator =
```
bdice, line 181: `comparator` is the same in both paths, right? Can we declare it outside the `if`?

bdice (changes requested):
Let's wait to merge until #13805 is merged, then the diff should get better here. Ideally we should rebase over `branch-23.10` once #13805 merges.

bdice (commented):
Code looks fine. I'd like to wait to approve until #13805 is merged, so I can see the final diff again.

cpp/src/lists/contains.cu
```diff
@@ -105,60 +105,19 @@ __device__ auto element_index_pair_iter(size_type const size)
...
-  }
-};
-
 /**
  * @brief Functor to perform searching for index of a key element in a given list, specialized
  * for nested types.
  */
 template <typename KeyValidityIter, typename EqComparator>
-struct search_list_nested_types_fn {
+struct search_list {
```
ttnghia, line 113: Typically I would prefer to have some suffix like `_fn` or `_functor` to let others know that this is a functor struct.

bdice: I am rather fond of that convention, it does help me a lot when skimming the code.

cpp/src/lists/contains.cu
```diff
@@ -190,129 +151,91 @@ struct search_list_nested_types_fn {
...
-      out_positions->set_null_mask(std::move(null_mask), null_count);
-    }
-    return out_positions;
+  auto const keys_tview  = cudf::table_view{{search_keys}};
+  auto const child_tview = cudf::table_view{{child}};
+  auto const has_nulls   = has_nested_nulls(child_tview) || has_nested_nulls(keys_tview);
+  auto const comparator =
+    cudf::experimental::row::equality::two_table_comparator(child_tview, keys_tview, stream);
+  if (search_keys.type().id() == cudf::type_id::STRUCT or
+      search_keys.type().id() == cudf::type_id::LIST) {
```
bdice, line 220: Use `is_nested` rather than explicitly listing the nested types. https://github.com/rapidsai/cudf/blob/dcc8d9195d62c64f0c132d0b67f9abf17207305e/cpp/src/utilities/traits.cpp#L352
```suggestion
  if (cudf::is_nested(search_keys.type())) {
```

bdice (approved):
Nice work. 🧘‍♂️ The zen of non-type-dispatching (043ae299f9246c7ec708b9544fbe416fcfdacf99) is now yours to cherish. Let it blossom throughout the code base.

ttnghia (approved):

PointKernel (approved):

PR #13811, opened 2023-08-02
Fix unbounded sequence issue in `DataFrame` constructor

galipremsagar:
## Description
In `cudf`, we currently have a hang in this scenario:
```python
In [1]: import cudf

In [2]:     class A:
   ...:         def __getitem__(self, key):
   ...:             return 1
   ...: 

In [3]: cudf.DataFrame([A()])
```

This PR introduces additional checks before letting the list-like inputs pass onto `itertools` for transposing.
## Checklist

### DIFF
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -843,8 +843,13 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
             data = DataFrame.from_pandas(pd.DataFrame(data))
             self._data = data._data
         else:
-            data = list(itertools.zip_longest(*data))
-
+            if all(
+                isinstance(a, abc.Iterable) or isinstance(a, abc.Sequence)
+                for a in data
+            ):
+                data = list(itertools.zip_longest(*data))
+            else:
+                raise TypeError("Inputs should be a bounded-sequence.")
             if columns is not None and len(data) == 0:
                 data = [
                     cudf.core.column.column_empty(row_count=0, dtype=None)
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/dataframe.py
```diff
@@ -843,8 +843,13 @@ def _init_from_list_like(self, data, index=None, columns=None):
             data = DataFrame.from_pandas(pd.DataFrame(data))
             self._data = data._data
         else:
-            data = list(itertools.zip_longest(*data))
-
+            if all(
+                isinstance(a, (abc.Iterable, abc.Sequence))
+                for a in data
```
galipremsagar, line 848:
```suggestion
                isinstance(col, (abc.Iterable, abc.Sequence))
                for col in data
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -843,8 +843,13 @@ def _init_from_list_like(self, data, index=None, columns=None):
...
         else:
-            data = list(itertools.zip_longest(*data))
-
+            if all(
+                isinstance(col, (abc.Iterable, abc.Sequence))
+                for col in data
+            ):
+                data = list(itertools.zip_longest(*data))
+            else:
+                raise TypeError("Inputs should be an iterable or sequence.")
```
bdice, line 852: Nit: maybe write “if not… raise” and then there’s no else required. Control flow can continue with “data = …”

bdice (approved):

python/cudf/cudf/core/dataframe.py
galipremsagar: Yup, updated 👍

PR #13812, opened 2023-08-02
Avoid use of CUDF_EXPECTS in libcudf unit tests outside of helper functions with return values

vuule:
## Description
`CUDF_EXPECTS` should be used to report internal errors in libcudf. Google test framework has macros that should be used to report failures. The `ASSERT_` macros replace `CUDF_EXPECTS` as they exit the current function (not by throwing as `CUDF_EXPECTS`, but the effect is very similar).

This PR replaces the erroneous use of CUDF_EXPECTS with ASSERT_XYZ macros in unit tests.

## Checklist

### DIFF
--- a/cpp/tests/io/arrow_io_source_test.cpp
+++ b/cpp/tests/io/arrow_io_source_test.cpp
@@ -97,7 +97,7 @@ TEST_F(ArrowIOTest, S3FileSystem)
 
     close_s3_func = reinterpret_cast<decltype(close_s3_func)>(
       dlsym(whole_app, "_ZN5arrow2fs17EnsureS3FinalizedEv"));
-    if (close_s3_func) { CUDF_EXPECTS(close_s3_func().ok(), "Failed to finalize s3 filesystem"); }
+    if (close_s3_func) { ASSERT_TRUE(close_s3_func().ok()); }
     dlclose(whole_app);
   }
 }
--- a/cpp/tests/io/arrow_io_source_test.cpp
+++ b/cpp/tests/io/arrow_io_source_test.cpp
@@ -97,7 +97,7 @@ TEST_F(ArrowIOTest, S3FileSystem)
 
     close_s3_func = reinterpret_cast<decltype(close_s3_func)>(
       dlsym(whole_app, "_ZN5arrow2fs17EnsureS3FinalizedEv"));
-    if (close_s3_func) { ASSERT_TRUE(close_s3_func().ok()); }
+    if (close_s3_func) { EXPECT_TRUE(close_s3_func().ok()); }
     dlclose(whole_app);
   }
 }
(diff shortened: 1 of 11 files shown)


cpp/tests/io/arrow_io_source_test.cpp
```diff
@@ -97,7 +97,7 @@ TEST_F(ArrowIOTest, S3FileSystem)
 
     close_s3_func = reinterpret_cast<decltype(close_s3_func)>(
       dlsym(whole_app, "_ZN5arrow2fs17EnsureS3FinalizedEv"));
-    if (close_s3_func) { CUDF_EXPECTS(close_s3_func().ok(), "Failed to finalize s3 filesystem"); }
+    if (close_s3_func) { ASSERT_TRUE(close_s3_func().ok()); }
```
davidwendt, line 100: Seems this one should be `EXPECT_TRUE` so as to not prevent the call to `dlclose()`.

cpp/tests/transpose/transpose_test.cpp
```diff
@@ -146,12 +146,10 @@ void run_test(size_t ncols, size_t nrows, bool add_nulls)
...
   auto result_view = std::get<1>(result);
 
-  CUDF_EXPECTS(result_view.num_columns() == expected_view.num_columns(),
-               "Expected same number of columns");
+  ASSERT_EQ(result_view.num_columns(), expected_view.num_columns());
   for (cudf::size_type i = 0; i < result_view.num_columns(); ++i) {
     CUDF_TEST_EXPECT_COLUMNS_EQUAL(result_view.column(i), expected_view.column(i));
-    CUDF_EXPECTS(result_view.column(i).null_count() == expected_nulls[i],
-                 "Expected correct null count");
+    ASSERT_EQ(result_view.column(i).null_count(), expected_nulls[i]);
```
davidwendt, line 152: This one seems safe to use `EXPECT_EQ` instead. Then the test will report all errors instead of just the first one.

cpp/tests/transform/row_bit_count_test.cu
```diff
@@ -751,7 +751,7 @@ TEST_F(RowBitCount, EmptyTable)
   {
     cudf::table_view empty;
     auto result = cudf::row_bit_count(empty);
-    CUDF_EXPECTS(result != nullptr && result->size() == 0, "Expected an empty column");
+    ASSERT_TRUE(result != nullptr && result->size() == 0);
```
davidwendt, line 754: This one does not seem like it should be stopping the entire test.

cpp/tests/transform/row_bit_count_test.cu
```diff
@@ -760,6 +760,6 @@ TEST_F(RowBitCount, EmptyTable)
     cudf::table_view empty({*strings, *ints});
 
     auto result = cudf::row_bit_count(empty);
-    CUDF_EXPECTS(result != nullptr && result->size() == 0, "Expected an empty column");
+    ASSERT_TRUE(result != nullptr && result->size() == 0);
```
davidwendt, line 763: Seems `EXPECT_TRUE` would be safe here.

davidwendt (changes requested):

davidwendt (approved):

ttnghia (approved):

PR #13818, opened 2023-08-04
Set native handles to null on close in Java wrapper classes

jlowe:
## Description
This updates a few Java wrapper classes that track native resources to zero the native handles when they are closed to make it easier to track down use-after-close errors.

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/ChunkedPack.java
+++ b/java/src/main/java/ai/rapids/cudf/ChunkedPack.java
@@ -88,7 +88,11 @@ public class ChunkedPack implements AutoCloseable {
 
   @Override
   public void close() {
-    chunkedPackDelete(nativePtr);
+    try {
+      chunkedPackDelete(nativePtr);
+    } finally {
+      nativePtr = 0;
+    }
   }
 
   private static native long chunkedPackGetTotalContiguousSize(long nativePtr);
--- a/java/src/main/java/ai/rapids/cudf/ColumnVector.java
+++ b/java/src/main/java/ai/rapids/cudf/ColumnVector.java
@@ -264,6 +264,7 @@ public final class ColumnVector extends ColumnView {
       eventHandler.onClosed(this, refCount);
     }
     if (refCount == 0) {
+      super.close();
       offHeap.clean(false);
     } else if (refCount < 0) {
       offHeap.logRefCountDebug("double free " + this);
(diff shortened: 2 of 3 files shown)


java/src/main/java/ai/rapids/cudf/ChunkedPack.java
```diff
@@ -88,7 +88,11 @@ public PackedColumnMetadata buildMetadata() {
 
   @Override
   public void close() {
-    chunkedPackDelete(nativePtr);
+    try {
+      chunkedPackDelete(nativePtr);
+    } finally {
+      nativePtr = 0;
```
ttnghia, line 94: `finally` is always executed right? So what if `chunkedPackDelete` failed? Will `nativePtr` be orphaned?

jlowe: The problem is that this is a destructor throwing, and we don't know how far along in the destruction it progressed before stopping.  If we get an exception while trying to delete a native resource, we don't know whether it's safe to try to delete it again (and it's very likely to throw again if we try again).  That's why we null it out, since we explicitly do not want to try to use this native resource again, even for destruction.  The other wrapper classes do the same thing.

ttnghia: But this `try` block will contain the exception within it. So if there was exception, should we rethrow it to let the application know that there was something wrong, instead of just keeping things silent like this?

gerashegalov: finally does not prevent the exception being thrown. It's important to keep the finally block free of exceptions which is the case here.

gerashegalov (approved):
LGTM

java/src/main/java/ai/rapids/cudf/ChunkedPack.java
ttnghia: I see. Thought that the `try{}` block always contains exceptions.

ttnghia (approved):

java/src/main/java/ai/rapids/cudf/ChunkedPack.java
gerashegalov: > I see. Thought that the `try{}` block always contains exceptions.

only if it matches a `catch` block

Issue #11559, opened 2022-08-17
[BUG] Arrow S3 tests don't throw an error when CUDF_ENABLE_ARROW_S3 is off


bdice:
**Describe the bug**

The C++ test `ArrowIOTest.S3URIWhenNotEnabled` expects that if the CMake variable `CUDF_ENABLE_ARROW_S3` is off, creating an `cudf::io::arrow_io_source` from an S3 path will throw a `cudf::logic_error`. However, that isn't necessarily the case. The new default setting of `CUDF_ENABLE_ARROW_S3=OFF` means that Arrow S3 support is not enabled _if Arrow is built from source_. If a pre-built Arrow library is found, it will be used and the value of `CUDF_ENABLE_ARROW_S3` is irrelevant. In most cases, I expect that pre-built Arrow libraries will have S3 support enabled (Arrow conda packages have it turned on).

```
[ RUN      ] ArrowIOTest.S3URIWhenNotEnabled
/home/bdice/rapids1/cudf/cpp/tests/io/arrow_io_source_test.cpp:91: Failure
Expected: std::make_unique<cudf::io::arrow_io_source>(s3_uri) throws an exception of type cudf::logic_error.
  Actual: it throws nothing.
[  FAILED  ] ArrowIOTest.S3URIWhenNotEnabled (6740 ms)
```

**Expected behavior**
This test should not fail. I'm not sure if we can determine at runtime whether the Arrow library being used has S3 support or not, but that's what we should be checking against rather than the [compile-time definition of `S3_ENABLED`](https://github.com/rapidsai/cudf/blob/abd4302d6d57c080323068f7b58f26b2d76e26fb/cpp/tests/io/arrow_io_source_test.cpp#L64).

We can't know from `CUDF_ENABLE_ARROW_S3` alone if the Arrow library will be able to read S3 files or not, unless we know that libcudf's CMake also built Arrow from source. We could have the opposite problem with `S3FileSystem` when the `CUDF_ENABLE_ARROW_S3` flag is enabled, if the found Arrow library wasn't built with S3 support.

**Environment overview (please complete the following information)**
 - Environment location: rapids-compose container, cudf 22.10 development branch, Arrow from conda package
 - Method of cuDF install: source

PR #13819, opened 2023-08-04
No need to dlsym EnsureS3Finalized we can call it directly

robertmaynard:
## Description
Previously I didn't realize that the `ARROW_S3` define was provided to consumers via `arrow/util/config.h`. We can remove the dlopen hack and just guard the entire S3 logic.

## Checklist

### DIFF
--- a/cpp/tests/io/arrow_io_source_test.cpp
+++ b/cpp/tests/io/arrow_io_source_test.cpp
@@ -95,7 +95,7 @@ TEST_F(ArrowIOTest, S3FileSystem)
     // https://github.com/apache/arrow/issues/36974
     // This needs to be in a separate conditional to ensure we call
     // finalize after all arrow_io_source instances have been deleted.
-    arrow::fs::EnsureS3Finalized();
+    [[maybe_unused]] arrow::fs::EnsureS3Finalized();
   }
 #endif
 }
--- a/cpp/tests/io/arrow_io_source_test.cpp
+++ b/cpp/tests/io/arrow_io_source_test.cpp
@@ -95,7 +95,7 @@ TEST_F(ArrowIOTest, S3FileSystem)
     // https://github.com/apache/arrow/issues/36974
     // This needs to be in a separate conditional to ensure we call
     // finalize after all arrow_io_source instances have been deleted.
-    [[maybe_unused]] arrow::fs::EnsureS3Finalized();
+    [[maybe_unused]] auto _ = arrow::fs::EnsureS3Finalized();
   }
 #endif
 }
--- a/cpp/tests/io/arrow_io_source_test.cpp
+++ b/cpp/tests/io/arrow_io_source_test.cpp
@@ -27,6 +27,7 @@
 
 #include <arrow/filesystem/filesystem.h>
 #include <arrow/io/api.h>
+#include <arrow/util/config.h>
 
 #include <fstream>
 #include <memory>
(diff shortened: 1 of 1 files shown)


cpp/tests/io/arrow_io_source_test.cpp
```diff
@@ -87,19 +88,16 @@ TEST_F(ArrowIOTest, S3FileSystem)
     ASSERT_EQ(1, tbl.tbl->num_columns());  // Only single column specified in reader_options
     ASSERT_EQ(244, tbl.tbl->num_rows());   // known number of rows from the S3 file
   }
+
+#ifdef ARROW_S3
```
bdice, line 92: I think we had some issues before with this, and decided it needs to be a runtime check. If you have a precompiled Arrow, you don't get this definition and you don't know how your precompiled Arrow was built, right?

See https://github.com/rapidsai/cudf/issues/11559 for related discussion.

bdice: (Note that this is using `ARROW_S3` rather than `S3_ENABLED`, so it might be different.)

robertmaynard: > (Note that this is using `ARROW_S3` rather than `S3_ENABLED`, so it might be different.)

Correct. This is the big difference compared to #11559 ( which steered me towards using dlopen ). `ARROW_S3` is a define provided by arrow itself ( not CMake ) and maps to the exact specification on how Arrow was built. So in this case we won't have any mismatches.

bdice: Okay good! Approving.

bdice: Ohhh it's in `<arrow/util/config.h>`. I overlooked that above. That makes lots more sense.

bdice (approved):

cpp/tests/io/arrow_io_source_test.cpp
```diff
@@ -87,19 +88,16 @@ TEST_F(ArrowIOTest, S3FileSystem)
...
     // This needs to be in a separate conditional to ensure we call
     // finalize after all arrow_io_source instances have been deleted.
-    void* whole_app                                       = dlopen(NULL, RTLD_LAZY);
-    decltype(arrow::fs::EnsureS3Finalized)* close_s3_func = nullptr;
-
-    close_s3_func = reinterpret_cast<decltype(close_s3_func)>(
-      dlsym(whole_app, "_ZN5arrow2fs17EnsureS3FinalizedEv"));
-    if (close_s3_func) { CUDF_EXPECTS(close_s3_func().ok(), "Failed to finalize s3 filesystem"); }
-    dlclose(whole_app);
+    arrow::fs::EnsureS3Finalized();
```
bdice, line 98: Probably need this to fix the build warning from `[[nodiscard]]`. Or you can check the status explicitly (not sure exactly how to write that without more thinking).
```suggestion
    [[maybe_unused]] arrow::fs::EnsureS3Finalized();
```

vuule (approved):

cpp/tests/io/arrow_io_source_test.cpp
```diff
@@ -87,19 +88,16 @@ TEST_F(ArrowIOTest, S3FileSystem)
...
     // This needs to be in a separate conditional to ensure we call
     // finalize after all arrow_io_source instances have been deleted.
-    void* whole_app                                       = dlopen(NULL, RTLD_LAZY);
-    decltype(arrow::fs::EnsureS3Finalized)* close_s3_func = nullptr;
-
-    close_s3_func = reinterpret_cast<decltype(close_s3_func)>(
-      dlsym(whole_app, "_ZN5arrow2fs17EnsureS3FinalizedEv"));
-    if (close_s3_func) { CUDF_EXPECTS(close_s3_func().ok(), "Failed to finalize s3 filesystem"); }
-    dlclose(whole_app);
+    [[maybe_unused]] arrow::fs::EnsureS3Finalized();
```
vyasr, line 98: Still need to assign the nodiscard output to something, then you can make it maybe_unused.
```suggestion
    [[maybe_unused]] auto _ = arrow::fs::EnsureS3Finalized();
```

vyasr (approved):
One small issue, but should be good to go once that is fixed.

Issue #13809, opened 2023-08-02
[BUG] Groupby-apply returns DataFrame rather than Series when taking the JIT path


shwina:
Depending on whether I choose `engine='jit'` or `engine='cudf'`, I get back a `DataFrame` rather than the expected `Series`:

```python
In [3]: df = cudf.DataFrame({'a': [1, 1, 1, 2, 2], 'b': [1, 2, 3, 4, 5]})

In [4]: df.groupby('a').apply(lambda group: group['a'].sum())
Out[4]:
   a  None
0  1     3
1  2     4

In [5]: df.groupby('a').apply(lambda group: group['a'].sum(), engine='cudf')
Out[5]:
a
1    3
2    4
dtype: int64
```

PR #13820, opened 2023-08-04
Return a Series from JIT GroupBy apply, rather than a DataFrame

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/13809

### DIFF
--- a/python/cudf/cudf/tests/test_groupby.py
+++ b/python/cudf/cudf/tests/test_groupby.py
@@ -395,7 +395,7 @@ def groupby_jit_data():
 
 
 def run_groupby_apply_jit_test(data, func, keys, *args):
-    expect_groupby_obj = data.to_pandas().groupby(keys, as_index=False)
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/tests/test_groupby.py
```diff
@@ -494,19 +494,20 @@ def func(df):
...
-    sorted = grouped._grouped()[3].to_pandas()
-    expect_vals = sorted["key1"].drop_duplicates().index
-    expect[None] = expect_vals
-
-    got = grouped.apply(func, engine="jit")
-    assert_eq(expect, got)
+    expect = groupby_jit_data.to_pandas().groupby("key1").apply(func)
+    got = groupby_jit_data.groupby("key1").apply(func, engine="jit")
+
+    if special_val is np.nan:
```
bdice, line 500: This sounds like a test that should be marked as `xfail`. What do you think?

galipremsagar (approved):

bdice (approved):

PR #13824, opened 2023-08-05
DataFrame with namedtuples uses ._field as column names

### LINKED ISSUE #13823: [BUG]: cudf.DataFrame does not map namedtuple field names to column names like pandas
**Describe the bug**
cudf.DataFrame does not map namedtuple field names to column names like pandas

Follow this guide http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports to craft a minimal bug report. This helps us reproduce the issue you're having and resolve the issue more quickly.

```
from collections import namedtuple
import cudf
import pandas

Point = namedtuple("Point", "x y")

pandas.DataFrame([Point(0, 0), Point(0, 3), (2, 3)])
#    x  y
# 0  0  0
# 1  0  3
# 2  2  3

cudf.DataFrame([Point(0, 0), Point(0, 3), (2, 3)])
#    0  1
# 0  0  0
# 1  0  3
# 2  2  3
```

**Expected behavior**
The cudf result should match the result

```
# Name                    Version                   Build  Channel
pandas                    1.5.3           py310h9b08913_1    conda-forge

# Name                    Version                   Build  Channel
cudf                      23.10.00a       cuda11_py310_230804_gd8bf9d2534_53    rapidsai-nightly
libcudf                   23.10.00a       cuda11_230803_g15cc501190_51    rapidsai-nightly
```

**Environment details**

**Additional context**

mroeschke:
## Description

Allow namedtuple's `_field` attribute to be mapped to DataFrame column labels like pandas

closes #13823

## Checklist

### DIFF
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -708,7 +708,10 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
                     )
                 else:
                     self._init_from_list_like(
-                        data, index=index, columns=columns
+                        data,
+                        index=index,
+                        columns=columns,
+                        nan_as_null=nan_as_null,
                     )
                 self._check_data_index_length_match()
             else:
@@ -827,7 +830,9 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
             self._data = self._data.select_by_label(columns)
 
     @_cudf_nvtx_annotate
-    def _init_from_list_like(self, data, index=None, columns=None):
+    def _init_from_list_like(
+        self, data, index=None, columns=None, nan_as_null=None
+    ):
         if index is None:
             index = RangeIndex(start=0, stop=len(data))
         else:
@@ -842,6 +847,21 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
         elif len(data) > 0 and isinstance(data[0], pd._libs.interval.Interval):
             data = DataFrame.from_pandas(pd.DataFrame(data))
             self._data = data._data
+        # namedtuple in a list
+        elif (
+            len(data) > 0
+            and columns is None
+            and isinstance(data[0], tuple)
+            and hasattr(data[0], "_fields")
+        ):
+            # pandas behavior is to use the fields from the first tuple
+            # as the column names
+            columns = data[0]._fields
+            values = itertools.zip_longest(*data)
+            data = dict(zip(columns, values))
+            self._init_from_dict_like(
+                data, index=index, nan_as_null=nan_as_null
+            )
         else:
             if any(
                 not isinstance(col, (abc.Iterable, abc.Sequence))
--- a/python/cudf/cudf/tests/test_dataframe.py
+++ b/python/cudf/cudf/tests/test_dataframe.py
@@ -10,7 +10,7 @@ import re
 import string
 import textwrap
 import warnings
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, namedtuple
 from copy import copy
 
 import cupy
@@ -10261,3 +10261,15 @@ def test_dataframe_constructor_unbounded_sequence():
 
     with pytest.raises(TypeError):
         cudf.DataFrame({"a": A()})
+
+
+def test_dataframe_constructor_from_namedtuple():
+    Point1 = namedtuple("Point1", ["a", "b", "c"])
+    Point2 = namedtuple("Point1", ["x", "y"])
+
+    data = [Point1(1, 2, 3), Point2(4, 5)]
+    idx = ["a", "b"]
+    gdf = cudf.DataFrame(data, index=idx)
+    pdf = pd.DataFrame(data, index=idx)
+
+    assert_eq(gdf, pdf)


python/cudf/cudf/core/dataframe.py
```diff
@@ -842,6 +847,21 @@ def _init_from_list_like(self, data, index=None, columns=None):
...
+            and hasattr(data[0], "_fields")
+        ):
+            # pandas behavior is to use the fields from the first tuple
+            # as the column names
+            columns = data[0]._fields
+            values = itertools.zip_longest(*data)
+            data = dict(zip(columns, values))
+            self._init_from_dict_like(
+                data, index=index, nan_as_null=nan_as_null
+            )
```
galipremsagar, line 864:
```suggestion
        if (
            len(data) > 0
            and columns is None
            and isinstance(data[0], tuple)
            and hasattr(data[0], "_fields")
        ):
            # pandas behavior is to use the fields from the first tuple
            # as the column names
            columns = data[0]._fields
```

Will just doing this work?

python/cudf/cudf/core/dataframe.py
```diff
@@ -842,6 +847,21 @@ def _init_from_list_like(self, data, index=None, columns=None):
...
+        # namedtuple in a list
+        elif (
+            len(data) > 0
+            and columns is None
+            and isinstance(data[0], tuple)
+            and hasattr(data[0], "_fields")
+        ):
+            # pandas behavior is to use the fields from the first tuple
+            # as the column names
+            columns = data[0]._fields
```
wence-, line 859: This seems like a bug (or at least, potentially confusing behaviour). If I pass a list of dicts with disjoint keys I get a column index that is the union of the keys with nans in all the missing slots.

OTOH, if I pass a list of named tuples they are just treated like normal tuples except for picking out the names of the columns. So no disjointness is applied (and the keys are kind of "meaningless").

python/cudf/cudf/core/dataframe.py
```diff
@@ -842,6 +847,21 @@ def _init_from_list_like(self, data, index=None, columns=None):
...
+            values = itertools.zip_longest(*data)
```
wence-, line 860: This inserts `None` whenever a tuple is too short.

python/cudf/cudf/core/dataframe.py
```diff
@@ -842,6 +847,21 @@ def _init_from_list_like(self, data, index=None, columns=None):
...
+            data = dict(zip(columns, values))
```
wence-, line 861: And this silently drops columns if the first tuple is shorter than subsequent ones.

Pandas is non-symmetric wrt the handling of named tuples of different lengths. If the first tuple has fewer fields than subsequent ones, we get a `ValueError`, if the first tuple has the same, or more, fields than subsequent ones we get `None`.

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -10261,3 +10261,15 @@ def __getitem__(self, key):
 
     with pytest.raises(TypeError):
         cudf.DataFrame({"a": A()})
+
+
+def test_dataframe_constructor_from_namedtuple():
+    Point1 = namedtuple("Point1", ["a", "b", "c"])
+    Point2 = namedtuple("Point1", ["x", "y"])
+
+    data = [Point1(1, 2, 3), Point2(4, 5)]
```
wence-, line 10270: Can you test with the points the other way round in the list? I think it will not work.

wence- (changes requested):
I think this doesn't mimic pandas behaviour exactly, and has some edge cases where it silently drops data.

python/cudf/cudf/core/dataframe.py
mroeschke: I was curious about this too; since this code path just assigns `self.columns = columns` I wasn't sure if it was okay if `columns` was a tuple or needed to be a different type

python/cudf/cudf/core/dataframe.py
mroeschke: Yeah this partially seems intentional when namedtuple support was implemented as it's documented in https://pandas.pydata.org/pandas-docs/stable/user_guide/dsintro.html#from-a-list-of-namedtuples

python/cudf/cudf/tests/test_dataframe.py
mroeschke: I can add a test that this fails like pandas does when the points are the other way around

```
In [1]: import numpy as np, pandas as pd

In [2]: from collections import namedtuple

In [3]:     Point1 = namedtuple("Point1", ["a", "b", "c"])
   ...:     Point2 = namedtuple("Point1", ["x", "y"])

In [4]: pd.DataFrame([Point2(4, 5), Point1(1, 2, 3)])
ValueError: 2 columns passed, passed data had 3 columns
```

python/cudf/cudf/core/dataframe.py
mroeschke: Okay this looks to work and produces a column Index too so I changed it to this.

galipremsagar (approved):

python/cudf/cudf/tests/test_dataframe.py
wence-: Thanks.

PR #13829, opened 2023-08-07
Fix cuFile I/O factories

vuule:
## Description
Factories that instantiate `cufile_input_impl`/`cufile_output_impl` objects did not return the created objects, and instead always returned `nullptr`.
This PR fixes this bug, which made the "GDS" and "OFF" cuFile use policies effectively the same.
The default policy value, "KVIKIO", should be unaffected.

## Checklist


ttnghia (approved):
Wow, such a nice catch!

vuule:
> Wow, such a nice catch!

We can thank @ayushdg for finding the issue. I just caused it :D

karthikeyann (approved):

Issue #13827, opened 2023-08-07
[BUG] `np.all` doesn't perform logical AND over all dimensions when `axis=None`


galipremsagar:
**Describe the bug**
`np.all` ufunc seems to be requiring to perform a logical AND reduction over all dimensions, but however in cudf the ufunc seems to be performing `all` operation only once when `axis=None`.

**Steps/Code to reproduce bug**
```python
In [1]: import numpy as np

In [2]: import cudf

In [3]: df = cudf.DataFrame({'a': [3, 2, 3, 4], 'b': [7, 0, 10, 10]})

In [4]: df
Out[4]: 
   a   b
0  3   7
1  2   0
2  3  10
3  4  10

In [5]: df.all()
Out[5]: 
a     True
b    False
dtype: bool

In [6]: df.all().all()
Out[6]: False

In [7]: np.all(df)
Out[7]: 
a     True
b    False
dtype: bool

In [8]: np.all(df.to_pandas())
Out[8]: False

In [9]: np.all(df.to_pandas(), axis=0)
Out[9]: 
a     True
b    False
dtype: bool

In [10]: np.all(df.to_pandas(), axis=1)
Out[10]: 
0     True
1    False
2     True
3     True
dtype: bool

In [11]: np.all(df, axis=0)
Out[11]: 
a     True
b    False
dtype: bool

In [12]: np.all(df, axis=1)
Out[12]: 
0     True
1    False
2     True
3     True
dtype: bool
```
**Expected behavior**
```python
In [8]: np.all(df)
Out[8]: False
```

**Environment overview (please complete the following information)**
 - Environment location: [Bare-metal]
 - Method of cuDF install: [from source]

**Environment details**
Please run and paste the output of the `cudf/print_env.sh` script here, to gather any other relevant environment details

**Additional context**

PR #13831, opened 2023-08-08
Fix `any`, `all` reduction behavior for `axis=None` and warn for other reductions

galipremsagar:
## Description
Fixes: #13827 

This PR:

- [x] Fixes `axis=None` behavior for `any` & `all` reductions.
- [x] Introduces `FutureWarning` for upcoming change in behavior for the rest of the reductions since some of the reductions are only updated in pandas-2.0 and the rest would be updated in pandas-3.0.
- [x] Fixed numpy array function inconsistency because of mismatching default for `axis` parameter.

## Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -1333,6 +1333,11 @@ def __array_function__(self, func, types, args, kwargs):
             return NotImplemented
 
         try:
+            if func.__name__ in {"any", "all"}:
+                # Numpy default for `axis` is
```
bdice, line 1337:
```suggestion
                # NumPy default for `axis` is
```

python/cudf/cudf/core/frame.py
```diff
@@ -2089,7 +2090,20 @@ def product(
         b    5040
         dtype: int64
         """
-        axis = self._get_axis_from_axis_arg(axis)
+        if axis in {no_default, None}:
+            if axis is None:
+                # Do not remove until pandas 3.0 support is added.
+                warnings.warn(
```
bdice, line 2096: Why doesn't this run through the same logic as the other `_reduce` calls, and get the warning from `_reduce`?

Could we instead replace the logic below, and write the cupy dispatch as `"product" if (axis == 0 or axis is None) else "prod"`?

python/cudf/cudf/tests/test_array_function.py
```diff
@@ -67,6 +67,8 @@ def test_array_func_cudf_series(np_ar, func):
         lambda x: np.sum(x, axis=0),
```
bdice, line 67: Should we explicitly test `product` (with various `axis` options?), since its dispatches to CuPy are special? Does that need to be tested with NumPy dispatches as well as "normal" pandas calls?

python/cudf/cudf/core/frame.py
galipremsagar: Yup, we can rely on `_reduce`. I also simplified the `prod` vs `product` logic selection.

python/cudf/cudf/tests/test_array_function.py
galipremsagar: Yup, added tests for this.

python/cudf/cudf/core/dataframe.py
```diff
@@ -5879,12 +5903,16 @@ def _reduce(
...
+            if axis == 2:
+                return getattr(as_column(result), op)(**kwargs)
+            else:
+                return Series._from_data(
+                    {None: result}, as_index(source._data.names)
+                )
         elif axis == 1:
             return source._apply_cupy_method_axis_1(op, **kwargs)
+        else:
+            raise ValueError(f"Incorrect value of {axis=} received for {op}")
```
bdice, line 5915:
```suggestion
            raise ValueError(f"Invalid value of {axis=} received for {op}")
```

bdice (approved):

Issue #13836, opened 2023-08-08
[BUG] Binary op between `Index` and `Series` should result in `Series`


galipremsagar:
**Describe the bug**
Binary op between two different `SingleColumnFrame` objects should result in `Series` results

**Steps/Code to reproduce bug**
```python

In [1]: import cudf

In [2]: i = cudf.Index([1, 2, 3], dtype='datetime64[ns]')

In [3]: i
Out[3]: 
DatetimeIndex(['1970-01-01 00:00:00.000000001',
               '1970-01-01 00:00:00.000000002',
               '1970-01-01 00:00:00.000000003'],
              dtype='datetime64[ns]')

In [4]: s = cudf.Series([1, 2, 3], dtype='timedelta64[ns]')

In [5]: i + s
Out[5]: 
DatetimeIndex(['1970-01-01 00:00:00.000000002',
               '1970-01-01 00:00:00.000000004',
               '1970-01-01 00:00:00.000000006'],
              dtype='datetime64[ns]')

In [6]: i.to_pandas() + s.to_pandas()
Out[6]: 
0   1970-01-01 00:00:00.000000002
1   1970-01-01 00:00:00.000000004
2   1970-01-01 00:00:00.000000006
dtype: datetime64[ns]

In [7]: i == s
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[7], line 1
----> 1 i == s

File /nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/cudf/core/mixins/mixin_factory.py:11, in _partialmethod.<locals>.wrapper(self, *args2, **kwargs2)
     10 def wrapper(self, *args2, **kwargs2):
---> 11     return method(self, *args1, *args2, **kwargs1, **kwargs2)

File /nvme/0/pgali/envs/cudfdev/lib/python3.10/site-packages/cudf/core/index.py:1032, in GenericIndex._binaryop(self, other, op, fill_value, *args, **kwargs)
   1030 if operands is NotImplemented:
   1031     return NotImplemented
-> 1032 ret = _index_from_data(self._colwise_binop(operands, op))
   1034 # pandas returns numpy arrays when the outputs are boolean. We
   1035 # explicitly _do not_ use isinstance here: we want only boolean
   1036 # GenericIndexes, not dtype-specific subclasses.
   1037 if type(ret) is GenericIndex and ret.dtype.kind == "b":

[...]

PR #13842, opened 2023-08-09
Fix binary operations between `Series` and `Index`

galipremsagar:
## Description
Fixes: #13836 

This PR fixes binary operations between `Series` & `Index` (and vice-versa) by properly handling the return type and the names being assigned to the return results. This PR also adds an early exit that will return `False` for `eq` operations when the other column being compared is not of the same type.

## Checklist


python/cudf/cudf/core/series.py
```diff
@@ -1505,9 +1505,14 @@ def _make_operands_and_index_for_binop(
...
             lhs, other = _align_indices([self, other], allow_non_unique=True)
-            can_use_self_column_name = self.name == other.name
         else:
             lhs = self
+
+        try:
+            can_use_self_column_name = cudf.utils.utils._is_same_name(
+                self.name, other.name
+            )
+        except Exception:
```
bdice, line 1515: This can only raise `AttributeError` from what I can tell.
```suggestion
        except AttributeError:
```

python/cudf/cudf/core/column/column.py
```diff
@@ -386,6 +386,25 @@ def _get_mask_as_column(self) -> ColumnBase:
             self.base_mask, self.offset, self.offset + len(self)
         )
 
+    def _early_exit_eq_op(self, other):
```
bdice, line 389: It seems like a better name would be `_create_all_false_column_with_nulls` or something to describe the nature of the output.

Also, should this be a method of a Column? I think it could be better as a free function in a utility namespace.

bdice (changes requested):

python/cudf/cudf/core/column/timedelta.py
```diff
@@ -202,6 +202,11 @@ def _binaryop(self, other: ColumnBinaryOperand, op: str) -> ColumnBase:
         elif other.dtype.kind in {"f", "i", "u"}:
             if op in {"__mul__", "__mod__", "__truediv__", "__floordiv__"}:
                 out_dtype = self.dtype
+            elif op in {"__eq__", "NULL_EQUALS"}:
```
mroeschke, line 205: What about `__ne__`?

Generally pandas behavior for an invalid comparison between types is

1. Return array of `True`s for ne
2. Return array of `False`s for eq
3. TypeError for other ops

galipremsagar: Thanks @mroeschke for catching this. I got it handled now.

python/cudf/cudf/core/column/column.py
galipremsagar: Separated it out as a free function utility and also changed the name.

python/cudf/cudf/utils/utils.py
```diff
@@ -384,3 +384,28 @@ def wrapper(self):
         return fn(self)
 
     return wrapper
+
+
+def _is_same_name(left_name, right_name):
+    # Internal utility to compare if two names are same.
+    return (left_name is right_name) or (left_name == right_name)
```
bdice, line 391: Just for my understanding — when would an equality check not be sufficient here, and the identity check required? Is it for non-string names?

bdice (approved):
Minor questions/comments but LGTM.

python/cudf/cudf/utils/utils.py
galipremsagar: Yup, `np.nan == np.nan` result to `False`. Thus an equality check of `names` would just not be sufficient, hence the identity check.

python/cudf/cudf/core/index.py
```diff
@@ -1029,12 +1030,24 @@ def _binaryop(
         operands = self._make_operands_for_binop(other, fill_value, reflect)
         if operands is NotImplemented:
             return NotImplemented
-        ret = _index_from_data(self._colwise_binop(operands, op))
+        result = self._colwise_binop(operands, op)
```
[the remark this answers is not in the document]
galipremsagar: Done 👍

python/cudf/cudf/utils/utils.py
bdice: Welllll if it’s for nans, is `np.nan` supposed to be the same as `float("NaN")` identity-wise? Should that be handled by checking isnan for both inputs? If this is specific to NaN we might want to reconsider how this is written.

galipremsagar: `nan` is one example of it. `NA` is another:
```
In [9]: pd.NA == pd.NA
Out[9]: <NA>

In [10]: pd.NA is pd.NA
Out[10]: True
```
I think `np.nan` & `float("NaN")` are two different values when it comes to name equality comparisons in pandas. AFAIK the pytests I've come across treat them as different values i.e., they just do an identity check & equality check.

bdice: Sounds good. If we should have specific expectations, it’d be good to have tests for a variety of values like those.

galipremsagar: I'm wrong. 
```python

In [25]: s = pd.Series([1, 2, 3, None], name=np.nan)

In [26]: p = pd.Series([1, 2, 3, None], name=float("NaN"))

In [27]: s + p
Out[27]: 
0    2.0
1    4.0
2    6.0
3    NaN
Name: nan, dtype: float64

In [28]: (s + p).name
Out[28]: nan

In [29]: type((s + p).name)
Out[29]: float

In [30]: p.name = "a"

In [31]: p
Out[31]: 
0    1.0
1    2.0
2    3.0
3    NaN
Name: a, dtype: float64

In [32]: s + p
Out[32]: 
0    2.0
1    4.0
2    6.0
3    NaN
dtype: float64
```

python/cudf/cudf/core/column/datetime.py
```diff
@@ -424,12 +424,8 @@ def _binaryop(self, other: ColumnBinaryOperand, op: str) -> ColumnBase:
...
-            "__eq__",
-            "NULL_EQUALS",
-        }:
-            out_dtype = cudf.dtype(np.bool_)
-        elif (
+
+        if (
             op
             in {
                 "__ne__",
```
mroeschke, line 431: Is this `__ne__` redundant with the one you added below?

mroeschke: Thanks 👍

mroeschke (approved):
LGTM

python/cudf/cudf/utils/utils.py
bdice: 😅🥲

galipremsagar: Added name handling. Had to open this can of worms by adding extensive name coverage in the test. 😅

bdice (approved):

PR #13844, opened 2023-08-09
Simplify implementation of interval_range() and fix behaviour for floating `freq`

### LINKED ISSUE #13843: [BUG] interval_range produces incorrect results when `freq=` is a float
There are a number of bugs in our implementation of `interval_range` causing various issues when `freq=` is a float:

```python
>>> cudf.interval_range(1, periods=2, freq=0.5)
ValueError: range() arg 3 must not be zero

>>> pd.interval_range(1, periods=2, freq=1.5)
 IntervalIndex([(1.0, 2.5], (2.5, 4.0]], dtype='interval[float64, right]')
>>> cudf.interval_range(1, periods=2, freq=1.5)  # cudf produces the wrong result
IntervalIndex([(1.0, 2.5], (2.5, 4.0], (4.0, 5.5]], dtype='interval')

>>> cudf.interval_range(1, periods=2, freq=2.5)
ArrowInvalid: Struct child array #1 has length smaller than expected for struct array (2 < 3)
```

I think our i[mplementation of `interval_range` can greatly be simplified](https://github.com/rapidsai/cudf/blob/e8df03754021e3decfc6640b58bd7a0770b0c230/python/cudf/cudf/core/index.py#L2945). We should also add tests to prevent regressions.

shwina:
Closes #13843 
Closes https://github.com/rapidsai/cudf/issues/13847

This PR simplifies the implementation of `interval_range()` and fixes a few different bugs in the process. It also moves all tests for interval indexes to `tests/indexes/test_interval.py`. Finally, while working on this PR, I ran into #13847; a fix for that is also included in this PR.

## Checklist


python/cudf/cudf/tests/indexes/test_interval.py
```diff
@@ -16,3 +19,273 @@ def test_interval_to_arrow():
...
+        (0.0, 1.0, None, 5),
+        # (0.0, 1.0, 0.2, None), # Pandas returns only 4 intervals here
+        (None, 1.0, 0.2, 5),
+        # (0.0, 1.0, 0.1, None), # Pandas returns the wrong result here
+        (0.0, 1.0, None, 10),
+        (0.0, None, 0.25, 4),
+        (1.0, None, 2.5, 2),
+    ],
+)
+def test_interval_range_floating(start, stop, freq, periods):
```
shwina, line 286: These are the only new tests introduced in this PR. The other tests have been moved to this file from elsewhere.

python/cudf/cudf/core/index.py
```diff
@@ -2991,89 +2990,56 @@ def interval_range(
...
+
+    start = cudf.Scalar(start) if start is not None else start
+    end = cudf.Scalar(end) if end is not None else end
+    periods = cudf.Scalar(int(periods)) if periods is not None else periods
+    freq = cudf.Scalar(freq) if freq is not None else freq
+
+    if start is None:
+        start = end - freq * periods
+    elif freq is None:
+        quotient, remainder = divmod((end - start).value, periods.value)
```
shwina, line 3014: `divmod` seems not to work with `cudf.Scalar`

python/cudf/cudf/tests/test_interval.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
```
galipremsagar, line 1:
```suggestion
# Copyright (c) 2020-2022, NVIDIA CORPORATION.
```

Undo copyright update since no changes in this file?

galipremsagar (approved):
LGTM, minor copyright comment.

python/cudf/cudf/tests/indexes/test_interval.py
```diff
@@ -16,3 +19,273 @@ def test_interval_to_arrow():
...
+
+    assert_eq(pindex, gindex)
+
+
+@pytest.mark.parametrize(
+    "start, stop, freq, periods",
+    [
+        (0.0, None, 0.2, 5),
+        (0.0, 1.0, None, 5),
+        # (0.0, 1.0, 0.2, None), # Pandas returns only 4 intervals here
```
shwina, line 278: @galipremsagar what's the right way to handle tests that will eventually pass with Pandas 2.x?

https://github.com/pandas-dev/pandas/pull/54477

galipremsagar, line 278:
```suggestion
        pytest.param(
            (0.0, 1.0, 0.2, None),
            marks=pytest.mark.xfail(
                condition=not PANDAS_GE_210,
                reason="https://github.com/pandas-dev/pandas/pull/54477",
            ),
        )
```

We can use this

galipremsagar: Suggested a change below: https://github.com/rapidsai/cudf/pull/13844/files#r1290106880

galipremsagar: We might want to introduce `PANDAS_GE_210` variable in `cudf.core._compat.py` too

galipremsagar (approved):

python/cudf/cudf/core/index.py
wence-: Can you do all this manipulation on the host before scalarising at the end, that would have far fewer syncs.

python/cudf/cudf/core/index.py
```diff
@@ -2991,89 +2990,56 @@ def interval_range(
...
+            freq = (end - start) / periods
+        else:
+            freq = cudf.Scalar(int(quotient))
+    elif periods is None:
+        periods = cudf.Scalar(int((end - start) / freq))
+    elif end is None:
+        end = start + periods * freq
+
     if any(
         not _is_non_decimal_numeric_dtype(x.dtype) if x is not None else False
```
wence-, line 3025: At this point all args must be not None.

python/cudf/cudf/core/index.py
```diff
@@ -2991,89 +2990,56 @@ def interval_range(
...
+    start = start.astype(common_dtype)
+    freq = freq.astype(common_dtype)
+
+    bin_edges = sequence(
+        size=periods + 1,
+        init=start.device_value,
+        step=freq.device_value,
+    )
+    left_col = bin_edges.slice(0, len(bin_edges) - 1)
+    right_col = bin_edges.slice(1, len(bin_edges))
```
wence-, line 3041: You're doing `linspace` here with `arange`-like calls, are we sure the edge cases are handled correctly (specifically if `periods` has not been provided).

wence- (commented):
Does it make sense to copy all the scalars to device, only to manipulate and copy back. Would it be better do do all the preprocessing on the host and then move stuff once?

python/cudf/cudf/core/scalar.py
```diff
@@ -353,14 +353,17 @@ def _binaryop(self, other, op: str):
 
     def _dispatch_scalar_binop(self, other, op):
         if isinstance(other, Scalar):
-            other = other.value
+            rhs = other.value
+        else:
+            rhs = other
+        lhs = self.value
```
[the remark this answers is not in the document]
brandon-b-miller: I think we'll always need to come up to the host do do binaryops between scalars because there's no device side binaryop for the underlying device scalar to target in the first place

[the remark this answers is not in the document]
wence-: Ah, ok.

python/cudf/cudf/core/index.py
shwina: We have tests written already that pass `Scalar` objects as input :(

python/cudf/cudf/core/index.py
wence-: Ah ok, they are also wrong then:

```
In [10]: pd.interval_range(start=1, end=3.0, freq=0.5)
Out[10]: IntervalIndex([(1.0, 1.5], (1.5, 2.0], (2.0, 2.5], (2.5, 3.0]], dtype='interval[float64, right]')

In [11]: pd.interval_range(start=1, end=3.0, freq=0.1)
Out[11]: IntervalIndex([(1.0, 1.1055555555555556], (1.1055555555555556, 1.211111111111111], (1.211111111111111, 1.3166666666666667], (1.3166666666666667, 1.4222222222222223], (1.4222222222222223, 1.5277777777777777] ... (2.3722222222222222, 2.477777777777778], (2.477777777777778, 2.583333333333333], (2.583333333333333, 2.688888888888889], (2.688888888888889, 2.7944444444444443], (2.7944444444444443, 2.9]], dtype='interval[float64, right]')
```

wence-: Note how I asked for a frequency of 0.1 (so the intervals should be `(1, 1.1]` etc...)

wence-: Ah, because the implementation does loads of stuff with floats that suffers from rounding errors

python/cudf/cudf/core/index.py
brandon-b-miller: The scalar caching should save us here, no? There shouldn't be a sync until `device_value` is called later on. `.value` should just return the cached host value if the scalar was constructed from a host scalar

python/cudf/cudf/core/index.py
mroeschke: Yup should hopefully be fixed in 2.1 in pandas to just do an arange in these cases https://github.com/pandas-dev/pandas/pull/54477

python/cudf/cudf/core/index.py
shwina: Fix'd

PR #13846, opened 2023-08-10
Generate unified Python/C++ docs

### LINKED ISSUE #11481: [FEA] Use a single integrated documentation solution across components
Currently the [C++ documentation](https://docs.rapids.ai/api/libcudf/stable/) and [Python documentation](https://docs.rapids.ai/api/cudf/stable/) are managed and published completely separately. The Python documentation uses Sphinx, while the C++ documentation uses doxygen. Sphinx docs make it significantly easier to include documentation beyond API docs (e.g. user guides or detailed topic references), while doxygen is much more focused on API documentation alone. #11475 demonstrates how non-API documentation can be integrated with doxygen docs, but this approach is limited relative to the flexibility that Sphinx supports. Additionally, Sphinx styling is easier to modify due to the large number of available themes and the knobs that can be easily turned for them. It would be nice if all of our documentation for the different language libraries (as well as different components like developer docs and API docs) could be centralized and presented in a unified manner. 

We should consider migrating all of our documentation to use the new [OmniVerse documentation system](http://omniverse-docs.s3-website-us-east-1.amazonaws.com/repo_docs/0.10.0-pre/index.html). It provides a single, unified platform for building both C++ and Python documentation into a Sphinx document. It supports the exact sets of documentation that we already use (doxygen for C++ API docs and rST for Python API docs) while also making it easy to add all the extra pages that we might wish (and which already exist for the Python documentation). 

One oft-cited benefit of our current documentation layout is that it maintains an alignment with pandas documentation. This makes it easier for users to find the corresponding APIs between the two libraries. While migrating to the OmniVerse documentation system would be a great solution for unifying our documentation and providing a layout and style that is very on-brand for NVIDIA tooling, the different styling may cause some dissonance for readers. If we think this is a significant issue (although I don't anticipate this being the case) we could consider using [Breathe](https://breathe.readthedocs.io/en/latest/) directly in our Sphinx docs. Breathe is what allows Sphinx docs to talk to doxygen and parse those API docs (it's what the Omniverse documentation system uses under the hood), and we could leverage it directly in our existing Sphinx documentation. This approach would allow us to have a unified approach to documentation while still retaining the pandas-compatible style.

**Additional context**
Migrating all of our documentation -- whether to OmniVerse or to Breathe -- is a large change that will need to be synchronized across all of RAPIDS. It will affect both user- and developer-facing documentation, so the effects should be carefully considered. Moreover, we should expect that the combination of Breathe and the PyData Sphinx theme that we use will have some incompatibilities that will need to be addressed, and Breathe may affect formatting in surprising ways so we'll need to do a thorough review. As such, any effort to modify the cudf documentation in this manner should be viewed as a POC to be demoed to and discussed across all of RAPIDS before any changes are finalized.

One additional minor point: we need to make sure that whatever system we choose supports documentation in Cython files appropriately. This shouldn't be a problem for direct usage of Breathe, but I don't know enough about how the Omniverse documentation system works under the hood to be entirely certain that it doesn't make additional assumptions that we would need relaxed to support Cython docstrings.

vyasr:
## Description

This PR leverages [Breathe](https://breathe.readthedocs.io/en/latest/) to pull the cudf C++ API documentation into the python Sphinx docs build, generating a single unified build of the documentation that supports cross-linking between language libraries and also simplifies cross-linking from other libraries that wish to link here.

This PR also revealed numerous other issues with our doxygen docs. I've submitted those as separate patches to control the diff here, but it's worth noting that Sphinx is much louder with warnings than doxygen and will help us avoid many more issues with broken documentation than doxygen alone could.

Resolves #11481 

## Checklist


docs/cudf/source/conf.py
```diff
@@ -195,11 +279,13 @@
...
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3", None),
     "cupy": ("https://docs.cupy.dev/en/stable/", None),
+    "dlpack": ("https://dmlc.github.io/dlpack/latest/", None),
     "numpy": ("https://numpy.org/doc/stable", None),
-    "pyarrow": ("https://arrow.apache.org/docs/", None),
     "pandas": ("https://pandas.pydata.org/docs/", None),
+    "pyarrow": ("https://arrow.apache.org/docs/", None),
+    "python": ("https://docs.python.org/3", None),
+    "rmm": ("https://docs.rapids.ai/api/rmm/nightly/", None),
```
ajschmidt8, line 288: Does this line mean we can remove the Doxygen step that fetches the `rmm.tag` file below?

- https://github.com/rapidsai/cudf/blob/8dca25c782bbe239ed6e9b6317cc3a01b15a2b42/ci/build_docs.sh#L35

ajschmidt8 (commented):
Awesome! There's a few other changes we'll want to include with this PR:

- Remove these lines: https://github.com/rapidsai/cudf/blob/8dca25c782bbe239ed6e9b6317cc3a01b15a2b42/ci/build_docs.sh#L37-L38
- Turn off `GENERATE_HTML` in `Doxyfile`: https://github.com/rapidsai/cudf/blob/8dca25c782bbe239ed6e9b6317cc3a01b15a2b42/cpp/doxygen/Doxyfile#L1148
- Disable `libcudf` nightly on our docs site since it will be included with `cudf`: https://github.com/rapidsai/docs/blob/1b7875ac8d983bd13481e2d3f9e1dd27e0dd937c/_data/docs.yml#L174

docs/cudf/source/conf.py
vyasr: Yes, we don't need that anymore.

vyasr:
@ajschmidt8 good call, took care of those.

ajschmidt8 (approved):

shwina (approved):

cpp/doxygen/developer_guide/BENCHMARKING.md
```diff
@@ -23,7 +23,7 @@ benchmarks in `cpp/benchmarks/copying/scatter.cu`.
 In the interest of improving compile time, whenever possible, test source files should be `.cpp`
 files because `nvcc` is slower than `gcc` in compiling host code. Note that `thrust::device_vector`
 includes device code, and so must only be used in `.cu` files. `rmm::device_uvector`,
-`rmm::device_buffer` and the various `column_wrapper` types described in [Testing](TESTING.md)
+`rmm::device_buffer` and the various `column_wrapper` types described in [Testing](testing)
```
karthikeyann, line 26: problem with this notation is that while browsing the github repository, these links are broken. is there any workaround for this?

cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md
```diff
@@ -894,6 +889,7 @@ the "breaking" tag. This ensures that the "Breaking" section of the release note
 description of what has broken from the past release. Label pull requests that contain deprecations
 with the "non-breaking" tag.
 
+(error-handling)=
 
 # Error Handling {#errors}
```
karthikeyann, line 894: is `{#errors}` still required?

karthikeyann (commented):
if a new documented function or a new doxygen group is not added in one of these `.rst` files, will this workflow be able to find it? (& throw a warning or failure).
How?

cpp/doxygen/developer_guide/BENCHMARKING.md
vyasr: Unfortunately I don't see a way to keep all three (Sphinx, doxygen, and Github) views of this working. When we're finally fully happy with the Sphinx docs and disable the doxygen builds altogether, I'm happy to revisit this and see if there's a way to at least keep Sphinx + GH working, but I don't think we should try until we've removed doxygen HTML altogether.

cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md
vyasr: I think it's still allowing some of the links on GH to work. Having GH links work on a best-effort basis seems fine to me until we remove doxygen HTML.

vyasr:
In the interest of moving forward in more incremental steps, I've put back doxygen HTML generation, so this PR won't disable those docs yet. We can do that at a later date when we're happier with the output of the Sphinx C++ documentation. Merging this PR in closer to its current state will allow us to iterate and improve more easily than having to perfect the appearance in this PR before removing the doxygen HTML.

karthikeyann (approved):

davidwendt (approved):
Approving C++ changes

PR #13848, opened 2023-08-10
Expand statistics support in ORC writer

### LINKED ISSUE #7087: [FEA] Support for missing ORC column statistics
The column statistics encoding in the writer is missing support for a few fields:

- [ ] hasNull;
- [ ] sum field of doubleStatistics;
- [x] sum field of stringStatistics;
- [ ] decimalStatistics;

Also, the ProtobufReader does not support bool fields (needed to read the hasNull field without a Python API).

### LINKED ISSUE #13793: [BUG] ORC statistics are wrong when a double column is all NULL.
**Describe the bug**
Report "Skipping ORC PPD" error when reading a ORC file with a double column is all NULL.
The statistic in the ORC file is wrong, refer to the following sections.
Note the query result is correct, but the PPD was skipped due to this error.

Generate a ORC file on GPU:
```cpp
TEST_F(OrcWriterTest, TestAllNulls)
{
  bool mask1[] = {false, false, false};  // all are null
  
  std::vector<char const*> h_strings1{"1", "2", "3"};
  str_col strings1(h_strings1.begin(), h_strings1.end());
  
  std::vector<double> hd{1.1, 2.2, 3.3};
  float64_col d2(hd.begin(), hd.end(), mask1);  // values of this double column are all null

  cudf::table_view expected({strings1, d2});

  cudf::io::table_input_metadata expected_metadata(expected);
  expected_metadata.column_metadata[0].set_name("c1");
  expected_metadata.column_metadata[1].set_name("c2");

  auto filepath = "/tmp/TestAllNulls.orc";
  cudf::io::orc_writer_options out_opts =
    cudf::io::orc_writer_options::builder(cudf::io::sink_info{filepath}, expected)
      .metadata(expected_metadata);
  cudf::io::write_orc(out_opts);
}
```

Read the ORC file from Spark.
```
SPARK_HOME/bin/pyspark 
df = spark.read.orc("/path/to/TestAllNulls.orc")
df.createOrReplaceTempView("tab")
spark.sql("select * from tab").show()
+---+----+
| c1|  c2|
+---+----+
|  1|null|
|  2|null|
|  3|null|
+---+----+

spark.sql("select * from tab where c2 is NOT NULL").show()
23/08/01 14:14:08 WARN RecordReaderImpl: ClassCastException when evaluating predicate. Skipping ORC PPD. Stats: numberOfValues: 0
 Predicate: (IS_NULL c2)
java.lang.ClassCastException: org.apache.orc.impl.ColumnStatisticsImpl cannot be cast to org.apache.orc.DoubleColumnStatistics
	at org.apache.orc.impl.RecordReaderImpl.evaluatePredicateProto(RecordReaderImpl.java:655)
	at org.apache.orc.impl.RecordReaderImpl$SargApplier.pickRowGroups(RecordReaderImpl.java:1144)
	at org.apache.orc.impl.RecordReaderImpl.pickRowGroups(RecordReaderImpl.java:1219)
	at org.apache.orc.impl.RecordReaderImpl.readStripe(RecordReaderImpl.java:1239)
	at org.apache.orc.impl.RecordReaderImpl.advanceStripe(RecordReaderImpl.java:1291)
	at org.apache.orc.impl.RecordReaderImpl.advanceToNextRow(RecordReaderImpl.java:1334)
	at org.apache.orc.impl.RecordReaderImpl.<init>(RecordReaderImpl.java:355)
	at org.apache.orc.impl.ReaderImpl.rows(ReaderImpl.java:879)
	at org.apache.spark.sql.execution.datasources.orc.OrcColumnarBatchReader.initialize(OrcColumnarBatchReader.java:130)
	at org.apache.spark.sql.execution.datasources.orc.OrcFileFormat.$anonfun$buildReaderWithPartitionValues$1(OrcFileFormat.scala:185)
	at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.org$apache$spark$sql$execution$datasources$FileScanRDD$$anon$$readCurrentFile(FileScanRDD.scala:209)
	at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.nextIterator(FileScanRDD.scala:270)
	at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.hasNext(FileScanRDD.scala:116)
	at org.apache.spark.sql.execution.FileSourceScanExec$$anon$1.hasNext(DataSourceScanExec.scala:553)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.columnartorow_nextBatch_0$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$1.hasNext(WholeStageCodegenExec.scala:760)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$getByteArrayRdd$1(SparkPlan.scala:364)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2(RDD.scala:890)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2$adapted(RDD.scala:890)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:365)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:329)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:90)
	at org.apache.spark.scheduler.Task.run(Task.scala:136)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:548)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1504)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:551)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
+---+---+ 
| c1| c2|
+---+---+   
+---+---+   
```

**Expected behavior**
Fix "Skipping ORC PPD" warning.
Check other types besides the double type.

**Environment details**
cuDF 23.08 branch  
Spark 3.3.0
orc-core-1.7.4.jar

**Additional context**
The error is:
```
java.lang.ClassCastException: org.apache.orc.impl.ColumnStatisticsImpl cannot be cast to org.apache.orc.DoubleColumnStatistics
```
Seems the ORC file does not contain `DoubleColumnStatistics`, so by default it's a `ColumnStatisticsImpl`.

### LINKED ISSUE #13899: [BUG] ORC write does not include full timestamp metrics which causes spark to not do predicate push down
**Describe the bug**
I am not totally sure if this is a bug or a new feature, but it ends up being a performance problem for Spark.

ORC statistics for timestamps are confusing. They include a minimum, maximum, minimum UTC, maximum UTC, and then min and max nanoseconds.

CUDF only sets minimum and maximum UTC. Spark and java ORC by default use only minimum and maximum +min/max nanos for predicate push down.  This results in predicate push down not working at all for timestamps in ORC.

https://github.com/NVIDIA/spark-rapids/pull/9068

The Spark CPU writer fills in all of the fields. It would really be great if CUDF could do the same. The nano seconds is probably not that important, but timestamp really is important.

Write a file out with CUDF in ORC. Try to read the data in with spark along with a filter where the value is outside of the range for any value in a timestamp column.

This will result in all of the data being read when CUDF wrote the file, but if the CPU wrote the file none of it would have been read.

**Expected behavior**
Predicate push down works no matter who wrote the data.

vuule:
## Description
Closes #7087, closes #13793, closes #13899

This PR adds support for several cases and statistics types:
- sum statistics are included even when all elements are null (no minmax);
- sum statistics are included in double stats;
- minimum/maximum and minimumNanos/maximumNanos are included in timestamp stats;
- hasNull field is written for all columns.
- decimal statistics

Added tests for all supported stats.

## Checklist


cpp/src/io/orc/stats_enc.cu
```diff
@@ -126,9 +126,6 @@ struct stats_state_s {
   statistics_chunk chunk;
   statistics_merge_group group;
   statistics_dtype stats_dtype;  //!< Statistics data type for this column
-  // ORC stats
-  uint64_t numberOfValues;
-  uint8_t hasNull;
```
vuule, line 131: was unused

python/cudf/cudf/tests/test_orc.py
```diff
@@ -633,16 +633,19 @@ def test_orc_write_statistics(tmpdir, datadir, nrows, stats_freq):
     for col in gdf:
         if "minimum" in file_stats[0][col]:
             stats_min = file_stats[0][col]["minimum"]
-            actual_min = gdf[col].min()
-            assert normalized_equals(actual_min, stats_min)
+            if stats_min is not None:
```
wence-, line 636: Under what circumstances does `read_orc_statistics` now return `None` in these slots when it didn't before?

wence- (approved):
One query about the python test, but approving python changes nonetheless

python/cudf/cudf/tests/test_orc.py
vuule: Great question!
The change in behavior is when a column only contains nulls. Previously we did not return any statistics for such column so we would not perform this comparison in the test. This PR changes the behavior so that statistics containing only the sum are included if there are not valid elements. So now the test need to correctly check "partial" statistics, i.e. min and max are not present, but sum is.

revans2 (approved):
From what I can see it looks correct, but I am not an ORC stats expert yet.

cpp/src/io/statistics/typed_statistics_chunk.cuh
```diff
@@ -244,9 +244,9 @@ get_untyped_chunk(typed_statistics_chunk<T, include_aggregate> const& chunk)
   stat.null_count = chunk.null_count;
   stat.has_minmax = chunk.has_minmax;
   stat.has_sum    = [&]() {
-    if (!chunk.has_minmax) return false;
     // invalidate the sum if overflow or underflow is possible
     if constexpr (std::is_floating_point_v<E> or std::is_integral_v<E>) {
+      if (!chunk.has_minmax) { return true; }
```
karthikeyann, line 249: is this change right?
For the same condition, return type is inverted.

karthikeyann (commented):
cpp review.

cpp/src/io/statistics/typed_statistics_chunk.cuh
vuule: This part is meant to disable the sum when there's a chance of overflow. We had the old condtion because we treated the statistics parts as indivisible - either return both the sum and min/max, or none. This does not really need to be the case. So, now, we return the sum even without min/max. The new logic is - if there's no minmax, we have no valid elements and the sum can be returned (as zero).

vuule:
Converted back to draft since there are further requests to expand the statistics in ORC that I can add to the PR.

cpp/src/io/orc/stats_enc.cu
```diff
@@ -87,21 +89,30 @@ __global__ void __launch_bounds__(block_size, 1)
         case dtype_int8:
         case dtype_int16:
         case dtype_int32:
-        case dtype_date32:
         case dtype_int64:
-        case dtype_timestamp64:
           stats_len = pb_fldlen_common + pb_fld_hdrlen + 3 * (pb_fld_hdrlen + pb_fldlen_int64);
           break;
+        case dtype_date32:
+          stats_len = pb_fldlen_common + pb_fld_hdrlen + 2 * (pb_fld_hdrlen + pb_fldlen_int64);
```
vuule, line 96: date statistics don't have the sum, used to be wrongly grouped with ints

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1858,8 +1858,8 @@ __device__ std::pair<void const*, uint32_t> get_extremum(statistics_val const* s
     }
     case dtype_int64:
     case dtype_timestamp64:
-    case dtype_float64:
-    case dtype_decimal64: return {stats_val, sizeof(int64_t)};
+    case dtype_float64: return {stats_val, sizeof(int64_t)};
```
vyasr, line 1861: So before decimal64 was being treated like int/float, and now it's instead being treated like decimal128?

cpp/src/io/statistics/statistics_type_identification.cuh
```diff
@@ -125,7 +125,7 @@ class extrema_type {
 
   using non_arithmetic_extrema_type = typename std::conditional_t<
     cudf::is_fixed_point<T>() or cudf::is_duration<T>() or cudf::is_timestamp<T>(),
-    typename std::conditional_t<std::is_same_v<T, numeric::decimal128>, __int128_t, int64_t>,
+    typename std::conditional_t<cudf::is_fixed_point<T>(), __int128_t, int64_t>,
```
vyasr, line 128: I assume this is related to the change in page_enc.cu where decimal64 is now treated the same as decimal128? Do we need any logic to cast down results for 64 bit decimals vs 128?

cpp/src/io/orc/stats_enc.cu
```diff
@@ -87,21 +89,32 @@ __global__ void __launch_bounds__(block_size, 1)
...
-          stats_len = pb_fldlen_common + pb_fld_hdrlen16 + 3 * (pb_fld_hdrlen + pb_fldlen_decimal);
-          break;
+        case dtype_decimal128: {
+          auto const scale    = groups[idx].col_dtype.scale();
+          auto const min_size = fixed_point_string_size(chunks[idx].min_value.d128_val, scale);
+          auto const max_size = fixed_point_string_size(chunks[idx].max_value.d128_val, scale);
+          auto const sum_size = fixed_point_string_size(chunks[idx].sum.d128_val, scale);
+          // common + total field length + encoded string lengths + strings
+          stats_len = pb_fldlen_common + pb_fld_hdrlen32 + 3 * (pb_fld_hdrlen + pb_fld_hdrlen32) +
+                      min_size + max_size + sum_size;
```
vyasr, line 114: So the new decimal statistics are min, max, and sum, and now we're reserving sufficient new space for them?

cpp/src/io/orc/stats_enc.cu
```diff
@@ -186,6 +205,15 @@ __device__ inline uint8_t* pb_put_fixed64(uint8_t* p, uint32_t id, void const* r
   return p + 9;
 }
 
+// Splits a nanosecond timestamp into milliseconds and nanoseconds
+__device__ std::pair<int64_t, int32_t> split_nanosecond_timestamp(int64_t nano_count)
```
vyasr, line 209: Why are the milliseconds encoded as 64 bit while nanoseconds are 32 bit? Is it because >1e6 ns adds to the ms, whereas the ms can grow unbounded?

vyasr (approved):
Everything seems solid, but I don't have much background on this code. I left a few questions but they're mostly for my benefit.

cpp/src/io/orc/stats_enc.cu
vuule: That's correct. The nanoseconds part is in [0, 999999] range. 
I'm open to suggestions to improve naming/comment here, I also wasn't 100% happy with clarity.

cpp/src/io/orc/stats_enc.cu
vuule: Yes, the previous computation of `stats_len` was basically unused since we did not write the stats we left space for.
The `case dtype_string:` case below has the similar logic as it also stores strings. The difference is that the string lengths are known from the column, and the sum is a number, not a string.

cpp/src/io/parquet/page_enc.cu
vuule: Yes. It used to use int64, now it uses int128.
the specs specify (as they do) that sum of int columns should be left out if it overflows int64_t. So we use this type and check for possible overflow.
Decimal sum does not have this limitation (saved as string, right) so we want to use the largest possible type.
Now that I'm going over this again, we could have kept the dec64 min/max at int64 and only used int128 for the sum. I don't think this detail is impactful and I like the consistency between min/max and sum for decimal types with the currently used types.

cpp/src/io/statistics/statistics_type_identification.cuh
vuule: Yup, using int128 for all decimal types. 
We cast elements up to int128 and never cast down, AFAICT. The resulting number is converted to a string. Let me know if I misunderstood the question.

cpp/tests/io/orc_test.cpp
```diff
@@ -1000,47 +1010,77 @@ TEST_F(OrcStatisticsTest, Basic)
 
   auto const stats = cudf::io::read_parsed_orc_statistics(cudf::io::source_info{filepath});
 
-  auto const expected_column_names =
-    std::vector<std::string>{"", "_col0", "_col1", "_col2", "_col3", "_col4"};
+  auto expected_column_names = std::vector<std::string>{""};
+  std::generate_n(std::back_inserter(expected_column_names), expected.num_columns(), []() mutable {
+    static int starting_index = 0;
+    return "_col" + std::to_string(starting_index++);
+  });
```
karthikeyann, line 1017: nit: 
is static required here?
how about?
```suggestion
  std::generate_n(std::back_inserter(expected_column_names), expected.num_columns(), [starting_index = 0]() mutable {
    return "_col" + std::to_string(starting_index++);
  });
```

cpp/include/cudf/strings/detail/convert/fixed_point_to_string.cuh
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @param value The value of the fixed point number
+ * @param scale The scale of the fixed point number
+ * @param out_ptr The pointer to the output string
+ */
+__device__ inline void fixed_point_to_string(__int128_t const& value, int32_t scale, char* out_ptr)
+{
+  if (scale >= 0) {
+    out_ptr += integer_to_string(value, out_ptr);
+    thrust::generate_n(thrust::seq, out_ptr, scale, []() { return '0'; });  // add zeros
```
karthikeyann, line 58: 👍

cpp/src/io/orc/stats_enc.cu
```diff
@@ -338,12 +404,22 @@ __global__ void __launch_bounds__(encode_threads_per_block)
...
+        //  optional int32 minimumNanos = 5; // lower 6 TS digits for min/max to achieve nanosecond
+        //  precision optional int32 maximumNanos = 6;
         // }
         if (s->chunk.has_minmax) {
           cur[0] = 9 * 8 + ProtofType::FIXEDLEN;
           cur += 2;
-          cur          = pb_put_int(cur, 3, s->chunk.min_value.i_val);  // minimumUtc
-          cur          = pb_put_int(cur, 4, s->chunk.max_value.i_val);  // maximumUtc
+          auto [min_ms, min_ns_remainder] = split_nanosecond_timestamp(s->chunk.min_value.i_val);
+          auto [max_ms, max_ns_remainder] = split_nanosecond_timestamp(s->chunk.max_value.i_val);
```
karthikeyann, line 414: nit: `auto const`

karthikeyann (approved):
Looks good to me.
very minor nitpicks & idiom.

PR #13854, opened 2023-08-11
Improve GroupBy JIT error handling

brandon-b-miller:
This PR implements general error handling for features that are missing dtype overloads in JIT GroupBy. Before, if a feature was not supported for a certain dtype, a somewhat confusing numba error is raised. With this PR however, a clear error is raised, for instance in the case of missing `float` dtypes for the `corr` aggregation:

```python
import pandas as pd
import cudf
df = pd.DataFrame({
    'a': [1, 1, 1, 2, 2, 2],
    'b': range(6),
    'c': range(6)
})
df['b'] = df['b'].astype('float64')
df['c'] = df['c'].astype('float64')


gdf = cudf.from_pandas(df, nan_as_null=False)
got = gdf.groupby('a').apply(lambda x: x['b'].corr(x['c']), engine='jit')
```
Now prints
```
Series.corr is not supported between float64 and float64 within JIT GroupBy apply.
Please file an issue requesting support for this feature at cuDF's GitHub page.
```


python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,16 +155,29 @@ def caller(data, index, size):
...
 
 
+class GroupUnaryAttrBase(AbstractTemplate):
+    def make_error_string(self, args):
+        fname = self.key.split(".")[-1]
+        return (
+            f"\nSeries.{fname} is not supported for "
+            f"{self.this.group_scalar_type} within JIT GroupBy "
+            "apply.\nPlease file an issue "
+            "requesting support for this feature at cuDF's GitHub page."
```
bdice, line 165: Do we intend to support all possible inputs that can raise this error? It seems like we're asking for a lot of small issues and some of them might even be "wontfix."

brandon-b-miller: I think this is a great question and worth discussion. From my point of view, the API is trying to convince the user as best as possible that their function is simply being run in a loop over groups, as in the iterative approach. In an ideal world then, every function we support in JIT groupby apply would support the same set of dtypes that the real `Series.api` supports. We have big holes in that right now, for example `corr` which is `int` only and the rest of the functions which don't support `int16`, `int8` or any unsigned types yet. Each one of these missing overloads is a missing feature IMO and it will take some time to close the gap. 

In raising as such I mainly intend to clarify what is going wrong, however in suggesting the issue I was aiming to create a feedback mechanism that might aid in our prioritization. However maybe as you suggest this is clumsy as we'd be waiting for users to file issues we already know exist.

What would you think of creating an umbrella issue that tracks the gap between supported dtypes and linking to it in the error instead, perhaps suggesting a github reply? Or, if pointing to github seems unnecessary in its entirety I can remove it.

bdice: A single issue would be better! Maybe you can start it and populate it with known gaps. But generally it’s unusual for cudf to point to GitHub issues in other circumstances such as when calling unsupported keyword arguments from pandas.

bdice: @shwina What are your thoughts here?

shwina: I think perhaps pointing a single issue in the error message so folks can provide input as comments is maybe a good compromise? Also it would be a real clickable link in the error message, which is nice. Perhaps something along:

> The `<func>()` function is not supported for the input types `<...>` by cuDF's JIT engine. To see what's currently supported or request support for a new function, go to https://github.com/rapidsai/cudf/issues/42

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -47,6 +52,15 @@ class GroupType(numba.types.Type):
     """
 
     def __init__(self, group_scalar_type, index_type=index_default_type):
+        if (
+            group_scalar_type not in SUPPORTED_GROUPBY_NUMBA_TYPES
+            and not isinstance(group_scalar_type, types.Poison)
+        ):
+            # an frame containing an column with an unsupported dtype
```
bdice, line 59:
```suggestion
            # A frame containing an column with an unsupported dtype
```

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -27,6 +30,8 @@
     numpy_support.as_dtype(dt) for dt in SUPPORTED_GROUPBY_NUMBA_TYPES
 ]
 
+docs = "https://docs.rapids.ai/api/cudf/stable/user_guide/guide-to-udfs/"
```
bdice, line 33:
```suggestion
_UDF_DOC_URL = "https://docs.rapids.ai/api/cudf/stable/user_guide/guide-to-udfs/"
```

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -47,6 +52,15 @@ class GroupType(numba.types.Type):
...
 
     def __init__(self, group_scalar_type, index_type=index_default_type):
+        if (
+            group_scalar_type not in SUPPORTED_GROUPBY_NUMBA_TYPES
+            and not isinstance(group_scalar_type, types.Poison)
+        ):
+            # an frame containing an column with an unsupported dtype
+            # is calling groupby apply. Construct a GroupType with
+            # a poisoned type so we can later error if this group is
+            # used in the UDAF body
```
bdice, line 62: Is UDAF something else? Or a typo?
```suggestion
            # used in the UDF body
```

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,15 +174,86 @@ def caller(data, index, size):
...
-def _make_unary_attr(funcname):
-    class GroupUnaryReductionAttrTyping(AbstractTemplate):
-        key = f"GroupType.{funcname}"
+class GroupOpBase(AbstractTemplate):
+    def make_error_string(self, args):
+        fname = self.key.__name__
+        sr_err = ", ".join(["Series" for _ in range(len(args))])
+        return (
+            f"{fname}({sr_err}) is not supported within JIT GroupBy "
+            f"apply. To see what's available, visit\n{docs}"
```
bdice, line 183:
```suggestion
            f"{fname}({sr_err}) is not supported by JIT GroupBy "
            f"apply. Supported features are listed at:\n{_UDF_DOC_URL}"
```

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,15 +174,86 @@ def caller(data, index, size):
...
+        )
+
+    def generic(self, args, kws):
+        # earlystop to make sure typing doesn't fail for normal
```
bdice, line 187:
```suggestion
        # early exit to make sure typing doesn't fail for normal
```

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,15 +174,86 @@ def caller(data, index, size):
...
+        # non-group ops
+        if not all(isinstance(arg, GroupType) for arg in args):
+            return None
+        # check if any groups are poisioned for this op
```
bdice, line 191:
```suggestion
        # check if any groups are poisoned for this op
```

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,15 +174,86 @@ def caller(data, index, size):
...
+        for arg in args:
+            if isinstance(arg.group_scalar_type, types.Poison):
+                raise UDFError(
+                    f"Use of a column of {arg.group_scalar_type.ty} detected "
+                    "within UDAF body. \nOnly columns of the following dtypes "
```
bdice, line 196:
```suggestion
                    "within UDF body. \nOnly columns of the following dtypes "
```

bdice: This appears a few times.

python/cudf/cudf/core/udf/groupby_typing.py
bdice: This appears a few times.

python/cudf/cudf/core/udf/groupby_typing.py
bdice: This appears a few times.

python/cudf/cudf/core/udf/utils.py
```diff
@@ -360,3 +361,7 @@ def set_malloc_heap_size(size=None):
 def column_to_string_view_array_init_heap(col):
     # lazily allocate heap only when a string needs to be returned
     return column_to_string_view_array(col)
+
+
+class UDFError(Exception):
```
bdice, line 366: Should this inherit from `RuntimeError`? Or some other exception class?

python/cudf/cudf/core/udf/groupby_typing.py
brandon-b-miller: Yeah, "user defined aggregation function". I figured these were distinct from scalar functions. on second thought though there's no reason to introduce extra terminology here.

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,15 +176,86 @@ def caller(data, index, size):
...
+            return None
+        # check if any groups are poisoned for this op
+        for arg in args:
+            if isinstance(arg.group_scalar_type, types.Poison):
+                raise UDFError(
+                    f"Use of a column of {arg.group_scalar_type.ty} detected "
+                    "within UDF body. \nOnly columns of the following dtypes "
+                    "may be used through the \nGroupBy.apply() JIT engine: "
+                    f"{[str(x) for x in SUPPORTED_GROUPBY_NUMPY_TYPES]}"
+                )
```
vyasr, line 201: Rather than erroring on the first encountered error, maybe collect all errors and then raise with the full set in case there are more than one? Saves the user having to iteratively remove parts of the UDF.

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,15 +176,86 @@ def caller(data, index, size):
...
+        fname = self.key.__name__
+        if fname in call_cuda_functions:
+            for sig in call_cuda_functions[fname].keys():
+                if all(
+                    arg.group_scalar_type == ty for arg, ty in zip(args, sig)
+                ):
+                    return nb_signature(sig[0], *args)
```
vyasr, line 208: Good place for a walrus
```suggestion
        if (funcs := call_cuda_funcs.get(self.key.__name__)):
            for sig in funcs.keys():
                if all(
                    arg.group_scalar_type == ty for arg, ty in zip(args, sig)
                ):
                    return nb_signature(sig[0], *args)
```

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,15 +176,86 @@ def caller(data, index, size):
...
+                    f"{[str(x) for x in SUPPORTED_GROUPBY_NUMPY_TYPES]}"
+                )
+        fname = self.key.split(".")[-1]
+        for sig in call_cuda_functions[fname].keys():
+            retty, selfty, *argtys = sig
+            if self.this.group_scalar_type == selfty and all(
+                arg.group_scalar_type == ty for arg, ty in zip(args, argtys)
+            ):
+                return nb_signature(retty, *args, recvr=self.this)
+        raise UDFError(self.make_error_string(args))
```
vyasr, line 245: Same suggestions from above apply here too for iterating over args and using the walrus.

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,15 +176,86 @@ def caller(data, index, size):
...
+            f"\nSeries.{fname}({sr_err}) is not supported for "
+            f"({dtype_err}) within JIT GroupBy apply.\nTo see "
+            f"what's available, visit {_UDF_DOC_URL}"
+        )
 
-        def generic(self, args, kws):
-            for retty, inputty in call_cuda_functions[funcname.lower()].keys():
-                if self.this.group_scalar_type == inputty:
-                    return nb_signature(retty, recvr=self.this)
+    def generic(self, args, kws):
```
vyasr, line 224: Probably not necessary right now since there's only two, but if you start adding more of these types that are just error case we should try to abstract the common code into either a common parent class or some standalone helper functions to call so that the error checking and messages are as standardized as possible.

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -206,11 +300,30 @@ def generic(self, args, kws):
...
-        return nb_signature(types.float64, args[0], recvr=self.this)
+
+def _make_unsupported_dataframe_attr(attrname):
+    def _attr(self, mod):
+        raise UDFError(
+            f"JIT GroupBy.apply() does not support DataFrame.{attrname}(). "
+            f"To see what's available, visit {_UDF_DOC_URL}"
+        )
+
+    return _attr
```
vyasr, line 314: I don't see this function used anywhere.

vyasr (changes requested):
Looks pretty close, couple of small things left that are worth doing though.

python/cudf/cudf/core/udf/groupby_typing.py
brandon-b-miller: That's a great suggestion and I hadn't thought about it from that perspective. In fact, normally this is what numba does by default. It goes through these typing classes `generic` functions one by one and if none of them return a signature (signaling success) it raises a general `TypingError`. This usually prints the big numba error string we're used to seeing, and contained within it are any actual errors raised by the `generic` functions themselves.

The thing is that this PR was kind of designed to break us out of that mechanism as convenient as it is for what you're suggesting - that is, raising and torpedoing typing phase _immediately_ with the goal of communicating something minimal and clear to the user instead. It seems like its worth exploring going back to an approach that mixes the existing mechanism with this new one and maybe just cleans up the default numba string, but I'd probably need to dive a little deeper into the numba internals next release to explore that in a way that bears fruit.

python/cudf/cudf/core/udf/groupby_typing.py
brandon-b-miller: Agree. I spent a little time trying to do this for a while, specifically having one base class for handling errors both for non-attr ops and attrs as well. The issue I ran into with it was that I couldn't make it simpler than computing a boolean `_IS_ATTR` up front and then branching `if`s based off it everywhere. In the end it felt like a class that was meant to do two separate things and it led me back to keeping it separate. But it's certainly possible that I just didn't boil the pot long enough :)

python/cudf/cudf/core/udf/groupby_typing.py
brandon-b-miller: Removed!

python/cudf/cudf/core/udf/groupby_typing.py
brandon-b-miller: Done

vyasr (approved):

brandon-b-miller:
I think this is ready to merge - any last thoughts anyone?

python/cudf/cudf/core/udf/groupby_typing.py
```diff
@@ -153,15 +176,87 @@ def caller(data, index, size):
...
+            return None
+        # check if any groups are poisoned for this op
+        for arg in args:
+            if isinstance(arg.group_scalar_type, types.Poison):
+                raise UDFError(
+                    f"Use of a column of {arg.group_scalar_type.ty} detected "
+                    "within UDF body. \nOnly columns of the following dtypes "
+                    "may be used through the \nGroupBy.apply() JIT engine: "
+                    f"{[str(x) for x in SUPPORTED_GROUPBY_NUMPY_TYPES]}"
+                )
```
bdice, line 201: I wouldn't use newlines in an error message unless it's meant to be an "announcement" with header lines like `*******` and some big error/warning paragraph with lots of instructions. This is just a normal error, in my view, so I'd write it as a single line string. Same for other errors in this PR.
```suggestion
                raise UDFError(
                    f"Use of a column of {arg.group_scalar_type.ty} detected "
                    "within UDF body. Only columns of the following dtypes "
                    "may be used through the GroupBy.apply() JIT engine: "
                    f"{[str(x) for x in SUPPORTED_GROUPBY_NUMPY_TYPES]}"
                )
```

bdice (approved):
I suggested simplifying the error message text. Otherwise I don't have strong opinions on this PR and we should merge it to unblock further work on migration to Numba 0.58.

PR #13862, opened 2023-08-11
Raise error when `astype(object)` is called in pandas compatibility mode

galipremsagar:
## Description
This PR disables type-casting to `object` dtype for non-string columns. It's a no-op for string columns. We disable this only in pandas-compatibility mode since the usage of `np.dtype("O")` is wide-spread and not quite possible to remove it's support entirely at this point.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/column/column.py
+++ b/python/cudf/cudf/core/column/column.py
@@ -994,7 +994,7 @@ class ColumnBase(Column, Serializable, BinaryOperand, Reducible):
                 dtype
             ).type in {np.object_}:
                 raise ValueError(
-                    f"{dtype} is an unsupported dtype to type-cast to, use "
+                    f"Casting to {dtype} is not supported, use "
                     "`.astype('str')` instead."
                 )
             return self.as_string_column(dtype, **kwargs)
--- a/python/cudf/cudf/core/column/column.py
+++ b/python/cudf/cudf/core/column/column.py
@@ -990,9 +990,11 @@ class ColumnBase(Column, Serializable, BinaryOperand, Reducible):
             np.object_,
             str,
         }:
-            if cudf.get_option("mode.pandas_compatible") and np.dtype(
-                dtype
-            ).type in {np.object_}:
+            if (
+                cudf.get_option("mode.pandas_compatible")
+                and not isinstance(self, cudf.core.column.StringColumn)
+                and np.dtype(dtype).type in {np.object_}
+            ):
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/column/column.py
```diff
@@ -990,6 +990,15 @@ def astype(self, dtype: Dtype, **kwargs) -> ColumnBase:
             np.object_,
             str,
         }:
+            if (
+                cudf.get_option("mode.pandas_compatible")
+                and not isinstance(self, cudf.core.column.StringColumn)
```
bdice, line 995: Why should we allow string columns to be cast as `object`? I expected that _nothing_ should be cast-able to `object` since cudf doesn't truly support it.
```suggestion
```

python/cudf/cudf/tests/test_series.py
```diff
@@ -2224,3 +2224,14 @@ def __getitem__(self, key):
 def test_series_constructor_error_mixed_type():
     with pytest.raises(pa.ArrowTypeError):
         cudf.Series(["abc", np.nan, "123"], nan_as_null=False)
+
+
+def test_series_typecast_error():
+    actual = cudf.Series([1, 2, 3], dtype="datetime64[ns]")
+    with cudf.option_context("mode.pandas_compatible", True):
```
bdice, line 2231: Should we add tests that this does _not_ fail when `pandas_compatible` is `False`? (With the expectation that behavior could change in the future)

python/cudf/cudf/core/column/column.py
galipremsagar: Agreed 👍 This however wouldn't be in-effect until we change our dtype to `np.dtype('str')` fully everywhere.

python/cudf/cudf/tests/test_series.py
galipremsagar: Done 👍

python/cudf/cudf/core/column/column.py
```diff
@@ -990,6 +990,13 @@ def astype(self, dtype: Dtype, **kwargs) -> ColumnBase:
             np.object_,
             str,
         }:
+            if cudf.get_option("mode.pandas_compatible") and np.dtype(
+                dtype
+            ).type in {np.object_}:
+                raise ValueError(
+                    f"{dtype} is an unsupported dtype to type-cast to, use "
```
bdice, line 997: This simplifies the error text. Might require re-wrapping.
```suggestion
                    f"Casting to {dtype} is not supported, use "
```

bdice (approved):

Issue #13867, opened 2023-08-12
[FEA] Change `NA` to `NaT` for `timedelta` & `datetime` types


galipremsagar:
**Is your feature request related to a problem? Please describe.**
Pandas array with `timedelta` & `datetime` types too have `NaT` as a null value, so this is a feature request to change our `<NA>`  to `NaT` just for these two types. Also, there are no other null values possible for these type and they don't have a corresponding nullable dtype.

**Describe the solution you'd like**
Change repr from `<NA>`  to `NaT` for `timedelta64` & `datetime64` dtypes.

PR #13868, opened 2023-08-12
Change `NA` to `NaT` for `datetime` and `timedelta` types

galipremsagar:
## Description
Resolves: #13867 

This PR:

- [x] Changes all repr code of `datetime` & `timdelta` types to switch to represent null values as `NaT` from `<NA>`.
- [x] Changes Scalar getitem to return `cudf.NaT` instead of `cudf.NA` 

## Checklist


python/cudf/cudf/_lib/scalar.pyx
```diff
@@ -178,7 +178,7 @@ cdef class DeviceScalar:
         return self.get_raw_ptr()[0].is_valid()
 
     def __repr__(self):
-        if self.value is NA:
+        if self.value is NA or self.value is NaT:
```
bdice, line 181: Do we want to have a helper function like `is_na_like` (to avoid confusion with the pandas name `isna`)? That would help reduce repeated implementation of the same logic. We also have some places where we write `x in {None, NA, NaT}` and other places where we write `x is None or x is NA or x is NaT`. Those seem like they should be cleaned up and use standard/common logic in this PR -- maybe a couple flavors of that logic, depending on `None` treatment.

python/cudf/cudf/core/column/column.py
```diff
@@ -605,7 +605,7 @@ def __setitem__(self, key: Any, value: Any):
             self._mimic_inplace(out, inplace=True)
 
     def _wrap_binop_normalization(self, other):
-        if other is NA or other is None:
+        if other is NA or other is None or other is NaT:
```
bdice, line 608:
```suggestion
        if other in {None, NA, NaT}:
```

bdice (commented):
This looks good overall. I'm a little surprised the changeset isn't larger than this. I would like to think about refactoring the ways we check if values are "None, NA, NaT"-like. I understand it might vary by the calling code, but we might be able to reduce repeated logic a little. If you don't see an opportunity to improve that, I am fine with accepting this PR as-is.

python/cudf/cudf/_lib/scalar.pyx
galipremsagar: Yup, added `is_na_like` utility to remove this common pattern repetition.

galipremsagar:
> This looks good overall. I'm a little surprised the changeset isn't larger than this. I would like to think about refactoring the ways we check if values are "None, NA, NaT"-like. I understand it might vary by the calling code, but we might be able to reduce repeated logic a little. If you don't see an opportunity to improve that, I am fine with accepting this PR as-is.

I was able to add `is_na_like` utility and replaced all such repetitive patterns.

python/cudf/cudf/core/dataframe.py
```diff
@@ -1694,7 +1694,16 @@ def _clean_nulls_from_dataframe(self, df):
...
+                if isinstance(
+                    df._data[col],
+                    (
+                        cudf.core.column.DatetimeColumn,
+                        cudf.core.column.TimeDeltaColumn,
+                    ),
+                ):
+                    df[col] = df._data[col].astype("str").fillna(cudf._NAT_REP)
+                else:
+                    df[col] = df._data[col].astype("str").fillna(cudf._NA_REP)
```
bdice, line 1706: Could we write this like
```python
fill_value = cudf._NAT_REP if ... else cudf._NA_REP
df[col] = df._data[col].astype("str").fillna(fill_value)
```
Just want to avoid repeating the bulk of the `fillna` logic. Only the value changes.

python/cudf/cudf/utils/dtypes.py
```diff
@@ -16,6 +16,7 @@
 from cudf.core.missing import NA
 
 _NA_REP = "<NA>"
+_NAT_REP = "NaT"
```
bdice, line 19: I'm not _really_ convinced that we need `_NA_REP`. You can write `str(cudf.NA)` to get this value. Same for NaT. Is this actually helpful, or does it just result in needless imports when you can call `str`? (This can be refactored in a separate PR if you agree it's a good idea.)

python/cudf/cudf/utils/dtypes.py
```diff
@@ -16,6 +16,7 @@
...
```
bdice, line 20: Do we need to do something about this line: https://github.com/galipremsagar/cudf/blob/09466815b0d4da61b80115de4e0612965ccde82a/python/cudf/cudf/utils/dtypes.py#L643

python/cudf/cudf/tests/test_repr.py
```diff
@@ -41,9 +41,7 @@ def test_null_series(nrows, dtype):
 
     pd.options.display.max_rows = int(nrows)
     psrepr = repr(ps)
-    psrepr = psrepr.replace("NaN", "<NA>")
-    psrepr = psrepr.replace("NaT", "<NA>")
-    psrepr = psrepr.replace("None", "<NA>")
+    psrepr = psrepr.replace("NaN", "<NA>").replace("None", "<NA>")
```
bdice, line 44: Lines 44 and 72 are splitting the computation differently. One calls `repr` first, then replaces. The other calls `repr` and replaces inline. I'd prefer sticking to one of those options.

python/cudf/cudf/core/series.py
```diff
@@ -1404,7 +1404,19 @@ def __repr__(self):
             cudf.core.column.timedelta.TimeDeltaColumn,
         ):
             output = repr(
-                preprocess.astype("O").fillna(cudf._NA_REP).to_pandas()
+                preprocess.astype("O")
```
bdice, line 1407: Is this affected by #13862?
```suggestion
                preprocess.astype("str")
```

python/cudf/cudf/core/dataframe.py
bdice: Or even copy the style in `core/index.py` from this PR:
```python
df[col] = df._data[col].astype("str").fillna(cudf._NAT_REP if ... else cudf._NA_REP)
```

python/cudf/cudf/tests/test_repr.py
galipremsagar: Changed to make it consistent with the rest of the file.

python/cudf/cudf/utils/dtypes.py
galipremsagar: Yup, updated it.

python/cudf/cudf/core/dataframe.py

python/cudf/cudf/__init__.py
```diff
@@ -124,6 +123,7 @@
     "IntervalIndex",
     "ListDtype",
     "MultiIndex",
+    "NaT",
     "NA",
```
bdice, line 127:
```suggestion
    "NA",
    "NaT",
```

bdice (approved):
Great! Thanks for the iterations.

Issue #13861, opened 2023-08-11
[BUG] MultiIndex.isin returns True instead of False for partial element matching tuples


mroeschke:
**Describe the bug**
MultiIndex.isin returns True for partial element matching tuples

**Steps/Code to reproduce bug**
```
In [41]: import cudf

In [42]: cudf.__version__
Out[42]: '23.10.00'

In [43]: mi = cudf.MultiIndex.from_product([[0, 1], ['a', 'b', 'c']])

In [44]: mi.isin([(1, 'a'), (2, 'b'), (0, 'c')])
Out[44]: array([ True,  True,  True,  True,  True,  True])

In [45]: mi
Out[45]: 
MultiIndex([(0, 'a'),
            (0, 'b'),
            (0, 'c'),
            (1, 'a'),
            (1, 'b'),
            (1, 'c')],
           )
```

**Expected behavior**
isin should only return True if all elements in a tuple a MultiIndex label

```
In [47]: mi = pandas.MultiIndex.from_product([[0, 1], ['a', 'b', 'c']])

In [48]: mi.isin([(1, 'a'), (2, 'b'), (0, 'c')])
Out[48]: array([False, False,  True,  True, False, False])
```

**Environment overview (please complete the following information)**
 - Environment location: Bare-metal
 - Method of cuDF install: conda
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**
Please run and paste the output of the `cudf/print_env.sh` script here, to gather any other relevant environment details

**Additional context**

PR #13879, opened 2023-08-14
Check for the presence of all values in `MultiIndex.isin`

galipremsagar:
## Description
Fixes: #13861 

This PR fixes an issue in `MulitIndex.isin` where, previously we calculated partial matches. With this PR we will only return `True` for complete matches.

## Checklist


bdice (approved):
It looks like this covers the case in that bug. I tried to think if there was a cheaper/faster way than to call `merge`, but I didn't immediately see anything.

Issue #13875, opened 2023-08-14
[BUG] JIT `GroupBy.apply` `corr()` aggregation wrong results when one variable has zero variance


brandon-b-miller:
**Describe the bug**
When executing a `corr` aggregation in a groupby apply UDF, the result can diverge from pandas in some cases, specifically when one of the variables to be correlated has no variance.

**Steps/Code to reproduce bug**
```python
import pandas as pd
import cudf

import numpy as np

df = pd.DataFrame(
    {
        'a':[0,0,0],
        'b':[1,1,1],
        'c':[2,2,2] 
    }
)
gdf = cudf.from_pandas(df)

def func(group):

    return group['b'].corr(group['c'])

pandas_result = df.groupby('a').apply(func)
cudf_result = gdf.groupby('a').apply(func, engine='jit')

print(pandas_result)
print(cudf_result)

```

```
a
0   NaN
dtype: float64
a
0    0.0
dtype: float64
```

**Expected behavior**
Results should be the same as pandas.

**Environment overview (please complete the following information)**
Bare-metal, 23.10

**Additional Context**
came up during https://github.com/rapidsai/cudf/pull/13813. This must be due to mishandling the denominator in `corr`.

PR #13813, opened 2023-08-03
Expand JIT groupby test suite

brandon-b-miller:
This PR reorganizes and expands the test suite for groupby apply functions using the JIT engine to include nan cases and cases where the groups are larger than a single thread block.

PR #13884, opened 2023-08-15
Return `nan` when one variable to be correlated has zero variance in JIT GroupBy Apply

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/13875


brandon-b-miller:
This was found during https://github.com/rapidsai/cudf/pull/13813

python/cudf/udf_cpp/shim.cu
```diff
@@ -643,9 +643,8 @@ __device__ double BlockCorr(T* const lhs_ptr, T* const rhs_ptr, int64_t size)
 {
   auto numerator   = BlockCoVar(lhs_ptr, rhs_ptr, size);
   auto denominator = BlockStd(lhs_ptr, size) * BlockStd<T>(rhs_ptr, size);
-
   if (denominator == 0.0) {
-    return 0.0;
+    return std::numeric_limits<double>::quiet_NaN();
```
galipremsagar, line 647: Can we have a pytest added to test this?

vyasr (approved):
Approving assuming that a new test is added as per Prem's request.

python/cudf/udf_cpp/shim.cu
brandon-b-miller: Added a simple test in https://github.com/rapidsai/cudf/pull/13884/commits/26ad24a7530563f553bbff3c297e4794aa3fee53. More tests pending, including one which covers this case, at https://github.com/rapidsai/cudf/pull/13854.

galipremsagar (approved):

galipremsagar (approved):

brandon-b-miller:
@galipremsagar jinx :)

Issue #13501, opened 2023-06-02
[FEA] Support V2 encodings in Parquet reader and writer


GregoryKimball:
Parquet V1 format supports three types of page encodings: PLAIN, DICTIONARY, and RLE (run-length encoded) ([reference from Spark Jira](https://issues.apache.org/jira/browse/SPARK-36879)). The newer and evolving Parquet V2 specification adds support for several [additional encodings](https://parquet.apache.org/docs/file-format/data-pages/encodings/), including DELTA_BINARY_PACKED for `INT32` and `INT64` types, DELTA_BYTE_ARRAY for `strings` logical type, and DELTA_LENGTH_BYTE_ARRAY for `strings` logical type. 

In the parquet reader and writer, libcudf should support V2 metadata as well as the three variants of DELTA encoding.

| Feature | Status | Notes | 
|---|---|---|
| Add V2 reader support |  ✅ #11778 | |
| Multi-warp decode of Dremel data streams | ✅ #13203 |  | 
| Use efficient strings column factory in decoder | ✅ #13302 |  | 
| Implement DELTA_BINARY_PACKED decoding |  ✅  #13637 | see #12948 for reference |
| Implement DELTA_BYTE_ARRAY decoding | ✅ #14101 | see #12948 for reference |
| Add V2 writer support | ✅ #13751 | |
| Implement DELTA_BINARY_PACKED encoding | ✅ #14100 | | 
| Add python bindings for V2 header and options | ✅ #14316 | |
| Implement DELTA_BYTE_ARRAY encoding | ✅ #15239 | some outdated reviews in #14938 |
| Implement DELTA_LENGTH_BYTE_ARRAY encoding and decoding for unsorted data | ✅ #14590 | |
| Add C++ API support for specifying encodings | ✅ #15081 | |
| Add cuDF-python API support for specifying encodings |  | |
| Add BYTE_STREAM_SPLIT encoding and decoding  | ✅ #15311 | see issue #15226 and [parquet reference](https://github.com/apache/parquet-format/blob/master/Encodings.md#byte-stream-split-byte_stream_split--9) |

etseidl:
@GregoryKimball should there be an entry for adding python bindings for the V2 options?

GregoryKimball:
Congratulations @etseidl! Everyone, please stay tuned for a technical blog on this topic! 😄

PR #13886, opened 2023-08-15
Enable RLE boolean encoding for v2 Parquet files

etseidl:
## Description
While working on #13707 it was noticed that RLE encoding of booleans had been implemented and then disabled (see [this comment](https://github.com/rapidsai/cudf/pull/13707#discussion_r1275194215) for details). This PR re-enables RLE encoding for booleans, but only when V2 headers are being used.

Part of #13501.

## Checklist

### DIFF
--- a/cpp/src/io/parquet/page_enc.cu
+++ b/cpp/src/io/parquet/page_enc.cu
@@ -403,6 +403,11 @@ __global__ void __launch_bounds__(128)
       num_pages = 1;
     }
     __syncwarp();
+
+    // page padding needed for RLE encoded boolean data
+    auto const rle_pad =
+      write_v2_headers && col_g.physical_type == BOOLEAN ? RLE_LENGTH_FIELD_LEN : 0;
+
     // This loop goes over one page fragment at a time and adds it to page.
     // When page size crosses a particular limit, then it moves on to the next page and then next
     // page fragment gets added to that one.
@@ -446,12 +451,12 @@ __global__ void __launch_bounds__(128)
       // override this_max_page_size if the requested size is smaller
       this_max_page_size = min(this_max_page_size, max_page_size_bytes);
 
-      // subtract size of rep and def level vectors
-      auto num_vals = values_in_page + frag_g.num_values;
-      this_max_page_size =
-        underflow_safe_subtract(this_max_page_size,
-                                max_RLE_page_size(col_g.num_def_level_bits(), num_vals) +
-                                  max_RLE_page_size(col_g.num_rep_level_bits(), num_vals));
+      // subtract size of rep and def level vectors and RLE length field
+      auto num_vals      = values_in_page + frag_g.num_values;
+      this_max_page_size = underflow_safe_subtract(
+        this_max_page_size,
+        max_RLE_page_size(col_g.num_def_level_bits(), num_vals) +
+          max_RLE_page_size(col_g.num_rep_level_bits(), num_vals) + rle_pad);
 
       // checks to see when we need to close the current page and start a new one
       auto const is_last_chunk          = num_rows >= ck_g.num_rows;
@@ -493,11 +498,7 @@ __global__ void __launch_bounds__(128)
           page_g.num_values         = values_in_page;
           auto const def_level_size = max_RLE_page_size(col_g.num_def_level_bits(), values_in_page);
           auto const rep_level_size = max_RLE_page_size(col_g.num_rep_level_bits(), values_in_page);
-          if (write_v2_headers && col_g.physical_type == BOOLEAN) {
-            // if RLE encoding booleans then will need an extra 4 bytes to encode the length
-            page_size += RLE_LENGTH_FIELD_LEN;
-          }
-          auto const max_data_size = page_size + def_level_size + rep_level_size;
+          auto const max_data_size  = page_size + def_level_size + rep_level_size + rle_pad;
           // page size must fit in 32-bit signed integer
           if (max_data_size > std::numeric_limits<int32_t>::max()) {
             CUDF_UNREACHABLE("page size exceeds maximum for i32");
@@ -1379,8 +1380,8 @@ __global__ void __launch_bounds__(128, 8)
   // save RLE length if necessary
   if (s->rle_len_pos != nullptr && t < 32) {
     // size doesn't include the 4 bytes for the length
-    auto const rle_size = static_cast<uint32_t>(s->cur - s->rle_len_pos) - 4;
-    if (t < 4) { s->rle_len_pos[t] = rle_size >> (t * 8); }
+    auto const rle_size = static_cast<uint32_t>(s->cur - s->rle_len_pos) - RLE_LENGTH_FIELD_LEN;
+    if (t < RLE_LENGTH_FIELD_LEN) { s->rle_len_pos[t] = rle_size >> (t * 8); }
     __syncwarp();
   }
 
--- a/cpp/src/io/parquet/page_enc.cu
+++ b/cpp/src/io/parquet/page_enc.cu
@@ -47,12 +47,6 @@ namespace parquet {
 namespace gpu {
 
 namespace {
-// Spark doesn't support RLE encoding for BOOLEANs
-#ifdef ENABLE_BOOL_RLE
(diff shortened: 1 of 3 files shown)


cpp/src/io/parquet/page_enc.cu
```diff
@@ -1188,7 +1216,7 @@ __global__ void __launch_bounds__(128, 8)
         }
         rle_numvals += rle_numvals_in_block;
         __syncthreads();
-        if ((!enable_bool_rle) && (physical_type == BOOLEAN)) {
+        if ((!is_v2) && (physical_type == BOOLEAN)) {
```
bdice, line 1219: Extraneous parens can be removed.
```suggestion
        if (!is_v2 && physical_type == BOOLEAN) {
```

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -676,12 +677,14 @@ void InitEncoderPages(cudf::detail::device_2dspan<EncColumnChunk> chunks,
  * @param[out] comp_in Compressor input buffers
  * @param[out] comp_in Compressor output buffers
  * @param[out] comp_stats Compressor results
+ * @param[in] write_v2_headers True if V2 page headers should be written
```
bdice, line 680: Should this "in" parameter be listed in the function signature before the "out" parameters? (Is this docstring correct about what it labels "in" and "out"?)

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -676,12 +677,14 @@ void InitEncoderPages(cudf::detail::device_2dspan<EncColumnChunk> chunks,
  * @param[out] comp_in Compressor input buffers
  * @param[out] comp_in Compressor output buffers
  * @param[out] comp_stats Compressor results
```
bdice, line 679:
```suggestion
 * @param[out] comp_res Compressor results
```

cpp/src/io/parquet/parquet_gpu.hpp
etseidl: I can move it up. I think all the docstrings in this file could use some love. I tried cleaning up some in a prior PR, but there's still more to do :(

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -676,12 +677,14 @@ void InitEncoderPages(cudf::detail::device_2dspan<EncColumnChunk> chunks,
...
+ * @param[in] write_v2_headers True if V2 page headers should be written
  * @param[in] stream CUDA stream to use, default 0
```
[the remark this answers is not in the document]
etseidl: I think the default was removed some time ago but the docs weren't updated. I'll clean up some more.

cpp/src/io/parquet/parquet_gpu.hpp
bdice: No worries, and thank you for leaving the code in a better state than you found it. 👍 I'll defer to @vuule on the decision of parameter order if he has other preferences. This is fine by me.

bdice (approved):

cpp/src/io/parquet/parquet_gpu.hpp
vuule: AFAICT, the order is consistent with most functions in cuIO so I'm fine with the current location. Having `write_v2_headers` as the first parameter would somehow be more odd to me (even though in theory that's the right spot).

etseidl: should I revert the change to the param order then?

vuule: it's fine :D

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1132,9 +1154,15 @@ __global__ void __launch_bounds__(128, 8)
     s->rle_pos     = 0;
     s->rle_numvals = 0;
     s->rle_out     = dst;
+    s->page.encoding =
+      determine_encoding(s->page.page_type, physical_type, s->ck.use_dictionary, write_v2_headers);
     if (dict_bits >= 0 && physical_type != BOOLEAN) {
       dst[0]     = dict_bits;
       s->rle_out = dst + 1;
+    } else if (is_v2 && physical_type == BOOLEAN) {
+      // save space for RLE length. we don't know the total length yet.
```
vuule, line 1163: why is such adjustment only needed for bool RLE?

vuule (commented):
Looks good, just a few comments.

cpp/src/io/parquet/page_enc.cu
etseidl: So RLE is only used for 1) rep/def levels, 2) dictionary indexes, and 3) booleans. So booleans are the only _data_ type that will be RLE encoded...everything else will be plain or dictionary (or eventually delta).

There's actually a table in the [spec](https://github.com/apache/parquet-format/blob/master/Encodings.md#run-length-encoding--bit-packing-hybrid-rle--3)  that lists when a length is prepended and when it is not. 
```
+--------------+------------------------+-----------------+
| Page kind    | RLE-encoded data kind  | Prepend length? |
+--------------+------------------------+-----------------+
| Data page v1 | Definition levels      | Y               |
|              | Repetition levels      | Y               |
|              | Dictionary indices     | N               |
|              | Boolean values         | Y               |
+--------------+------------------------+-----------------+
| Data page v2 | Definition levels      | N               |
|              | Repetition levels      | N               |
|              | Dictionary indices     | N               |
|              | Boolean values         | Y               |
+--------------+------------------------+-----------------+
```

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1132,9 +1154,15 @@ __global__ void __launch_bounds__(128, 8)
...
+      s->rle_out     = dst + 4;
```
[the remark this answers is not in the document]
etseidl: | I assume this is the answer to the question above and this is the same 4

yep.

| Could you add a constant to give name to the RLE length size (assuming that's what 4 is)?

sure, sounds sensible

cpp/src/io/parquet/page_enc.cu
vuule: I think I get it, thanks.
And this is something that does not make sense to include in fragment size?
I'm asking because this "max RLE page size" is implemented as a util function in the ORC writer, so this manual padding looked odd in comparison.

BTW, must be nice not having RLE in almost every data type (cries in ORC).

cpp/src/io/parquet/page_enc.cu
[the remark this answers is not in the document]
etseidl: done

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1345,29 +1376,45 @@ __global__ void __launch_bounds__(128, 8)
 
   uint32_t const valid_count = block_reduce(temp_storage.reduce_storage).Sum(num_valid);
 
+  // save RLE length if necessary
+  if (s->rle_len_pos != nullptr && t < 32) {
+    // size doesn't include the 4 bytes for the length
+    auto const rle_size = static_cast<uint32_t>(s->cur - s->rle_len_pos) - 4;
```
vuule, line 1382: is this 4 also RLE_LENGTH_FIELD_LEN?

cpp/src/io/parquet/page_enc.cu
etseidl: > And this is something that does not make sense to include in fragment size?

yeah, because it's just once per page, not every fragment. although, yeah, maybe it should be included in the calculation at line 451.  I'll ponder that.

cpp/src/io/parquet/page_enc.cu
etseidl: >.< yes

cpp/src/io/parquet/page_enc.cu
etseidl: ok, I think that fixed it.  nice catch!  That would have been a lot of fun tracking down later :P

vuule (approved):
:fire:

PR #13890, opened 2023-08-16
Disable creation of `DatetimeIndex` when `freq` is passed to `cudf.date_range`

galipremsagar:
## Description
In https://github.com/rapidsai/cudf/pull/13857/, we disabled the construction of `DatetimeIndex` & `TimedeltaIndex` when their pandas counter-parts have `freq` set. However, when `freq` is passed to `date_range`, the expectation and pandas-behavior is to return a `DateTimeIndex` with `freq` set. This isn't possible with cudf currently, hence disabled the construction of `DatetimeIndex` in this case.

## Checklist


bdice (approved):
This looks fine. Also, I see that `Timestamp.freq` was removed in pandas 2 (probably because it's meaningless as far as I can tell). That doesn't really help with the `DatetimeIndex` but might help with the timestamps themselves.

PR #13893, opened 2023-08-16
Allow explicit `shuffle="p2p"` within dask-cudf API

rjzamora:
## Description
This PR allows explicit `shuffle="p2p"` usage within the dask-cudf API now that https://github.com/dask/distributed/pull/7743 is in.


rjzamora:
>Should we try to get this in for 23.10 ?

The only reason we may want to hold off on this PR "as is"  is that p2p will become the default on distributed systems, because we will obey [get_default_shuffle_method](https://github.com/dask/dask/blob/29ee29c38630fa7f7a5c57bfd065103ec5d88110/dask/utils.py#L2107) as long as it doesn't return `"disk"`.  I personally think this is a good thing in the long run, since "p2p" should be more stable than "tasks". However, (1) there may be users who are still expecting the `"tasks"` default, and (2) there may be edge cases where "p2p" doesn't return the same result (I'm thinking about index preservation).

If we want to target 23.10, then perhaps it would be best to enable `shuffle="p2p"` without making it default behavior?

wence- (approved):
I agree that flipping the default is probably not the right thing for 23.10 (at low worker/partition counts I think that users would likely notice a performance regression right now).

rjzamora:
Just leaving a note that this is currently blocked by #14159

CHANGELOG.md
```diff
@@ -3314,6 +3314,7 @@
 - Fixing empty null lists throwing explode_outer for a loop. ([#7649](https://github.com/rapidsai/cudf/pull/7649)) [@hyperbolic2346](https://github.com/hyperbolic2346)
 - Fix internal compiler error during JNI Docker build ([#7645](https://github.com/rapidsai/cudf/pull/7645)) [@jlowe](https://github.com/jlowe)
 - Fix Debug build break with device_uvectors in grouped_rolling.cu ([#7633](https://github.com/rapidsai/cudf/pull/7633)) [@mythrocks](https://github.com/mythrocks)
+- Parquet reader: Fix issue when using skip_rows on non-nested columns containing nulls ([#7627](https://github.com/rapidsai/cudf/pull/7627)) [@nvdbaranec](https://github.com/nvdbaranec)
```
wence-, line 3317: This should disappear I think.

CHANGELOG.md
```diff
@@ -3413,6 +3414,7 @@
 - Add groupby scan operations (sort groupby) ([#7387](https://github.com/rapidsai/cudf/pull/7387)) [@karthikeyann](https://github.com/karthikeyann)
 - Add cudf::explode_position ([#7376](https://github.com/rapidsai/cudf/pull/7376)) [@hyperbolic2346](https://github.com/hyperbolic2346)
 - Add string conversion to/from decimal values libcudf APIs ([#7364](https://github.com/rapidsai/cudf/pull/7364)) [@davidwendt](https://github.com/davidwendt)
+- Add groupby SUM_OF_SQUARES support ([#7362](https://github.com/rapidsai/cudf/pull/7362)) [@karthikeyann](https://github.com/karthikeyann)
```
wence-, line 3417: And this one.

CHANGELOG.md
```diff
@@ -3421,6 +3423,7 @@
 - Add support for special tokens in nvtext::subword_tokenizer ([#7254](https://github.com/rapidsai/cudf/pull/7254)) [@davidwendt](https://github.com/davidwendt)
 - Fix inplace update of data and add Series.update ([#7201](https://github.com/rapidsai/cudf/pull/7201)) [@galipremsagar](https://github.com/galipremsagar)
 - Implement `cudf::group_by` (hash) for `decimal32` and `decimal64` ([#7190](https://github.com/rapidsai/cudf/pull/7190)) [@codereport](https://github.com/codereport)
+- Adding support to specify &quot;level&quot; parameter for `Dataframe.rename` ([#7135](https://github.com/rapidsai/cudf/pull/7135)) [@skirui-source](https://github.com/skirui-source)
```
wence-, line 3426: And this one.

CHANGELOG.md
```diff
@@ -3561,6 +3564,7 @@
 - Adding Interval Dtype ([#6984](https://github.com/rapidsai/cudf/pull/6984)) [@marlenezw](https://github.com/marlenezw)
 - Cleaning up `for` loops with `make_(counting_)transform_iterator` ([#6546](https://github.com/rapidsai/cudf/pull/6546)) [@codereport](https://github.com/codereport)
 
+
```
wence-, line 3567:
```suggestion
```

rjzamora:
@wence- - I know you approved this PR a few days ago, but the scope has changed a bit. Let me know if you still think this is good to go.

wence- (approved):
One small suggested simplification.

PR #13895, opened 2023-08-16
Use `thread_index_type` to avoid index overflow in grid-stride loops

PointKernel:
## Description
This PR checks all related files under `src/hash`, `src/bitmask` and `src/transform` folders and fixes potential index overflow issues by using `thread_index_type`.

## Checklist

### DIFF
--- a/cpp/src/bitmask/null_mask.cu
+++ b/cpp/src/bitmask/null_mask.cu
@@ -108,9 +108,11 @@ __global__ void set_null_mask_kernel(bitmask_type* __restrict__ destination,
   auto const last_word    = word_index(end_bit) - word_index(begin_bit);
   bitmask_type fill_value = valid ? 0xffff'ffff : 0;
 
-  for (size_type destination_word_index = threadIdx.x + blockIdx.x * blockDim.x;
+  thread_index_type const stride = blockDim.x * gridDim.x;
+
+  for (thread_index_type destination_word_index = threadIdx.x + blockIdx.x * blockDim.x;
        destination_word_index < number_of_mask_words;
-       destination_word_index += blockDim.x * gridDim.x) {
+       destination_word_index += stride) {
     if (destination_word_index == 0 || destination_word_index == last_word) {
       bitmask_type mask = ~bitmask_type{0};
       if (destination_word_index == 0) {
@@ -189,9 +191,10 @@ __global__ void copy_offset_bitmask(bitmask_type* __restrict__ destination,
                                     size_type source_end_bit,
                                     size_type number_of_mask_words)
 {
-  for (size_type destination_word_index = threadIdx.x + blockIdx.x * blockDim.x;
+  thread_index_type const stride = blockDim.x * gridDim.x;
+  for (thread_index_type destination_word_index = threadIdx.x + blockIdx.x * blockDim.x;
        destination_word_index < number_of_mask_words;
-       destination_word_index += blockDim.x * gridDim.x) {
+       destination_word_index += stride) {
     destination[destination_word_index] = detail::get_mask_offset_word(
       source, destination_word_index, source_begin_bit, source_end_bit);
   }
@@ -261,14 +264,15 @@ __global__ void count_set_bits_kernel(bitmask_type const* bitmask,
 
   auto const first_word_index{word_index(first_bit_index)};
   auto const last_word_index{word_index(last_bit_index)};
-  auto const tid         = threadIdx.x + blockIdx.x * blockDim.x;
-  auto thread_word_index = tid + first_word_index;
+  thread_index_type const tid         = threadIdx.x + blockIdx.x * blockDim.x;
+  thread_index_type const stride      = blockDim.x * gridDim.x;
+  thread_index_type thread_word_index = tid + first_word_index;
   size_type thread_count{0};
 
   // First, just count the bits in all words
   while (thread_word_index <= last_word_index) {
     thread_count += __popc(bitmask[thread_word_index]);
-    thread_word_index += blockDim.x * gridDim.x;
+    thread_word_index += stride;
   }
 
   // Subtract any slack bits counted from the first and last word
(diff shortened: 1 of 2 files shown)


PointKernel:
Question for reviewers: I don't think we have tests exercising an input that could cause overflow. I hesitated to add those tests since it requires relatively large memory and may take a long time to execute. What do you think?

cpp/src/bitmask/null_mask.cu
```diff
@@ -261,14 +264,15 @@ __global__ void count_set_bits_kernel(bitmask_type const* bitmask,
 
   auto const first_word_index{word_index(first_bit_index)};
   auto const last_word_index{word_index(last_bit_index)};
-  auto const tid         = threadIdx.x + blockIdx.x * blockDim.x;
-  auto thread_word_index = tid + first_word_index;
+  thread_index_type const tid         = threadIdx.x + blockIdx.x * blockDim.x;
+  thread_index_type const stride      = blockDim.x * gridDim.x;
+  thread_index_type thread_word_index = tid + first_word_index;
```
ttnghia, line 269: Probably we also need to upgrade `first_` and `last_` indices above.

cpp/src/bitmask/null_mask.cu
```diff
@@ -108,9 +108,11 @@ __global__ void set_null_mask_kernel(bitmask_type* __restrict__ destination,
   auto const last_word    = word_index(end_bit) - word_index(begin_bit);
   bitmask_type fill_value = valid ? 0xffff'ffff : 0;
 
-  for (size_type destination_word_index = threadIdx.x + blockIdx.x * blockDim.x;
+  thread_index_type const stride = blockDim.x * gridDim.x;
```
ttnghia, line 111: May also need to upgrade `last_word`.

karthikeyann:
> Question for reviewers: I don't think we have tests exercising an input that could cause overflow. I hesitated to add those tests since it requires relatively large memory and may take a long time to execute. What do you think?

The for-loop comparison happens on short-width type `size_type`. So, is overflow possible to test?

PointKernel:
> The for-loop comparison happens against shorter-width type size_type. So, is overflow possible to test?

It should be possible if the input size is smaller than `INT_MAX` and larger than `INT_MAX - stride`, then the "last" iteration with `int32_t` index will produce negative indices when doing `tid += stride;` which is smaller than the end condition thus causing illegal memory access in the loop after.

cpp/src/bitmask/null_mask.cu
PointKernel: We can reply on implicit conversions for `<` or `<=` since they are always cast to the larger data type for comparison. `first_word_index` and `last_word_index` are explicitly used as `size_type` later in the loop thus setting them to `thread_index_type` seems not worth the effort.

karthikeyann:
Benchmarks could be a good place for very large sizes (not for testing, but to catch illegal accesses). But Only unit tests are run through memcheck regularly. So, it won't be caught regularly unless it's added in unit tests. Besides, memcheck will be very slow to run on very large inputs.

Almost all of our algorithms aren't tested for INT32 max sizes. So, it may be okay to not add unit tests for this. 
@PointKernel  Were you able to verify with a unit test case locally? (How much time does it take to run?)

PointKernel:
> @PointKernel Were you able to verify with a unit test case locally? (How much time does it take to run?)

I wrote a test as below and realized bitmask is a special case: the loop end condition is no smaller than the number of bitmask words which is `total_bits / num_bits_per_word`. Since `total_bits` is `size_type`, the thread index would almost never overflow (unless we change the "word" type to something 1 bit only). 
```cpp
TEST_F(SetBitmaskTest, index_overflow)
{
  auto const begin = 0;
  auto const end   = INT_MAX - 10;
  auto const valid = true;
  auto const size  = end - begin;

  thrust::host_vector<bool> expected(size, valid);
  rmm::device_buffer mask = create_null_mask(size, cudf::mask_state::UNINITIALIZED);

  auto bitmask = static_cast<cudf::bitmask_type*>(mask.data());
  cudf::set_null_mask(bitmask, begin, end, valid);

  auto stream = cudf::get_default_stream();

  rmm::device_uvector<bool> output(size, stream);
  auto counting_iter = thrust::counting_iterator<cudf::size_type>{0};
  thrust::transform(rmm::exec_policy(stream),
                    counting_iter,
                    counting_iter + size,
                    output.begin(),
                    valid_bit_functor{bitmask});

  auto const result = thrust::all_of(
    rmm::exec_policy(stream), output.begin(), output.end(), thrust::identity<bool>{});

  EXPECT_EQ(result, valid);
}
```

Not sure if it's still relevant, to answer your question about runtime, the test took about 700 ms to run.
```bash
(base) yunsongw@yunsongw-dt:~/dev/rapids/cudf/cpp/build/release/gtests$ ./BITMASK_TEST --gtest_filter=SetBitmaskTest.index_overflow
Note: Google Test filter = SetBitmaskTest.index_overflow
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SetBitmaskTest
[ RUN      ] SetBitmaskTest.index_overflow
[       OK ] SetBitmaskTest.index_overflow (708 ms)
[----------] 1 test from SetBitmaskTest (708 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (708 ms total)
[  PASSED  ] 1 test.
```

ttnghia (approved):

cpp/src/bitmask/null_mask.cu
mythrocks: I think I understand, but I'd better confirm:
1. `first_word_index` and `last_word_index` are `int32_t`.
2. `tid` is `uint32_t`. So `thread_word_index` is eventually `uint32_t`.
3. `while (thread_word_idx <= last_word_index)` compares the two as `uint32_t`.
4. The place where `thread_word_index` might overflow is at line 275 (`thread_word_index += stride`). But both `thread_word_index` and `stride` are `uint32_t` already.

mythrocks (approved):
LGTM!

ttnghia (approved):

karthikeyann (approved):

PR #13279, opened 2023-05-02
Add an event handler for ColumnVector.close

abellina:
This adds an event handler for `ColumnVector` to be invoked when `.close()` is invoked.

I also made `getRefCount` public since it is useful for debugging and I couldn't think of a good reason to keep this hidden.

This event handler is used to keep track of what columns are spillable at a time and is part of https://github.com/NVIDIA/spark-rapids/issues/7672.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13898, opened 2023-08-17
[JNI] Adds HostColumnVector.EventHandler for spillability checks

abellina:
This change adds the exact same API for event handling that was added to the device-side `ColumnVector` here https://github.com/rapidsai/cudf/pull/13279. We are going to need this to make `HostColumnVector`, or a batch of them,  spillable in this spark-rapids feature: https://github.com/NVIDIA/spark-rapids/issues/8882.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -543,6 +543,7 @@ dependencies:
             packages:
               - cuda-version=12.0
               - cuda-nvtx
+              - cuda-nvtx-dev
           - matrix:
               cuda: "11.8"
             packages:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -289,12 +289,14 @@ dependencies:
               - cuda-version=12.0
               - cuda-cudart-dev
               - cuda-nvrtc-dev
+              - cuda-nvtx-dev
               - libcurand-dev
           - matrix:
               cuda: "11.8"
             packages:
               - cuda-version=11.8
               - cudatoolkit
+              - cuda-nvtx=11.8
               - libcurand-dev=10.3.0.86
               - libcurand=10.3.0.86
           - matrix:
@@ -302,6 +304,7 @@ dependencies:
             packages:
               - cuda-version=11.5
               - cudatoolkit
+              - cuda-nvtx=11.5
                 # Can't hard pin the version since 11.x is missing many
                 # packages for specific versions
               - libcurand-dev>=10.2.6.48,<=10.2.7.107
@@ -311,6 +314,7 @@ dependencies:
             packages:
               - cuda-version=11.4
               - cudatoolkit
+              - &cudanvtx114 cuda-nvtx=11.4
               - &libcurand_dev114 libcurand-dev>=10.2.5.43,<=10.2.5.120
               - &libcurand114 libcurand>=10.2.5.43,<=10.2.5.120
           - matrix:
@@ -321,6 +325,7 @@ dependencies:
                 # The NVIDIA channel doesn't publish pkgs older than 11.4 for
                 # these libs, so 11.2 uses 11.4 packages (the oldest
                 # available).
+              - *cudanvtx114
               - *libcurand_dev114
               - *libcurand114
       - output_types: conda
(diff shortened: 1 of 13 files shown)


abellina:
Tests pass for me locally. I am seeing a CI failure related to a missing nvtx3 header, but I wasn't upmerged. Trying again:

```
     [exec] /opt/conda/envs/test/include/cudf/detail/nvtx/nvtx3.hpp:33:10: fatal error: nvtx3/nvToolsExt.h: No such file or directory
```

abellina:
Failed again, cuDF jni isn't compiling in CI, but it runs fine locally :( It seems the `nvToolsExt` header is no longer available in the conda environment...?

java/src/main/java/ai/rapids/cudf/HostColumnVector.java
```diff
@@ -39,12 +39,31 @@
...
+  public interface EventHandler {
+    /**
+     * `onClosed` is invoked with the updated `refCount` during `close`.
+     * The last invocation of `onClosed` will be with `refCount=0`.
+     *
+     * @note the callback is invoked with this `HostColumnVector`'s lock held.
+     *
+     * @param cv - a reference to the HostColumnVector we are closing
+     * @param refCount - the updated ref count for this HostColumnVector at
+     *                   the time of invocation
```
ttnghia, line 55:
```suggestion
     * @param cv Reference to the HostColumnVector we are closing
     * @param refCount The updated ref count for this HostColumnVector at
     *        the time of invocation
```

java/src/main/java/ai/rapids/cudf/HostColumnVector.java
```diff
@@ -39,12 +39,31 @@
...
+  /**
+   * Interface to handle events for this HostColumnVector. Only invoked during
+   * close, hence `onClosed` is the only event.
+   */
+  public interface EventHandler {
+    /**
+     * `onClosed` is invoked with the updated `refCount` during `close`.
+     * The last invocation of `onClosed` will be with `refCount=0`.
+     *
+     * @note the callback is invoked with this `HostColumnVector`'s lock held.
```
ttnghia, line 51:
```suggestion
     * @note The callback is invoked with this `HostColumnVector`'s lock held.
```

abellina:
I am trying a suggestion from @robertmaynard here https://github.com/rapidsai/cudf/pull/13898/commits/6fb3f1b8a702aa03bec65f7163592cb6651273cd to get the conda environment setup for the java test run with cuda 12, which is the one that is failing because the nvtx header we need is in a different conda package now.

dependencies.yaml
```diff
@@ -543,6 +543,7 @@ dependencies:
             packages:
               - cuda-version=12.0
               - cuda-nvtx
+              - cuda-nvtx-dev
```
revans2, line 546: where did this come from?

gerashegalov (approved):
LGTM

revans2 (commented):
Looks good, just curious about the deps.

revans2 (approved):

java/src/main/java/ai/rapids/cudf/HostColumnVector.java
abellina: We tend to favor starting with lowercase and we also favor indenting the description so that it reads as one block, but I do think the hyphen is discouraged. How about a compromise? (https://github.com/rapidsai/cudf/pull/13898/commits/a0944eebf501006f9f542e09a9dfc1934dbdf4b2)

ttnghia: It's not very important since we are not using the generated docds. But if we are the users who read the generated docs, we will see they are ugly with bad style: Sentences start without first letter being Capped.

ttnghia: and sentences starting without cap. letter is not good IMO.

abellina: Since our code styles on this are all over the place, I am going to suggest filing a follow on if you feel strongly about it. We should then in that issue fix all the comments and make sure that the pre commit hooks exercise them, so we have something consistent and easy to test.

abellina:
@robertmaynard that worked! thanks! Also I do need an approval from `ops-codeowners`

abellina:
@raydouglass would you mind taking a look at this from the ops-codeowner perspective?  The change that prompted this review request was: https://github.com/rapidsai/cudf/pull/13898/commits/6fb3f1b8a702aa03bec65f7163592cb6651273cd

abellina:
I'll wait for https://github.com/rapidsai/cudf/pull/13901/files#diff-5475a6d76de4c506ee92cf6f941bba30a6a07b5881cfea531d42a6ec035095a6 then re-test and rollback my change here likely.

abellina:
Removed `ops-codeowners` because this patch no longer has diffs in dependencies.yaml, it's all java files.

PR #13901, opened 2023-08-17
Use cuda-nvtx-dev CUDA 12 package.

bdice:
## Description
Fixes the CUDA 12 `cuda-nvtx-dev` package name. This issue was found by @davidwendt and I verified that this fixes it.

## Checklist

### DIFF
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -16,6 +16,7 @@ dependencies:
 - cachetools
 - cmake>=3.26.4
 - cubinlinker
+- cuda-nvtx=11.8
 - cuda-python>=11.7.1,<12.0a0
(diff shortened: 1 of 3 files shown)


vuule (approved):

jakirkham:
Thanks Bradley! 🙏

Do we need to rerun `rapids-dependency-file-generator`?

raydouglass (approved):

bdice:
@jakirkham Good catch. There were no changes being rendered to the environment, which was not what I expected. I found and fixed several issues with the placement of cuda-nvtx in the dependency lists in commit d5005dbb0c596cd44cf96455585eaffbd8404a93. We should've known this was a problem before because we had a workaround specific to cudf's Java builds, which is now corrected for all builds.

jakirkham (approved):
Thanks Bradley! 🙏

PR #13903, opened 2023-08-17
Remove the internal use of the cudf's default stream in cuIO

vuule:
## Description
`cudf::get_default_stream()` is used as the default stream when calling `detail` APIs. From this point, the stream passed to this `detail` implementation should be propagated and used consistently. The exceptions are the parts of the implementation that use a separate stream pool.
All in all, libcudf code should not use `cudf::get_default_stream()` outside of dispatching the detail APIs. This PR removes most of such uses in cuIO.
One notable exception is `datasource::host_read`, as it simply does not have a stream parameter (and adding it would be a breaking change).
Also removed comments that mention default stream value that no longer exists.

## Checklist


bdice (approved):
Code looks fine -- just make sure this is internal-only, otherwise mark it as `breaking`.

ttnghia (approved):

PR #13904, opened 2023-08-17
Add noSanitizer tag to Java reduction tests failing with sanitizer in CUDA 12

jlowe:
## Description
Relates to NVIDIA/spark-rapids-jni#1349.  The Java ReductionTest unit tests are failing when run under CUDA 12's compute-sanitizer but pass when run with the CUDA 11 version.  To unblock CI, marking the affected tests to be run without the sanitizer in the interim while this is being investigated.

## Checklist

### DIFF
--- a/java/src/test/java/ai/rapids/cudf/ReductionTest.java
+++ b/java/src/test/java/ai/rapids/cudf/ReductionTest.java
@@ -18,6 +18,7 @@
 package ai.rapids.cudf;
 
 import com.google.common.collect.Lists;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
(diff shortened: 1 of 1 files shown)


ttnghia (approved):

java/src/test/java/ai/rapids/cudf/ReductionTest.java
```diff
@@ -430,6 +431,7 @@ private static void assertEqualsDelta(ReductionAggregation op, Scalar expected,
     }
   }
 
+  @Tag("noSanitizer")
```
gerashegalov, line 434: Since it's passing under CUDA11 Sanitizer, maybe we can call the tag accordingly `cuda11Sanitizer`. If the user has access to both 11 and 12 on CI they can choose accordingly?

jlowe: Sure, although when the tag was created it wasn't expected the sanitizer would run so differently across CUDA versions.  Note that this is just a short-term fix to unblock CI, we don't have plans to run it in a split form long-term unless it's a bug in compute-sanitizer itself that we would need to work around.

gerashegalov (approved):

PR #13905, opened 2023-08-17
Use cudf::thread_index_type in replace.cu.

bdice:
## Description
This PR uses `cudf::thread_index_type` in `replace.cu` to avoid risk of overflow.

## Checklist


cpp/src/replace/replace.cu
```diff
@@ -126,41 +126,42 @@ __global__ void replace_strings_first_pass(cudf::column_device_view input,
...
     uint32_t bitmask = __ballot_sync(active_mask, output_is_valid);
     if (0 == lane_id) {
-      output_valid[cudf::word_index(i)] = bitmask;
+      output_valid[cudf::word_index(idx)] = bitmask;
       valid_sum += __popc(bitmask);
     }
 
-    i += blockDim.x * gridDim.x;
-    active_mask = __ballot_sync(active_mask, i < nrows);
+    tid += blockDim.x * gridDim.x;
```
PointKernel, line 163: In cuco, we explicitly set `blockDim.x * gridDim.x;` to `thread_index_type` type as well to minimize the number of implicit type conversions:
```cuda
  thread_index_type const stride = blockDim.x * gridDim.x;
  ...
  tid += stride;
```
This shouldn't affect runtime performance though.

PointKernel (approved):
LGTM

cpp/src/replace/replace.cu
davidwendt: The `blockDim` and `gridDim` are `dim3` types with members `unsigned x,y,z`. 
The multiplication may not promote these values to `int64_t` before multiplying.
https://godbolt.org/z/f39j9YWEh

cpp/src/replace/replace.cu
```diff
@@ -126,41 +126,42 @@ __global__ void replace_strings_first_pass(cudf::column_device_view input,
                                            cudf::bitmask_type* output_valid,
                                            cudf::size_type* __restrict__ output_valid_count)
 {
-  cudf::size_type nrows = input.size();
-  cudf::size_type i     = blockIdx.x * blockDim.x + threadIdx.x;
-  uint32_t active_mask  = 0xffff'ffffu;
-  active_mask           = __ballot_sync(active_mask, i < nrows);
+  cudf::size_type nrows       = input.size();
+  cudf::thread_index_type tid = blockIdx.x * blockDim.x + threadIdx.x;
```
vuule, line 130: This should be
```suggestion
  cudf::thread_index_type tid = blockIdx.x * blockDim.x + cudf::thread_index_type{threadIdx.x};
```
so we promote before adding the stride?

bdice: I think promoting `blockIdx.x` first would be the best move. That way the multiplication is done in `int64_t` (which gets the addition, too).
```suggestion
  cudf::thread_index_type tid = static_cast<cudf::thread_index_type>(blockIdx.x) * blockDim.x + threadIdx.x;
```

bdice: @vuule Do you agree with that analysis?

vuule: We're wandering close to the shores of the nitpick sea, but... could `blockIdx.x * blockDim.x` ever overflow? I assume we don't launch kernels with blocks that are completely outside of the row range. Also, the type is called `thread_index_type`, to me it looks odd applied to the block index.

vuule: This is what I'm going with 
https://github.com/rapidsai/cudf/pull/13910/files#diff-c0d67e637bd07fdcf4833b0da7fb7aa7aebd6f992736416a00d86f9c5e6ee32c
would be nice to align somehow :D

davidwendt: Promoting just the addend does not seem to be enough to promote the multiplication
https://godbolt.org/z/coqPKqfWe

vuule: sure, but when do we need to promote the multiplication?

davidwendt: Sorry, I thought the point of going with `thread_index_type` was for this very purpose.
Otherwise, we should be ok with `size_type` or perhaps `uint32_t` if the overflow never actually occurs?
Maybe I misunderstood what we are doing this for.

vuule: Per offline discussion, it seems like I'm just not forward-looking wrt `size_type` size.

cpp/src/replace/replace.cu
```diff
@@ -127,40 +127,42 @@ __global__ void replace_strings_first_pass(cudf::column_device_view input,
                                            cudf::size_type* __restrict__ output_valid_count)
 {
   cudf::size_type nrows = input.size();
-  cudf::size_type i     = blockIdx.x * blockDim.x + threadIdx.x;
-  uint32_t active_mask  = 0xffff'ffffu;
-  active_mask           = __ballot_sync(active_mask, i < nrows);
+  auto tid = cudf::thread_index_type{blockIdx.x} * cudf::thread_index_type{blockDim.x} +
+             cudf::thread_index_type{threadIdx.x};
```
ttnghia, line 131: `cudf::detail::grid_1d::global_thread_id();`?

cpp/src/replace/replace.cu
```diff
@@ -127,40 +127,42 @@ __global__ void replace_strings_first_pass(cudf::column_device_view input,
...
     uint32_t bitmask = __ballot_sync(active_mask, output_is_valid);
     if (0 == lane_id) {
-      output_valid[cudf::word_index(i)] = bitmask;
+      output_valid[cudf::word_index(idx)] = bitmask;
       valid_sum += __popc(bitmask);
     }
 
-    i += blockDim.x * gridDim.x;
-    active_mask = __ballot_sync(active_mask, i < nrows);
+    tid += cudf::thread_index_type{blockDim.x} * cudf::thread_index_type{gridDim.x};
```
ttnghia, line 164: The mul operation can be extracted like:
```
auto const stride = blockDim.x * gridDim.x;
......
for(....) {
  ...
  tid += stride;
}

File: cpp/src/replace/replace.cu [hunk 5]
```diff
@@ -189,27 +191,32 @@ __global__ void replace_strings_second_pass(cudf::column_device_view input,
                                             cudf::mutable_column_device_view indices)
 {
   cudf::size_type nrows = input.size();
-  cudf::size_type i     = blockIdx.x * blockDim.x + threadIdx.x;
+  auto tid = cudf::thread_index_type{blockIdx.x} * cudf::thread_index_type{blockDim.x} +
+             cudf::thread_index_type{threadIdx.x};
```
ttnghia, line 195: Similar to above.

cpp/src/replace/replace.cu
```diff
@@ -189,27 +191,32 @@ __global__ void replace_strings_second_pass(cudf::column_device_view input,
...
+      cudf::string_view output = (replace_idx == -1)
+                                   ? input.element<cudf::string_view>(idx)
+                                   : replacement.element<cudf::string_view>(replace_idx);
+      std::memcpy(strings.data<char>() + offsets.data<cudf::size_type>()[idx],
                   output.data(),
                   output.size_bytes());
     }
 
-    i += blockDim.x * gridDim.x;
+    tid += cudf::thread_index_type{blockDim.x} * cudf::thread_index_type{gridDim.x};
```
ttnghia, line 219: Same here.

cpp/src/replace/replace.cu
bdice: This is resolved by #13996 and associated utility functions.

cpp/src/replace/replace.cu
bdice: This is resolved by https://github.com/rapidsai/cudf/pull/13996 and associated utility functions.

vuule (approved):

ttnghia (approved):

PR #13910, opened 2023-08-17
Use `cudf::thread_index_type` in cuIO to prevent overflow in row indexing

vuule:
## Description
Use wider type for indexing when rows are indexed in pattern 
```
for (auto row = start_row; row < num_rows; row += block_size) {
if (is_within_bounds) ...
}
```
or
```
auto t = threadIdx.x;
auto row = block_start + t;
if (is_within_bounds) ...
```
Overflow can happen when the number of rows is so close to `max<size_type>` that adding block size pushes the index over the max.

Also sprinkled auto where increased size is not needed.
Also removed a few redundant conditions.

## Checklist


cpp/src/io/orc/dict_enc.cu
```diff
@@ -217,9 +217,7 @@ __global__ void __launch_bounds__(block_size)
 
   auto cur_row = start_row + t;
   while (cur_row < end_row) {
-    auto const is_valid = cur_row < col.size() and col.is_valid(cur_row);
```
vuule, line 220: redundant check with `cur_row < end_row`

cpp/src/io/csv/csv_gpu.cu
```diff
@@ -317,8 +316,8 @@ __global__ void __launch_bounds__(csvparse_block_dim)
   auto const raw_csv = data.data();
   // thread IDs range per block, so also need the block id.
   // this is entry into the field array - tid is an elements within the num_entries array
-  long const rec_id      = threadIdx.x + (blockDim.x * blockIdx.x);
-  long const rec_id_next = rec_id + 1;
+  auto const rec_id      = cudf::thread_index_type{threadIdx.x} + (blockDim.x * blockIdx.x);
```
davidwendt, line 319: Promoting just the `threadId.x` does not appear to be enough to promote the multiplication
https://godbolt.org/z/coqPKqfWe

vuule: moved the promotion to the multiplication

cpp/include/cudf/detail/utilities/cuda.cuh
```diff
@@ -65,6 +65,12 @@ class grid_1d {
     CUDF_EXPECTS(num_threads_per_block > 0, "num_threads_per_block must be > 0");
     CUDF_EXPECTS(num_blocks > 0, "num_blocks must be > 0");
   }
+  static constexpr thread_index_type global_thread_id(unsigned int thread_id,
+                                                      unsigned int block_id,
+                                                      unsigned int num_threads_per_block)
+  {
+    return thread_id + thread_index_type{block_id} * num_threads_per_block;
+  }
```
PointKernel, line 73:
```suggestion
  static constexpr thread_index_type global_thread_id()
  {
    return threadIdx.x + thread_index_type{blockIdx.x} * blockDim.x;
  }
```
Correct me if I'm wrong, this is slightly simpler and would work also. 

Can you please add docs as well?

cpp/src/io/orc/dict_enc.cu
```diff
@@ -130,7 +130,7 @@ __global__ void __launch_bounds__(block_size)
   size_type entry_count{0};
   size_type char_count{0};
   // all threads should loop the same number of times
-  for (auto cur_row = start_row + t; cur_row - t < end_row; cur_row += block_size) {
+  for (int64_t cur_row = start_row + t; cur_row - t < end_row; cur_row += block_size) {
```
PointKernel, line 133:
```suggestion
  for (thread_index_type cur_row = start_row + t; cur_row - t < end_row; cur_row += block_size) {
```
prefer `thread_index_type` over plain integer type. The same to other use cases.

cpp/include/cudf/detail/utilities/cuda.cuh
```diff
@@ -65,6 +65,12 @@ class grid_1d {
     CUDF_EXPECTS(num_threads_per_block > 0, "num_threads_per_block must be > 0");
     CUDF_EXPECTS(num_blocks > 0, "num_blocks must be > 0");
   }
+  static constexpr thread_index_type global_thread_id(unsigned int thread_id,
+                                                      unsigned int block_id,
+                                                      unsigned int num_threads_per_block)
+  {
+    return thread_id + thread_index_type{block_id} * num_threads_per_block;
```
harrism, line 72: Why not upcast all 3 parameters to thread_index_type?

cpp/include/cudf/detail/utilities/cuda.cuh
```diff
@@ -65,6 +65,12 @@ class grid_1d {
     CUDF_EXPECTS(num_threads_per_block > 0, "num_threads_per_block must be > 0");
     CUDF_EXPECTS(num_blocks > 0, "num_blocks must be > 0");
   }
+  static constexpr thread_index_type global_thread_id(unsigned int thread_id,
```
harrism, line 68: Blank line between functions (and also document this).

```suggestion

  static constexpr thread_index_type global_thread_id(unsigned int thread_id,
```

cpp/include/cudf/detail/utilities/cuda.cuh
harrism: Could provide both overloads, in case they want to customize.  But @PointKernel's solution should be the common case.

cpp/src/io/orc/stripe_data.cu
```diff
@@ -1206,7 +1206,7 @@ __global__ void __launch_bounds__(block_size)
       if (row_in < first_row && t < 32) {
         uint32_t skippedrows = min(static_cast<uint32_t>(first_row - row_in), nrows);
         uint32_t skip_count  = 0;
-        for (uint32_t i = t * 32; i < skippedrows; i += 32 * 32) {
+        for (int64_t i = t * 32; i < skippedrows; i += 32 * 32) {
```
harrism, line 1209: `t` is defined above as `int`. Should it not be `thread_index_type` since it is assigned `threadIdx.x`?

harrism (changes requested):
Looks good, just a couple of comments.

cpp/src/io/orc/stripe_data.cu
vuule: Not sure if I understood the comment correctly.
Should every variable based on `threadIdx.x` be a `thread_index_type`?
I tried to keep changes to a minimum; since `t` is never larger than the number of row, it's not an overflow concern.

cpp/include/cudf/detail/utilities/cuda.cuh
vuule: Thank you for the suggestions.
Added an overload with `threadIdx.x`, `blockIdx.x`, `blockDim.x`.

harrism (approved):

PR #13911, opened 2023-08-17
Translate column size overflow exception to JNI

mythrocks:
When a CUDF operation causes a column's row count to exceed the size limit imposed by `cudf::size_type`, the operation throws a `std::overflow_error` exception. However, prior to this commit, CUDF JNI did not translate this to a separate Java exception. Because of handling this condition as any generic exception, there was no way to attempt case specific recovery for overflow conditions.

This commit translates `std::overflow_error` into a new Java exception (`CudfColumnOverflowException`) that may be caught in user space to attempt recovery/retry.

This is a non-breaking change. The user-facing change is minimal in that existing failure handling based on catching `CudfException` will continue to work as before. The user will now have more fine grained error handling by catching `CudfColumnOverflowException`.

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/CudfColumnOverflowException.java
+++ b/java/src/main/java/ai/rapids/cudf/CudfColumnOverflowException.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2023, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ai.rapids.cudf;
+
+/**
+ * Exception thrown when CUDF operation results in a column size
+ * exceeding CUDF column size limits
+ */
+public class CudfColumnOverflowException extends CudfException {
+  CudfColumnOverflowException(String message) {
+    super(message);
+  }
+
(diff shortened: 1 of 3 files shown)


java/src/main/native/include/jni_utils.hpp
```diff
@@ -901,6 +902,10 @@ inline void jni_cuda_check(JNIEnv *const env, cudaError_t cuda_status) {
     JNI_CHECK_THROW_CUDF_EXCEPTION(env, cudf::jni::CUDF_DTYPE_ERROR_CLASS, e.what(),               \
                                    e.stacktrace(), ret_val);                                       \
   }                                                                                                \
+  catch (std::overflow_error const &e) {                                                           \
+    JNI_CHECK_THROW_CUDF_EXCEPTION(env, cudf::jni::CUDF_OVERFLOW_ERROR_CLASS, e.what(),            \
```
ttnghia, line 906: `JNI_CHECK_THROW_CUDF_EXCEPTION` needs to access the Java exception with constructor `(String, String)`.

java/src/main/java/ai/rapids/cudf/CudfColumnOverflowException.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package ai.rapids.cudf;
+
+/**
+ * Exception thrown when CUDF operation results in a column size
+ * exceeding CUDF column size limits
+ */
+public class CudfColumnOverflowException extends CudfException {
```
ttnghia, line 22: I feel that "ColumnOverflow" is not a good name. Maybe `....SizeOverflow....`?

mythrocks: Right. I was thinking that also `ColumnSizeOverflow`.

mythrocks: I have renamed this to `CudfColumnSizeOverflowException`. It's a mouthful, but it's more accurate.

gerashegalov (approved):

ttnghia (approved):

PR #13914, opened 2023-08-18
Enable hugepage for arrow host allocations 

madsbk:
This PR enables Transparent Huge Pages (THP) for large (>4MB) arrow allocations (host memory only).

### Performance results  on a DGX-1 (`dgx14`) 
|                |  8MB  |  80MB | 800MB |  8GB  | Method                             |
|:--------------:|:-----:|:-----:|:-----:|:-----:|------------------------------------|
| cudf-native    | 0.006 | 0.049 | 0.485 | 4.787 | `df.to_arrow()` (branch-23.10)     |
| Dask-serialize | 0.004 | 0.032 | 0.310 | 3.122 | `distributed.protocol.serialize(df)` |
| cudf-hugepage  | 0.004 | 0.030 | 0.299 | 3.046 | `df.to_arrow()` (this PR)          |
| speedup        | 1.5   | 1.63  | 1.62  | 1.57  |  cudf-native vs. cudf-hugepage         |

Notice, Dask-serialize also use THP, which is why its performance is on par with cudf-hugepage.

## Checklist

cc. @pentschev (I think you did something similar in NumPy)


cpp/src/interop/detail/arrow_allocator.cpp
```diff
@@ -30,7 +61,7 @@ std::unique_ptr<arrow::Buffer> allocate_arrow_buffer(int64_t const size, arrow::
   */
   auto result = arrow::AllocateBuffer(size, ar_mr);
   CUDF_EXPECTS(result.ok(), "Failed to allocate Arrow buffer");
-  return std::move(result).ValueOrDie();
+  return enable_hugepage(std::move(result).ValueOrDie());
```
harrism, line 64: What is the reason for `std::move` here? I'm not sure you need that.

Generally `return std::move(...)` is an antipattern. When you are returning a local variable, usually the copy will be elided anyway. If there is a move ctor, it can be used automatically.

harrism (commented):
One comment.

cpp/src/interop/detail/arrow_allocator.cpp
madsbk: Notice, we use `std::move` on the intermediate `result`, not the value returned by `ValueOrDie()`. This _moves_ the value (`std::unique_ptr<arrow::Buffer>`) out of `result` .
I have added explicit types to make this more clear.

harrism: Right, I was assuming you just copied the std::move from the code that was there before, which just had `return std::move`, which I think was unnecessary (and disables the compiler from eliding the move with RVO, IIUC).

harrism (approved):

cpp/src/interop/detail/arrow_allocator.cpp
vuule: The compiler does not allow `return std::move(retval)` in libcudf because it disables RVO. But `std::move(result).ValueOrDie()` is slightly different, as it does not move the return value, just casts `result` to rvalue, presumably so that `ValueOrDie` can move the value out of `result`. The moved value is most likely then RVOd to the caller (not sure how this is actually called :)).

cpp/src/interop/detail/arrow_allocator.cpp
```diff
@@ -16,9 +16,40 @@
...
+{
+  if (buf->size() < (1u << 22u)) {  // Smaller than 4 MB
+    return buf;
+  }
+
+#ifdef MADV_HUGEPAGE
+  const auto pagesize = sysconf(_SC_PAGESIZE);
+  void* addr          = const_cast<uint8_t*>(buf->data());
+  if (addr == nullptr) { return buf; }
+  std::size_t length{static_cast<std::size_t>(buf->size())};
```
vuule, line 43: The type is already spelled out in the same line
```suggestion
  auto length{static_cast<std::size_t>(buf->size())};
```

cpp/src/interop/detail/arrow_allocator.cpp
```diff
@@ -16,9 +16,40 @@
...
 namespace detail {
 
+/*
+  Enable Transparent Huge Pages (THP) for large (>4MB) allocations.
+  `buf` is returned untouched.
+  Enabling THP can improve performance of device-host memory transfers
+  significantly, see <https://github.com/rapidsai/cudf/pull/13914>.
+*/
+template <typename T>
+T enable_hugepage(T buf)
```
vuule, line 33: Should we limit the type here, since we don't want to ever copy T when calling the function?
```suggestion
T enable_hugepage(T&& buf)
```

cpp/src/interop/detail/arrow_allocator.cpp
```diff
@@ -42,9 +73,9 @@ std::shared_ptr<arrow::Buffer> allocate_arrow_bitmap(int64_t const size, arrow::
   To work around this issue we compile an allocation shim in C++ and use
   that from our cuda sources
   */
-  auto result = arrow::AllocateBitmap(size, ar_mr);
+  arrow::Result<std::shared_ptr<arrow::Buffer>> result = arrow::AllocateBitmap(size, ar_mr);
```
vuule, line 76: is this just for readability?

vuule (commented):
a few questions/suggestions

cpp/src/interop/detail/arrow_allocator.cpp
madsbk: Yes, just to make the `arrow::Result` wrapper obvious

vuule:
Looks like RVO is not happy with the recent parameter type change to T&&?

cpp/src/interop/detail/arrow_allocator.cpp
```diff
@@ -16,9 +16,40 @@
...
 namespace detail {
 
+/*
+  Enable Transparent Huge Pages (THP) for large (>4MB) allocations.
+  `buf` is returned untouched.
+  Enabling THP can improve performance of device-host memory transfers
+  significantly, see <https://github.com/rapidsai/cudf/pull/13914>.
+*/
+template <typename T>
+T enable_hugepage(T&& buf)
```
vuule, line 33: could we have enable_hugepage just take a host_span? Do we really need to forward `std::unique_ptr<arrow::Buffer>` through the function?

madsbk: We don't need to but it means we can avoid an in-place function and have a simply return wrapper `return enable_hugepage(...)`.

vuule: I still don't see the benefit, and I'm fine with that.

PR #13919, opened 2023-08-18
Add java API to get size of host memory needed to copy column view

revans2:
## Description
To help with work for host memory management in java this provides an API to know how much memory is needed on the host to copy the data before it happens.

This was written by @jbrennan333 but I am taking over the patch to get it in.

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/ColumnView.java
+++ b/java/src/main/java/ai/rapids/cudf/ColumnView.java
@@ -5161,6 +5161,58 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
   }
 
   /**
+   * Calculate the total space required to copy the data to the host.
+   */
+  public long getHostBytesRequired() {
+    return getHostBytesRequiredHelper(this);
+  }
+
+  /*
+   * Align given size to account for host allocation alignment of 64
+   */
+  private static long alignAllocSize(long size) {
+    final long align = 64; // must be a power of two
+    return (size + (align - 1)) & ~(align - 1);
+  }
+
+  private static long getHostBytesRequiredHelper(
+      ColumnView deviceCvPointer) {
+    if (deviceCvPointer == null) {
+      return 0;
+    }
+    BaseDeviceMemoryBuffer valid = deviceCvPointer.getValid();
+    BaseDeviceMemoryBuffer offsets = deviceCvPointer.getOffsets();
+    BaseDeviceMemoryBuffer data = null;
+    DType type = deviceCvPointer.getType();
+    if (!type.isNestedType()) {
+      data = deviceCvPointer.getData();
+    }
+    long validityLength = 0;
+    long offsetsLength = 0;
+    long dataLength = 0;
+    long childrenLength = 0;
+    if (valid != null) {
+      validityLength = alignAllocSize(valid.getLength());
+    }
+    if (offsets != null) {
+      offsetsLength = alignAllocSize(offsets.getLength());
+    }
+    // If a strings column is all null values there is no data buffer allocated
+    if (data != null) {
+      dataLength = alignAllocSize(data.length);
+      data.close();
+    }
+    if (type.isNestedType()) {
+      for (int i = 0; i < deviceCvPointer.getNumChildren(); i++) {
+        try (ColumnView childDevPtr = deviceCvPointer.getChildColumnView(i)) {
+          childrenLength += getHostBytesRequiredHelper(childDevPtr);
+        }
+      }
+    }
+    return validityLength + offsetsLength + dataLength + childrenLength;
+  }
+
+  /**
    * Exact check if a column or its descendants have non-empty null rows
    *
    * @return Whether the column or its descendants have non-empty null rows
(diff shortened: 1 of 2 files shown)


ttnghia (approved):

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -5160,6 +5160,58 @@ public HostColumnVector copyToHost() {
...
+   */
+  public long getHostBytesRequired() {
+    return getHostBytesRequiredHelper(this);
+  }
+
+  /*
+   * Align given size to account for host allocation alignment of 64
+   */
+  private static long alignAllocSize(long size) {
+    final long align = 64; // must be a power of two
```
gerashegalov, line 5174: nit: a power of two requirement can be emphasized
with a `1 << 6` or a binary literal `0B1000000`

Where is 64 coming from? Is there a mnemonic constant for this?  Can this be represented as some DTypeEnum.<type>64.sizeInBytes * 8?

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -5160,6 +5160,58 @@ public HostColumnVector copyToHost() {
...
+    return (size + (align - 1)) & ~(align - 1);
```
gerashegalov, line 5175: maybe comment that this is the closest size that is a multiple of `align`?

revans2:
@gerashegalov and @ttnghia sorry to change the code on you. I realized that we already had some C++ code to do almost the exact same thing, so I switched over to use it instead. Just to remove duplication.

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -307,7 +307,15 @@ public final int getNumChildren() {
...
   public long getDeviceMemorySize() {
-    return getDeviceMemorySize(getNativeView());
+    return getDeviceMemorySize(getNativeView(), false);
+  }
+
+  /**
+   * Returns the amount of memory used by this, but padded for 64-bit alignment. This makes it
+   * so it could be used as the amount of memory needed to copy the data to the host.
+   */
+  public long getDeviceMemorySizeAligned() {
```
gerashegalov, line 317: nit: can we incorporate 64 in the method name?

```suggestion
  public long getDeviceMemorySizeAligned64() {
```
or 

```suggestion
  public long getDeviceMemorySizePadded64() {
```

gerashegalov (approved):
LGTM, just optional nits

gerashegalov (approved):

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -5160,6 +5168,13 @@ public HostColumnVector copyToHost() {
     }
   }
 
+  /**
+   * Calculate the total space required to copy the data to the host.
+   */
+  public long getHostBytesRequired() {
```
razajafri, line 5174: nit: why is the method name `getHostBytesRequired ` not indicative that it is returning padded memory? 
Is this an oversight or will that be redundant?

revans2: Technically all of the versions of the APIs would be more accurate when padding the data to 64-bit boundaries. The memory allocation APIs that we use don't take any form of alignment. For C malloc says that the value returned will be aligned at least as much as `std::max_align_t`

https://en.cppreference.com/w/cpp/memory/c/malloc

If we really wanted to be accurate we should use this value when computing the alignment. But most of the time 64-bit is good enough. The GPU has similar requirements for alignment on memory allocations.

We have an existing API for returning GPU memory that matches exactly with what the buffers say their size is. Because of this I thought it best not to change it, but add a new API. For host memory there is no existing API, so I thought it better to just have it default to the more correct thing.

But if someone feels strongly about it I can change it.

java/src/test/java/ai/rapids/cudf/ColumnVectorTest.java
```diff
@@ -1031,7 +1031,9 @@ void testGetDeviceMemorySizeNonStrings() {
     try (ColumnVector v0 = ColumnVector.fromBoxedInts(1, 2, 3, 4, 5, 6);
          ColumnVector v1 = ColumnVector.fromBoxedInts(1, 2, 3, null, null, 4, 5, 6)) {
       assertEquals(24, v0.getDeviceMemorySize()); // (6*4B)
+      assertEquals(64, v0.getHostBytesRequired()); // account for alignment padding
```
razajafri, line 1034: Isn't 24 bytes already 64-bit aligned? 8+8+8 = 24 bytes?

razajafri, line 1034: Isn't 24 bytes already 64-bit aligned? 8+8+8 = 24 bytes?

java/src/main/java/ai/rapids/cudf/ColumnView.java
razajafri:
```
      assertEquals(24, v0.getDeviceMemorySize()); // (6*4B)
      assertEquals(64, v0.getHostBytesRequired()); // account for alignment padding
```
When I read those lines in the test I came to the code to read what is the difference, the doc didn't specify. It's your decision

revans2:
Great catch @razajafri I updated the code to follow the alignment that C++ says it should be.  Turns out, at least for my desktop, it is 32-byte aligned.  So 64-bytes was not too far off for what we really wanted.

razajafri (approved):

razajafri:
build

PR #13920, opened 2023-08-18
Create table_input_metadata from a table_metadata

etseidl:
## Description
When round-tripping data through cuDF (e.g. read a parquet file with `read_parquet()`, then write slices using the `chunked_parquet_writer`) column nullability information can be lost. The parquet writers will accept a `table_input_metadata` object as an optional parameter, and this object can be used to preserve the nullability. Creating the `table_input_metadata` can be a challenge, however. This PR addresses this problem by adding the ability to create a `table_input_metadata` using the `table_metadata` returned by `read_parquet()`. 

## Checklist


cpp/src/io/functions.cpp
```diff
@@ -503,6 +503,11 @@ std::unique_ptr<std::vector<uint8_t>> merge_row_group_metadata(
   return detail_parquet::writer::merge_row_group_metadata(metadata_list);
 }
 
+table_input_metadata table_with_metadata::get_table_input_metadata()
```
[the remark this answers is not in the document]
etseidl: syntactic sugar for the user.  `read_parquet` returns the `table_with_metadata`, so it seems easier for the user (me!) to just call a member on the returned struct than to call the `table_input_metadata` ctor or rely on implicit conversion.  Not a must have.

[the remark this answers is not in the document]
vuule: yeah, that's what I'm thinking. The implicit ctor has to be the least verbose option.
But I'm not sure if it should be implicit.

cpp/tests/io/parquet_test.cpp
```diff
@@ -6599,4 +6599,68 @@ TEST_F(ParquetWriterTest, TimestampMicrosINT96NoOverflow)
...
+    cudf::io::parquet_reader_options::builder(cudf::io::source_info{filepath});
+  auto const result        = cudf::io::read_parquet(in_opts);
+  auto const read_metadata = cudf::io::table_input_metadata{result};
+
+  // test that expected_metadata matches read_metadata
+  std::function<void(cudf::io::column_in_metadata, cudf::io::column_in_metadata)>
+    compare_names_and_nullability = [&](auto lhs, auto rhs) {
+      EXPECT_EQ(lhs.get_name(), rhs.get_name());
+      EXPECT_EQ(lhs.nullable(), rhs.nullable());
+      EXPECT_EQ(lhs.num_children(), rhs.num_children());
```
vuule, line 6652: This one should probably be an ASSERT, test could crash if `lhs.num_children() > rhs.num_children()`.

etseidl: Iright you are

cpp/include/cudf/io/types.hpp
```diff
@@ -201,21 +201,55 @@ enum dictionary_policy {
...
+   * @return True if the nullability is set, false otherwise
+   */
+  bool is_set_nullability() const { return is_nullable.has_value(); }
+
+  /**
+   * @brief Get the nullability for this column.
+   *
+   * @return True if the column is nullable, false if not, or the nullability has not been set.
+   */
+  bool get_nullablity() const { return is_nullable.value_or(false); }
```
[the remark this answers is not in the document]
vuule: If we do want to have this getter, it should return an optional.
But the name does not have a getter/setter, so :shrug:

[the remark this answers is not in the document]
etseidl: I guess accessors are a bit sporty here. I think I'll remove them. Otherwise I'd want to promote this to a class.

cpp/include/cudf/io/types.hpp
```diff
@@ -201,20 +201,33 @@ enum dictionary_policy {
...
    */
   column_name_info(std::string const& _name) : name(_name) {}
+
+  /**
+   * @brief Construct a column name info with a name, nullabilty, and no children
+   *
+   * @param _name Column name
+   * @param _is_nullable True if column is nullable
+   */
+  column_name_info(std::string const& _name, bool _is_nullable)
```
vuule, line 226: Default to unspecified nullability?
```suggestion
  column_name_info(std::string const& _name, std::optional<bool> _is_nullable = std::nullopt)
```

cpp/src/io/utilities/column_buffer.cpp
```diff
@@ -149,14 +149,17 @@ std::unique_ptr<column> make_column(column_buffer_base<string_policy>& buffer,
...
+    schema_info->is_nullable = buffer.is_nullable;
+  }
 
   switch (buffer.type.id()) {
     case type_id::STRING:
       if (schema.value_or(reader_column_schema{}).is_enabled_convert_binary_to_strings()) {
         if (schema_info != nullptr) {
-          schema_info->children.push_back(column_name_info{"offsets"});
-          schema_info->children.push_back(column_name_info{"chars"});
+          schema_info->children.push_back(column_name_info{"offsets", false});
```
vuule, line 161: default value in the ctor would clean up these uses (see other comment).

vuule (commented):
Looks good, aside from a single detail.
Bit sad about `column_name_info` containing nullability. However, it looks like `column_name_info` and related types will change a lot, and this awkward naming is temporary :)

cpp/include/cudf/io/types.hpp
etseidl: I think that will collide with the other ctor?

vuule: made me look!

cpp/src/io/utilities/column_buffer.cpp
etseidl: I think we want these specified explicitly. At least in the past these extra children have been non-nullable. Could be convinced otherwise.

etseidl: ok, removed the nullabilty on cudf-only bits...had to change the test but output schema still seems fine

cpp/include/cudf/io/types.hpp
```diff
@@ -798,7 +805,17 @@ class table_input_metadata {
...
    * @param table The table_view to construct metadata for
    */
-  table_input_metadata(table_view const& table);
+  explicit table_input_metadata(table_view const& table);
+
+  /**
+   * @brief Construct a new table_input_metadata from a table_metadata object.
+   *
+   * The constructed table_input_metadata has the same structure as the passed table_with_metadata,
+   * and also preserves any naming and nullability info from the original table.
```
vuule, line 814:
```suggestion
   * The constructed table_input_metadata has the same structure, column names and nullability as the passed table_metadata.
```

vuule (approved):

cpp/src/io/functions.cpp
```diff
@@ -517,6 +517,26 @@ table_input_metadata::table_input_metadata(table_view const& table)
     table.begin(), table.end(), std::back_inserter(this->column_metadata), get_children);
 }
 
+table_input_metadata::table_input_metadata(table_metadata const& metadata)
+{
+  auto const& names = metadata.schema_info;
+
+  // Create a metadata hierarchy with naming and nullability using `table_and_metadata`
+  std::function<column_in_metadata(column_name_info const&)> get_children =
```
hyperbolic2346, line 525: Very nit picking here, but `get_children` doesn't just get children. It is also setting nullability on the metadata. Maybe something like `process_node` or `set_nullability`?

hyperbolic2346 (approved):
A very small nit related to lambda names, nothing to hold up the review. This seems very useful for round-trips.

cpp/src/io/functions.cpp
etseidl: Cut and paste strikes again :sweat_smile: I like `process_node`

Issue #13509, opened 2023-06-05
[FEA] Improve cudf::gather scalability as number of columns increases


abellina:
As the number of columns increases for `cudf::gather` with the same gather map, we see the number of kernels called increase proportionally and the runtime increases linearly. We are wondering if there are better ways to group or "batch" these calls so we perform less kernel invocations that can do more work all at once, in hopes of amortizing some of the cost with many columns or deeply nested schemas.

A very simple example is below. This creates a column of 10 `int32_t` rows and adds it to a struct N times (where `N` is between 2 and 1024):

```c++
#include <cudf/table/table.hpp>
#include <cudf_test/column_wrapper.hpp>

#include <rmm/mr/device/cuda_memory_resource.hpp>
#include <rmm/mr/device/device_memory_resource.hpp>
#include <rmm/mr/device/pool_memory_resource.hpp>

#include <memory>
#include <string>
#include <vector>
#include <nvtx3/nvToolsExt.h>

int main(int argc, char** argv)
{
  rmm::mr::cuda_memory_resource cuda_mr{};
  rmm::mr::pool_memory_resource mr{&cuda_mr};
  rmm::mr::set_current_device_resource(&mr);
  using col_t = cudf::test::fixed_width_column_wrapper<int32_t>;

  auto const values = std::vector<int32_t>{1,2,3,4,5,6,7,8,9,10};
  for (int num_cols = 2; num_cols <= 1024; num_cols *= 2) {
    std::vector<std::unique_ptr<cudf::column>> members(num_cols);
    for (auto i = 0; i < num_cols; ++i) {
      auto wrapper = col_t(values.begin(), values.end());
      members[i] = wrapper.release();
    }
    auto struct_col = cudf::test::structs_column_wrapper(std::move(members));
    auto gather_map = std::vector<cudf::offset_type>{1}; // gather 1 row
    std::stringstream msg;
    nvtxRangePush(msg.str().c_str()); 
    auto result = cudf::gather(
      cudf::table_view{{struct_col}}, 
      cudf::test::fixed_width_column_wrapper<int32_t>(gather_map.begin(), gather_map.end()),
      cudf::out_of_bounds_policy::NULLIFY);
    nvtxRangePop();
    std::cout << "Result: rows: " << result->num_rows() << " cols: " << result->num_columns() << std::endl;

  }
  return 0;
}
```
As the column count increases by 2x, the gather kernel takes 2x longer:

[...]

PR #13922, opened 2023-08-18
Global stream pool

etseidl:
## Description
#13637 added a static stream pool object for use by the Parquet reader. This PR expands upon that by:

- Moving the stream pool to the `cudf::detail` namespace.
- Adding a debugging implementation that always returns the default stream.
- Hiding implementation details behind a more streamlined interface.
- Using cuda events for synchronization.

## Checklist


etseidl:
The stream pool could be moved to io/utilities and then be used by `multibyte_split` (which a lot of this is taken from anyway).

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,94 @@
...
+// doing lazy initialization to avoid allocating this before cuInit is called (was
+// causing issues with compute-sanitizer).
+// deliberately allowing the pool to leak to avoid deleting streams after cuda shutdown.
+rmm::cuda_stream_pool* pool = nullptr;
+
+std::mutex stream_pool_mutex;
+
+void init()
+{
+  std::lock_guard<std::mutex> lock(stream_pool_mutex);
```
vuule, line 38: I think we don't need a mutex at any point (except possibly in get_streams) if the singleton is implemented with a static variable, similar to `logger_wrapper`.

cpp/src/io/parquet/stream_pool.cpp
vuule, line 1: hear me out - a singleton, abstract, `cudf::stream_pool`, that can wrap different pools, depending on the config. By default we would use rmm stream pool, but there could be a dummy pool that always returns the same (default?) stream, and maybe other types. Takes a bit of extra code, but could help debug later on.

vuule (commented):
parquet code looks so much better with the new pool!
I think we could beef up the pool and make use of it at libcudf level.

cpp/src/io/parquet/stream_pool.cpp
etseidl: I'd be happy to get rid of the `init()` if possible.  Didn't think to try static variable like the old version...I had it defined in the anonymous namespace.

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+std::vector<rmm::cuda_stream_view> get_streams(uint32_t count)
+{
+  init();
+
+  // TODO maybe add mutex to be sure streams don't overlap
+  auto streams = std::vector<rmm::cuda_stream_view>();
+  for (uint32_t i = 0; i < count; i++) {
+    streams.emplace_back(pool->get_stream((stream_idx++)));
```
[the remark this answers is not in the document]
etseidl: I'm trying to minimize the risk of getting duplicate streams in the vector assuming concurrent calls to `get_streams` and the size of the vector is less than `STREAM_POOL_SIZE`. If we wrap this in a mutex, then we don't have to make `stream_idx` an atomic, but I think we should still use a global index so we don't keep handing out the same streams.

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,94 @@
...
+  return pool->get_stream();
+}
+
+rmm::cuda_stream_view get_stream(std::size_t stream_id)
+{
+  init();
+  return pool->get_stream(stream_id);
+}
+
+std::vector<rmm::cuda_stream_view> get_streams(uint32_t count)
```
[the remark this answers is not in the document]
etseidl: I'm good with a warning...then at least we'd know if we need to beef up `STREAM_POOL_SIZE`. 

I wish the stream pool was growable...but then `get_stream(std::size_t stream_id)` would no longer be able to guarantee the same stream for a given id.

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+  init();
+
+  // TODO maybe add mutex to be sure streams don't overlap
```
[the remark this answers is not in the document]
etseidl: yes, I think we can make that guarantee if we use the mutex, subject to `count` being less than `STREAM_POOL_SIZE`.

cpp/src/io/parquet/stream_pool.cpp
etseidl: > hear me out

I'm all ears...tell me more :)

cpp/src/io/parquet/stream_pool.cpp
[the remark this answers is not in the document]
vuule: > would no longer be able to guarantee the same stream for a given id.
Why not, does the index wrap around?

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+// TODO: these next 2 (3?) can go away if we expose global_cuda_stream_pool()
+rmm::cuda_stream_view get_stream() { return global_cuda_stream_pool().get_stream(); }
+
+rmm::cuda_stream_view get_stream(std::size_t stream_id)
+{
+  return global_cuda_stream_pool().get_stream(stream_id);
+}
+
+std::vector<rmm::cuda_stream_view> get_streams(uint32_t count)
```
vuule, line 89: I think `get_streams` need to be in the class interface. For example, the warning does not make sense for the debug pool.

cpp/src/io/parquet/stream_pool.cpp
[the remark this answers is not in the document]
etseidl: Yes, they mod the stream id with the pool size. So if the size of the pool changes...

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,127 @@
...
+  cudaEvent_t event;
+  CUDF_CUDA_TRY(cudaEventCreate(&event));
+  CUDF_CUDA_TRY(cudaEventRecord(event, stream));
+  std::for_each(streams.begin(), streams.end(), [&](auto& strm) {
+    CUDF_CUDA_TRY(cudaStreamWaitEvent(strm, event, 0));
+  });
+  CUDF_CUDA_TRY(cudaEventDestroy(event));
+}
+
+void join_streams(std::vector<rmm::cuda_stream_view>& streams, rmm::cuda_stream_view stream)
```
vuule, line 99:
```suggestion
void join_streams(host_span<rmm::cuda_stream_view> streams, rmm::cuda_stream_view stream)
```

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,127 @@
...
+  return global_cuda_stream_pool().get_stream(stream_id);
+}
+
+std::vector<rmm::cuda_stream_view> get_streams(uint32_t count)
+{
+  return global_cuda_stream_pool().get_streams(count);
+}
+#endif
+
+void fork_streams(std::vector<rmm::cuda_stream_view>& streams, rmm::cuda_stream_view stream)
```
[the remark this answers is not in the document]
etseidl: to operate on a subset of stream?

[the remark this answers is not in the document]
vuule: That, and host_span can be created from any contiguous buffer.
Not too important, but there are basically no drawbacks, it just makes the API a bit more applicable.

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,119 @@
...
+cuda_stream_pool& global_cuda_stream_pool()
+{
+  static cuda_stream_pool* pool = create_global_cuda_stream_pool();
+  return *pool;
+}
+
+void fork_streams(host_span<rmm::cuda_stream_view> streams, rmm::cuda_stream_view stream)
+{
+  cudaEvent_t event;
+  CUDF_CUDA_TRY(cudaEventCreate(&event));
```
jrhemstad, line 100: I'd avoid the overhead of creating/destroying the event every time by making the event a function local static.

Also, you should disable timing on the event to reduce the overhead.

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    if (count > STREAM_POOL_SIZE) {
+      CUDF_LOG_WARN("get_streams called with count ({}) > pool size ({})", count, STREAM_POOL_SIZE);
+    }
+    auto streams = std::vector<rmm::cuda_stream_view>();
+    std::lock_guard<std::mutex> lock(stream_pool_mutex);
+    for (uint32_t i = 0; i < count; i++) {
+      streams.emplace_back(_pool.get_stream());
+    }
+    return streams;
```
jrhemstad, line 60: What's with the mutex/locks here? `rmm::cuda_stream_pool::get_stream` is thread safe, so I don't believe these are needed.

cpp/src/io/parquet/stream_pool.cpp
etseidl: sounds good. thanks!

cpp/src/io/parquet/stream_pool.cpp
etseidl: > What's with the mutex/locks here?

The idea is to make sure a given thread gets a set of unique streams (assuming the count is below the stream pool size). My fear is several threads calling this a the same time.  Even though the underlying `_pool.get_stream()` is thread safe, I can imagine enough threads calling concurrently that the stream counter rolls over and one thread winds up with one or more duplicate streams in it's vector, leading to kernels that could overlap operation serializing instead. Low probability, but would be a head scratcher if it actually occurred.

bdice:
We ought to benchmark and profile how `cudf::gather` on a table would change with a stream pool, with a stream for each column. https://github.com/rapidsai/cudf/issues/13509

cpp/src/io/parquet/stream_pool.cpp
etseidl: Added, but then got worried about different threads recording events on the same object.  @vuule whipped up an event pool that still gets rid of the event creation overhead, but returns a different event object per thread.

cpp/src/io/parquet/stream_pool.cpp
```diff
@@ -0,0 +1,148 @@
...
+ private:
+  std::unordered_map<std::thread::id, cudaEvent_t> event_map_;
+  std::mutex map_mutex_;
+};
+
+cudaEvent_t event_for_thread()
+{
+  static cuda_event_map instance;
+  return instance.find(std::this_thread::get_id());
+}
```
jrhemstad, line 120: No need for the `cuda_event_map`. `thread_local` will handle creating an object per thread for you. 

```suggestion
struct cuda_event {
  cuda_event()
      : e_{[]() {
          cudaEvent_t event;
          CUDF_CUDA_TRY(
              cudaEventCreateWithFlags(&event, cudaEventDisableTiming));
          return event;
        }()} {}

 private:
  struct deleter {
    using pointer = cudaEvent_t;
    auto operator()(cudaEvent_t e) { cudaEventDestroy(e); }
  };
  std::unique_ptr<cudaEvent_t, deleter> e_;
};


cudaEvent_t event_for_thread()
{
  thread_local cuda_event  thread_event;
  return thread_event;
}
```

cpp/src/io/parquet/stream_pool.cpp
jrhemstad: [remark] Hm, okay, I see what you're going for.  Just know that threads can still get overlapping streams even with this logic.

Furthermore, streams themselves are just an abstraction over the hardware channels, so two independent streams can still be artificially serialized.

bdice (commented):
This generally looks like it's going in the right direction. I'll try and run some benchmarks with table gathers using this stream pool (probably later this week).

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  std::vector<rmm::cuda_stream_view> get_streams(uint32_t count) override
+  {
+    static std::mutex stream_pool_mutex;
+
+    if (count > STREAM_POOL_SIZE) {
+      CUDF_LOG_WARN("get_streams called with count ({}) > pool size ({})", count, STREAM_POOL_SIZE);
+    }
+    auto streams = std::vector<rmm::cuda_stream_view>();
+    std::lock_guard<std::mutex> lock(stream_pool_mutex);
```
vuule, line 55: I think I see a bit of a problem. This lock only protects against other `get_streams` calls, but concurrent `get_stream` calls can still "steal" a stream in the middle of the expected order.
I'm starting to think that we should drop the mutex for now. If this turns out to be important, maybe a case can be made to have the feature in the rmm pool.

bdice: I still think that we only need a simple wrapper around the RMM stream pool that allows for debugging with 1 stream. I don't really understand why the locks are important. RMM already has an atomically incrementing counter for each call to `get_stream()` that ensures work is cycled across the stream pool regardless of the calling thread. https://github.com/rapidsai/rmm/blob/228330543cba4921e6b71a52dd1fa81c63921420/include/rmm/cuda_stream_pool.hpp#L96

cpp/include/cudf/detail/utilities/stream_pool.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * `LIBCUDF_USE_DEBUG_STREAM_POOL`.
+ *
+ * Access to the global `cuda_stream_pool` is granted via `cudf::detail::global_cuda_stream_pool()`.
+ *
+ * Example usage:
+ * @code{.cpp}
+ * auto stream = cudf::get_default_stream();
+ * auto const num_streams = 2;
+ * // do work on stream
+ * auto streams = cudf::detail::global_cuda_stream_pool().get_streams(num_streams);
```
bdice, line 42: Let's consider a simpler API. What pros/cons do you see here?

```c++
auto stream = cudf::get_default_stream();
auto const num_streams = 2;
auto streams = cudf::detail::fork_streams(stream, num_streams);
// do work on streams[0], streams[1]
cudf::detail::join_streams(streams);
```

Here, the `fork_streams` call would return a vector of streams taken from the global pool -- no need to refer to the global pool directly at all.

bdice: I also considered (and rejected) making `fork_streams` return an RAII object that automatically joins. That might be too much for readers to expect implicit joining -- and maybe not stream-safe by design, if it encourages bad habits where host objects go out of scope before the join occurs... that's already a case where we expect to require a manual stream sync, but it'd require an explicit join before that sync happens as well.

etseidl: I'm not opposed.  @vuule?

vuule: Looks neat. Would this be an overload of fork_streams or a replacement? I wonder if we ever need the original, to fork streams that don't come directly from the pool.

etseidl: And should we keep `get_streams()` in the stream pool interface, or just roll that functionality into `fork_streams()`?

bdice: I vote to replace the existing fork_streams. And to keep the get_streams pool interface, just internally for use by fork_streams.

I tried to think of cases where you’d want a pool without forking and still want to sync with another stream, but I didn’t see anything immediately.

etseidl: @bdice I've pushed a new minimalist interface...is this what you were after, or is this too far?

cpp/include/cudf/detail/utilities/stream_pool.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * // do work on streams[0] and streams[1]
+ * // wait for event on streams before continuing to do work on stream
+ * cudf::detail::join_streams(streams, stream);
+ * @endcode
+ *
+ * @param stream Stream to synchronize the returned streams to, usually the default stream.
+ * @param count The number of stream views to return.
+ * @return Vector containing `count` stream views.
+ */
+std::vector<rmm::cuda_stream_view> fork_streams(rmm::cuda_stream_view stream, uint32_t count);
```
vuule, line 50: should this be `fork_stream`?

etseidl: I guess that would be consistent with the input

cpp/include/cudf/detail/utilities/stream_pool.hpp
vuule: I think the new interface is good; pretty hard to misuse.
I've thought about whether we'd want to keep get and fork separate, but could not find a use case that can't use the current API. So +1 for the simple API from me :)

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+ private:
+  struct deleter {
+    using pointer = cudaEvent_t;
+    auto operator()(cudaEvent_t e) { cudaEventDestroy(e); }
+  };
+  std::unique_ptr<cudaEvent_t, deleter> e_;
+};
+
+cudaEvent_t event_for_thread()
```
vuule, line 157: this could use some light docs
otherwise the whole thing looks :fire:

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,205 @@
...
+   * An attempt is made to ensure that the returned vector does not contain duplicate
+   * streams, but this cannot be guaranteed if `count` is greater than the value returned by
+   * `get_stream_pool_size()`.
+   *
+   * This function is thread safe with respect to other calls to the same function.
+   *
+   * @param count The number of stream views to return.
+   * @return Vector containing `count` stream views.
+   */
+  virtual std::vector<rmm::cuda_stream_view> get_streams(uint32_t count) = 0;
```
harrism, line 72: **💡 suggestion:** ‏ Use a signed integer here. https://google.github.io/styleguide/cppguide.html#Integer_Types

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * @brief Get a `cuda_stream_view` of the stream associated with `stream_id`.
+   *
+   * Equivalent values of `stream_id` return a `cuda_stream_view` to the same underlying stream.
+   * This function is thread safe with respect to other calls to the same function.
+   *
+   * @param stream_id Unique identifier for the desired stream
+   * @return Requested stream view.
+   */
+  virtual rmm::cuda_stream_view get_stream(std::size_t stream_id) = 0;
```
harrism, line 58: **💡 suggestion:** ‏ Define a type for stream IDs with a `using` alias.

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,205 @@
...
+   *
+   * @return the number of stream objects in the pool
+   */
+  virtual std::size_t get_stream_pool_size() const = 0;
+};
+
+/**
+ * @brief Implementation of `cuda_stream_pool` that wraps an `rmm::cuda_stram_pool`.
+ */
+class rmm_cuda_stream_pool : public cuda_stream_pool {
```
harrism, line 87: **💡 suggestion:** ‏ I see that you want to have multiple stream pool implementations of a common interface. But this `rmm_cuda_stream_pool` mostly just calls into `rmm::cuda_stream_pool`, and extends it with `get_streams()`. So why not instead extend RMM by adding a base class, adding `get_streams()` to `rmm::cuda_stream_pool`, and adding a second derived class `constant_stream_pool` that always returns the same stream (which you initialize with `cudf::default_stream`).

The reason I suggest this is that it makes the same useful functionality available to all users of RMM, rather than just to cuDF.  Adding the fork/join functions to RMM would be useful too, I think.

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,205 @@
...
+ */
+cuda_stream_pool* create_global_cuda_stream_pool()
+{
+  if (getenv("LIBCUDF_USE_DEBUG_STREAM_POOL")) return new debug_cuda_stream_pool();
+
+  return new rmm_cuda_stream_pool();
+}
+
+/**
+ * @brief RAII struct to wrap a cuda event and ensure it's proper destruction.
```
harrism, line 143:
```suggestion
 * @brief RAII struct to wrap a cuda event and ensure its proper destruction.
```

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,205 @@
...
+      }()}
+  {
+  }
+
+  operator cudaEvent_t() { return e_.get(); }
+
+ private:
+  struct deleter {
+    using pointer = cudaEvent_t;
+    auto operator()(cudaEvent_t e) { cudaEventDestroy(e); }
```
harrism, line 160: **💡 suggestion:** ‏ Should assert success here, so that you catch errors in a debug build. Can't throw because this is likely to be called in a destructor.

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,205 @@
...
+  };
+  std::unique_ptr<cudaEvent_t, deleter> e_;
+};
+
+/**
+ * @brief Returns a cudaEvent_t for the current thread.
+ */
+cudaEvent_t event_for_thread()
+{
+  thread_local cuda_event thread_event;
```
harrism, line 170: **🚩 issue:** ‏ There's a tricky issue here. On multi-device systems, you will only have one event for all devices that this thread may use. Events are associated with a single device (the current device when `cudaEventCreate` is called. If this thread switches devices after `event_for_thread()` is called, and then it tries to `cudaEventRecord`, there will be a `cudaErrorInvalidResourceHandle`.  I'm currently working on an RMM PR to fix a similar issue, which I fixed by making `thread_local std::vector<cuda_event>` sized to the number of devices, instead of a single event. See https://github.com/rapidsai/rmm/pull/1333/files#diff-0d9e7691f84fe22b13d13977fe225d29f7824feee8e5b5763d98d8569b87634eR295

harrism (changes requested):
Thanks for this work, it is a nice addition. However, there is one critical issue for multi-device systems. Also, I would request that this be split into two PRs:

1. Extend the stream pool functionality in RMM to make the useful additions in this PR available to all RMM users.
2. Use the extended RMM pools in cuDF.

etseidl:
@harrism thanks for the great feedback! It might take me a while to digest it all. 😅

cpp/src/utilities/stream_pool.cpp
bdice: This is a good plan, I like the proposed design change. Thank you for these ideas @harrism!

cpp/src/utilities/stream_pool.cpp
etseidl: I changed it `std::size_t`. That's still unsigned, but more consistent with `std::vector` sizing.

cpp/src/utilities/stream_pool.cpp
etseidl: @harrism I'm trying to implement something similar to your solution, but I'm worried about races between threads. Would it be appropriate to do something like:
```c++
std::vector<cuda_event> create_events_per_device()
{
  std::vector<cuda_event> result(get_num_cuda_devices());
  for (int i=0; i < result.size(); i++) {
    set_current_cuda_device(rmm::cuda_device_id{i});
    result[i] = cuda_event{};
  }
  return result;
}

cudaEvent_t event_for_thread()
{
  thread_local std::vector<cuda_event> thread_events = create_events_per_device();
  return thread_events[get_current_cuda_device().value()];
}
```

etseidl: I took a stab at duplicating what you are doing in `stream_ordered_memory_resource`, but now I'm wondering if we also need a stream pool per device?

harrism: You definitely need either a stream pool per device or a stream pool that is multidevice aware. But the former can be done at the CUDF level outside of your stream pool class.  Even with that done, `thread_local` implies `static` so even if you create a pool per device your events need to be associated with a specific device.

Devices, streams and threads, Oh My!

cpp/include/cudf/detail/utilities/stream_pool.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * // do work on streams[0] and streams[1]
+ * // wait for event on streams before continuing to do work on stream
+ * cudf::detail::join_streams(streams, stream);
+ * @endcode
+ *
+ * @param stream Stream to synchronize the returned streams to, usually the default stream.
+ * @param count The number of stream views to return.
+ * @return Vector containing `count` stream views.
+ */
+std::vector<rmm::cuda_stream_view> fork_stream(rmm::cuda_stream_view stream, uint32_t count);
```
harrism, line 50: **💡 suggestion:** ‏ Multiple streams are. Also use consistent integer types.

```suggestion
std::vector<rmm::cuda_stream_view> fork_streams(rmm::cuda_stream_view stream, std::size_t count);
```

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,236 @@
...
+  do {                                                                          \
+    cudaError_t const status__ = (_call);                                       \
+    if (status__ != cudaSuccess) {                                              \
+      std::cerr << "CUDA Error detected. " << cudaGetErrorName(status__) << " " \
+                << cudaGetErrorString(status__) << std::endl;                   \
+    }                                                                           \
+    /* NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay) */   \
+    assert(status__ == cudaSuccess);                                            \
+  } while (0)
+#endif
```
harrism, line 53: **💡 suggestion:** ‏ Should be in cuDF's error header?

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+class cuda_stream_pool {
+ public:
+  // matching type used in rmm::cuda_stream_pool::get_stream(stream_id)
+  using stream_id_t = std::size_t;
```
harrism, line 58: **🤓 nitpick (non-blocking):** ‏ I think either `stream_id` or `stream_id_type` would be more idiomatic modern C++.

See: 
* https://stackoverflow.com/a/26400652
* https://www.reddit.com/r/cpp/comments/61yaf5/posixs_t_rule_and_modern_c_types/?rdt=33192

cpp/include/cudf/detail/utilities/stream_pool.hpp
harrism: Why singular? It forks multiple streams from a single input. In the context of unix "fork", this is like calling `fork` multiple times. 

Plus, plural is consistent with `join_streams`

cpp/src/utilities/stream_pool.cpp
```diff
@@ -161,36 +161,45 @@ cuda_stream_pool* create_global_cuda_stream_pool()
...
 struct cuda_event {
-  cuda_event()
-    : e_{[]() {
-        cudaEvent_t event;
-        CUDF_CUDA_TRY(cudaEventCreateWithFlags(&event, cudaEventDisableTiming));
-        return event;
-      }()}
-  {
-  }
+  cuda_event() { CUDF_CUDA_TRY(cudaEventCreateWithFlags(&e_, cudaEventDisableTiming)); }
```
harrism, line 183: **👏 praise:** ‏ I like this better than the previous approach with unique_ptr. Nice and simple.

cpp/src/utilities/stream_pool.cpp
```diff
@@ -161,36 +161,45 @@ cuda_stream_pool* create_global_cuda_stream_pool()
...
-  struct deleter {
-    using pointer = cudaEvent_t;
-    auto operator()(cudaEvent_t e) { CUDF_ASSERT_CUDA_SUCCESS(cudaEventDestroy(e)); }
-  };
-  std::unique_ptr<cudaEvent_t, deleter> e_;
+  cudaEvent_t e_;
 };
 
 /**
  * @brief Returns a cudaEvent_t for the current thread.
```
harrism, line 193: **💡 suggestion:** ‏ I think This will fail the new doxygen tests because the return value is not documented. There is a new `briefreturn` doxygen macro created by @vyasr that you can use to keep the docs a one-liner.

harrism: Nevermind, the doxygen tests apply to RMM only so far.

cpp/src/utilities/stream_pool.cpp
```diff
@@ -161,36 +161,45 @@ cuda_stream_pool* create_global_cuda_stream_pool()
...
  */
 cudaEvent_t event_for_thread()
 {
-  thread_local cuda_event thread_event;
-  return thread_event;
+  thread_local std::vector<std::unique_ptr<cuda_event>> thread_events(get_num_cuda_devices());
```
harrism, line 197: **❓ question:** ‏ Curious to discuss `unique` vs. `shared` here. I used `shared` in my concurrent RMM PR, but I don't feel like an expert about C++ smart pointers.

I *think* `unique_ptr` is right here because the ptr itself is never copied. In my code I used `shared_ptr` because I was thinking "multiple threads, multiple copies", but really there's a separate ptr (and event) per thread in `thread_local`, not copies of the same pointer. So I could change my code to `unique_ptr`. Am I right?

harrism (changes requested):
Looking good! I discussed with @vuule offline that in the interest of using it sooner in cuDF, I would accept this PR as is (modulo my new comments) in cuDF, with a follow-up PR to add it to RMM, and then a third PR to update cuDF to use it from RMM.

@bdice liked my design suggestion. Curious if you are OK with a multi-step migration to RMM, @bdice?

cpp/src/utilities/stream_pool.cpp
bdice: I vote `stream_id_type`.

cpp/include/cudf/detail/utilities/stream_pool.hpp
etseidl: Forking from the singular stream that is passed in as input? Joining multiple streams back to the singular stream. :shrug: I'm bad at naming so I'm happy either way.

cpp/src/utilities/stream_pool.cpp
bdice: I think that sounds right.

bdice:
> @bdice liked my design suggestion. Curious if you are OK with a multi-step migration to RMM, @bdice?

Yes, that would be fine with me.

cpp/src/utilities/stream_pool.cpp
etseidl: My thinking was it's temporary anyway, but if others feel there's a desire to keep this I'll move it.

cpp/include/cudf/detail/utilities/stream_pool.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @brief Synchronize a stream to an event on a set of streams.
+ *
+ * @param streams Vector of streams to synchronize to.
+ * @param stream Stream to synchronize, usually the default stream.
```
bdice, line 56: Does this help clarify the docs? (Not sure if this is better or worse.)
```suggestion
 * @param streams Streams to wait on.
 * @param stream Joined stream that synchronizes with the waited-on streams.
```

cpp/src/utilities/stream_pool.cpp
etseidl: Stole it straight from `event_wrapper` in `stream_ordered_memory_resource` :wink:

cpp/src/utilities/stream_pool.cpp
etseidl: IIUC, you return a {cudaStream_t, cudaEvent_t} tuple, not the smart pointer itself. So yes, I _think_ you could just use `unique_ptr` as well.  @vuule or @bdice , do you care to comment?

cpp/include/cudf/detail/utilities/stream_pool.hpp
vuule: Interesting. My interpretation was that the single stream is forked into multiple streams. Hence singular.
I'd be okay with `fork_streams` as well.

bdice: At first glance, I see this the same way as @etseidl and @vuule. It seems like you fork one stream to make multiple streams, and join multiple streams to make one stream. Hence `fork_stream` is singular, `join_streams` is plural based on their inputs. I'm not sure that the parallel to UNIX `fork` is strong enough to interpret it as requiring a plural. But I also see that it sounds nice to have both be plural.

cpp/src/utilities/stream_pool.cpp
bdice: Is this materially different from `CUDF_CUDA_TRY`? There's an `assert` vs. a `throw`, but I think we _want_ the `CUDF_CUDA_TRY` behavior if I'm not mistaken. https://github.com/rapidsai/cudf/blob/e81d79e94f268499c8656eba1fe8de8122589780/cpp/include/cudf/utilities/error.hpp#L256

cpp/src/utilities/stream_pool.cpp
vuule: Yeah, no need for `shared_ptr`, since the pointer itself is never copied. The thread local vector is the sole owner of the events.

cpp/src/utilities/stream_pool.cpp
bdice: Ah, is the difference that we need the event destructor calling this to be `noexcept` and thus can't throw? If so, can we mark the destructor `noexcept` explicitly?

etseidl: I think the biggest difference is that the ASSERT version only operates when compiled for debug.  Otherwise it's a no-op.

vuule: > can we mark the destructor noexcept explicitly?

I'm mildly against this since it's redundant.

bdice: You’re right. It’s redundant and not necessary. My mistake.

bdice: Now that I’ve been corrected (thanks), I don’t have strong feelings here. Resolve as you see fit.

cpp/include/cudf/detail/utilities/stream_pool.hpp
harrism: **💡 suggestion:** ‏ The disagreement is really about whether we are naming the functions in reference to their input or their output. I have a suggestion that may resolve this difference of opinions: let the input and output types (and plurality) speak for themselves. Having `stream` or `streams` in the name is not necessary, since they are part of the type signature. So just name the functions like so:

```cpp
std::vector<rmm::cuda_stream_view> fork(rmm::cuda_stream_view stream, uint32_t count);
void join(host_span<rmm::cuda_stream_view> streams, rmm::cuda_stream_view stream);
```


**💭 thought (non-blocking):** ‏ I can also envision another version of `join` that doesn't take the singular output stream as input, and instead just chooses one stream to synchronize all the others to (e.g. the first in the span):

```cpp
rmm::cuda_stream_view join(host_span<rmm::cuda_stream_view> streams);
```

cpp/src/utilities/stream_pool.cpp
etseidl: Ok, I'm going to leave it as is for now. When the stream pool moves to RMM the macro will go away.

cpp/include/cudf/detail/utilities/stream_pool.hpp
harrism: I like it. I don't really like "usually the default stream".

etseidl:
@harrism @bdice I _think_ I captured all your suggestions. Please take another look. :pray:

cpp/include/cudf/detail/utilities/stream_pool.hpp
etseidl: Ahhh, missed this comment. `<luke>`They're coming too fast!`</luke>`

I'm ok with the renaming. I'd prefer to pass on the other `join`s and address them in the follow-on PRs.

cpp/src/utilities/stream_pool.cpp
harrism: Actually, when I went to change to `unique_ptr` in my code here: https://github.com/rapidsai/rmm/pull/1333/files#diff-0d9e7691f84fe22b13d13977fe225d29f7824feee8e5b5763d98d8569b87634eR295-R304

I realized that I need `shared_ptr` because I actually stash a copy of the shared pointer in another container. The purpose of that copy is because the events may still be needed when a block is deallocated, *after* the thread exits and the TLS is freed.

Thinking about whether you need to do the same in this PR, I don't think so. However it made me wonder why you are using TLS at all. Is it just in case the stream passed to `fork` is the PTDS?  Could you just have a static event for each device? That would mean if you create multiple stream pools for a single device they would share an event, but hopefully nobody will do that...

etseidl: I'm sorry, I'm not thoroughly steeped in cuda terminology yet. PTDS I get, but what is TLS?

harrism: Yes, Thread Local Storage. That's C++, not CUDA.  Sorry for the unexplained TLA (Three Letter Acronym).

etseidl: Thanks, at least I guessed correctly 😅 I had to go back to look at some offline conversations with @vuule, but the reason for the TLS was a fear that if two threads called either fork or join at the same time, there could be a condition where thread 1 records an event on stream 1, then thread 2 records an event on stream 2, but both calls use the same static event. My understanding of the docs was that the second call to cudaEventRecord would overwrite the state on the event, so calls to cudaStreamWaitEvent from thread 1 would actually be waiting on stream 2. @vuule wrote a quick per-thread event store, but then @jrhemstad pointed out we could just use TLS to get an event per thread.

harrism: OK yes this is similar to the reason for the TLS in RMM's stream_ordered_memory_resource.  And I don't think you need the extra shared copy that we need there because the event is only used within the scope of one function. 👍

cpp/include/cudf/detail/utilities/stream_pool.hpp
etseidl: @bdice @vuule I think this is the last outstanding issue from this pass. Do you agree to the compromise naming suggestion?

bdice: I have reservations about the name `cudf::detail::join` because it sounds like a dataframe join and not something with streams. If we can keep “stream” somewhere in the name (or namespace), I don’t care if it’s singular or plural.

etseidl: So let it be written, so shall it be done 😀

bdice (approved):

cpp/src/utilities/stream_pool.cpp
```diff
@@ -0,0 +1,250 @@
...
+ * limitations under the License.
+ */
+
+#include <cudf/detail/utilities/logger.hpp>
+#include <cudf/detail/utilities/stream_pool.hpp>
+#include <cudf/utilities/default_stream.hpp>
+#include <cudf/utilities/error.hpp>
+
+#include <rmm/cuda_stream_pool.hpp>
+
```
PointKernel, line 23:
```suggestion

#include <algorithm>
#include <cstddef>
#include <memory>
#include <mutex>
#include <vector>

```

Add involved C++ headers for `std::for_each`, `std::size_t`, `std::unique_ptr`, `std::mutex` and `std::vector` to avoid transitive inclusions.

cpp/include/cudf/detail/utilities/stream_pool.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cudf/utilities/span.hpp>
+
+#include <rmm/cuda_stream_view.hpp>
+
```
PointKernel, line 22:
```suggestion

#include <cstddef>
#include <vector>

```

vuule (approved):
:fire:

harrism (changes requested):
Great work. Just one more missing `const`, I think.

cpp/include/cudf/detail/utilities/stream_pool.hpp
```diff
@@ -0,0 +1,64 @@
...
+[[nodiscard]] std::vector<rmm::cuda_stream_view> fork_streams(rmm::cuda_stream_view stream,
+                                                              std::size_t count);
+
+/**
+ * @brief Synchronize a stream to an event on a set of streams.
+ *
+ * @param streams Streams to wait on.
+ * @param stream Joined stream that synchronizes with the waited-on streams.
+ */
+void join_streams(host_span<rmm::cuda_stream_view> streams, rmm::cuda_stream_view stream);
```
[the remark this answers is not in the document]
etseidl: done

harrism (approved):

Issue #13902, opened 2023-08-17
[BUG] LocIndexing a float index with integer slice does not respect end point


mroeschke:
**Describe the bug**
LocIndexing a float index with integer slice does not respect end point

**Steps/Code to reproduce bug**
```
In [53]: import cudf

In [54]: cudf.__version__
Out[54]: '23.10.00'

In [79]: ser = cudf.Series(range(3), index=[2.0, 3.0, 4.5])

In [80]: ser
Out[80]: 
2.0    0
3.0    1
4.5    2
dtype: int64

In [81]: ser.loc[2:4]
Out[81]: 
2.0    0
3.0    1
4.5    2
dtype: int64

In [82]: ser.to_pandas().loc[2:4]
Out[82]: 
2.0    0
3.0    1
dtype: int64
```

**Expected behavior**
Though there could be int/float precision issues, in this case 4.5 should be clearly excluded with a slice endpoint of 4

**Environment overview (please complete the following information)**
 - Environment location: Bare-metal
 - Method of cuDF install: conda
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**
Please run and paste the output of the `cudf/print_env.sh` script here, to gather any other relevant environment details

**Additional context**

PR #13925, opened 2023-08-19
Fix handling of typecasting in `searchsorted`

galipremsagar:
## Description

Fixes: #13902 

This PR fixes a type-casting issue with `searchsorted` where typecast was done to the `values` dtype instead of inspecting both input and values columns and converting them to common dtypes.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/frame.py
+++ b/python/cudf/cudf/core/frame.py
@@ -1409,12 +1409,23 @@ class Frame(BinaryOperand, Scannable):
         if len(values) != len(self._data):
             raise ValueError("Mismatch number of columns to search for.")
 
-        sources = [
-            col
-            if is_dtype_equal(col.dtype, val.dtype)
-            else col.astype(val.dtype)
+        common_dtype_list = [
+            find_common_type([col.dtype, val.dtype])
             for col, val in zip(self._columns, values)
         ]
+        sources = [
+            col
+            if is_dtype_equal(col.dtype, common_dtype)
+            else col.astype(common_dtype)
+            for col, common_dtype in zip(self._columns, common_dtype_list)
+        ]
+        values = [
+            val
+            if is_dtype_equal(val.dtype, common_dtype)
+            else val.astype(common_dtype)
+            for val, common_dtype in zip(values, common_dtype_list)
+        ]
+
         outcol = libcudf.search.search_sorted(
             sources,
             values,
--- a/python/cudf/cudf/core/frame.py
+++ b/python/cudf/cudf/core/frame.py
@@ -1409,6 +1409,8 @@ class Frame(BinaryOperand, Scannable):
         if len(values) != len(self._data):
             raise ValueError("Mismatch number of columns to search for.")
 
+        # TODO: Change behavior based on the decision in
+        # https://github.com/pandas-dev/pandas/issues/54668
         common_dtype_list = [
             find_common_type([col.dtype, val.dtype])
             for col, val in zip(self._columns, values)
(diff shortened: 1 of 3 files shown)


python/cudf/cudf/core/frame.py
```diff
@@ -1409,12 +1409,23 @@ def searchsorted(
...
-            for col, val in zip(self._columns, values)
+            if is_dtype_equal(col.dtype, common_dtype)
+            else col.astype(common_dtype)
+            for col, common_dtype in zip(self._columns, common_dtype_list)
         ]
+        values = [
+            val
+            if is_dtype_equal(val.dtype, common_dtype)
+            else val.astype(common_dtype)
+            for val, common_dtype in zip(values, common_dtype_list)
```
wence-, line 1426: This allows:
```
s = cudf.Series([1, 2, 3], index=[1, 2, 3])
s.loc["1":"3"]  # same as if we wrote s.loc[1:3]
```
to succeed, even though `s.loc["1"]` and `s.loc["3"]` raise `KeyError`. But I note that pandas does the same thing here, so perhaps this is safe (@mroeschke is this a bug?). If I make a series:
```
import pandas as pd
s = pd.Series([1, 2, "1"])
s.sort_values() # raise ValueError
# But
s = pd.Series([1, 2])
s.searchsorted("1") # => array([0]), whereas I was expecting ValueError
```

mroeschke: Interesting. I think we are mirroring this behavior from numpy but I think pandas should be raising a ValueError here too

```python
In [4]: import numpy as np

In [5]: arr = np.array([1, 2])

In [6]: arr.searchsorted("1")
Out[6]: 0
```

galipremsagar: @mroeschke Would it be possible for you to open a pandas issue to track this? I can link it in our code for future change/reference.

mroeschke: Sure, https://github.com/pandas-dev/pandas/issues/54668

galipremsagar: Thanks @mroeschke ! 

@wence- I added a TODO to update our code behavior in future based on how https://github.com/pandas-dev/pandas/issues/54668 progresses. Let me know if you think that's okay.

python/cudf/cudf/tests/test_indexing.py
```diff
@@ -2067,3 +2067,16 @@ def test_loc_index_notinindex_slice(
             expect = pdf.loc[lo:hi:take_order]
             actual = df.loc[lo:hi:take_order]
             assert_eq(expect, actual)
+
```
wence-, line 2070: If pandas upgrades are an intentional thing, then we can encode the pandas behaviour, and hence the TODO, in a test:

```
@pytest.mark.xfail(reason="https://github.com/pandas-dev/pandas/issues/54668")
def test_searchsorted_mixed_str_int():
    s = pd.Series([1, 2, 3], dtype="int")
    c = cudf.from_pandas(s)
    with pytest.raises(ValueError):
        actual = s.searchsorted("a")
    with pytest.raises(ValueError):
        expect = c.searchsorted("a")
    assert_eq(expect, actual)
```

This test will fail once that pandas issue is addressed, and remind us (when upgrading pandas) to change the cudf behaviour.

wence- (approved):
Minor suggestion for a test, but looks good.

python/cudf/cudf/tests/test_indexing.py
galipremsagar: Done 👍

Issue #13918, opened 2023-08-18
[BUG] DataFrame iloc with a scalar can return different outputs based on the combination of dtypes


beckernick:
`iloc[scalar]` can produce inconsistent outputs based on the dtypes of the inputs. Perhaps we should either throw an error or coerce, rather than sometimes return a DataFrame? 

With an int64 and object dtypes, we get a DataFrame while pandas gets a Series:

```python
import cudf
import pandas as pd

df = cudf.DataFrame({"a": ["a", "b"], "b":[0, 1]})
print(df.iloc[0])
print(df.to_pandas().iloc[0])
   a  b
0  a  0
a    a
b    0
Name: 0, dtype: object
```

With an int64 and a categorical, we throw an error (pandas first this time)

```python
a    a
b    0
Name: 0, dtype: object

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [3], in <cell line: 8>()
      5 df["a"] = df["a"].astype("category")
      7 print(df.to_pandas().iloc[0])
----> 8 print(df.iloc[0])

File ~/miniconda3/envs/cudf-2310/lib/python3.10/site-packages/cudf/core/dataframe.py:446, in _DataFrameIlocIndexer.__getitem__(self, arg)
    441 try:
    442     # Behaviour difference from pandas, which will merrily
    443     # turn any heterogeneous set of columns into a series if
    444     # you only ask for one row.
    445     new_name = result.index[0]
--> 446     result = Series._concat(
    447         [result[name] for name in column_names],
    448         index=result.keys(),
    449     )
    450     result.name = new_name
    451     return result

File ~/miniconda3/envs/cudf-2310/lib/python3.10/site-packages/nvtx/nvtx.py:101, in annotate.__call__.<locals>.inner(*args, **kwargs)
     98 @wraps(func)
     99 def inner(*args, **kwargs):
    100     libnvtx_push_range(self.attributes, self.domain.handle)
--> 101     result = func(*args, **kwargs)
    102     libnvtx_pop_range(self.domain.handle)
    103     return result

File ~/miniconda3/envs/cudf-2310/lib/python3.10/site-packages/cudf/core/series.py:1619, in Series._concat(cls, objs, axis, index)
   1616         common_dtype = find_common_type([obj.dtype for obj in objs])
   1617         objs = [obj.astype(common_dtype) for obj in objs]
-> 1619 col = concat_columns([o._column for o in objs])
   1621 if len(objs):
[...]

Issue #11477, opened 2022-08-05
[BUG] `DataFrame.iloc` returns the wrong type of object when a string column is present


brandon-b-miller:
**Describe the bug**
In the cases described in the title, we are getting a `DataFrame` rather than a `Series` object like we should.

**Steps/Code to reproduce bug**
```python
>>> df1 = cudf.DataFrame({
...     'a': [1, 2, 3],
...     'b': ['a', 'b', 'c']
... })
>>> df2 = cudf.DataFrame({
...     'a': [1, 2, 3],
...     'b': [4, 5, 6]
... })
>>> type(df1.iloc[0])
<class 'cudf.core.dataframe.DataFrame'>
>>> type(df2.iloc[0]) 
<class 'cudf.core.series.Series'>
>>> type(df1.to_pandas().iloc[0])
<class 'pandas.core.series.Series'>
>>> type(df2.to_pandas().iloc[0])
<class 'pandas.core.series.Series'>
```

**Expected behavior**
We should get the same object as pandas (a series)

**Additional Information**
Branch-22.10

wence-:
This is not possible because `Series` objects contain data of a single homogenous dtype. In this case (mixed string and numeric columns), the series that we want to get would contain heterogeneous dtypes, which is not supported. Pandas can happily store `object` arrays that really support heterogeneous dtypes so it is not a problem there.

shwina:
Should we {raise, warn, something_else}?

wence-:
I think we should never try and convert to a `Series`. That makes `.i?loc`-based indexing have consistent semantics in all cases, and try and convince pandas that that should also be the behaviour there. The isomorphism between a 1xN dataframe and length-N series in pandas is not really that strong anyway which is probably a cause of a bunch of special-casing.

e.g.
```python
import pandas as pd
df = pd.DataFrame({"a": [1, 2, 3, 4],
                   "b": [1, 2, 1, 2],
                   "c": [1, 1, 2, 2],
                   "d": [1, 2, 3, 4]}, index=["e", "f", "g", "h"])

print(df.loc["e"]["a"] * df)    # single number is broadcast
#    a  b  c  d
# e  1  1  1  1
# f  2  2  1  2
# g  3  1  2  3
# h  4  2  2  4
print(df.loc[["e"]]["a"] * df)  # (1,) series is not aligned
#     a   b   c   d   e
# e NaN NaN NaN NaN NaN
# f NaN NaN NaN NaN NaN
[...]

PR #13926, opened 2023-08-19
Fix `DataFrame.iloc` to return consistent output types for scalars

galipremsagar:
## Description
Fixes: #13918 

This PR raises an error when there are mixed column types that will be constructed as a result of calling `.iloc[scalar]` instead of returning a `DataFrame`.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -447,21 +447,13 @@ class _DataFrameIlocIndexer(_DataFrameIndexer):
             return frame._slice(row_spec.key)
         elif isinstance(row_spec, indexing_utils.ScalarIndexer):
             result = frame._gather(row_spec.key, keep_index=True)
-            # Attempt to turn into series.
-            try:
-                # Behaviour difference from pandas, which will merrily
-                # turn any heterogeneous set of columns into a series if
-                # you only ask for one row.
-                new_name = result.index[0]
-                result = Series._concat(
-                    [result[name] for name in column_names],
-                    index=result.keys(),
-                )
-                result.name = new_name
-                return result
-            except TypeError:
-                # Couldn't find a common type, just return a 1xN dataframe.
-                return result
+            new_name = result.index[0]
+            result = Series._concat(
+                [result[name] for name in column_names],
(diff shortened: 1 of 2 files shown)


wence-:
This can close https://github.com/rapidsai/cudf/issues/11477 I think.

python/cudf/cudf/core/dataframe.py
```diff
@@ -437,21 +437,13 @@ def __getitem__(self, arg):
...
-            except TypeError:
-                # Couldn't find a common type, just return a 1xN dataframe.
-                return result
+            new_name = result.index[0]
+            result = Series._concat(
+                [result[name] for name in column_names],
+                index=result.keys(),
+            )
+            result.name = new_name
+            return result
```
wence-, line 446: Do you want to gate this behind pandas-compat? Note also that this is a breaking change in that some cudf-only workflows will now no-longer work.

Additionally, should the `TypeError` here report a useful error message (recommending to use a singleton index/slice to get a 1xN dataframe, for example)?

galipremsagar: > Do you want to gate this behind pandas-compat? Note also that this is a breaking change in that some cudf-only workflows will now no-longer work.

Do we have to support this since we are anyway not able to match pandas here? After going through https://github.com/rapidsai/cudf/issues/11477, I feel we should just raise an useful error message rather that gating this inconsistency with pandas-compat/

> Additionally, should the `TypeError` here report a useful error message (recommending to use a singleton index/slice to get a 1xN dataframe, for example)?

+1 to raising this useful error message.


cc: @bdice | @shwina any thoughts?

galipremsagar: cc: @beckernick | @bdice | @shwina any thoughts?

bdice: I’m not entirely sure what to say, but I think I would propose this:

- Make breaking changes (like raising errors) only for pandas compat mode
- If desired, raise deprecation warnings in “normal” cudf mode and then change behavior later

galipremsagar:
closing this PR

PR #13927, opened 2023-08-21
Rewrite `DataFrame.stack` to support multi level column names

### LINKED ISSUE #13739: [BUG] `stack()` on a DataFrame with multi-level column results in an unprintable object
Reproducer:

```python
df = cudf.DataFrame({('A', 'a'): [1, 2], ('A', 'b'): [3, 4]})
result = df.stack()  # No error
print(result)  # TypeError: unhashable type: 'numpy.ndarray'
```

Expected behaviour:

```python
In [21]: df = pd.DataFrame({('A', 'a'): [1, 2], ('A', 'b'): [3, 4]})

In [22]: df
Out[22]:
   A
   a  b
0  1  3
1  2  4

In [23]: df.stack()
Out[23]:
     A
0 a  1
  b  3
1 a  2
  b  4
```

isVoid:
## Description

This PR rewrites `DataFrame.stack()`. Adding support to stacking multiple levels in the dataframe.
User can now specify one or more levels from the column names to stack. Example:

```python
>>> multicol1 = pd.MultiIndex.from_tuples([('weight', 'kg'),
...                                        ('weight', 'pounds')])
>>> df_multi_level_cols1 = cudf.DataFrame([[1, 2], [2, 4]],
...                                     index=['cat', 'dog'],
...                                     columns=multicol1)
>>> df_multi_level_cols1.stack(0)
            kg  pounds
cat weight   1       2
dog weight   2       4
>>> df_multi_level_cols1.stack([0, 1])
cat  weight  kg        1
             pounds    2
dog  weight  kg        2
             pounds    4
dtype: int64
```

The implementation heavily uses pandas index methods on the column axis. This assumes
that the width of the cudf column is limited.

The combination of `len(level) > 1 and dropna=False` is currently unsupported. The corresponding
behavior in pandas is due to be deprecated in 3.0. See pandas-dev/pandas#53515.

closes #13739

## Checklist


wence- (commented):
Thanks. This is some pretty tricksy code, I wonder if there is a way to make the algorithmic pattern at all more clear, or if it is by necessity messy in this way.

python/cudf/cudf/tests/test_reshape.py
```diff
@@ -151,6 +151,69 @@ def test_df_stack_reset_index():
...
+    ],
+)
+@pytest.mark.parametrize("dropna", [True, False])
+def test_df_stack_multiindex_column_axis(columns, level, dropna):
+    if isinstance(level, list) and len(level) > 1 and not dropna:
+        pytest.skip(
+            "Stacking multiple levels with dropna==False is unsupported."
+        )
+
+    pdf = pd.DataFrame(data=[[1, 2, 3, 4], [2, 4, 6, 8]], columns=columns)
```
mroeschke, line 188: Could you make `pdf` have an `Index` with a `name=...` (same as below)? I noticed recently the resulting Index names were not always preserved and I think this fixes it

python/cudf/cudf/core/dataframe.py
```diff
@@ -6380,71 +6380,294 @@ def to_orc(
...
+        dog  height  m     4.0
+             weight  kg    3.0
+        dtype: float64
         """
-        assert level in (None, -1)
-        repeated_index = self.index.repeat(self.shape[1])
-        name_index = libcudf.reshape.tile(
-            [as_column(self._column_names)], self.shape[0]
+
+        level = [level] if not isinstance(level, list) else level
```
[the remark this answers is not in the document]
isVoid: I like this approach. One thing I'd like to add is that we can use internal type hinting to make code more readable.

python/cudf/cudf/core/dataframe.py
```diff
@@ -6380,71 +6380,294 @@ def to_orc(
...
+                "in original dataframe. This behavior is unsupported in "
+                "cuDF. See pandas deprecation note: "
+                "https://github.com/pandas-dev/pandas/issues/53515"
+            )
+
+        # Compute the columns to stack based on specified levels
+        if any(not isinstance(lv, int) for lv in level):
+            # Convert level names to indices, assume a list of labels
+            names = pd.Index(self._data.level_names)
+            level = [names.get_loc(lv) for lv in level]
```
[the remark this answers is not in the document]
isVoid: You are right here. I was under the shadow of the broad range of label types that pandas allow in the column. If we limit the column names to (int, str) as above, we can avoid calling the pandas helper.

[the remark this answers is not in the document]
isVoid: There's also the slight concern to that `.index` is O(N), but considering we are promised to work with a short list here the performance should still outperform dispatching to pandas.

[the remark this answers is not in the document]
isVoid: > This doesn't work if the level list is a mix of int positions and str positions.

Pandas doesn't allow `level` to be specified with mixed integer position and str names as well:
```python

In [24]: midx = pd.MultiIndex.from_arrays([[1, 2, 3], [10, 11, 12]], names=[128, "a"])

In [25]: df = pd.DataFrame(columns=midx)

In [28]: df.stack(level=[0, "a"])

ValueError: level should contain all level names or all level numbers, not a mixture of the two.
```

However the following is allowed in pandas (which doesn't in this PR):

```python
In [17]: midx = pd.MultiIndex.from_arrays([[1, 2, 3], [10, 11, 12]], names=[128, 129])

In [18]: df = pd.DataFrame(columns=midx)

In [20]: df.stack(level=[128])
Out[20]: 
Empty DataFrame
Columns: [10, 11, 12]
Index: []
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -6380,71 +6380,294 @@ def to_orc(
...
+        repeated_index = self.index.repeat(len(unique_named_levels))
+
+        # Each column name should repeat itself by len(df) times
+        tiled_index = libcudf.reshape.tile(
+            [
+                as_column(unique_named_levels.get_level_values(i))
+                for i in range(unique_named_levels.nlevels)
+            ],
+            self.shape[0],
+        )
```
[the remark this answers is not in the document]
isVoid: Unfortunately while `unique_named_levels` is short, `self.shape[0]` can be large. Microbenchmark:

```python
repeat = 10

arr = pd.Index([1, 2, 3])

for e in [4, 5, 6, 7]:
    n = 10**e
    
    t1 = 0
    for _ in range(repeat):
        start = time.time()
        as_column(np.tile(arr, n))
        t1 += time.time() - start
    t1 /= repeat

    t2 = 0
    for _ in range(repeat):
        start = time.time()
        libcudf.reshape.tile([as_column(arr)], 1000000)
        t2 += time.time() - start
    t2 /= repeat

    print(n, f"np: {t1}", f"cu: {t2}", f"ratio: {t1 / t2}")
```
```
10000 np: 0.0011644840240478515 cu: 0.0014286041259765625 ratio: 0.8151201602136181
100000 np: 0.0015172481536865235 cu: 0.001217818260192871 ratio: 1.2458740382544686
1000000 np: 0.00784306526184082 cu: 0.0010487556457519532 ratio: 7.478448667818495
10000000 np: 0.14060163497924805 cu: 0.0011314630508422851 ratio: 124.26533493478307
```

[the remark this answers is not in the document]
isVoid: Cutting point is when self.shape[0] > 100000, the cost of "tile on H -> H2D" is greater than "H2D -> tile on D".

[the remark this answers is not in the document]
isVoid: The microbenchmark shows that "tile on H -> d2h" cost is always higher than "d2h -> tile on D".

python/cudf/cudf/core/dataframe.py
```diff
@@ -6380,71 +6380,294 @@ def to_orc(
...
+            homogenized: list(column) = []
+            for col in columns:
+                if col is None:
+                    if all_nans is None:
+                        all_nans = column_empty(
+                            self.shape[0], common_type, masked=True
+                        )
+                    col = all_nans
+                elif not np.issubdtype(col.dtype, common_type):
+                    col = col.as_type(common_type)
```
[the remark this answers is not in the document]
isVoid: Good catch. I simply overly trusted the old code here and didn't figured this is simply incorrect. `np.result_type` doesn't return a dtype family so this line will always evaluate to false. While `interleave_columns` explicitly requires all input columns to have the same type.

python/cudf/cudf/tests/test_reshape.py
isVoid: Done!

python/cudf/cudf/core/dataframe.py
```diff
@@ -6380,71 +6380,294 @@ def to_orc(
...
+            data=range(len(self._data)), index=named_levels
         )
 
-        result = Series._from_data(
-            {
-                None: libcudf.reshape.interleave_columns(
-                    [*homogenized._columns]
+        column_indices: list[list[int]] = []
+        if len(unnamed_levels_indices) == 0:
+            column_idx_df = column_idx_df.sort_index()
```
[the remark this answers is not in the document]
isVoid: Did best effort to factor them out here.

isVoid:
>  I wonder if there is a way to make the algorithmic pattern at all more clear, or if it is by necessity messy in this way.

On a high level, the method can break down to these several components:

1. Compute the new indices
2. Figuring out the correct column order to call `interleave_columns`
3. Call `interleave_columns`
4. Compute the new column names and assemble series/dataframe
5. dropna.

2, 3 might be able to be merged in a single step, but could create even more spaghetti code than it is now.

I'm not sure if it's appropriate to separate each step as a sub function in cuDF, also there are many shared variables between steps, so figured keeping them in a mega function with proper comment might be a good idea.

Thoughts welcome.

python/cudf/cudf/core/dataframe.py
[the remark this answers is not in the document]
wence-: Thanks!

python/cudf/cudf/core/dataframe.py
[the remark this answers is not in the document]
wence-: >  Since integers can be both a name or a position, pandas always preemptively check if the integer is a column name first, if it doesn't then assumes the integer is a position.

Ah yes, I hate this behaviour...

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,314 @@ def to_orc(
...
+                # assigns -1 to these key combinations, representing an
+                # all-null column that is used in the subsequent libcudf call.
+                grpdf_aligned = df.reindex(
+                    unique_named_levels, axis=0, fill_value=-1
+                )
+            grpdf_aligned.sort_index(inplace=True)  # this needed?
+            indices = []
+            for label in grpdf_aligned.index:
+                (column_idx,) = grpdf_aligned.loc[label]
+                indices.append(int(column_idx))
```
wence-, line 6684: Why is this not just:
```
grpdf_aligned.values
```
?

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,314 @@ def to_orc(
...
+            column_indices.append(indices)
+
+        # For each of the group constructed from the unnamed levels,
+        # invoke `interleave_columns` to stack the values.
+        stacked = []
+        all_nans = None  # lazily created when needed
```
wence-, line 6690: Here's a perhaps more functional way:
```
from functools import cache, partial

all_nans = cache(partial(column_empty, self.shape[0], common_type, masked=True))

...
if col is None:
    col = all_nans() # always the same object
else:
   ...
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,314 @@ def to_orc(
...
+
+        # Construct the resulting dataframe / series
+        if len(unnamed_levels_indices) == 0:
+            result = Series._from_data(
+                data={None: stacked[0]}, index=new_index
+            )
+        else:
+            unnamed_level_values = unnamed_level_values.unique().sort_values()
+            if unnamed_level_values.nlevels == 1:
+                unnamed_level_values = unnamed_level_values.get_level_values(0)
```
wence-, line 6730:
```suggestion
            if unnamed_level_values.nlevels == 1:
                unnamed_level_values = unnamed_level_values.get_level_values(0)
            unnamed_level_values = unnamed_level_values.unique().sort_values()
```

Let's not hit multiindices unless we have to :).

Maybe also:

```
assert len(stacked) == len(unnamed_level_values)
```
?

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,314 @@ def to_orc(
...
+            homogenized: list(column) = []
+            for col in columns:
+                if col is None:
+                    if all_nans is None:
+                        all_nans = column_empty(
+                            self.shape[0], common_type, masked=True
+                        )
+                    col = all_nans
+                elif col.dtype != common_type:
+                    col = col.astype(common_type)
```
wence-, line 6716: `astype` checks for equality and returns self in that case. So (with the `all_nans` suggestion above) we could write this as:
```
homogenized = [col.astype(common_type) if col is not None else all_nans() for col in columns]
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,314 @@ def to_orc(
...
+            # Collect columns based on indices, append None for -1 indices.
+            for i in column_idx:
+                if i == -1:
+                    columns.append(None)
+                else:
+                    columns.append(self._data.select_by_index(i).columns[0])
+
+            # Collect datatypes and cast columns as that type
+            common_type = np.result_type(
+                *[col.dtype for col in columns if col is not None]
```
wence-, line 6703:
```suggestion
                *(col.dtype for col in columns if col is not None)
```
No need to cons the list to immediately splat it.

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,314 @@ def to_orc(
...
+                (column_idx,) = grpdf_aligned.loc[label]
+                indices.append(int(column_idx))
+            column_indices.append(indices)
+
+        # For each of the group constructed from the unnamed levels,
+        # invoke `interleave_columns` to stack the values.
+        stacked = []
+        all_nans = None  # lazily created when needed
+        for column_idx in column_indices:
+            columns: list(column | None) = []
```
wence-, line 6692: I don't think you mean the `column` module. you probably want:
```suggestion
            columns: list[column.ColumnBase | None] = []
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,314 @@ def to_orc(
...
+
+            # Collect columns based on indices, append None for -1 indices.
+            for i in column_idx:
+                if i == -1:
+                    columns.append(None)
+                else:
+                    columns.append(self._data.select_by_index(i).columns[0])
```
wence-, line 6699: Use a list comprehension maybe?
```
columns = [None if i == -1 else self._data.select_by_index(i).columns[0]]
```

(This is kind of closest to the C++ mantra of "no raw loops" given that python doesn't have nice functional programming facilities).

python/cudf/cudf/core/dataframe.py
wence-: Although perhaps it is better named `all_nulls`?

wence- (commented):
Thanks, some small suggested cleanups in the body which I'll let you apply if you think it makes sense.

python/cudf/cudf/core/dataframe.py
isVoid: Nice, didn't thought that list comprehension is also cons.

isVoid:
@wence- Thanks! All are very useful comments, addressed them all.

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,304 @@ def to_orc(
...
+                unnamed_level_values
+            )
+
+        def unnamed_group_generator():
+            if len(unnamed_levels_indices) == 0:
+                yield column_idx_df
+            else:
+                for _, grpdf in column_idx_df.groupby(by=unnamed_level_values):
+                    # When stacking part of the levels, some combinations
+                    # of keysmay not present in this group but can present
```
wence-, line 6665:
```suggestion
                    # of keys may not be present in this group but can be present
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,304 @@ def to_orc(
...
+        else:
+            # Must be a list of positions, normalize negative positions
+            level_indices = [
+                lv + self._data.nlevels if lv < 0 else lv for lv in level
+            ]
+
+        unnamed_levels_indices = [
+            i for i in range(self._data.nlevels) if i not in level_indices
+        ]
+
```
wence-, line 6609:
```suggestion
        has_unnamed_levels = len(unnamed_levels_indices) > 0
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,304 @@ def to_orc(
...
+                        unique_named_levels, axis=0, fill_value=-1
+                    )
+                    grpdf_aligned.sort_index(inplace=True)
+                    yield grpdf_aligned
+
+        column_indices = [
+            grpdf_aligned.values for grpdf_aligned in unnamed_group_generator()
+        ]
+        # for grpdf_aligned in unnamed_group_generator():
+        #     column_indices.append(grpdf_aligned.values)
```
wence-, line 6680:
```suggestion
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,304 @@ def to_orc(
...
+            map(column_name_idx.get_level_values, unnamed_levels_indices)
         )
+        if unnamed_level_values:
+            unnamed_level_values = pd.MultiIndex.from_arrays(
+                unnamed_level_values
+            )
+
+        def unnamed_group_generator():
+            if len(unnamed_levels_indices) == 0:
+                yield column_idx_df
```
wence-, line 6661:
```suggestion
                yield column_idx_df.values
```

python/cudf/cudf/core/dataframe.py
```diff
@@ -6431,71 +6431,304 @@ def to_orc(
...
+                    # is used in the subsequent libcudf call.
+                    grpdf_aligned = grpdf.reindex(
+                        unique_named_levels, axis=0, fill_value=-1
+                    )
+                    grpdf_aligned.sort_index(inplace=True)
+                    yield grpdf_aligned
+
+        column_indices = [
+            grpdf_aligned.values for grpdf_aligned in unnamed_group_generator()
+        ]
```
wence-, line 6678:
```suggestion
        column_indices = list(unnamed_group_generator())
```

wence- (approved):
Thanks again, some minor cleanups to apply (the only substantive one is the `has_unnamed_levels` change)

mroeschke (approved):

wence-:
Thanks @isVoid!

Issue #13864, opened 2023-08-11
[BUG] LocIndexer of a DataFrame with a MultiIndex columns with a single label tuple returns DataFrame


mroeschke:
**Describe the bug**
LocIndexer of a DataFrame with a MultiIndex columns with a single label tuple returns DataFrame instead of a Series

**Steps/Code to reproduce bug**
```
In [50]: import cudf

In [51]: cudf.__version__
Out[51]: '23.10.00'

In [52]: dfmi = cudf.DataFrame([list('abcd'),
    ...: 
    ...:                      list('efgh'),
    ...: 
    ...:                      list('ijkl'),
    ...: 
    ...:                      list('mnop')],
    ...: 
    ...:                     columns=cudf.MultiIndex.from_product([['one', 'two'],
    ...: 
    ...:                                                         ['first', 'second']]))

In [53]: dfmi.loc[:, ('one', 'second')]
Out[53]: 
  (one, second)
0             b
1             f
2             j
3             n

In [54]: dfmi.to_pandas().loc[:, ('one', 'second')]
Out[54]: 
0    b
1    f
2    j
3    n
Name: (one, second), dtype: object
```

**Expected behavior**
Like pandas, a single tuple label should be treated like a scalar label instead of a container of labels and return a Series instead of a DataFrame

**Environment overview (please complete the following information)**
 - Environment location: Bare-metal
 - Method of cuDF install: conda
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**
Please run and paste the output of the `cudf/print_env.sh` script here, to gather any other relevant environment details

**Additional context**

PR #13929, opened 2023-08-21
Fix an issue with `loc` when column names is `MultiIndex`

galipremsagar:
## Description
Fixes: #13864 

This PR fixes an issue with `loc` indexer where some special handling needs to be done when `columns` is of type `MultiIndex`.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/column_accessor.py
+++ b/python/cudf/cudf/core/column_accessor.py
@@ -511,7 +511,7 @@ class ColumnAccessor(abc.MutableMapping):
     def _select_by_label_grouped(self, key: Any) -> ColumnAccessor:
         result = self._grouped_data[key]
         if isinstance(result, cudf.core.column.ColumnBase):
-            return self.__class__({key: result})
+            return self.__class__({key: result}, multiindex=self.multiindex)
         else:
             if self.multiindex:
                 result = _to_flat_dict(result)
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -173,11 +173,13 @@ class _DataFrameIndexer(_FrameIndexer):
             all_numeric = all(is_numeric_dtype(t) for t in dtypes)
             if all_numeric:
                 return True
+            if isinstance(arg[1], tuple):
+                return True
         if ncols == 1:
             if type(arg[1]) is slice:
                 return False
             if isinstance(arg[1], tuple):
-                return True
+                return len(arg[1]) == df._data.nlevels
             if not (is_list_like(arg[1]) or is_column_like(arg[1])):
                 return True
         return False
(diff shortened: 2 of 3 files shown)


python/cudf/cudf/core/dataframe.py
```diff
@@ -287,7 +288,13 @@ def _getitem_tuple_arg(self, arg):
                         BooleanMask(tmp_arg[0], len(columns_df))
                     )
                 else:
-                    tmp_col_name = str(uuid4())
+                    if columns_df._data.multiindex:
+                        tmp_col_name = (
+                            str(uuid4()),
+                            "" * columns_df._data.nlevels,
```
bdice, line 294: This doesn't seem to actually use the number of levels. `"" * n` is always `""` (repeating an empty string).

wence-: This should be:
```
tmp_col_name = (str(uuid4()), *("" for _ in range(columns_df._data.nlevels - 1)))
```
You're trying to produce a column index that has the same number of levels as `columns_df` to avoid this warning:

```
In [61]: other.join(df, how="inner")
/home/wence/Documents/src/rapids/cudf/python/cudf/cudf/core/join/join.py:438: FutureWarning: merging between different levels is deprecated and will be removed in a future version. (1 levels on the left, 2 on the right)
  warnings.warn(
```

python/cudf/cudf/tests/test_indexing.py
```diff
@@ -2067,3 +2067,24 @@ def test_loc_index_notinindex_slice(
...
+@pytest.mark.parametrize(
+    "arg",
+    [
+        (2, ("one", "second")),
+        (slice(None, None, None), ("two", "first")),
+        (1, ("one", "first")),
+        (slice(None, None, None), ("two", "second")),
+    ],
+)
+def test_loc_dataframe_column_multiindex(arg):
```
wence-, line 2081: Can you add a three-level multiindex (to catch the bug in the name length above).

python/cudf/cudf/core/column_accessor.py
```diff
@@ -511,7 +511,7 @@ def _select_by_label_list_like(self, key: Any) -> ColumnAccessor:
     def _select_by_label_grouped(self, key: Any) -> ColumnAccessor:
         result = self._grouped_data[key]
         if isinstance(result, cudf.core.column.ColumnBase):
-            return self.__class__({key: result})
+            return self.__class__({key: result}, multiindex=self.multiindex)
```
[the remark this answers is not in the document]
wence-: `ColumnAccessor` has used `self.__class__` since its inception AFAICT. The only time `type(x)` and `x.__class__` are different is if `__class__` is overridden, which isn't the case here I think.

I guess this is just mimicking existing style.

wence- (changes requested):
I think this makes sense, but some minor logic cleanup in the column name generation is needed.

python/cudf/cudf/tests/test_indexing.py
galipremsagar: Done

python/cudf/cudf/core/dataframe.py
```diff
@@ -296,6 +303,7 @@ def _getitem_tuple_arg(self, arg):
                         map(str, columns_df._data.names)
                     )
                     columns_df[cantor_name] = column.arange(len(columns_df))
+                    cantor_name = columns_df._data.names[-1]
```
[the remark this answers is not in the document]

python/cudf/cudf/core/dataframe.py
galipremsagar: Sorry for the incorrect logic, Thank you both for catching this issue.

python/cudf/cudf/core/column_accessor.py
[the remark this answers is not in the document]
galipremsagar: For consistency with the rest of the code in this file too, I choose to leave it as `self.__class__`.

wence- (approved):
Thanks @galipremsagar

PR #13932, opened 2023-08-22
Fix construction of `Grouping` objects

### LINKED ISSUE #13390: [BUG] groupby object is unusable after calling `.cumcount`
**Describe the bug**

```python
import cudf
df = cudf.DataFrame({"key": [0] * 4, "values": range(4)})

g = df.groupby("key")

expect = g.cumsum()
counts = g.cumcount()
got = g.cumsum()

print(expect)
#   values
# 0       0
# 1       1
# 2       3
# 3       6
print(got)
#   None
# 0     0
# 1     0
# 2     0
# 3     0
```

What went on? `cumcount` is implemented as:
```python
    def cumcount(self):
        """
        Return the cumulative count of keys in each group.
        """
        return (
            cudf.Series(
                cudf.core.column.column_empty(
                    len(self.obj), "int8", masked=False
                ),
                index=self.obj.index,
            )
            .groupby(self.grouping, sort=self._sort)
            .agg("cumcount")
        )
```

So we build an empty column, group it with the original grouper and call an aggregation, looks relatively benign, but... `self.grouping` carries around the grouped values and the `GroupBy` constructor does this:

```python
        if isinstance(by, _Grouping):
            by._obj = self.obj
            self.grouping = by
```

This is horrible action at a distance because we now have modified the original `Groupby.grouping` object that we first thought of to act on this nonsense empty column rather that the original dataframe.

**Expected behavior**

This should work.

### LINKED ISSUE #13923: [BUG] Groupby object returns erroneous results for subsequent nth calls
**Describe the bug**
Groupby object returns erroneous results for subsequent nth calls

```
In [63]: import cudf; import numpy as np

In [64]: df = cudf.DataFrame([[1, np.nan], [1, 4], [5, 6]], columns=["A", "B"])

In [65]: g = df.groupby("A")

In [66]: g.nth(-1)
Out[66]: 
   B
A   
1  4
5  6

In [67]: g.nth(0)  # same result for any n
Out[67]: 
   None
A      
1     0
5     0

In [68]: pdf = df.to_pandas()

In [69]: g = pdf.groupby("A")

In [70]: g.nth(-1)
Out[70]: 
     B
A     
1  4.0
5  6.0

In [71]: g.nth(0)
Out[71]: 
     B
A     
1  NaN
5  6.0
```

**Expected behavior**
Looks like there's some errant state held in the groupby object that makes subsequent groupby.nth calls incorrect. Should match the pandas result

   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**

**Additional context**

galipremsagar:
## Description
This PR fixes multiple issues with groupby where the consequent operations on a groupby object always return incorrect results due to an inplace modification that happens in the `Groupby` constructor. To fix this, take a copy of the `by` argument whenever it is an (internal) `_Grouping` object.

- Closes #13923
- Closes #13390
## Checklist

### DIFF
--- a/python/cudf/cudf/core/groupby/groupby.py
+++ b/python/cudf/cudf/core/groupby/groupby.py
@@ -1,5 +1,6 @@
 # Copyright (c) 2020-2023, NVIDIA CORPORATION.
 
+import copy
 import itertools
 import pickle
 import textwrap
@@ -261,17 +262,17 @@ class GroupBy(Serializable, Reducible, Scannable):
         """
         self.obj = obj
         self._as_index = as_index
-        self._by = by
+        self._by = by.copy(deep=True) if isinstance(by, _Grouping) else by
         self._level = level
         self._sort = sort
         self._dropna = dropna
         self._group_keys = group_keys
 
-        if isinstance(by, _Grouping):
-            by._obj = self.obj
-            self.grouping = by
+        if isinstance(self._by, _Grouping):
+            self._by._obj = self.obj
+            self.grouping = self._by
         else:
-            self.grouping = _Grouping(obj, by, level)
+            self.grouping = _Grouping(obj, self._by, level)
 
     def __iter__(self):
         if isinstance(self._by, list) and len(self._by) == 1:
@@ -2545,6 +2546,13 @@ class _Grouping(Serializable):
         out.names = names
         out._named_columns = _named_columns
         out._key_columns = key_columns
+        return out
+
+    def copy(self, deep=True):
+        out = _Grouping.__new__(_Grouping)
+        out.names = copy.deepcopy(self.names)
+        out._named_columns = copy.deepcopy(self._named_columns)
+        out._key_columns = [col.copy(deep=deep) for col in self._key_columns]
         return out
 
 
--- a/python/cudf/cudf/core/groupby/groupby.py
+++ b/python/cudf/cudf/core/groupby/groupby.py
@@ -262,7 +262,7 @@ class GroupBy(Serializable, Reducible, Scannable):
         """
         self.obj = obj
         self._as_index = as_index
-        self._by = by
+        self._by = by.copy(deep=True) if isinstance(by, _Grouping) else by
         self._level = level
         self._sort = sort
         self._dropna = dropna
@@ -388,9 +388,7 @@ class GroupBy(Serializable, Reducible, Scannable):
                     len(self.obj), "int8", masked=False
                 )
             )
-            .groupby(
-                self.grouping.copy(), sort=self._sort, dropna=self._dropna
-            )
+            .groupby(self.grouping, sort=self._sort, dropna=self._dropna)
             .agg("size")
         )
 
@@ -406,7 +404,7 @@ class GroupBy(Serializable, Reducible, Scannable):
                 ),
                 index=self.obj.index,
             )
-            .groupby(self.grouping.copy(), sort=self._sort)
+            .groupby(self.grouping, sort=self._sort)
             .agg("cumcount")
         )
 
@@ -1909,7 +1907,7 @@ class GroupBy(Serializable, Reducible, Scannable):
 
         column_pair_groupby = cudf.DataFrame._from_data(
             column_pair_structs
-        ).groupby(by=self.grouping.keys.copy())
+        ).groupby(by=self.grouping.keys)
 
         try:
             gb_cov_corr = column_pair_groupby.agg(func)
@@ -2293,7 +2291,7 @@ class GroupBy(Serializable, Reducible, Scannable):
             )
 
         filled = self.fillna(method=fill_method, limit=limit)
-        fill_grp = filled.groupby(self.grouping.copy())
+        fill_grp = filled.groupby(self.grouping)
         shifted = fill_grp.shift(periods=periods, freq=freq)
         return (filled / shifted) - 1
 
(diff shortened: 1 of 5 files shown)


python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -261,17 +262,17 @@ def __init__(
         """
         self.obj = obj
         self._as_index = as_index
-        self._by = by
+        self._by = by.copy(deep=True) if isinstance(by, _Grouping) else by
```
galipremsagar, line 265: Should we rather move the task of making a deep copy to the caller's of this constructor?

bdice: Can you discuss a bit more what this would mean? How many places call this code? If you delegate this task to the callers of the constructor, are there places where the deep copy _isn't_ needed?

mroeschke: I don't know how this code really works, but there is something fishy about just setting `_Grouping._obj` to the new `self.obj` since `_Grouping._key_columns` and `_Grouping.names` could have been derived from the old `self.obj`

galipremsagar: @bdice Made a change such that only the caller's take the responsibility to make deep copies: https://github.com/rapidsai/cudf/pull/13932/commits/e256d9660f1cf7d98e62565a9b04149d3664c023

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -2336,7 +2339,7 @@ class DataFrameGroupBy(GroupBy, GetAttrGetItemMixin):
 
     def __getitem__(self, key):
         return self.obj[key].groupby(
-            by=self.grouping.keys,
+            by=self.grouping.keys.copy(),
```
wence-, line 2342: This doesn't need to be copied I think.

python/cudf/cudf/core/resample.py
```diff
@@ -95,6 +121,35 @@ class _ResampleGrouping(_Grouping):
 
     bin_labels: cudf.core.index.Index
 
+    def copy(self, deep=True):
+        out = super().copy(deep=deep)
+        out.bin_labels = self.bin_labels.copy(deep=deep)
+        return out
+
+    def serialize(self):
```
wence-, line 129: Are these new methods needed for the `copy` implementation, or something else?

python/cudf/cudf/core/series.py
```diff
@@ -3096,7 +3096,7 @@ def value_counts(
             res = self.groupby(series_bins, dropna=dropna).count(dropna=dropna)
             res = res[res.index.notna()]
         else:
-            res = self.groupby(self, dropna=dropna).count(dropna=dropna)
+            res = self.groupby(self.copy(), dropna=dropna).count(dropna=dropna)
```
wence-, line 3099: This is a series and doesn't need to be copied.

python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -1906,7 +1909,7 @@ def _cov_or_corr(self, func, method_name):
 
         column_pair_groupby = cudf.DataFrame._from_data(
             column_pair_structs
-        ).groupby(by=self.grouping.keys)
+        ).groupby(by=self.grouping.keys.copy())
```
wence-, line 1912: I think this doesn't need copied.

python/cudf/cudf/core/groupby/groupby.py
wence-: Sorry, I don't like this change in e256d96. `groupby(by=existing_grouper)` is now still unsafe by default and requires the caller to remember to make the copy. Since, AFAICT, they _always_ have to make the copy, let's do it for them. Additionally, they only have to make the copy when `by` is an existing `_Grouping` object (if it's just the keys, that's fine since those won't be modified), so to correctly determine when to copy needs inappropriate intimacy with the internal implementation of `GroupBy` (they need to inspect the `grouper._by` slot and check it's a `_Grouping` which feels like an internal implementation detail).

wence- (changes requested):

python/cudf/cudf/core/groupby/groupby.py
bdice: > Since, AFAICT, they _always_ have to make the copy, let's do it for them.

Right, this is the crux of what I was asking earlier:

> If you delegate this task to the callers of the constructor, are there places where the deep copy _isn't_ needed?

I'm still parsing through this but my initial impression aligns with what @wence- is saying. Maybe revert e256d96, then? @wence- Are you generally happy with the state before e256d96?

galipremsagar: Looks like we are tending towards the initial version. I reverted the changes and moved copies to the constructor only when the `by` object is a `Grouping`.

wence-: Thanks

wence- (approved):

wence-:
Serialization test looks like it fails https://github.com/rapidsai/cudf/actions/runs/5953195610/job/16147414724?pr=13932

galipremsagar:
> Serialization test looks like it fails https://github.com/rapidsai/cudf/actions/runs/5953195610/job/16147414724?pr=13932

Fixed it 👍

bdice (approved):

Issue #13873, opened 2023-08-14
[BUG] JIT `GroupBy.apply` aggregations with `int32` dtype may return different results from pandas when large values are present


brandon-b-miller:
**Describe the bug**
When executing a `sum` aggregation in a groupby apply UDF, the result can diverge from pandas in some cases, specifically when large values are present. This is likely due to some result not being promoted to `int64` somewhere it ought to be. 

**Steps/Code to reproduce bug**
```python
import pandas as pd
import cudf
import numpy as np

int32_max = np.iinfo(np.int32).max

df = cudf.DataFrame(
    {
        'a':[0,0,0,0,0,0,0,0,0,0],
        'b':[int32_max,2,3,4,5,6,7,8,9,10]        
    }
)

df['b'] = df['b'].astype('int32')
pdf = df.to_pandas()

def func(group):
    return group['b'].sum()

pandas_result = pdf.groupby('a').apply(func)
cudf_result = df.groupby('a').apply(func, engine='jit')

print(pandas_result)
print(cudf_result)
```

**Expected behavior**
Results should be the same as pandas.

**Environment overview (please complete the following information)**
Bare-metal, 23.10

**Additional context**
came up during https://github.com/rapidsai/cudf/pull/13813. The above test fails for `mean`, `var`, and `std` as well. Since these three reductions also involve a summation, I suspect the root cause could be the same.

PR #13943, opened 2023-08-23
Fix integer overflow in shim `device_sum` functions

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/13873


python/cudf/udf_cpp/shim.cu
```diff
@@ -388,25 +388,58 @@ __device__ bool are_all_nans(cooperative_groups::thread_block const& block,
   return count == 0;
 }
 
-template <typename T>
-__device__ void device_sum(cooperative_groups::thread_block const& block,
-                           T const* data,
-                           int64_t size,
-                           T* sum)
+template <typename T, std::enable_if_t<std::is_integral_v<T>, int> = 0>
+__device__ int64_t device_sum(cooperative_groups::thread_block const& block,
```
bdice, line 392: Rather than writing an entirely new function, could we have a single kernel and write something like this in a second template parameter:

`typename AccumT = std::conditional<std::is_integral_v<T>, int64_t, T>`

brandon-b-miller: yes! was hoping someone would comment with something like this :)

bdice: giphy

brandon-b-miller: Should be updated! Looking much cleaner now, thanks.

wence-:
I think this makes sense. Of course, there are many ways to still get a bad result, but pandas also gets a bad result in those cases. Overflowing int64 is the natural one, but pandas hilariously converts a sum of _`unsigned`_ items to a signed type, so that's a more natural other one.

wence- (approved):
Python test changes

python/cudf/udf_cpp/shim.cu
```diff
@@ -388,38 +388,38 @@ __device__ bool are_all_nans(cooperative_groups::thread_block const& block,
...
 
-template <typename T>
-__device__ T BlockSum(T const* data, int64_t size)
+template <typename T, typename AccumT = std::conditional_t<std::is_integral_v<T>, int64_t, T>>
+__device__ AccumT BlockSum(T const* data, int64_t size)
 {
   auto block = cooperative_groups::this_thread_block();
 
-  if constexpr (std::is_floating_point_v<T>) {
+  if constexpr (std::is_floating_point_v<AccumT>) {
```
bdice, line 418: This should still be `T`, I think.
```suggestion
  if constexpr (std::is_floating_point_v<T>) {
```

bdice (approved):

Issue #13735, opened 2023-07-23
[FEA] Add `bytes_per_second` to all libcudf benchmarks


GregoryKimball:
Many libcudf benchmarks report the `bytes_per_second` processed as part of the output data. This value is useful for comparing benchmarks because it normalizes the increasing execution time from processing more data. Also, `bytes_per_second` and `real_time_s` together let us compute `bytes_processed` values which serve as a useful X-axis.

As of the end of 23.08 development, many benchmarks still do not report `bytes_per_second` in the output data. Here is a figure summarizing the metrics reported by the benchmark suite.

![image](https://github.com/rapidsai/cudf/assets/12725111/c132ef5f-2a1b-4c89-b6e1-a03fa19fe287)

| benchmark | status | notes | 
|---|---|---|
| `BINARYOP` | | |
| `COPY_IF_ELSE` | | |
| `GROUPBY` | | |
| `GROUPBY_NV` | | |
| `HASHING` | | |
| `JOIN` | | |
| `JOIN_NV` | | |
| `QUANTILES` | | |
| `REDUCTION` | | |
| `REPLACE` | | |
| `SEARCH` | | |
| `SEARCH_NV` | | |
| `SET_OPS_NV` | | |
| `SHIFT` | | |
| `SORT` | | |
| `SORT_NV` | | |
| `STREAM_COMPACTION_NV` | | |
| `TRANSPOSE` | | |

Note: For this tracking list, cuIO benchmarks are omitted because "encoded file size" serves a similar purpose.

Blonck:
Hi, I want to start contributing to cudf and would like to work on this issue.

Should I create one PR per benchmark or put all in one PR?

jihoonson:
It is a good idea to have some metrics for data size. I wonder whether it would be even better if we have separate metrics for input and output data. For example, we could add custom metrics like `input_items_per_second`, `output_items_per_second`, and `input_bytes_per_second` and `output_bytes_per_second`. The first two are the number of items (usually rows), and the last two are the byte sizes of input and output.

GregoryKimball:
Thank you @jihoonson for your suggestion. I agree that we need better and more granular throughput metrics. I would like to start with an input+output bytes per second metric that is consistently implemented in all benchmarks. From there we can add the separate input/output and bytes/items metrics as you suggest.

jihoonson:
[...]

PR #13950, opened 2023-08-24
Add `bytes_per_second` to shift benchmark

Blonck:
Adds `bytes_per_second` to `SHIFT_BENCH`. 

This patch relates to #13735.

## Checklist


cpp/benchmarks/copying/shift.cu
```diff
@@ -68,6 +68,17 @@ static void BM_shift(benchmark::State& state)
     cuda_event_timer raii(state, true);
     auto output = cudf::shift(input, offset, *fill);
   }
+
+  auto const elems_read    = (size - offset);
+  auto const elems_written = use_validity ? (size - offset) : size;
```
karthikeyann, line 73: nit: Why is the `elems_written` changing based on `use_validity`?

Blonck: If `use_validity` is false, the fill value is a number, and the full column must be written. If `use_validity` is true, the filled values can be null, and writing only the elements which can be shifted + the bitmask would be sufficient.

karthikeyann (approved):
LGTM 👍

cpp/benchmarks/copying/shift.cu
ttnghia: Can you add this comment into the code please?

cpp/benchmarks/copying/shift.cu
```diff
@@ -68,6 +68,17 @@ static void BM_shift(benchmark::State& state)
...
+  auto const bytes_read    = elems_read * sizeof(int);
+  auto const bytes_written = elems_written * sizeof(int);
+
+  // null bytes for input and output
+  auto const null_bytes = use_validity ? 2 * cudf::bitmask_allocation_size_bytes(size) : 0;
```
ttnghia, line 78: Please add a comment here too, clarifying why we multiply by 2.

Blonck: Done.

PS: here is the perf log:  [perf_log.txt](https://github.com/rapidsai/cudf/files/12568659/perf_log.txt)

ttnghia (approved):

Blonck:
CI for conda-python-build / build (12.0.1, ubuntu22.04, amd64, 3.9) with a conda error. I assume that it's not related to this patch :). I would rebase the PR in a few days onto branch-23.10 and push it again.

cpp/benchmarks/copying/shift.cu
```diff
@@ -68,6 +68,19 @@ static void BM_shift(benchmark::State& state)
     cuda_event_timer raii(state, true);
     auto output = cudf::shift(input, offset, *fill);
   }
+
+  auto const elems_read = (size - offset);
+  auto const bytes_read = elems_read * sizeof(int);
```
harrism, line 73: I would like to make the same suggestion that I made on this PR: https://github.com/rapidsai/cudf/pull/14170

Please don't hard code the type here in case the type used in the benchmark changes. Thanks!

harrism (changes requested):
One small request a la #14170.

harrism:
Remember to run `pre-commit run` before commiting to avoid clang-format errors. (Or format on save).

harrism (approved):

cpp/benchmarks/copying/shift.cu
```diff
@@ -56,10 +56,9 @@ static void BM_shift(benchmark::State& state)
...
-  auto const input_table =
-    create_sequence_table({cudf::type_to_id<int>()},
-                          row_count{size},
-                          use_validity ? std::optional<double>{1.0} : std::nullopt);
+  auto constexpr column_type_id = cudf::type_to_id<int>();
+  auto const input_table        = create_sequence_table(
+    {column_type_id}, row_count{size}, use_validity ? std::optional<double>{1.0} : std::nullopt);
   cudf::column_view input{input_table->get_column(0)};
 
   auto fill = use_validity ? make_scalar<int>() : make_scalar<int>(777);
```
harrism, line 64: Match the transpose benchmark approach.

```suggestion
  constexpr auto column_type_id = cudf::type_id::INT32;
  using column_type = cudf::id_to_type<column_type_id>;
  auto const input_table        = create_sequence_table(
    {column_type_id}, row_count{size}, use_validity ? std::optional<double>{1.0} : std::nullopt);
  cudf::column_view input{input_table->get_column(0)};

  auto fill = use_validity ? make_scalar<column_type>() : make_scalar<column_type>(777);
```

cpp/benchmarks/copying/shift.cu
```diff
@@ -68,6 +67,19 @@ static void BM_shift(benchmark::State& state)
     cuda_event_timer raii(state, true);
     auto output = cudf::shift(input, offset, *fill);
   }
+
+  auto const elems_read = (size - offset);
+  auto const bytes_read = elems_read * sizeof(int);
```
harrism, line 72:
```suggestion
  auto const bytes_read = elems_read * sizeof(column_type);
```

cpp/benchmarks/copying/shift.cu
```diff
@@ -68,6 +67,19 @@ static void BM_shift(benchmark::State& state)
...
+
+  // If 'use_validity' is false, the fill value is a number, and the entire column
+  // (excluding the null bitmask) needs to be written. On the other hand, if 'use_validity'
+  // is true, only the elements that can be shifted are written, along with the full null bitmask.
+  auto const elems_written = use_validity ? (size - offset) : size;
+  auto const bytes_written = elems_written * sizeof(cudf::id_to_type<column_type_id>);
```
harrism, line 78:
```suggestion
  auto const bytes_written = elems_written * sizeof(column_type);
```

harrism (changes requested):
Sorry, just trying to consolidate the use of types.

Blonck:
> Sorry, just trying to consolidate the use of types.

No, problem.

Just to mention it: With the latest change, technically, it is not longer the same benchmark as it does not longer depend on the width of `int` :) (I guess on platform which anyways are not supported.)

harrism (approved):

Issue #13897, opened 2023-08-17
[BUG] Groupby-apply ignores `as_index=` argument


shwina:
The `as_index=` argument is ignored by groupby-apply:

```python
In [2]: df = cudf.DataFrame({'a': [1, 1, 2], 'b': [1, 2, 3]})
In [3]: df.groupby('a', as_index=False).apply(lambda x: x['b'].sum())
Out[3]:
a
1    3
2    3
dtype: int64

In [4]: df.to_pandas().groupby('a', as_index=False).apply(lambda x: x['b'].sum())
Out[4]:
   a  None
0  1     3
1  2     3
```

brandon-b-miller:
I think I know what might be happening here, will take a look!

PR #13951, opened 2023-08-24
Handle `as_index` in `GroupBy.apply`

brandon-b-miller:
Closes https://github.com/rapidsai/cudf/issues/13897

### DIFF
--- a/python/cudf/cudf/core/groupby/groupby.py
+++ b/python/cudf/cudf/core/groupby/groupby.py
@@ -1436,6 +1436,9 @@ class GroupBy(Serializable, Reducible, Scannable):
 
         if self._sort:
             result = result.sort_index()
+        if self._as_index is False:
+            result = result.reset_index()
+            result[None] = result.pop(0)
         return result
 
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/groupby/groupby.py
```diff
@@ -1436,6 +1436,9 @@ def mult(df):
 
         if self._sort:
             result = result.sort_index()
+        if self._as_index is False:
+            result = result.reset_index()
+            result[None] = result.pop(0)
```
galipremsagar, line 1441:
```suggestion
            result = result.reset_index(drop=True)
```
Could we do this instead?

brandon-b-miller: When I try that, I get a `Series` still. We're trying to end up with a `DataFrame` with a column named `None`:

```
0    1.0
1    2.5
dtype: float64
```

vs
```
   y  None
0  0   1.0
1  1   2.5
```

galipremsagar (approved):

Issue #8175, opened 2021-05-06
[BUG] Series unique should return an array rather than a dataframe


beckernick:
Pandas's unique method returns an array, while ours returns a Series.

```python
import cudf
print(cudf.__version__)
​
df = cudf.DataFrame({
    "a": [0,1,2,0,1,2,0]
})
type(df['a'].unique()), type(df['a'].to_pandas().unique())
0.20.0a+248.g7623f3978b
(cudf.core.series.Series, numpy.ndarray)
```

kkraus14:
Hmm, this is problematic for us in handling things like strings or decimal values which can't be supported by arrays as of now. @shwina is scoping building `cudf.arrays` similar to `pandas.arrays` which would potentially give a path forward in the future.

vyasr:
I'm not sure what we should do about this. I don't think extension arrays are on our roadmap for any time soon. @galipremsagar WDYT? We could match the type for types supported by cupy, but returning a different object based on the data type seems even worse than the status quo. Fundamentally I don't know that we'll ever be able to solve this since we have other types that cupy will probably never support (list and struct, for instance).

PR #13959, opened 2023-08-25
Add pandas compatible output to `Series.unique`

galipremsagar:
## Description
Partially addresses #8175
This PR makes changes to `Series.unique`, where a cupy array is returned to match `pd.Series.unique` where a numpy array is returned.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/series.py
+++ b/python/cudf/cudf/core/series.py
@@ -2970,6 +2970,8 @@ class Series(SingleColumnFrame, IndexedFrame, Serializable):
         dtype: object
         """
         res = self._column.unique()
+        if cudf.get_option("mode.pandas_compatible"):
+            return res.values
         return Series(res, name=self.name)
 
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/series.py
```diff
@@ -2970,6 +2970,8 @@ def unique(self):
         dtype: object
         """
         res = self._column.unique()
+        if cudf.get_option("mode.pandas_compatible"):
+            return res.values
         return Series(res, name=self.name)
```
wence-, line 2975: Only doing this in pandas-compat mode because this marks the buffer as exposed, and may not work for string/struct types?

wence- (approved):
I think this makes sense.

python/cudf/cudf/core/series.py
galipremsagar: Yup, will only work for numeric types (int, float..)

Issue #13771, opened 2023-07-26
[BUG] Errors converting tables from arrow to cuDF


pentschev:
**Describe the bug**
We've been investigating an issue with Dask's new P2P shuffle algorithm where arrow tables are [converted to cuDF in this line](https://github.com/rapidsai/cudf/blob/branch-23.08/python/cudf/cudf/core/frame.py#L1104), ultimately causing the following error:

```c++
*** MemoryError: std::bad_alloc: CUDA error at: /datasets/pentschev/miniconda3/envs/cudf-invalid-address-src-230726/include/rmm/mr/device/cuda_async_view_memory_resource.hpp:121: cudaErrorIllegalAddress an illegal memory access was encountered
```

The error above is irrecoverable from a CUDA perspective, if we attach a debugger and attempt to call any CUDA functions we get the same `cudaErrorIllegalAddress` for any subsequent calls.  We don't have a clear understanding as to what causes that, various attempts at understanding the real source were frustrated. The closest we've got to understand what is going on was by converting the data to another format (e.g., pandas) and back, which succeeds. All workers fail at the same time while executing the same line but on different chunks of data, also hinting at something going on with the input arrow data. Essentially, the following change allows the workflow to complete successfully:

<details><summary>Workaround patch</summary>

```python
diff --git a/python/cudf/cudf/core/frame.py b/python/cudf/cudf/core/frame.py
index dcf44344cd..0ce092dc53 100644
--- a/python/cudf/cudf/core/frame.py
+++ b/python/cudf/cudf/core/frame.py
@@ -1101,7 +1101,11 @@ class Frame(BinaryOperand, Scannable):
         cudf_non_category_frame = {
             name: col
             for name, col in zip(
-                data.column_names, libcudf.interop.from_arrow(data)
+                data.column_names,
+                cudf._lib.interop.from_arrow(
+                    pa.Table.from_pandas(data.to_pandas())
+                ),
+
             )
         }

```

</details>

That seems to point at some potential error when parsing the input data that is somehow fixed when converting to pandas and back. With that in mind, I attempted to store the data to disk to both ORC and parquet and see what would happen. With ORC, storing to disk and reading it back also completes successfully, just like pandas. However, with parquet data we get a different error:

```python
*** OverflowError: CUDF failure at: /datasets/pentschev/src/cudf-invalid-address/cudf/cpp/src/copying/concatenate.cu:421: Total number of concatenated chars exceeds the column size limit
```

[...]

PR #13968, opened 2023-08-25
Use cudf::thread_index_type in strings custom kernels

davidwendt:
## Description
Adds `cudf::thread_index_type` usage when calculating the thread index in custom kernels in `src/strings/attributes.cu` and `src/strings/convert/convert_urls.cu`

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #13971, opened 2023-08-25
Use thread_index_type to avoid out of bounds accesses in conditional joins

vyasr:
## Description

See #10368 (and more recently #13771

## Checklist


cpp/src/join/conditional_join_kernels.cuh
```diff
@@ -67,20 +67,21 @@ __global__ void compute_conditional_join_output_size(
     &intermediate_storage[threadIdx.x * device_expression_data.num_intermediates];
 
   std::size_t thread_counter{0};
-  cudf::size_type const start_idx      = threadIdx.x + blockIdx.x * block_size;
-  cudf::size_type const stride         = block_size * gridDim.x;
-  cudf::size_type const left_num_rows  = left_table.num_rows();
-  cudf::size_type const right_num_rows = right_table.num_rows();
-  auto const outer_num_rows            = (swap_tables ? right_num_rows : left_num_rows);
-  auto const inner_num_rows            = (swap_tables ? left_num_rows : right_num_rows);
+  cudf::thread_index_type const start_idx      = threadIdx.x + blockIdx.x * block_size;
```
bdice, line 70: The results of these computations can overflow. You have to cast the rhs values before doing math on them.

cpp/src/join/conditional_join_kernels.cuh
```diff
@@ -67,20 +67,21 @@ __global__ void compute_conditional_join_output_size(
...
+  cudf::thread_index_type const stride         = thread_index_type{block_size} * gridDim.x;
+  cudf::thread_index_type const left_num_rows  = left_table.num_rows();
+  cudf::thread_index_type const right_num_rows = right_table.num_rows();
```
bdice, line 73: FYI: You don't _have_ to upcast these values, because normal integer casting rules for the comparisons (`outer_row_index < outer_num_rows`) will make sure that they're upcast.

cpp/src/join/conditional_join_kernels.cuh
```diff
@@ -161,18 +162,18 @@ __global__ void conditional_join(table_device_view left_table,
...
+  cudf::thread_index_type const outer_num_rows = (swap_tables ? right_num_rows : left_num_rows);
+  cudf::thread_index_type const inner_num_rows = (swap_tables ? left_num_rows : right_num_rows);
 
   if (0 == lane_id) { current_idx_shared[warp_id] = 0; }
 
   __syncwarp();
 
-  cudf::size_type outer_row_index = threadIdx.x + blockIdx.x * block_size;
+  cudf::thread_index_type outer_row_index = threadIdx.x + blockIdx.x * block_size;
```
bdice, line 176: The results of these computations can overflow. You have to cast the rhs values before doing math on them.

cpp/src/join/conditional_join_kernels.cuh
davidwendt: You can replace this with one line
```
auto const start_idx = cudf::detail::grid_1d::global_thread_id();
```
Reference:
https://github.com/rapidsai/cudf/pull/13968/files#diff-b601b88e04ff05d909c800466b031c6e097fef812300cb2350c8709d6cd37fe2R114

vyasr: Nice thanks David. Done.

cpp/src/join/conditional_join_kernels.cuh
vyasr: Switched to `cudf::detail::grid_1d::global_thread_id()`, which upcasts the parameters appropriately.

cpp/src/join/conditional_join_kernels.cuh
vyasr: Figured it was safer in case there was any arithmetic I missed.

cpp/src/join/conditional_join_kernels.cuh
```diff
@@ -67,20 +67,21 @@ __global__ void compute_conditional_join_output_size(
...
-  cudf::size_type const left_num_rows  = left_table.num_rows();
-  cudf::size_type const right_num_rows = right_table.num_rows();
-  auto const outer_num_rows            = (swap_tables ? right_num_rows : left_num_rows);
-  auto const inner_num_rows            = (swap_tables ? left_num_rows : right_num_rows);
+  auto const start_idx                         = cudf::detail::grid_1d::global_thread_id();
+  cudf::thread_index_type const stride         = thread_index_type{block_size} * gridDim.x;
+  cudf::thread_index_type const left_num_rows  = left_table.num_rows();
+  cudf::thread_index_type const right_num_rows = right_table.num_rows();
+  auto const outer_num_rows                    = (swap_tables ? right_num_rows : left_num_rows);
+  auto const inner_num_rows                    = (swap_tables ? left_num_rows : right_num_rows);
```
davidwendt, line 75: The `auto` decls here look like an int64 type will be passed to the `evaluator.evaluate()` function which only accepts `size_type`. What happens in this case? Seems the compiler should've complained about this.

cpp/src/join/conditional_join_kernels.cuh
```diff
@@ -67,20 +67,21 @@ __global__ void compute_conditional_join_output_size(
...
 
   std::size_t thread_counter{0};
-  cudf::size_type const start_idx      = threadIdx.x + blockIdx.x * block_size;
-  cudf::size_type const stride         = block_size * gridDim.x;
-  cudf::size_type const left_num_rows  = left_table.num_rows();
-  cudf::size_type const right_num_rows = right_table.num_rows();
-  auto const outer_num_rows            = (swap_tables ? right_num_rows : left_num_rows);
-  auto const inner_num_rows            = (swap_tables ? left_num_rows : right_num_rows);
+  auto const start_idx                         = cudf::detail::grid_1d::global_thread_id();
+  cudf::thread_index_type const stride         = thread_index_type{block_size} * gridDim.x;
```
bdice, line 71: Are we sure that `block_size` can be replaced with `blockDim.x`? If so, let's wait for #13996 and use `cudf::detail::grid_1d::grid_stride()`. If not, the use of `global_thread_id` isn't valid either.

cpp/src/join/conditional_join_kernels.cuh
vyasr: I'll change the declarations of `[left|right]_row_index` below to perform the cast so that it's explicit. I don't think the compiler can warn in this case, narrowing larger integral types to smaller ones doesn't usually does it? At least from a few quick godbolt experiments I can't trigger an error without going from float->int or by introducing arithmetic operations that trigger overflow (at which point it's really warning about the overflow, not the narrowing).

davidwendt: I thought I saw a warning for this kind of thing when compiled in debug mode but I also was not able to recreate it in godbolt. So I think your solution looks good.

cpp/src/join/conditional_join_kernels.cuh
```diff
@@ -67,23 +67,25 @@ __global__ void compute_conditional_join_output_size(
...
 
   std::size_t thread_counter{0};
-  cudf::size_type const start_idx      = threadIdx.x + blockIdx.x * block_size;
-  cudf::size_type const stride         = block_size * gridDim.x;
-  cudf::size_type const left_num_rows  = left_table.num_rows();
-  cudf::size_type const right_num_rows = right_table.num_rows();
-  auto const outer_num_rows            = (swap_tables ? right_num_rows : left_num_rows);
-  auto const inner_num_rows            = (swap_tables ? left_num_rows : right_num_rows);
+  auto const start_idx                 = cudf::detail::grid_1d::global_thread_id();
+  cudf::thread_index_type const stride = cudf::detail::grid_1d::grid_stride();
```
PointKernel, line 71:
```suggestion
  atuo const stride = cudf::detail::grid_1d::grid_stride();
```
non-blocking nit

vyasr, line 71:
```suggestion
  auto const stride = cudf::detail::grid_1d::grid_stride();
```

vyasr: Done

cpp/src/join/conditional_join_kernels.cuh
```diff
@@ -67,23 +67,25 @@ __global__ void compute_conditional_join_output_size(
...
   std::size_t thread_counter{0};
-  cudf::size_type const start_idx      = threadIdx.x + blockIdx.x * block_size;
-  cudf::size_type const stride         = block_size * gridDim.x;
-  cudf::size_type const left_num_rows  = left_table.num_rows();
-  cudf::size_type const right_num_rows = right_table.num_rows();
-  auto const outer_num_rows            = (swap_tables ? right_num_rows : left_num_rows);
-  auto const inner_num_rows            = (swap_tables ? left_num_rows : right_num_rows);
+  auto const start_idx = cudf::detail::grid_1d::global_thread_id();
+  auto const stride    = cudf::detail::grid_1d::grid_stride();
+  ;
```
bdice, line 72:
```suggestion
```

bdice (approved):

PointKernel (approved):

davidwendt (approved):

PR #13983, opened 2023-08-28
Fix Arrow package resolution when Arrow is pre-installed on the system.

cwharris:
## Description
Call `find_package` for Arrow before `rapids_cpm_find` to prevent duplicate target declarations while still setting Arrow variables in the appropriate scope on systems where Arrow is pre-installed.

## Checklist

### DIFF
--- a/cpp/cmake/thirdparty/get_arrow.cmake
+++ b/cpp/cmake/thirdparty/get_arrow.cmake
@@ -161,10 +161,10 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
   endif()
 
   # Arrow sets a number of variables that we need in this scope which won't be set via
-  # rapids_cpm_find because it is a function. Calling find_package will import targets which will
-  # duplicate targets generated from rapids_cpm_find builds, so we must call it before
-  # rapids_cpm_find has a chance to build and generate those targets or we get errors when an
-  # appropriate version of Arrow is found pre-installed on the system.
+  # rapids_cpm_find because it is a function. Calling find_package will import targets which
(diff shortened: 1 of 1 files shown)


cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -160,6 +160,13 @@ function(find_and_configure_arrow VERSION BUILD_STATIC ENABLE_S3 ENABLE_ORC ENAB
     list(APPEND ARROW_PARQUET_OPTIONS "ARROW_DEPENDENCY_SOURCE AUTO")
   endif()
 
+  # Arrow sets a number of variables that we need in this scope which won't be set via
+  # rapids_cpm_find because it is a function. Calling find_package will import targets which
+  # will duplicate targets generated from rapids_cpm_find builds, so we must call it before
+  # rapids_cpm_find has a chance to build and generate those targets or we get errors when
+  # an appropriate version of Arrow is found pre-installed on the system.
+  find_package(Arrow QUIET)
```
robertmaynard, line 168: By moving this before `rapids_cpm_find` it breaks all the cpm controls over arrow. Specifically it will not allow users to force a source build of arrow when a local arrow exists.

You will need to do the following:

- Specify the version of arrow we need, as we can't allow searches for differing values
- check the CPM variables for global and per project build from source and obey those.

robertmaynard (changes requested):

cpp/cmake/thirdparty/get_arrow.cmake
cwharris: Per talking offline, it sounds like the right approach is to make Arrow's config idempotent.

PR #13984, opened 2023-08-28
Add `bytes_per_second` to groupby max benchmark.

Blonck:
This patch adds memory statistics for the `GROUPBY_NVBENCH` benchmark using the `max` aggregation.

For this purpose helper functions are introduced to compute the payload size for:
  - Column
  - Table
  - Groupby execution results

This patch relates to #13735.

## Checklist

### DIFF
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -40,8 +40,9 @@ target_include_directories(
 
 # Use an OBJECT library so we only compile these helper source files only once
 add_library(
-  cudf_benchmark_common OBJECT "${CUDF_SOURCE_DIR}/tests/utilities/base_fixture.cpp"
-                               synchronization/synchronization.cpp io/cuio_common.cpp
+  cudf_benchmark_common OBJECT
+  "${CUDF_SOURCE_DIR}/tests/utilities/base_fixture.cpp" synchronization/synchronization.cpp
+  io/cuio_common.cpp common/memory_statistics.cpp
 )
 target_link_libraries(cudf_benchmark_common PRIVATE cudf_datagen $<TARGET_NAME_IF_EXISTS:conda_env>)
 add_custom_command(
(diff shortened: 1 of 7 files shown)


Blonck:
Hi, I added a few helper function which I would use also for the other groupby benchmarks.

PS: Should I add the other `groupby` benchmarks to this PR to reduce the amount of PR for #13735 a bit?

PointKernel:
Would other benchmarks benefit from those helper functions as well? If not, we can move them to `benchmarks
/groupby/group_common.hpp`.

> Should I add the other groupby benchmarks to this PR to reduce the amount of PR for #13735 a bit?

Sounds reasonable to me.

Blonck:
> Would other benchmarks benefit from those helper functions as well? If not, we can move them to `benchmarks

I think so. They could be used in a few of the benchmarks to simplify the code.

Blonck:
Hi, the PR is not ready, I had not enough time the last two days. So no need to review it :smiley:.

Blonck:
Hi @PointKernel , @davidwendt,

in the current state of the PR every NVBENCH benchmark has `bytes_per_second` added. However, I'm a bit unhappy with its current state. Specifically, I'd like to introduce test code to validate the calculations for required_bytes, especially when dealing with nested column types.
I'm considering adding these utility functions in the `test/utilities` directoriy and creating corresponding tests there. I've noticed that similar utility functions used in the tests are also employed within the benchmarks. However, I'm think that housing benchmark-specific helper functions under the test directories is not optimal. Do you have any better ideas?

jihoonson:
Hi @Blonck, this is nice work and I'd like to get this change merged. However, it seems that you haven't had a chance to finish up this. Would you mind if I take this from where you left off and finish up?

davidwendt:
You may want to look into using `cudf::row_bit_count` to get the column/table sizes: https://docs.rapids.ai/api/libcudf/stable/group__transformation__transform.html#gaa78354f8eda093519182149710215d6f
which already handles all cudf types. The output is the number of bits per row but you can use `cudf::reduce` to get the total size and divide by 8 to get the number of bytes.

jihoonson:
Thanks @davidwendt, this is very useful! I will check out the function you suggested.

vyasr:
Thanks for working on this @Blonck! It looks like this was a helpful starting point for #16126, which ultimately superseded this PR.

PR #13991, opened 2023-08-29
Use dev_string_lengths.

bdice:
## Description
It seems like a device lambda is accessing a host `std::vector` when it should be accessing a device uvector instead.

## Checklist

### DIFF
--- a/java/src/main/native/src/row_conversion.cu
+++ b/java/src/main/native/src/row_conversion.cu
@@ -2220,7 +2220,7 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const &input,
     std::vector<rmm::device_uvector<char>> string_data_cols;
     std::vector<size_type *> string_col_offset_ptrs;
     std::vector<char *> string_data_col_ptrs;
-    for (auto &col_string_lengths : string_lengths) {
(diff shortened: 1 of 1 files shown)


java/src/main/native/src/row_conversion.cu
```diff
@@ -2220,7 +2220,7 @@ std::unique_ptr<table> convert_from_rows(lists_column_view const &input,
     std::vector<rmm::device_uvector<char>> string_data_cols;
     std::vector<size_type *> string_col_offset_ptrs;
     std::vector<char *> string_data_col_ptrs;
-    for (auto &col_string_lengths : string_lengths) {
+    for (auto &col_string_lengths : dev_string_lengths) {
```
bdice, line 2223: @hyperbolic2346 On second thought, I'm not sure about this change. I think the `string_lengths` vector is a host vector of device pointers, so it seems like binding `auto &col_string_lengths : string_lengths` by value into the lambda `tmp` is actually just binding a device pointer by value into the lambda -- there's no issue with using that on device. It's awkward, yes, but this is probably correct with no changes. If you agree let's close this PR.

bdice:
Discussed offline with @hyperbolic2346. This code is fine as-is. I'll close this PR.

PR #14003, opened 2023-08-30
Fix map column can not be non-nullable for java

res-life:
## Description
Make map column non-nullable for java.

Changes:
- Add a new method to pass nullable; Deprecate the old one.
- Update the tests.

## Checklist
Update existing tests
Bug fix, no need to update document.


java/src/main/java/ai/rapids/cudf/ColumnWriterOptions.java
```diff
@@ -522,7 +522,10 @@ protected String[] getFlatColumnNames(String[] ret) {
    * Maps are List columns with a Struct named 'key_value' with a child named 'key' and a child
    * named 'value'. The caller of this method doesn't need to worry about this as this method will
    * take care of this without the knowledge of the caller.
+   * Note: this method always returns a nullabe column, can not return non-nullable column.
```
ttnghia, line 525:
```suggestion
   *
   * Note: this method always returns a nullabe column, can not return non-nullable column.
```

java/src/main/java/ai/rapids/cudf/ColumnWriterOptions.java
```diff
@@ -537,6 +540,30 @@ public static ColumnWriterOptions mapColumn(String name, ColumnWriterOptions key
...
 
+  /**
+   * Add a Map Column to the schema.
+   * <p>
+   * Maps are List columns with a Struct named 'key_value' with a child named 'key' and a child
+   * named 'value'. The caller of this method doesn't need to worry about this as this method will
+   * take care of this without the knowledge of the caller.
+   * @param isNullable is the returned map nullable.
+   * Note: If this map column is a key of another map, should pass isNullable = false.
+   * e.g.: map1(map2(int, int), int) the map2 should be non-nullable.
```
ttnghia, line 551:
```suggestion
   * Add a Map Column to the schema.
   *
   * Maps are LIST columns with a STRUCT named 'key_value' with a child named 'key' and a child
   * named 'value'. The caller of this method doesn't need to worry about this as this method will
   * take care of this without the knowledge of the caller.
   *
   * Note: If this map column is a key of another map, should pass isNullable = false.
   * e.g.: map1(map2(int, int), int) the map2 should be non-nullable.
   *
   * @param isNullable is the returned map nullable.
```

java/src/main/java/ai/rapids/cudf/ColumnWriterOptions.java
```diff
@@ -537,6 +540,30 @@ public static ColumnWriterOptions mapColumn(String name, ColumnWriterOptions key
...
+   */
+  public static ColumnWriterOptions mapColumn(String name, ColumnWriterOptions key,
+                                              ColumnWriterOptions value, Boolean isNullable) {
+    StructColumnWriterOptions struct = structBuilder("key_value").build();
+    if (key.isNullable) {
+      throw new IllegalArgumentException("key column can not be nullable");
+    }
```
ttnghia, line 558:
```suggestion
    if (key.isNullable) {
      throw new IllegalArgumentException("key column can not be nullable");
    }
    StructColumnWriterOptions struct = structBuilder("key_value").build();
```

java/src/main/java/ai/rapids/cudf/ColumnWriterOptions.java
```diff
@@ -537,6 +540,30 @@ public static ColumnWriterOptions mapColumn(String name, ColumnWriterOptions key
...
+    struct.childColumnOptions = new ColumnWriterOptions[]{key, value};
+    ColumnWriterOptions opt = listBuilder(name, isNullable)
```
revans2, line 560: I am confused why we need a new API for this? Can we not just update the old API? It looks like it was a missed requirement and making something not support nulls should be fine for maps.

revans2 (commented):
Technically this is not a breaking change for CUDF, but because the plugin code fails if it is using a deprecated API it will break the spark plugin build.  Could you please have a PR ready to go in the Spark plugin so that the time we are broken is very short.

java/src/main/java/ai/rapids/cudf/ColumnWriterOptions.java
revans2: Sorry forgot to delete this one when I understood the issue better.

res-life:
Corresponding PR: https://github.com/NVIDIA/spark-rapids/pull/9147

java/src/main/java/ai/rapids/cudf/ColumnWriterOptions.java
res-life: Done

java/src/main/java/ai/rapids/cudf/ColumnWriterOptions.java
res-life: Done

java/src/main/java/ai/rapids/cudf/ColumnWriterOptions.java
res-life: Done

revans2 (approved):
I would like to see the changes the Spark plugin PR approved before this is merged in.

res-life:
ok to test

res-life:
build

res-life:
ok to test

res-life:
rerurn tests

java/src/main/java/ai/rapids/cudf/ColumnWriterOptions.java
```diff
@@ -522,7 +522,11 @@ protected String[] getFlatColumnNames(String[] ret) {
    * Maps are List columns with a Struct named 'key_value' with a child named 'key' and a child
    * named 'value'. The caller of this method doesn't need to worry about this as this method will
    * take care of this without the knowledge of the caller.
+   *
+   * Note: this method always returns a nullabe column, can not return non-nullable column.
```
ttnghia, line 526:
```suggestion
   * Note: This method always returns a nullabe column, cannot return non-nullable column.
```

revans2 (approved):

res-life:
@gpucibot merge

PR #14010, opened 2023-08-30
Expose streams in public replace APIs

vyasr:
## Description

Contributes to #925

## Checklist


cpp/tests/replace/replace_tests.cpp
```diff
@@ -47,8 +47,11 @@ TEST_F(ReplaceErrorTest, SizeMismatch)
   cudf::test::fixed_width_column_wrapper<int32_t> values_to_replace_column{{10, 11, 12, 13}};
   cudf::test::fixed_width_column_wrapper<int32_t> replacement_values_column{{15, 16, 17}};
 
-  EXPECT_THROW(cudf::find_and_replace_all(
-                 input_column, values_to_replace_column, replacement_values_column, mr()),
+  EXPECT_THROW(cudf::find_and_replace_all(input_column,
+                                          values_to_replace_column,
+                                          replacement_values_column,
+                                          cudf::test::get_default_stream(),
+                                          mr()),
```
davidwendt, line 54: Why was mr being passed here? Maybe this should be removed?

vyasr: Good point, I assumed this was a meaningful mr but it's just pulling [the mr set on the base fixture](https://github.com/rapidsai/cudf/blob/branch-23.10/cpp/include/cudf_test/base_fixture.hpp#L59). Do you know why that mr even exists? It's used pretty sparingly, and given its definition it seems like we could probably just remove it altogether. What do you think?

davidwendt: I think we can remove it. I'm not seeing a purpose for it here.

vyasr: I'll update these files in this PR and make a follow-up to remove from the fixture.

vyasr: Done.

cpp/src/replace/nans.cu
```diff
@@ -202,7 +204,7 @@ std::unique_ptr<column> normalize_nans_and_zeros(column_view const& input,
 
   // from device. unique_ptr which gets automatically cleaned up when we leave.
   auto out_view = out->mutable_view();
-  normalize_nans_and_zeros(out_view, stream);
+  detail::normalize_nans_and_zeros(out_view, stream);
```
mythrocks, line 207: I'm a little confused with this part. The enclosing function is in `cudf::detail`. Is `detail::` necessary here? 
The intended target is the preceding function in the same namespace, no?

bdice: It makes that namespace lookup a bit more explicit. I think this is slightly better than before.

bdice (approved):
Good reviews @davidwendt and @mythrocks -- I think this is ready.

mythrocks (approved):

cpp/tests/streams/replace_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+  auto high = cudf::numeric_scalar<int>(7, true, cudf::test::get_default_stream());
+  cudf::clamp(input, low, high, cudf::test::get_default_stream());
+}
+
+TEST_F(ReplaceTest, NormalizeNansAndZeros)
+{
+  auto nan          = std::numeric_limits<double>::quiet_NaN();
+  auto input_column = cudf::test::make_type_param_vector<double>({-0.0, 0.0, -nan, nan, nan});
+  cudf::test::fixed_width_column_wrapper<double> input(input_column.begin(), input_column.end());
+  cudf::normalize_nans_and_zeros(static_cast<cudf::column_view>(input),
```
davidwendt, line 98:
```suggestion
  cudf::normalize_nans_and_zeros(input,
```
Just a minor nit. I don't think this cast is needed.

davidwendt (approved):

vyasr:
Can't find the comment anymore, but for the record the reason the tests include explicit casts from `column_wrapper` to `cudf::column_view ` and `cudf::mutable_column_view` when calling `normalize_nans_and_zeros` is that both casts operators are defined and the function also has overloads for both, so the calls are ambiguous.

PR #14018, opened 2023-08-31
Explicitly depend on zlib in conda recipes

### LINKED ISSUE #14021: [BUG] libcudf CI builds failing due to lack of `zlib` package
**Describe the bug**

[Recently](https://github.com/conda-forge/cmake-feedstock/pull/191/files), the conda feedstock for `cmake` updated its run exports to remove dependency on (amongst other things) the `zlib` conda package.

It seems like we were relying on this transitively being pulled in for our builds. See for example a recent failure: https://github.com/rapidsai/cudf/actions/runs/6033351272/job/16385609807#step:7:635

We should therefore probably make this dependency explicit in `dependencies.yaml` and the conda recipe `meta.yaml` for libcudf.

wence-:
## Description

We were previously obtaining zlib transitively through our cmake dependency, but since the 3.27.4 conda package, this dependency no longer exists. Therefore we must depend on zlib ourselves.

- Closes #14021 

## Checklist

### DIFF
--- a/conda/recipes/libcudf/meta.yaml
+++ b/conda/recipes/libcudf/meta.yaml
@@ -73,7 +73,7 @@ requirements:
     - benchmark {{ gbench_version }}
     - gtest {{ gtest_version }}
     - gmock {{ gtest_version }}
-    - zlib
+    - zlib {{ zlib_version }}
 
 outputs:
   - name: libcudf
--- a/conda/recipes/libcudf/meta.yaml
+++ b/conda/recipes/libcudf/meta.yaml
@@ -45,6 +45,7 @@ requirements:
     - cuda-version ={{ cuda_version }}
     - ninja
     - sysroot_{{ target_platform }} {{ sysroot_version }}
+    - zlib
   host:
     - librmm ={{ minor_version }}
     - libkvikio ={{ minor_version }}
@@ -90,6 +91,7 @@ outputs:
     requirements:
       build:
         - cmake {{ cmake_version }}
+        - zlib
       run:
         {% if cuda_major == "11" %}
         - cudatoolkit
--- a/conda/recipes/libcudf/meta.yaml
+++ b/conda/recipes/libcudf/meta.yaml
@@ -45,7 +45,6 @@ requirements:
     - cuda-version ={{ cuda_version }}
     - ninja
     - sysroot_{{ target_platform }} {{ sysroot_version }}
-    - zlib
   host:
     - librmm ={{ minor_version }}
     - libkvikio ={{ minor_version }}
@@ -74,6 +73,7 @@ requirements:
     - benchmark {{ gbench_version }}
     - gtest {{ gtest_version }}
     - gmock {{ gtest_version }}
+    - zlib
 
 outputs:
   - name: libcudf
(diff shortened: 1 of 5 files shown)


dependencies.yaml
```diff
@@ -179,6 +179,7 @@ dependencies:
           - c-compiler
           - cxx-compiler
           - dlpack>=0.5,<0.6.0a0
+          - zlib
```
bdice, line 182: @wence- Can you comment on where/how we use zlib? Does it need to be in the conda package dependencies (meta.yml for libcudf, or cudf, …) as well?

bdice: Is there an issue documenting the problems you see in builds?

conda/recipes/libcudf/meta.yaml
```diff
@@ -90,6 +91,7 @@ outputs:
     requirements:
       build:
         - cmake {{ cmake_version }}
+        - zlib
```
ajschmidt8, line 94:
```suggestion
```

`zlib` can be removed from here. because of the way our C++ conda packages are structured, the build dependencies are only needed in the top-level requirements section.

`cmake` is typically the only dependency needed in the sub-packages of the recipe.

conda/recipes/libcudf/meta.yaml
```diff
@@ -45,6 +45,7 @@ requirements:
     - cuda-version ={{ cuda_version }}
     - ninja
     - sysroot_{{ target_platform }} {{ sysroot_version }}
+    - zlib
```
ajschmidt8, line 48: I believe this should be moved to the `host` section.

The conda docs below describe the difference between the `host` and `build` dependency sections:

- https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#requirements-section

Even though we don't do any cross-compliation as mentioned in those docs, we've generally followed this convention for most of the RAPIDS projects.

dependencies.yaml
vyasr: We support zlib compression/decompression in ORC reading/writing, so we should be including it in our recipe I think.

bdice: This is what I was missing: https://github.com/rapidsai/cudf/issues/14021#issuecomment-1701327693

vyasr (changes requested):
@vuule do you know if there's a specific version of zlib that we require for our use cases? Wondering if we need a lower bound here.

vuule:
> @vuule do you know if there's a specific version of zlib that we require for our use cases? Wondering if we need a lower bound here.

I'm pretty sure we go back quite a bit and have the same API available. I'm looking at their changelog, and there are bug fixes in every release. Based on this, should we make 1.2.13 the lower bound? It was released Oct 2022.

wence-:
> > @vuule do you know if there's a specific version of zlib that we require for our use cases? Wondering if we need a lower bound here.
> 
> I'm pretty sure we go back quite a bit and have the same API available. I'm looking at their changelog, and there are bug fixes in every release. Based on this, should we make 1.2.13 the lower bound? It was released Oct 2022.

Done.

vyasr (approved):
Lower bound looks reasonable to me.

Issue #13733, opened 2023-07-22
[FEA] Increase maximum characters in strings columns


GregoryKimball:
In libcudf, strings columns have child columns containing character data and offsets, and the offsets child column uses a 32-bit signed size type. This limits strings columns to containing ~2.1 billion characters. For LLM training, documents have up to 1M characters, and a median around 3K characters. Due to the size type limit, LLM training pipelines have to carefully batch the data down to a few thousand rows to stay comfortably within the size type limit. We have a general issue open to explore a 64-bit size type in libcudf (#13159). For size issues with LLM training pipelines, we should consider a targeted change to only address the size limit for strings columns.

### Requirements
* We must maintain or improve throughput for functions processing strings columns with <2.1 billion characters. This requirement prevents us from using 64-bit offsets for all strings columns. It does not prevent us from using 64-bit offsets for strings columns with >2.1 billion characters.
* We must not introduce a new data type or otherwise increase compile times significantly. This requirement prevents us from dispatching between "strings" types and "large strings" types. 

### Proposed solution
One idea that satisfies these requirements would be to represent the character data as an `int64` typed column instead of an `int8` typed column. This would allow us to store 8x more bytes of character data. To access the character bytes, we would use an offset-normalizing iterator (inspired by ["indexalator"](https://github.com/rapidsai/cudf/blob/branch-23.08/cpp/include/cudf/detail/indexalator.cuh)) to identify byte positions using an `int64` iterator output. Please note that the row count 32-bit size type would still apply to the proposed "large strings" columns.

We should also consider an "unbounded" character data allocation that is not typed, but rather a single buffer up to 2^64 bytes in size. The 64-bit offset type would be able to index into much larger allocations.

Please note that this solution will not impact the offsets for list columns. We believe that the best design to allow for more than 2.1B elements in lists will be to use 64-bit size type in libcudf as discussed in #13159.

### Creating strings columns
[...]

PR #14030, opened 2023-09-01
Use cudf::make_empty_column instead of column_view constructor

davidwendt:
## Description
Replaces places where the `cudf::column_view(type,size,...)` constructor was used to create an empty view with a call to `cudf::make_column_view(type)->view()`.

This helps minimize the dependency on calling the constructors directly as part of the work needed for #13733 which may require an update to the `column_view` classes and its constructor(s).

Most of the changes occur in strings gtests source files.
No functionality or behavior has changed.

## Checklist

### DIFF
--- a/cpp/tests/strings/array_tests.cpp
+++ b/cpp/tests/strings/array_tests.cpp
@@ -47,8 +47,8 @@ TEST_F(StringsColumnTest, Sort)
 
 TEST_F(StringsColumnTest, SortZeroSizeStringsColumn)
 {
-  cudf::column_view zero_size_strings_column(
-    cudf::data_type{cudf::type_id::STRING}, 0, nullptr, nullptr, 0);
+  auto const zero_size_strings_column = cudf::make_empty_column(cudf::type_id::STRING)->view();
+
   auto results = cudf::sort(cudf::table_view({zero_size_strings_column}));
   cudf::test::expect_column_empty(results->view().column(0));
 }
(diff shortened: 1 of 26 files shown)


cpp/tests/strings/array_tests.cpp
```diff
@@ -47,8 +47,8 @@ TEST_F(StringsColumnTest, Sort)
 
 TEST_F(StringsColumnTest, SortZeroSizeStringsColumn)
 {
-  cudf::column_view zero_size_strings_column(
-    cudf::data_type{cudf::type_id::STRING}, 0, nullptr, nullptr, 0);
+  auto const zero_size_strings_column = cudf::make_empty_column(cudf::type_id::STRING)->view();
```
vuule, line 50: is this safe? we're keeping a view of a column that's a temporary object. The pattern in `[gather_str_tests.cpp](https://github.com/rapidsai/cudf/pull/14030/files#diff-8c7735ba97a898ebe6d635f6606af58136f7e266a4aad7594b29c4ae18b055c7)` looks more intuitive to me.

davidwendt: This is indeed safe. The return `column_view` contains no data. It has member data like type, etc. But the size and other values are set to null/zero. So there is no dependency on the column it came from.

vuule (approved):

hyperbolic2346 (approved):

PR #14033, opened 2023-09-04
Implement `sort_remaining` for `sort_index`

### LINKED ISSUE #14011: [FEA] Allow DataFrame.sort_index(level=n) to tiebreak on lexsortedness in mode.pandas_compatable
```python
In [1]: import cudf

In [2]:         df = cudf.DataFrame(
   ...:             {
   ...:                 "col1": ["b", "d", "b", "a"],
   ...:                 "col2": [3, 1, 1, 2],
   ...:                 "data": ["one", "two", "three", "four"],
   ...:             }
   ...:         )
   ...: 
   ...:         df2 = df.set_index(["col1", "col2"])

In [3]: df2
Out[3]: 
            data
col1 col2       
b    3       one
d    1       two
b    1     three
a    2      four

In [4]: df2.sort_index(level=0, key=None)
Out[4]: 
            data
col1 col2       
a    2      four
b    3       one
     1     three
d    1       two

In [7]: df2.to_pandas().sort_index(level=0, key=None)
Out[7]: 
            data
col1 col2       
a    2      four
b    1     three
     3       one
d    1       two
```

The current behavior is still technically correct, but it appears that pandas tiebreaks on lexsortedness. It would be nice to mirror this behavior in pandas compatible mode

Manually lexsort after `sort_index`

**Additional context**
Add any other context, code examples, or references to existing implementations about the feature request here.

wence-:
## Description

Previously, the `sort_remaining` argument to `sort_index` was ignored. Passing `sort_remaining=False` would raise a `NotImplementedError`. Moreover, for a multiindex, `sort_remaining=True` was not handled correctly: if not all levels were requested as sorted, `sort_index` would behave as if `sort_remaining=False` had been passed.

To fix this case, construct the sort order based on first the provided levels and, if `sort_remaining=True`, the left-over levels (in index order).

To facilitate this, refactor the internal `_get_columns_by_label` function to always return a `Frame`-like object (previously, if we had a `Frame` we would get back a `ColumnAccessor`, and it was only for `IndexedFrame` and above that we'd get something of `Self`-like type back). This meant that calling `_get_sorted_inds` with `by != None` was not possible on an `Index` or `MultiIndex` (the code assumed we'd get a `Frame` back).

- Closes #14011

## Checklist

### DIFF
--- a/python/cudf/cudf/tests/test_multiindex.py
+++ b/python/cudf/cudf/tests/test_multiindex.py
@@ -1897,3 +1897,26 @@ def test_multiindex_empty_slice_pandas_compatibility():
     with cudf.option_context("mode.pandas_compatible", True):
         actual = cudf.from_pandas(expected)
     assert_eq(expected, actual, exact=False)
+
+
+@pytest.mark.parametrize(
+    "levels",
+    itertools.chain.from_iterable(
+        itertools.permutations(range(3), n) for n in range(1, 4)
+    ),
+    ids=str,
+)
+def test_multiindex_sort_index_partial(levels):
+    df = pd.DataFrame(
+        {
+            "a": [3, 3, 3, 1, 1, 1, 2, 2],
+            "b": [4, 2, 7, -1, 11, -2, 7, 7],
+            "c": [4, 4, 2, 3, 3, 3, 1, 1],
+            "val": [1, 2, 3, 4, 5, 6, 7, 8],
+        }
+    ).set_index(["a", "b", "c"])
+    cdf = cudf.from_pandas(df)
+
+    expect = df.sort_index(level=levels, sort_remaining=True)
+    got = cdf.sort_index(level=levels, sort_remaining=True)
+    assert_eq(expect, got)
--- a/python/cudf/cudf/core/dataframe.py
+++ b/python/cudf/cudf/core/dataframe.py
@@ -36,7 +36,7 @@ from pandas._config import get_option
 from pandas.core.dtypes.common import is_float, is_integer
 from pandas.io.formats import console
 from pandas.io.formats.printing import pprint_thing
-from typing_extensions import assert_never
+from typing_extensions import Self, assert_never
 
 import cudf
 import cudf.core.common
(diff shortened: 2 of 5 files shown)


python/cudf/cudf/tests/test_multiindex.py
```diff
@@ -1897,3 +1897,26 @@ def test_multiindex_empty_slice_pandas_compatibility():
...
+            "b": [4, 2, 7, -1, 11, -2, 7, 7],
+            "c": [4, 4, 2, 3, 3, 3, 1, 1],
+            "val": [1, 2, 3, 4, 5, 6, 7, 8],
+        }
+    ).set_index(["a", "b", "c"])
+    cdf = cudf.from_pandas(df)
+
+    expect = df.sort_index(level=levels, sort_remaining=True)
+    got = cdf.sort_index(level=levels, sort_remaining=True)
+    assert_eq(expect, got)
```
galipremsagar, line 1922:
```suggestion
)
@pytest.mark.parametrize("sort_remaining", [True, False])
def test_multiindex_sort_index_partial(levels):
    df = pd.DataFrame(
        {
            "a": [3, 3, 3, 1, 1, 1, 2, 2],
            "b": [4, 2, 7, -1, 11, -2, 7, 7],
            "c": [4, 4, 2, 3, 3, 3, 1, 1],
            "val": [1, 2, 3, 4, 5, 6, 7, 8],
        }
    ).set_index(["a", "b", "c"])
    cdf = cudf.from_pandas(df)

    expect = df.sort_index(level=levels, sort_remaining=sort_remaining)
    got = cdf.sort_index(level=levels, sort_remaining=sort_remaining)
    assert_eq(expect, got)
```

Should we just parametrize this test by testing both `sort_remaining=True & False`?

galipremsagar (approved):
LGTM, minor comment.

python/cudf/cudf/tests/test_multiindex.py
wence-: I am not sure if there are any guarantees on matching in that case. (I think whether the sort is stable is an implementation detail).

galipremsagar: Ah, alright. We can ignore this suggestion then. Thanks !

galipremsagar (approved):

mroeschke (approved):

vyasr (approved):

PR #14037, opened 2023-09-06
Raise NotImplementedError for to_datetime with z format

mroeschke:
## Description
Avoids timezone information from being dropped in `to_datetime` when the z directive is provided

```python
In [1]: import cudf

In [2]: fmt = '%Y-%m-%d %H:%M:%S %Z'
   ...: dates = ['2010-01-01 12:00:00 UTC', '2010-01-01 12:00:00 UTC']

In [3]: cudf.to_datetime(dates, format=fmt)
Out[3]: DatetimeIndex(['2010-01-01 12:00:00', '2010-01-01 12:00:00'], dtype='datetime64[ns]')
```

## Checklist

### DIFF
--- a/python/cudf/cudf/core/tools/datetimes.py
+++ b/python/cudf/cudf/core/tools/datetimes.py
@@ -147,8 +147,13 @@ def to_datetime(
     if utc:
         raise NotImplementedError("utc is not yet implemented")
 
-    if format is not None and "%f" in format:
-        format = format.replace("%f", "%9f")
+    if format is not None:
+        if "%Z" in format or "%z" in format:
+            raise NotImplementedError(
+                "cuDF does not yet support timezone-aware datetimes"
+            )
+        elif "%f" in format:
+            format = format.replace("%f", "%9f")
 
     try:
         if isinstance(arg, cudf.DataFrame):
--- a/python/cudf/cudf/tests/test_datetime.py
+++ b/python/cudf/cudf/tests/test_datetime.py
@@ -2154,5 +2154,5 @@ def test_daterange_pandas_compatibility():
 def test_format_timezone_not_implemented(code):
     with pytest.raises(NotImplementedError):
         cudf.to_datetime(
-            ["2020-01-01 00:00:00 UTC"], format=f"%Y-%m-%d %H-%M-%S %{code}"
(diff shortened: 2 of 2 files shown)


python/cudf/cudf/tests/test_datetime.py
```diff
@@ -2148,3 +2148,11 @@ def test_daterange_pandas_compatibility():
             "2010-01-01", "2010-02-01", periods=10, name="times"
         )
     assert_eq(expected, actual)
+
+
+@pytest.mark.parametrize("code", ["z", "Z"])
+def test_format_timezone_not_implemented(code):
+    with pytest.raises(NotImplementedError):
+        cudf.to_datetime(
+            ["2020-01-01 00:00:00 UTC"], format=f"%Y-%m-%d %H-%M-%S %{code}"
```
wence-, line 2157:
```suggestion
            ["2020-01-01 00:00:00 UTC"], format=f"%Y-%m-%d %H:%M:%S %{code}"
```

Format string doesn't match the provided date string without it (and once timezones are implemented, we would expect this test to `raise ValueError` because of it).

python/cudf/cudf/core/tools/datetimes.py
```diff
@@ -147,8 +147,13 @@ def to_datetime(
...
 
-    if format is not None and "%f" in format:
-        format = format.replace("%f", "%9f")
+    if format is not None:
+        if "%Z" in format or "%z" in format:
+            raise NotImplementedError(
+                "cuDF does not yet support timezone-aware datetimes"
+            )
+        elif "%f" in format:
+            format = format.replace("%f", "%9f")
```
wence-, line 156: Note: This is fine, but I'll just note that `%%Z` is a valid string for `strptime` that means "look for a literal `%Z`". So this replacement (and the existing `%f` replacement) is not robust against some of the wackier formats a user might provide.

wence- (changes requested):

python/cudf/cudf/tests/test_datetime.py
mroeschke: Thanks for the catch

wence- (approved):
Thanks @mroeschke

PR #14043, opened 2023-09-06
Refactor libcudf indexalator to typed normalator

davidwendt:
## Description
Creates generic normalizing-iterator for integer types for use by the indexalator and the future offsets normalizing iterator.
Mostly code has been moved around or renamed so the normalizing-iterator part can take type template parameter to identify which integer type to normalize to. For the indexalator this type is `cudf::size_type` for the offsets iterator this type would be `int64_t`.

## Checklist

### DIFF
--- a/cpp/include/cudf/detail/indexalator.cuh
+++ b/cpp/include/cudf/detail/indexalator.cuh
@@ -16,14 +16,13 @@
 
 #pragma once
 
+#include "normalizing_iterator.cuh"
+
 #include <cudf/column/column_view.hpp>
 #include <cudf/detail/iterator.cuh>
 #include <cudf/scalar/scalar.hpp>
 #include <cudf/utilities/traits.hpp>
 
-#include <thrust/binary_search.h>
-#include <thrust/execution_policy.h>
-#include <thrust/functional.h>
 #include <thrust/iterator/constant_iterator.h>
 #include <thrust/iterator/transform_iterator.h>
 #include <thrust/optional.h>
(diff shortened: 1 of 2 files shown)


cpp/include/cudf/detail/indexalator.cuh
```diff
@@ -32,193 +31,6 @@
 namespace cudf {
 namespace detail {
 
-/**
```
davidwendt, line 35: This code moved/refactored to `normalizing_iterator.cuh`

cpp/include/cudf/detail/normalizing_iterator.cuh
```diff
@@ -0,0 +1,362 @@
+/*
```
davidwendt, line 1: This code moved/refactored from `indexalator.cuh`

cpp/include/cudf/detail/normalizing_iterator.cuh
```diff
@@ -0,0 +1,362 @@
...
+   * to capture the `operator=(size_type)` calls.
+   */
+  __device__ inline output_normalator const& operator*() const { return *this; }
+
+  /**
+   * @brief Array subscript operator returns an iterator instance at the specified `idx` position.
+   *
+   * This allows capturing the subsequent `operator=(size_type)` call in this class.
+   */
+  __device__ inline output_normalator const operator[](size_type idx) const
```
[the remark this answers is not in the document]
davidwendt: It's a good question but I don't remember all the details of how I figured this out.
If it helps, this is not new code but is just moved from the other header.

bdice (approved):
Thanks for helping me think through my previous review comments. This looks good to me.

cpp/include/cudf/detail/normalizing_iterator.cuh
```diff
@@ -0,0 +1,367 @@
...
+  CUDF_HOST_DEVICE inline bool operator>=(Derived const& rhs) const
+  {
+    return static_cast<Derived const&>(*this).p_ >= rhs.p_;
+  }
+
+ protected:
+  /**
+   * @brief Constructor assigns width and type member variables for base class.
+   */
+  base_normalator(data_type dtype) : width_(size_of(dtype)), dtype_(dtype) {}
```
mythrocks, line 209: Silly question: Should this single argument constructor be `explicit`?

cpp/include/cudf/detail/normalizing_iterator.cuh
```diff
@@ -0,0 +1,367 @@
...
+
+  output_normalator()                                    = default;
+  output_normalator(output_normalator const&)            = default;
+  output_normalator(output_normalator&&)                 = default;
+  output_normalator& operator=(output_normalator const&) = default;
+  output_normalator& operator=(output_normalator&&)      = default;
+
+  /**
+   * @brief Indirection operator returns this iterator instance in order
+   * to capture the `operator=(size_type)` calls.
```
mythrocks, line 310: I'm not sure I've understood the brief.  A user might attempt to assign a `size_type` to an `output_normalator`? Or to `*my_output_normalator`?

cpp/include/cudf/detail/normalizing_iterator.cuh
```diff
@@ -0,0 +1,367 @@
...
+   */
+  __device__ inline output_normalator const& operator*() const { return *this; }
```
mythrocks, line 312: I'm fairly certain I haven't understood this part. By returning `*this`, then for any `output_normalator` `p`,
`p` == `*p` == `**p`?

cpp/include/cudf/detail/normalizing_iterator.cuh
davidwendt: The comment needs to be updated. The `size_type` in the comment should be `Integer` now.
This function basically redirects any `*itr = value` call to use the `operator=(Integer)` function below.

cpp/include/cudf/detail/normalizing_iterator.cuh
mythrocks: I would have thought we'd have to support semantics like:
```c++
output_normalator p = ...;
*p = int16_t{5};
*(p+1) = int32_t{10};
p[2] = int64{15};
```
I haven't understood how current implementation for `operator*()` and `operator[]()` work.

davidwendt: We cannot return a reference to the underlying array element which is what you would normally see.
From what I remember, this is a way to capture that request `*itr = value` and redirect it to the `operator=(Integer)` function below. It's a subtle thing for sure.

davidwendt: The output iterator expects a specific type but can store it as the underlying type.
The `operator[]()` works normally in that it handles `p[2] = value`.
I've tried to explain the `operator*()` and `operator=` relationship in a few separate comments now.

mythrocks: Oh, maybe I see it now?

`*p = 5;`  is evaluated as:
1. `output_normalator::operator *()`, which returns `p`.
2. `p = 5` calls `output_normalator::operator =(Integer)`.

Do I have this right?

Were I to nitpick, I would have preferred it if `*p` returned an `integer_reference` object, and `integer_reference::operator=()` did the assignment.  As it stands, doesn't the user run the risk of assigning integers to the normalator directly, instead of to `*normalator`?

That said, yours has the benefit of brevity.

bdice: That reference idea sounds brilliant to me! I think that’s much clearer in intent.

davidwendt: I think there is more to it than what you have described.

*p = 5; is evaluated as:

1. `output_normalator::operator*()`, which returns _reference_ to `p`.
2. `&p = 5` calls `output_normalator::operator =(Integer)`.

I'm not seeing what the `integer_reference` object would be defined to be.

bdice: Yes, let's do this in a later PR. I'd propose the following. If you agree, opening an issue for this would be good.

- `output_normalator::reference` is a member type defined as a class `integer_reference`
- `integer_reference` requirements:
  - holds some state to know what it's referencing (perhaps the output normalator and the index which it references?)
  - defines `integer_reference::operator=(Integer)` which normalizes and writes to the referenced location
- `output_normalator::operator*` is changed to return an `integer_reference`
- `output_normalator::operator=(Integer)` is no longer defined -- because assignment of a value to a (non-dereferenced) iterator is an unclear design that we can avoid by using `integer_reference`

mythrocks (approved):
Barring a minor suggestion regarding how to deal with `output_normalator` assignments, this looks good to ship. Thanks, David!

cpp/include/cudf/detail/normalizing_iterator.cuh
mythrocks: > let's do this in a later PR

Yes, sir. That works.

mythrocks:
Sorry if I muddied the waters here with my last comment.  Let's not hold this PR up for the `integer_reference` suggestion. We can tackle that later.

PR #14055, opened 2023-09-07
Adds pylibcudf Scalar

vyasr:
## Checklist


python/cudf/cudf/_lib/pylibcudf/scalar.pyx
```diff
@@ -0,0 +1,99 @@
...
+        self.mr = get_current_device_resource()
+
+        if value is None:
+            # TODO: This early return is not something we really want to
+            # support, but it here for now to ease the transition of
+            # DeviceScalar.
+            return
+
+        # Convert the value to a cudf object via pyarrow
+        arr = pyarrow.lib.array([value.as_py()], type=value.type)
```
wence-, line 47: More effort, but I think we can use the C++-level [`ArrayBuilder`](https://arrow.apache.org/docs/cpp/api/builder.html#_CPPv4N5arrow12ArrayBuilder12AppendScalarERK6Scalar) interface to get an arrow array from which we can make a table to pass to libcudf. If we push it all onto C++ then we can probably avoid some back and forth too

wence-: This (untested) might be a useful starting point:

```c++
std::unique_ptr<scalar> from_arrow(arrow::Scalar const &scalar, rmm::mr::device_memory_resource* mr) {
  auto b = arrow::MakeBuilder(scalar.type);
  if (b.ok()) {
    auto builder = std::move(b.ValueOrDie());
    arrow::Status status = builder->AppendScalar(scalar);
    if (status != arrow::Status::OK()) {
      CUDF_FAIL("Arrow ArrayBuilder::AppendScalar failed");
    }
    auto array_result = builder->Finish();
    if (array_result.ok()) {
      auto array = array_result.ValueOrDie();
      auto type = detail::arrow_to_cudf_type(*array->type().get());
      auto col = detail::get_column(*array.get(), type, true, cudf::get_default_stream(), mr);
      auto cv = col->view();
      return detail::get_element(cv, 0, cudf::get_default_stream(), mr);
    } else {
      CUDF_FAIL("Arrow ArrayBuilder::Finish failed");
    }
  } else {
    CUDF_FAIL("Arrow MakeBuilder failed");
  }
}
```

vyasr: I was hoping to avoid the C++ API, but at this point I think you're right that the most straightforward solution is to implement at this level. I'll look into it.

vyasr:
The core of this PR is in #14121, #14124, and #14133. The only remaining bit that's only on this branch is the beginnings of a scatter implementation, which didn't actually get very far. I'll handle that properly in another PR once #14133 is merged, but at this point this PR can be closed since it was primarily to help demonstrate the approach I was taking before I had a chance to really clean up the code.

PR #14057, opened 2023-09-07
Remove the mr from the base fixture

vyasr:
## Description

This mr is just an alias for the current memory resource, so we don't really need it. This came up in https://github.com/rapidsai/cudf/pull/14010#discussion_r1312405952. This PR removes all uses of it, but does not actually remove the mr yet. That will be done in a follow-up (see https://github.com/rapidsai/cudf/pull/14057#issuecomment-1712340714).

## Checklist


davidwendt (approved):

vyasr:
There's something very strange happening here. I can remove all uses of the mr from the fixture, but removing the mr itself is causing a single test failure in a test that isn't otherwise modified in this PR and does not use the mr function. Furthermore, the test only fails on arm, and only for certain types. In the interest of debugging a minimal example, I'm going to revert just that part of this PR, merge it, and then debug the issue separately (will probably need to get access to an arm machine).

Issue #14042, opened 2023-09-06
[BUG] Should cudf.to_datetime(dayfirst=True) raise NotImplementedError?


mroeschke:
**Describe the bug**
Should cudf.to_datetime(dayfirst=True) raise NotImplementedError?

**Steps/Code to reproduce bug**
```python
In [6]: cudf.to_datetime(["10-02-2014"], dayfirst=True)
Out[6]: DatetimeIndex(['2014-10-02'], dtype='datetime64[ns]')
```

**Expected behavior**
```python
In [6]: cudf.to_datetime(["10-02-2014"], dayfirst=True)
Out[6]: DatetimeIndex(['2014-02-10'], dtype='datetime64[ns]')
```

In `DatetimeIndex.__init__`, this specifically raises a NotImplementedError

```python
        if dayfirst is not False:
            raise NotImplementedError("dayfirst == True is not yet supported")
```

**Environment overview (please complete the following information)**
 - Environment location: Bare-metal
 - Method of cuDF install: conda
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**
Please run and paste the output of the `cudf/print_env.sh` script here, to gather any other relevant environment details

**Additional context**

PR #14058, opened 2023-09-07
Raise NotImplementedError in to_datetime with dayfirst without infer_format

mroeschke:
## Description
Raises a `NotImplementedError` to avoid this incorrect behavior (which seems to actually not be implemented)

```python
In [6]: cudf.to_datetime(["10-02-2014"], dayfirst=True)
Out[6]: DatetimeIndex(['2014-10-02'], dtype='datetime64[ns]')
```

closes https://github.com/rapidsai/cudf/issues/14042

## Checklist


vyasr (commented):
Looks like the modified test is failing to catch the expected error, perhaps the condition isn't quite right?

vyasr (approved):

Issue #13744, opened 2023-07-24
[FEA] Expose public stream-ordered C++ APIs


vyasr:
**Is your feature request related to a problem? Please describe.**
Currently most libcudf APIs operate implicitly on the default CUDA stream. As documented in #925, we would like to expose stream-ordered APIs publicly. Prior to doing so, we needed to do a great deal of work to ensure that libcudf's internals were properly stream-ordered. Among the important tasks here were removing implicit APIs that ran on a stream (see #11968) and implementing a strategy for verifying that streams are properly forwarded from APIs all the way down to the lowest level functions (see #11943). Now that all such work that we are aware of has been completed, libcudf should start exposing streams in public APIs. 

We originally considered adding stream support via a monolithic feature branch where APIs would be modified one at a time. From various discussions, however, we decided that the overhead of maintaining a feature branch would grow too large to be worthwhile, so we have instead decided to proceed by incrementally exposing streams in libcudf APIs over time. PRs adding new APIs should always include a stream in the public API. Existing APIs will be transitioned one header file at a time, with PRs going directly into the current main branch of cudf.

The purpose of this PR is to document progress on the task of transitioning existing headers, as well as to describe the process by which new stream-ordered APIs should be added.

**Describe the solution you'd like**
When adding stream support to an API, the stream parameter should be placed just before the mr parameter with a default value of `cudf::get_default_stream()`. A new test should be added that validates that the API properly forwards the stream to all CUDA calls, which can be done by following the instructions at the bottom of [libcudf's testing developer docs](https://docs.rapids.ai/api/libcudf/nightly/md_developer_guide_testing). 

To divide up the work, we consider public headers organized by directories (all relative to `cpp/include/cudf`). Devs may assign themselves a directory within which to own all headers and the public APIs contained therein.

### /cudf APIs

- [x] aggregation.hpp
- [x] ast/ #16452
- [x] binaryop.hpp #14187
- [x] column/
    - [x] column_device_view.hpp
    - [x] column_factories.hpp #14354
    - [x] column_view.hpp #14354
    - [x] column.hpp #14354
- [x] concatenate.hpp #13987 
- [x] contiguous_split.hpp
- [x] copying.hpp #13629
- [x] datetime.hpp #16775
- [x] dictionary/
    - [x] dictionary_column_view.hpp
    - [x] dictionary_factories.hpp #14115
    - [x] encode.hpp #14115
    - [x] search.hpp #14115
    - [x] update_keys.h #14115
[...]

PR #14060, opened 2023-09-07
Expose stream parameter in public strings find APIs

davidwendt:
## Description
Add stream parameter to public APIs:

- `cudf::strings::find()`
- `cudf::strings::rfind()`
- `cudf::strings::contains()`
- `cudf::strings::starts_with()`
- `cudf::strings::ends_with()`
- `cudf::strings::findall()`
- `cudf::strings::find_multiple()`

Also cleaned up some of the doxygen comments. 

Reference #13744 

## Checklist


vyasr (changes requested):
Could we add tests for the new stream-based APIs (see any of my recent PRs for what those should look like)? They're mostly just to be safe, but on occasion they have revealed places where we're internally not forwarding input streams along and are just passing `cudf::get_default_stream()` instead.

vyasr (approved):

cpp/include/cudf/strings/find.hpp
```diff
@@ -166,14 +174,16 @@ std::unique_ptr<column> contains(
  *
  * Any null string entries return corresponding null entries in the output columns.
  *
- * @param strings Strings instance for this operation.
- * @param target UTF-8 encoded string to search for in each string.
+ * @param input Strings instance for this operation
+ * @param target UTF-8 encoded string to search for in each string
+ * @param stream CUDA stream used for device memory operations and kernel launches
  * @param mr Device memory resource used to allocate the returned column's device memory.
```
vuule, line 180: missed a period :D 
```suggestion
 * @param mr Device memory resource used to allocate the returned column's device memory
```

vuule (approved):

PR #14063, opened 2023-09-07
Remove debug print in a Parquet test

vuule:
## Description
Removed a debug print. That's it.

## Checklist


karthikeyann (approved):

cpp/tests/io/parquet_test.cpp
```diff
@@ -52,6 +52,9 @@
 #include <random>
 #include <type_traits>
 
+// Set this to true to enable printing, for debugging.
+auto constexpr print_enabled = false;
```
ttnghia, line 56: Should this be read from env variable?

cpp/tests/io/parquet_test.cpp
```diff
@@ -6534,7 +6537,7 @@ TEST_F(ParquetReaderTest, FilterFloatNAN)
   auto col0 = cudf::test::fixed_width_column_wrapper<float>(elements, elements + num_rows);
   auto col1 = cudf::test::fixed_width_column_wrapper<double>(elements, elements + num_rows);
 
-  cudf::test::print(col0);
+  if constexpr (print_enabled) { cudf::test::print(col0); }
```
davidwendt, line 6540: Is this line even necessary? Seems this was added for debugging something specific and was just forgotten.
It does not seem difficult to re-add the line if debugging is needed in the future.
Also, future debugging would probably want to print a different variable in a different place and enabling the constant could produce confusing output.

karthikeyann:
```suggestion
```

cpp/tests/io/parquet_test.cpp
vuule: I copied the mechanism from structs_tests. Either way, this is removed now :)

davidwendt (approved):

ttnghia (approved):

Issue #12261, opened 2022-11-29
[FEA] Refactor hash-based algorithms with new cuco data structures


PointKernel:
**Is your feature request related to a problem? Please describe.**
`cuco::static_map` and `cuco::static_multimap` are used to perform hash-based operations in libcudf. Depending on https://github.com/NVIDIA/cuCollections/issues/110, all the existing use cases should be replaced `cuco::static_set` or `cuco:static_multiset` since the payload part in the hash map has never been used.

Moreover, some libcudf implementations are still using `concurrent_unordered_map` or `unordered_multiset` and they should be refactored with `cuco::static_set/multiset` as well.

**Describe the solution you'd like**
Replace existing implementations with new data structures like `cuco::static_set`, `cuco::static_multiset` or `cuco::static_map`.

|Algorithm|Desired Data Structure|Related APIs|Notes|
|---|---|---|---|
|join|`cuco::static_multiset`|`insert`, `insert_if`, `size`, `count`, `count_outer`, `retrieve`, `retrieve_outer`| |
|mixed join| TBD | TBD | |
|search/contains_column|`cuco::static_set`|`insert`, `insert_if`, `contains`| not sure why it's using `unordered_multiset`, should be able to refactor with `cuco:static_set`. `unordered_multiset` can be removed once this is done.|
|search/contains_table|`cuco::static_set`|`insert`, `insert_if`, `contains`| |
|hash-based groupby|`cuco::static_set`|`ref::insert_and_find`, `retrieve_all`|using `concurrent_unordered_map`|
|byte_pair_encoding|`cuco::static_set`|`insert`, `size`, `ref::find`| |
|distinct|`cuco::static_set`|`insert`, `retrieve_all`, `ref::find`| |
|distinct_count|`cuco::static_set`|`insert`, `insert_if`, `size`| |
|parquet dictionary encoding|`cuco::static_set`|`ref::insert`, `ref::find`| |
|orc dictionary encoding|`cuco::static_set`|`ref::insert`, `ref::find`| similar to parquet dictionary encoding but the current implementation is still using a custom dictionary |
|json_tree|`cuco::static_set`|`insert_if`, `ref::find`| |
|json reader|`cuco::static_map`|`insert`, `ref::find`| looks like a rational use of hash map. To be replaced by `cuco::static_map` |

bdice:
[...]

PR #14064, opened 2023-09-08
Refactor `contains_table` with cuco::static_set

PointKernel:
## Description
Contributes to #12261

This PR refactors `contains_table` to use the new `cuco::static_set` data structure. It also adds a `contains_table` benchmark to track the performance before and after this work.

## Checklist


cpp/src/search/contains_table.cu
```diff
@@ -133,23 +138,40 @@ std::pair<rmm::device_buffer, bitmask_type const*> build_row_bitmask(table_view
   return std::pair(rmm::device_buffer{0, stream}, nullable_columns.front().null_mask());
 }
 
+template <bool haystack_has_nulls, bool has_any_nulls, typename Func>
+void dispatch(
+  null_equality compare_nulls, auto self_comp, auto two_table_comp, auto nan_comp, Func func)
+{
+  auto const d_self_eq = self_comp.equal_to<haystack_has_nulls>(
```
PointKernel, line 145: Got build issues at this line
```bash
/home/yunsongw/dev/rapids/cudf/cpp/src/search/contains_table.cu (145): error: constant "haystack_has_nulls" is not a type name

/home/yunsongw/dev/rapids/cudf/cpp/src/search/contains_table.cu (145): error: type name is not allowed
```
Not sure where the issue came from.

cpp/src/search/contains_table.cu
```diff
@@ -173,124 +194,98 @@ rmm::device_uvector<bool> contains(table_view const& haystack,
...
+      cuco::experimental::extent{compute_hash_table_size(haystack.num_rows())},
+      cuco::empty_key{lhs_index_type{-1}},
+      d_equal,
+      cuco::experimental::linear_probing<1, hasher_type>{d_hasher},
+      detail::hash_table_allocator_type{default_allocator<lhs_index_type>{}, stream},
+      stream.value()};
+
+    if (haystack_has_nulls && compare_nulls == null_equality::UNEQUAL) {
+      auto const row_bitmask =
+        cudf::detail::bitmask_and(haystack, stream, rmm::mr::get_current_device_resource()).first;
```
ttnghia, line 238: `bitmask_and` only exec `and` of the top level columns in the table. If we have nested columns with nulls nested inside, this `row_bitmask` will be all valid. What we really want is to exclude any row having any nested nulls from being inserted into the set.

ttnghia: Imagine that we have a structs column `struct<int, int>` where the second `int` child is all nulls. In such situation, we will hit the performance issue reported in the issues at lines 203-204 (deleted) above.

ttnghia: There is still one more potential situation that we need to explore/verify: `nulls` at top level are considered unequal but nested nulls are always equal. In such situation, we must update our row comparator to support that before we can use `bitmask_and` here.

PointKernel: > What we really want is to exclude any row having any nested nulls from being inserted into the set.

Is this the desired behavior of spark? If so, we should probably move `build_row_bitmask` to a detail header thus it can be used in other places like join or distinct, etc. Also, would null equality make a difference here? 

For my own education, there are several details I would like to confirm:

- If nulls are unequal, and two tables are not nullable at the top level, Are the below rows considered equal?
```
list{ list{ list{ 1, null } } }  v.s.  list{ list{ list{ 1, null } } }
``` 
- The same condition as above, but the top-level lists are nullable, would this make a difference?
- I vaguely recall for certain stream compaction operations, Spark treats all nulls as equal


> nulls at top level are considered unequal but nested nulls are always equal.

Maybe I'm missing something obvious, do we distinct top-level nulls and nested nulls in libcudf?

ttnghia: >> What we really want is to exclude any row having any nested nulls from being inserted into the set.

> Is this the desired behavior of spark? 
This is not asked by Spark. This is due to the performance issue reported in the deleted lines 203-204.

Indeed, in spark, in some cases, nested nulls are considered **equal** even top-level nulls are **unequal**. However, we currently cannot compare nulls with both equal and unequal in cudf.

In summary, when nulls are unequal, we should avoid inserting rows having any nulls at any nested level, for performance reason.

PointKernel: Done.

cpp/src/search/contains_table.cu
```diff
@@ -134,22 +124,49 @@ std::pair<rmm::device_buffer, bitmask_type const*> build_row_bitmask(table_view
...
 /**
- * @brief Invoke an `operator()` template with a row equality comparator based on the specified
- * `compare_nans` parameter.
+ * @brief Invokes the given `func` with desired comparators based on the specified `compare_nans`
+ parameter.
  *
- * @param compare_nans The flag to specify whether NaNs should be compared equal or not
+ * @tparam HaystackHasNested Flag indicating whether there are nested columns in haystack
+ * @tparam HasAnyNested Flag indicating whether there are nested columns in either haystack or
+ needles
```
ttnghia, line 132:
```suggestion
 * @tparam HasAnyNested Flag indicating whether there are nested columns in either haystack or
          needles
```

cpp/src/search/contains_table.cu
```diff
@@ -173,124 +190,98 @@ rmm::device_uvector<bool> contains(table_view const& haystack,
...
-        } else {
-          auto const d_eqcomp = strong_index_comparator_adapter{comparator.equal_to<false>(
-            nullate::DYNAMIC{haystack_has_nulls}, compare_nulls, value_comp)};
-          map.insert(
-            haystack_it, haystack_it + haystack.num_rows(), d_hasher, d_eqcomp, stream.value());
-        }
-      };
-
-      dispatch_nan_comparator(compare_nans, insert_map);
+      set.insert_if_async(haystack_iter,
```
ttnghia, line 235: Can you add back the deleted comments to clarify this `_if`:
```
   // If the haystack table has nulls but they are compared unequal, don't insert them.
    // Otherwise, it was known to cause performance issue:
    // - https://github.com/rapidsai/cudf/pull/6943
    // - https://github.com/rapidsai/cudf/pull/8277
```

cpp/src/search/contains_table.cu
PointKernel: Not sure it's worth the effort. I assume we shouldn't rely on any manual alignment for code formatting/linting.

ttnghia (approved):

cpp/src/search/contains_table.cu
```diff
@@ -173,124 +190,103 @@ rmm::device_uvector<bool> contains(table_view const& haystack,
...
+  auto const needles_iter = cudf::detail::make_counting_transform_iterator(
+    size_type{0}, [] __device__(auto idx) { return rhs_index_type{idx}; });
+
+  auto const helper_func = [&](auto const& d_self_equal, auto const& d_two_table_equal) {
+    auto const d_equal = comparator_adapter{d_self_equal, d_two_table_equal};
 
-    auto const comparator =
-      cudf::experimental::row::equality::self_comparator(preprocessed_haystack);
+    auto set = cuco::experimental::static_set{
+      cuco::experimental::extent{compute_hash_table_size(haystack.num_rows())},
```
ttnghia, line 225: Side comment: Should `compute_hash_table_size` be added as a public utility API in `cuco`? I see that it is needed everywhere that is constructing a cuco map/set.

PointKernel: Good point! it's something I wanted to add for a while.

PointKernel: see https://github.com/NVIDIA/cuCollections/pull/369

PointKernel:
The peak memory usage drops about 50% after this PR in all test cases.

As for performance, there is a trade-off between nested and flat types. The legacy map uses a hardcoded CG size = 4 and there is no way to change it on the user side. With the new set data structure and CG size = 4, the new code outperforms the legacy implementation for both int column and list column:
|      type       |  null_probability  |  table_size  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |        Diff |   %Diff |  Status  |
|-----------------|--------------------|--------------|------------|-------------|------------|-------------|-------------|---------|----------|
|       I32       |         0          |    10000     |  38.075 us |      14.80% |  26.917 us |      15.44% |  -11.158 us | -29.31% |   FAIL   |
|       I32       |        0.1         |    10000     |  35.490 us |       6.37% |  25.740 us |       7.00% |   -9.750 us | -27.47% |   FAIL   |
|       I32       |         0          |    100000    |  38.219 us |      16.96% |  28.168 us |       6.23% |  -10.051 us | -26.30% |   FAIL   |
|       I32       |        0.1         |    100000    |  40.250 us |      10.13% |  30.142 us |       6.66% |  -10.108 us | -25.11% |   FAIL   |
|       I32       |         0          |   1000000    |  89.140 us |       2.51% |  67.346 us |       2.73% |  -21.794 us | -24.45% |   FAIL   |
|       I32       |        0.1         |   1000000    | 105.566 us |       2.34% |  83.149 us |       4.47% |  -22.417 us | -21.23% |   FAIL   |
|       I32       |         0          |   10000000   | 613.279 us |       0.93% | 478.395 us |       0.83% | -134.883 us | -21.99% |   FAIL   |
|       I32       |        0.1         |   10000000   | 756.407 us |       0.70% | 634.450 us |       0.68% | -121.957 us | -16.12% |   FAIL   |
| cudf::list_view |         0          |    10000     | 500.887 us |       0.51% | 488.441 us |       0.65% |  -12.447 us |  -2.48% |   FAIL   |
| cudf::list_view |        0.1         |    10000     | 624.067 us |       0.44% | 619.780 us |       0.36% |   -4.287 us |  -0.69% |   FAIL   |
| cudf::list_view |         0          |    100000    | 527.369 us |       0.54% | 513.879 us |       0.34% |  -13.490 us |  -2.56% |   FAIL   |
| cudf::list_view |        0.1         |    100000    | 557.109 us |       0.66% | 539.606 us |       0.60% |  -17.503 us |  -3.14% |   FAIL   |
| cudf::list_view |         0          |   1000000    |   3.363 ms |       0.14% |   3.348 ms |       0.06% |  -15.632 us |  -0.46% |   FAIL   |
| cudf::list_view |        0.1         |   1000000    |   4.276 ms |       0.15% |   4.317 ms |       0.07% |   41.135 us |   0.96% |   FAIL   |
| cudf::list_view |         0          |   10000000   |   2.257 ms |       0.20% |   2.249 ms |       0.13% |   -8.445 us |  -0.37% |   FAIL   |
| cudf::list_view |        0.1         |   10000000   |   2.862 ms |       0.14% |   2.886 ms |       0.12% |   24.101 us |   0.84% |   FAIL   |

If we reduce the CG size to 1 or 2, the performance for int column can be further improved to about 3.5x speedups while the performance for list column can be as much as 30% slower than the current implementation:
|      type       |  null_probability  |  table_size  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |        Diff |   %Diff |  Status  |
|-----------------|--------------------|--------------|------------|-------------|------------|-------------|-------------|---------|----------|
|       I32       |         0          |    10000     |  37.817 us |      15.00% |  26.301 us |       9.96% |  -11.516 us | -30.45% |   FAIL   |
|       I32       |        0.1         |    10000     |  35.636 us |       6.63% |  25.913 us |       6.84% |   -9.722 us | -27.28% |   FAIL   |
|       I32       |         0          |    100000    |  36.421 us |       6.37% |  26.731 us |       6.48% |   -9.690 us | -26.61% |   FAIL   |
|       I32       |        0.1         |    100000    |  38.886 us |       8.68% |  28.775 us |       6.04% |  -10.111 us | -26.00% |   FAIL   |
|       I32       |         0          |   1000000    |  89.063 us |       2.69% |  38.619 us |       4.26% |  -50.444 us | -56.64% |   FAIL   |
|       I32       |        0.1         |   1000000    | 105.316 us |       2.32% |  46.594 us |       3.75% |  -58.723 us | -55.76% |   FAIL   |
|       I32       |         0          |   10000000   | 612.065 us |       0.75% | 168.778 us |       1.13% | -443.287 us | -72.42% |   FAIL   |
|       I32       |        0.1         |   10000000   | 757.799 us |       0.77% | 217.097 us |       0.81% | -540.702 us | -71.35% |   FAIL   |
| cudf::list_view |         0          |    10000     | 487.529 us |       0.59% | 488.617 us |       0.56% |    1.087 us |   0.22% |   PASS   |
| cudf::list_view |        0.1         |    10000     | 608.179 us |       0.48% | 621.207 us |       0.32% |   13.028 us |   2.14% |   FAIL   |
| cudf::list_view |         0          |    100000    | 526.683 us |       0.52% | 630.362 us |       0.32% |  103.680 us |  19.69% |   FAIL   |
| cudf::list_view |        0.1         |    100000    | 557.078 us |       0.69% | 745.262 us |       0.50% |  188.184 us |  33.78% |   FAIL   |
| cudf::list_view |         0          |   1000000    |   3.355 ms |       0.08% |   3.605 ms |       0.06% |  250.828 us |   7.48% |   FAIL   |
| cudf::list_view |        0.1         |   1000000    |   4.267 ms |       0.08% |   4.548 ms |       0.33% |  281.113 us |   6.59% |   FAIL   |
| cudf::list_view |         0          |   10000000   |   2.252 ms |       0.17% |   2.426 ms |       0.10% |  174.065 us |   7.73% |   FAIL   |
| cudf::list_view |        0.1         |   10000000   |   2.861 ms |       0.19% |   3.014 ms |       0.10% |  153.529 us |   5.37% |   FAIL   |

ttnghia:
> If we reduce the CG size to 1 or 2, the performance for int column can be further improved to about 3.5x speedups while the performance for list column can be as much as 30% slower than the current implementation:

Can we selectively create a map (using separate code path) with different CG if the input is flat/nested?

cpp/src/search/contains_table.cu
davidwendt: Regardless of alignment, I think all lines should start with a single space and an asterisk.
https://github.com/rapidsai/cudf/blob/branch-23.10/cpp/doxygen/developer_guide/DOCUMENTATION.md#block-comments
I think this should apply to blank lines within the block comment as well.

PointKernel: > all lines should start with a single space and an asterisk.

Oh, good catch. Definitely yes, something went wrong with my IDE formatting. Fixed

PointKernel:
> Can we selectively create a map (using separate code path) with different CG if the input is flat/nested?

That's exactly what we are doing in `contains_column` where it uses cudf multiset for flat types and invokes `contains_table` if it's a nested type. The cudf multiset doesn't have the CG concept and the performance should be similar to the CG size = 1 implementation with new cuco set.

PointKernel:
The latest performance with CG tuning based on haystack data type:
|      type       |  null_probability  |  table_size  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |        Diff |   %Diff |  Status  |
|-----------------|--------------------|--------------|------------|-------------|------------|-------------|-------------|---------|----------|
|       I32       |         0          |    10000     |  38.075 us |      14.80% |  26.776 us |      14.09% |  -11.299 us | -29.68% |   FAIL   |
|       I32       |        0.1         |    10000     |  35.490 us |       6.37% |  25.960 us |       6.55% |   -9.530 us | -26.85% |   FAIL   |
|       I32       |         0          |    100000    |  38.219 us |      16.96% |  26.678 us |       6.56% |  -11.541 us | -30.20% |   FAIL   |
|       I32       |        0.1         |    100000    |  40.250 us |      10.13% |  28.822 us |       6.07% |  -11.427 us | -28.39% |   FAIL   |
|       I32       |         0          |   1000000    |  89.140 us |       2.51% |  38.517 us |       4.52% |  -50.623 us | -56.79% |   FAIL   |
|       I32       |        0.1         |   1000000    | 105.566 us |       2.34% |  46.443 us |       3.72% |  -59.123 us | -56.01% |   FAIL   |
|       I32       |         0          |   10000000   | 613.279 us |       0.93% | 169.362 us |       2.67% | -443.916 us | -72.38% |   FAIL   |
|       I32       |        0.1         |   10000000   | 756.407 us |       0.70% | 215.990 us |       0.82% | -540.417 us | -71.45% |   FAIL   |
| cudf::list_view |         0          |    10000     | 500.887 us |       0.51% | 487.216 us |       0.46% |  -13.671 us |  -2.73% |   FAIL   |
| cudf::list_view |        0.1         |    10000     | 624.067 us |       0.44% | 618.258 us |       0.26% |   -5.809 us |  -0.93% |   FAIL   |
| cudf::list_view |         0          |    100000    | 527.369 us |       0.54% | 498.654 us |       0.44% |  -28.715 us |  -5.44% |   FAIL   |
| cudf::list_view |        0.1         |    100000    | 557.109 us |       0.66% | 539.606 us |       0.59% |  -17.503 us |  -3.14% |   FAIL   |
| cudf::list_view |         0          |   1000000    |   3.363 ms |       0.14% |   3.340 ms |       0.07% |  -23.658 us |  -0.70% |   FAIL   |
| cudf::list_view |        0.1         |   1000000    |   4.276 ms |       0.15% |   4.307 ms |       0.06% |   31.480 us |   0.74% |   FAIL   |
| cudf::list_view |         0          |   10000000   |   2.257 ms |       0.20% |   2.233 ms |       0.09% |  -24.332 us |  -1.08% |   FAIL   |
| cudf::list_view |        0.1         |   10000000   |   2.862 ms |       0.14% |   2.867 ms |       0.08% |    4.870 us |   0.17% |   FAIL   |

ttnghia (approved):

cpp/src/search/contains_table.cu
```diff
@@ -134,22 +124,59 @@ std::pair<rmm::device_buffer, bitmask_type const*> build_row_bitmask(table_view
 }
 
 /**
- * @brief Invoke an `operator()` template with a row equality comparator based on the specified
- * `compare_nans` parameter.
+ * @brief Invokes the given `func` with desired comparators based on the specified `compare_nans`
+ * parameter
+ *
+ * @tparam HaystackHasNested Flag indicating whether there are nested columns in haystack
+ * @tparam HasAnyNested Flag indicating whether there are nested columns in either haystack or
```
bdice, line 131: Why do we need two separate template parameters for "haystack has nested" and "haystack has nested or needles has nested"? I would have expected this to only require one dispatch. Is there a meaningful benefit to the case where "haystack has nested and not needles has nested"?

cpp/src/search/contains_table.cu
```diff
@@ -173,124 +200,106 @@ rmm::device_uvector<bool> contains(table_view const& haystack,
...
+                                        compare_nans,
+                                        haystack_has_nulls,
+                                        has_any_nulls,
+                                        self_equal,
+                                        two_table_equal,
+                                        d_hasher,
+                                        helper_func);
+  } else {
+    if (cudf::detail::has_nested_columns(needles)) {
+      dispatch_nan_comparator<false, true>(compare_nulls,
```
bdice, line 285: This is the case I'm interested in investigating in the other thread. Is the `<false, true>` dispatch providing special functionality or performance advantages that can't be achieved with 2 dispatches (`<true, true>` or `<false, false>`)?

PointKernel: Yeah, I see your point. 

Ideally, users should always check nested against nested. Having the `<false, true>` dispatch allows users to check nested needles against a flat haystack (though I'm not sure either if it's a legit use case), and the haystack set insertion would be faster than the `<true, true>` dispatch.

For my education, in this particular case, I assume `<true, true>` would work fine but less efficiently since we do redundant nested-related checks for a flat column. But would `<false, false>` produce proper results?

PointKernel: >  But would <false, false> produce proper results?

Checked with @divyegala, this will cause runtime errors.

bdice: > Having the `<false, true>` dispatch allows users to check nested needles against a flat haystack (though I'm not sure either if it's a legit use case)

Maybe I'm misunderstanding the problem -- but it seems like a nested needle could never be found in a haystack without nesting. Shouldn't that fail because the types of the needle and haystack don't match?

PointKernel: > Shouldn't that fail because the types of the needle and haystack don't match?

I'm not sure. The current `contains_table` allows this combination and returns 0 in this case. I'm trying to align with the existing behavior. Maybe @ttnghia knows more about what is the expected behavior here.

ttnghia: Sorry, I shouldn't approve this. I almost forgot about the behavior of the comparator regarding the `has_nulls`.

Here is the thing: when the column has nulls, the validity (true/false) of the column will be hashed into the output hash:
```
if (_check_nulls) {
          auto validity_it = detail::make_validity_iterator<true>(curr_col);
          hash             = detail::accumulate(..., hash,...);
```

When using hash table (map/set), we must ensure that the hash values for rows in both tables are consistent. That means, if one side table has nulls and hashed with nulls, the other side must also be hashed with the same `has_nulls`. Otherwise, one table is hashed differently than the other and you will not find the correct "contains" output.

In summary, the same `hash_nulls` value should be used for both inserting `lhs` and finding contains of `rhs`.

PointKernel: I think that's a different issue. Null handling is properly propagated to both self and two-table comparators.

The focus was whether it's legit to have mismatched column types between haystack and needles. After checking the public `contains` API, I think the answer is clearer now: I've updated the implementation so now it will throw if column types mismatch. The implementation is therefore simplified as @bdice suggested. Nice catch, thanks! :pray:

bdice (approved):
Thanks for fixing the dispatch logic here. I knew that seemed fishy! Everything looks good here to me. I haven't looked at benchmarks or compile times yet. Anything noteworthy there to consider before merging?

davidwendt:
A snapshot of the compile times near the beginning of 23.10 shows `src/search/contains_table.cu.o` at between 10.5 to 12 minute range when doing a full build. So I think the build time here would be about on par or a little better.

bdice (approved):
Approving CMake changes.

PR #14069, opened 2023-09-08
Update pyarrow-related dispatch logic in dask_cudf

rjzamora:
## Description
Updates `dask_cudf` dispatch logic to avoid breakage from https://github.com/dask/dask/pull/10500.
Also removes stale `try`/`except` logic.

## Checklist


python/dask_cudf/dask_cudf/backends.py
```diff
@@ -378,35 +371,37 @@ def percentile_cudf(a, q, interpolation="linear"):
...
-            )
-        return obj.from_arrow(table)
+@pyarrow_schema_dispatch.register((cudf.DataFrame,))
+def _get_pyarrow_schema_cudf(obj, preserve_index=True, **kwargs):
+    if kwargs:
+        warnings.warn(
+            "Ignoring the following arguments to "
+            f"`pyarrow_schema_dispatch`: {list(kwargs)}"
+        )
+    return obj.to_arrow(preserve_index=preserve_index).schema
```
wence-, line 382: **thought (non-blocking):** ‏If dask asks for the schema separately from the table it would be good to figure out a way to provide a pyarrow schema for a cudf dataframe without necessarily copying the full frame to host. Ideally producing this metadata should be O(1) rather than (as it is now) O(size-of-frame).

wence- (approved):
Thanks

python/dask_cudf/dask_cudf/backends.py
rjzamora: Yeah, good call. The specific logic in `_get_pyarrow_schema_cudf` wasn't really part of this PR, but that's an excellent point. It definitely feels like we should be able to produce a schema without moving **all** the data!

rjzamora: Using `meta_nonempty(obj)` seems to do the trick here.

PR #14070, opened 2023-09-08
Raise NotImplementedError for datetime strings with UTC offset

mroeschke:
## Description
Avoids e.g. DatetimeIndex(["2022-07-22 00:00:00+02:00"]) from dropping the +02:00 since timezones are not supported

## Checklist


python/cudf/cudf/core/column/column.py
```diff
@@ -2533,6 +2533,20 @@ def _construct_array(
...
             # Only way to construct an Interval column.
             return pd.array(arbitrary)
+        elif (
+            inferred_dtype == "string" and getattr(dtype, "kind", None) == "M"
+        ):
+            # We may have date-like strings with timezones
+            try:
+                pd_arbitrary = pd.to_datetime(arbitrary)
+            except pd.errors.OutOfBoundsDatetime:
+                # Not an issue in pandas>=2.0
```
vyasr, line 2543: Does this comment mean that this error won't be raised when we upgrade to pandas 2.0? If so, one suggestion I would have is that we implement a utility function that will raise an exception everywhere that we have a pandas<2.0 code path if the detected pandas version is 2.0. Then when we actually upgrade pandas we can validate that we changed everywhere that needed changing.

python/cudf/cudf/tests/test_datetime.py
```diff
@@ -2150,6 +2150,12 @@ def test_daterange_pandas_compatibility():
     assert_eq(expected, actual)
 
 
+@pytest.mark.filterwarnings("ignore:parsing timezone:DeprecationWarning")
```
vyasr, line 2153: Can we use `pytest.warns` here instead? See https://docs.rapids.ai/api/cudf/stable/developer_guide/testing/#testing-code-that-throws-warnings.

vyasr (changes requested):

python/cudf/cudf/core/column/column.py
mroeschke: Ah so actually this doesn't work yet in pandas (it was related to non-nanosecond support in pandas 2.0, but it's not implemented everywhere yet). I linked the issue to this exception path

python/cudf/cudf/tests/test_datetime.py
mroeschke: Sure thing, added

python/cudf/cudf/core/column/column.py
```diff
@@ -2533,6 +2533,20 @@ def _construct_array(
...
+        elif (
+            inferred_dtype == "string" and getattr(dtype, "kind", None) == "M"
+        ):
+            # We may have date-like strings with timezones
+            try:
+                pd_arbitrary = pd.to_datetime(arbitrary)
+            except pd.errors.OutOfBoundsDatetime:
+                # https://github.com/pandas-dev/pandas/issues/55096
+                pass
+            else:
```
wence-, line 2545: **nitpick (non-blocking):** ‏style

FWIW I tend to prefer not using else on try/except (I can never remember if it applies to the try or the except), so perhaps move this validation into the try block?

python/cudf/cudf/tests/test_datetime.py
```diff
@@ -2150,6 +2150,12 @@ def test_daterange_pandas_compatibility():
     assert_eq(expected, actual)
 
 
+def test_strings_with_utc_offset_not_implemented():
+    with pytest.warns(DeprecationWarning, match="parsing timezone"):
+        with pytest.raises(NotImplementedError):
```
wence-, line 2155: This one is cudf.

python/cudf/cudf/tests/test_datetime.py
```diff
@@ -2150,6 +2150,12 @@ def test_daterange_pandas_compatibility():
     assert_eq(expected, actual)
 
 
+def test_strings_with_utc_offset_not_implemented():
+    with pytest.warns(DeprecationWarning, match="parsing timezone"):
```
wence-, line 2154: And this one is pandas in the pd.to_datetime call?

wence- (approved):
Minor style suggestion

python/cudf/cudf/tests/test_datetime.py
mroeschke: This one is in cupy. I'll add a comment referencing this

python/cudf/cudf/core/column/column.py
mroeschke: Sure thing. Moved to the try block

vyasr (approved):

PR #14072, opened 2023-09-08
Remove header tests

ajschmidt8:
From some internal Slack discussions, it was determined that the `headers_test.sh` file is no longer necessary.

This PR removes it and its associated checks.


raydouglass (approved):

conda/recipes/libcudf/meta.yaml
```diff
@@ -107,252 +107,6 @@ outputs:
         - dlpack {{ dlpack_version }}
         - gtest {{ gtest_version }}
         - gmock {{ gtest_version }}
-    test:
-      commands:
-        - test -f $PREFIX/lib/libcudf.so
```
bdice, line 112: I'd like to retain the test that checks for the existence of `libcudf.so`, and maybe one header like `$PREFIX/include/cudf/column/column.hpp`. That means we have a spot-check that installation worked correctly, but we don't have to keep this list updated with all the headers (it should not need to change when new headers are added).

Same for `libcudf_kafka.so` below.

ajschmidt8: fixed in  4bd518f

PR #14079, opened 2023-09-11
Support for progressive parquet chunked reading.

nvdbaranec:
Previously, the parquet chunked reader operated by controlling the size of output chunks only.  It would still ingest the entire input file and decompress it, which can take up a considerable amount of memory.  With this new 'progressive' support, we also 'chunk' at the input level.  Specifically, the user can pass a `pass_read_limit` value which controls how much memory is used for storing compressed/decompressed data.  The reader will make multiple passes over the file, reading as many row groups as it can to attempt to fit within this limit.  Within each pass, chunks are emitted as before. 

From the external user's perspective, the chunked read mechanism is the same.  You call `has_next()` and `read_chunk()`.  If the user has specified a value for `pass_read_limit` the set of chunks produced might end up being different (although the concatenation of all of them will still be the same). 

The core idea of the code change is to add the idea of the internal `pass`.  Previously we had a `file_intermediate_data` which held data across `read_chunk()` calls.   There is now a `pass_intermediate_data` struct which holds information specific to a given pass.  Many of the invariant things from the file level before (row groups and chunks to process) are now stored in the pass intermediate data.  As we begin each pass, we take the subset of global row groups and chunks that we are going to process for this pass, copy them to out intermediate data, and the remainder of the reader reference this instead of the file-level data. 

In order to avoid breaking pre-existing interfaces, there's a new contructor for the `chunked_parquet_reader` class:

```
  chunked_parquet_reader(
    std::size_t chunk_read_limit,
    std::size_t pass_read_limit,
    parquet_reader_options const& options,
```

## Checklist


cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -306,33 +306,74 @@ struct ColumnChunkDesc {
 };
 
 /**
- * @brief Struct to store raw/intermediate file data before parsing.
+ * @brief The row_group_info class
+ */
+struct row_group_info {
+  size_type index;
+  size_t start_row;  // TODO source index
```
PointKernel, line 313: Is the TODO item still valid?

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -306,33 +306,74 @@ struct ColumnChunkDesc {
...
+  size_type source_index;
+
+  row_group_info() = default;
+
+  row_group_info(size_type index, size_t start_row, size_type source_index)
+    : index(index), start_row(start_row), source_index(source_index)
```
PointKernel, line 319:
```suggestion
    : index{index}, start_row{start_row}, source_index{source_index}
```

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -101,6 +101,7 @@ class reader::impl {
    * @param mr Device memory resource to use for device memory allocation
    */
   explicit impl(std::size_t chunk_read_limit,
+                std::size_t pass_read_limit,
```
PointKernel, line 104: docs for `pass_read_limit`

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -702,38 +702,39 @@ void reader::impl::allocate_nesting_info()
...
   }
 }
 
-std::pair<bool, std::vector<std::future<void>>> reader::impl::create_and_read_column_chunks(
-  cudf::host_span<row_group_info const> const row_groups_info, size_type num_rows)
+std::pair<bool, std::vector<std::future<void>>> reader::impl::read_and_decompress_column_chunks()
 {
-  auto& raw_page_data = _file_itm_data.raw_page_data;
-  auto& chunks        = _file_itm_data.chunks;
+  auto row_groups_info = _pass_itm_data->row_groups;
```
PointKernel, line 729:
```suggestion
  auto const& row_groups_info = _pass_itm_data->row_groups;
```
Copying the whole vector can be expensive

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -747,13 +748,68 @@ std::pair<bool, std::vector<std::future<void>>> reader::impl::create_and_read_co
...
+
+  CUDF_EXPECTS(remaining_rows == 0, "All rows data must be read.");
+
+  return {total_decompressed_size > 0, std::move(read_chunk_tasks)};
+}
+
+void reader::impl::load_global_chunk_info()
+{
+  auto const num_rows  = _file_itm_data.global_num_rows;
+  auto row_groups_info = _file_itm_data.row_groups;
```
PointKernel, line 799:
```suggestion
  auto const& row_groups_info = _file_itm_data.row_groups;
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -792,92 +843,171 @@ std::pair<bool, std::vector<std::future<void>>> reader::impl::create_and_read_co
...
-                                                         0,
-                                                         chunks.size(),
-                                                         column_chunk_offsets,
-                                                         chunk_source_map,
-                                                         _stream));
+void reader::impl::compute_input_pass_row_group_info()
+{
+  // at this point, row_groups has already been filtered down to just the row groups we need to
+  // handle optional skip_rows/num_rows parameters.
+  auto row_groups_info = _file_itm_data.row_groups;
```
PointKernel, line 856:
```suggestion
  auto const& row_groups_info = _file_itm_data.row_groups;
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -792,92 +843,171 @@ std::pair<bool, std::vector<std::future<void>>> reader::impl::create_and_read_co
...
+  //
+  std::size_t const read_limit =
+    _input_pass_read_limit > 0 ? _input_pass_read_limit : std::numeric_limits<std::size_t>::max();
+  std::size_t cur_read      = 0;
+  std::size_t cur_rg_start  = 0;
+  std::size_t cur_row_count = 0;
+  _input_pass_row_group_indices.push_back(0);
+  _input_pass_row_count.push_back(0);
+
+  // for (auto const& rg : row_groups_info) {
```
PointKernel, line 875:
```suggestion
```

cpp/src/io/parquet/reader_impl_preprocess.cu
nvdbaranec: Whoops. Good catch.

cpp/src/io/parquet/parquet_gpu.hpp
nvdbaranec: Nope.  Will fix.

cpp/include/cudf/io/detail/parquet.hpp
```diff
@@ -102,17 +103,21 @@ class chunked_reader : private reader {
...
-   * If `chunk_read_limit == 0` (i.e., no reading limit), a call to `read_chunk()` will read the
-   * whole file and return a table containing all rows.
+   * If `chunk_read_limit == 0` (i.e., no output limit), and `pass_read_limit == 0` (no input
+   * temporary memory size limit) a call to `read_chunk()` will read the whole file and return a
+   * table containing all rows.
    *
    * @param chunk_read_limit Limit on total number of bytes to be returned per read,
    *        or `0` if there is no limit
+   * @param pass_read_limit Limit on total amount of memory used for temporary computations during
+   * loading, or `0` if there is no limit.
```
ttnghia, line 113: Do not end a `@param` with period `.`. People will complain about it ;)

cpp/include/cudf/io/parquet.hpp
```diff
@@ -445,6 +445,30 @@ class chunked_parquet_reader {
...
+   * output table for each reading, and a byte limit on the amount of temporary memory to use
+   * when reading. pass_read_limit affects how many row groups we can read at a time by limiting
+   * the amount of memory dedicated to decompression space. pass_read_limit is a hint, not an
+   * absolute limit - if a single row group cannot fit within the limit given, it will still be
+   * loaded.
+   *
+   * @param chunk_read_limit Limit on total number of bytes to be returned per read,
+   *        or `0` if there is no limit
+   * @param pass_read_limit Limit on the amount of memory used for reading and decompressing data or
+   * `0` if there is no limit.
```
ttnghia, line 462: No period `.`.

ttnghia, line 462: Please use consistent alignment.

cpp/include/cudf/io/detail/parquet.hpp
ttnghia, line 113: Also please use consistent alignment.

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -306,33 +306,74 @@ struct ColumnChunkDesc {
...
+ * @brief Structs to identify the reading row range for each chunk of rows in chunked reading.
+ */
+struct chunk_read_info {
+  size_t skip_rows;
+  size_t num_rows;
+};
+
+/**
+ * @brief Struct to store pass-level data that remains constant for
+ * a single pass.
```
ttnghia, line 352:
```suggestion
 * @brief Struct to store pass-level data that remains constant for a single pass.
```

cpp/src/io/parquet/reader.cpp
```diff
@@ -43,12 +43,14 @@ table_with_metadata reader::read(parquet_reader_options const& options)
...
 chunked_reader::chunked_reader(std::size_t chunk_read_limit,
+                               std::size_t pass_read_limit,
                                std::vector<std::unique_ptr<datasource>>&& sources,
                                parquet_reader_options const& options,
                                rmm::cuda_stream_view stream,
                                rmm::mr::device_memory_resource* mr)
 {
-  _impl = std::make_unique<impl>(chunk_read_limit, std::move(sources), options, stream, mr);
+  _impl = std::make_unique<impl>(
+    chunk_read_limit, pass_read_limit, std::move(sources), options, stream, mr);
```
ttnghia, line 53: Can we move this into the initializer list of this constructor?

cpp/src/io/utilities/hostdevice_vector.hpp
```diff
@@ -96,6 +96,14 @@ class hostdevice_vector {
   [[nodiscard]] size_t size_bytes() const noexcept { return sizeof(T) * size(); }
   [[nodiscard]] bool empty() const noexcept { return size() == 0; }
 
+  void resize(size_t new_size, rmm::cuda_stream_view stream)
+  {
+    CUDF_EXPECTS(new_size <= capacity(),
+                 "hostdevice_vector resize must fit within existing capacity");
+    stream.synchronize();
```
ttnghia, line 103: Why?

vuule (commented):
review.flush()

cpp/include/cudf/io/detail/parquet.hpp
```diff
@@ -102,17 +103,21 @@ class chunked_reader : private reader {
...
    * @param sources Input `datasource` objects to read the dataset from
    * @param options Settings for controlling reading behavior
    * @param stream CUDA stream used for device memory operations and kernel launches.
    * @param mr Device memory resource to use for device memory allocation
    */
   explicit chunked_reader(std::size_t chunk_read_limit,
+                          std::size_t pass_read_limit,
```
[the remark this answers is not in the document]
hyperbolic2346: I see from context below that this is making multiple passes at reading the file. Maybe detail of that should bubble up to the header somehow?

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -747,13 +748,68 @@ std::pair<bool, std::vector<std::future<void>>> reader::impl::create_and_read_co
...
+  size_type chunk_count = 0;
   for (auto const& rg : row_groups_info) {
     auto const& row_group       = _metadata->get_row_group(rg.index, rg.source_index);
-    auto const row_group_start  = rg.start_row;
     auto const row_group_source = rg.source_index;
     auto const row_group_rows   = std::min<int>(remaining_rows, row_group.num_rows);
 
+    // generate ColumnChunkDesc objects for everything to be decoded (all input columns)
+    for (size_t i = 0; i < num_input_columns; ++i) {
+      auto col = _input_columns[i];
```
hyperbolic2346, line 760: Should this be a reference?

hyperbolic2346 (commented):
Some comments on first pass. I'll make another one tomorrow with a clear head.

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -278,12 +276,28 @@ class reader::impl {
...
+
   cudf::io::parquet::gpu::file_intermediate_data _file_itm_data;
-  cudf::io::parquet::gpu::chunk_intermediate_data _chunk_itm_data;
-  std::vector<cudf::io::parquet::gpu::chunk_read_info> _chunk_read_info;
-  std::size_t _chunk_read_limit{0};
-  std::size_t _current_read_chunk{0};
+  std::unique_ptr<cudf::io::parquet::gpu::pass_intermediate_data> _pass_itm_data;
+
+  // an array of indices into _file_itm_data::global_chunks. Each pair of indices represents
+  // the start/end of the chunks to be loaded for a given pass.
```
vuule, line 292: FWIW I did not understand this comment before looking at how `_input_pass_row_group_indices` is used. I don't think "pair" is the right term here.

vuule (commented):
another flush, mostly comments coming from difficulty to understand the code. 
Still got some parts to dig into.

cpp/src/io/utilities/hostdevice_vector.hpp
nvdbaranec: I wanted to avoid the complexity of having to actually reallocate/copy existing elements since the use case here is very narrow.

cpp/include/cudf/io/parquet.hpp
```diff
@@ -445,6 +445,30 @@ class chunked_parquet_reader {
...
+   * @param options The options used to read Parquet file
+   * @param mr Device memory resource to use for device memory allocation
+   */
+  chunked_parquet_reader(
```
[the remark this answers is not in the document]
nvdbaranec: I think the other one is still useful.  If you just want to limit how big your output is you can use that.  The new one is a bit more tweaky/tune-y where you're playing games with a much more nebulous set of internal state.

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -306,33 +306,74 @@ struct ColumnChunkDesc {
 };
 
 /**
- * @brief Struct to store raw/intermediate file data before parsing.
+ * @brief The row_group_info class
+ */
+struct row_group_info {
+  size_type index;
```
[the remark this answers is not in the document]
nvdbaranec: Right.  This and `source_index]` are the params passed to `aggregate_reader_metadata::get_row_group`. I'll add comments.

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -306,33 +306,74 @@ struct ColumnChunkDesc {
...
 
-/**
- * @brief Structs to identify the reading row range for each chunk of rows in chunked reading.
- */
-struct chunk_read_info {
+  std::vector<gpu::chunk_read_info> output_chunk_read_info;
+  std::size_t current_output_chunk{0};
+
+  rmm::device_buffer level_decode_data;
+  int level_type_size;
```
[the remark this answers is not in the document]
nvdbaranec: No.  These are values that get initialized deep down during the decode process.  Any default would be meaningles.  Whereas `current_output_chunk` being initialized to 0 makes the code (very) slighly simpler.

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -306,33 +306,74 @@ struct ColumnChunkDesc {
...
+  size_t global_skip_rows;
+  size_t global_num_rows;
+};
+
+/**
+ * @brief Structs to identify the reading row range for each chunk of rows in chunked reading.
+ */
+struct chunk_read_info {
+  size_t skip_rows;
+  size_t num_rows;
```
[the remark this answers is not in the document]
nvdbaranec: I don't think so.  These are just generated during the chunking process for the output chunks 
https://github.com/rapidsai/cudf/blob/31e56702fe15f44b3e849207d31d0bb79c307367/cpp/src/io/parquet/reader_impl_preprocess.cu#L1184
This PR doesn't really touch that stuff -  it adds the "pass" concept in an attempt to de-overload the term "chunk"

cpp/include/cudf/io/parquet.hpp
```diff
@@ -445,6 +445,30 @@ class chunked_parquet_reader {
...
+   * `cudf::read_parquet()`, with additional parameters to specify the size byte limit of the
+   * output table for each reading, and a byte limit on the amount of temporary memory to use
+   * when reading. pass_read_limit affects how many row groups we can read at a time by limiting
+   * the amount of memory dedicated to decompression space. pass_read_limit is a hint, not an
+   * absolute limit - if a single row group cannot fit within the limit given, it will still be
+   * loaded.
+   *
+   * @param chunk_read_limit Limit on total number of bytes to be returned per read,
+   *        or `0` if there is no limit
+   * @param pass_read_limit Limit on the amount of memory used for reading and decompressing data or
```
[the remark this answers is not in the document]
nvdbaranec: This is a soft limit because it will attempt to continue even if it can't meet the limit.  For example, if the user specified `1 MB` and it can't fit even one row group (say `50 MB`) into that size, it will still attempt to read/decompress one row group at a time.

cpp/src/io/parquet/reader_impl_preprocess.cu
nvdbaranec: Yep.

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -278,12 +276,28 @@ class reader::impl {
...
 
-  // Variables used for chunked reading:
+  // chunked reading happens in 2 parts:
+  //
+  // At the top level there is the "pass" in which we try and limit the
+  // total amount of temporary memory (compressed data, decompressed data) in use
+  // via _input_pass_read_limit.
+  //
+  // Within a pass, we produce one or more chunks of output, whose maximum total
+  // byte size is controlled by _output_chunk_read_limit.
```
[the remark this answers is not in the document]
nvdbaranec: Yes. But it's worth pointing out that that was the case even without passes.  Prior to this PR, we just had one big "pass". Now we just have a number of them, where each one is limited by input memory usage.  Internal to a pass, the chunking system doesn't know the difference.

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -278,12 +276,28 @@ class reader::impl {
   std::optional<std::vector<reader_column_schema>> _reader_column_schema;
   data_type _timestamp_type{type_id::EMPTY};
 
-  // Variables used for chunked reading:
+  // chunked reading happens in 2 parts:
+  //
+  // At the top level there is the "pass" in which we try and limit the
+  // total amount of temporary memory (compressed data, decompressed data) in use
+  // via _input_pass_read_limit.
```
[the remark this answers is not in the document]
nvdbaranec: Not exactly. Chunks are limited only by the output chunk read limit (the limit on how many bytes of data can be in a chunk).  The only thing that this would affect is the number of chunks you might get back.  For example, if you had a case where the input limit allowed everything to happen in 1 pass, you might get 1 chunk.  But if the input limit caused it to go through 3 passes, you'd get back at least 3 chunks.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -792,92 +843,170 @@ std::pair<bool, std::vector<std::future<void>>> reader::impl::create_and_read_co
...
+
+  auto const chunks_per_rowgroup = _input_columns.size();
+  auto const num_chunks          = chunks_per_rowgroup * num_row_groups;
+
+  auto chunk_start = _file_itm_data.chunks.begin() + (row_group_start * chunks_per_rowgroup);
+  auto chunk_end   = _file_itm_data.chunks.begin() + (row_group_end * chunks_per_rowgroup);
+
+  _pass_itm_data->chunks =
+    cudf::detail::hostdevice_vector<gpu::ColumnChunkDesc>(0, num_chunks, _stream);
+  _pass_itm_data->chunks.resize(num_chunks, _stream);
```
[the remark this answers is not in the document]
nvdbaranec: Good catch.  That's a remnant of an earlier structure of the code where we were allocating things earlier and then resizing. Unnecessary now (and I can also remove the change to `hostdevice_vector`)

cpp/src/io/utilities/hostdevice_vector.hpp
nvdbaranec: Also, as @vuule pointed out down below, the usage for this is unnecessary.  An earlier iteration of the code had the vector initialized early on and then resized downwards later.  But in the current iteration, it is all done in one place so this change is not necessary. I'll remove it.

cpp/include/cudf/io/detail/parquet.hpp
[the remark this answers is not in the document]
nvdbaranec: Added a big chunk of comments detailing passes vs chunks.

cpp/include/cudf/io/detail/parquet.hpp
```diff
@@ -102,17 +103,21 @@ class chunked_reader : private reader {
...
    *
-   * If `chunk_read_limit == 0` (i.e., no reading limit), a call to `read_chunk()` will read the
-   * whole file and return a table containing all rows.
+   * If `chunk_read_limit == 0` (i.e., no output limit), and `pass_read_limit == 0` (no input
+   * temporary memory size limit) a call to `read_chunk()` will read the whole file and return a
+   * table containing all rows.
    *
    * @param chunk_read_limit Limit on total number of bytes to be returned per read,
    *        or `0` if there is no limit
+   * @param pass_read_limit Limit on total amount of memory used for temporary computations during
```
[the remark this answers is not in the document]
nvdbaranec: Done.

cpp/src/io/parquet/reader.cpp
nvdbaranec: Can't initialize a parent's member as part of your own initialization list.

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -133,22 +136,22 @@ class reader::impl {
                     host_span<std::vector<size_type> const> row_group_indices,
                     std::optional<std::reference_wrapper<ast::expression const>> filter);
 
+  void load_global_chunk_info();
+  void compute_input_pass_row_group_info();
+  void setup_pass();
```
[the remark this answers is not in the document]
nvdbaranec: If these were made free functions they would all have to be passed large numbers of parameters, which would the code a lot harder to read.

cpp/src/io/parquet/reader_impl.hpp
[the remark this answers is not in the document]
vuule: > if you had a case where the input limit allowed everything to happen in 1 pass, you might get 1 chunk. But if the input limit caused it to go through 3 passes, you'd get back at least 3 chunks.

Chunks can't span passes. This answers my question, thanks.

cpp/src/io/parquet/reader_impl.hpp
vuule: hmm, maybe just changing to "each pair of adjacent indices" is enough here.

PointKernel (approved):

vuule (commented):
few more small suggestions, looks good overall.
I think the biggest issue for me is the fragility of the test, which we discussed offline and found a solution for.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -792,92 +843,169 @@ std::pair<bool, std::vector<std::future<void>>> reader::impl::create_and_read_co
...
+    // can we add this row group
+    if (cur_read + row_group.total_byte_size >= read_limit) {
+      // always need to add at least 1 row group, so add ourselves
+      if (cur_rg_start == cur_rg_index) {
+        _input_pass_row_group_offsets.push_back(cur_rg_index + 1);
+        _input_pass_row_count.push_back(cur_row_count + row_group.num_rows);
+        cur_rg_start = cur_rg_index + 1;
+        cur_read     = 0;
+      }
+      // add the previous group
```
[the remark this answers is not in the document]
vuule: could also be "start of current row group", wasn't sure what is clearer.

vuule (approved):

PR #14082, opened 2023-09-11
Add fallback matrix for nvcomp.

bdice:
## Description
Some platforms (such as aarch64 + CUDA 12) don't have a matching matrix entry for nvcomp. This PR adds a fallback matrix entry so it is possible to attempt local development on aarch64 with CUDA 12.

## Checklist

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -241,6 +241,9 @@ dependencies:
               cuda: "11.8"
             packages:
               - *nvcomp
+          # Fallback matrix for aarch64 CUDA 12
+          - matrix:
+            packages:
   build_wheels:
     common:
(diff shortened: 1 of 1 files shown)


dependencies.yaml
```diff
@@ -241,6 +241,9 @@ dependencies:
               cuda: "11.8"
             packages:
               - *nvcomp
+          # Fallback matrix for aarch64 CUDA 12
```
bdice, line 244:
```suggestion
          # TOD: Temporary fallback matrix for aarch64 CUDA 12. After migrating to nvcomp 3,
          # all CUDA/arch combinations should be supported by existing packages.
```

PR #14093, opened 2023-09-12
Support negative preceding/following for ROW window functions

mythrocks:
## Description

This commit adds support for "offset" ROW windows, where the preceding and following
window bounds are allowed to have negative values.  This allows window definitions to
exclude the current row entirely.

Prior to this change, ROW-based windows *had* to include the current row, causing
`preceding` and `following` to support only non-negative values.  Additionally, the
inclusion of the current row would count against the `min_periods` check.

The following is an example of the new "negative" semantics.  Consider the input:
```c++
auto const row = ints_column{1, 2, 3, 4};
```
If the window bounds are specified as (preceding=3, following=-1), then the window
for the third row (`3`) is `{1, 2}`.
`following=-1` indicates a "following" row *before* the current row.

A negative value for `preceding` follows the existing convention of including the
current row.  This makes it slightly more involved:
  1. `preceding=2` indicates *one* row before the current row.
  2. `preceding=1` indicates the current row.
  3. `preceding=0` indicates one row past (i.e. after) the current row.
  4. `preceding=-1` indicates two rows after the current row.
Et cetera.

`min_periods` checks continue to be honoured as before, but the requirement for
positive `min_periods` is dropped.  `min_periods` only need be non-negative.

## Checklist


mythrocks:
Moved this back to draft. There are a couple of things pending:
1. A Spark test had revealed a possible bug at the group boundaries. That test needs to be formalized in the CUDF test.
2. One if-check can be moved out of the kernel for fixed window bounds.
3. Must merge latest from `origin/branch-23.10`.

I'll update these very shortly.

mythrocks:
A note to the reviewer:
1. Most of the bulk seems to be the tests.
2. In `grouped_rolling.cu`: The new `row_based_preceding_calc` and `row_based_following_calc` are a result of refactoring what used to be a simpler lambda.  This refactor allows the `if (following < 0)` check to be `constexpr`.
3. [Here](https://github.com/rapidsai/cudf/pull/14093/files#diff-7d124511e62781670dc53992c820cc982d594f5ce16fb385a79d084116cd7c99R75), an additional change was made to materialize the `preceding_column` and `following_column`, so that the rolling kernels need not be instantiated with complicated iterators.

cpp/tests/rolling/offset_row_window_test.cpp
```diff
@@ -0,0 +1,294 @@
...
+                                 ints_column{{2, 2, 2, 2, 1, 0, 2, 2, 1, 0}, no_nulls()});
+
+  CUDF_TEST_EXPECT_COLUMNS_EQUAL(
+    *run_rolling(*AGG_COLLECT_LIST),
+    lists_column{{{1, 2}, {2, 3}, {3, 4}, {4, 5}, {5}, {}, {7, 8}, {8, 9}, {9}, {}}, no_nulls});
+}
+
+TEST_F(OffsetRowWindowTest, OffsetRowWindow_Ungrouped_0_to_2)
+{
+  auto const preceding = cudf::window_bounds::get(0);
```
robertmaynard, line 218: I don't see a test where verify negative `preceding` values like -1 and -200 ( verify all clamp ).

cpp/tests/rolling/offset_row_window_test.cpp
```diff
@@ -0,0 +1,294 @@
...
+  cudf::make_count_aggregation<cudf::rolling_aggregation>(cudf::null_policy::INCLUDE);
+auto const AGG_MIN          = cudf::make_min_aggregation<cudf::rolling_aggregation>();
+auto const AGG_MAX          = cudf::make_max_aggregation<cudf::rolling_aggregation>();
+auto const AGG_SUM          = cudf::make_sum_aggregation<cudf::rolling_aggregation>();
+auto const AGG_COLLECT_LIST = cudf::make_collect_list_aggregation<cudf::rolling_aggregation>();
+
+TEST_F(OffsetRowWindowTest, OffsetRowWindow_Grouped_3_to_Minus_1)
+{
+  auto const preceding = cudf::window_bounds::get(3);
+  auto const following = cudf::window_bounds::get(-1);
```
robertmaynard, line 97: If I am understanding `following` correctly it could be -4 to capture 4 rows before the current row. We shouldn't have all the 'before' tests use same `following` value.

robertmaynard (changes requested):
We need to update https://github.com/rapidsai/cudf/blob/branch-23.10/cpp/include/cudf/rolling.hpp#L181 to document the relaxation of preceding_window and  following_window

cpp/tests/rolling/offset_row_window_test.cpp
mythrocks: Added a better test for this, for varying values of `preceding`.

cpp/tests/rolling/offset_row_window_test.cpp
mythrocks: Added better tests for different following values.

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -94,6 +93,109 @@ std::unique_ptr<column> grouped_rolling_window(table_view const& group_keys,
 
 namespace detail {
 
+/// Preceding window calculation functor.
+template <bool preceding_less_than_1>
```
divyegala, line 97: What is the advantage and disadvantage of having this template?

mythrocks: Ah, right. This is to enable the `if constexpr()` in `operator()`:
```c++
    if constexpr (preceding_less_than_1) {  // where 1 indicates only the current row.
      auto group_end = _group_offsets_begin[group_label + 1];
      return thrust::maximum{}(_preceding_window, -(group_end - 1 - idx));
    } else { ... }
```
When this used to be a lambda, this used to be:
```c++
    if (preceding < 1) { ... } else { ... }
```
That if check would have run once per row in the window. Or worst case N**2 times, over the column. The template was so that this is checked once.

divyegala: Has there been a tangible benefit to runtime perf with this change?

mythrocks: Sorry if the refactor isn't immediately obvious. This commit shows the refactor: 589f2a57c4a979c832cf733d030b51cdc097ac54

mythrocks:
Ah, I haven't updated the header yet.  Will do shortly.

cpp/src/rolling/grouped_rolling.cu
mythrocks: Ah, I don't have a benchmark in place for this. 

I did assume that lifting this if check out would be beneficial.

mythrocks: Thanks for accommodating, @divyegala.

I've measured how much longer `grouped_rolling.cu` takes to compile, with all the changes in this PR thrown in.
Old: 98s
New: 93s

It's taking less time to compile now.  This might have to do with my materializing the preceding/following columns, to simplify the iterators passed to `cudf::detail::rolling_window`.

divyegala (approved):

cpp/src/rolling/grouped_rolling.cu
mythrocks: As an aside, the compile time for `rolling_fixed_window.cu` has reduced from 80s to 65s.

So @divyegala's instinct was right, to keep an eye on the compile times.  I think I've mitigated it for now.

mythrocks:
> We need to update https://github.com/rapidsai/cudf/blob/branch-23.10/cpp/include/cudf/rolling.hpp#L181 to document the relaxation of preceding_window and following_window

I've updated the docs in `rolling.hpp`.

cpp/include/cudf/rolling.hpp
```diff
@@ -199,6 +199,24 @@ struct window_bounds {
...
+ *                     |
+ *               current_row
+ *
+ * Similarly, `preceding` could have a negative value, indicating that the window begins at a
+ * position after the current row.  It differs slightly from the semantics for `following`, because
+ * `preceding` includes the current row. Therefore:
+ *   1. preceding=1  => Window starts at the current row.
+ *   2. preceding=0  => Window starts at 1 past the current row.
+ *   3. preceding=-1 => Window starts at 2 past the current row. Etc.
+ *
```
robertmaynard, line 219: I think the short help for `preceding_window` and `following_window` still need a tweak. Something like ` The static rolling window size in the backward direction when positive`

robertmaynard (changes requested):

cpp/include/cudf/rolling.hpp
mythrocks: Ah, right. Just a moment.

robertmaynard (approved):

PR #14095, opened 2023-09-12
Refactor `hash_reduce_by_row`

ttnghia:
This PR extracts `hash_reduce_by_row` function from `distinct_reduce.*` files. Previously, that function was designed specifically to work with `distinct` in stream compaction with `size_type` output. Now, it becomes more generic and can support more generic reduction operations and various output types. 

No new functionality was added.

The changes in this work pave the way for implementing histogram/merge histogram aggregations, which also rely on hash-base reduction.


cpp/src/reductions/hash_reduce_by_row.cuh
```diff
@@ -0,0 +1,164 @@
+/*
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
ttnghia, line 2: This file is extracted from `distinct_reduce.cu` thus I keep the copyright as-is.

karthikeyann (approved):
LGTM 👍

cpp/src/reductions/hash_reduce_by_row.cuh
PointKernel, line 1: Shall we move the header to `include` since it's used by both `stream_compaction` and `reductions`?

cpp/src/stream_compaction/distinct_reduce.hpp
```diff
@@ -68,11 +66,13 @@ auto constexpr reduction_init_value(duplicate_keep_option keep)
...
  * @param keep The parameter to determine what type of reduction to perform
  * @param nulls_equal Flag to specify whether null elements should be considered as equal
+ * @param nans_equal Flag to specify whether NaN values in floating point column should be
+ *        considered equal.
  * @param stream CUDA stream used for device memory operations and kernel launches
  * @param mr Device memory resource used to allocate the returned vector
  * @return A device_uvector containing the reduction results
  */
-rmm::device_uvector<size_type> hash_reduce_by_row(
+rmm::device_uvector<size_type> distinct_reduce(
```
PointKernel, line 75: `distinct` already implies removing duplicates within the group (or reduction) thus `distinct_reduce` doesn't sound right to me.

ttnghia: Do you have a suggestion for the new name?

cpp/src/reductions/hash_reduce_by_row.cuh
ttnghia: Sound good.

cpp/src/stream_compaction/distinct_reduce.hpp
ttnghia: Probably I'll merge this into `distinct.cu` if compiling this file is not an issue.

ttnghia: Alright, merging this inflates the compile time from 40s to 90s. Not too bad if compared with many minutes of compilation time of many others. Then I'll merge it.

PointKernel: Ah good point. Build time is another reason that we are doing this file splitting. Maybe `reduce_by_row` or `reduce_helper`?

cpp/include/cudf/detail/hash_reduce_by_row.cuh
```diff
@@ -14,61 +14,56 @@
  * limitations under the License.
  */
 
-#include "distinct_reduce.cuh"
+#include <stream_compaction/stream_compaction_common.cuh>
```
PointKernel, line 17: Including a `src` header in `include` is fishy. Can we move the relevant code to this file and get rid of this inclusion?

ttnghia: The code in this file needs to use `compaction_hash` in `<src/>` header (`stream_compaction_common.cuh`). I overlooked that. So we must put this `hash_reduce_by_row.cuh` in the `src` folder.

PointKernel: `compaction_hash` is a hash value adaptor that avoids getting the sentinel value as the hashing output. This is a generic functionality (not stream compaction specific) and can be placed in the detail header. The `compaction_hash` logic can be removed once we use e.g. `-1` as the key sentinel but we can defer this change when migrating `distinct` to use new cuco set/map.

To make this right and facilitate the future refactoring, I would suggest that we:

- Move the `reduce_by_row.cuh` back to `include`
- Move `compaction_hash` functor to the `reduce_by_row.cuh` file and rename it as `sentinel_hash_adaptor` (or simply `hash_adaptor``)

PointKernel: https://github.com/rapidsai/cudf/blob/07845438855c3787dc5daec0195bd75a399cd598/cpp/src/stream_compaction/stream_compaction_common.hpp#L33

If you are worrying about the stream compaction key sentinel value, Just change it to `-1` and we don't need  `compaction_hash` at all.

ttnghia: Then there will be some small changes to other files. But that should be good, since we are only removing code.

PointKernel (approved):
Looks great!

Thanks for the iteration and your effort in addressing reviews.

bdice (approved):
Approving CMake changes.

PR #14098, opened 2023-09-13
Validate ignore_index type in drop_duplicates

mroeschke:
## Description
Currently allows odd behavior like

```python
In [1]: import cudf

In [4]: df = cudf.DataFrame({"a": [1, 2, 1, 3]})

In [6]: df.drop_duplicates(ignore_index="True")
Out[6]: 
   a
0  1
1  2
2  3
```

## Checklist

### DIFF
--- a/python/cudf/cudf/core/indexed_frame.py
+++ b/python/cudf/cudf/core/indexed_frame.py
@@ -1961,7 +1961,7 @@ class IndexedFrame(Frame):
         ignore_index: bool, default False
             If True, the resulting axis will be labeled 0, 1, ..., n - 1.
         """
-        if not isinstance(ignore_index, bool):
+        if not isinstance(ignore_index, (np.bool_, bool)):
             raise ValueError(
                 f"{ignore_index=} must be bool, "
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1961,6 +1961,11 @@ def drop_duplicates(
         ignore_index: bool, default False
             If True, the resulting axis will be labeled 0, 1, ..., n - 1.
         """
+        if not isinstance(ignore_index, bool):
```
galipremsagar, line 1964: Does pandas allow `np.bool_`'s too? If yes, we will need to allow that too.

mroeschke: Good catch. Yes pandas accepts `np.bool_` so added that here

galipremsagar (approved):

python/cudf/cudf/core/indexed_frame.py
galipremsagar: Thanks @mroeschke !

PR #14100, opened 2023-09-13
Add DELTA_BINARY_PACKED encoder for Parquet writer

etseidl:
## Description
Part of #13501. Adds ability to fall back on DELTA_BINARY_PACKED encoding when V2 page headers are selected and dictionary encoding is not possible.

## Checklist


vuule (commented):
Thank you for making the effort to make this complex PR readable. Really appreciate the amount of comments.
Bunch of small comments so far. Still need to look into `page_enc.cu`.

cpp/tests/io/parquet_test.cpp
```diff
@@ -6732,4 +6734,34 @@ TEST_P(ParquetV2Test, CheckEncodings)
...
+using DeltaBinaryTypes =
+  cudf::test::Concat<cudf::test::IntegralTypesNotBool, cudf::test::ChronoTypes, DeltaDecimalTypes>;
+using SupportedDeltaTestTypes =
+  cudf::test::RemoveIf<cudf::test::ContainedIn<UnsupportedChronoTypes>, DeltaBinaryTypes>;
+TYPED_TEST_SUITE(ParquetWriterDeltaTest, SupportedDeltaTestTypes);
+
+TYPED_TEST(ParquetWriterDeltaTest, SupportedDeltaTestTypes)
+{
+  using T   = TypeParam;
+  auto col0 = testdata::ascending<T>();
```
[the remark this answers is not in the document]
etseidl: Not really, just need one column with a constant sign for the delta and one that's random.

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,276 @@
...
+  void* _bitpack_tmp;  // pointer to shared scratch memory used in bitpacking
+
+  // write the delta binary header. only call from thread 0
+  inline __device__ void write_header(T first_value)
+  {
+    delta::put_uleb128(_dst, delta::block_size);
+    delta::put_uleb128(_dst, delta::num_mini_blocks);
+    delta::put_uleb128(_dst, _num_values);
+    delta::put_zz128(_dst, first_value);
+  }
```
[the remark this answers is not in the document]
etseidl: Looks right. Hold over from when this was not in a class.

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,276 @@
...
+{
+  zigzag128_t s = (v < 0);
+  put_uleb128(p, (v ^ -s) * 2 + s);
+}
+
+// a block size of 128, with 4 mini-blocks of 32 values each fits nicely without consuming
+// too much shared memory.
+constexpr int block_size            = 128;
+constexpr int num_mini_blocks       = 4;
+constexpr int values_per_mini_block = block_size / num_mini_blocks;
```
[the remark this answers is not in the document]
etseidl: Which? block_size has to be a multiple of 128, and values_per_mb has to be a multiple of 32.

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,276 @@
...
+#include "parquet_gpu.hpp"
+
+#include <cudf/detail/utilities/cuda.cuh>
+#include <cudf/detail/utilities/integer_utils.hpp>
+
+#include <cub/cub.cuh>
+
+namespace cudf::io::parquet::gpu {
+
+namespace delta {
```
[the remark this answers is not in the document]
etseidl: I put stuff only for use in this file in the delta namespace. I would have used an anonymous namespace if it weren't a header.

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,276 @@
...
+
+    // now each warp encodes it's data...can calculate starting offset with _mb_bits
+    uint8_t* mb_ptr = _dst;
+    switch (warp_id) {
+      case 3: mb_ptr += _mb_bits[2] * delta::values_per_mini_block / 8; [[fallthrough]];
+      case 2: mb_ptr += _mb_bits[1] * delta::values_per_mini_block / 8; [[fallthrough]];
+      case 1: mb_ptr += _mb_bits[0] * delta::values_per_mini_block / 8;
+    }
+
+    // encoding happens here....will have to update pack literals to deal with larger numbers
```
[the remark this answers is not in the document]
etseidl: They were, I'll do another pass and see what can be deleted (once power comes back on 🫠)

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -90,6 +90,32 @@ struct input_column_info {
...
+constexpr uint32_t BitAnd(Enum a, Enum b)
+{
+  return static_cast<uint32_t>(a) & static_cast<uint32_t>(b);
+}
+
+template <class Enum, typename std::enable_if_t<is_scoped_enum<Enum>::value, bool> = true>
+constexpr uint32_t BitAnd(uint32_t a, Enum b)
+{
+  return a & static_cast<uint32_t>(b);
+}
```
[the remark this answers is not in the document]
etseidl: will have to think more on this one...`is_scoped_enum` is unhappy if it's passed an `int32_t`

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,276 @@
...
+    __syncthreads();
+
+    // now each warp encodes it's data...can calculate starting offset with _mb_bits
+    uint8_t* mb_ptr = _dst;
+    switch (warp_id) {
+      case 3: mb_ptr += _mb_bits[2] * delta::values_per_mini_block / 8; [[fallthrough]];
+      case 2: mb_ptr += _mb_bits[1] * delta::values_per_mini_block / 8; [[fallthrough]];
+      case 1: mb_ptr += _mb_bits[0] * delta::values_per_mini_block / 8;
+    }
+
```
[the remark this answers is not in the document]
etseidl: comparing the assembler, I think I'll stick with the switch 😉

cpp/src/io/parquet/parquet_gpu.hpp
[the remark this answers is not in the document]
etseidl: done. thanks stack overflow :)

cpp/src/io/parquet/delta_enc.cuh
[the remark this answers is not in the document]
vuule: Specifically referring to `values_per_mini_block`. We have the loopswitch that scales with delta::values_per_mini_block / 8 in every step. If we know that delta::values_per_mini_block % 8 == 0, then we can just scale the final value.
Don't know if this is relevant, haven't looked into the generated code.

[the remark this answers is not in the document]
etseidl: some implementations will bump `values_per_mini_block` to 64 for `INT64`. I'm not convinced that's a win, so for now sticking with 32 in the encoder.  In either case, it would always be divisible by 8.

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,282 @@
...
+  // zero out scratch
+  scratch[lane_id] = 0;
+  __syncwarp();
+
+  // TODO: see if there is any savings using special packing for easy bitwidths (1,2,4,8,16...)
+  // like what's done for the RLE encoder.
+  if (nbits == div) {
+    if (lane_id < count) {
+      for (int i = 0; i < sizeof(scratch_type); i++) {
```
vuule, line 85: This might be faster for small values
```suggestion
      for (int i = 0; val != 0 and i < sizeof(scratch_type); i++) {
```

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,282 @@
...
+        dst[lane_id * sizeof(scratch_type) + i] = val & 0xff;
+        val >>= 8;
+      }
+    }
+    __syncwarp();
```
vuule, line 90: why is this sync needed (there's a `__syncthreads` immediately after)?

etseidl: defensive sync...I need to check them all...when things go wrong I start throwing in syncs all over the place. This one can for sure go away.

cpp/src/io/parquet/delta_enc.cuh
etseidl: this only gets called if the bitwidth is 32 or 64. if that's the case, then there likely aren't many values that will have leading zeros. I'm also worried that the column buffer might not be initialized to 0...I'll have to check that.

cpp/src/io/parquet/page_enc.cu
```diff
@@ -239,6 +247,62 @@ struct BitwiseOr {
...
+template <typename I>
+struct delta_enc {
+  using output_type = std::conditional_t<std::is_signed_v<I>, zigzag128_t, uleb128_t>;
+
+  page_enc_state_s<0>* const s;
+  uint32_t const valid_count;
+  uint64_t* const buffer;
+  void* const temp_space;
+
+  __device__ thrust::pair<bool, size_type> calc_idx_and_validity(uint32_t cur_val_idx)
```
vuule, line 260: The order is wrong in the name :D
```suggestion
  __device__ thrust::pair<bool, size_type> calc_validity_and_idx(uint32_t cur_val_idx)
```

cpp/tests/io/parquet_test.cpp
```diff
@@ -6732,4 +6734,34 @@ TEST_P(ParquetV2Test, CheckEncodings)
...
+// removing duration_D, duration_s, and timestamp_s as they don't appear to be supported properly.
+// see definition of UnsupportedChronoTypes above.
+using DeltaDecimalTypes = cudf::test::Types<numeric::decimal32, numeric::decimal64>;
+using DeltaBinaryTypes =
+  cudf::test::Concat<cudf::test::IntegralTypesNotBool, cudf::test::ChronoTypes, DeltaDecimalTypes>;
+using SupportedDeltaTestTypes =
+  cudf::test::RemoveIf<cudf::test::ContainedIn<UnsupportedChronoTypes>, DeltaBinaryTypes>;
+TYPED_TEST_SUITE(ParquetWriterDeltaTest, SupportedDeltaTestTypes);
+
+TYPED_TEST(ParquetWriterDeltaTest, SupportedDeltaTestTypes)
```
vuule, line 6746: I think we need a test with a sliced table as well.

cpp/src/io/parquet/page_enc.cu
etseidl: nit, meet comb :rofl:

cpp/src/io/parquet/page_enc.cu
```diff
@@ -239,6 +247,62 @@ struct BitwiseOr {
...
+  {
+    size_type const val_idx_in_block    = cur_val_idx + threadIdx.x;
+    size_type const val_idx_in_leaf_col = s->page_start_val + val_idx_in_block;
+
+    uint32_t const is_valid = (val_idx_in_leaf_col < s->col.leaf_column->size() &&
```
vuule, line 265: why not bool?

etseidl: ah, just wait for it :rofl: I just changed it with the name :D

cpp/tests/io/parquet_test.cpp
etseidl: done

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1395,37 +1693,146 @@ __global__ void __launch_bounds__(128, 8)
...
+      case 8: {
+        // only DURATIONS map to 8 bytes, so safe to just use signed here?
+        delta_enc<int64_t> encoder{s, valid_count, delta_shared, &temp_storage};
+        delta_ptr = encoder.encode();
+        break;
+      }
+      case 4: {
+        if (type_id == type_id::UINT32) {
+          delta_enc<uint32_t> encoder{s, valid_count, delta_shared, &temp_storage};
+          delta_ptr = encoder.encode();
```
vuule, line 1795: hate to say this, but it seems like it would be cleaner to call a function here instead of creating the object. Should delta_enc be a function? Not sure.

vuule (commented):
Rest of the review (if this UI works as I assume).

cpp/tests/io/parquet_test.cpp
```diff
@@ -6732,4 +6734,34 @@ TEST_P(ParquetV2Test, CheckEncodings)
   }
 }
 
+// removing duration_D, duration_s, and timestamp_s as they don't appear to be supported properly.
+// see definition of UnsupportedChronoTypes above.
+using DeltaDecimalTypes = cudf::test::Types<numeric::decimal32, numeric::decimal64>;
+using DeltaBinaryTypes =
+  cudf::test::Concat<cudf::test::IntegralTypesNotBool, cudf::test::ChronoTypes, DeltaDecimalTypes>;
+using SupportedDeltaTestTypes =
```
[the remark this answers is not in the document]
etseidl: Not explicitly for delta encoding. I was more concerned about making sure all data types that map to INT32 or INT64 were covered correctly. I'll see what I can whip up.

cpp/src/io/parquet/page_enc.cu
```diff
@@ -791,30 +905,35 @@ inline __device__ void PackLiterals(
  * @param[in] flush nonzero if last batch in block
  * @param[in] t thread id (0..127)
  */
+template <typename state_buf>
```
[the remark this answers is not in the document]
etseidl: we don't

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1124,29 +1237,122 @@ __global__ void __launch_bounds__(128, 8)
...
+  size_t const skip_comp_size =
+    write_v2_headers ? s->page.def_lvl_bytes + s->page.rep_lvl_bytes : 0;
+
+  if (t == 0) {
+    // only need num_nulls for v2 data page headers
+    if (write_v2_headers) { s->page.num_nulls = s->page.num_values - valid_count; }
+    uint8_t const* const base   = s->page.page_data + s->page.max_hdr_size;
+    auto const actual_data_size = static_cast<uint32_t>(end_ptr - base);
+    if (actual_data_size > s->page.max_data_size) {
+      printf("data corruption %d %d\n", actual_data_size, s->page.max_data_size);
```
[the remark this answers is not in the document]
etseidl: this is another place where we need to return an error code to the host. right now it just silently overwrites and we're left scratching our heads when things go wrong.

cpp/src/io/parquet/page_enc.cu
etseidl: yeah, I'm not super happy with this. I was hoping you'd have some clever dispatcher magic or something.

vuule: we do have dispacth magic in libcudf, but it doesn't apply her that well. I'll take another look, but I don't think we can avoid having a switch somewhere.

etseidl: ok, that's a little better.

vuule (approved):
Looks great, thank you for addressing the comments!

GregoryKimball:
Hello @karthikeyann, would you please take a look over this contribution?

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,290 @@
...
+#include <cub/cub.cuh>
+
+namespace cudf::io::parquet::detail {
+
+namespace delta {
+
+inline __device__ void put_uleb128(uint8_t*& p, uleb128_t v)
+{
+  while (v > 0x7f) {
+    *p++ = v | 0x80;
```
ttnghia, line 33: Typically I hate this style. It is very confusing. I have to scratch my head to identify what operator is called first? 

Please put in parenthesis so anybody can understand it easily.

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,290 @@
...
+inline __device__ void put_zz128(uint8_t*& p, zigzag128_t v)
+{
+  zigzag128_t s = (v < 0);
+  put_uleb128(p, (v ^ -s) * 2 + s);
+}
+
+// a block size of 128, with 4 mini-blocks of 32 values each fits nicely without consuming
+// too much shared memory.
+// the parquet spec requires block_size to be a multiple of 128, and values_per_mini_block
+// to be a multiple of 32.
```
ttnghia, line 48: Nit: It is much cleaner to capitalize the first letter in any sentence.

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,290 @@
...
+constexpr int block_size            = 128;
+constexpr int num_mini_blocks       = 4;
+constexpr int values_per_mini_block = block_size / num_mini_blocks;
+constexpr int buffer_size           = 2 * block_size;
+
+// extra sanity checks to enforce compliance with the parquet specification
```
ttnghia, line 54: Nit: It is cleaner to have any sentence (including comments) begins with a capitalized letter and ends with a period. Sorry if I'm nick picking too much.

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -0,0 +1,290 @@
...
+// T is the input data type (either zigzag128_t or uleb128_t)
+template <typename T>
+class delta_binary_packer {
+ private:
+  uint8_t* _dst;                             // sink to dump encoded values to
+  size_type _current_idx;                    // index of first value in buffer
+  uint32_t _num_values;                      // total number of values to encode
+  size_type _values_in_buffer;               // current number of values stored in _buffer
+  T* _buffer;                                // buffer to store values to be encoded
+  uint8_t _mb_bits[delta::num_mini_blocks];  // bitwidth for each mini-block
```
ttnghia, line 138: Can you group pointers and values please?

cpp/src/io/parquet/delta_enc.cuh
etseidl: No worries...hitting caps is an extra keystroke and I'm lazy :sweat_smile:

ttnghia (commented):
LGTM. I just left a few comments about coding style.

ttnghia:
The tests here seem to do round trip write/read. Can we have file written somewhere else that can be tested here? I'm not sure if we can do that in C++ code or have to do it in python?

etseidl:
> The tests here seem to do round trip write/read. Can we have file written somewhere else that can be tested here? I'm not sure if we can do that in C++ code or have to do it in python?

So the inverse of `test_delta_binary` in `test_parquet.py`, which writes with pandas and reads with cudf? So write with cudf and read with pyarrow or pandas? The issue is we don't yet expose the options necessary to write V2 with python. Can we use the arrow compatibility stuff? So on the C++ side, write with cudf, then read with arrow-cpp, convert arrow to cudf and compare?

vuule:
> > The tests here seem to do round trip write/read. Can we have file written somewhere else that can be tested here? I'm not sure if we can do that in C++ code or have to do it in python?
> 
> So the inverse of `test_delta_binary` in `test_parquet.py`, which writes with pandas and reads with cudf? So write with cudf and read with pyarrow or pandas? The issue is we don't yet expose the options necessary to write V2 with python. Can we use the arrow compatibility stuff? So on the C++ side, write with cudf, then read with arrow-cpp, convert arrow to cudf and compare?

FWIW, we do have reader tests for this encoding that does not depend on the cudf writer. I guess it does not prove that reader only reads correctly encoded data, so writer could write something only cudf can read. Might be best to add the V2 API to Python...

ttnghia:
> Might be best to add the V2 API to Python...

Agree. The test can be added here or either in a separate PR.

ttnghia (approved):

PR #11657, opened 2022-09-06
Revert removal of skip_rows / num_rows options from the Parquet reader.

nvdbaranec:
Reverts:

https://github.com/rapidsai/cudf/pull/11480
https://github.com/rapidsai/cudf/pull/11503

PR #11480, opened 2022-08-05
[REVIEW] Drop support for `skiprows` and `num_rows` in `cudf.read_parquet`

galipremsagar:
## Description
This PR removes support for `skiprows` & `num_rows` in parquet reader. A continuation of https://github.com/rapidsai/cudf/pull/11218

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #14101, opened 2023-09-13
Add decoder for DELTA_BYTE_ARRAY to Parquet reader

etseidl:
## Description
Part of #13501. Adds ability to decode DELTA_BYTE_ARRAY encoded pages.

## Checklist


etseidl:
This one cannot be tested with python in same way the DELTA_BINARY_PACKED decoder is because pyarrow doesn't yet support writing DELTA_BYTE_ARRAY (as of version 13.0). The only option at this point is to add yet another small parquet file to the python test data directory, but that's not a very thorough test.

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -1286,6 +1286,11 @@ def test_parquet_reader_v2(tmpdir, simple_pdf):
     assert_eq(cudf.read_parquet(pdf_fname), simple_pdf)
 
 
+def test_parquet_delta_byte_array(datadir):
```
vuule, line 1289: Should we have tests with skip_rows and/or nested columns? I saw some logic related to this in the kernels.

vuule (commented):
Great stuff (again :P)!
Bunch of small suggestions and questions, I think tests might be the biggest topic.

vuule (commented):
Great stuff (again :P)!
Bunch of small suggestions and questions, I think tests might be the biggest topic.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -748,27 +953,93 @@ __global__ void __launch_bounds__(decode_block_size)
...
+    auto const page_sizes = [] __device__(PageInfo const& page) { return page.temp_string_size; };
+    int64_t total_size    = thrust::transform_reduce(rmm::exec_policy(stream),
+                                                  pages.d_begin(),
+                                                  pages.d_end(),
+                                                  page_sizes,
+                                                  0L,
+                                                  thrust::plus<int64_t>{});
+
+    rmm::device_uvector<int64_t> page_string_offsets(pages.size(), stream);
+    thrust::transform_exclusive_scan(rmm::exec_policy(stream),
```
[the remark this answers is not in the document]
etseidl: TIL `exec_policy_nosync`. Might be able to use this in `calculate_page_string_offsets()` too.  That one ends with a memcpy followed by a sync.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -30,10 +29,15 @@ namespace cudf::io::parquet::detail {
...
-  auto& page_nesting_decode = _pass_itm_data->page_nesting_decode_info;
+  auto& chunks               = _pass_itm_data->chunks;
+  auto& pages                = _pass_itm_data->pages_info;
+  auto& page_nesting         = _pass_itm_data->page_nesting_info;
+  auto& page_nesting_decode  = _pass_itm_data->page_nesting_decode_info;
+  auto const level_type_size = _pass_itm_data->level_type_size;
+
+  // temporary space for DELTA_BYTE_ARRAY decoding. this only needs to live until
+  // gpu::DecodeDeltaByteArray returns.
+  rmm::device_buffer delta_temp_buf;
```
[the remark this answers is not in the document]
etseidl: I'll give it a whirl.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -748,27 +953,93 @@ __global__ void __launch_bounds__(decode_block_size)
   if (t == 0 and s->error != 0) { set_error(s->error, error_code); }
 }
 
+struct page_tform_functor {
+  void* const data;
+
+  __device__ PageInfo operator()(PageInfo& page, int64_t offset)
+  {
+    if (page.temp_string_size != 0) { page.temp_string_buf = static_cast<uint8_t*>(data) + offset; }
```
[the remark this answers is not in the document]
etseidl: Yes, it's initialized when the initial page header parsing is done.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -163,6 +168,7 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
   chunks.host_to_device_async(_stream);
   chunk_nested_valids.host_to_device_async(_stream);
   chunk_nested_data.host_to_device_async(_stream);
+  if (has_strings) { chunk_nested_str_data.host_to_device_async(_stream); }
```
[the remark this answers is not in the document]
etseidl: Well, the vector has size 0 if `has_strings` is false...it's asynch so I suppose a 0 length memcpy won't hurt.

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -291,7 +308,11 @@ struct PageInfo {
   // level decode buffers
   uint8_t* lvl_decode_buf[level_type::NUM_LEVEL_TYPES];
 
-  uint32_t kernel_mask;
+  // temporary space for decoding DELTA_BYTE_ARRAY encoded strings
+  int64_t temp_string_size;
+  uint8_t* temp_string_buf;
```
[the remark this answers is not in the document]
etseidl: The size is set pretty far from where the pointer is initialized, I'm not sure if device_span will work here.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -450,12 +458,110 @@ __device__ size_t totalPlainEntriesSize(uint8_t const* data,
...
+  // now sum up total_bytes from the two warps
+  auto const final_bytes =
+    cudf::detail::single_lane_block_sum_reduce<delta_preproc_block_size, 0>(total_bytes);
+
+  // sum up prefix and suffix max lengths
+  auto temp_bytes =
+    cudf::detail::single_lane_block_sum_reduce<delta_preproc_block_size, 0>(max_len);
+  if (t == 0) {
+    // save enough for one mimi-block plus some extra to save the last_string
+    temp_bytes *= db->values_per_mb + 1;
```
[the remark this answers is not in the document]
etseidl: No, it's what it should be.  I rewrote it to make it clearer what's going on (will fix formatting later).

cpp/src/io/parquet/page_delta_decode.cu
```diff
@@ -154,6 +424,154 @@ __global__ void __launch_bounds__(96)
...
+
+  auto const mask = decode_kernel_mask::DELTA_BYTE_ARRAY;
+  if (!setupLocalPageInfo(
+        s, &pages[page_idx], chunks, min_row, num_rows, mask_filter{mask}, true)) {
+    return;
+  }
+
+  bool const has_repetition = s->col.max_level[level_type::REPETITION] > 0;
+
+  // TODO(ets) assert string_data != nullptr
```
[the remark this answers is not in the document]
etseidl: I think that's a leftover...but the TODO a little lower should set the error code.

cpp/src/io/parquet/page_delta_decode.cu
```diff
@@ -27,6 +27,280 @@ namespace cudf::io::parquet::detail {
...
+
+  // dump strings before start_val to temp buf
+  __device__ void skip(bool use_char_ll)
+  {
+    using cudf::detail::warp_size;
+    int const t       = threadIdx.x;
+    int const lane_id = t % warp_size;
+
+    // is this even necessary? return if asking to skip the whole block.
+    if (start_val >= prefixes.num_encoded_values(true)) { return; }
```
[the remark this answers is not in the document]
etseidl: It's probably safe to remove. There's already a test earlier in the kernel that will return if there are no in-range values in the page. When we add a skip_rows test we can remove it.

python/cudf/cudf/tests/test_parquet.py
etseidl: So there's a can of worms. When I added the V2 headers, I modified a few roundtrip tests to run with both V1 and V2 headers. Now as we add more encodings, I think there really needs to be a lot more tests that run over a range of parameters, more like nvbench. But parquet already has the longest test to run and compile. Maybe this should be brought up for discussion elsewhere. 

Also, we can't use python to test because arrow doesn't support writing DELTA_BYTE_ARRAY yet (coming in version 14 IIRC), and we don't have an encoder ready yet either. I don't know if we can add a parquet file big enough to test all the weird edge cases.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -748,27 +953,93 @@ __global__ void __launch_bounds__(decode_block_size)
   if (t == 0 and s->error != 0) { set_error(s->error, error_code); }
 }
 
+struct page_tform_functor {
```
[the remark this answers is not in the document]
etseidl: done

cpp/src/io/parquet/reader_impl.cpp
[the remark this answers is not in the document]
etseidl: Yay, that was better.  I can never remember which containers have `resize`.

cpp/src/io/parquet/reader_impl.cpp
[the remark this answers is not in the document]
vuule: I think we can leave this for now; as a future work, these functions should guarantee a no-op when empty.

python/cudf/cudf/tests/test_parquet.py
vuule: Agreed.
I would be okay with doing what we can with one test file and opening an issue for test coverage of new features.
My suggestion would be to have a test file that is a list of strings, and read it with and without skip_rows (and any other relevant option which I may be missing).

etseidl: Except #11480 :(

GregoryKimball:
@nvdbaranec would you please review this PR with the recent changes from @etseidl?

vuule (approved):
Unfortunate testing situation, but we can improve it over time as requirements fall into place.

etseidl:
Added some more tests now that we've moved to arrow 14.  Should we think about reverting #11480 so we can test skip rows, or just wait for the encoder? @vuule

cpp/src/io/parquet/page_data.cu
```diff
@@ -449,8 +449,9 @@ __global__ void __launch_bounds__(decode_block_size)
   int out_thread0;
   [[maybe_unused]] null_count_back_copier _{s, t};
 
+  auto const mask = decode_kernel_mask::GENERAL;
   if (!setupLocalPageInfo(
-        s, &pages[page_idx], chunks, min_row, num_rows, mask_filter{KERNEL_MASK_GENERAL}, true)) {
+        s, &pages[page_idx], chunks, min_row, num_rows, mask_filter{mask}, true)) {
```
nvdbaranec, line 454: I have a mild preference for just 
`mask_filter{decode_kernel_mask::GENERAL}`
so your eyes don't have to go hunting for where mask is declared.

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -675,6 +701,28 @@ void DecodeDeltaBinary(cudf::detail::hostdevice_vector<PageInfo>& pages,
...
+ * @param[in] num_rows Total number of rows to read
+ * @param[in] min_row Minimum number of rows to read
+ * @param[in] level_type_size Size in bytes of the type for level decoding
+ * @param[out] error_code Error code for kernel failures
+ * @param[in] stream CUDA stream to use
+ */
+void DecodeDeltaByteArray(cudf::detail::hostdevice_vector<PageInfo>& pages,
+                          cudf::detail::hostdevice_vector<ColumnChunkDesc> const& chunks,
+                          size_t num_rows,
+                          size_t min_row,
```
nvdbaranec, line 721: To be consistent with other places:
```suggestion
                          size_t min_row,
                          size_t num_rows,
```

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -748,27 +951,101 @@ __global__ void __launch_bounds__(decode_block_size)
...
+  // synchronize the streams
+  cudf::detail::join_streams(streams, stream);
+
+  // check for needed temp space for DELTA_BYTE_ARRAY
+  auto const need_sizes = thrust::any_of(
+    rmm::exec_policy(stream), pages.d_begin(), pages.d_end(), [] __device__(auto& page) {
+      return page.temp_string_size != 0;
+    });
+
+  if (need_sizes) {
```
nvdbaranec, line 1017: Just to verify: page.temp_string_size > 0 only true if we have pages that need the DELTA_BYTE_ARRAY decoder?

cpp/src/io/parquet/page_delta_decode.cu
```diff
@@ -27,6 +27,280 @@ namespace cudf::io::parquet::detail {
...
 namespace {
 
+constexpr int decode_block_size = 128;
+
+// DELTA_BYTE_ARRAY encoding (incremental encoding or front compression), is used for BYTE_ARRAY
+// columns. For each element in a sequence of strings, a prefix length from the preceding string
+// and a suffix is stored. The prefix lengths are DELTA_BINARY_PACKED encoded. The suffixes are
+// encoded with DELTA_LENGTH_BYTE_ARRAY encoding, which is a DELTA_BINARY_PACKED list of suffix
+// lengths, followed by the concatenated suffix data.
+struct delta_byte_array_decoder {
```
nvdbaranec, line 37: How big is this struct? It's declared as shared and it seems pretty heavy. Looks like maybe on the order of 2k?

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -748,27 +951,101 @@ __global__ void __launch_bounds__(decode_block_size)
...
+    // sum up all of the temp_string_sizes
+    auto const page_sizes = [] __device__(PageInfo const& page) { return page.temp_string_size; };
+    int64_t total_size    = thrust::transform_reduce(rmm::exec_policy(stream),
```
nvdbaranec, line 1020:
```suggestion
    int64_t const total_size    = thrust::transform_reduce(rmm::exec_policy(stream),
```

cpp/src/io/parquet/delta_binary.cuh
```diff
@@ -179,6 +173,28 @@ struct delta_binary_decoder {
     }
   }
 
+  // given start/end pointers in the data, find the end of the binary encoded block. when done,
+  // `this` will be initialized with the correct start and end positions. returns the end, which is
+  // start of data/next block. should only be called from thread 0.
+  inline __device__ uint8_t const* find_end_of_block(uint8_t const* start, uint8_t const* end)
```
nvdbaranec, line 179: Out of curiosity, how hard would parallelizing this be, even at the warp level?  Is this a similar thing to the level data where you can't easily do it because you can't know where position N+1 is without processing position N? It seems to do a decent amount of work and it shows up in a number of places.  I don't have a good sense how many mini blocks you might see in a typical page.

cpp/src/io/parquet/page_delta_decode.cu
```diff
@@ -154,6 +424,159 @@ __global__ void __launch_bounds__(96)
...
 
+// Decode page data that is DELTA_BYTE_ARRAY packed. This encoding consists of a DELTA_BINARY_PACKED
+// array of prefix lengths, followed by a DELTA_BINARY_PACKED array of suffix lengths, followed by
+// the suffixes (technically the suffixes are DELTA_LENGTH_BYTE_ARRAY encoded). The latter two can
+// be used to create an offsets array for the suffix data, but then this needs to be combined with
+// the prefix lengths to do the final decode for each value. Because the lengths of the prefixes and
+// suffixes are not encoded in the header, we're going to have to first do a quick pass through them
+// to find the start/end of each structure.
+template <typename level_t>
+__global__ void __launch_bounds__(decode_block_size)
```
nvdbaranec, line 435: What is the relationship between this and the code in page_string_decode.cu?  Is DELTA_BYTE_ARRAY always treated as strings and then the code in page_string_decode.cu is just for the preprocess/size computation?

cpp/src/io/parquet/parquet_gpu.hpp
etseidl: It's actually consistent with all the other `DecodeXXX` calls, but the `ComputePageXXXSizes` swap them.

cpp/src/io/parquet/page_string_decode.cu
etseidl: Yes, but also only if rows are being skipped for the given page.

cpp/src/io/parquet/delta_binary.cuh
etseidl: Sadly this is just like trying to count PLAIN encoded string sizes, or as you point out trying to find where runs start in the RLE decoder.  The header for each block has a varint in it, and then each of the miniblocks can have its own encoding bitwidth. So you have to traverse the entire thing to find the end of current section/start of the next. Other implementations usually just allocate an array to hold all the values, since they're only dealing with one page per column chunk at a time.

Pages these days are usually capped at 20,000 rows (at least cudf and parquet-mr), with 128 values per block, so on the order of 157 block headers to parse and skip per page.  Of course this can get pretty big if lists are involved.

cpp/src/io/parquet/page_delta_decode.cu
etseidl: DELTA_BYTE_ARRAY is just another encoding for strings, joining PLAIN and DICTIONARY, and soon to be joined by DELTA_LENGTH_BYTE_ARRAY.  So yes, the code in page_string_decode.cu is preprocessing that's to some extent common to all the string decoders (which is why some DELTA stuff wound up in there rather than all living here).

cpp/src/io/parquet/page_data.cu
etseidl: I was trying to keep all the args on one line 😭 I'll change it.

cpp/src/io/parquet/page_delta_decode.cu
etseidl: Yep, a little bigger than 2k. I can give you exact numbers for shared mem use per kernel tomorrow, but yes, this one gets pretty herfty.

vuule:
> Added some more tests now that we've moved to arrow 14. Should we think about reverting #11480 so we can test skip rows, or just wait for the encoder? @vuule

We probably should, since we already maintain the whole feature in libcudf. I'll let you know after an internal discussion.

cpp/src/io/parquet/page_delta_decode.cu
```diff
@@ -27,6 +27,277 @@ namespace cudf::io::parquet::detail {
...
+    // let p(n) === length(prefix(string_n))
+    //
+    // if p(n-1) > p(n), then string_n can be completed when string_n-2 is completed. likewise if
+    // p(m) > p(n), then string_n can be completed with string_m-1. however, if p(m) < p(n), then m
+    // is a "blocker" for string_n; string_n can be completed only after string_m is.
+    //
+    // we will calculate the nearest blocking position for each lane, and then fill in string_0. we
+    // then iterate, finding all lanes that have had their "blocker" filled in and completing them.
+    // when all lanes are filled in, we return. this will still hit the worst case if p(n-1) < p(n)
+    // for all n
```
nvdbaranec, line 80: I appreciate all the detailed comments in this PR, btw.

cpp/src/io/parquet/page_delta_decode.cu
```diff
@@ -154,6 +410,159 @@ __global__ void __launch_bounds__(96)
...
+  while (!s->error && (s->input_value_count < s->num_input_values || s->src_pos < s->nz_count)) {
+    uint32_t target_pos;
+    uint32_t const src_pos = s->src_pos;
+
+    if (t < 3 * warp_size) {  // warp 0..2
+      target_pos = min(src_pos + 2 * batch_size, s->nz_count + s->first_row + batch_size);
+    } else {  // warp 3
+      target_pos = min(s->nz_count, src_pos + batch_size);
+    }
+    __syncthreads();
```
nvdbaranec, line 498: Is this sync necessary? There's one already at the bottom of the loop and there's nothing here that seems to require synchronization.   I just looked at the related loop in page_data.cu and we're doing it there as well.  Hmm. This feels very vestigial.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -450,12 +458,108 @@ __device__ size_t totalPlainEntriesSize(uint8_t const* data,
...
+  }
+  __syncthreads();
+
+  // now sum up total_bytes from the two warps
+  auto const final_bytes =
+    cudf::detail::single_lane_block_sum_reduce<delta_preproc_block_size, 0>(total_bytes);
+
+  // Sum up prefix and suffix max lengths to get a max possible string length. Multiply that
+  // by the number of strings in a mini-block, plus one to save the last string.
+  auto temp_bytes =
```
nvdbaranec, line 548:
```suggestion
  auto const temp_bytes =
```

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -450,12 +458,108 @@ __device__ size_t totalPlainEntriesSize(uint8_t const* data,
...
+        }
+      }
+
+      if (lane_id == 0) { db->setup_next_mini_block(true); }
+      __syncwarp();
+    }
+
+    // get sum for warp.
+    // note: warp_sum will only be valid on lane 0.
+    using cudf::detail::warp_size;
```
nvdbaranec, line 531: Unused?
```suggestion
```

cpp/src/io/parquet/page_delta_decode.cu
etseidl: Never thought about it, but I think you're right. Maybe back in the day target_pos was shared or something. I'll take a look at all the decoders.

nvdbaranec: Actually, let's leave them in for now and remove them in a separate PR.  I'd rather let it soak for a full milestone (until 24.02) to make sure there's not something subtle here that we're missing.

etseidl: Ok I'll put them back

etseidl: done...and added some TODOs

nvdbaranec (approved):

python/cudf/cudf/tests/test_parquet.py
galipremsagar: We actually reverted the `skip_rows` here: https://github.com/rapidsai/cudf/pull/11657 and we now have arrow-14 support, would it make sense for now to add the tests?

galipremsagar (approved):
Looks good, left an optional comment here: https://github.com/rapidsai/cudf/pull/14101/files#r1396338749

python/cudf/cudf/tests/test_parquet.py
etseidl: Thanks @galipremsagar! How did I miss _that_??? 

@vuule Do you think we have time to add skip rows tests now or just wait for DELTA_LENGTH_BYTE_ARRAY.

etseidl: > We actually reverted the skip_rows here: https://github.com/rapidsai/cudf/pull/11657

Oh, that only reverted the C++ side. The num_rows/skip_rows options are still missing from the python side of the house, which is what I have to use to generate the test files right now.

Another option is to use arrow-cpp for write and libcudf for read in the C++ tests.

FWIW I'm working on the DELTA_LENGTH_BYTE_ARRAY encoder/decoder right now, and will be able to test skip rows directly on the C++ side.

galipremsagar: Ah, my bad. Scratch that. Sorry for the confusion @etseidl ! This looks good to me then.

PR #14107, opened 2023-09-13
Some additional kernel thread index refactoring.

bdice:
## Description
This PR refactors a few kernels to use `thread_index_type` and associated utilities. I started this before realizing how much scope was still left in issue #10368 ("Part 2 - Take another pass over more challenging kernels"), and then I stopped working on this due to time constraints. For the moment, I hope this PR makes a small dent in the number of remaining kernels to convert to using `thread_index_type`.

## Checklist


cpp/include/cudf/detail/copy_if_else.cuh
bdice, line 1: This is the refactor I'm least sure about -- please give it a close eye. I feel like I might have a bug here, or that the kernel can be further simplified, but can't decide exactly what to do or what's in scope for this change.

cpp/include/cudf/detail/copy_if_else.cuh
```diff
@@ -44,18 +44,19 @@ __launch_bounds__(block_size) __global__
                            mutable_column_device_view out,
                            size_type* __restrict__ const valid_count)
 {
-  size_type const tid            = threadIdx.x + blockIdx.x * block_size;
-  int const warp_id              = tid / warp_size;
+  auto tidx                      = cudf::detail::grid_1d::global_thread_id();
```
harrism, line 47: **💡 suggestion:** ‏ This change could have an effect on performance. You should probably benchmark. Templating the block size rather than using blockDim.x is a classical kernel micro-optimization. It typically eliminates a register and enables some other compiler optimizations. Very valid when a kernel always uses the same block size.

You could add a template utility for this:

```
template <thread_index_type block_size>
grid_1d::global_thread_id() {
  return threadIdx.x + blockIdx.x * block_size;
}
```

cpp/include/cudf/detail/copy_if_else.cuh
```diff
@@ -44,18 +44,19 @@ __launch_bounds__(block_size) __global__
...
+  auto const stride              = cudf::detail::grid_1d::grid_stride();
+  int const warp_id              = tidx / warp_size;
```
harrism, line 49: **💡 suggestion:** ‏ Use `cudf::detail::warp_size`. CUDA's built-in `warp_size` is not a compile-time constant. (The reason is forever a mystery to me). Therefore we created a `constexpr` in `detail/utilities/cuda.cuh`. Could save multiple divisions here. Not sure how this slipped by in the past -- the constexpr is used in most cudf kernels.

cpp/include/cudf/detail/copy_if_else.cuh
```diff
@@ -44,18 +44,19 @@ __launch_bounds__(block_size) __global__
...
-  size_type end   = out.size();
+  size_type const begin = 0;
+  size_type const end   = out.size();
   // warp indices.  since 1 warp == 32 threads == sizeof(bitmask_type) * 8,
   // each warp will process one (32 bit) of the validity mask via
   // __ballot_sync()
-  size_type warp_begin = cudf::word_index(begin);
-  size_type warp_end   = cudf::word_index(end - 1);
+  size_type const warp_begin = cudf::word_index(begin);
+  size_type const warp_end   = cudf::word_index(end - 1);
```
harrism, line 59: **❓ question:** ‏ Should any of these use `thread_index_type` instead of `size_type`?

cpp/include/cudf/detail/copy_if_else.cuh
harrism: **❓ question:** ‏ What sort of bug are you feeling? I don't see anything obvious, just optimization opportunities.

harrism (changes requested):
Thanks for the fixups!

cpp/include/cudf/detail/copy_if_else.cuh
mythrocks: > Could save multiple divisions here. 

Sorry, I'm not sure I followed how multiple divisions might be saved here.  `tidx` isn't a `constexpr`, is it?

cpp/include/cudf/detail/copy_if_else.cuh
mythrocks: This is a question that I had as well.  I was wondering if we should convert all these to `thread_index_type`.

I suppose `warp_curr` is less likely to overflow at line#87, so maybe it's fine as a `size_type`?

cpp/include/cudf/detail/copy_if_else.cuh
harrism: If the compiler knows that warp_size is 32, it will use a SHR (right shift) instead of a DIV.

mythrocks: Wow, that's brilliant.  Noted.

harrism: Should have said this is the case for any power of 2.

cpp/benchmarks/join/generate_input_tables.cuh
```diff
@@ -45,13 +46,14 @@ __global__ static void init_build_tbl(key_type* const build_tbl,
...
-  auto const start_idx = blockIdx.x * blockDim.x + threadIdx.x;
-  auto const stride    = blockDim.x * gridDim.x;
+  auto const start_idx = cudf::detail::grid_1d::global_thread_id();
+  auto const stride    = cudf::detail::grid_1d::grid_stride();
   assert(start_idx < num_states);
 
   curandState localState = state[start_idx];
 
-  for (size_type idx = start_idx; idx < build_tbl_size; idx += stride) {
+  for (thread_index_type tidx = start_idx; tidx < build_tbl_size; tidx += stride) {
```
ttnghia, line 55: Should we add a range-based for loop to replace this pattern?
```
for(auto const tidx: grid_1d::thread_indices()) {...}
```
So we won't need to have `stride` like this.

cpp/include/cudf/detail/copy_if_else.cuh
bdice: If you trace the types, the value of `warp_cur` in `while (warp_cur <= warp_end) {` is the one to watch. Theoretically this value could be `thread_index_type`. However, it's compared to `warp_end` which is `size_type` (its maximum value is `std::numeric_limits<cudf::size_type>::max() / cudf::detail::warp_size`). So really I don't think we need to do anything here, otherwise we'll be casting up some computations "for safety reasons" only to unsafely downcast in a comparison.

cpp/benchmarks/join/generate_input_tables.cuh
bdice: For now I'm going to leave that out of scope for this PR. Please feel free to comment on https://github.com/rapidsai/cudf/issues/10368 -- the topic of a range-based loop has already come up there.

cpp/include/cudf/detail/copy_if_else.cuh
bdice: Done in 9e4154dd176c677ad7168c9f0770a7e2363c6970.

cpp/include/cudf/detail/copy_if_else.cuh
bdice: Done in 9e4154dd176c677ad7168c9f0770a7e2363c6970.

mythrocks (approved):
LGTM!

harrism (approved):
Looks good!

davidwendt (approved):

bdice:
I pushed a couple small fixes for benchmark builds (I hadn't run that locally yet). This should be ready to merge after CI passes.

PR #14108, opened 2023-09-13
Correct numerous 20054-D: dynamic initialization errors found on arm+12.2

robertmaynard:
## Description
Compile issues found by compiling libcudf with the `rapidsai/devcontainers:23.10-cpp-gcc9-cuda12.2-ubuntu20.04` docker container. 

## Checklist


bdice:
@robertmaynard Do you think these are "real issues" that we should fix for C++ standards compliance (like avoiding UB) or are they potentially compiler bugs? Do you see these issues with newer GCC versions or older CUDA versions on ARM?

cpp/src/io/comp/unsnap.cu
```diff
@@ -64,15 +66,17 @@ struct unsnap_queue_s {
...
-  device_span<uint8_t const> src;  ///< input for current block
-  device_span<uint8_t> dst;        ///< output for current block
+  constexpr unsnap_state_s() noexcept {}  // required to compile on ctk-12.2 + aarch64
+
+  uint8_t const* base        = nullptr;   ///< base ptr of compressed stream
+  uint8_t const* end         = nullptr;   ///< end of compressed stream
+  uint32_t uncompressed_size = 0;         ///< uncompressed stream size
+  uint32_t bytes_left        = 0;         ///< remaining bytes to decompress
+  int32_t error              = 0;         ///< current error status
+  uint32_t tstart            = 0;         ///< start time for perf logging
```
davidwendt, line 76:
```suggestion
  uint8_t const* base{};         ///< base ptr of compressed stream
  uint8_t const* end{};          ///< end of compressed stream
  uint32_t uncompressed_size{};  ///< uncompressed stream size
  uint32_t bytes_left{};         ///< remaining bytes to decompress
  int32_t error{};               ///< current error status
  uint32_t tstart{};             ///< start time for perf logging
```
These can all be initialized the same way.

robertmaynard:
> @robertmaynard Do you think these are "real issues" that we should fix for C++ standards compliance (like avoiding UB) or are they potentially compiler bugs? Do you see these issues with newer GCC versions or older CUDA versions on ARM?


These are all real issues that aren't exposed by the x86 compiler. I am still digging into why, but here is the background on why these failures are valid:


From the CUDA programming guide `14.5.3.1. Device Memory Space Specifiers`:

A `__device__`, `__constant__`, `__managed__` or `__shared__` variable definition cannot have a class type with a non-empty constructor or a non-empty destructor. A constructor for a class type is considered empty at a point in the translation unit, if it is either a trivial constructor or it satisfies all of the following conditions:

  - The constructor function has been defined.
  - The constructor function has no parameters, the initializer list is empty and the function body is an empty compound statement.
  - Its class has no virtual functions, no virtual base classes and no non-static data member initializers.
  - The default constructors of all base classes of its class can be considered empty.
  - For all the nonstatic data members of its class that are of class type (or array thereof), the default constructors can be considered empty.


The simpliest way to understand the above is that the `__shared__` type must pass the `std::is_trivially_constructible` check, if it doesn't it fails to meet the requirements and could generate the warning. As an example if we look at the `block_desc_s` specification we can see that it fails to meet the is_trivially_constructible requirements ( https://godbolt.org/z/75rv9jd3s )

```cpp
struct block_desc_s {
  block_desc_s() {};

  explicit constexpr block_desc_s(size_t offset_, uint32_t size_,
                                  uint32_t row_offset_, uint32_t first_row_,
                                  uint32_t num_rows_)
      : offset(offset_), size(size_), row_offset(row_offset_),
        first_row(first_row_), num_rows(num_rows_) {}

  size_t offset;
  uint32_t size;
  uint32_t row_offset;
  uint32_t first_row;
  uint32_t num_rows;
};
```

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -228,7 +228,7 @@ struct PageInfo {
  * @brief Struct describing a particular chunk of column data
  */
 struct ColumnChunkDesc {
-  ColumnChunkDesc() = default;
+  constexpr ColumnChunkDesc() noexcept {};
```
karthikeyann, line 231: does `default` also not meet `14.5.3.1. Device Memory Space Specifiers` specifications?
is `default` not considered as trivial constructor if all non-static members have trivial constructors?

I see usage of `default` in this same PR.

robertmaynard: I need the constexpr constructor for the `byte_stream_s` type in page_hdr.cu. I agree that on initial read the `= default` constructor should be sufficient, but the compiler disagrees.

harrism (approved):

davidwendt (approved):

bdice (approved):

vyasr (approved):

Issue #14088, opened 2023-09-12
[BUG] `Columns` name not being preserved in `DataFrame` constructor


galipremsagar:
**Describe the bug**
When an index object is passed to `columns` parameter in `DataFrame` constructor, the column name is not being preserved. I know there are some design issues are not being able to assign column name to column objects, but this is probably a bug.

**Steps/Code to reproduce bug**
```python

In [1]: import cudf

In [2]: df = cudf.DataFrame(columns=cudf.Index(['a'], name='n'))
df

In [3]: df
Out[3]: 
Empty DataFrame
Columns: [a]
Index: []

In [4]: df.columns
Out[4]: Index(['a'], dtype='object')

In [5]: import pandas as pd

In [6]: pdf = pd.DataFrame(columns=pd.Index(['a'], name='n'))

In [7]: pdf.columns
Out[7]: Index(['a'], dtype='object', name='n')

In [8]: df = cudf.DataFrame(columns=pd.Index(['a'], name='n'))

In [9]: df.columns
Out[9]: Index(['a'], dtype='object')
```
**Expected behavior**
```python
In [7]: df.columns
Out[7]: Index(['a'], dtype='object', name='n')
```

**Environment overview (please complete the following information)**
 - Environment location: [Bare-metal]
 - Method of cuDF install: [from source]

PR #14110, opened 2023-09-13
Preserve name of the column while initializing a `DataFrame`

galipremsagar:
## Description
Fixes: #14088 

This PR preserves `names` of `column` object while constructing a `DataFrame` through various constructor flows.

## Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -712,6 +715,11 @@ def __init__(
 
                     self._data = new_df._data
                     self._index = new_df._index
+                    self._data._level_names = (
```
bdice, line 718: Do we need some kind of helper to normalize this? I see this repeated a lot of times.

shwina: While the helper does reduce the amount of code somewhat, I think it reduces the overall comprehensibility sadly :(  I also find it a bit out-of-place for this helper to live in `column_accessor.py`.

I think either we should:

- Rethink the helper's name and API
- Go back to not having a helper (more explicit but also more code)

@bdice would you concur?

bdice: I was on the fence about whether a helper was appropriate here. I feel pretty neutral about the implementation in 88ccc48 -- I would accept the PR with or without the helper, since the DRY benefit and benefit of explicit logic are about even to me. @shwina If you think we should revert that change, I'm fine with that. I don't have any clear improvements to propose for names or APIs of the helper.

bdice (approved):

shwina (approved):

galipremsagar:
Thanks @shwina & @bdice !

Issue #14091, opened 2023-09-12
[BUG] Default type of empty list for `Index` should be `object` dtype


galipremsagar:
**Describe the bug**
When an empty list is passed to `Index` constructor `cudf` defaults to `float64`, whereas pandas defaults to `object`. We should match the same.

**Steps/Code to reproduce bug**
```python
In [1]: import cudf

In [2]: cudf.Index([])
Out[2]: Float64Index([], dtype='float64')

In [3]: import pandas as pd

In [4]: pd.Index([])
Out[4]: Index([], dtype='object')
```

**Expected behavior**
```
In [4]: cudf.Index([])
Out[4]: StringIndex([], dtype='object')
```

**Environment overview (please complete the following information)**
 - Environment location: [Bare-metal]
 - Method of cuDF install: [from source]

mroeschke:
Note that in 2.0, an empty `pandas.Series` also defaulted to `object`

PR #14116, opened 2023-09-18
Fix type of empty `Index` and raise warning in `Series` constructor

galipremsagar:
## Description
Fixes: #14091 
This PR fixes empty inputs dtype in `Index` to default to `str` instead of `float64`. Another change is there is a deprecation warning for `Series` constructor to match pandas.

## Checklist


python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -2001,7 +2002,7 @@ def test_series_shape():
 
 def test_series_shape_empty():
     ps = pd.Series(dtype="float64")
-    cs = cudf.Series([])
+    cs = _create_cudf_series([])
```
bdice, line 2005: Rather than call this helper function to make cudf use a float64, shouldn't we alter this test to call `pd.Series([])`?

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -2840,7 +2841,7 @@ def test_series_all_null(num_elements, null_type):
 @pytest.mark.parametrize("num_elements", [0, 2, 10, 100])
 def test_series_all_valid_nan(num_elements):
     data = [np.nan] * num_elements
-    sr = cudf.Series(data, nan_as_null=False)
+    sr = _create_cudf_series(data, nan_as_null=False)
```
bdice, line 2844: Why is this change needed?

For `num_elements == 0`, the null count should still be zero even if it interprets the dtype as string/object (so the test is still valid, though it might warn?).

For `num_elements > 0`, the data should be non-empty and thus be correctly interpreted as a floating type.

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -4073,28 +4074,28 @@ def test_empty_dataframe_describe():
 
 
 def test_as_column_types():
-    col = column.as_column(cudf.Series([]))
+    col = column.as_column(_create_cudf_series([]))
```
bdice, line 4077: Same here and below -- I'm not sure if I understand why these helper functions are needed. It seems like the test construction might need subtle changes but I don't think this casting helper function is the right answer. It's not announcing what it is actually doing -- the differences between `cudf.Series` and `_create_cudf_series` should be in the name, if this is needed (like `_create_cudf_series_float64_default`).

bdice (changes requested):

python/cudf/cudf/tests/test_dataframe.py
galipremsagar: Changed it to `Series([], dtype="float64")` and other places below. But retained the helper function calls in few places which can be used generically.

python/cudf/cudf/tests/test_dataframe.py
galipremsagar: Agreed, reverted the change here and passed explicit dtypes. Renamed the helper to `_create_cudf_series_float64_default` and `_create_pandas_series_float64_default`. These will be removed in `pandas-2.0` upgrade.

bdice (approved):
Approving -- but let's make sure to remove these testing helper functions early in the pandas 2.0 migration.

PR #14121, opened 2023-09-18
Enable direct ingestion and production of Arrow scalars

vyasr:
## Description

This PR adds overloads of `from_arrow` and `to_arrow` for scalars to enable interoperability on par with Arrow Arrays. The new public APIs accept streams, and for consistency streams have also been added to the corresponding column APIs, so this PR contributes to #925.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #14124, opened 2023-09-18
Replace Python scalar conversions with libcudf

vyasr:
## Description

This PR replaces the various Cython converters for different libcudf scalar types by using the new libcudf `[to|from]_arrow` overloads for scalars introduced in #14121. This change dramatically simplifies the Cython code and paves the way for implementation of a pylibcudf.Scalar object.

## Checklist


vyasr:
C++ codeowners (@mythrocks and @bdice are requested), please disregard review requests on this PR. This PR depends on #14121 and should not be merged until after it. However, it can be reviewed now for Python changes. All the C++ changes will be merged as part of the other PR.

Python reviewers can go ahead and review this PR as is, just ignore all changes to the cpp/ directory.

python/cudf/cudf/_lib/interop.pyx
```diff
@@ -182,3 +199,79 @@ def from_arrow(object input_table):
...
+def to_arrow_scalar(DeviceScalar source_scalar):
+    """Convert a scalar to a PyArrow scalar.
+
+    Parameters
+    ----------
+    source_scalar : the scalar to convert
+
+    Returns
+    -------
+    pyarrow scalar
```
brandon-b-miller, line 214:
```suggestion
    pyarrow.lib.Scalar
```

python/cudf/cudf/_lib/interop.pyx
```diff
@@ -182,3 +199,79 @@ def from_arrow(object input_table):
...
+    """Convert from PyArrow scalar to a cudf scalar.
+
+    Parameters
+    ----------
+    input_scalar : PyArrow scalar
+    output_dtype : output type to cast to, ignored except for decimals
+
+    Returns
+    -------
+    A cudf scalar
```
brandon-b-miller, line 241:
```suggestion
    cudf::scalar
```

python/cudf/cudf/_lib/interop.pyx
```diff
@@ -182,3 +199,79 @@ def from_arrow(object input_table):
...
+    cdef shared_ptr[CScalar] cpp_arrow_scalar = (
+        pyarrow_unwrap_scalar(input_scalar)
+    )
+    cdef unique_ptr[scalar] c_result
+
+    with nogil:
+        c_result = move(cpp_from_arrow(cpp_arrow_scalar.get()[0]))
+
+    cdef type_id ctype = c_result.get().type().id()
+    if ctype == type_id.DECIMAL128:
```
brandon-b-miller, line 252: This function is really clean and symmetric vs `from_pyarrow_scalar`, it would be even cleaner if we could factor this out into a `handle_decimal_type(c_result)`

python/cudf/cudf/_lib/scalar.pyx
```diff
@@ -97,61 +86,61 @@ cdef class DeviceScalar:
             A NumPy dtype.
         """
```
brandon-b-miller, line 87: Is the description of `value` still up to date with respect to handling of nulls? I don't see where we fail at this point for `None` without a `dtype` also being passed.

python/cudf/cudf/_lib/scalar.pyx
```diff
@@ -97,61 +86,61 @@ cdef class DeviceScalar:
...
-        elif isinstance(dtype, cudf.StructDtype):
-            _set_struct_from_pydict(self.c_value, value, dtype, valid)
+
+        if cudf.utils.utils.is_na_like(value):
+            value = None
+        else:
+            # TODO: For now we always deepcopy the input value to avoid
+            # overwriting the input values when replacing nulls. Since it's
+            # just host values it's not that expensive, but we could consider
+            # alternatives.
```
brandon-b-miller, line 96: It seems like the most natural thing here is for `pa.scalar` to accept an `na_value` in its constructor. How do we handle this in the column case? Can you remind me what the consequences would be for just allowing this to fail at the `pa.scalar` call below?

python/cudf/cudf/_lib/scalar.pyx
```diff
@@ -97,61 +86,61 @@ cdef class DeviceScalar:
...
-        elif pd.api.types.is_string_dtype(self.dtype):
-            result = _get_py_string_from_string(self.c_value)
-        elif pd.api.types.is_numeric_dtype(self.dtype):
-            result = _get_np_scalar_from_numeric(self.c_value)
-        elif pd.api.types.is_datetime64_dtype(self.dtype):
-            result = _get_np_scalar_from_timestamp64(self.c_value)
-        elif pd.api.types.is_timedelta64_dtype(self.dtype):
-            result = _get_np_scalar_from_timedelta64(self.c_value)
+        is_datetime = cudf.api.types.is_datetime64_dtype(self.dtype)
+        is_timedelta = cudf.api.types.is_timedelta64_dtype(self.dtype)
```
brandon-b-miller, line 118: Might be able to avoid going up through the top level `cudf` python namespace with 

```suggestion
        is_datetime = self.dtype.kind == 'M'
        is_timedelta = self.dtype.kind == 'm'
```

python/cudf/cudf/_lib/interop.pyx
vyasr: Hold that though. This interop module is an intermediate state, it's being replaced in #14133 (the follow-up PR that actually introduces a pylibcudf.Scalar). Let's do this cleanup in that PR since it'll need to look a bit different there, if that's OK with you?

python/cudf/cudf/_lib/scalar.pyx
vyasr: Nice I like it.

python/cudf/cudf/_lib/scalar.pyx
vyasr: I'm not sure that I understand your question. Are you asking whether pyarrow understands pandas.NA? If so, the answer is no:

```
>>> pa.scalar(pd.NA)
  File "<stdin>", line 1, in <module>
  File "pyarrow/scalar.pxi", line 1120, in pyarrow.lib.scalar
  File "pyarrow/error.pxi", line 144, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 100, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Could not convert <NA> with type NAType: did not recognize Python value type when inferring an Arrow data type
```
so we have to replace those NAs with normal Python None. The consequence of not doing this would be failing for any scalar that contained NA (or NaT), including nested scalars.

python/cudf/cudf/_lib/scalar.pyx
vyasr: It is, yes. Now the failure first happens at the C++ level, see https://github.com/rapidsai/cudf/blob/branch-23.10/cpp/src/interop/from_arrow.cu#L520. However the old failure mode in Python also still exists. I think this docstring was never quite right, because the failure happens at the Scalar layer before it ever reaches DeviceScalar: https://github.com/rapidsai/cudf/blob/branch-23.10/python/cudf/cudf/core/scalar.py#L237

python/cudf/cudf/_lib/scalar.pyx
brandon-b-miller: We connected offline and chatted about this 👍

brandon-b-miller (approved):

PR #14125, opened 2023-09-19
Avoid circular cimports in _lib/cpp/reduce.pxd

vyasr:
## Description

This Cython modules contains some cimports from higher-level modules than it should, which introduces the possibility for circular import issues. Also it contains an unused import of DeviceScalar that can cause similar issues.

## Checklist


wence- (approved):
Thanks!

PR #14133, opened 2023-09-19
Add pylibcudf.Scalar that interoperates with Arrow scalars

vyasr:
## Description

This PR adds a new Scalar object to pylibcudf that will function as the pylibcudf equivalent of cudf::scalar. Unlike columns, which are typically operated on in the form of views rather than owning types by libcudf, owning scalars are accepted by (const) ref by libcudf APIs and no corresponding view type exists. Therefore, pylibcudf.Scalar differs from pylibcudf.Column by actually owning an instance of the underlying libcudf type (cudf::scalar). Construction of pylibcudf Scalars is expected to be done from an Arrow scalar.

This PR relies on #14124 and should not be merged until after that one.
## Checklist


python/cudf/cudf/_lib/scalar.pyx
```diff
@@ -61,15 +60,32 @@ def _replace_nested(obj, check, replacement):
                 _replace_nested(v, check, replacement)
 
 
-# The DeviceMemoryResource attribute could be released prematurely
-# by the gc if the DeviceScalar is in a reference cycle. Removing
-# the tp_clear function with the no_gc_clear decoration prevents that.
-# See https://github.com/rapidsai/rmm/pull/931 for details.
-@cython.no_gc_clear
+def gather_metadata(dtypes):
```
shwina, line 63: Can we add a comment explaining what this function does/returns?

python/cudf/cudf/_lib/pylibcudf/scalar.pxd
```diff
@@ -0,0 +1,32 @@
...
+
+    cdef const scalar* get(self) except *
+
+    cpdef DataType type(self)
+    cpdef bool is_valid(self)
+
+    @staticmethod
+    cdef Scalar from_libcudf(unique_ptr[scalar] libcudf_scalar, dtype=*)
+
+    # TODO: Make sure I'm correct to avoid typing the metadata as
```
shwina, line 30: Should we resolve before merging?

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
```diff
@@ -0,0 +1,119 @@
...
+
+    @staticmethod
+    def from_pyarrow_scalar(pyarrow.lib.Scalar value, DataType data_type=None):
+        # Allow passing a dtype, but only for the purpose of decimals for now
+
+        # Need a local import here to avoid a circular dependency because
+        # from_arrow_scalar returns a Scalar.
+        from .interop import from_arrow_scalar
+
+        cdef Scalar s = from_arrow_scalar(value)
```
shwina, line 53: We should eliminate one of `Scalar.from_pyarrow_scalar` or `interop.from_arrow_scalar`. Currently they are both part of the "public" pylibcudf API, and it's not clear when one should use which. It seems that `Scalar.from_pyarrow_scalar()` is the correct way to construct a `Scalar` object, while the `interop.from_arrow_scalar` is just a helper for the former.

shwina: Subsuming `interop.from_arrow_scalar` here will have the added advantage of not needing a function-local import.

python/cudf/cudf/_lib/pylibcudf/scalar.pxd
vyasr: Yup. This comment is correct and can be removed. Depending on the resolution of your other comment though (combining the Scalar method and the interop function) this may not be an issue, let me see.

python/cudf/cudf/_lib/scalar.pyx
```diff
@@ -108,18 +124,26 @@ cdef class DeviceScalar:
...
-        cdef DeviceScalar obj = from_arrow_scalar(pa_scalar, self._dtype)
-        self.c_value.swap(obj.c_value)
+        data_type = None
+        if isinstance(dtype, cudf.core.dtypes.DecimalDtype):
+            tid = pylibcudf.TypeId.DECIMAL128
+            if isinstance(dtype, cudf.core.dtypes.Decimal32Dtype):
+                tid = pylibcudf.TypeId.DECIMAL32
+            elif isinstance(dtype, cudf.core.dtypes.Decimal64Dtype):
+                tid = pylibcudf.TypeId.DECIMAL64
+            data_type = pylibcudf.DataType(tid, -dtype.scale)
```
wence-, line 134: **thought (non-blocking):** ‏Can we invert control here and let dtypes deliver their pylibcudf equivalent? Or is that not possible because the `DataType` is a `cdef` object that can't be built from python?

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
```diff
@@ -0,0 +1,119 @@
...
+    decimal64,
+    decimal128,
+    scale_type,
+)
+
+from .types cimport DataType, type_id
+
+
+# The DeviceMemoryResource attribute could be released prematurely
+# by the gc if the DeviceScalar is in a reference cycle. Removing
```
wence-, line 23:
```suggestion
# by the gc if the Scalar is in a reference cycle. Removing
```

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
```diff
@@ -0,0 +1,119 @@
...
+# the tp_clear function with the no_gc_clear decoration prevents that.
+# See https://github.com/rapidsai/rmm/pull/931 for details.
```
wence-, line 25: **thought (non-blocking):** ‏Perhaps we should (in a followup) document this issue more discoverably by adding some text to the dev docs.

wence- (changes requested):
Mostly minor comments with one issue to do with memory-resource management going back and forth between py- and libcudf

python/cudf/cudf/_lib/pylibcudf/interop.pxd
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+from pyarrow.lib cimport Scalar as pa_Scalar, Table as pa_Table
```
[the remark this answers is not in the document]
vyasr: Nice, I like that. It'll clean up the pyx file too.

python/cudf/cudf/_lib/pylibcudf/interop.pyx
```diff
@@ -0,0 +1,98 @@
...
+from .scalar cimport Scalar
+from .table cimport Table
+
+
+cdef class ColumnMetadata:
+    def __init__(self, name):
+        self.name = name
+        self.children_meta = []
+
+    cdef column_metadata to_c_metadata(self):
```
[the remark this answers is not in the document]
vyasr: Yup, done.

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
```diff
@@ -0,0 +1,119 @@
...
+            )
+        elif tid == type_id.DECIMAL64:
+            s.c_obj.reset(
+                new fixed_point_scalar[decimal64](
+                    (<fixed_point_scalar[decimal128]*> s.c_obj.get()).value(),
+                    scale_type(-value.type.scale),
+                    s.c_obj.get().is_valid()
+                )
+            )
+        return s
```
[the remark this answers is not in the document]
vyasr: Changed your suggestion a tiny bit but largely implemented as is. Agree that this is much clearer.

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
```diff
@@ -0,0 +1,119 @@
...
+
+    def __init__(self, pyarrow.lib.Scalar value=None):
+        # TODO: This case is not something we really want to
+        # support, but it here for now to ease the transition of
+        # DeviceScalar.
+        if value is not None:
+            raise ValueError("Scalar should be constructed with a factory")
+
+    @staticmethod
+    def from_pyarrow_scalar(pyarrow.lib.Scalar value, DataType data_type=None):
```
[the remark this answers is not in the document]
vyasr: Agreed, will change.

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
vyasr: Yeah perhaps there's a more discoverable place for this, @shwina WDYT?

python/cudf/cudf/_lib/scalar.pyx
vyasr: Yup.

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
vyasr: This is done.

vyasr: For consistency I've done the same for Table.

python/cudf/cudf/_lib/pylibcudf/scalar.pxd
vyasr: OK, with the changes to make the class methods the primary means of going between arrow and pylibcudf Scalar/Table objects this limitation is no longer present. I've typed the method appropriately.

python/cudf/cudf/_lib/pylibcudf/__init__.pxd
```diff
@@ -1,9 +1,13 @@
 # Copyright (c) 2023, NVIDIA CORPORATION.
 
 # TODO: Verify consistent usage of relative/absolute imports in pylibcudf.
-from . cimport copying
+# TODO: Cannot import interop because it introduces a build-time pyarrow header
+# dependency for everything that cimports pylibcudf. See if there's a way to
+# avoid that before polluting the whole package.
+from . cimport copying  # , interop
```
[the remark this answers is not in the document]
vyasr: I think that's fair, and I agree that's how you'd solve it in C/C++. It's a question of how you'd want a Cython package to look; in Python it feels more natural to import a package and be able to access subpackages/submodules without needing to do a separate import, wouldn't you say?

For now this particular question is no longer imminently pressing. The changes to move the interop code directly into Column/Scalar means that all of cudf's Cython now requires these headers anyway, so for consistency I'll do the same in pylibcudf since it's no longer so onerous relatively speaking. I definitely want to come back to this discussion when pylibcudf is closer to being complete and we really start thinking about the public API though. There's some tricky questions about to what extent it's a C library vs a Python library when used in pure Cython mode.

python/cudf/cudf/_lib/scalar.pyx
vyasr: The issue is more that dtype could be a non-cudf dtype as well and we don't have control over those (e.g. numpy.dtype objects).

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
```diff
@@ -0,0 +1,119 @@
...
+        if tid == type_id.DECIMAL128:
+            return s
+
+        if tid == type_id.DECIMAL32:
+            s.c_obj.reset(
+                new fixed_point_scalar[decimal32](
+                    (<fixed_point_scalar[decimal128]*> s.c_obj.get()).value(),
+                    scale_type(-value.type.scale),
+                    s.c_obj.get().is_valid()
+                )
```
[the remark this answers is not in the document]
vyasr: Hmm that's an interesting question I hadn't considered. In theory I suppose in a multi-threaded application it's possible for `s` to be created, then this thread to be preempted by another thread that does a `set_per_device_resource`, then control flow resume so that the second constructor calls use a different device resource. So while the mr of s can't change, the current resource could have changed. I don't think that's a problem though, at least not in this scenario. Do you see a way it could cause issues?

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
```diff
@@ -0,0 +1,119 @@
...
+
+    @staticmethod
+    cdef Scalar from_libcudf(unique_ptr[scalar] libcudf_scalar, dtype=None):
+        """Construct a Scalar object from a libcudf scalar.
+
+        This method is for pylibcudf's functions to use to ingest outputs of
+        calling libcudf algorithms, and should generally not be needed by users
+        (even direct pylibcudf Cython users).
+        """
+        cdef Scalar s = Scalar.__new__(Scalar)
```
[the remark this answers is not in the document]
vyasr: That's a great find. I think you are correct. Unfortunately I don't know if we have a great way to solve this problem right now. rmm DeviceMemoryResource objects can't be easily constructed from librmm objects AFAIK, so we can't extract the one from the underlying scalar to create a suitable Python object to store to the pylibcudf.Scalar instance. The way I see it we have two options:
1. Implement that sort of constructor in some way
2. Modify librmm scalars to keep their memory resources alive via a shared pointer or so

2 is a very far-reaching change that I don't see happening anytime soon (I assume there was a reason for the current design where rmm::device_buffer doesn't share ownership of the allocating mr, but we may want to revisit that design), so we may have to wait on 1 to fix this. Note that this isn't a new issue for cudf, just something we're noticing now.

@shwina WDYT?

vyasr:
Retargeting to 23.12.

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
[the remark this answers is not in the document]
wence-: >  so we may have to wait on 1 to fix this.

I don't  think being able to construct a `DeviceMemoryResource` from a librmm `device_memory_resource *` (which is what `device_buffer` and `device_scalar` carry around) would work. Since on the librmm side we just have a raw pointer, we have no means of keeping it alive. AIUI, the RMM philosophy is that memory resources and device_buffers are RAII and it is up to the user code to manage the lifetime of the memory resource attached to any given device buffer.

I think this means that if you have some object:

```c++
struct foo {
    ...
    rmm::device_buffer<...> _data;
}
```
and your library creates objects of type `foo` with unique ownership semantics (i.e. it hands back `std::unique_ptr<foo>` from creation routines), there isn't a safe way to take ownership of that object unless the library _also_ provides a way of transferring (or sharing) ownership of the `memory_resource` it used when making the `foo` things.

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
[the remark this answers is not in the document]
wence-: I think this potentially suffers from the same use-after-free discussed below.

Suppose `s` is created with `memory_resource` `A` (obtained via `get_current_device_resource`) and, before we get to this line, (in a multithreaded setting) the device resource is changed to `B`, before we get back here.

Now suppose that on the cudf side (which we'll assume had an owning handle on `A`) we now drop `A` (and move to using `B`) and then drop all objects which were allocated with `A`. So `~A` runs, and does one of two things:

1. deletes itself (so the dangling pointer to `A` in `s` is invalid) => when we come to free `s`, we'll get a memory error, potentially
2. if `A` is a pool, releases the pool allocation, so now `s` points to invalid memory.

I think we are reasonably safe from all this because the way cudf drives libcudf, it is relatively in control of lifetime of the memory resources but we probably want to think about whether cudf should explicitly pass memory resources into all libcudf APIs that allocate data? I _suspect_ the answer is yes, though this is a huge job (ala streams). If we did that, this problem would go away because we would have control over the memory resource being used for all allocations (and would know it's one whose lifetime we can control).

python/cudf/cudf/_lib/pylibcudf/scalar.pyx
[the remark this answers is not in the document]
vyasr: You're right, but let's continue discussion of mrs on the other open thread.

python/cudf/cudf/_lib/pylibcudf/__init__.pxd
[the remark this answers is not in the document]
wence-: > in Python it feels more natural to import a package and be able to access subpackages/submodules without needing to do a separate import, wouldn't you say?

yes/no? scipy used to require you to explicitly import subpackages (although now they do on-demand lazy-loading so it's no longer the case). Some large packages I've worked on have had subpackage setups with an "import-what-you-need" approach for namespacing and to reduce import times.

But agreed we don't need to do anything now.

wence- (approved):
Thanks! Good to go from my side. I think we can just open a doc issue for the outstanding comment about the `@no_gc_clear` decorator.

PR #14137, opened 2023-09-20
Add validation to reader benchmarks

vuule:
## Description
Validate that cuIO readers successfully round trip the input tables.
Validation is only done in the first iteration, and it is not included in the timing.
If there is a difference, a warning is logged.
Setting `CUDF_BENCH_OUTPUT_DIFF` environment variable adds diff to the standard output. Valid values are `FIRST_ERROR` and `ALL_ERRORS`.

Benchmarks currently report differences for data types that can't be preserved through the given formats. There are not differences caused by data corruption.

## Checklist


vyasr (changes requested):
The implementation seems sensible, but why are we doing this in benchmarks rather than tests?

vuule:
> The implementation seems sensible, but why are we doing this in benchmarks rather than tests?

To verify that the benchmarks are doing what they're supposed to. Also, it's helpful in situations where we benchmark different configurations (thread pool size, GDS policy, compression policy...). Basically enabling us to benchmark changes without jumping back to tests for each change.

PR #14138, opened 2023-09-20
Reduce memory usage of as_categorical_column

wence-:
## Description

The main culprit is in the way the codes returned from _label_encoding were being ordered. We were generating an int64 column for the order, gathering through the left gather map, and then argsorting, before using that ordering as a gather map for the codes.

We note that gather(y, with=argsort(x)) is equivalent to sort_by_key(y, with=x) so use that instead (avoiding an unnecessary gather). Furthermore we also note that gather([0..n), with=x) is just equivalent to x, so we can avoid a gather too.

This reduces the peak memory footprint of categorifying a random column of 500_000_000 int32 values where there are 100 unique values from 24.75 GiB to 11.67 GiB.

### Test code

```python
import cudf
import cupy as cp

K = 100
N = 500_000_000
rng = cp.random._generator.RandomState()
column = cudf.core.column.as_column(rng.choice(cp.arange(K, dtype="int32"), size=(N,), replace=True))
column = column.astype("category", ordered=False)
```

### Before

Screenshot from 2023-09-20 14-49-27

### After

Screenshot from 2023-09-20 14-49-42

## Checklist

### DIFF
--- a/python/cudf/cudf/core/column/column.py
+++ b/python/cudf/cudf/core/column/column.py
@@ -1390,20 +1390,19 @@ class ColumnBase(Column, Serializable, BinaryOperand, Reducible):
         except ValueError:
             return _return_sentinel_column()
 
-        codes = arange(len(cats), dtype=dtype)
         left_gather_map, right_gather_map = cpp_join(
             [self], [cats], how="left"
         )
-        codes = codes.take(
-            right_gather_map, nullify=True, check_bounds=False
-        ).fillna(na_sentinel.value)
-
+        codes = libcudf.copying.gather(
+            [arange(len(cats), dtype=dtype)], right_gather_map, nullify=True
+        )
+        del right_gather_map
         # reorder `codes` so that its values correspond to the
         # values of `self`:
-        order = arange(len(self))
-        order = order.take(left_gather_map, check_bounds=False).argsort()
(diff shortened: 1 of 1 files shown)


galipremsagar (approved):

bdice (approved):
Great! Note that this is an example of the performance antipattern discussed in #13557.

shwina (approved):

harrism:
Is performance affected?

wence-:
> Is performance affected?

Yes, but positively, I run:
```python
import time
import cupy as cp
import cudf
import rmm

rmm.reinitialize(pool_allocator=True)

rng = cp.random._generator.RandomState(seed=108)
for K in [2**4, 2**10, 2**12, 2**14, 2**16]:
    for N in [1_000_000, 10_000_000, 100_000_000, 250_000_000]:
        col = cudf.core.column.as_column(rng.choice(cp.arange(K, dtype="uint32"), size=N, replace=True))
        start = time.time()
        for _ in range((reps := 1_000_000_000 // N)):
            y = col.astype("category", ordered=False)
            del y
        end = time.time()
        del col
```

performance-improvement

Across column sizes and number of unique values, the new code is between 25 and 30% faster.

harrism:
Excellent!

PR #14158, opened 2023-09-21
Fix calls to copy_bitmask to pass stream parameter

davidwendt:
## Description
Fixes a couple places where `cudf::copy_bitmask` was called instead of `cudf::detail::copy_bitmask` to pass the available stream (and mr) parameters.

Found while reviewing #14121
Reference: https://github.com/rapidsai/cudf/pull/14121#discussion_r1332332391

## Checklist


cpp/src/lists/count_elements.cu
```diff
@@ -52,7 +52,7 @@ std::unique_ptr<column> count_elements(lists_column_view const& input,
   // create output column
   auto output = make_fixed_width_column(data_type{type_to_id<size_type>()},
                                         input.size(),
-                                        copy_bitmask(input.parent()),
+                                        copy_bitmask(input.parent(), stream, mr),
```
bdice, line 55: Do we need to be explicit about `detail::` here? Once we add streams to the public API, will the signatures of public and detail be the same?

cpp/src/lists/count_elements.cu
bdice, line 1: Possibly out of scope but wanted to mention it: we could update the docstring `@return New INT32 column with lengths.` to indicate `size_type` instead.

cpp/src/lists/count_elements.cu
davidwendt: The compiler does not have a problem with it. I'm happy to be explicit if it helps with reading the code.
Yes, once we add streams to the public API, the signatures will be the same minus the namespace.

Which brings up the question: Perhaps we should consider removing detail APIs that match public APIs in the future?

bdice: > Perhaps we should consider removing detail APIs that match public APIs in the future?

We’ve ruled against this in past conversations because of the NVTX ranges in the public API that we don’t want at the detail level.

bdice (approved):

mythrocks (approved):

Issue #13656, opened 2023-07-03
[BUG] libcudf fails to recognized malformed dictionary during Parquet read


jlowe:
**Describe the bug**
Using libcudf to load a Parquet file with a malformed dictionary does not result in an error. 

**Steps/Code to reproduce bug**
Read the [nation.dict-malformed.parquet file from apache/parquet-testing](https://github.com/apache/parquet-testing/blob/master/data/nation.dict-malformed.parquet)  with libcudf and note that it does not error. 

**Expected behavior**
Loading this from Apache Spark or via the [parquet-cli tool](https://github.com/apache/parquet-mr/tree/master/parquet-cli) results in a decode error as expected:
```
java -cp '/home/jlowe/src/parquet-mr/parquet-cli/target/parquet-cli-1.13.1.jar:/home/jlowe/src/parquet-mr/parquet-cli/target/dependency/*' org.apache.parquet.cli.Main cat parquet-testing/data/nation.dict-malformed.parquet
Unknown error
java.lang.RuntimeException: Failed on record 0
	at org.apache.parquet.cli.commands.CatCommand.run(CatCommand.java:86)
	at org.apache.parquet.cli.Main.run(Main.java:163)
	at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:76)
	at org.apache.parquet.cli.Main.main(Main.java:193)
Caused by: java.lang.RuntimeException: Failed while reading Parquet file: /home/jlowe/src/spark-rapids/thirdparty/parquet-testing/data/nation.dict-malformed.parquet
	at org.apache.parquet.cli.BaseCommand$1$1.advance(BaseCommand.java:367)
	at org.apache.parquet.cli.BaseCommand$1$1.<init>(BaseCommand.java:344)
	at org.apache.parquet.cli.BaseCommand$1.iterator(BaseCommand.java:342)
	at org.apache.parquet.cli.commands.CatCommand.run(CatCommand.java:73)
	... 3 more
Caused by: java.io.EOFException
	at org.apache.parquet.bytes.SingleBufferInputStream.sliceBuffers(SingleBufferInputStream.java:134)
	at org.apache.parquet.hadoop.ParquetFileReader$Chunk.readAsBytesInput(ParquetFileReader.java:1688)
	at org.apache.parquet.hadoop.ParquetFileReader$Chunk.readAllPages(ParquetFileReader.java:1604)
[...]

PR #14167, opened 2023-09-21
Propagate errors from Parquet reader kernels back to host

vuule:
## Description
Pass the error code to the host when a kernel detects invalid input.
If multiple errors types are detected, they are combined using a bitwise OR so that caller gets the aggregate error code that includes all types of errors that occurred.

Does not change the kernel side checks.

## Checklist


vuule:
CC @nvdbaranec @etseidl 
Please let me know if you see any issues with this approach to error reporting in the reader.

cpp/src/io/parquet/page_data.cu
```diff
@@ -596,6 +602,10 @@ __global__ void __launch_bounds__(decode_block_size) gpuDecodePageData(
     }
     __syncthreads();
   }
+  if (!t and s->error != 0) {
```
etseidl, line 605: t != 0 while we're at it?

etseidl (commented):
I can't think of a better way to do this. Do we want to define some constants for the error codes?

vuule:
> Do we want to define some constants for the error codes?

Definitely, just not sure if it should be in this PR. Related - should we return the error code as a bitmask? Would returning multiple errors even be useful?

etseidl:
> Definitely, just not sure if it should be in this PR. Related - should we return the error code as a bitmask? Would returning multiple errors even be useful?

I think a bitmask might be a bit much, and limits us to 32 errors. There will probably be more ways to fail than that, esp if we also return errors from the preprocessing kernels.

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -68,6 +69,12 @@ struct page_state_s {
   PageNestingDecodeInfo nesting_decode_cache[max_cacheable_nesting_decode_info];
   // points to either nesting_decode_cache above when possible, or to the global source otherwise
   PageNestingDecodeInfo* nesting_info;
+
+  inline __device__ void set_error_code(int32_t err) volatile
```
divyegala, line 73: Am I correct in assuming that different threads in a block may find different errors, and we set any one of the error codes here due to `memory_order_relaxed`? And then we extend this to `thread_scope_device` where different blocks may find different errors and we set the final error returned to the user as any one of them?

Which means errors are not deterministic?

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -68,6 +69,12 @@ struct page_state_s {
...
+  {
+    cuda::atomic_ref<int32_t, cuda::thread_scope_block> ref{const_cast<int&>(error)};
+    ref.store(err, cuda::std::memory_order_relaxed);
```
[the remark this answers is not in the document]
vuule: I made it atomic since we probably don't need to worry about performance when failing. This seemed like a safe option for future checks as well.

About the error code as mask - Ed is concerned about the limit on the number of errors that this would impose. I could be convinced to go either way, don't expect the trade-off to be relevant in practice.

cpp/src/io/parquet/page_decode.cuh
vuule: Not quite deterministic, yes. 
And this is regardless of the relaxed order, since we call three kernels with the same output error code address.
This may be a good argument to use a bitwise mask for the errors.

cpp/src/io/parquet/page_decode.cuh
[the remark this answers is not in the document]
etseidl: TBH the most common error condition is going to be a buffer overrun detected somewhere. We could probably get away without codes at all and have a single error bit. The host code calling the kernel can report which kernel failed. It just comes down to how fine grained you want the error reporting to be.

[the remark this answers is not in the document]
nvdbaranec: I could see it either way. It's so hard to even know what thread failed and the context of why (possibly because some other thread did something wrong) having a set of bits could act as bread-crumbs to lead you to where things really went wrong. But on the other hand, you're a lot more limited on what you can report.  I'm fine either way.  Parallel error reporting is amusing in any case.

cpp/src/io/parquet/page_decode.cuh
divyegala: How many items would the bitset report errors for? 1 bit per block or 1 bit per thread? Maybe we should also allow the user to "enable" host-side error reporting

vuule: We wouldn't separate errors from different threads/blocks, just errors from different root causes but applying a bitwise OR when updating the error code. Final error code would represent all types of errors we encountered in any thread/block/kernel.

divyegala: Ah, yeah, we should definitely do this

vuule:
Looks like we're leaning towards a mask to aggregate errors. I'll make the changes.

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -68,6 +79,18 @@ struct page_state_s {
...
+  inline __device__ void set_error_code(decode_error err) volatile
+  {
+    cuda::atomic_ref<int32_t, cuda::thread_scope_block> ref{const_cast<int&>(error)};
+    ref.fetch_or(static_cast<int32_t>(err), cuda::std::memory_order_relaxed);
+  }
+
+  inline __device__ void reset_error_code() volatile
+  {
+    cuda::atomic_ref<int32_t, cuda::thread_scope_block> ref{const_cast<int&>(error)};
+    ref.store(0, cuda::std::memory_order_relaxed);
```
vuule, line 92: @PointKernel is it okay to used relaxed memory order for this one?

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -21,10 +21,21 @@
 
 #include <io/utilities/block_utils.cuh>
 
+#include <cuda/atomic>
 #include <cuda/std/tuple>
 
 namespace cudf::io::parquet::gpu {
 
+enum class decode_error {
```
etseidl, line 29: Should this go in `parquet_gpu.hpp`? I'm thinking ahead to using the same codes in `gpuDecodePageHeaders` to detect the error from #13656.

vuule: Yeah, you're right.
Tangential to the suggestion: I think current check already cover #13656. Will verify in a bit.

nvdbaranec (approved):
Yeah I like this mechanism.  The explicit names also remove some of the mystery when reading the code itself too.

cpp/src/io/parquet/page_decode.cuh
vuule: I mixed it up with another "should fail instead of returning garbage" issue. No current checks catch #13656.

cpp/src/io/parquet/page_decode.cuh
PointKernel:
```suggestion
    ref.store(0, cuda::std::memory_order_release);
```
This should be `release` since the stored value must be visible for all reads/writes after it. Glad you ask!

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -487,7 +500,14 @@ __device__ void gpuDecodeStream(
             cur++;
           }
         }
-        if (cur > end || level_run <= 1) { s->error = 0x10; }
+        if (cur > end) {
+          s->set_error_code(decode_error::STREAM_OVERRUN);
```
etseidl, line 504: this is decoding RLE level data...should be RLE_STREAM_OVERRUN (in the current naming...)

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -1279,22 +1301,23 @@ inline __device__ bool setupLocalPageInfo(page_state_s* const s,
           // first 4 bytes are length of RLE data
           int const len = (cur[0]) + (cur[1] << 8) + (cur[2] << 16) + (cur[3] << 24);
           cur += 4;
-          if (cur + len > end) { s->error = 2; }
+          if (cur + len > end) { s->set_error_code(decode_error::RLE_STREAM_OVERRUN); }
```
etseidl, line 1304: Here we're doing RLE encoded data. Perhaps it would be better to instead have the errors split up along level data vs actual data lines, so `LEVEL_STREAM_OVERRUN` and `DATA_STREAM_OVERRUN` vs the current `RLE_STREAM/STREAM_OVERRUN`

etseidl (commented):
Looking good, just a few naming nits :D

cpp/src/io/parquet/page_decode.cuh
vuule: Made the suggested changes, please check if I'm using the right codes in all places.

etseidl: Yep, I think it's good to go now.

divyegala (approved):

PointKernel (approved):
LGTM

bdice (approved):

PR #14168, opened 2023-09-22
Fix assert failure for range window functions

### LINKED ISSUE #13855: [BUG] Rolling window function tests fail assert in DEBUG
`ROLLING_TEST` seems to be failing asserts in `DEBUG` builds, even though the results are correct.

```
gtests/ROLLING_TEST --gtest_filter=TypedCollectListTest/0.BasicGroupedTimeRangeRollingWindow
```
Fails this assert in `grouped_rolling.cu`.
```
    cudf_assert(type_id_matches_device_storage_typeT(col.type().id()) &&
                "the data type mismatch");
```
The instantiation type `T=int32_t` does not match the `col.type().id() == TIMESTAMP_DAYS`.

mythrocks:
Fixes #13855.

This commit fixes an erroneous assert check in range window functions, where the wrong data-type is checked for timestamp order-by columns.

For timestamps, it's the `duration::rep` type that should be checked, when accessing the order by column.

This fix allows ROLLING_TEST to complete correctly, in DEBUG mode.

## Checklist


divyegala (approved):

cpp/src/rolling/grouped_rolling.cu
```diff
@@ -357,15 +357,28 @@ template <typename T>
...
    */
   explicit __device__ device_value_accessor(column_device_view const& col_) : col{col_}
   {
-    cudf_assert(type_id_matches_device_storage_type<T>(col.type().id()) &&
-                "the data type mismatch");
+    // For non-timestamp types, T must match the order-by column's type.
+    // For timestamp types, T must match the range rep type for the order-by column.
+    cudf_assert((type_id_matches_device_storage_type<T>(col.type().id()) or
+                 cudf::type_dispatcher(col.type(), is_correct_range_rep{})) &&
+                "data type mismatch when accessing the order-by column");
```
davidwendt, line 381: I'm not sure I agree with retrieving the data (via `element()`) using the rep-type and doing operations on said rep-type. I would rather the actual type be read (via `element()`) and used directly.
I feel the original error is correct and this is a flaw in the design of this code perhaps.

mythrocks: The reason we're comparing the rep-type values here is that the order-by columns are timestamps, and the preceding/following values are durations (e.g. 2 days ago). 

For integral order-by columns, preceding/following are of the same types. But not for timestamps.

mythrocks: Hmm. That said, I could stand to tighten this up a little more... I'll see about a stricter check for timestamps.

davidwendt (approved):

PointKernel (approved):

cpp/src/rolling/grouped_rolling.cu
mythrocks: Ah, I just realized that one unintended consequence of doing order-by comparisons without switching to the rep-types is that all those templates will be instantiated for all timestamp types, duration types, fixed-point, etc. :/

mythrocks:
This change has been merged. Thank you all for the reviews.

PR #14170, opened 2023-09-22
Add `bytes_per_second` to transpose benchmark

Blonck:
This patch relates to #13735.

Benchmark: [transpose_benchmark.txt](https://github.com/rapidsai/cudf/files/12699834/transpose_benchmark.txt)

## Checklist


cpp/benchmarks/transpose/transpose.cpp
```diff
@@ -40,16 +40,29 @@ static void BM_transpose(benchmark::State& state)
     cuda_event_timer raii(state, true);
     auto output = cudf::transpose(input);
   }
+
+  // Collect memory statistics.
+  auto const bytes_read    = input.num_columns() * input.num_rows() * (sizeof(int32_t));
```
harrism, line 45: I would like to avoid potential future type mismatches that result in wrong bytes/s reports. So I think you should stash the `type_id` in a variable above:
```
constexpr auto column_type = cudf::type_id::INT32;
```

And then here use CUDF's `id_to_type` utility:
```suggestion
  auto const bytes_read    = input.num_columns() * input.num_rows() * (sizeof(cudf::id_to_type(column_type)));
```

See https://docs.rapids.ai/api/libcudf/stable/group__utility__dispatcher#gad7e12b8accf60e7c0e500294e1ee8536

harrism (changes requested):
Thanks for contributing this. Just one request for maintainability.

cpp/benchmarks/transpose/transpose.cpp
```diff
@@ -42,7 +44,7 @@ static void BM_transpose(benchmark::State& state)
   }
 
   // Collect memory statistics.
-  auto const bytes_read    = input.num_columns() * input.num_rows() * (sizeof(int32_t));
+  auto const bytes_read    = input.num_columns() * input.num_rows() * cudf::size_of(column_type);
```
harrism, line 47: **💡 suggestion:** ‏ This is one way to do it. But I think the way I suggested is a bit better because it all happens at compile time, whereas `cudf::size_of()` invokes the type dispatcher at run time. Not that it will affect benchmarks, but it just seems cleaner to use `sizeof(cudf::id_to_type<column_type>)`.

harrism (approved):
Thanks @Blonck !

harrism (approved):
Thanks @Blonck !

PointKernel (approved):

PointKernel:
@Blonck Can you please rebase with the latest `branch-23.12` and fix the formatting issues?

cpp/benchmarks/transpose/transpose.cpp
```diff
@@ -40,6 +42,17 @@ static void BM_transpose(benchmark::State& state)
     cuda_event_timer raii(state, true);
     auto output = cudf::transpose(input);
   }
+
+  // Collect memory statistics.
+  auto const bytes_read =
+    input.num_columns() * input.num_rows() * sizeof(cudf::id_to_type<column_type_id>);
```
bdice, line 48: Do we need to worry about this product overflowing `size_type`? I think the number of columns and rows are both `size_type`, so their product could overflow before multiplying by the `sizeof`. Maybe need to cast everything to `std::size_t`?

harrism: Good spotting. I think you are correct.

Blonck: I've verified that both `.num_columns()` and `.num_rows()` return a `cudf::size_type`. To avoid any potential ambiguities, I would cast it to `uint64_t`. The attached performance log in this PR remains Ok, given that the current benchmark uses `32768` for both row and column sizes. This value, when squared, fits within int32_t.

cpp/benchmarks/transpose/transpose.cpp
```diff
@@ -40,6 +42,17 @@ static void BM_transpose(benchmark::State& state)
...
     auto output = cudf::transpose(input);
   }
+
+  // Collect memory statistics.
+  auto const bytes_read = static_cast<uint64_t>(input.num_columns()) * input.num_rows() *
+                          sizeof(cudf::id_to_type<column_type_id>);
+  auto const bytes_written = bytes_read;
+  // Account for nullability in input and output.
+  auto const null_bytes =
+    2 * input.num_columns() * cudf::bitmask_allocation_size_bytes(input.num_rows());
```
wence-, line 52: **suggestion:** ‏This one could also overflow, I think, perhaps:
```suggestion
    2 * static_cast<uint64_t>(input.num_columns()) * cudf::bitmask_allocation_size_bytes(input.num_rows());
```
?

Blonck: Are you sure about this one? Since the return type of `cudf::bitmask_allocation_size_bytes` is `std::size_t` which is either `unsigned long` or `unsigned long long` so for reasonable input sizes the integer type promotion will avoid the overflow (https://cppinsights.io/s/26f977cb).

That said, just having this discussion indicates I should've included an explicit cast upfront to clear up any potential confusion.

wence-: Left-to-right associativity means that this is evaluated as `(2 * ncol) * nrow`, the first multiplication is performed in `size_type` (AKA, `int32_t`), so that could overflow, no? Although I think these benchmarks are generally run with fewer than $2^{30}$ rows, there's in general no reason why they couldn't be (although the transpose performance will be _terrible_ I grant you).

harrism: Why not just always put the thing that returns `size_t` (`sizeof` or `bitmask_allocation_size_bytes`) first in the arithmetic in all of these PRs?

Blonck: Personally, I would keep the cast explicit to make visible what is happening, but I don't have a strong stance on it.

ttnghia:
It means now github starts to realize that you are no longer cudf developer :laughing:

Issue #13251, opened 2023-04-29
Improve ORC reader performance for decimal types


vuule:
Decode of decimal files is an order of magnitude slower than decode of integral types. 
The reason is the use of a single thread to find the sizes of the next batch of elements, which are then decoded using the whole block. To improve kernel performance, it needs to use multiple threads to find varint boundaries:

Pass 1: every thread runs `is_boundary_byte` (highest bit == 0) to find if it's at the last byte of a varint element.
```
A = 0 0 0 0 1 0 0 1 0 0 0 1
```
 
Pass 2: Scan A to produce B. Also gets the number of elements (3 in this case).
```
B = 0 0 0 0 1 1 1 2 2 2 2 3
    ^       ^     ^
    t0      t4    t7
```
Pass 3: Threads that are on a boundary decode the element that starts at their index and store it at col[t].
t=0 writes to [0]
t=4 writes to [1]
t=7 writes to [2]
Alternatively, step 3 can store the offsets of each element so they can be decoded in parallel.

vuule:
Additional optimization:

- divide your threads into 2 chunks based on the average length of a varint.
- Let's say 2 bytes / 1 varint.   2:1
- Divide your block of say 768 threads into 2 chunks of 512 and 256. 
- Overlap generation of the next set of offsets (512 threads) with decoding the last set (256 threads).

Issue #12086, opened 2022-11-07
[FEA] Migrate to Thrust `nosync` stream policy for performance.


bdice:
### Motivation & Description

Thrust 1.16 added an execution policy `thrust::cuda::par_nosync` that removes unnecessary internal stream synchronizations, except when required for correctness (e.g. if the algorithm returns a value to the host like `thrust::reduce`, a sync is required).

In PR #11577, I experimented with a bulk find-replace of `rmm::exec_policy` with `rmm::exec_policy_nosync`. This resulted in notable performance improvements, particularly for small data sizes. See: https://gist.github.com/bdice/bbeae4d28a45bedf0f53a13304714f70

However, a plain find-replace may leave issues with stream safety. For example, an internal detail API could be constructing host memory whose lifetime doesn't guarantee a safe copy to device without a stream sync before returning (thought experiment identified by @jrhemstad). Changing to `nosync` execution policies requires analysis of every use case individually, at both the detail and public API levels.

(As a reminder of the current stream policy: [libcudf APIs called on the host do not guarantee that the stream is synchronized before returning](https://github.com/rapidsai/cudf/blob/branch-22.12/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#treat-libcudf-apis-as-if-they-were-asynchronous), but this does not mean we can always use `nosync` safely.)

### Tasks

I am planning to open PRs to use `nosync` across the libcudf codebase. This task list is sorted by largest impact for small data sizes (or whatever fixed data sizes are in the benchmark), which indicates overhead that we can systematically eliminate. Note that these performance improvements use `nosync` everywhere, which may not be stream-safe in all cases, so the real performance gains may be lower if not all executions can use `nosync`. Additionally, the performance improvements for a given algorithm may rely on improvements in other algorithms, so the full improvement may not be achieved until all tasks are complete.

- [x] Gather/scatter. #12038
  - [Gather is 42% faster for 1024 rows, 1 column.](https://gist.github.com/bdice/bbeae4d28a45bedf0f53a13304714f70#file-nosync_benchmarks_all-txt-L449)
  - [Scatter is 37% faster for 1024 rows, 1 column.](https://gist.github.com/bdice/bbeae4d28a45bedf0f53a13304714f70#file-nosync_benchmarks_all-txt-L2052)
- [ ] Merge.
[...]

PR #14190, opened 2023-09-26
Optimize ORC writer for decimal columns

vuule:
## Description
Use thread blocks to scan over decimal element sizes in row groups instead of using a single CUDA thread.

Up to 60% performance improvement with decimal columns.

## Checklist


vuule:
64 decimal columns, original profile:
original
with the optimization:
optimized

cpp/src/io/orc/writer_impl.cu
```diff
@@ -1870,6 +1870,60 @@ hostdevice_2dvector<rowgroup_rows> calculate_rowgroup_bounds(orc_table_view cons
...
+  uint32_t col_idx;
+  device_span<uint32_t> sizes;
+  decimal_column_element_sizes(uint32_t col_idx, device_span<uint32_t> sizes)
+    : col_idx(col_idx), sizes(sizes)
+  {
+  }
+};
+
+template <int block_size>
+__global__ void rowgroup_inclusive_scan(device_2dspan<rowgroup_rows const> rg_bounds,
```
vuule, line 1883: need to move this to the file with device code.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -1882,7 +1882,7 @@ encoder_decimal_info decimal_chunk_sizes(orc_table_view& orc_table,
       auto& current_sizes =
         elem_sizes.insert({orc_col.index(), rmm::device_uvector<uint32_t>(orc_col.size(), stream)})
           .first->second;
-      thrust::tabulate(rmm::exec_policy(stream),
+      thrust::tabulate(rmm::exec_policy_nosync(stream),
```
ttnghia, line 1885: When do we know to add `_nosync`?

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1368,6 +1402,27 @@ std::optional<writer_compression_statistics> CompressOrcDataStreams(
...
+void decimal_sizes_to_offsets(device_2dspan<rowgroup_rows const> rg_bounds,
+                              std::map<uint32_t, rmm::device_uvector<uint32_t>>& elem_sizes,
+                              rmm::cuda_stream_view stream)
+{
+  if (rg_bounds.count() == 0) return;
+  std::vector<decimal_column_element_sizes> h_sizes;
+  h_sizes.reserve(elem_sizes.size());
+  for (auto& [col_idx, current_sizes] : elem_sizes) {
+    h_sizes.emplace_back(col_idx, current_sizes);
+  }
```
ttnghia, line 1414: Probably this?
```
auto const h_sizes = std::vector<decimal_column_element_sizes>(elem_sizes.begin(), elem_sizes.end());
```

ttnghia:
So is this basically replacing a thrust call by a custom kernel? It makes me worried much about my thrust calls and whether should I write custom kernel instead :smiling_face_with_tear:

vuule:
> So is this basically replacing a thrust call by a custom kernel? It makes me worried much about my thrust calls and whether should I write custom kernel instead 🥲

This is a special case, with a single thread per row group being replaced with a whole block.
I think one common way we lose performance is by launching a kernel (from thrust usually) per column when processing a table. Using a stream pool is useful in that case, but having a single kernel launch would still be a bit faster.

cpp/src/io/orc/writer_impl.cu
vuule: You can find some information here: https://github.com/rapidsai/cudf/issues/12086
With nosync, thrust::tabulate will not sync at the end of execution. This is fine in this case, as the results are on the GPU, and available for the next kernel in the same stream.
We can use this in many, if not all, places in cuIO.

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1409,9 +1409,9 @@ void decimal_sizes_to_offsets(device_2dspan<rowgroup_rows const> rg_bounds,
   if (rg_bounds.count() == 0) return;
   std::vector<decimal_column_element_sizes> h_sizes;
   h_sizes.reserve(elem_sizes.size());
-  for (auto& [col_idx, current_sizes] : elem_sizes) {
-    h_sizes.emplace_back(col_idx, current_sizes);
-  }
+  std::transform(elem_sizes.begin(), elem_sizes.end(), std::back_inserter(h_sizes), [](auto& p) {
+    return decimal_column_element_sizes{p.first, p.second};
+  });
```
ttnghia, line 1414: This seems even less elegant to me.

vuule: yeah, but it follows the "no raw loops" guideline :)

ttnghia: The for each style is also "no raw loop"  though. As long as you don't use any explicit index, it is good.

vuule: (in theory) transform is better as it's a less general tool.
I'm not defending this piece for code specifically, talking about the guideline.
Should I revert the change?

ttnghia: No need for it. They are about the same, you didn't have to changed to this, and don't have to change back :P

vuule: Glad we agree :D

hyperbolic2346:
> > So is this basically replacing a thrust call by a custom kernel? It makes me worried much about my thrust calls and whether should I write custom kernel instead 🥲
> 
> This is a special case, with a single thread per row group being replaced with a whole block. I think one common way we lose performance is by launching a kernel (from thrust usually) per column when processing a table. Using a stream pool is useful in that case, but having a single kernel launch would still be a bit faster.

I think it's a downside of the abstraction that thrust brings you, which is thrust's power. Recently I wrote a kernel and used thrust calls to do it. It was very simple and easy to understand, but the performance was 10x slower than a hand-rolled kernel. The reason mostly hinged around cross-thread communication/coordination. I wasn't able to do ballots and things between threads to generate larger writes and I was mostly stuck with a single thread per row. Changing my kernel to have a warp per row was a huge win for me. It has made me rethink the times I go towards thrust as well.

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1260,6 +1261,39 @@ __global__ void __launch_bounds__(1024)
   }
 }
 
+struct decimal_column_element_sizes {
+  uint32_t col_idx;
```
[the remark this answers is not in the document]
vuule: I prefer not to mark data members as const, based on
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-constref
https://quuxplusone.github.io/blog/2022/01/23/dont-const-all-the-things/#data-members-never-const

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1368,6 +1402,27 @@ std::optional<writer_compression_statistics> CompressOrcDataStreams(
   }
 }
 
+void decimal_sizes_to_offsets(device_2dspan<rowgroup_rows const> rg_bounds,
```
ttnghia, line 1405: I can't figure out where is the return object? Or the output is overwritten into this `rg_bounds`? Is there any memory allocation that needs `mr`?

ttnghia:
Since decimals are fixed width types, I wonder if there is similar optimization can be applied to other types?

cpp/src/io/orc/stripe_enc.cu
vuule: `elem_sizes` is the output here - we're doing a scan of the element sizes (within each rowgroup) to get the offset in the output stream.
Maybe I should rename to something like `scan_decimal_sizes`?

vuule:
> Since decimals are fixed width types, I wonder if there is similar optimization can be applied to other types?

Decimal types are the only types where we determine the encoded element size before encoding. Leads to a simple path in the encode kernel, but we do have these extra steps (that weren't efficient) and require extra memory to keep offsets.
I'm not aware of similar opportunities in the ORC writer.

ttnghia:
What if we bitcast (reinterpret cast) an int32 column into decimal32 column? Will that column be processed faster with this optimization trick?

vuule:
> What if we bitcast (reinterpret cast) an int32 column into decimal32 column? Will that column be processed faster with this optimization trick?

Great question. Decimal type in ORC does not use RLE so we can encode each element separately (and thus can determine offset for each element). We can't do the same for integral types.
The fact that decimals don't use RLE is why we also have a perf bottleneck on the reader side  https://github.com/rapidsai/cudf/issues/13251

Also, I doubt decimal encode is faster than int encode. It's just less slow after this PR :D

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1260,6 +1261,39 @@ __global__ void __launch_bounds__(1024)
...
+
+  uint32_t initial_value = 0;
+  // do a series of block sums, storing results in the array as we go
+  for (int pos = 0; pos < row_group.size(); pos += block_size) {
+    auto const tidx    = pos + t;
+    auto tval          = tidx < row_group.size() ? elem_sizes[tidx] : 0u;
+    uint32_t block_sum = 0;
+    block_scan(scan_storage).InclusiveSum(tval, tval, block_sum);
+    if (tidx < row_group.size()) { elem_sizes[tidx] = tval + initial_value; }
+    initial_value += block_sum;
```
hyperbolic2346, line 1293: Interesting, I've seen this mostly with the last thread writing this value into shared memory. I didn't know cub offered the block sum to each thread. I wonder if that impacts performance at all.

cpp/src/io/orc/stripe_enc.cu
hyperbolic2346: Is it though? I don't see a sync back from the gpu data in `d_sizes`. Am I missing the trick?

hyperbolic2346 (changes requested):

cpp/src/io/orc/stripe_enc.cu
vuule: I'm not sure. I stumbled around a bit while implementing this, it's possible there is a better way. We do have similar uses in places like `page_bounds` and `gpuUpdatePageSizes`.

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1260,6 +1261,39 @@ __global__ void __launch_bounds__(1024)
...
+  __shared__ typename block_scan::TempStorage scan_storage;
+  int const t = threadIdx.x;
+
+  auto const& col_elem_sizes = sizes[blockIdx.x];
+  auto const& row_group      = rg_bounds[blockIdx.y][col_elem_sizes.col_idx];
+  auto const elem_sizes      = col_elem_sizes.sizes.data() + row_group.begin;
+
+  uint32_t initial_value = 0;
+  // do a series of block sums, storing results in the array as we go
+  for (int pos = 0; pos < row_group.size(); pos += block_size) {
```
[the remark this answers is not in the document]
vuule: Not in practice, but I'll use a long here anyway. Good catch.

cpp/src/io/orc/stripe_enc.cu
vuule: `h_sizes`/`d_sizes` hold `device_span`s of the `device_uvector`s in `elem_sizes`. It's just a mutable view that the kernel uses to convert the sizes in `elem_sizes` into offsets within row groups, using the scan.
`rg_bounds` contain the row bound of each row group (for each column). We don't change this here, just use it correctly scan within row groups.

I'll take these questions as a sign that the code is very unclear :(
Some comments are in order, it seems.

cpp/src/io/orc/stripe_enc.cu
```diff
@@ -1368,6 +1402,27 @@ std::optional<writer_compression_statistics> CompressOrcDataStreams(
   }
 }
 
+void decimal_sizes_to_offsets(device_2dspan<rowgroup_rows const> rg_bounds,
+                              std::map<uint32_t, rmm::device_uvector<uint32_t>>& elem_sizes,
```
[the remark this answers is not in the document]
vuule: no, this is an inout parameter. The sizes are scanned in-place.

cpp/src/io/orc/stripe_enc.cu
hyperbolic2346: Oh, I don't think there is anything wrong with this code at all. I'm just wondering out loud.

They both have to synchronize at some point. One does it before exiting the cub call and one does it at the bottom of the loop. I would think 6 of one type of situation.

cpp/src/io/orc/stripe_enc.cu
vuule: added comments in a few places

hyperbolic2346 (approved):

PR #14202, opened 2023-09-26
Move chars column to parent data buffer in strings column

karthikeyann:
## Description
Eliminates chars column and moves chars data to parent string column's _data buffer.

Summary of changes 
- chars child column is removed, chars buffer is added to parent column
- Adds stream to `chars_size()`, `chars_end()` in `strings_column_view` and their invocations
- Remove `chars_column_index`, and deprecate `chars()` from `strings_column_view`
- Replace `chars_col.begin<char>()` with `static_cast<char*>(parent.head())`
- Adds string column factory which accepts `rmm::device_buffer` instead of chars column
- Deprecate string column factory which accepts chars column
- IO changes - contiguous split (From @nvdbaranec ), to_arrow, parquet writer.
- Fix binary ops, column_view,  interleave columns, byte cast, strings APIs, text APIs
- Fix tests, benchmarks (mostly adding `stream` parameter to chars_size)
- Java fixes (From @andygrove)
- Python changes
  - .data special case for string column
  - get size from offsets column for rmm.DeviceBuffer in column
  - special condition for string slice
  - Pickle file update for string column
  - a few unit tests updates

Preparing for https://github.com/rapidsai/cudf/issues/13733


karthikeyann:
@galipremsagar 6 failing pytests
FAILED test_pickling.py::test_pickle_string_column[slices1] - AssertionError: Series are different
FAILED test_pickling.py::test_pickle_string_column[slices3] - AssertionError: Series are different
FAILED test_serialize.py::test_deserialize_cudf_0_16 - RuntimeError: CUDF failure at: /home/knataraj/dev/rapids/cudf/cpp/src/column/column_view.cpp:57: Null data pointer.
FAILED test_testing.py::test_assert_column_memory_basic_same[arrow_arrays1] - ValueError: Buffer size must be divisible by element size
FAILED test_orc.py::test_writer_timestamp_stream_size - AssertionError: numpy array are different
FAILED test_orc.py::test_orc_writer_negative_timestamp - AssertionError: numpy array are different

cpp/include/cudf/strings/strings_column_view.hpp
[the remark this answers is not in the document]
karthikeyann: I searched and removed all the occurences. I didn't remove it completely yet because at some places `[1]` was used. So, this will be kept as a comment with a warning.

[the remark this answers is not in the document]
jlowe: I'm confused by this comment since we're removing the column at index 1 when creating a strings column, per the changes to make_strings_column.  Do you have a code example where we do not want to remove the use of this identifier yet?

jlowe (approved):
Java approval

GregoryKimball:
Hello @karthikeyann , it looks like the high column count concat is the biggest outstanding performance concern 
image

python/cudf/cudf/_lib/column.pyx
```diff
@@ -106,8 +121,36 @@ cdef class Column:
...
+        # this is true for STRING, STRUCT. how to ensure it's string only?
+        if self.dtype != cudf.api.types.dtype("object"):
+            raise TypeError(
+                "Expected a Strings column, "
+                f"got {self.dtype}"
+            )
+        if self.base_data is None:
+            return None
+        cv = self.view()
+        if cv.num_children() == 0:
```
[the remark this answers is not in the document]
karthikeyann: Not all the time. It may have 0 children when number of rows is zero.

davidwendt:
Most of the `nvbench` source files have this line:
```
```
What do you think about creating `auto stream = cudf::get_default_stream()` variable and use this instead of creating new stream objects in the `chars_size()` calls (other places)?

cpp/benchmarks/string/url_decode.cu
```diff
@@ -67,15 +67,15 @@ auto generate_column(cudf::size_type num_rows, cudf::size_type chars_per_row, do
...
+  auto chars_col  = result_col->mutable_view();
   auto offset_col = result_col->child(cudf::strings_column_view::offsets_column_index).view();
 
   auto engine = thrust::default_random_engine{};
   thrust::for_each_n(thrust::device,
                      thrust::make_zip_iterator(offset_col.begin<cudf::size_type>(),
                                                offset_col.begin<cudf::size_type>() + 1),
                      num_rows,
-                     url_string_generator{chars_col.begin<char>(), esc_seq_chance, engine});
+                     url_string_generator{chars_col.data<char>(), esc_seq_chance, engine});
```
davidwendt, line 78: This looks odd to me. I had not considered how accessing the chars data directly from the strings column would appear. Also, the `data()` function adds the `offset()` value and so this would not generally be correct (although it works here since we know the column is not sliced). I think I might prefer something like this instead:
```
   static_cast<char*>(chars_col.head());
```
Let me know what you think.

cpp/include/cudf/column/column_device_view.cuh
```diff
@@ -443,7 +443,7 @@ class alignas(16) column_device_view : public detail::column_device_view_base {
   __device__ T element(size_type element_index) const noexcept
   {
     size_type index       = element_index + offset();  // account for this view's _offset
-    char const* d_strings = d_children[strings_column_view::chars_column_index].data<char>();
+    char const* d_strings = head<char>();
```
davidwendt, line 446: This seems ok to me.
I was expecting something like `d_strings = static_cast<char const*>(_data);` since we have access the member `_data` here.

cpp/include/cudf/utilities/traits.hpp
```diff
@@ -474,7 +475,7 @@ template <typename T>
 constexpr bool is_rep_layout_compatible()
 {
   return cudf::is_numeric<T>() or cudf::is_chrono<T>() or cudf::is_boolean<T>() or
-         cudf::is_byte<T>();
+         cudf::is_byte<T>() or std::is_same_v<T, char>;
```
davidwendt, line 478: This seems a little hacky. I think I'd prefer casting `head()` rather than allowing someone calling `data<char>()`.

cpp/src/interop/to_arrow.cu
```diff
@@ -295,8 +297,10 @@ std::shared_ptr<arrow::Array> dispatch_to_arrow::operator()<cudf::string_view>(
     return std::make_shared<arrow::StringArray>(
       0, std::move(tmp_offset_buffer), std::move(tmp_data_buffer));
   }
-  auto offset_buffer = child_arrays[0]->data()->buffers[1];
-  auto data_buffer   = child_arrays[1]->data()->buffers[1];
+  auto offset_buffer = child_arrays[strings_column_view::offsets_column_index]->data()->buffers[1];
+  // auto data_buffer   = child_arrays[1]->data()->buffers[1];
+  auto data_buffer =
+    fetch_data_buffer<int8_t>(strings_column_view{input_view}.chars(stream), ar_mr, stream);
```
davidwendt, line 303: Could we have a `fetch_data_buffer` that accepts a pointer, size (e.g. `device_span`) so we don't need to create a `column_view` here?

cpp/src/io/json/write_json.cu
```diff
@@ -19,6 +19,7 @@
  * @brief cuDF-IO JSON writer implementation
  */
 
+#include "cudf/types.hpp"
```
davidwendt, line 22:
```suggestion
```

cpp/include/cudf/utilities/traits.hpp
karthikeyann: Right. This is a hack.
will replace `data<char>()` with static cast `head()` everywhere.

cpp/include/cudf/strings/strings_column_view.hpp
```diff
@@ -107,19 +107,21 @@ class strings_column_view : private column_view {
    * @brief Returns the internal column of chars
    *
    * @throw cudf::logic error if this is an empty column
+   * @param stream CUDA stream used for device memory operations and kernel launches
    * @return The chars column
    */
-  [[nodiscard]] column_view chars() const;
+  [[nodiscard]] column_view chars(rmm::cuda_stream_view stream) const;
```
davidwendt, line 113: This member should be removed. Did you mean to add a stream parameter to `chars_begin()`?

cpp/src/strings/strings_column_view.cpp
```diff
@@ -42,26 +45,30 @@ strings_column_view::offset_iterator strings_column_view::offsets_end() const
...
 }
 
-size_type strings_column_view::chars_size() const noexcept
+size_type strings_column_view::chars_size(rmm::cuda_stream_view stream) const noexcept
 {
   if (size() == 0) return 0;
-  return chars().size();
+  // return chars().size();
+  // TODO
+  return detail::get_value<size_type>(offsets(), offset() + size(), stream);
```
davidwendt, line 60: This is only correct if `offset()==0`
```suggestion
  // chars_size = offsets[offset()+size()] - offsets[offset()];
  auto const adjust = (offset()==0) ? 0 : detail::get_value<size_type>(offsets(), offset(), stream);
  return detail::get_value<size_type>(offsets(), offset() + size(), stream) - adjust;
```

cpp/src/strings/strings_column_view.cpp
```diff
@@ -42,26 +45,30 @@ strings_column_view::offset_iterator strings_column_view::offsets_end() const
...
 }
 
 strings_column_view::chars_iterator strings_column_view::chars_begin() const
 {
-  return chars().begin<char>();
+  return head<char>();
```
davidwendt, line 65:
```
strings_column_view::chars_iterator strings_column_view::chars_begin(rmm::cuda_stream_view stream) const
{
  auto const adjust = (offset()==0) ? 0 : detail::get_value<size_type>(offsets(), offset(), stream);
  return head<char>() + adjust;
}
```

cpp/src/strings/strings_column_view.cpp
```diff
@@ -42,26 +45,30 @@ strings_column_view::offset_iterator strings_column_view::offsets_end() const
...
 }
 
-strings_column_view::chars_iterator strings_column_view::chars_end() const
+strings_column_view::chars_iterator strings_column_view::chars_end(
+  rmm::cuda_stream_view stream) const
 {
-  return chars_begin() + chars_size();
+  return chars_begin() + chars_size(stream);
```
davidwendt, line 71:
```suggestion
  return head<char>() + detail::get_value<size_type>(offsets(), offset() + size(), stream);
```

cpp/benchmarks/text/jaccard.cpp
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include "cudf/utilities/default_stream.hpp"
```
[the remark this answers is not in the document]
karthikeyann: done.

cpp/include/cudf/column/column_device_view.cuh
karthikeyann: replaced with `_data`

cpp/src/interop/to_arrow.cu
karthikeyann: replaced with `device_span`

cpp/src/io/json/write_json.cu
karthikeyann: removed it.

cpp/include/cudf/utilities/traits.hpp
karthikeyann: This hack is removed.

cpp/benchmarks/string/url_decode.cu
karthikeyann: Sounds good. Replaced all occurances with this suggestion.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -106,8 +121,36 @@ cdef class Column:
...
+            return self.base_data
+        # get the size from offset child column (device to host copy)
+        offsets_column_index = 0
+        offset_child_column = cv.child(offsets_column_index)
+        start = 0
+        end = self.base_data.size
+        if offset_child_column.size() != 0:
+            start = get_element(offset_child_column, cv.offset()).value
+            end = get_element(offset_child_column, cv.offset()+cv.size()).value
+        return self.base_data[start:end]
```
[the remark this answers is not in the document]
karthikeyann: @galipremsagar I have to uniquely identify string columns (from struct and other nested columns). How do I do that in Cython column.pyx ?

cpp/src/strings/strings_column_view.cpp
```diff
@@ -42,26 +45,21 @@ strings_column_view::offset_iterator strings_column_view::offsets_end() const
...
-  CUDF_EXPECTS(num_children() > 0, "strings column has no children");
-  return child(chars_column_index);
-}
-
-size_type strings_column_view::chars_size() const noexcept
+size_type strings_column_view::chars_size(rmm::cuda_stream_view stream) const noexcept
 {
   if (size() == 0) return 0;
-  return chars().size();
+  return detail::get_value<size_type>(offsets(), offset() + size(), stream);
```
ttnghia, line 51: This function seems to be called a lot. Can we cache it into a member variable?

karthikeyann: Yes. caching is required. That change is moved to next PR to not overload the work of reviewers of this PR.

cpp/benchmarks/hashing/hash.cpp
```diff
@@ -43,7 +43,7 @@ static void bench_hash(nvbench::state& state)
 
   // collect statistics
   cudf::strings_column_view input(data->get_column(1).view());
-  auto const chars_size = input.chars_size();
+  auto const chars_size = input.chars_size(cudf::get_default_stream());
```
davidwendt, line 46:
```suggestion
  auto const chars_size = input.chars_size(stream);
```
I think we can just use the variable declared at line 41

cpp/src/strings/case.cu
```diff
@@ -227,16 +227,15 @@ std::unique_ptr<column> convert_case(strings_column_view const& input,
...
   // The count_if is faster than any_of or all_of: https://github.com/NVIDIA/thrust/issues/1016
   bool const multi_byte_chars =
-    thrust::count_if(
-      rmm::exec_policy(stream), input.chars_begin(), input.chars_end(), [] __device__(auto chr) {
-        return is_utf8_continuation_char(chr);
-      }) > 0;
+    thrust::count_if(rmm::exec_policy(stream),
+                     input.chars_begin(),
+                     input.chars_end(stream),
+                     [] __device__(auto chr) { return is_utf8_continuation_char(chr); }) > 0;
```
ttnghia, line 233: `proclaim_return_type`?

GregoryKimball:
After @davidwendt made the concat improvement in #14540, now this PR now longer has a performance regression. Please note: There is still a small gain to contiguous split
image

cpp/src/strings/case.cu
karthikeyann: added.

cpp/tests/strings/array_tests.cpp
```diff
@@ -219,7 +220,8 @@ TEST_F(StringsColumnTest, OffsetsBeginEnd)
   scv = cudf::strings_column_view(cudf::slice(input, {1, 5}).front());
   EXPECT_EQ(std::distance(scv.offsets_begin(), scv.offsets_end()),
             static_cast<std::ptrdiff_t>(scv.size() + 1));
-  EXPECT_EQ(std::distance(scv.chars_begin(), scv.chars_end()), 16L);
+  // char data does not have size stored.
+  // EXPECT_EQ(std::distance(scv.chars_begin(), scv.chars_end(cudf::get_default_stream())), 16L);
```
davidwendt, line 224:
```suggestion
```

cpp/src/strings/strings_column_view.cpp
davidwendt, line 51:
```suggestion
  return detail::get_value<size_type>(offsets(), offsets.size()-1, stream);
```

ttnghia: Does `offsets.size()` account for non-zero `offset()`?

davidwendt: No. Offset due to slice/split is not applied to child columns like `offsets()` (as well as `chars()`).
Since `chars_begin()` does not account for parent offset (as mentioned in the doxygen), then `chars_end()` should not as well for consistency (as also documented in the doxygen).

davidwendt (approved):

cpp/include/cudf/column/column_device_view.cuh
```diff
@@ -445,7 +445,7 @@ class alignas(16) column_device_view : public detail::column_device_view_base {
   __device__ T element(size_type element_index) const noexcept
   {
     size_type index       = element_index + offset();  // account for this view's _offset
-    char const* d_strings = d_children[strings_column_view::chars_column_index].data<char>();
+    char const* d_strings = static_cast<char const*>(_data);
```
ttnghia, line 448: Should we change the next line (`offsets`) to be consistent with this?

cpp/include/cudf/strings/strings_column_view.hpp
```diff
@@ -128,21 +131,22 @@ class strings_column_view : private column_view {
    * The offsets child must be used to properly address the char bytes.
    *
    * For example, to access the first character of string `i` (accounting for
-   * a sliced column offset) use: `chars_begin()[offsets_begin()[i]]`.
+   * a sliced column offset) use: `chars_begin(stream)[offsets_begin()[i]]`.
    *
    * @return Iterator pointing to the first char byte.
    */
-  [[nodiscard]] chars_iterator chars_begin() const;
+  [[nodiscard]] chars_iterator chars_begin(rmm::cuda_stream_view) const;
```
ttnghia, line 138: No name parameter? So it actually doesn't need a stream, right? Just added for consistency?

davidwendt: Mostly for consistency but also for possible future use.

cpp/include/cudf/strings/strings_column_view.hpp
```diff
@@ -106,20 +106,23 @@ class strings_column_view : private column_view {
...
    * @brief Returns the internal column of chars
    *
-   * @throw cudf::logic_error if this is an empty column
+   * @throw cudf::logic error if this is an empty column
+   * @param stream CUDA stream used for device memory operations and kernel launches
    * @return The chars column
    */
-  [[nodiscard]] column_view chars() const;
+  [[deprecated]] [[nodiscard]] column_view chars(
+    rmm::cuda_stream_view stream = cudf::get_default_stream()) const;
```
ttnghia, line 114: If it is deprecated, why its signature is changed now? And I also see that its implementation is deleted from the source file.

cpp/include/cudf/column/column_device_view.cuh
davidwendt: The `offsets` still need to be retrieved from the `d_children`. I'm not sure what you are suggesting.

cpp/src/column/column_view.cpp
```diff
@@ -51,7 +51,8 @@ column_view_base::column_view_base(data_type type,
     CUDF_EXPECTS(nullptr == data, "EMPTY column should have no data.");
     CUDF_EXPECTS(nullptr == null_mask, "EMPTY column should have no null mask.");
   } else if (is_compound(type)) {
-    CUDF_EXPECTS(nullptr == data, "Compound (parent) columns cannot have data");
+    if (type.id() != type_id::STRING)
+      CUDF_EXPECTS(nullptr == data, "Compound (parent) columns cannot have data");
```
ttnghia, line 55:
```suggestion
    if (type.id() != type_id::STRING) {
      CUDF_EXPECTS(nullptr == data, "Compound (parent) columns cannot have data");
    }
```

cpp/src/interop/to_arrow.cu
```diff
@@ -49,16 +51,16 @@ namespace {
  * @brief Create arrow data buffer from given cudf column
  */
 template <typename T>
-std::shared_ptr<arrow::Buffer> fetch_data_buffer(column_view input_view,
+std::shared_ptr<arrow::Buffer> fetch_data_buffer(device_span<T const> input_span,
```
ttnghia, line 54: Just call `input`.

cpp/include/cudf/strings/strings_column_view.hpp
davidwendt: Removing this completely will be done in the next release. 
The signature is not _really_ changed since the parameter is defaulted and so should cause no upstream breakages.

cpp/src/io/orc/dict_enc.cu
```diff
@@ -67,7 +67,7 @@ void rowgroup_char_counts(device_2dspan<size_type> counts,
   CUDF_CUDA_TRY(
     cudaOccupancyMaxPotentialBlockSize(&min_grid_size, &block_size, rowgroup_char_counts_kernel));
   auto const grid_size =
-    dim3(cudf::util::div_rounding_up_unsafe<unsigned int>(num_rowgroups, block_size),
+    dim3(cudf::util::div_rounding_up_safe<unsigned int>(num_rowgroups, block_size),
```
ttnghia, line 70: Why changing from `unsafe` to `safe`? A bug?

cpp/src/strings/strings_column_factories.cu
```diff
@@ -131,10 +131,33 @@ std::unique_ptr<column> make_strings_column(size_type num_strings,
...
+
+std::unique_ptr<column> make_strings_column(size_type num_strings,
+                                            std::unique_ptr<column> offsets_column,
+                                            rmm::device_buffer&& chars_buffer,
+                                            size_type null_count,
+                                            rmm::device_buffer&& null_mask)
+{
+  CUDF_FUNC_RANGE();
+
+  if (null_count > 0) CUDF_EXPECTS(null_mask.size() > 0, "Column with nulls must be nullable.");
```
ttnghia, line 150:
```suggestion
  if (null_count > 0) { CUDF_EXPECTS(null_mask.size() > 0, "Column with nulls must be nullable."); }
```

cpp/src/io/orc/dict_enc.cu
karthikeyann: I changed this during a debugging. I left it since it's safe. This may be reverted.

cpp/include/cudf/strings/strings_column_view.hpp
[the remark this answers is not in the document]
karthikeyann: the param is not used, so no name for it. since there is no name, doxygen will throw a warning for param without a name.
`warning: unexpected token in comment block while parsing the argument of command param`
When we skip this, doxygen doesn't mind.

python/cudf/cudf/_lib/column.pyx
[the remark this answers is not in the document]
shwina: I'm surprised we need to implement this in Cython. This kind of logic for getting the chars of a slice must live _somewhere_ in libcudf, right?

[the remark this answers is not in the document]
karthikeyann: Cython column is formed using rmm.DeviceBuffer. So, we don't have column_view. 
Also, there is no existing function for this in string column in libcudf. Should we move this logic to libcudf?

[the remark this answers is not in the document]
davidwendt: What would be moved to `libcudf`? 
The purpose of this PR is to remove the column wrapper for `chars` from libcudf.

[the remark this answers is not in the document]
karthikeyann: `the logic for getting the chars of a slice`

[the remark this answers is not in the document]
davidwendt: How will the `chars of a slice` be returned?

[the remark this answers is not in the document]
shwina: A buffer?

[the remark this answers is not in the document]
davidwendt: Ok, thanks. Just checking.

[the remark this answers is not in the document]
davidwendt: Here is the only place I can think of that libcudf does this internally.
https://github.com/rapidsai/cudf/pull/14202/files#diff-3ab40a1842ddd5ecded7155994a65580ca0bcfad07bbddc2bf17a59df78394eaR71-R72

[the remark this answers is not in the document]
davidwendt: So I think this could be done by exposing the existing `cudf::strings::detail::copy_slice` API as a public API (`cudf::strings::copy_slice`) and then just retrieve the data-buffer from the returned column.

ttnghia (approved):

python/cudf/cudf/_lib/column.pyx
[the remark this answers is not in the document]
karthikeyann: used same logic as `def children(self)`. performs deep copy.

python/cudf/cudf/core/column/string.py
```diff
@@ -5464,6 +5464,7 @@ class StringColumn(column.ColumnBase):
 
     def __init__(
         self,
+        data: Union[Buffer, None] = None,
```
shwina, line 5467: Prefer `Optional[Buffer]` as the type here.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -106,6 +120,18 @@ cdef class Column:
             start = self.offset * self.dtype.itemsize
             end = start + self.size * self.dtype.itemsize
             self._data = self.base_data[start:end]
+            # special case of string column
```
shwina, line 123: At this point, is there anywhere that we're attempting to access the `.data` attribute for a strings column?

If not, I'd prefer we just raise `NotImplementedError` when someone attempts to do that, instead of setting `.data` to be the same as `.base_data` for strings column specially. It seems confusing.

karthikeyann: `.data` is used by`ColumnBase.serialize()` on all type of columns.
https://github.com/rapidsai/cudf/blob/d311a6806d12f6f325880b6a9674c486336a7dc6/python/cudf/cudf/core/column/column.py#L1213

If setting `.data` to be same as `.base_data` is the only concern, we could choose to copy it always.

python/cudf/cudf/core/column/string.py
```diff
@@ -5546,14 +5546,14 @@ def end_offset(self) -> int:
...
-        if len(self.base_children) == 2:
+        if len(self.base_children) == 1:
             child0_size = (self.size + 1) * self.base_children[
                 0
             ].dtype.itemsize
 
             child1_size = (
                 self.end_offset - self.start_offset
-            ) * self.base_children[1].dtype.itemsize
+            ) * cudf.api.types.dtype("int8").itemsize
```
wence-, line 5556: **issue*: This code makes no sense I think, there's one child and one `data` object. Why not:
```
return self.data.memory_usage() + self.base_children[0].memory_usage() + bitmask_usage
```
?

python/cudf/cudf/core/column/string.py
```diff
@@ -5938,15 +5930,15 @@ def view(self, dtype) -> "cudf.core.column.ColumnBase":
         str_end_byte_offset = self.base_children[0].element_indexing(
             self.offset + self.size
         )
-        char_dtype_size = self.base_children[1].dtype.itemsize
+        char_dtype_size = cudf.api.types.dtype("int8").itemsize
```
wence-, line 5933: **issue (throughout)**: Please don't hardcode `"int8"` as the dtype for string column characters everywhere. There must be a data buffer we can obtain it from?

python/cudf/cudf/core/column/string.py
```diff
@@ -5938,15 +5930,15 @@ def view(self, dtype) -> "cudf.core.column.ColumnBase":
...
 
         n_bytes_to_view = (
             str_end_byte_offset - str_byte_offset
         ) * char_dtype_size
 
         to_view = column.build_column(
-            self.base_children[1].data,
-            dtype=self.base_children[1].dtype,
+            self.base_data,  # self.base_children[1].data,
+            dtype=cudf.api.types.dtype("int8"),  # self.base_children[1].dtype,
```
wence-, line 5941: Why not `self.base_data.dtype`?

python/cudf/cudf/tests/test_testing.py
```diff
@@ -431,8 +431,8 @@ def test_assert_column_memory_basic_same(arrow_arrays):
     data = cudf.core.column.ColumnBase.from_arrow(arrow_arrays)
     buf = cudf.core.buffer.as_buffer(data.base_data)
 
-    left = cudf.core.column.build_column(buf, dtype=np.int32)
-    right = cudf.core.column.build_column(buf, dtype=np.int32)
```
wence-, line 435: Why this change?

wence- (changes requested):
Minor changes requested for python, mostly to do with whether there are ways to remove special-casing and not wanting to hardcode `int8` everywhere.

python/cudf/cudf/core/column/string.py
```diff
@@ -5515,7 +5515,7 @@ def copy(self, deep: bool = True):
     def start_offset(self) -> int:
         if self._start_offset is None:
             if (
-                len(self.base_children) == 2
+                len(self.base_children) == 1
```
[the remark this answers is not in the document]
karthikeyann: Then `self._start_offset = 0` (the else code path executes).

python/cudf/cudf/core/column/string.py
karthikeyann: In the older code itself, they don't include the entire second child's full memory. They calculate only the part of memory taken by start_offset and end_offset. 

Now, since we calculate `.data` properly for string column, the above suggestion will work. But `self.data` may incur a deep copy.

python/cudf/cudf/core/column/string.py
karthikeyann: `self.base_data` is cudf.core.buffer. Buffer type. It does not have a dtype.

python/cudf/cudf/tests/test_testing.py
karthikeyann: `data.base_data` is always held None or nullptr for string column. Now this PR changes that to a byte buffer. It may not always hold size that's divisible by `sizeof(int32)`. That's why here it is changed to a type of size 1 byte.

python/cudf/cudf/core/column/string.py
karthikeyann: data buffer is of type `cudf.core.buffer.Buffer`, it does not hold dtype.

python/cudf/cudf/_lib/column.pyx
```diff
@@ -652,11 +678,29 @@ cdef class Column:
...
+        if is_string_column:
+            # get the size from offset child column (device to host copy)
+            offsets_column_index = 0
+            offset_child_column = cv.child(offsets_column_index)
+            if offset_child_column.size() == 0:
+                base_nbytes = 0
+            else:
+                chars_size = get_element(
+                    offset_child_column, offset_child_column.size()-1).value
+                base_nbytes = chars_size * 1
```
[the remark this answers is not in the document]
karthikeyann: Not required. pylibcudf column is plain wrapper around libcudf `cudf::column`. It does not manipulate column's data. Its data is not used so far. So, it does not need this change.

python/cudf/cudf/core/column/column.py
```diff
@@ -1503,17 +1503,11 @@ def column_empty(
...
-        data = None
+        data = as_buffer(
+            rmm.DeviceBuffer(size=row_count * cudf.dtype("int8").itemsize)
+        )
         children = (
             full(row_count + 1, 0, dtype=libcudf.types.size_type_dtype),
-            build_column(
-                data=as_buffer(
-                    rmm.DeviceBuffer(
-                        size=row_count * cudf.dtype("int8").itemsize
```
karthikeyann, line 1512: Question: does `dtype.kind in "OU" and not isinstance(dtype, DecimalDtype):` allow only string type?
I didn't change the logic here. For an empty string column, why do we need a chars buffer of size `row_count` ? it could be an empty buffer since the offsets are all `0`.

python/cudf/cudf/core/column/string.py
karthikeyann: Used `.data` for memory usage calculation

python/cudf/cudf/_lib/column.pyx
karthikeyann: Moved this to string.py. Simplified.

python/cudf/cudf/core/column/string.py
karthikeyann: done.

python/cudf/cudf/core/column/column.py
wence-: In general this condition can also be true for (at least) List and Struct columns. But, those are handled by specific cases above.

From a quick test, I think we don't need an empty chars buffer, so can you try using `rmm.DeviceBuffer(0)`?

python/cudf/cudf/core/column/string.py
```diff
@@ -5567,6 +5565,24 @@ def base_size(self) -> int:
...
+    # override for string column
+    @property
+    def data(self):
+        if self.base_data is None:
+            return None
+        if self._data is None:
+            if (
+                self.offset == 0
+                and len(self.base_children) > 0
+                and self.size == self.base_children[0].size - 1
```
wence-, line 5577: **question**: can it be the case that `base_data is not None`  but yet `len(base_children) == 0`? That seems like an impossible case (because that would mean we have a strings column with char data, but no offset data).

python/cudf/cudf/core/column/string.py
wence-: Ah, ok. Can we introduce (like `cudf._lib.types.size_type_dtype`) a single source of truth for the type of the string char buffer, perhaps `cudf._lib.types.char_type_dtype`?

python/cudf/cudf/tests/test_testing.py
wence-: Thanks!

python/cudf/cudf/core/column/string.py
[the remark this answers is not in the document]
wence-: Thanks!

python/cudf/cudf/_lib/column.pyx
```diff
@@ -52,6 +53,19 @@ from cudf._lib.cpp.scalar.scalar cimport scalar
 from cudf._lib.scalar cimport DeviceScalar
 
 
+cdef get_element(column_view col_view, size_type index):
```
wence-, line 56: I briefly thought this was identical to `copying.pyx:get_element` ,but that takes a `cudf.Column` :(

wence- (approved):
Approving python changes with (non-blocking) suggestion to introduce a single type definition for the char type.

python/cudf/cudf/core/column/string.py
karthikeyann: That case seems impossible. Even if it happens, `self._data = self.base_data[0:0]` will be executed. so, it won't cause an issue.

python/cudf/cudf/_lib/column.pyx
karthikeyann: Yes. That's why I couldn't reuse it.

python/cudf/cudf/core/column/string.py
karthikeyann: `int8` is used only in this place. Ideally we would want to get rid of it, since `int8` is actually a workaround to hold byte data buffer in strings column. In follow up PRs, we will eliminate references of `int8` related to character buffer.

karthikeyann: Here the`int8` indicates a byte data buffer as column. It is not related to `char` type.

davidwendt:
We'll probably want to update the developer's guide once this is merged as well
https://github.com/rapidsai/cudf/blob/branch-24.02/cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md#strings-columns

cpp/doxygen/developer_guide/DEVELOPER_GUIDE.md
```diff
@@ -1197,17 +1197,15 @@ This is related to [Arrow's "Variable-Size List" memory layout](https://arrow.ap
 
 ## Strings columns
 
-Strings are represented in much the same way as lists, except that the data child column is always
-a non-nullable column of `INT8` data. The parent column's type is `STRING` and contains no data,
+Strings are represented as a column with data device buffer which holds all the characters across all the strings packed together and a child offsets column.
+The parent column's type is `STRING` and its data holds all the characters
```
davidwendt, line 1201:
```suggestion
Strings are represented as a column with a data device buffer and a child offsets column.
The parent column's type is `STRING` and its data holds all the characters across all the strings packed together 
```

python/cudf/cudf/core/column/string.py
mroeschke: I think it would be worth adding a `# TODO` comment noting that int8 is a workaround

mroeschke (approved):
Optional comment otherwise LGTM

python/cudf/cudf/core/column/string.py
karthikeyann: removed int8 here.

shwina (approved):

PR #14205, opened 2023-09-27
[Java] Add JNI bindings for `integers_to_hex`

### LINKED ISSUE #14081: [FEA][JNI] Add JNI bindings for cudf::strings::integers_to_hex()
We need a way to convert a string col to a hex string col

libcudf already supports this in convert_integers::integers_to_hex(). We should expose this method to Java

None

razajafri:
## Description
This PR adds a method to ColumnView class to allow for conversion from Integers to hex
closes #14081 

## Checklist

### DIFF
--- a/java/src/main/java/ai/rapids/cudf/ColumnView.java
+++ b/java/src/main/java/ai/rapids/cudf/ColumnView.java
@@ -5235,7 +5235,23 @@ public class ColumnView implements AutoCloseable, BinaryOperable {
   }
 
   /**
-   * Convert this integer column to hexadecimal column
+   * Convert this integer column to hexadecimal column and return a new strings column
+   *
+   * Any null entries will result in corresponding null entries in the output column.
+   *
+   * The output character set is '0'-'9' and 'A'-'F'. The output string width will
+   * be a multiple of 2 depending on the size of the integer type. A single leading
+   * zero is applied to the first non-zero output byte if it is less than 0x10.
+   *
+   * Example:
+   * input = [123, -1, 0, 27, 342718233] 
+   * s = input.toHex()
+   * s is [ '04D2', 'FFFFFFFF', '00', '1B', '146D7719']
+   *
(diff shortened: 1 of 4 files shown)


kuhushukla (approved):
lgtm

java/src/main/java/ai/rapids/cudf/ColumnView.java
```diff
@@ -5231,4 +5233,29 @@ static ColumnView[] getColumnViewsFromPointers(long[] nativeHandles) {
...
+   * s is [ '04D2', 'FFFFFFFF', '00', '1B', '146D7719']
+   *
+   * The example above shows an `INT32` type column where each integer is 4 bytes.
+   * Leading zeros are suppressed unless filling out a complete byte as in
+   * `123 -> '04D2'` instead of `000004D2` or `4D2`. 
+   * 
+   * @return new string ColumnVector
+   */
+  public ColumnVector toHex() {
+    assert getType().isIntegral() : "Only integers are supported";
```
revans2, line 5258: nit: technically I don't think we need this. The native code should handle checking for this and throw an exception if needed.

revans2 (approved):

java/src/main/java/ai/rapids/cudf/ColumnView.java
razajafri: Yes, but the native code is allowing bool8 which will return invalid results.

revans2: That is a CUDF bug and there is a PR up to fix it. https://github.com/rapidsai/cudf/pull/14208  Plus the assert will only run if you enable them. They are not on by default.

razajafri: I understand that. By convention we have been using asserts, do you want me to revert this entire assert and related change?

razajafri: Thank you very much

Issue #13749, opened 2023-07-25
[BUG] Decimal128 uses a fixed precision of 18 when converting to an arrow array


jihoonson:
**Describe the bug**
Hi all. When a decimal128 column is converted to an arrow array, the precision is set to 18 in [this line](https://github.com/rapidsai/cudf/blob/branch-23.08/cpp/src/interop/to_arrow.cu#L203), even though the max precision decimal128 supports is 38. This seems like a bug rather than intentional behavior.

codereport:
So I'm not 💯 percent confident this is the story but I am pretty certain is is the case. TLDR: It was a bug when it was introduced in https://github.com/rapidsai/cudf/pull/9986. 

___

Timeline:

* https://github.com/rapidsai/cudf/pull/5704 was merged Jul 2020
* https://github.com/rapidsai/cudf/pull/7609 was merged Apr 2021 ⬅️ introduced the 18 for decimal64 
* https://github.com/rapidsai/cudf/pull/9483 was merged Nov 2021
* https://github.com/rapidsai/cudf/pull/9986 was merged Apr 2022 ⬅️ introduced the 18 for decimal6128

My guess is this was just a copy and paste issue. Note though that `libcudf` doesn't actually have the concept of precision, but `Python CUDF` does seem to set it to 38 in the case of 128. 

```py
class Decimal32Dtype(DecimalDtype):
    name = "decimal32"
    MAX_PRECISION = np.floor(np.log10(np.iinfo("int32").max))
    ITEMSIZE = 4


@doc_apply(
    decimal_dtype_template.format(
        size="64",
    )
)
class Decimal64Dtype(DecimalDtype):
    name = "decimal64"
    MAX_PRECISION = np.floor(np.log10(np.iinfo("int64").max))
    ITEMSIZE = 8


@doc_apply(
    decimal_dtype_template.format(
        size="128",
    )
)
class Decimal128Dtype(DecimalDtype):
    name = "decimal128"
    MAX_PRECISION = 38
    ITEMSIZE = 16
```

My suggestion is to "fix" it, and then write a few tests to see if it works when roundtripping. IIRC, we had planned to test most of the overflow  / close to overflow use cases from the Python side of things cause it was much easier that way. Probably why it wasn't caught immediately in the C++ test cases (as they are quite bare bones)

jihoonson:
Thanks @codereport. I created https://github.com/rapidsai/cudf/pull/14230 to fix this issue.

PR #14230, opened 2023-09-28
Fix the precision when converting a decimal128 column to an arrow array

jihoonson:
## Description

This PR fixes https://github.com/rapidsai/cudf/issues/13749. As discussed in the issue linked, the precision is unnecessarily being limited to 18 when converting decimal128 to arrow. 

Implementation-wise, I wasn't sure where is the best place to define the max precision for decimal types. Given that the decimal types don't store the precision in libcudf, I thought it would be better to not expose the max precision to the outside of to-arrow conversion. However, this led to replicating the definition of max precision across multiple places. Appreciate any suggestion.

Finally, it was suggested in https://github.com/rapidsai/cudf/issues/13749#issuecomment-1686406749 to add some tests for round tripping. The existing tests look sufficient to me for round trip tests, so I just modified them instead of adding new tests. Please let me know if we need new tests in addition to them.

I'm also not sure whether any documentation should be fixed. Please let me know.

## Checklist


jihoonson:
Uh, should I create this PR against another branch than the `main`?

davidwendt:
> Uh, should I create this PR against an other branch than the `main`?

The PR should go against `branch-23.12`

jihoonson:
Thanks @davidwendt. Working on the rebase and conflict.

vyasr:
@jihoonson thanks for the PR! Style checks are currently failing. Could you please run our pre-commit hooks locally on your system to fix those? If you don't have those set up, the steps are (I'm making no assumptions about your environment, you may want to change how you install pre-commit if you use conda for instance):
```
cd /path/to/repo
pip install pre-commit
python -m pre-commit install
pre-commit run --all-files
```
Then commit and push those changes.

> Implementation-wise, I wasn't sure where is the best place to define the max precision for decimal types. Given that the decimal types don't store the precision in libcudf, I thought it would be better to not expose the max precision to the outside of to-arrow conversion. However, this led to replicating the definition of max precision across multiple places. Appreciate any suggestion.

That's fine. Since as you say libcudf doesn't have a concept of the precision the only real place it makes sense to include is in the conversion function. I wouldn't sweat the mild duplication.

> Finally, it was suggested in [#13749 (comment)](https://github.com/rapidsai/cudf/issues/13749#issuecomment-1686406749) to add some tests for round tripping. The existing tests look sufficient to me for round trip tests, so I just modified them instead of adding new tests. Please let me know if we need new tests in addition to them.

I agree with your approach, modifying is fine.

> I'm also not sure whether any documentation should be fixed. Please let me know.

At the moment this behavior is not specifically documented. What might be nice is to add a note to the [`to_arrow` C++ documentation](https://github.com/rapidsai/cudf/blob/branch-23.12/cpp/include/cudf/interop.hpp#L123) indicating that decimal types will be converted to the widest precision supported by that type. Perhaps an [`@note` section](https://www.doxygen.nl/manual/commands.html#cmdnote) would be good.

cpp/src/interop/to_arrow.cu
```diff
@@ -220,6 +220,7 @@ std::shared_ptr<arrow::Array> dispatch_to_arrow::operator()<numeric::decimal128>
   rmm::cuda_stream_view stream)
 {
   using DeviceType = __int128_t;
+  auto constexpr max_precision = 38;
```
ttnghia, line 223: Hard coding the number of max precision is a bit confusing because somebody may not know at all "where is the number (18) 38 coming from?". 

It makes more sense to formulate it, like:
```
template<typename T>
constexpr std::size_t max_precision() {
  auto const num_bits = sizeof(T) * 8;
  return std::floor(num_bits * std::log(2) / std::log(10));
}
```

cpp/tests/interop/to_arrow_test.cpp
```diff
@@ -355,6 +355,9 @@ TEST_F(ToArrowTest, StructColumn)
 template <typename T>
 using fp_wrapper = cudf::test::fixed_point_column_wrapper<T>;
 
+auto constexpr decimal64_max_precision = 18;
+auto constexpr decimal128_max_precision = 38;
```
ttnghia, line 359: Same as my comment above. Just define a template function.

cpp/src/interop/to_arrow.cu
jihoonson: Good idea. Added a util function as suggested and replaced the hard-coded max precision with calling the util function.

jihoonson:
Thanks @vyasr and @ttnghia for the review! I addressed your comments in the last 3 commits.

hyperbolic2346 (approved):

ttnghia (approved):

jihoonson:
Thanks @hyperbolic2346 and @ttnghia! What will be the next step? Please let me know if there is anything I need to do :slightly_smiling_face:

jihoonson:
Hmm the codecov failed for [python/cudf/cudf/io/orc.py](https://app.codecov.io/gh/rapidsai/cudf/commit/21d9f3266267c705b9059e4af3e09c3c448ff1da/blob/python/cudf/cudf/io/orc.py), which seems unrelated to my change in this PR.

Issue #14209, opened 2023-09-27
[FEA] Improve exception message when unknown Parquet page encoding detected


jlowe:
**Is your feature request related to a problem? Please describe.**
A user of the RAPIDS Accelerator for Apache Spark reported the following exception:
```
ai.rapids.cudf.CudfException: CUDF failure at: /home/jenkins/agent/workspace/jenkins-spark-rapids-jni_nightly-dev-424-cuda11/thirdparty/cudf/cpp/src/io/parquet/reader_impl_preprocess.cu:346: Unsupported page encoding detected
	at ai.rapids.cudf.ParquetChunkedReader.readChunk(Native Method)
	at ai.rapids.cudf.ParquetChunkedReader.readChunk(ParquetChunkedReader.java:111)
	at com.nvidia.spark.rapids.ParquetTableReader.$anonfun$next$1(GpuParquetScan.scala:2639)
	at com.nvidia.spark.rapids.Arm$.withResource(Arm.scala:29)
	at com.nvidia.spark.rapids.ParquetTableReader.next(GpuParquetScan.scala:2638)
	at com.nvidia.spark.rapids.ParquetTableReader.next(GpuParquetScan.scala:2615)
	at com.nvidia.spark.rapids.CachedGpuBatchIterator$.$anonfun$apply$1(GpuDataProducer.scala:161)
	at com.nvidia.spark.rapids.Arm$.withResource(Arm.scala:29)
	at com.nvidia.spark.rapids.CachedGpuBatchIterator$.apply(GpuDataProducer.scala:158
```

The exception message from libcudf is not very helpful in that it says an unsupported page encoding was detected but not what that unexpected page encoding was (i.e.: the enum value).  Without this information, we're left guessing what encoding was found in the file and usually have to request users to share a sample file to find out.  Not all users are willing to share sample files.

**Describe the solution you'd like**
Exception messages for an unexpected/unsupported value should show the value as part of the exception message.

ZelboK:
Hi, I'd like to try doing this task.

GregoryKimball:
Thank you @ZelboK. It's possible that https://github.com/rapidsai/cudf/pull/14237 partially or fully addressed this. @etseidl would you please share your thoughts?

etseidl:
[...]

PR #14237, opened 2023-09-29
Detect and report errors in Parquet header parsing

### LINKED ISSUE #13656: [BUG] libcudf fails to recognized malformed dictionary during Parquet read

etseidl:
## Description
Fixes #13656.  Uses the error reporting introduced in #14167 to report errors in header parsing.

## Checklist


etseidl:
Note: we could also detect unsupported encodings earlier, but then that would make addressing #14209 a bit harder.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -967,13 +972,31 @@ void reader::impl::load_and_decompress_data()
...
   }
 
+  rmm::device_scalar<int32_t> error_code(0, _stream);
+
   // Process dataset chunk pages into output columns
-  auto const total_pages = count_page_headers(chunks, _stream);
+  auto const total_pages = count_page_headers(chunks, error_code.data(), _stream);
   if (total_pages <= 0) { return; }
+
+  auto decode_error = error_code.value(_stream);
```
vuule, line 981: do we need to check the error twice?

etseidl: Well, the same kernel is called twice. I suppose we could initialize the error code to 0 outside the kernel, and then since errors are OR'd in we could check once. After both are called. Or just exit early from the second call if error is non zero.

vuule: the error is initialized outside of the kernel. I think having a single check should work, as long as we're not worried about a segfault in decode_page_headers if there's an error in count_page_headers.

etseidl: hmm, so the code is initialized, so no worries there.  but if something goes wrong in count_page_headers and it returns 0, then we'll just return without the error. So we kind of have to check after count_page_headers.  Right now I don't think there's an error that won't happen in the first call, that can happen in the second, but down the road...

I think it's safest to just leave both in (unless it turns out that pulling back the error code is expensive).

vuule: Let's assume error checking is cheap and keep both checkpoints:
1) Please move the common code to a util function (we got three copies now)?
2) Please run benchmarks with the new checks.

cpp/src/io/parquet/error.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include <rmm/device_scalar.hpp>
+
+#include <sstream>
+
+namespace cudf::io::parquet {
+
+namespace {
+
+rmm::device_scalar<int32_t> error_code(0, cudf::get_default_stream());
```
vuule, line 29: I don't think this should be global.

Suggestion:
class kernel_error{
rmm::device_scalar<int32_t> error_code;
public:
kernel_error(stream); // inits to zero
int32_t data();
int32_t value(); // use same stream as in ctor?
void reset(); // use same stream as in ctor?
operator std::string() const; // explicit?
};

etseidl: I originally went with global thinking we could avoid some memsets to 0. Initialize one at startup and then assume it stays zero since non zero will halt execution. But that didn't account for multiple invocations as I found while running the python tests. No need for it to be global now, since it has to be reset to zero before every use. Now it's just the time to allocate and free 4 bytes of device memory, but with pooled memory that shouldn't be a big deal.

vuule: I'm scared of going global because of multi-threaded use cases. Opens the solution to many new challenges.

etseidl:
Here's the `parquet_read_decode` benchmark for some selected types.  Overall a little better, due mainly to getting rid of an unnecessary DtoH copy of the pages.
```
|  data_type  |      io       |  cardinality  |  run_length  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |        Diff |   %Diff |  Status  |
|-------------|---------------|---------------|--------------|------------|-------------|------------|-------------|-------------|---------|----------|
|  INTEGRAL   | DEVICE_BUFFER |       0       |      1       |  14.130 ms |       1.58% |  13.814 ms |       1.67% | -316.585 us |  -2.24% |   FAIL   |
|  INTEGRAL   | DEVICE_BUFFER |     1000      |      1       |  13.284 ms |       0.41% |  13.025 ms |       1.30% | -258.603 us |  -1.95% |   FAIL   |
|  INTEGRAL   | DEVICE_BUFFER |       0       |      32      |  11.915 ms |       1.53% |  11.661 ms |       0.50% | -253.702 us |  -2.13% |   FAIL   |
|  INTEGRAL   | DEVICE_BUFFER |     1000      |      32      |  11.381 ms |       0.38% |  11.210 ms |       1.50% | -170.723 us |  -1.50% |   FAIL   |
|    FLOAT    | DEVICE_BUFFER |       0       |      1       |   8.413 ms |       1.96% |   8.283 ms |       0.54% | -130.085 us |  -1.55% |   FAIL   |
|    FLOAT    | DEVICE_BUFFER |     1000      |      1       |   8.349 ms |       0.42% |   8.168 ms |       1.22% | -180.579 us |  -2.16% |   FAIL   |
|    FLOAT    | DEVICE_BUFFER |       0       |      32      |   7.095 ms |       0.49% |   6.950 ms |       1.65% | -144.718 us |  -2.04% |   FAIL   |
|    FLOAT    | DEVICE_BUFFER |     1000      |      32      |   7.312 ms |       2.40% |   7.184 ms |       1.61% | -128.323 us |  -1.75% |   FAIL   |
|   STRING    | DEVICE_BUFFER |       0       |      1       |  45.079 ms |       1.53% |  44.930 ms |       1.55% | -149.336 us |  -0.33% |   PASS   |
|   STRING    | DEVICE_BUFFER |     1000      |      1       |  14.408 ms |       1.64% |  14.313 ms |       1.54% |  -95.200 us |  -0.66% |   PASS   |
|   STRING    | DEVICE_BUFFER |       0       |      32      |  45.644 ms |       1.42% |  45.160 ms |       1.40% | -483.791 us |  -1.06% |   PASS   |
|   STRING    | DEVICE_BUFFER |     1000      |      32      |  14.297 ms |       1.85% |  14.235 ms |       1.89% |  -61.915 us |  -0.43% |   PASS   |
|    LIST     | DEVICE_BUFFER |       0       |      1       |  76.528 ms |       1.94% |  76.663 ms |       1.10% |  134.970 us |   0.18% |   PASS   |
|    LIST     | DEVICE_BUFFER |     1000      |      1       |  63.407 ms |       1.48% |  63.571 ms |       1.38% |  164.160 us |   0.26% |   PASS   |
|    LIST     | DEVICE_BUFFER |       0       |      32      |  53.909 ms |       1.72% |  53.933 ms |       1.83% |   24.582 us |   0.05% |   PASS   |
|    LIST     | DEVICE_BUFFER |     1000      |      32      |  54.842 ms |       1.96% |  54.926 ms |       1.70% |   84.101 us |   0.15% |   PASS   |
```

cpp/src/io/parquet/error.hpp
```diff
@@ -0,0 +1,77 @@
...
+   * @throws `rmm::cuda_error` if copying `initial_value` to device memory fails.
+   *
+   * @param CUDA stream to use
+   */
+  kernel_error(rmm::cuda_stream_view stream) : _error_code{0, stream} {}
+
+  /**
+   * @brief Return a pointer to the device memory for the error
+   */
+  auto data() { return _error_code.data(); }
```
vuule, line 54:
```suggestion
  [[nodiscard]] auto data() { return _error_code.data(); }
```

cpp/src/io/parquet/error.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * @brief Return the current value of the error
+   *
+   * This uses the stream used to create this instance. This does a synchronize on the stream
+   * this object was instantiated with.
+   */
+  auto value() const { return _error_code.value(_error_code.stream()); }
```
vuule, line 62:
```suggestion
  [[nodiscard]] auto value() const { return _error_code.value(_error_code.stream()); }
```

cpp/src/io/parquet/error.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * @brief Return a hexadecimal string representation of the current error code
+   *
+   * Returned string will have "0x" prepended.
+   */
+  std::string str() const
```
vuule, line 69:
```suggestion
  [[nodiscard]] std::string str() const
```

vuule (approved):
a utility is born!
Looks good, just a few minor suggestions.

cpp/src/io/parquet/error.hpp
```diff
@@ -0,0 +1,77 @@
...
+ */
+class kernel_error {
+ private:
+  rmm::device_scalar<int32_t> _error_code;
+
+ public:
+  /**
+   * @brief Construct a new `kernel_error` with an initial value of 0.
+   *
+   * Note: setting the initial value is done asynchronously.
```
bdice, line 42:
```suggestion
   * Note: the initial value is set asynchronously.
```

cpp/src/io/parquet/page_hdr.cu
```diff
@@ -356,16 +340,19 @@ struct gpuParsePageHeader {
...
+  gpuDecodePageHeaders(ColumnChunkDesc* chunks, int32_t num_chunks, int32_t* error_code)
 {
   gpuParsePageHeader parse_page_header;
   __shared__ byte_stream_s bs_g[4];
 
+  int error[4]            = {0};
   int lane_id             = threadIdx.x % 32;
-  int chunk               = (blockIdx.x * 4) + (threadIdx.x / 32);
-  byte_stream_s* const bs = &bs_g[threadIdx.x / 32];
+  int warp_id             = threadIdx.x / 32;
```
bdice, line 350: While refactoring this, let's move to use `cudf::detail::warp_size` instead of hardcoding `32`.

cpp/src/io/parquet/page_hdr.cu
```diff
@@ -356,16 +340,19 @@ struct gpuParsePageHeader {
  */
 // blockDim {128,1,1}
 __global__ void __launch_bounds__(128)
-  gpuDecodePageHeaders(ColumnChunkDesc* chunks, int32_t num_chunks)
+  gpuDecodePageHeaders(ColumnChunkDesc* chunks, int32_t num_chunks, int32_t* error_code)
 {
   gpuParsePageHeader parse_page_header;
   __shared__ byte_stream_s bs_g[4];
 
+  int error[4]            = {0};
```
bdice, line 348: Do we need to be explicit about `int` vs. `int32_t`, for the purpose of aligning our conventions? I'd like to use the same type for the `error` value as the type we use in the parameters of `set_error`.

bdice (commented):
Thanks for this. I have a few relatively small requests.

cpp/src/io/parquet/page_hdr.cu
```diff
@@ -412,6 +403,9 @@ __global__ void __launch_bounds__(128)
         bs->page.lvl_bytes[level_type::DEFINITION] = 0;
         bs->page.lvl_bytes[level_type::REPETITION] = 0;
         if (parse_page_header(bs) && bs->page.compressed_page_size >= 0) {
+          if (not is_supported_encoding(bs->page.encoding)) {
+            error[warp_id] |= static_cast<int>(decode_error::UNSUPPORTED_ENCODING);
```
bdice, line 407: Do these also need to match the type of `int32_t error[4]`? This cast occurs in at least two places that I can see.
```suggestion
            error[warp_id] |= static_cast<int32_t>(decode_error::UNSUPPORTED_ENCODING);
```

cpp/src/io/parquet/page_hdr.cu
```diff
@@ -440,20 +434,25 @@ __global__ void __launch_bounds__(128)
           }
           bs->page.page_data = const_cast<uint8_t*>(bs->cur);
           bs->cur += bs->page.compressed_page_size;
+          if (bs->cur > bs->end) {
+            error[warp_id] |= static_cast<int>(decode_error::DATA_STREAM_OVERRUN);
```
bdice, line 438:
```suggestion
            error[warp_id] |= static_cast<int32_t>(decode_error::DATA_STREAM_OVERRUN);
```

bdice (approved):
Great! Just a couple casts needed, I think.

cpp/src/io/parquet/page_hdr.cu
etseidl: yep, they sure do. thanks!

PR #14242, opened 2023-10-02
Fix inaccurate ceil/floor and inaccurate rescaling casts of fixed-point values.

bdice:
## Description
This is a follow-up PR to #14233. This PR fixes a bug where floating-point values were used as intermediates in ceil/floor unary operations and cast operations that require rescaling for fixed-point types, giving inaccurate results.

See also:
- https://github.com/rapidsai/cudf/pull/14233#discussion_r1340786769
- https://github.com/rapidsai/cudf/issues/14243

## Checklist


harrism:
Why in a subsequent release? Why not just fix them now?

bdice:
> Why in a subsequent release? Why not just fix them now?

This PR is not ready for review yet, I'm still moving things around. I'm targeting a fix for the problems with floating precision loss in fixed point operations and tests that catch these bugs for 23.10. Those fixes and tests will go in this PR. The refactoring of intpow operations across libcudf will target 23.12 after this PR and #14233 are forward-merged.

cpp/src/unary/cast_ops.cu
```diff
@@ -199,7 +199,13 @@ std::unique_ptr<column> rescale(column_view input,
       }
       return output_column;
     }
-    auto const scalar = make_fixed_point_scalar<T>(std::pow(10, -diff), scale_type{diff}, stream);
+
+    RepType scalar_value = 10;
```
hyperbolic2346, line 203: Is the next step to consolidate this logic into a single function somewhere?

hyperbolic2346 (approved):

cpp/src/unary/math_ops.cu
```diff
@@ -295,7 +295,11 @@ std::unique_ptr<column> unary_op_with(column_view const& input,
     input.type(), input.size(), copy_bitmask(input, stream, mr), input.null_count(), stream, mr);
 
   auto out_view = result->mutable_view();
-  Type const n  = std::pow(10, -input.type().scale());
+
+  Type n = 10;
+  for (int i = 1; i < -input.type().scale(); ++i) {
+    n *= 10;
```
vuule, line 301: do we need to worry about overflow?

cpp/src/unary/cast_ops.cu
bdice: Yes. See #14243.

cpp/src/unary/math_ops.cu
bdice: No, I don't think so. We know that the value `10^-scale` is representable by the device storage type (e.g. `__int128_t` or `int64_t`). And in the worst case, this won't introduce a _new_ overflow if one already existed.

PR #14250, opened 2023-10-04
Reimplement `cudf::merge` for nested types without using comparators

divyegala:
## Description
Part of #11844 

This PR also uses new experimental comparators for non-nested types by introducing a new device constructor for `cudf::experimental::row::lexicographic::device_row_comparator`. In the case of non-nested types, preprocessing can be skipped so comparators can be created on the fly. This solution helps us avoid creating 3 comparator types because `thrust::merge` can call the operator with indices from either side of the table.

Furthermore, the PR reworks `cudf/detail/merge.cuh` by removing any CUDA headers/components to expose a true detail API of the form `cudf/detail/merge.hpp`.

[Benchmark comparison for non-nested types](https://github.com/rapidsai/cudf/pull/14250#issuecomment-1747620300)

Compilation time increases from ~6 mins to ~7 mins.

## Checklist


divyegala:
```
(cudf_10_04) dgala@dgx05:~/cudf$ python ../benchmark/tools/compare.py benchmarks base_merge.json merge_new_algo.json 
Comparing base_merge.json to merge_new_algo.json
Benchmark                                          Time             CPU      Time Old      Time New       CPU Old       CPU New
-------------------------------------------------------------------------------------------------------------------------------
Merge/pow2tables/2/manual_time                  +0.8867         +0.8072             0             0             0             0
Merge/pow2tables/4/manual_time                  +1.0211         +0.9913             1             1             1             1
Merge/pow2tables/8/manual_time                  +1.1014         +1.0894             2             4             2             4
Merge/pow2tables/16/manual_time                 +1.1943         +1.1890             4            10             4            10
Merge/pow2tables/32/manual_time                 +1.2540         +1.2513             9            21            10            21
Merge/pow2tables/64/manual_time                 +1.3223         +1.3211            22            52            22            52
Merge/pow2tables/128/manual_time                +1.3936         +1.3930            52           124            52           124
OVERALL_GEOMEAN                                 +1.1613         +1.1404             0             0             0             0
```

cpp/src/merge/merge.cu
```diff
@@ -187,42 +190,83 @@ index_vector generate_merged_indices(table_view const& left_table,
...
+  auto left_indices_mutable_begin = left_indices_mutable.begin<cudf::size_type>();
+
+  auto const total_counter = thrust::make_counting_iterator(0);
+  thrust::for_each(
+    rmm::exec_policy_nosync(stream),
+    total_counter,
+    total_counter + total_size,
+    [merged = merged_indices.data(), left = left_indices_begin, left_size, right_size] __device__(
+      auto const idx) {
+      if (idx < right_size) {
```
bdice, line 255:
```suggestion
      // We split threads into two groups.
      // Threads in [0, right_size) will insert right indices in sorted order.
      // Threads in [right_size, total_size) will insert left indices in sorted order.
      if (idx < right_size) {
```

cpp/src/merge/merge.cu
```diff
@@ -418,8 +462,15 @@ table_ptr_type merge(cudf::table_view const& left_table,
...
-    index_left_view, index_right_view, column_order, null_precedence, nullable, stream);
+  index_vector merged_indices(0, stream);
+  if (cudf::detail::has_nested_columns(left_table) or
+      cudf::detail::has_nested_columns(right_table)) {
+    merged_indices = generate_merged_indices_nested(
+      index_left_view, index_right_view, column_order, null_precedence, nullable, stream);
+  } else {
+    merged_indices = generate_merged_indices(
+      index_left_view, index_right_view, column_order, null_precedence, nullable, stream);
+  }
```
bdice, line 473: Use an IIFE to avoid requiring a stream-ordered constructor. This also lets us mark the result as `const`.
```suggestion
  auto const merged_indices = [&](){
    if (cudf::detail::has_nested_columns(left_table) or
        cudf::detail::has_nested_columns(right_table)) {
      return generate_merged_indices_nested(
        index_left_view, index_right_view, column_order, null_precedence, nullable, stream);
    } else {
      return generate_merged_indices(
        index_left_view, index_right_view, column_order, null_precedence, nullable, stream);
    }
  }();
```

bdice (commented):
This is great. I have a couple early suggestions on the draft.

cpp/src/merge/merge.cu
```diff
@@ -45,8 +53,57 @@
...
+    CUDF_EXPECTS(_lhs.num_columns() == _rhs.num_columns(), "Mismatched number of columns.");
+  }
+
+  __device__ bool operator()(index_type lhs_tagged_index,
+                             index_type rhs_tagged_index) const noexcept
+  {
+    auto const [l_side, l_indx] = lhs_tagged_index;
+    auto const [r_side, r_indx] = rhs_tagged_index;
+
+    // Not sure why `const_cast` is needed here
```
bdice, line 77: Does this throw a compiler error without the `const_cast`? Do we need to mark this as `TODO`?

cpp/src/merge/merge.cu
```diff
@@ -187,18 +244,31 @@ index_vector generate_merged_indices(table_view const& left_table,
...
+
   auto lhs_device_view = table_device_view::create(left_table, stream);
   auto rhs_device_view = table_device_view::create(right_table, stream);
 
   auto d_column_order = cudf::detail::make_device_uvector_async(
     column_order, stream, rmm::mr::get_current_device_resource());
 
-  if (nullable) {
+  if (has_nulls) {
+    auto new_null_precedence = [&]() {
```
bdice, line 257: Can this be `const`?
```suggestion
    auto const new_null_precedence = [&]() {
```

cpp/src/merge/merge.cu
bdice: What if you change the member `table_device_view _lhs` to be `const`, and change `ptr_[left|right]_dview` to be `const`, too? Pretty much everything in the constructor, members, etc. can be `const`.

bdice (commented):
Awesome work. I have a few comments.

cpp/benchmarks/sort/rank_lists.cpp
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#include "nested_types_common.hpp"
 #include "rank_types_common.hpp"
+#include <benchmarks/common/generate_nested_types.hpp>
```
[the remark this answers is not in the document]
divyegala: The developer guide recommends using `#include ""` pattern only when the header in question is in the same directory as the source including the header. So this looks correct to me as-is.

cpp/src/merge/merge.cu
divyegala: That worked, although I had no idea why it didn't work previously without the `const_cast` (there indeed were compiler errors)

cpp/benchmarks/sort/rank_lists.cpp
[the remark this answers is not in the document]
bdice: Ah. Of course. `benchmarks/sort` is not the same directory as `benchmarks/common`. That would've helped me to recognize. 😅

cpp/src/merge/merge.cu
```diff
@@ -45,8 +53,51 @@
...
 
+template <bool has_nulls>
+struct row_lexicographic_tagged_comparator {
+  row_lexicographic_tagged_comparator(table_device_view const lhs,
+                                      table_device_view const rhs,
+                                      device_span<order const> const column_order,
+                                      device_span<null_order const> const null_precedence)
+    : _lhs{lhs}, _rhs{rhs}, _column_order{column_order}, _null_precedence{null_precedence}
+  {
+    // Add check for types to be the same.
```
bdice, line 67: Is this comment a TODO for this PR?

bdice (approved):
Approving. I had one comment about whether a comment is really a TODO or not, but otherwise LGTM.

cpp/src/merge/merge.cu
divyegala: Oh it was already there before my changes, but now I realize it's a relevant comment still. I'll add fulfill this TODO.

divyegala: Okay looks like these checks were already performed in the call stack before this

cpp/include/cudf/merge.hpp
```diff
@@ -34,7 +34,10 @@ namespace cudf {
  * @brief Merge a set of sorted tables.
  *
  * Merges sorted tables into one sorted table
- * containing data from all tables.
+ * containing data from all tables. The key columns
+ * of each table must be sorted according to the
+ * parameters (cudf::column_order and cudf::null_order)
+ * specified for that column.
  *
```
mythrocks, line 41: :+1:

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -305,6 +309,44 @@ class device_row_comparator {
...
+   * @param rhs The second table (may be the same table as `lhs`)
+   * @param column_order Optional, device array the same length as a row that indicates the desired
+   * ascending/descending order of each column in a row. If `nullopt`, it is assumed all columns are
+   * sorted in ascending order.
+   * @param null_precedence Optional, device array the same length as a row and indicates how null
+   * values compare to all other for every column. If `nullopt`, then null precedence would be
+   * `null_order::BEFORE` for all columns.
+   * @param comparator Physical element relational comparison functor.
+   */
+  template <bool nested_disable = not has_nested_columns, CUDF_ENABLE_IF(nested_disable)>
```
mythrocks, line 330: Interesting phrasing. I'll remember this for when I need it.

Is `nested_disable` used elsewhere?
If not, would `CUDF_ENABLE_IF(not has_nested_columns)` be less readable?

cpp/include/cudf/table/experimental/row_operators.cuh
```diff
@@ -370,6 +414,13 @@ class device_row_comparator {
                              std::numeric_limits<int>::max());
     }
 
+    /**
+     * @brief Throws run-time error when columns types cannot be compared
+     *        or if this class is instantiated with `has_nested_columns = false` but
+     *        passed tables with nested columns
+     *
+     * @return Ordering
+     */
```
mythrocks, line 423: :+1: Thank you for adding docs for these overloads.

cpp/src/merge/merge.cu
```diff
@@ -45,8 +53,49 @@
...
+      if (has_nulls) {
+        return cudf::experimental::row::lexicographic::device_row_comparator<false, bool>{
+          has_nulls, *ptr_left_dview, *ptr_right_dview, _column_order, _null_precedence};
+      } else {
+        return cudf::experimental::row::lexicographic::device_row_comparator<false, bool>{
+          has_nulls, *ptr_left_dview, *ptr_right_dview, _column_order};
+      }
+    }();
+
+    auto weak_order = comparator(l_indx, r_indx);
```
mythrocks, line 87: Should we consider adding `const` to the definitions of `comparator` and `weak_order`?

mythrocks (commented):
This is looking very good. I have a couple of trivial nitpicks.

I'm not done. I've just hit the meat of the change. I'll resume the review shortly.

cpp/src/merge/merge.cu
```diff
@@ -45,8 +53,49 @@
...
+  __device__ bool operator()(index_type lhs_tagged_index,
+                             index_type rhs_tagged_index) const noexcept
+  {
+    auto const [l_side, l_indx] = lhs_tagged_index;
+    auto const [r_side, r_indx] = rhs_tagged_index;
+
+    table_device_view const* ptr_left_dview{l_side == side::LEFT ? &_lhs : &_rhs};
+    table_device_view const* ptr_right_dview{r_side == side::LEFT ? &_lhs : &_rhs};
+    auto comparator = [&]() {
+      if (has_nulls) {
```
mythrocks, line 78: Silly question, for my edumacation: Given that `has_nulls` is `constexpr`, is there any value at all making this `if` an `if constexpr`? My guess is no, but I'd like to confirm.

By my understanding:
1. `if constexpr` is to be used where either the `if` or `else` branch might not compile, depending on the value of `has_nulls`.
2. The advantage of "optimizing away the `if`-check" is independent of whether this is a `constexpr` check.

Just confirming my understanding here.

cpp/src/merge/merge.cu
```diff
@@ -223,6 +285,56 @@ index_vector generate_merged_indices(table_view const& left_table,
...
+index_vector generate_merged_indices_nested(table_view const& left_table,
+                                            table_view const& right_table,
+                                            std::vector<order> const& column_order,
+                                            std::vector<null_order> const& null_precedence,
+                                            bool nullable,
+                                            rmm::cuda_stream_view stream)
+{
+  size_type const left_size  = left_table.num_rows();
+  size_type const right_size = right_table.num_rows();
+  size_type const total_size = left_size + right_size;
```
mythrocks, line 297: Another impertinent question: Do we need to protect against the case where the `total_size` overflows?

cpp/include/cudf/table/experimental/row_operators.cuh
divyegala: I can change the variable name if you'd like, but I cannot use the class-template `has_nested_columns` directly here.

Let me explain why, and my apologies if you already knew this but I misunderstood what you are suggesting.

If I did:
```
template <CUDF_ENABLE_IF(not has_nested_columns)`
```
then we'd not be able to trigger SFINAE. This is because `has_nested_columns` is a class-template. So, when you write `device_comparator<true> d_comp`, the substitution for the enable-if on the constructor will trigger a failure as `enable_if::type` will be ill-formed.

As a result of which, we want to defer the SFINAE at the member function level so that it behaves the way it is intended to: as an overload.

cpp/src/merge/merge.cu
divyegala: Your understanding is spot-on, actually, thanks for pointing this out! This should be `if constexpr`. Not only can we reduce compilation time this way, we also make the device function less bloated by not having to instantiate two versions.

mythrocks (approved):
LGTM. A couple of minor questions.

cpp/src/merge/merge.cu
divyegala: Good point! I added a check here https://github.com/rapidsai/cudf/pull/14250/files#diff-bea7124eddd4e287917be21bd0a0b7cb946816ad08c2a77943177a54569d87d0R641-R647

cpp/src/merge/merge.cu
```diff
@@ -493,6 +638,13 @@ table_ptr_type merge(std::vector<table_view> const& tables_to_merge,
 
   CUDF_EXPECTS(key_cols.size() == column_order.size(),
                "Mismatched size between key_cols and column_order");
+  CUDF_EXPECTS(std::accumulate(tables_to_merge.cbegin(),
+                               tables_to_merge.cend(),
+                               cudf::size_type{0},
+                               [](auto const& running_sum, auto const& tbl) {
+                                 return running_sum + tbl.num_rows();
+                               }) <= std::numeric_limits<cudf::size_type>::max(),
+               "Total number of merged rows exceeds row limit");
```
bdice, line 647: @divyegala (cc: @mythrocks) I don't think this actually protects against overflow. By the definition of overflow, the sum will always be less than the max limit of `size_type`. You need to use a _larger_ type and accumulate in `std::size_t` precision. Can you file a follow-up PR with this change?

```suggestion
  CUDF_EXPECTS(std::accumulate(tables_to_merge.cbegin(),
                               tables_to_merge.cend(),
                               size_t{0},
                               [](auto const& running_sum, auto const& tbl) {
                                 return running_sum + static_cast<size_t>(tbl.num_rows());
                               }) <= static_cast<size_t>(std::numeric_limits<cudf::size_type>::max()),
               "Total number of merged rows exceeds row limit");
```

Examples of this pattern:
- https://github.com/rapidsai/cudf/blob/2a923dfff84a8ce3422a126e9d0035282a0bea2f/cpp/src/strings/copying/concatenate.cu#L221
- https://github.com/rapidsai/cudf/blob/2a923dfff84a8ce3422a126e9d0035282a0bea2f/cpp/src/copying/scatter.cu#L324
- https://github.com/rapidsai/cudf/blob/2a923dfff84a8ce3422a126e9d0035282a0bea2f/cpp/src/scalar/scalar.cpp#L70

divyegala: Thanks for pointing that out @bdice! I'll raise a PR

PR #14251, opened 2023-10-05
cudf::detail::pinned_allocator doesn't throw from `deallocate`

### LINKED ISSUE #14165: [BUG] pinned_host_vector can cause abrupt program termination
**Describe the bug**
pinned_host_vector can throw from within it's destructor, causing the application process to terminate abruptly.  For example, this is causing abnormal termination of Spark executors when a GPU illegal access occurs during a Parquet read.  Since the executor process is abruptly terminated via low-level `abort()`, there's no chance to convey a useful message from the driver to the executor otherwise try to handle at the application level the `cudf::fatal_cuda_error` exception that is being thrown.

Compile and run the following program:
```c++
#include <stdexcept>
#include <cudf/detail/utilities/pinned_host_vector.hpp>

__global__ void crash(int *x) {
  int *y = NULL;
  *x = y[12345678];
}

int main() {
  int *x;
  try {
    auto p = cudf::detail::pinned_host_vector<uint8_t>(4096);
    cudaMalloc(&x, sizeof(*x));
    crash<<<1,1>>>(x);
    cudaDeviceSynchronize();
  } catch (std::exception const& e) {
  }
  return 0;
}
```
which will produce the following output:
```
terminate called after throwing an instance of 'cudf::fatal_cuda_error'
  what():  Fatal CUDA error encountered at: /home/jlowe/src/spark-rapids-jni/target/libcudf-install/include/cudf/detail/utilities/pinned_host_vector.hpp:172: 700 cudaErrorIllegalAddress an illegal memory access was encountered
Aborted (core dumped)
```

**Expected behavior**
Exceptions should not be thrown from destructors and cause an application to be terminated abruptly with no chance for the application to shutdown gracefully.

robertmaynard:
## Description
Fixes #14165

The deallocate function is called by the `pinned_host_vector`. Throwing from destructors is bad since they can't be caught, and generally get converted into runtime sig aborts.

## Checklist

### DIFF
--- a/cpp/include/cudf/detail/utilities/pinned_host_vector.hpp
+++ b/cpp/include/cudf/detail/utilities/pinned_host_vector.hpp
@@ -173,7 +173,7 @@ class pinned_allocator {
   {
     auto dealloc_worked = cudaFreeHost(p);
     (void)dealloc_worked;
-    assert(status__ == cudaSuccess);
+    assert(dealloc_worked == cudaSuccess);
   }
 
   /**
--- a/cpp/include/cudf/detail/utilities/pinned_host_vector.hpp
+++ b/cpp/include/cudf/detail/utilities/pinned_host_vector.hpp
@@ -169,7 +169,12 @@ class pinned_allocator {
    *        It is the responsibility of the caller to destroy
    *        the objects stored at \p p.
    */
-  __host__ inline void deallocate(pointer p, size_type /*cnt*/) { CUDF_CUDA_TRY(cudaFreeHost(p)); }
+  __host__ inline void deallocate(pointer p, size_type /*cnt*/)
+  {
+    auto dealloc_worked = cudaFreeHost(p);
+    (void)dealloc_worked;
+    assert(status__ == cudaSuccess);
+  }
 
   /**
    * @brief This method returns the maximum size of the \c cnt parameter


davidwendt (approved):

cpp/include/cudf/detail/utilities/pinned_host_vector.hpp
```diff
@@ -169,7 +169,12 @@ class pinned_allocator {
    *        It is the responsibility of the caller to destroy
    *        the objects stored at \p p.
    */
-  __host__ inline void deallocate(pointer p, size_type /*cnt*/) { CUDF_CUDA_TRY(cudaFreeHost(p)); }
+  __host__ inline void deallocate(pointer p, size_type /*cnt*/)
+  {
+    auto dealloc_worked = cudaFreeHost(p);
+    (void)dealloc_worked;
```
divyegala, line 175: I have never seen a type cast without assignment. What does this line do?

davidwendt: Generally, in Release mode, I believe this will suppress a variable-is-set-but-never-used warning from the compiler since the assert is compiled out.

robertmaynard: Correct on all accounts

divyegala: And where does the `status__` variable come from, seeing that it's never declared?

robertmaynard: Thanks @divyegala, that is a typo. I pushed a correction

divyegala (approved):

hyperbolic2346 (approved):

Issue #14227, opened 2023-09-28
[BUG] JSON reader with recover_with_null fails to detect incomplete JSON as invalid


andygrove:
**Describe the bug**
The following input has one invalid JSON record on line 3 (missing the closing `}`).

```
{ "number": 1 }
{ "number": 2 }
{ "number": 3
{ "number": 4 }
```

When reading this file with cuDF and specifying a schema with column `number` of type `int` and also specifying `recover_with_null`, the column returned is a struct instead of an int and has null values for the first two records and then has a struct with a child int containing the number 4.

```
COLUMN input - STRUCT
0 NULL
1 NULL
COLUMN input:CHILD_0 - INT64
0 NULL
1 NULL
2 4
```

It appears that line 3 is treated as valid and multiline, and line 4 is read as a child.

**Steps/Code to reproduce bug**
I am testing via Spark RAPIDS, but I suspect that this issue could be reproduced by adding this scenario to the `TEST_F(JsonReaderTest, JSONLinesRecovering)` test in `cpp/tests/io/json_test.cpp`.

**Expected behavior**
I would expect line 3 to be treated as invalid and return a NULL.

**Environment overview (please complete the following information)**
N/A

**Environment details**
N/A

**Additional context**

elstehle:
I've opened https://github.com/rapidsai/cudf/pull/14252 to address this issue. Feel free to test against the PR to verify if the new behaviour matches Spark for JSON lines with incomplete records.

PR #14252, opened 2023-10-05
Fixes behaviour for incomplete lines when `recover_with_nulls` is enabled

elstehle:
## Description

Closes https://github.com/rapidsai/cudf/issues/14227. Adapts the behaviour of the JSON finite-state transducer (FST) when `recover_with_nulls` is `true` to be more strict and reject lines that contain incomplete JSON objects (aka records) or JSON arrays (aka lists).

## Checklist

### DIFF
--- a/cpp/src/io/json/nested_json_gpu.cu
+++ b/cpp/src/io/json/nested_json_gpu.cu
@@ -697,7 +697,9 @@ auto get_transition_table(json_format_cfg_t format)
 
 /**
  * @brief Getting the translation table
- * @param recover_from_error Whether or not the tokenizer should recover from invalid lines
+ * @param recover_from_error Whether or not the tokenizer should recover from invalid lines. If
+ * `recover_from_error` is true, invalid JSON lines end with the token sequence (`ErrorBegin`,
+ * `LineEn`) and incomplete JSON lines (e.g., `{"a":123\n`) are treated as invalid lines.
  */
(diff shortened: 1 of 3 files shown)


cpp/src/io/json/nested_json_gpu.cu
vuule, line 1: could you leave a comment on the changes in logic in the FST (how does this translate into the intended behavior change)? I can glean some logic from the code, but it's possibly made up :D

elstehle: I've added comments to the code (`get_translation_table` and the `nl_tokens` lambda. 

Essentially, we just changed the behaviour when we're in `recover_from_error`, such that we fail if the JSON line is an incomplete JSON value. E.g.,
- when we see a newline while we are still parsing a `LIST` or a `STRUCT`, we emit an `ErrorBegin` token that will mark this incomplete JSON line as invalid.
- when we see a newline while we are still parsing a `string` or `field name` (e.g., `{"a":"123\n`) , we emit an `ErrorBegin` token that will mark this incomplete JSON line as invalid.

vuule (approved):
Thank you for the comments.
It's still challenging to understand case-to-case why `regular_tokens` and `recovering_tokens` might be different, but I think I got it.

ttnghia (approved):

PR #14254, opened 2023-10-05
Add in java bindings for DataSource

revans2:
## Description
This PR adds DataSource Java bindings. It also fixes a small bug in CUDF that made it so the bindings would not work for anything but CSV.

## Checklist


java/src/main/native/src/DataSourceHelperJni.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+namespace cudf {
+namespace jni {
+bool cache_data_source_jni(JNIEnv *env) {
+  jclass cls = env->FindClass(DATA_SOURCE_CLASS);
+  if (cls == nullptr) {
+    return false;
+  }
+
+  hostRead_method = env->GetStaticMethodID(cls, "hostRead", "(L" DATA_SOURCE_CLASS ";JJJ)J");
```
jlowe, line 42: Nit: Seems like it would be more straightforward on the Java end if these were member functions rather than static functions, and the JNI invoked it as a member function directly since the native code already has the object pointer.  Curious if that was considered.

java/src/main/java/ai/rapids/cudf/MultiBufferDataSource.java
```diff
@@ -0,0 +1,237 @@
...
+
+    interface DoCopy<T extends MemoryBuffer> {
+        void copyFromHostBuffer(T dest, long destOffset, HostMemoryBuffer src,
+                                long srcOffset, long srcAmount);
+    }
+
+    private <T extends MemoryBuffer> long read(long offset, long amount,
+                                               T dest, DoCopy<T> doCopy) {
+        long realOffset = Math.min(offset, sizeInBytes);
+        long realAmount = Math.min(sizeInBytes - realOffset, amount);
```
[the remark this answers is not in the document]
revans2: But then why does the CUDF API bother to return the amount read at all? That feels like too much of a departure from the CUDF APIs for me.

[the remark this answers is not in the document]
jlowe: OK that makes sense, but what about negative values which the cudf C++ interface does not allow since it's unsigned values?  Pretty sure we're going to do some bad and hard to debug things if someone accidentally passes a negative offset or length to this.

java/src/main/java/ai/rapids/cudf/MultiBufferDataSource.java
```diff
@@ -0,0 +1,225 @@
...
+      }
+    }
+
+    return realAmount;
+  }
+
+  @Override
+  public HostMemoryBuffer hostRead(long offset, long amount) {
+    long realOffset = Math.min(offset, sizeInBytes);
+    long realAmount = Math.min(sizeInBytes - realOffset, amount);
```
jlowe, line 142: Negative offset or amount will do some bad things here.  Since this is public, seems prudent to check here rather than assume the entry point will always be via `DataSource::hostReadBuff`.

java/src/main/java/ai/rapids/cudf/DataSource.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    @Override
+    public boolean isClean() {
+      return true;
+    }
+  }
+  private static final NoopCleaner cleaner = new NoopCleaner();
+
+  private void onHostBufferDone(long bufferId) {
```
jlowe, line 139: It would be good to have a comment on these private methods to note that they are called from JNI.  An IDE flags them as unused.

revans2:
@jlowe please take another look

java/src/main/java/ai/rapids/cudf/MultiBufferDataSource.java
jlowe: This comment has not been addressed.

revans2: It was addressed in a different place by https://github.com/rapidsai/cudf/pull/14254/commits/8e66e63a3eb1d2ff8fc9c941c7d23beea14d5f74

I thought it was more of a generic problem so I fixed it in a more generic place.

jlowe: I saw that commit, but it doesn't address the issue here.  This function is public and therefore could be called directly, rather than only from the JNI.  The fix in 8e66e63 only works if it's called via hostReadBuff, and even then that fix only addresses negative amounts.  I believe negative offsets are still problematic even in the JNI-called path.

revans2: If `offset` or `amount` is negative we will get a `realAmount` that is either too large or negative and/or a `realOffset` that is negative. In all of those cases we are going through a HostMemoryBuffer API that has asserts on it to protect against going off the limits if assertions are enabled and if they are not you are out of luck. This code is not going to violate that. But just to make debugging simpler if ti does happen I will add in asserts here too.

jlowe (approved):

revans2:
@davidwendt @vyasr could you please take a quick look at the C++ fix I had to put in to allow parquet/orc reads with a custom data source.

vuule:
Nice!
What is the motivating use case?

revans2:
> Nice! What is the motivating use case?

I did most of it to try and see if it could help with some memory management pain points. This is one of the few places where we hold onto host memory while allocating GPU memory. It didn't work out exactly how we wanted, but the plan is to try and avoid an extra copy when stitching parts of ORC and parquet together into a final buffer we send to CUDF.  It might get expanded to other things in the future.

revans2:
@vuule could you review the change I had to make to cpp/src/io/utilities/datasource.cpp? It is a C++ change to CUDF do I need a reviewer for it.

vuule (approved):

davidwendt (approved):
Approving libcudf C++ change.

PR #14262, opened 2023-10-09
Centralize chunked reading code in the parquet reader to reader_impl_chunking.cu

nvdbaranec:
As a precursor to further chunked reader work, this PR centralizes chunk-related code (mostly from the `reader::impl` class) into `reader_impl_chunking.cu` and `reader_impl_chunking.hpp`.  Also cleans up some variable naming and locations in `reader::impl` and the `file_intermediate_data` and `pass_intermediate_data classes`.


cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -250,6 +247,39 @@ class reader::impl {
    */
   void decode_page_data(size_t skip_rows, size_t num_rows);
 
+  /*
+   *
+   *
+    Functions related to computing chunks and passes (reader_impl_chunking.cu)
+   *
+   *
+   */
```
vuule, line 256: not a big fan of these headings in the code. IMO the names already suggest what the functions are related to.

vuule (approved):
Looks good, assuming this is mostly code shuffle. Hard to compare when the code is moved between files.

robertmaynard (approved):
Approving CMake changes

cpp/src/io/parquet/reader_impl_chunking.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cudf/types.hpp>
+
+#include "reader_impl_helpers.hpp"
```
ttnghia, line 21: Reverse their order.

PR #14263, opened 2023-10-09
Expose streams in public null mask APIs

vyasr:
## Description

Contributes to #925

## Checklist


cpp/src/bitmask/null_mask.cu
```diff
@@ -157,16 +157,21 @@ void set_null_mask(bitmask_type* bitmask,
...
 // or null, otherwise;
-void set_null_mask(bitmask_type* bitmask, size_type begin_bit, size_type end_bit, bool valid)
+void set_null_mask(bitmask_type* bitmask,
+                   size_type begin_bit,
+                   size_type end_bit,
+                   bool valid,
+                   rmm::cuda_stream_view stream)
 {
-  return detail::set_null_mask(bitmask, begin_bit, end_bit, valid, cudf::get_default_stream());
+  return detail::set_null_mask(bitmask, begin_bit, end_bit, valid, stream);
```
ttnghia, line 174: The detail and non-detail functions have exactly the same signature. Should we remove the detail version now?

davidwendt: Should the public APIs contain `CUDF_FUNC_RANGE();` calls?

bdice: Yes, that is the distinction between public and detail APIs. We always annotate public APIs with NVTX ranges.

vyasr: Do we want to add the nvtx ranges to the public APIs now? I'm happy to do it, but I think we should do it in a separate PR where we audit all our public APIs for nvtx ranges. I don't think we should be adding them as part of the streams PRs since a more systematic search will probably turn up more places that they're missing.

davidwendt: Since the functions are being touched in this PR and it is minimal change, I'd prefer not delaying adding the nvtx calls for these specific APIs.

ttnghia (approved):

cpp/src/bitmask/null_mask.cu
vyasr: Done.

bdice (approved):

davidwendt (approved):

PR #14264, opened 2023-10-10
Refactor LogicalType for Parquet

etseidl:
## Description
Continuation of #14097, this PR refactors the LogicalType struct to use the new way of treating unions defined in the parquet thrift (more enum like than struct like).

## Checklist


etseidl:
The ultimate goal is to allow greater use of `LogicalType` in the parquet reader and writer, so we can rely less on special cases that test the input or output length.

cpp/src/io/parquet/compact_protocol_writer.cpp
```diff
@@ -124,20 +119,15 @@ size_t CompactProtocolWriter::write(SchemaElement const& s)
...
   }
-  if (s.field_id) { c.field_int(9, s.field_id.value()); }
-  auto const isset = s.logical_type.isset;
-  // TODO: add handling for all logical types
-  // if (isset.STRING or isset.MAP or isset.LIST or isset.ENUM or isset.DECIMAL or isset.DATE or
-  //    isset.TIME or isset.TIMESTAMP or isset.INTEGER or isset.UNKNOWN or isset.JSON or isset.BSON)
-  //    {
-  if (isset.TIMESTAMP or isset.TIME) { c.field_struct(10, s.logical_type); }
+  if (s.field_id.has_value()) { c.field_int(9, s.field_id.value()); }
+  if (s.logical_type.has_value()) { c.field_struct(10, s.logical_type.value()); }
```
etseidl, line 130:
```suggestion
  if (s.field_id) { c.field_int(9, *s.field_id); }
  if (s.logical_type) { c.field_struct(10, *s.logical_type); }
```

@vuule We could alternatively use a less verbose idiom here...do you like the more explicit form?

vuule: I really prefer `has_value()` over the implicit conversion. I immediately know what `s.field_id` is and what the condition means. No strong opinion on `value()` vs `*`.

cpp/src/io/parquet/parquet.hpp
```diff
@@ -46,79 +46,92 @@ struct file_ender_s {
...
+  };
+  Type type;
+  thrust::optional<DecimalType> decimal_type;
+  thrust::optional<TimeType> time_type;
+  thrust::optional<TimestampType> timestamp_type;
+  thrust::optional<IntType> int_type;
+
+  constexpr bool is_time_millis() const
+  {
+    return type == TIME and time_type->unit.type == TimeUnit::MILLIS;
```
vuule, line 100: we don't need to check has_value() here?
~Trying to understand if maybe the following condition is sufficient: `time_type.value_or(UNDEFINED) == TimeUnit::MICROS`~
value_or does not work here at all :(

cpp/src/io/parquet/compact_protocol_writer.cpp
```diff
@@ -84,31 +82,28 @@ size_t CompactProtocolWriter::write(IntType const& integer)
...
+      break;
+    case LogicalType::TIMESTAMP:
+      c.field_struct(LogicalType::TIMESTAMP, logical_type.timestamp_type.value());
+      break;
+    case LogicalType::INTEGER:
+      c.field_struct(LogicalType::INTEGER, logical_type.int_type.value());
+      break;
+    case LogicalType::UNKNOWN: c.field_empty_struct(LogicalType::UNKNOWN); break;
+    case LogicalType::JSON: c.field_empty_struct(LogicalType::JSON); break;
+    case LogicalType::BSON: c.field_empty_struct(LogicalType::BSON); break;
```
vuule, line 105: Most of these could be `c.field_empty_struct(logical_type.type);` ?
I wonder if we should simply catch all of these under `default`.

cpp/src/io/parquet/parquet.hpp
```diff
@@ -46,79 +46,92 @@ struct file_ender_s {
...
+  {
+    return type == TIMESTAMP and timestamp_type->unit.type == TimeUnit::MICROS;
+  }
+
+  constexpr bool is_timestamp_nanos() const
+  {
+    return type == TIMESTAMP and timestamp_type->unit.type == TimeUnit::NANOS;
+  }
+
+  constexpr int8_t bit_width() const { return type == INTEGER ? int_type->bitWidth : -1; }
```
vuule, line 128: same here, is type == INTEGER interchangeable with int_type.has_value()?

cpp/src/io/parquet/writer_impl.cu
```diff
@@ -381,133 +399,151 @@ struct leaf_schema_fn {
...
-    col_schema.type                                = Type::INT32;
-    col_schema.converted_type                      = ConvertedType::TIME_MILLIS;
-    col_schema.stats_dtype                         = statistics_dtype::dtype_int32;
-    col_schema.logical_type.isset.TIME             = true;
-    col_schema.logical_type.TIME.unit.isset.MILLIS = true;
+    col_schema.type                    = Type::INT32;
+    col_schema.converted_type          = ConvertedType::TIME_MILLIS;
+    col_schema.stats_dtype             = statistics_dtype::dtype_int32;
+    col_schema.logical_type            = LogicalType{LogicalType::TIME};
+    col_schema.logical_type->time_type = TimeType{false, TimeUnit::MILLIS};
```
vuule, line 488: We could have constructors to avoid multi-step initialization
LogicalType(TimeType) etc.

vuule (commented):
couple of small questions/suggestions

cpp/src/io/parquet/parquet.hpp
etseidl: If `type` is a given value, then I'm assuming the optional struct for that value will also be set. Since this can be used on host or device, I didn't want to throw an assert in here.  Could do something like
```c++
time_type.value_or(TimeType{false, TimeUnit::UNDEFINED}).unit.type == TimeUnit::MILLIS
```
 to be safe, I suppose.

cpp/src/io/parquet/compact_protocol_writer.cpp
etseidl: I'd prefer to enumerate them and fall through, and leave defaut as a no-op in case there's a weird value set for type.  Probably more of a concern on the read side where new logical types we don't understand can get added.

cpp/src/io/parquet/parquet.hpp
etseidl: It _should_ be. I'm ok with bad things happening if someone forgets to assign the struct, but I'm assuming you're not :sweat_smile:

cpp/src/io/parquet/writer_impl.cu
etseidl: yeah, I guess that would get rid of some of the anxiety with type being set, but the struct being left `nullopt`.

cpp/src/io/parquet/parquet.hpp
vuule: nah, it's fine. I just wanted to make sure I understand the invariant.
But if we do rely on them being interchangeable, we should try to enforce the invariant in construction.

cpp/src/io/parquet/compact_protocol_writer.cpp
vuule: Sounds good for the most part - should the default throw instead?

cpp/src/io/parquet/parquet.hpp
vuule: I think the current code is as good as this gets.

cpp/src/io/parquet/compact_protocol_writer.cpp
etseidl: Hmm, on the write side...yeah, that's probably a good idea.  We likely want to know when the struct isn't properly initialized.

cpp/src/io/parquet/compact_protocol_writer.cpp
```diff
@@ -124,20 +119,15 @@ size_t CompactProtocolWriter::write(SchemaElement const& s)
...
     }
   }
-  if (s.field_id) { c.field_int(9, s.field_id.value()); }
-  auto const isset = s.logical_type.isset;
-  // TODO: add handling for all logical types
-  // if (isset.STRING or isset.MAP or isset.LIST or isset.ENUM or isset.DECIMAL or isset.DATE or
-  //    isset.TIME or isset.TIMESTAMP or isset.INTEGER or isset.UNKNOWN or isset.JSON or isset.BSON)
-  //    {
-  if (isset.TIMESTAMP or isset.TIME) { c.field_struct(10, s.logical_type); }
+  if (s.field_id.has_value()) { c.field_int(9, s.field_id.value()); }
```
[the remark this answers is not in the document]
etseidl: That's a lot of template foo to avoid a few invocations of `has_value`. We can leave this for another day :D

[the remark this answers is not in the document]
vuule: I think it would have a larger impact than that, but maybe I'm mixing it up with the reader side. Either way it's not a suggestion for this PR.

vuule (approved):

cpp/src/io/parquet/compact_protocol_writer.cpp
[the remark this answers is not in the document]
ttnghia: (Not sure if I'm on the same page)
How about `std::visit` to deal with such template issue and optional?

cpp/src/io/parquet/parquet.hpp
```diff
@@ -46,79 +46,98 @@ struct file_ender_s {
...
-  EnumType ENUM;
-  DecimalType DECIMAL;
-  DateType DATE;
-  TimeType TIME;
-  TimestampType TIMESTAMP;
-  IntType INTEGER;
-  NullType UNKNOWN;
-  JsonType JSON;
-  BsonType BSON;
+  enum Type {
```
ttnghia, line 76: `enum class`?

etseidl: The `Type` enums are already buried inside the struct, so we're already getting the benefits of a scoped enum. And keeping it non-scoped allows me to use the enum values as the positional argument to the `field_struct` calls in the writer without having to do a cast.

cpp/src/io/parquet/compact_protocol_writer.cpp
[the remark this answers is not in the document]
etseidl: That can be in the next refactor of CompactProtocolReader/Writer 🤣 (or should I say compact_protocol_reader/writer?)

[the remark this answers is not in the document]
vuule: I don't think std::visit is applicable here. Outside of reflection that would allow us to iterate over data members (which AFAIK does not exist), I think an overload set is as good as this gets. I'd be happy to learn about other solutions.

ttnghia (approved):

rjzamora:
@galipremsagar - As far as I can tell the `test_create_metadata_file_inconsistent_schema` failures must be coming from a change in this PR. Test passes for the commit just before this was merged.

vuule:
Hi @rjzamora, we identified other issue with changes to timestamp logical type and opened https://github.com/rapidsai/cudf/pull/14322 to fix them. Do you think this change would fix your failing test?

rjzamora:
Not sure, but I don't immediately think so.  The test failure we are investigating suggests that the behavior of `cudf.io.merge_parquet_filemetadata` has changed. We have a test where we aggregate footer metadata into a global `_metadata` file. One of those files contains all null values in one of the columns, while the other file contains int. We expect the aggregated metadata to "promote" the null type to align with the int type.

vuule:
@rjzamora can you please point us to the failing tests? Is there an issue open?

rjzamora:
Failing test is [here](https://github.com/rapidsai/cudf/blob/bc4d38dc370a3ebe743bcc5c17581aa5cd73de6b/python/dask_cudf/dask_cudf/io/tests/test_parquet.py#L447 - I don't believe it is showing up in CI anywhere because all dask-cudf tests are being skipped for some reason.

I'll trying to boil this down to a simpler reproducer (if possible).

vuule:
> are being skipped for some reason.

oof

galipremsagar:
> > are being skipped for some reason.

🤯 🤯 🤯

etseidl:
@rjzamora I think I'm running this to ground...it seems that the blob passed in to `merge_row_group_metadata` has the logical type annotation `UNKNOWN`, which is used for all null columns where the schema type can't be discovered. The change in this PR is causes that `UNKNOWN` to be written back out in the merged metadata, even though the physical type is written as `INT32`. I suspect that having the logical type present is throwing off whatever dask is doing to infer the type when it uses the _metadata file. I can think of two ways to fix this...either change the thrift protocol writer to not write the logical type if the type is 'UNKNOWN', or override an `UNKNOWN` logical type in `merge_row_group_metadata`. I'll have to discuss with @vuule offline to see which approach to take.

rjzamora:
Thanks for looking into this @etseidl !

>I suspect that having the logical type present is throwing off whatever dask is doing to infer the type when it uses the _metadata file.

Dask is using whatever type `pyarrow.dataset` happens to infer. I wouldn't be surprised if it was just using the logical type.

>I can think of two ways to fix this...either change the thrift protocol writer to not write the logical type if the type is 'UNKNOWN', or override an UNKNOWN logical type in merge_row_group_metadata. I'll have to discuss with @vuule offline to see which approach to take.

Okay, either approach sounds reasonable to me. To be completely honest, I'm somewhat doubtful that there is anyone actually depending on the behavior covered in this test. In fact, we are already expecting the "wrong" result for `dask.dataframe`.  Therefore, if this proves tricky to fix, I think it should be fine to modify or xfail the test for now.

PR #14265, opened 2023-10-10
Update rapids-cmake functions to non-deprecated signatures

robertmaynard:
## Description
Update to use non deprecated signatures for `rapids_export` functions

## Checklist

### DIFF
--- a/cpp/cmake/thirdparty/get_kvikio.cmake
+++ b/cpp/cmake/thirdparty/get_kvikio.cmake
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


bdice (approved):

cpp/cmake/thirdparty/get_kvikio.cmake
```diff
@@ -25,10 +25,10 @@ function(find_and_configure_kvikio VERSION)
     OPTIONS "KvikIO_BUILD_EXAMPLES OFF"
   )
 
-  if(KvikIO_BINARY_DIR)
-    include("${rapids-cmake-dir}/export/find_package_root.cmake")
-    rapids_export_find_package_root(BUILD KvikIO "${KvikIO_BINARY_DIR}" cudf-exports)
-  endif()
+  include("${rapids-cmake-dir}/export/find_package_root.cmake")
```
harrism, line 28: The CONDITION is new to me. But this change means that we include a file unconditionally where we used to conditionally conclude it. Does this need a condition?

harrism (commented):
One question. Otherwise looks good.

cpp/cmake/thirdparty/get_kvikio.cmake
robertmaynard: The include was only conditional so that we never brought in functions we didn't call. 

We won't have any problems always bringing in the file.

Issue #14226, opened 2023-09-28
[FEA] Add JSON reader option to ignore all characters after a valid JSON record


andygrove:
**Is your feature request related to a problem? Please describe.**

When reading a newline-delimited JSON file in Spark (using default settings), all characters after a valid JSON record are ignored. For example, all of the following inputs equate to the same value `{ "number": 1 }`:

```
{ "number": 1 }
{ "number": 1 } xyz
{ "number": 1 } {}
{ "number": 1 } { "number": 4 }
```

However, when reading this data in cuDF (and when specifying the `recover_with_nulls` option), the lines with additional characters are rejected as invalid, and NULL values are returned.

**Describe the solution you'd like**

To support Spark compatibility, I would like a new JSON reader option that specifies to ignore all input after a valid JSON record and before the line delimiter.

**Describe alternatives you've considered**
It is perhaps possible that in Spark RAPIDS we could pre-process the input for some of these cases, but I am not sure that it is feasible to efficiently use regular expressions to completely validate JSON.

**Additional context**

elstehle:
Thanks for the example, @andygrove. My understanding of Spark JSON parser's behaviour is still a bit fuzzy here. Would it be correct to assume that it fails for any line whose _first_ record (including first record's decendants) does not conform to the JSON grammar and that Spark parser simply ignores anything that comes after the first record?

andygrove:
> Thanks for the example, @andygrove. My understanding of Spark JSON parser's behaviour is still a bit fuzzy here. Would it be correct to assume that it fails for any line whose _first_ record (including first record's decendants) does not conform to the JSON grammar and that Spark parser simply ignores anything that comes after the first record?

Yes, that is my understanding.

elstehle:
Thank you for confirming, @andygrove. I'm working on a PR that will change the behaviour to ignore any data that comes after the first JSON value on a JSON line and the next newline.

elstehle:
Just to follow up, I've opened https://github.com/rapidsai/cudf/pull/14279 to address this issue.

Now, when `recover_with_nulls` is `true`, it will _always_ ignore any excess data that comes after the first record on each JSON line. Is that sufficient, or do we need the behaviour to be toggleable via an extra option?

andygrove:
> Just to follow up, I've opened #14279 to address this issue.
> 
[...]

PR #14279, opened 2023-10-13
Changes JSON reader's recovery option's behaviour to ignore all characters after a valid JSON record

elstehle:
## Description
Closes https://github.com/rapidsai/cudf/issues/14226.

The new behvior of `JSON_LINES_RECOVER` will now ignore excess characters after the first valid JSON record on each JSON line.
```
{ "number": 1 } 
{ "number": 1 } xyz
{ "number": 1 } {}
{ "number": 1 } { "number": 4 }
```

**Implementation details:**
The JSON parser pushdown automaton was changed for `JSON_LINES_RECOVER` format such that when in state `PD_PVL` (`post-value`, "I have just finished parsing a value") and when the stack context is `ROOT` ("I'm not somewhere within a list or struct"), we just treat all characters as "white space" until encountering a newline character. `post-value` in stack context `ROOT` is exactly the condition we are in after having parsed the first valid record of a JSON line. _Thanks to @karthikeyann for suggesting to use `PD_PVL` as the capturing state._ 

As the stack context is generated upfront, we have to fix up and correct the stack context to set the stack context as `ROOT` stack context for all these excess characters. I.e., (`_` means `ROOT` stack context, `{` means within a `STRUCT` stack context):
```
in:    {"a":1}{"this is supposed to be ignored"}
stack: _{{{{{{_{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{
```
Needs to be fixed up to become:
```
in:    {"a":1}{"this is supposed to be ignored"}
stack: _{{{{{{__________________________________
```

## Checklist


cpp/src/io/fst/lookup_tables.cuh
```diff
@@ -753,15 +753,15 @@ class TranslationOp {
                                              RelativeOffsetT const relative_offset,
                                              SymbolT const read_symbol) const
   {
-    return translation_op(*this, state_id, match_id, relative_offset, read_symbol);
+    return translation_op(state_id, match_id, relative_offset, read_symbol);
```
elstehle, line 756: This is just a minor improvement that I piggy-backed into this PR. The delegate that implements `translation_op` has no benefit from the extra reference of `*this`, so we removed it from the arguments.

bdice (approved):
Just a couple small pieces of feedback. I think this makes sense. The overall complexity of the FST code is getting higher and higher, but I'm not sure if we can do much about that while continuing to add requested features.

cpp/src/io/json/nested_json_gpu.cu
```diff
@@ -91,6 +91,98 @@ void check_input_size(std::size_t input_size)
...
+  }
+};
+
+// Aliases for readability of the transition table
+constexpr auto TT_BEFR = dfa_states::BEFORE;
+constexpr auto TT_INSD = dfa_states::WITHIN;
+constexpr auto TT_POST = dfa_states::EXCESS;
+
+// Transition table
+std::array<std::array<dfa_states, NUM_SYMBOL_GROUPS>, TT_NUM_STATES> const transition_table{
```
ttnghia, line 176: Can this be `constexpr`?

cpp/src/io/json/nested_json_gpu.cu
```diff
@@ -733,6 +828,18 @@ auto get_translation_table(bool recover_from_error)
     return regular_tokens;
   };
 
+  /**
+   * @brief Helper function that returns `recovering_tokens` if `recover_from_error` is true and
+   * returns `regular_tokens` otherwise. This is used to ignore excess characters after the first
+   * vaue in the case of JSON lines that recover from invalid lines, as Spark does simply ignore any
+   * excess characters that follow the first record on a JSON line.
```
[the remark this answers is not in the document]
elstehle: Thanks! I've adopted your suggestion.

cpp/src/io/json/nested_json_gpu.cu
elstehle: Thanks, I've made it `constexpr` 👍

cpp/tests/io/json_test.cpp
```diff
@@ -2020,4 +2020,65 @@ TEST_F(JsonReaderTest, JSONLinesRecovering)
                     c_validity.cbegin()});
 }
 
+TEST_F(JsonReaderTest, JSONLinesRecoveringIgnoreExcessChars)
+{
+  /**
+   * @brief Spark has the specific need to ignore extra characters that come after the first record
+   * on a JSON line
+   */
+  std::string data =
```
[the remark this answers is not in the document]
elstehle: Good point, thanks. I've added the following three test cases:
```
    R"({"a":5}//Comment after record)"
    "\n"
    // 6 -> (valid)
    R"({"a":6} //Comment after whitespace)"
    "\n"
    // 7 -> (invalid)
    R"({"a":5 //Invalid Comment within record})";
 ```

**Comment by elstehle** [comment]
> . The overall complexity of the FST code is getting higher and higher, but I'm not sure if we can do much about that while continuing to add requested features

Yes, agreed. Unfortunately, the Pushdown Transducer for a JSON parser is quite complex by nature already. For most of the remaining Spark requests on our radar we will be introducing extra pre-/post-processing steps rather than changing the Pushdown machinery.

**Review by ttnghia — APPROVED** [review]

File: cpp/src/io/json/nested_json_gpu.cu [hunk 5]
```diff
@@ -91,6 +91,98 @@ void check_input_size(std::size_t input_size)
...
+  {
+    constexpr int32_t single_output_item = 1;
+    return single_output_item;
+  }
+};
+
+// Aliases for readability of the transition table
+constexpr auto TT_BEFR = dfa_states::BEFORE;
+constexpr auto TT_INSD = dfa_states::WITHIN;
+constexpr auto TT_POST = dfa_states::EXCESS;
```
karthikeyann, line 173: nit: how about `TT_BEFORE`, etc.? (all have same number of characters).
It gets a little difficult to read `TT_INSD` to `TT_INSIDE`.
It's a small table anyway.

karthikeyann (commented):
Great work! Especially with the additional processing step without complicating the existing JSON FST table.

cpp/src/io/json/nested_json_gpu.cu
elstehle: Thanks, good point. Renamed the entries 👍

karthikeyann (approved):

GregoryKimball:
@andygrove would you please evaluate this solution?

andygrove:
> @andygrove would you please evaluate this solution?

@GregoryKimball I have confirmed that this resolves the issue. The plugin tests for this issue now pass when run against this PR.

PR #13912, opened 2023-08-17
Improve performance of nvtext::edit_distance

davidwendt:
## Description

Improves performance of `nvtext::edit_distance` by reworking the algorithm with shorter working buffer and simpler logic.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #14283, opened 2023-10-13
Fix memset error in nvtext::edit_distance_matrix

davidwendt:
## Description
Fixes a bug in `nvtext::edit_distance_matrix` where the internal offsets vector is initialized to 0. 
This error was introduced in #13912 
The bug was found while working on a different PR which re-ordered the nvtext gtests execution causing device memory to be reused from the rmm pool in a different way.

## Checklist

### DIFF
--- a/cpp/src/text/edit_distance.cu
+++ b/cpp/src/text/edit_distance.cu
@@ -224,7 +224,7 @@ std::unique_ptr<cudf::column> edit_distance_matrix(cudf::strings_column_view con
   cudf::size_type n_upper = (strings_count * (strings_count - 1)) / 2;
   rmm::device_uvector<std::ptrdiff_t> offsets(n_upper, stream);
   auto d_offsets = offsets.data();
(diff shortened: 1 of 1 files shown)


bdice (approved):
LGTM. Would we want to consider a `using ...` type declaration, or `decltype`, or `::value_type`, or something to reduce the hardcoded types?

davidwendt:
> LGTM. Would we want to consider a `using ...` type declaration, or `decltype`, or `rmm::device_uvector<T>::value_type`, or something to reduce the hardcoded types?

I'm not sure it is worth doing. Every parameter to `cudaMemsetAsync` is really a potential error. 
I came up with this
```
using offsets_vector_type = rmm::device_uvector<std::ptrdiff_t>;
offsets_vector_type offsets( count, stream );
cudaMemsetAsync( offsets.begin(), 0, count * sizeof(typename offsets_vector_type::value_type), stream.value() );
```
and realized that I could just have easily put in the wrong `count` variable too.

harrism (approved):

ttnghia (approved):
This looks like something that the compiler can detect and issue a warning :)

PR #14284, opened 2023-10-13
Move and rename byte-pair-encoding source files

davidwendt:
## Description
Moves and renames the byte-pair-encoding source files. The source files are moved from the `text/subword` to `text/bpe` and the filenames have remove the `tokenize` since these functions only do encoding.
No functions names have been changed. The `nvtext::load_merge_pairs_file` API has been deprecated. Callers must load the pairs into a strings column (using the CSV or text readers in cuio) and call the new `nvtext::load_merge_pairs` API instead.

Follow on PRs will address function and performance issues.

## Checklist


cpp/include/nvtext/byte_pair_encoding.hpp
```diff
@@ -32,11 +32,11 @@ namespace nvtext {
 /**
  * @brief The table of merge pairs for the BPE encoder.
  *
- * To create an instance, call @ref nvtext::load_merge_pairs_file
+ * To create an instance, call @ref nvtext::load_merge_pairs
  */
 struct bpe_merge_pairs {
   struct bpe_merge_pairs_impl;
-  std::unique_ptr<bpe_merge_pairs_impl> impl{};  ///< Implementation of the BPE merge pairs table.
+  bpe_merge_pairs_impl* impl{};  ///< Implementation of the BPE merge pairs table.
```
ttnghia, line 39: So this will no longer manage the pointer which may cause memory leak. Is it intentional?

davidwendt: Ah, good catch. I meant for the destructor to free the pointer. I need to change this back.

bdice (approved):
Approving aside from the pointer issue mentioned above.

cpp/src/text/bpe/load_merge_pairs.cu
```diff
@@ -158,14 +158,14 @@ bpe_merge_pairs::bpe_merge_pairs_impl::bpe_merge_pairs_impl(
...
-  : impl(detail::create_bpe_merge_pairs_impl(std::move(input), stream))
+  : impl(detail::create_bpe_merge_pairs_impl(std::move(input), stream).release())
 {
 }
 
 bpe_merge_pairs::bpe_merge_pairs(cudf::strings_column_view const& input,
                                  rmm::cuda_stream_view stream,
                                  rmm::mr::device_memory_resource* mr)
-  : impl(detail::create_bpe_merge_pairs_impl(input, stream, mr))
+  : impl(detail::create_bpe_merge_pairs_impl(input, stream, mr).release())
```
karthikeyann, line 168: This is the memory leak pointed by @ttnghia

ttnghia (approved):

karthikeyann (approved):

PR #14287, opened 2023-10-16
Register ``partd`` encode dispatch in ``dask_cudf``

rjzamora:
## Description
This PR enables "disk"-based shuffling of `cudf`-backed Dask-DataFrame collections, but does **not** yet add the `shuffle="disk"` option to the `dask_cudf.DataFrame.shuffle/sort_values` APIs.

We now use basic (slow) `pickle` logic to convert `cudf.DataFrame` objects to/from `bytes` here, so I'd like to consider further optimizations before making the `shuffle="disk"` option "official".

## Checklist


wence- (approved):
Looks good to me.

rjzamora:
cc @charlesbluca - Since this partially addresses https://github.com/dask/partd/issues/61

wence- (approved):
Makes sense, thanks!

wence-:
Hmm, looks like the disk shuffle in dask wants some internal property of an `Index` that our indices do not provide:

```
FAILED python/dask_cudf/dask_cudf/tests/test_sort.py::test_disk_shuffle - AttributeError: 'Int64Index' object has no attribute '_attributes'
```

rjzamora:
>Hmm, looks like the disk shuffle in dask wants some internal property of an Index that our indices do not provide:

Ah, right - Sorry. The test requires a newer version of dask, so I'll need to add a version check to that test.

python/dask_cudf/dask_cudf/backends.py
```diff
@@ -481,6 +481,31 @@ def sizeof_cudf_series_index(obj):
...
+try:
+    from dask.dataframe.dispatch import partd_encode_dispatch
+
+    @partd_encode_dispatch.register(cudf.DataFrame)
+    def _simple_cudf_encode(_):
+        # Basic pickle-based encoding for a partd k-v store
+        import pickle
+        from functools import partial
+
+        import partd
```
galipremsagar, line 494: Can we add `partd` to our package requirements and conda recipes?

wence-: I think it comes in transitively through dask? Which is a dependency of dask-cudf.

rjzamora: Right exactly, [dask depends on partd](https://github.com/dask/dask/blob/91098a634967dfb44d67da54b1394d730f75e87d/pyproject.toml#L37). I think it should be safe for us to let dask worry about the partd dependency. If dask suddenly stops using partd for `shuffle="disk"`, it will also stop using `partd_encode_dispatch`.

galipremsagar (approved):

PR #14292, opened 2023-10-17
Refactor cudf_kafka to use skbuild

jdye64:
## Description
Refactor the currently outdated cudf_kafka build setup to use skbuild instead.


bdice:
Cross-linking for awareness: https://github.com/rapidsai/cudf/pull/14296/files#r1364688842

bdice:
Let's also fix the missing `README.md` for cudf_kafka in this PR.
```
SetuptoolsWarning: File '/opt/conda/conda-bld/work/python/cudf_kafka/README.md' cannot be found
```

bdice:
CI is failling. At this point, I think it's due to the problem we noticed previously where `libcudf_kafka` is shipping its headers in `include/include/cudf_kafka` instead of `include/cudf_kafka`. I will have to revisit that and figure out what's going wrong.

cpp/libcudf_kafka/CMakeLists.txt
```diff
@@ -94,7 +96,7 @@ install(
   EXPORT cudf_kafka-exports
 )
 
-install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include DESTINATION include)
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```
bdice, line 99:
```suggestion
install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```

build.sh
```diff
@@ -234,6 +234,11 @@ if [[ "${EXTRA_CMAKE_ARGS}" != *"DFIND_CUDF_CPP"* ]]; then
     EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DFIND_CUDF_CPP=ON"
 fi
 
+# Append `-DFIND_CUDF_KAFKA_CPP=ON` to EXTRA_CMAKE_ARGS unless a user specified the option.
+if [[ "${EXTRA_CMAKE_ARGS}" != *"DFIND_CUDF_KAFKA_CPP"* ]]; then
+    EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DFIND_CUDF_KAFKA_CPP=ON"
+fi
```
bdice, line 240: We only need to add this flag if `cudf_kafka` is being built.
```suggestion
# Append `-DFIND_CUDF_KAFKA_CPP=ON` to EXTRA_CMAKE_ARGS unless a user specified the option.
if buildAll || hasArg cudf_kafka && [[ "${EXTRA_CMAKE_ARGS}" != *"DFIND_CUDF_KAFKA_CPP"* ]]; then
    EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DFIND_CUDF_KAFKA_CPP=ON"
fi
```

build.sh
```diff
@@ -369,9 +374,9 @@ fi
 # build cudf_kafka Python package
 if hasArg cudf_kafka; then
     cd ${REPODIR}/python/cudf_kafka
-    SKBUILD_CONFIGURE_OPTIONS="-DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR}" \
+    SKBUILD_CONFIGURE_OPTIONS="-DCMAKE_PREFIX_PATH=${INSTALL_PREFIX} -DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR} -DCMAKE_CUDA_ARCHITECTURES=${CUDF_CMAKE_CUDA_ARCHITECTURES} ${EXTRA_CMAKE_ARGS}" \
         SKBUILD_BUILD_OPTIONS="-j${PARALLEL_LEVEL:-1}" \
-        python -m pip install --no-build-isolation --no-deps .
+        python -m pip install --no-build-isolation --no-deps . -vvv
```
bdice, line 379:
```suggestion
        python -m pip install --no-build-isolation --no-deps .
```

bdice:
This PR requested a lot of reviewers because it touches a lot of files in fairly small ways. I think we would be fine with a single review from @vyasr and an ops review, since it's mostly build system changes that Vyas is familiar with (or explicitly requested during conversations about this PR).

bdice:
@divyegala or @robertmaynard, your reviews would also be welcome here.

cpp/libcudf_kafka/CMakeLists.txt
```diff
@@ -94,7 +96,7 @@ install(
   EXPORT cudf_kafka-exports
 )
 
-install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include DESTINATION include)
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```
bdice, line 99: This is a pretty subtle change but fixes a major issue with packaging. Without this slash, `libcudf_kafka` has been shipping its headers in `include/include/cudf_kafka/...` which led to bad practices in the Python packaging for finding those headers.

python/cudf_kafka/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
...
+       "Search for existing cudf_kafka C++ installations before defaulting to local files" OFF
+)
+
+option(CUDF_KAFKA_BUILD_WHEELS "Whether this build is generating a Python wheel." OFF)
+
+# If the user requested it we attempt to find cudf_kafka.
+if(FIND_CUDF_KAFKA_CPP)
+  find_package(cudf_kafka ${cudf_kafka_version} REQUIRED)
+else()
+  set(cudf_kafka_FOUND OFF)
```
robertmaynard, line 43: Nitpick: This else isn't need. variables that don't exist evaluate to false

bdice, line 43:
```suggestion
```

python/cudf_kafka/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+# the License.
+# =============================================================================
+
+cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
+
+set(cudf_kafka_version 23.12.00)
+
+include(../../fetch_rapids.cmake)
+include(rapids-cuda)
+rapids_cuda_init_architectures(cudf-kafka-python)
```
vyasr, line 21: suggestion: As I said in build.sh, I don't think these are necessary.
```suggestion
```

python/cudf_kafka/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+
+project(
+  cudf-kafka-python
+  VERSION ${cudf_kafka_version}
+  LANGUAGES # TODO: Building Python extension modules via the python_extension_module requires the C
+            # language to be enabled here. The test project that is built in scikit-build to verify
+            # various linking options for the python library is hardcoded to build with C, so until
+            # that is fixed we need to keep C.
+            C CXX CUDA
```
vyasr, line 30:
```suggestion
            C CXX
```

python/cudf_kafka/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+# If the user requested it we attempt to find cudf_kafka.
+if(FIND_CUDF_KAFKA_CPP)
+  find_package(cudf_kafka ${cudf_kafka_version} REQUIRED)
+endif()
+
+if(NOT cudf_kafka_FOUND)
+  set(BUILD_TESTS OFF)
+  set(BUILD_BENCHMARKS OFF)
+  set(_exclude_from_all "")
+  if(CUDF_KAFKA_BUILD_WHEELS)
```
vyasr, line 48: I suspect a couple of the things in this block are superfluous, but since we're not actually building cudf_kafka wheels it's probably easiest to leave this block as is for now and make fixes later if/when we actually run and test it.

python/cudf_kafka/LICENSE
```diff
@@ -0,0 +1 @@
+../../LICENSE
```
vyasr, line 1: Question: Just double-checking since it's a limitation of the GH GUI that's bitten us before, this file is a symlink and not an actual file containing `../../LICENSE` as text, right?

python/cudf_kafka/README.md
```diff
@@ -0,0 +1 @@
+../../README.md
```
vyasr, line 1: Same question as for the license.

vyasr (changes requested):
praise: Few things left to clean up but overall this looks great!

build.sh
```diff
@@ -257,7 +262,7 @@ fi
 ################################################################################
 # Configure, build, and install libcudf
 
-if buildAll || hasArg libcudf || hasArg cudf || hasArg cudfjar; then
+if buildAll || hasArg libcudf || hasArg cudf || hasArg cudf_kafka || hasArg cudfjar; then
```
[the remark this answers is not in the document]
bdice: No, this is actually a new requirement. We weren't using CMake for the `cudf_kafka` build before, so this change adds support for building `cudf_kafka` with the appropriate flags. Note that `libcudf_kafka` does not use `LANGUAGES CUDA` so it doesn't need this, afaik, but `cudf_kafka` _does_ need `LANGUAGES CUDA` like all the wheel building scripts. If I'm misunderstanding this part of the CMake code, let me know!

build.sh
```diff
@@ -369,7 +374,7 @@ fi
 # build cudf_kafka Python package
 if hasArg cudf_kafka; then
     cd ${REPODIR}/python/cudf_kafka
-    SKBUILD_CONFIGURE_OPTIONS="-DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR}" \
+    SKBUILD_CONFIGURE_OPTIONS="-DCMAKE_PREFIX_PATH=${INSTALL_PREFIX} -DCMAKE_LIBRARY_PATH=${LIBCUDF_BUILD_DIR} -DCMAKE_CUDA_ARCHITECTURES=${CUDF_CMAKE_CUDA_ARCHITECTURES} ${EXTRA_CMAKE_ARGS}" \
```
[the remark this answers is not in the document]
bdice: Ah. First, see the comment above. https://github.com/rapidsai/cudf/pull/14292#discussion_r1390084465

The issue is that if libcudf is _not_ found, the `cudf_kafka` wheel has to compile libcudf, too. We probably _never_ want to ship cudf_kafka as a wheel for this reason -- only as a conda package where we can know that libcudf will be available. I'm torn about how to handle this case since we need scikit-build/CMake packaging (which implies wheel compatibility if we're keeping it aligned with other packages) but we would never want to ship all the bits for libcudf.so in both the cudf wheel and in a hypothetical cudf_kafka wheel.

dependencies.yaml
```diff
@@ -276,6 +278,11 @@ dependencies:
       - output_types: pyproject
         packages:
           - protoc-wheel
+  build_python_cudf_kafka:
+    common:
+      - output_types: [conda, requirements, pyproject]
+        packages:
+          - scikit-build>=0.13.1
```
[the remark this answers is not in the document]
bdice: I didn't do that because custreamz and dask-cudf also use that list but don't need scikit-build.

python/cudf_kafka/LICENSE
bdice: Yes, it's a symlink. Same for the README.

build.sh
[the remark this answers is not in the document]
bdice: @vyasr It would probably be easiest for us to review this together on a call, so that we can get it shipped as early as possible in burndown and talk over the confusing pieces.

dependencies.yaml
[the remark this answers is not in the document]
bdice: You're correct. I guess I read something wrong when I was doing this earlier.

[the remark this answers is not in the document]
bdice: Fixed in 64a5d618baf808286bb481dc9c2efe95728f9680.

build.sh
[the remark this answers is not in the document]
vyasr: For the record, we discussed offline and concluded that the easiest path forward here is to have cudf_kafka be much simpler than our other RAPIDS packages by not supporting building the C++ components from the Python build. This won't work for our typical wheel workflows, but we have no plans for building wheels of this package right now so that's a non-issue, and it will dramatically simplify these build scripts right now.

bdice:
@vyasr and I discussed this offline, and concluded that the `cudf_kafka` Python package should not be possible to build if the corresponding `libcudf_kafka` C++ package cannot be found. This greatly simplifies the code. This change makes it impossible to build pure wheels of `cudf_kafka` but we aren't shipping those today anyway -- and such a wheel would have to build and ship all of `libcudf` (and `libcudf_kafka`) anyway, which is undesirable. We only ship conda packages, and this won't affect conda.

vyasr:
I pushed one small fix that should hopefully get this passing CI.

vyasr (approved):
Much happier with this simplified version. The remaining test failures are unrelated to this PR and will be resolved once we can merge #14399. Nice work you two!

bdice (approved):
I was heavily involved in the development of this PR so my review is a bit unfair, but I did a round of self-review and all looks fine to me. I think we will be fine to merge this with a review from @rapidsai/ops.

bdice:
I did want to mark this PR as `breaking` because of the packaging changes for `libcudf_kafka` -- we moved some headers around (so that the include paths look like `include/cudf_kafka/...` instead of `include/include/cudf_kafka/...`. However, I don't expect any significant breakage for downstream use cases since the `cudf_kafka` Python package is the only tool I know of that builds on the C++ `libcudf_kafka` library.

ajschmidt8 (approved):

PR #14295, opened 2023-10-17
Sort dictionary data alphabetically in the ORC writer

vuule:
## Description
Strings in the dictionary data streams are now sorted alphabetically.
Reduces file size in some cases because compression can be more efficient.

Reduces throughput up to 22% when writing strings columns (3% speedup when dictionary encoding is not used, though!).
Benchmark data does not demonstrate the compression difference, but we have some user data that compresses almost 30% better.

## Checklist


cpp/src/io/orc/writer_impl.cu
```diff
@@ -867,16 +870,15 @@ encoded_data encode_columns(orc_table_view const& orc_table,
...
-        if (ck.type_kind == TypeKind::DECIMAL) { ck.decimal_offsets = column.decimal_offsets(); }
+        auto const is_str_dict =
+          ck.type_kind == TypeKind::STRING and ck.encoding_kind == DICTIONARY_V2;
+        ck.dict_index = is_str_dict ? column.host_stripe_dict(stripe.id).index.data() : nullptr;
+        ck.dict_data_order =
+          is_str_dict ? column.host_stripe_dict(stripe.id).data_order.data() : nullptr;
+        ck.dtype_len = (ck.type_kind == TypeKind::STRING) ? 1 : column.type_width();
+        ck.scale     = column.scale();
+        ck.decimal_offsets =
+          (ck.type_kind == TypeKind::DECIMAL) ? column.decimal_offsets() : nullptr;
```
vuule, line 881: Some of these were left uninitialized when unused, changed to always initialize.

vuule:
> Do you expect a 3% slow down to the write because of the sort for dictionary encoded data?

The slowdown is up to 22% unfortunately. Sorting is not cheap :(
FWIW, performance-wise we are still in a much better spot than the starting point - pre-cuco use.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -2133,15 +2152,55 @@ stripe_dictionaries build_dictionaries(orc_table_view& orc_table,
...
+                        sd.data.end(),
+                        sd.data_order.begin(),
+                        string_rows_less{orc_table.d_columns, sd.column_idx});
+
+    // Create the inverse permutation - i.e. the mapping from the original order to the sorted
+    auto order_copy = cudf::detail::make_device_uvector_async<uint32_t>(
+      sd.data_order, current_stream, rmm::mr::get_current_device_resource());
+    thrust::scatter(rmm::exec_policy_nosync(current_stream),
+                    thrust::make_counting_iterator<uint32_t>(0),
+                    thrust::make_counting_iterator<uint32_t>(sd.data_order.size()),
```
davidwendt, line 2191:
```suggestion
                    thrust::counting_iterator<uint32_t>(0),
                    thrust::counting_iterator<uint32_t>(sd.data_order.size()),
```
I've been told that if you specify the type, there is no reason to use the `make_counting_iterator` to make a `counting_iterator`.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -2023,21 +2025,37 @@ struct stripe_dictionaries {
...
+/**
+ * @brief Compares two rows in a strings column
+ */
+struct string_rows_less {
+  device_span<orc_column_device_view> cols;
+  uint32_t col_idx;
+  __device__ bool operator()(size_type lhs_idx, size_type rhs_idx) const
+  {
+    auto const& col = cols[col_idx];
+    return col.element<string_view>(lhs_idx).compare(col.element<string_view>(rhs_idx)) < 0;
```
davidwendt, line 2055:
```suggestion
    return col.element<string_view>(lhs_idx) < col.element<string_view>(rhs_idx);
```
This looks a bit clearer I think.

vuule: no idea why I thought `<` would not work.

davidwendt (approved):

cpp/src/io/orc/writer_impl.cu
```diff
@@ -2133,15 +2152,55 @@ stripe_dictionaries build_dictionaries(orc_table_view& orc_table,
...
+    if (not sd.is_enabled) { continue; }
+
+    sd.map_slots = {};
+    dict_order_owner.emplace_back(sd.entry_count, stream);
+    sd.data_order = dict_order_owner.back();
+  }
+  stripe_dicts.host_to_device_async(stream);
+
+  // Sort stripe dictionaries alphabetically
+  auto streams = cudf::detail::fork_streams(stream, std::min<size_t>(dict_order_owner.size(), 8));
```
divyegala, line 2170: Is 8 streams an empirical choice?

divyegala (approved):
Looks good, just one question

cpp/src/io/orc/writer_impl.cu
vuule: it is
I tried powers of two up to 32 and 8 was the fastest one. There wasn't a big difference compared to other 4+ values, though.

abellina:
Before we merge, mind if I run orc benchmarks on our stuff? I should be able to get these back to you tomorrow. @vuule

abellina:
1. I ran NDS at 3TB and, as expected, it didn't affect that benchmark at all. I compared the results for the tables and I don't see anything out of the ordinary.
2. I ran a simple transcode scenario (read all the source data, and write it out whole) with customer source data around 10GB and these were my findings: 
   - With this patch, we wrote 8.8GB instead of 11GB without (20% reduction in size)
   - With this patch it took us 115 seconds wall clock vs 106 seconds without to carry out my experiment (7.8% slower)
   - Isolating the write GPU time, the GPU spent ~12% more time (or 100ms per task more) with this patch, in order to do the sorting.

With the above we believe default=on makes sense but we really like having the flag you added @vuule, because it allows us to experiment easily and you never know what pathological cases we may run into.

Thank you!!

abellina (approved):

PR #14173, opened 2023-09-22
Pin to `aws-sdk-cpp<1.11`

pentschev:
## Description
Pin conda packages to `aws-sdk-cpp<1.11`. The recent upgrade in version `1.11.*` has caused several issues with cleaning up (more details on changes can be read in [this link](https://github.com/aws/aws-sdk-cpp#version-111-is-now-available)), leading to Distributed and Dask-CUDA processes to segfault. The stack for one of those crashes looks like the following:

```
(gdb) bt
#0  0x00007f5125359a0c in Aws::Utils::Logging::s_aws_logger_redirect_get_log_level(aws_logger*, unsigned int) () from /opt/conda/envs/dask/lib/python3.9/site-packages/pyarrow/../../.././libaws-cpp-sdk-core.so
#1  0x00007f5124968f83 in aws_event_loop_thread () from /opt/conda/envs/dask/lib/python3.9/site-packages/pyarrow/../../../././libaws-c-io.so.1.0.0
#2  0x00007f5124ad9359 in thread_fn () from /opt/conda/envs/dask/lib/python3.9/site-packages/pyarrow/../../../././libaws-c-common.so.1
#3  0x00007f519958f6db in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#4  0x00007f5198b1361f in clone () from /lib/x86_64-linux-gnu/libc.so.6
```

Such segfaults now manifest frequently in CI, and in some cases are reproducible with a hit rate of ~30%. Given the approaching release time, it's probably the safest option to just pin to an older version of the package while we don't pinpoint the exact cause for the issue and a patched build is released upstream.

The `aws-sdk-cpp` is statically-linked in the `pyarrow` pip package, which prevents us from using the same pinning technique. cuDF is currently pinned to `pyarrow=12.0.1` which seems to be built against `aws-sdk-cpp=1.10.*`, as per [recent build logs](https://github.com/apache/arrow/actions/runs/6276453828/job/17046177335?pr=37792#step:6:1372).

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #14296, opened 2023-10-18
Downgrade to Arrow 12.0.0 for aws-sdk-cpp and fix cudf_kafka builds for new CI containers

vyasr:
## Description

The aws-sdk-cpp pinning introduced in https://github.com/rapidsai/cudf/pull/14173 causes problems because newer builds of libarrow require a newer version of aws-sdk-cpp. Even though we restrict to libarrow 12.0.1, this restriction is insufficient to create solvable environments because the conda (mamba) solver doesn't seem to consistently reach far back enough into the history of builds to pull the last build that was compatible with the aws-sdk-cpp version that we need. For now, the safest way for us to avoid this problem is to downgrade to arrow 12.0.0, for which all conda package builds are pinned to the older version of aws-sdk-cpp that does not have the bug in question.

Separately, while the above issue was encountered we also got new builds of our CI images [that removed system installs of CTK packages from CUDA 12 images](https://github.com/rapidsai/ci-imgs/pull/77). This changes was made because for CUDA 12 we can get all the necessary pieces of the CTK from conda-forge. However, it turns out that the cudf_kafka builds were implicitly relying on system CTK packages, and the cudf_kafka build is in fact not fully compatible with conda-forge CTK packages because it is not using CMake via scikit-build (nor any other more sophisticated library discovery mechanism like pkg-config) and therefore does not know how to find conda-forge CTK headers/libraries. This PR introduces a set of temporary patches to get around this limitation. These patches are not a long-term fix, and are only put in place assuming that #14292 is merged in the near future before we cut a 23.12 release.

## Checklist


dependencies.yaml
```diff
@@ -225,7 +225,9 @@ dependencies:
           - &gmock gmock>=1.13.0
           # Hard pin the patch version used during the build. This must be kept
           # in sync with the version pinned in get_arrow.cmake.
-          - libarrow==12.0.1.*
+          # Temporarily building a specific build string in order to choose one
+          # that is compatible with the aws-sdk-cpp version required above
+          - libarrow==12.0.1.*=hb6645e0_10_cpu
```
bdice, line 230: Do we need to make this pinning architecture-specific? I think we’ll need a build pinning for x86-64 and aarch64. Shouldn’t need to worry about Python or CUDA versions for a C++ CPU build.

conda/environments/all_cuda-118_arch-x86_64.yaml
```diff
@@ -40,7 +40,7 @@ dependencies:
 - hypothesis
 - identify>=2.5.20
 - ipython
-- libarrow==12.0.1.*
+- libarrow==12.0.1.*=hb6645e0_10_cpu
```
jakirkham, line 43: Think one `=`s (instead of 2 `=`s) has the behavior of allowing `.*` when relevant. Idk if the `==`s is also causing us issues finding the intended version (hence mentioning this here)

```suggestion
- libarrow=12.0.1
```

conda/recipes/libcudf/conda_build_config.yaml
```diff
@@ -26,7 +26,7 @@ aws_sdk_cpp_version:
   - "<1.11"
 
 libarrow_version:
-  - "=12"
+  - "==12.0.1.*=hb6645e0_10_cpu"
```
jakirkham, line 29: Would it help to just spell out the full version here (as we only listed major before)?

```suggestion
  - "=12.0.1"
```

dependencies.yaml
```diff
@@ -263,7 +265,7 @@ dependencies:
       - output_types: [conda, requirements]
         packages:
           # Allow runtime version to float up to minor version
-          - libarrow==12.*
+          - libarrow==12.0.1.*=hb6645e0_10_cpu
```
jakirkham, line 268: Maybe `12.*` was too flexible

Also wondering if we should split this into Conda & Pip separately and then use `=`s with Conda & `==`s with Pip. It might be trying to consolidate this is loosing some amount of specificity that either tool has that would help with the problem

```suggestion
          - libarrow==12.0.1.*
```

vyasr: I'm not even sure why this is listed for pip, there is no libarrow package for that.

dependencies.yaml
vyasr: Done.

conda/recipes/cudf_kafka/meta.yaml
```diff
@@ -41,6 +41,9 @@ requirements:
     - {{ compiler('cxx') }}
     - ninja
     - sysroot_{{ target_platform }} {{ sysroot_version }}
+    {% if cuda_major == "12" %}
+    - cuda-gdb
```
bdice, line 45: Just leaving a comment to document that this is an issue caused by https://github.com/rapidsai/ci-imgs/pull/77. We removed the system CTK from CUDA 12 ci-conda images, and cudf_kafka's build script relied on a system `cuda-gdb` being found. That was a bug (!) exposed by my changes to the CI images. This should be removed in #14292 once CI is unbroken and we rewrite cudf_kafka's build system to use scikit-build.

dependencies.yaml
```diff
@@ -225,11 +226,23 @@ dependencies:
           - &gmock gmock>=1.13.0
           # Hard pin the patch version used during the build. This must be kept
           # in sync with the version pinned in get_arrow.cmake.
-          - libarrow==12.0.1.*
+          # Temporarily building a specific build string in order to choose one
+          # that is compatible with the aws-sdk-cpp version required above
```
bdice, line 230: This comment (and the one above) belong next to the `libarrow` pinning, not next to `librdkafka`.

dependencies.yaml
```diff
@@ -282,6 +301,7 @@ dependencies:
               - cuda-nvrtc-dev
               - cuda-nvtx-dev
               - libcurand-dev
+              - cuda-gdb
```
bdice, line 304:
```suggestion
              # TODO: Remove after PR 14292 updates cudf_kafka to use scikit-build
              - cuda-gdb
```

conda/recipes/cudf_kafka/meta.yaml
bdice, line 45:
```suggestion
    # TODO: Remove after PR 14292 updates cudf_kafka to use scikit-build
    - cuda-gdb
```

bdice:
@vyasr I left a review but I would accept doing the changes that I requested in a follow-up PR if this PR can unblock CI as it is currently.

conda/recipes/cudf_kafka/meta.yaml
```diff
@@ -41,13 +44,21 @@ requirements:
     - {{ compiler('cxx') }}
     - ninja
     - sysroot_{{ target_platform }} {{ sysroot_version }}
+    # TODO: Remove after https://github.com/rapidsai/cudf/pull/14292 updates
+    # cudf_kafka to use scikit-build
+    {% if cuda_major == "12" %}
+    - cuda-gdb
```
bdice, line 50: I think we can drop `cuda-gdb` from this dependency list if we're defining `CUDA_HOME`, since the only reason cudf_kafka uses `cuda-gdb` is to determine `CUDA_HOME` if that environment variable is not set.

dependencies.yaml
```diff
@@ -282,6 +281,9 @@ dependencies:
               - cuda-nvrtc-dev
               - cuda-nvtx-dev
               - libcurand-dev
+              # TODO: Remove after https://github.com/rapidsai/cudf/pull/14292 updates
+              # cudf_kafka to use scikit-build
+              - cuda-gdb
```
bdice, line 286: Same as above -- we can remove this since we are specifying `CUDA_HOME` in the cudf_kafka build script.

bdice (approved):
Approving as-is. I recommend merging in the current state to unblock CI, but there are several changes we will want to revert in #14292.

conda/recipes/cudf_kafka/meta.yaml
vyasr: Yes, that's correct. I had originally added cuda-gdb before realizing that the cuda-gdb-based detection in setup.py was not compatible with the conda-forge CUDA package layout anyway and therefore installing cuda-gdb doesn't really help. However, at this point I say we just merge this to unblock CI since we have incoming fixes in #14292.

conda/recipes/libcudf/conda_build_config.yaml
```diff
@@ -25,8 +25,8 @@ gtest_version:
 aws_sdk_cpp_version:
   - "<1.11"
 
-libarrow_version:
-  - "=12"
+libarrow:
```
raydouglass, line 28: Was it intentional to drop the `_version` suffix?

raydouglass (approved):

conda/recipes/libcudf/conda_build_config.yaml
vyasr: No 🙁 good catch. But now that it's done and we have CI passing, I'm inclined to merge and fix in a follow-up to unblock CI.

PR #14308, opened 2023-10-20
Minor improvements in `source_info`

vuule:
## Description
Apply a common parameter passing pattern (pass by value + move) to avoid an unnecessary `source_info` copy when passing a temporary object.
Also changed the default `io_type` in `source_info` to VOID so that the default-constructed objects are safe to use.

## Checklist


cpp/include/cudf/io/avro.hpp
```diff
@@ -56,7 +56,7 @@ class avro_reader_options {
    *
    * @param src source information used to read avro file
    */
-  explicit avro_reader_options(source_info const& src) : _source(src) {}
+  explicit avro_reader_options(source_info src) : _source{std::move(src)} {}
```
divyegala, line 59: When you pass-by-value here, isn't a copy created here? So you move after this to avoid a copy, but I don't understand how this change creates a situation where the object is being copied anyway as virtue of being pass-by-value.

vuule: Please see https://godbolt.org/z/3YPEoh9Tn
Pass-by-value does not always imply a copy. There is no copy when you pass an rvalue (e.g. temporary). The passed object is created with a move constructor. The overall effect is that we have an additional move to avoid a potentially unnecessary copy.
We use this pattern consistently in the cuIO APIs.

divyegala: Understood, that makes sense. I was trying to understand the pattern to see if this constructor would be called with a const lvalue, because that would be a copy now unlike the changes before this PR.

However, if the call site is always passing temporaries or moved non-const lvalues then we'll have no copies. 👍

divyegala (approved):

cpp/include/cudf/io/avro.hpp
shrshi: How about having the function accept a rvalue reference like `avro_reader_options(source_info &&src)` ? In that case, we can ensure that no lvalues are accepted and accidentally copied.

vuule: That is another option. But in that case we need two versions of each function - one that takes cont& and one that takes &&.
We've been using the pass-by-value + move for non-trivial options.

shrshi: Okay, got it, thanks!

shrshi (approved):

PR #14313, opened 2023-10-23
Added streams to JSON reader and writer api

shrshi:
## Description

This PR contributes to [#13744](https://github.com/rapidsai/cudf/issues/13744). 
- Added stream parameters to public APIs
`cudf::io::read_json`
`cudf::io::write_json`
- Added stream gtests
- Added copy constructor to internal JSON struct that was breaking for non-default streams

## Checklist


cpp/include/cudf/io/json.hpp
```diff
@@ -512,13 +512,15 @@ class json_reader_options_builder {
  * @endcode
  *
  * @param options Settings for controlling reading behavior
+ * @param stream Cuda stream to use for device memory operations
```
bdice, line 515: I think most parts of cudf are using this docstring. Can we align the docstrings in this PR?
```suggestion
 * @param stream CUDA stream used for device memory operations and kernel launches
```

cpp/tests/streams/io/functions_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+  std::string data = "[1, 1.1]\n[2, 2.2]\n[3, 3.3]\n";
+  cudf::io::json_reader_options in_options =
+    cudf::io::json_reader_options::builder(cudf::io::source_info{data.data(), data.size()})
+      .dtypes(std::vector<cudf::data_type>{cudf::data_type{cudf::type_id::INT32},
+                                           cudf::data_type{cudf::type_id::FLOAT64}})
+      .lines(true)
+      .legacy(true);
+  cudf::io::table_with_metadata result = cudf::io::read_json(
+    in_options, cudf::get_default_stream(), rmm::mr::get_current_device_resource());
```
bdice, line 43: The files in `cpp/tests/streams/*` need to use `cudf::test::get_default_stream()` to ensure they are also testing which stream is used. See docs: https://github.com/rapidsai/cudf/blob/e8cf0ebd517a9c81e294771a40c74ff5c4fb42da/cpp/include/cudf_test/default_stream.hpp#L27-L34
```suggestion
    in_options, cudf::test::get_default_stream(), rmm::mr::get_current_device_resource());
```
Please apply this change throughout the streams tests in this PR.

bdice (changes requested):
Nice! I have a couple of change requests, then this should be good to go.

cpp/include/cudf/io/json.hpp
shrshi: Docstring aligned

cpp/tests/streams/io/functions_test.cpp
shrshi: Change applied to streams test `cudf::get_default_stream` -> `cudf::test::get_default_stream`

bdice (approved):
Thanks!

bdice:
@shrshi Looks like there's a test failure ([link to logs](https://github.com/rapidsai/cudf/actions/runs/6619621643/job/17981358432?pr=14313#step:7:658)).

```
 C++ exception with description "cudf_identify_stream_usage found unexpected stream!" thrown in the test body.
```

This indicates that there might be an API that is using the default stream rather than a passed-in user stream.

vuule:
Can you please also remove unused `#include <cudf/utilities/default_stream.hpp>` from JSON code? Looks like a missed a few in https://github.com/rapidsai/cudf/pull/13903
IMO it fits in the PR as a part of "don't use the default stream in JSON".

shrshi:
Thank you for the feedback, @vuule and @bdice. I'm marking this PR as ready for review now.

bdice (approved):
Great job! Was the copy constructor of `column_to_strings_fn` the missing piece for the earlier errors?

shrshi:
Yes, the error was that the implicitly generated copy constructor of `column_to_strings_fn` was calling the constructor of the `string_scalar` member with the default stream.

cpp/tests/io/json_writer.cpp
```diff
@@ -542,7 +567,9 @@ TEST_F(JsonWriterTest, Unicode)
                            .lines(true)
                            .na_rep("null");
 
-  cudf::io::write_json(options_builder.build(), rmm::mr::get_current_device_resource());
+  cudf::io::write_json(options_builder.build(),
+                       cudf::test::get_default_stream(),
+                       rmm::mr::get_current_device_resource());
```
vuule, line 572: can these just be 
```suggestion
  cudf::io::write_json(options_builder.build());
```
We have the same values for the defaults. Not sure why `mr` was specified in the first place.

vuule (changes requested):
Tests can be cleaned up to exclude the default-able parameters.
Looks good otherwise!

cpp/tests/io/json_writer.cpp
vuule: now I see the difference, cudf::get_default_stream vs cudf::test::get_default_stream
does this matter for these tests? I don't see other PRs adding the stream parameter to all tests

vyasr: Yes, the stream parameter can be left as is for now. For the moment, `cudf::test::get_default_stream` is only necessary for the stream tests. That may change in the future, but probably isn't worth planning for right now.

shrshi: Thanks for the feedback! I've cleaned up the `rmm::mr` default parameter in both the calls.

vuule: Not sure I understand what "as is" is in @vyasr 's comment, but it looks like we might be adding the stream parameter to all API calls in the future so we might as do it now.

FWIW, the `mr` parameter can be removed from `write_json` calls in this file as well. I'm not blocking the PR on that, though.

vuule (approved):

cpp/src/io/json/write_json.cu
```diff
@@ -504,6 +504,25 @@ struct column_to_strings_fn {
...
+      mr_{other.mr_},
+      narep(other.narep, other.stream_),
+      struct_value_separator(other.struct_value_separator, other.stream_),
+      struct_row_begin_wrap(other.struct_row_begin_wrap, other.stream_),
+      struct_row_end_wrap(other.struct_row_end_wrap, other.stream_),
+      list_value_separator(other.list_value_separator, other.stream_),
+      list_row_begin_wrap(other.list_row_begin_wrap, other.stream_),
+      list_row_end_wrap(other.list_row_end_wrap, other.stream_)
+  {
+  }
```
karthikeyann, line 519: `column_to_strings_fn` is not meant to be copied. 
I just added the patch to fix the copy constructor from being invoked.

karthikeyann:
```
diff --git a/cpp/src/io/json/write_json.cu b/cpp/src/io/json/write_json.cu
index c98e93aec4..98b35fc943 100644
--- a/cpp/src/io/json/write_json.cu
+++ b/cpp/src/io/json/write_json.cu
@@ -633,17 +633,17 @@ struct column_to_strings_fn {
 
     auto child_string_with_null = [&]() {
       if (child_view.type().id() == type_id::STRUCT) {
-        return (*this).template operator()<cudf::struct_view>(
+        return this->template operator()<cudf::struct_view>(
           child_view,
           children_names.size() > child_index ? children_names[child_index].children
                                               : std::vector<column_name_info>{});
       } else if (child_view.type().id() == type_id::LIST) {
-        return (*this).template operator()<cudf::list_view>(child_view,
+        return this->template operator()<cudf::list_view>(child_view,
                                                             children_names.size() > child_index
                                                               ? children_names[child_index].children
                                                               : std::vector<column_name_info>{});
       } else {
-        return cudf::type_dispatcher(child_view.type(), *this, child_view);
+        return cudf::type_dispatcher<cudf::id_to_type_impl, column_to_strings_fn const&>(child_view.type(), *this, child_view);
       }
     };
     auto new_offsets = cudf::lists::detail::get_normalized_offsets(
@@ -706,17 +706,17 @@ struct column_to_strings_fn {
                      auto const& current_col = thrust::get<1>(i_current_col);
                      // Struct needs children's column names
                      if (current_col.type().id() == type_id::STRUCT) {
-                       return (*this).template operator()<cudf::struct_view>(
+                       return this->template operator()<cudf::struct_view>(
                          current_col,
                          children_names.size() > i ? children_names[i].children
                                                    : std::vector<column_name_info>{});
                      } else if (current_col.type().id() == type_id::LIST) {
-                       return (*this).template operator()<cudf::list_view>(
+                       return this->template operator()<cudf::list_view>(
                          current_col,
                          children_names.size() > i ? children_names[i].children
                                                    : std::vector<column_name_info>{});
                      } else {
-                       return cudf::type_dispatcher(current_col.type(), *this, current_col);
+                       return cudf::type_dispatcher<cudf::id_to_type_impl, column_to_strings_fn const&>(current_col.type(), *this, current_col);
                      }
                    });
 
```

karthikeyann:
```suggestion
  column_to_strings_fn(column_to_strings_fn const& other)      = delete;
```

karthikeyann (changes requested):

cpp/src/io/json/write_json.cu
karthikeyann: similarly other constructors and assignment operators can be deleted too.

shrshi: Thank you, @karthikeyann. The patch has been applied.

vuule: TIL you can get the type dispatcher to take the functor by reference.
Thanks @karthikeyann!

karthikeyann (approved):
LGTM 👍

cpp/src/io/json/write_json.cu
bdice: Very nice job on this, @karthikeyann. 👏

karthikeyann:
does codecov/patch check not stop the merge ?!

bdice:
@karthikeyann Codecov is purely informational. It tends to have a lot of noise and false positives / false negatives. We are hoping that switching to a new branching strategy (like always merging to `dev`) will someday fix these issues.

PR #14316, opened 2023-10-23
Add python tests for Parquet DELTA_BINARY_PACKED encoder

etseidl:
## Description
During the review of #14100 there was a suggestion to add a test of writing using cudf and then reading the resulting file back with pyarrow. This PR adds the necessary python bindings to perform this test.

NOTE: there is currently an issue with encoding 32-bit values where the deltas exceed 32-bits. parquet-mr and arrow truncate the deltas for the INT32 physical type and allow values to overflow, whereas cudf currently uses 64-bit deltas, which avoids the overflow, but can result in requiring 33-bits when encoding. The current cudf behavior is allowed by the specification (and in fact is readable by parquet-mr), but using the extra bit is not in the Parquet spirit of least output file size. This will be addressed in follow-on work.

## Checklist


cpp/src/io/parquet/page_enc.cu
```diff
@@ -396,7 +396,9 @@ __device__ size_t delta_data_len(Type physical_type, cudf::type_id type_id, uint
   // need max dtype_len + 1 bytes for min_delta
   // one byte per mini block for the bitwidth
   // and block_size * dtype_len bytes for the actual encoded data
-  auto const block_size = dtype_len + 1 + delta::num_mini_blocks + vals_per_block * dtype_len;
+  auto const max_bits        = dtype_len * 8 + 1;
+  auto const bytes_per_block = max_bits * vals_per_block / 8;
```
vuule, line 400: should this be rounded up?
```suggestion
  auto const bytes_per_block = cudf::util::div_rounding_up_unsafe(max_bits * vals_per_block, 8);
```

cpp/src/io/parquet/page_enc.cu
```diff
@@ -396,7 +396,9 @@ __device__ size_t delta_data_len(Type physical_type, cudf::type_id type_id, uint
   // need max dtype_len + 1 bytes for min_delta
   // one byte per mini block for the bitwidth
   // and block_size * dtype_len bytes for the actual encoded data
```
vuule, line 398: please update the comments to reflect the new math

vuule (changes requested):
a rounding concern

cpp/src/io/parquet/page_enc.cu
etseidl: Not really a concern because vals_per_block will always be a multiple of 8, but will change it.

vuule: I really thought I got something there :D

cpp/src/io/parquet/page_enc.cu
```diff
@@ -393,13 +393,20 @@ __device__ size_t delta_data_len(Type physical_type, cudf::type_id type_id, uint
...
-  // need max dtype_len + 1 bytes for min_delta
+  // need max dtype_len + 1 bytes for min_delta (because we only encode 7 bits per byte)
   // one byte per mini block for the bitwidth
-  // and block_size * dtype_len bytes for the actual encoded data
-  auto const block_size = dtype_len + 1 + delta::num_mini_blocks + vals_per_block * dtype_len;
+  auto const mini_block_header_size = dtype_len + 1 + delta::num_mini_blocks;
+  // each encoded value can be at most sizeof(type) * 8 + 1 bits
+  auto const max_bits = dtype_len * 8 + 1;
+  // each data block will then be max_bits * values per block. vals_per_block is guaranteed to be
+  // divisible by 128 (via static assert on delta::block_size), but do save division anyway.
```
vuule, line 402:
```suggestion
  // divisible by 128 (via static assert on delta::block_size), but do safe division anyway.
```

vuule (approved):

GregoryKimball:
@galipremsagar Would you please take a look at this testing improvement?

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -1335,6 +1332,26 @@ def test_delta_binary(nrows, add_nulls, tmpdir):
...
+        use_dictionary=False,
+    )
+
+    # FIXME(ets): should probably not use more bits than the data type
+    try:
+        cdf2 = cudf.from_pandas(pd.read_parquet(cudf_fname))
+    except OSError:
+        if dtype == "int32" and nrows == 100000:
+            pytest.mark.xfail(
+                reason="arrow does not support 33-bit delta encoding"
```
galipremsagar, line 1350:
```suggestion
                reason="arrow does not support 32-bit delta encoding"
```
typo?

etseidl: No, that's actually the correct wording. The cudf delta encoder will use 33 bits when encoding 32-bit integers.  Some parquet readers are ok with this (cudf, parquet-mr), some are not (arrow-cpp, arrow-rs). I have a follow-on to this PR that will restrict the encoding to 32 bits to make everyone happy, then the xfail will go away.

python/cudf/cudf/_lib/cpp/io/parquet.pxd
```diff
@@ -191,6 +199,8 @@ cdef extern from "cudf/io/parquet.hpp" namespace "cudf::io" nogil:
         void set_row_group_size_rows(size_type val) except +
         void set_max_page_size_bytes(size_t val) except +
         void set_max_page_size_rows(size_type val) except +
+        void enable_write_v2_headers(bool val) except +
```
mythrocks, line 202: I realize it isn't the fault of this current PR, but one does wish `enable_write_v2_headers` were named `set_write_v2_headers`.

mythrocks (approved):
Not a cuIO dev, but this looks good to me.

A minor nitpick regarding the naming of a function, but that's unrelated to this change.

python/cudf/cudf/_lib/cpp/io/parquet.pxd
vuule: We use enable_ for bool options, so this should be consistent (for better or for worse, apparently).

python/cudf/cudf/core/dataframe.py
```diff
@@ -6370,6 +6370,8 @@ def to_parquet(
         max_page_size_rows=None,
         storage_options=None,
         return_metadata=False,
+        use_dictionary=True,
```
galipremsagar, line 6373: Can we document these two parameters here: https://github.com/rapidsai/cudf/blob/16051a718509c218010e6912d6f8e0fca7a7aa24/python/cudf/cudf/utils/ioutils.py#L222

python/cudf/cudf/tests/test_parquet.py
galipremsagar: TIL. Thanks @etseidl !

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -1335,6 +1332,26 @@ def test_delta_binary(nrows, add_nulls, tmpdir):
...
+            )
+    else:
```
galipremsagar, line 1352:
```suggestion
    except OSError as e:
        if dtype == "int32" and nrows == 100000:
            pytest.mark.xfail(
                reason="arrow does not support 33-bit delta encoding"
            )
        else:
            raise e
    else:
```
We will have to raise for else case.

galipremsagar (changes requested):

python/cudf/cudf/core/dataframe.py
etseidl: done

etseidl: Thanks @galipremsagar. I would have never thought to look there for the docstring :sweat_smile:

vuule: First time hearing about it as well :man_shrugging: (just don't git blame `ioutils.py` :P )

galipremsagar (approved):

PR #14321, opened 2023-10-23
Avoid `pyarrow.fs` import for local storage

rjzamora:
## Description
This is not a resolution, but may help mitigate problems from https://github.com/aws/aws-sdk-cpp/issues/2681

## Checklist


pentschev (approved):
I'm +1000 on this change. Thanks @rjzamora !

python/cudf/cudf/utils/ioutils.py
```diff
@@ -1582,6 +1581,8 @@ def _open_remote_files(
         Key-word arguments to be passed to format-specific
         open functions.
     """
+    # Avoid top-level pyarrow.fs import
+    from pyarrow.fs import FSSpecHandler, PyFileSystem
```
wence-, line 1585: **suggestion:** ‏Can we scope this even narrower to just before line 1633? There are a bunch of cases where even if we're opening a remote file we don't do so via the pyarrow filesystem interface.

wence-:
```suggestion
    # Importing pyarrow.fs initializes a S3 SDK with a finalizer
    # that runs atexit. In some circumstances it appears this
    # runs a call into a logging system that is already shutdown.
    # To avoid this, we only import this subsystem if it is
    # really needed.
    # See https://github.com/aws/aws-sdk-cpp/issues/2681
    # Avoid top-level pyarrow.fs import
    from pyarrow.fs import FSSpecHandler, PyFileSystem
```

Let's signpost why this is needed.

wence- (approved):
Minor suggestions, but LGTM

python/cudf/cudf/utils/ioutils.py
pentschev: I was just thinking about it and was coming here to suggest this comment as well. +1

wence-:
Is this sufficient? The import of pyarrow.dataset (here https://github.com/rapidsai/cudf/blob/branch-23.12/python/cudf/cudf/io/parquet.py#L18) also seems to provoke importing `pyarrow._fs`.

Can we add a test (maybe run in a subprocess) that checks that `import cudf; "pyarrow._fs" in sys.modules` returns False?

python/cudf/cudf/utils/ioutils.py
rjzamora: Good call on both suggestions.

bdice (approved):
I support @wence-'s request for a test but otherwise LGTM.

rjzamora:
> Is this sufficient? The import of pyarrow.dataset (here https://github.com/rapidsai/cudf/blob/branch-23.12/python/cudf/cudf/io/parquet.py#L18) also seems to provoke importing pyarrow._fs.

This is a good point. However, do we care about `pyarrow._fs`, or just `pyarrow._s3fs`? As far as I can tell, importing `pyarrow.dataset` does bring in `pyarrow._fs`, but `pyarrow._s3fs` isn't loaded unless `pyarrow.fs` is imported.

```python
import sys
import pyarrow
pa_mods = set(sys.modules)
_fs = "pyarrow._fs"
_s3fs = "pyarrow._s3fs"

print("import pyarrow.dataset")
import pyarrow.dataset
ds_mods = set(sys.modules)
print(f"{_fs} imported: {_fs in (ds_mods - pa_mods)}")
print(f"{_s3fs} imported: {_s3fs in (ds_mods - pa_mods)}")

print("import pyarrow.fs")
import pyarrow.fs
fs_mods = set(sys.modules)
print(f"{_s3fs} imported: {_s3fs in (fs_mods - ds_mods)}")
```
```
import pyarrow.dataset
pyarrow._fs imported: True
pyarrow._s3fs imported: False
import pyarrow.fs
pyarrow._s3fs imported: True
```

>Can we add a test (maybe run in a subprocess) that checks that import cudf; "pyarrow._fs" in sys.modules returns False?

A test is a good idea. Let me know if you think that we can just check that `pyarrow._s3fs` isn't imported.

python/cudf/cudf/tests/test_s3.py
```diff
@@ -533,3 +533,17 @@ def test_write_chunked_parquet(s3_base, s3so):
             actual.sort_values(["b"]).reset_index(drop=True),
             cudf.concat([df1, df2]).sort_values(["b"]).reset_index(drop=True),
         )
+
+
+def test_no_s3fs_on_cudf_import():
```
rjzamora, line 538: This test is useful, but adding it has rained a bit on our mini parade. We now see that the `pyarrow._s3fs` problem pops up whenever we import `pyarrow.parquet` or `pyarrow.orc`. (So, much less constrained than the "remove storage only" case we originally expected).

pentschev: Thanks for doing that investigation Rick. I see you've already adjusted the remainder of the code to take that into consideration and this test is still passing, so I'm good with things as they are now.

pentschev (approved):

pentschev:
Given sentiment was positive regarding this change and to reduce probability of segfaults in Dask et al., I'm merging this. Thanks a lot @rjzamora for working on this change!

PR #14322, opened 2023-10-24
Parquet writer - fix logical type for timestamp columns

### LINKED ISSUE #14315: [BUG] Parquet writer encodes timestamp statistics incorrectly
**Describe the bug**
After #14264, Parquet files written with timestamp columns have the timestamp statistics for the column in the footer encoded incorrectly.  Apache Spark is no longer able to properly perform predicate pushdown on timestamps for these files.

Attached are two files with equivalent contents, one was created with the RAPIDS Accelerator for Apache Spark with cudf before #14264 and the other is the same software except cudf has been rolled back just one commit to before #14264.  Spark can properly evaluate timestamp predicates on the file created before #14264 and fails to do so afterwards, always loading all files.

You can see the failed predicate pushdown in Spark by running the following query in Spark and examining the Spark UI to see how many rows were produced by the file scan before it is filtered by the filter.  Replace "/tmp/pqgood" with the appropriate path to the good or bad Parquet files.
```
spark.read.parquet("/tmp/pqgood").filter(col("a") === java.sql.Timestamp.valueOf("1970-01-01 00:00:00.500")).collect
```

In the "good" case, Spark should load _less_ than the total number of rows because it can properly see via the footers in the separate files that only one file is worth loading.  In the "bad" case, Spark loads _all_ of the data because it cannot properly detect that only one file could possibly contain the value trying to be found by the query.

**Expected behavior**
Apache Spark should be able to correctly perform predicate pushdown on timestamp columns for Parquet files written by libcudf.

[pqbad.zip](https://github.com/rapidsai/cudf/files/13077994/pqbad.zip)
[pqgood.zip](https://github.com/rapidsai/cudf/files/13077997/pqgood.zip)

vuule:
## Description
Closes #14315

Parquet writer writes time and timestamp types with logical type with `isAdjustedToUTC` as `false`. However, timestamps in libcudf tables are implicitly in UTC and don't need to be adjusted.
This PR changes the `isAdjustedToUTC` to true.

## Checklist

### DIFF
--- a/cpp/src/io/parquet/parquet.hpp
+++ b/cpp/src/io/parquet/parquet.hpp
@@ -58,13 +58,15 @@ struct TimeUnit {
 };
 
 struct TimeType {
-  bool isAdjustedToUTC = false;
+  bool isAdjustedToUTC = true;
   TimeUnit unit;
+  explicit TimeType(TimeUnit::Type t = TimeUnit::MILLIS) : unit{t} {}
 };
 
 struct TimestampType {
-  bool isAdjustedToUTC = false;
+  bool isAdjustedToUTC = true;
   TimeUnit unit;
+  explicit TimestampType(TimeUnit::Type t = TimeUnit::MILLIS) : unit{t} {}
 };
 
 struct IntType {
--- a/cpp/src/io/parquet/writer_impl.cu
+++ b/cpp/src/io/parquet/writer_impl.cu
@@ -404,7 +404,7 @@ struct leaf_schema_fn {
     col_schema.ts_scale    = 1000;
     if (not timestamp_is_int96) {
       col_schema.converted_type = ConvertedType::TIMESTAMP_MILLIS;
-      col_schema.logical_type   = LogicalType{TimestampType{false, TimeUnit::MILLIS}};
+      col_schema.logical_type   = LogicalType{TimestampType{TimeUnit::MILLIS}};
     }
   }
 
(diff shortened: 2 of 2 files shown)


cpp/src/io/parquet/parquet.hpp
```diff
@@ -58,13 +58,15 @@ struct TimeUnit {
 };
 
 struct TimeType {
-  bool isAdjustedToUTC = false;
+  bool isAdjustedToUTC = true;
```
etseidl, line 61: Could you leave a comment here explaining why the default is true?

vuule:
CC @galipremsagar @rjzamora for impact on the Python side
We have failing tests because pandas sets `AdjustedToUTC` to false by default so we get type mismatch.

jlowe (commented):
Built a new libcudf with this change and verified the failing Spark Parquet predicate pushdown tests for timestamps are passing again.

vuule:
Thank you @etseidl for the `merge_row_group_metadata` patch!

galipremsagar (commented):
@vuule Can address the todo here(https://github.com/rapidsai/cudf/pull/14327/files#diff-ba3ea8b1c9c17ff8301a86d6a844b075f4f49feaa782f4a8f190e631123b007eR484) since this PR is fixing https://github.com/rapidsai/cudf/issues/14326

davidwendt (approved):

galipremsagar (approved):
Thank you both so much for tackling this @vuule @etseidl !

hyperbolic2346 (approved):

rjzamora:
Thanks so much @vuule and @etseidl !

PR #14323, opened 2023-10-24
Drop `pyorc` dependency and use `pandas`/`pyarrow` instead

galipremsagar:
## Description

This PR removes dependency on `pyorc` in `cudf` altogether by using drop-in replacements found in `pandas` & `pyarrow`.

## Checklist

### DIFF
--- a/python/cudf/cudf/_fuzz_testing/orc.py
+++ b/python/cudf/cudf/_fuzz_testing/orc.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
 
 import copy
 import io
@@ -6,14 +6,14 @@ import logging
 import random
 
 import numpy as np
-import pyorc
+import pyarrow.orc
+import pyarrow as pa
 
 import cudf
 from cudf._fuzz_testing.io import IOFuzz
 from cudf._fuzz_testing.utils import (
     ALL_POSSIBLE_VALUES,
     _generate_rand_meta,
-    pandas_to_orc,
     pyarrow_to_pandas,
 )
 from cudf.testing import dataset_generator as dg
@@ -82,12 +82,7 @@ class OrcReader(IOFuzz):
         logging.info(f"Shape of DataFrame generated: {table.shape}")
         self._df = df
         file_obj = io.BytesIO()
-        pandas_to_orc(
-            df,
-            file_io_obj=file_obj,
-            stripe_size=self._rand(len(df)),
-            arrow_table_schema=table.schema,
-        )
+        pa.orc.write_table(table, file_obj, stripe_size=self._rand(len(df)))
         file_obj.seek(0)
         buf = file_obj.read()
         self._current_buffer = copy.copy(buf)
@@ -109,8 +104,8 @@ class OrcReader(IOFuzz):
                     )
                 elif param == "stripes":
                     f = io.BytesIO(self._current_buffer)
-                    reader = pyorc.Reader(f)
-                    stripes = [i for i in range(reader.num_of_stripes)]
+                    orcFile = pa.orc.ORCFile(f)
+                    stripes = [i for i in range(orcFile.nstripes)]
                     params_dict[param] = np.random.choice(
                         [
                             None,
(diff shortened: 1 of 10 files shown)


python/cudf/cudf/tests/test_orc.py
```diff
@@ -1024,50 +992,59 @@ def generate_list_struct_buff(size=100_000):
...
         rd.choice(
             [
                 None,
-                (
-                    rd.choice([None, np.random.randint(0, 3)]),
-                    (
-                        rd.choice([None, np.random.randint(0, 3)]),
-                        np.random.randint(0, 3),
-                    ),
+                *(
```
bdice, line 1025: Is this unpacking needed? I think this is like saying `[*(a, b)]` instead of `[a, b]`.

python/cudf/cudf/tests/test_orc.py
```diff
@@ -1024,50 +992,59 @@ def generate_list_struct_buff(size=100_000):
...
+                    {
+                        "lvl1_struct": {
+                            "c": rd.choice([None, np.random.randint(0, 3)]),
+                            "d": np.random.randint(0, 3),
+                        },
+                    },
                 ),
             ]
         )
         for x in range(size)
```
bdice, line 1036: Is this another dummy variable that can be replaced with `_`?
```suggestion
        for _ in range(size)
```

python/cudf/cudf/tests/test_orc.py
```diff
@@ -1024,50 +992,59 @@ def generate_list_struct_buff(size=100_000):
...
     ]
     list_nests_struct = [
         [
-            [rd.choice(lvl1_struct), rd.choice(lvl1_struct)]
+            {"a": rd.choice(lvl1_struct), "b": rd.choice(lvl1_struct)}
             for y in range(np.random.randint(1, 4))
```
bdice, line 1041: Maybe do one more pass for dummy variables, and replace them with `_` if possible.
```suggestion
            for _ in range(np.random.randint(1, 4))
```

python/cudf/cudf/tests/test_orc.py
```diff
@@ -1160,107 +1129,89 @@ def gen_map_buff(size=10000):
...
+        [
+            rd.choice(
                 [
-                    (
-                        rd.choice(al),
-                        rd.choice(
-                            [
-                                None,
+                    None,
+                    *(
```
bdice, line 1153: Same question as above about unnecessary unpacking. Check for more instances of that in this file.

bdice (commented):
Great. I have a few questions. Also be aware that https://github.com/rapidsai/cudf/pull/14321 tells us we should be cautious about importing `pyarrow.orc` except where it's necessary, since it will trigger S3 initialization.

python/cudf/cudf/tests/test_orc.py
galipremsagar: We will need it here because it is a comprehension (for _ in range(2))

python/cudf/cudf/tests/test_orc.py
galipremsagar: This isn't needed, dropped it.

galipremsagar:
> Great. I have a few questions. Also be aware that #14321 tells us we should be cautious about importing `pyarrow.orc` except where it's necessary, since it will trigger S3 initialization.

Removed the imports where unnecessary 👍

vuule (approved):
Thanks for making this change!
LGTM, but I am a bit out of by Python depth here.

AyodeAwe (approved):

python/cudf/cudf/_fuzz_testing/orc.py
```diff
@@ -109,8 +103,8 @@ def set_rand_params(self, params):
                     )
                 elif param == "stripes":
                     f = io.BytesIO(self._current_buffer)
-                    reader = pyorc.Reader(f)
-                    stripes = [i for i in range(reader.num_of_stripes)]
+                    orcFile = pa.orc.ORCFile(f)
+                    stripes = [i for i in range(orcFile.nstripes)]
```
vyasr, line 107:
```suggestion
                    stripes = list(range(orcFile.nstripes))
```

vyasr (approved):
One small suggestion, otherwise the substitution LGTM. There are a few unnecessary extra changes, but they are cleanup so may as well merge them now that they're already done.

python/cudf/cudf/tests/test_orc.py
bdice: Yes, I see that now.

bdice (approved):

Issue #14326, opened 2023-10-25
[BUG] Null types are not promoted in `cudf.io.merge_parquet_filemetadata`


rjzamora:
**Describe the bug**
There is a minor bug in `cudf.io.merge_parquet_filemetadata` after https://github.com/rapidsai/cudf/pull/14264 was merged. Some details are discussed in that PR. E.g. https://github.com/rapidsai/cudf/pull/14264#issuecomment-1778311615

**Steps/Code to reproduce bug**
```python
import pyarrow.dataset as ds
import cudf
import fsspec
from dask_cudf.io.parquet import CudfEngine
import pandas as pd
paths = ["/tmp/part.0.parquet", "/tmp/part.1.parquet"]
pd.DataFrame({"a": [None] * 10}).to_parquet(paths[0])
pd.DataFrame({"a": range(10)}).to_parquet(paths[1])
md1 = CudfEngine.collect_file_metadata(paths[0], fsspec, "none")
md2 = CudfEngine.collect_file_metadata(paths[1], fsspec, "none")
md = cudf.io.merge_parquet_filemetadata([md1, md2])
with open("/tmp/_metadata", "wb") as fil:
    fil.write(memoryview(md))
schema = ds.parquet_dataset("/tmp/_metadata").schema
assert schema.field(0).type !=  “null”
```
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[3], line 15
     13     fil.write(memoryview(md))
     14 schema = ds.parquet_dataset("/tmp/_metadata").schema
---> 15 assert schema.field(0).type != "null"

AssertionError: 
```

**Expected behavior**
Null column metadata should be "promoted" when it is merged with integer metedata for the same column.

**Additional context**
This bug is unlikely to affect many users, and can be treated as a low priority if the fix proves difficult.

PR #14327, opened 2023-10-25
Run IO tests for Dask-cuDF

rjzamora:
## Description
We are not currently running any IO tests for `dask_cudf` in CI. This PR should correct this. It also modifies a test that *would* be failing due to https://github.com/rapidsai/cudf/issues/14326

## Checklist

### DIFF
--- a/ci/test_wheel_dask_cudf.sh
+++ b/ci/test_wheel_dask_cudf.sh
@@ -16,4 +16,6 @@ python -m pip install git+https://github.com/dask/dask.git@2023.9.2 git+https://
 # echo to expand wildcard before adding `[extra]` requires for pip
 python -m pip install $(echo ./dist/dask_cudf*.whl)[test]
 
+# Run tests in dask_cudf/tests and dask_cudf/io/tests
 python -m pytest -n 8 ./python/dask_cudf/dask_cudf/tests
+python -m pytest -n 8 ./python/dask_cudf/dask_cudf/io/tests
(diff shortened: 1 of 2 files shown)


ci/test_wheel_dask_cudf.sh
```diff
@@ -16,4 +16,6 @@ python -m pip install git+https://github.com/dask/dask.git@2023.9.2 git+https://
 # echo to expand wildcard before adding `[extra]` requires for pip
 python -m pip install $(echo ./dist/dask_cudf*.whl)[test]
 
+# Run tests in dask_cudf/tests and dask_cudf/io/tests
 python -m pytest -n 8 ./python/dask_cudf/dask_cudf/tests
+python -m pytest -n 8 ./python/dask_cudf/dask_cudf/io/tests
```
galipremsagar, line 21:
```suggestion
python -m pytest -n 8 ./python/dask_cudf/dask_cudf/
```

Should we rather do this?

cc: @vyasr if you have any thoughts on this

rjzamora:
cc @galipremsagar (This is why we have not been catching the recent `dask_cudf` failures in CI)

ci/test_wheel_dask_cudf.sh
rjzamora: Right - Was about to comment on this. Seems fine to me either way, so I'll go with whatever others like.

bdice: Let's adopt Prem's suggestion and rely on pytest discovery.

galipremsagar (approved):
Thanks for finding this @rjzamora !

raydouglass (approved):

Issue #14311, opened 2023-10-23
[BUG] LIBCUDF_IO_PREFER_PAGEABLE_TMP_MEMORY=1 causes cudaErrorIllegalAddress in parquet reads


abellina:
I have been trying to test some changes in the host memory side for parquet decode. One of the issues we see in traces with the help from the CUDA team is that `cudaFreeHost` is introducing synchronization that can be mitigated by pooling for example. Another test I'd like to make is to use pageable memory instead of pinned memory, specifically in `hostdevice_vector`. 

I see there is already a flag to make this happen: `LIBCUDF_IO_PREFER_PAGEABLE_TMP_MEMORY`, where I expect that we'd use pageable memory via a `thrust` vector. I tried this and I am getting all kinds of `cudaErrorIllegalAddress` errors immediately and thought to report since it seems that this flag should work. I am curious to know why just going from pinned to pageable could cause this (likely some inherent synchronization we are now missing??)

A simple repro case is:

`LIBCUDF_IO_PREFER_PAGEABLE_TMP_MEMORY=1 ./PARQUET_TEST` (run this under the `gtests` folder`).

Small snippet:

```
...
[  FAILED  ] ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/9, where GetParam() = 9 (704 ms)
[ RUN      ] ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/10
unknown file: Failure
C++ exception with description "Fatal CUDA error encountered at: /home/abellina/work/cudf/cpp/include/cudf/detail/utilities/vector_factories.hpp:100: 700 cudaErrorIllegalAddress an illegal memory access was encountered" thrown in the test body.

[  FAILED  ] ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/10, where GetParam() = 10 (710 ms)
[ RUN      ] ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/11
unknown file: Failure
C++ exception with description "Fatal CUDA error encountered at: /home/abellina/work/cudf/cpp/include/cudf/detail/utilities/vector_factories.hpp:100: 700 cudaErrorIllegalAddress an illegal memory access was encountered" thrown in the test body.

[  FAILED  ] ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/11, where GetParam() = 11 (737 ms)
[ RUN      ] ParquetDictionaryTest/ParquetSizedTest.DictionaryTest/12
unknown file: Failure
[...]

PR #14328, opened 2023-10-25
Fix host buffer access from device function in the Parquet reader

vuule:
## Description
Closes https://github.com/rapidsai/cudf/issues/14311
The host access becomes an issue when pageable memory is used for host side of `hostdevice_vector`.
This PR fixes the device lambda to use the device pointer instead.

## Checklist

### DIFF
--- a/cpp/src/io/parquet/reader_impl_preprocess.cu
+++ b/cpp/src/io/parquet/reader_impl_preprocess.cu
@@ -310,8 +310,8 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
   }
 
   // compute max bytes needed for level data
-  auto level_bit_size =
-    cudf::detail::make_counting_transform_iterator(0, [chunks = chunks.begin()] __device__(int i) {
+  auto level_bit_size = cudf::detail::make_counting_transform_iterator(
+    0, [chunks = chunks.d_begin()] __device__(int i) {
       auto c = chunks[i];
       return static_cast<int>(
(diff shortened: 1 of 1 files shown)


shrshi (commented):
Looks good to me, thank you!

shrshi (approved):

vyasr (approved):

ttnghia (approved):
Is there any test for the related function? If yes, how this didn't crash the test?

vuule:
> Is there any test for the related function? If yes, how this didn't crash the test?

The use of pinned memory was hiding the issue. When using pageable memory in `hostdevice_vector`s, all of the Parquet reader tests crash.

ttnghia:
> The use of pinned memory was hiding the issue. When using pageable memory in `hostdevice_vector`s, all of the Parquet reader tests crash.

Then can we have a test with both pinned memory/pageable memory?

vuule:
> > The use of pinned memory was hiding the issue. When using pageable memory in `hostdevice_vector`s, all of the Parquet reader tests crash.
> 
> Then can we have a test with both pinned memory/pageable memory? Such as through macro definition or toggling the corresponding env variable?

I've started working on the test and realized some problems. The env var is read only once and cached. We can't enable pageable memory use for only one test.
We would also need separate tests for each format.
I don't think we should be adding tests for this at this time. Maybe if/when the option gets some use.

vuule:
Merged to enable spark experiments; we can add tests in a separate PR.

PR #14332, opened 2023-10-25
Temporarily avoid the current build of pydata-sphinx-theme

vyasr:
## Description

There appears to be a bug in the latest release of pydata-sphinx theme where a warning that was intended to be thrown conditionally is appearing unconditionally in our builds, triggering a doc build failure because we build with `-W`. See https://github.com/pydata/pydata-sphinx-theme/issues/1539 for more information. We should be OK to simply avoid the current version for now. If the next release is a minor release then the warning will be removed and we can automatically upgrade. If they have another patch release then we can reevaluate the pinning, but the current pinning seems like the most likely to require no additional work going forward.

## Checklist

### DIFF
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -69,7 +69,7 @@ dependencies:
 - protobuf>=4.21,<5
 - ptxcompiler
 - pyarrow==12.0.1.*
-- pydata-sphinx-theme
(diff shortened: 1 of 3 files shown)


ajschmidt8 (approved):
LGTM.

This is a good time to point out the new RAPIDS sphinx theme below that we're currently working on. In the future, that theme will handle the PyData theme pin for all of RAPIDS. Allan and I are working on that as time permits.

- https://github.com/rapidsai/sphinx-theme/

PR #14336, opened 2023-10-26
Improve performance of nvtext::tokenize_with_vocabulary for long strings

davidwendt:
## Description
Improves `nvtext::tokenize_with_vocabulary` performance for long strings. Also adds additional tests and an nvbench benchmark.

## Checklist


davidwendt:
Performance numbers for long strings from the included benchmark
```
| width |  num_rows  |   Ref Time |   Cmp Time |          Diff |   %Diff |      |
|-------|------------|------------|------------|---------------|---------|------|
|  256  |   262144   |   3.267 ms |   1.694 ms |  -1572.363 us | -48.13% | 1.93 |
|  512  |   262144   |   8.746 ms |   3.193 ms |  -5553.180 us | -63.49% | 2.74 |
| 1024  |   262144   |  20.265 ms |   6.129 ms | -14135.536 us | -69.75% | 3.31 |
|  256  |   524288   |   6.344 ms |   3.326 ms |  -3018.149 us | -47.57% | 1.91 |
|  512  |   524288   |  18.740 ms |   6.387 ms | -12352.539 us | -65.92% | 2.93 |
| 1024  |   524288   |  42.879 ms |  12.394 ms | -30485.431 us | -71.10% | 3.46 |
|  256  |  1048576   |  11.448 ms |   6.693 ms |  -4755.292 us | -41.54% | 1.71 |
|  512  |  1048576   |  34.554 ms |  12.878 ms | -21676.282 us | -62.73% | 2.68 |
| 1024  |  1048576   |  80.785 ms |  24.909 ms | -55876.306 us | -69.17% | 3.24 |
|  256  |  2097152   |  23.507 ms |  13.482 ms | -10025.858 us | -42.65% | 1.74 |
|  512  |  2097152   |  69.627 ms |  26.090 ms | -43537.188 us | -62.53% | 2.67 |
|  256  |  4194304   |  45.730 ms |  27.060 ms | -18669.250 us | -40.83% | 1.69 |
```

cpp/src/text/vocabulary_tokenize.cu
```diff
@@ -162,6 +173,122 @@ std::unique_ptr<tokenize_vocabulary> load_vocabulary(cudf::strings_column_view c
...
+  }
+  __syncwarp();
+
+  for (auto itr = d_output + lane_idx + 1; itr < d_output_end; itr += cudf::detail::warp_size) {
+    auto index = static_cast<cudf::size_type>(thrust::distance(d_output, itr));
+    if (*itr) {
+      count += !(*(itr - 1));
+    } else if (itr + 1 == d_output_end) {
+      ++count;
+    }
```
bdice, line 284: Is this equivalent? It seems like there are a few cases that add 1, but they're spelled differently.
```suggestion
    if ((*itr && !(*(itr - 1))) || (itr + 1 == d_output_end)) {
      ++count;
    }
```

bdice (approved):
All looks good. I have one suggested rewrite for some of the counting math.

cpp/src/text/vocabulary_tokenize.cu
mythrocks: I initially thought the original version was more readable.  I'm now leaning towards preferring @bdice's version.

davidwendt: See the current implementation. I added a comment here as well that I helps.

mythrocks: Forgive me, but I think this one-liner is too terse now.

I'll go along with it, if the consensus is that this is best.  But I don't think this would be readable to me, had I not seen the evolution from the original version.

bdice: The key outcome I want is to have only one way to get to "count++". You could also achieve that by declaring two named bools and OR-ing them, rather than putting both conditions directly into the if.

davidwendt: Surely the comment above the line should alleviate any concerns?

bdice: Sorry. I was replying to Mithun on the general topic of terseness / one-liners. I like the current state of the code, the comment is helpful here.

mythrocks (approved):
Thanks for the comment and explanation. +1.

cpp/src/text/vocabulary_tokenize.cu
mythrocks: > The key outcome I want is to have only one way to get to "count++". 

Thanks, that helped me grok this better.

davidwendt: Turns out this is not equivalent to the original if-else.
Here is the desired logic table:
```
a = *itr
b = !(*(itr-1))
c = (itr + 1) == d_output_end
r = count
0 = false, 1 =true

 a  b  c    r
 0  0  0    -
 0  0  1   +1
 0  1  0    -
 0  1  1   +1
 1  0  0    -
 1  0  1    -     <<<< new logic incorrectly adds 1 to count here
 1  1  0   +1
 1  1  1   +1

if( (a && b) || c) {r++}  is not the same as  if(a) {r+=b} else if(c) {r++}
```
(Sorry, the table may be confusing since `!*(itr-1) == true` when the value is actually false).
Anyway, the new code is causing failures that my tests are not catching so I need to revert this change and add some additional tests.

bdice: Yes, I agree. I see why -- you need to skip checking `c` if `a` is true. I tried to work that truth table out in my head while reviewing but made an error in that case. Apologies for the incorrect suggestion.

PR #14337, opened 2023-10-26
Rework nvtext::byte_pair_encoding API

davidwendt:
## Description
Rewrite of the `nvtext::byte_pair_encoding` API to better match behavior and performance requirements.
The API now strictly does BPE without any tokenizing (word delimiting) on the input.
The code also has been rewritten to improve performance on very long strings.

## Checklist

### DIFF
--- a/cpp/include/nvtext/byte_pair_encoding.hpp
+++ b/cpp/include/nvtext/byte_pair_encoding.hpp
@@ -36,7 +36,7 @@ namespace nvtext {
  */
 struct bpe_merge_pairs {
   struct bpe_merge_pairs_impl;
-  std::unique_ptr<bpe_merge_pairs_impl> impl{};  ///< Implementation of the BPE merge pairs table.
+  bpe_merge_pairs_impl* impl{};  ///< Implementation of the BPE merge pairs table.
 
   /**
    * @brief Construct a new bpe merge pairs object
@@ -61,6 +61,7 @@ struct bpe_merge_pairs {
                   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
   ~bpe_merge_pairs();
+  bpe_merge_pairs();
 };
 
 /**
--- a/cpp/src/text/bpe/byte_pair_encoding.cu
+++ b/cpp/src/text/bpe/byte_pair_encoding.cu
@@ -21,8 +21,10 @@
 #include <cudf/column/column_device_view.cuh>
 #include <cudf/column/column_factories.hpp>
 #include <cudf/detail/get_value.cuh>
+#include <cudf/detail/null_mask.hpp>
 #include <cudf/detail/nvtx/ranges.hpp>
-#include <cudf/strings/detail/combine.hpp>
+#include <cudf/detail/sizes_to_offsets_iterator.cuh>
+#include <cudf/detail/utilities/cuda.cuh>
 #include <cudf/strings/detail/strings_children.cuh>
 #include <cudf/utilities/default_stream.hpp>
 #include <cudf/utilities/error.hpp>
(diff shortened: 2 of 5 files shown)


cpp/src/text/bpe/byte_pair_encoding.cu
```diff
@@ -21,477 +21,306 @@
...
+struct bpe_unpairable_offsets_fn {
+  cudf::device_span<char const> d_chars;
+  cudf::size_type offset;
+  MapRefType const d_map;
+  __device__ cudf::size_type operator()(cudf::size_type idx)
+  {
+    if (!cudf::strings::detail::is_begin_utf8_char(d_chars[idx])) { return 0; }
+
+    auto const itr  = d_chars.data() + idx;
+    auto const end  = d_chars.end();  // + chars_size;
```
bdice, line 72: What does the comment `+ chars_size` mean here?

cpp/src/text/bpe/byte_pair_encoding.cu
```diff
@@ -21,477 +21,306 @@
...
+    if (next < end) {
+      auto const rhs = cudf::string_view(next, cudf::strings::detail::bytes_in_utf8_byte(*next));
+      // see if both halves exist anywhere in the table, if not these are unpairable
+      if (d_map.find(lhs) == d_map.end() && d_map.find(rhs) == d_map.end()) {
+        output = idx + lhs.size_bytes() + offset;  // offset for artificial boundary
+      }
+    }
+    return output;
+
+    // Alternate solution that only checks one substring.
```
bdice, line 85: Do we need this alternate solution in a comment or can it be cleaned up?

cpp/src/text/bpe/byte_pair_encoding.cu
```diff
@@ -21,477 +21,306 @@
...
-      return;
+  // for finding the next half of a pair
+  auto next_substr = [d_str, d_spaces, end = end_spaces](int8_t* begin) {
+    auto const next = thrust::find(thrust::seq, begin + 1, end, 1);
+    auto const size = static_cast<cudf::size_type>(thrust::distance(begin, next));
+    return cudf::string_view(d_str.data() + thrust::distance(d_spaces, begin), size);
+  };
+  // for locating adjacent pairs after merging a pair
+  auto find_prev = [begin = d_spaces](int8_t* ptr) {
+    while (ptr > begin && *ptr == 0) {
```
bdice, line 179: Should this be a find algorithm on a reverse iterator? Can binary search be used instead of repeated decrements, or is that not appropriate here? (Still wrapping my head around the algorithm's behavior.)

cpp/src/text/bpe/byte_pair_encoding.cu
```diff
@@ -21,477 +21,306 @@
...
+    auto const next_itr = itr + lhs.size_bytes();
+    if (next_itr < end_spaces) {
+      auto const rhs = next_substr(next_itr);  // retrieve rhs of the pair
+      if (!rhs.empty()) {
+        auto rank          = max_rank;
+        auto const mp      = merge_pair_type{lhs, rhs};
+        auto const map_itr = d_map.find(mp);                     // lookup pair in merges table;
+        if (map_itr != d_map.end()) { rank = map_itr->second; }  // found a match;
+        d_ranks[thrust::distance(d_spaces, next_itr)] = rank;    // store the rank
+        if (rank < min_rank) min_rank = rank;
```
bdice, line 201:
```suggestion
        if (rank < min_rank) { min_rank = rank; }
```

cpp/src/text/bpe/byte_pair_encoding.cu
```diff
@@ -500,54 +329,120 @@ std::unique_ptr<cudf::column> byte_pair_encoding(cudf::strings_column_view const
...
+  }
+
+  // compute the output sizes and store them in the d_offsets vector
+  bpe_finalize<<<input.size(), block_size, 0, stream.value()>>>(
+    *d_strings, d_spaces.data(), d_offsets);
+
+  // convert sizes to offsets in-place
+  auto const bytes =
+    cudf::detail::sizes_to_offsets(d_offsets, d_offsets + input.size() + 1, d_offsets, stream);
+  CUDF_EXPECTS(bytes <= static_cast<int64_t>(std::numeric_limits<cudf::size_type>::max()),
```
bdice, line 414: It seems like we use `size_t` more often than `int64_t` here.

```suggestion
  CUDF_EXPECTS(bytes <= static_cast<std::size_t>(std::numeric_limits<cudf::size_type>::max()),
```

Examples:
- https://github.com/rapidsai/cudf/blob/2a923dfff84a8ce3422a126e9d0035282a0bea2f/cpp/src/copying/scatter.cu#L324
- https://github.com/rapidsai/cudf/blob/2a923dfff84a8ce3422a126e9d0035282a0bea2f/cpp/src/copying/gather.cu#L67
- https://github.com/rapidsai/cudf/blob/2a923dfff84a8ce3422a126e9d0035282a0bea2f/cpp/src/scalar/scalar.cpp#L70
- https://github.com/rapidsai/cudf/blob/2a923dfff84a8ce3422a126e9d0035282a0bea2f/cpp/src/text/normalize.cu#L218

cpp/src/text/bpe/byte_pair_encoding.cu
davidwendt: Yes, I forgot to remove this as well. Thanks.

cpp/src/text/bpe/byte_pair_encoding.cu
davidwendt: The algorithm is a challenge. Perhaps I can add examples into the comments.

bdice: Great. Yes, reverse find can be awkward to write. If you know the number of iterations will be small (around 5), then don't worry about using a smarter search algorithm.

cpp/src/text/bpe/byte_pair_encoding.cu
davidwendt: Except `bytes` is `int64_t` and I don't think we should cast both sides to `size_t`

bdice: Ok, feel free to resolve this.

cpp/src/text/bpe/byte_pair_encoding.cu
```diff
@@ -21,477 +21,305 @@
...
- * @brief Resolve a substring up to the first whitespace character.
+ * @brief Produces offsets to unpairable locations in the given chars array
  *
- * This will return a substring of the input starting with the first byte
- * up to the first whitespace character found or the end of the string.
- * Any whitespace is expected only at the end of the string.
+ * Launched as a thread per byte of the chars array.
+ * The output is non-zero offsets to locations of unpairable substrings.
+ * An unpairable substring does not exist in the given map and so will
+ * never be paired :-(. Fortunately, this can be used as an artificial
```
bdice, line 57: The sad face is funny but it has caught my attention twice while reviewing with the question of "what is this docstring syntax?" Maybe better to remove it since we have a more direct tone in most docstrings.
```suggestion
 * never be paired. Fortunately, this can be used as an artificial
```

cpp/src/text/bpe/byte_pair_encoding.cu
```diff
@@ -21,477 +21,305 @@
...
+
+  auto min_rank = max_rank;
+
+  // store all the initial ranks for each pair
+  // every character but the first and last one will have a pair and a rank
+  //
+  // Example:
+  // string:   abcdefghij
+  // spaces:   1111111111
+  // ranks:    *948516327*
```
bdice, line 179: I'm a bit confused. The first and last characters do not have a rank, if I understand this comment correctly. I believe that's what you're trying to indicate with the `*` meaning "no rank" for the character `a`. But why is there no character/space data above the second `*`? I expected to see this:
```suggestion
  // every character but the first and last one will have a pair and a rank
  //
  // Example:
  // string:   abcdefghijk
  // spaces:   11111111111
  // ranks:    *948516327*
```

cpp/src/text/bpe/byte_pair_encoding.cu
```diff
@@ -21,477 +21,305 @@
...
-      thrust::lower_bound(thrust::seq, d_offsets, d_offsets + d_strings.size() + 1, offset);
-    // ignore offsets at existing string boundaries
-    if (*itr == offset) { return false; }
-
-    // count only edges for valid strings
-    auto const index = static_cast<cudf::size_type>(thrust::distance(d_offsets, itr)) - 1;
-    return d_strings.is_valid(index);
+  // compute the output size for this string by counting the resulting separator positions
+  auto bytes = 0;
+  for (auto itr = d_spaces + lane_idx; itr < end_spaces; itr += block_size) {
```
bdice, line 316: This seems like it should be a `thrust::count_if`? Is the block stride the reason why that wasn't chosen?

cpp/src/text/bpe/byte_pair_encoding.cu
```diff
@@ -500,54 +328,120 @@ std::unique_ptr<cudf::column> byte_pair_encoding(cudf::strings_column_view const
...
+  // build the output: inserting separators to the input character data
+  auto chars   = cudf::strings::detail::create_chars_child_column(bytes, stream, mr);
+  auto d_chars = chars->mutable_view().data<char>();
+
+  auto const d_inserts     = d_working.data();  // stores the insert positions
+  auto offsets_at_non_zero = [d_spaces = d_spaces.data()] __device__(auto idx) {
+    return d_spaces[idx] > 0;  // separator to be inserted here
+  };
+  auto const copy_end = thrust::copy_if(
+    rmm::exec_policy(stream), chars_begin + 1, chars_end, d_inserts, offsets_at_non_zero);
```
bdice, line 426: Should we try to use `exec_policy_nosync` stream policies throughout this code?

cpp/src/text/bpe/byte_pair_encoding.cu
davidwendt: Right, a `thrust::count_if` here would perform terribly here.

cpp/src/text/bpe/byte_pair_encoding.cu
davidwendt: The trailing '*' was a mistake. I will correct this.

bdice (approved):

cpp/src/text/bpe/load_merge_pairs.cu
```diff
@@ -135,6 +154,15 @@ std::unique_ptr<bpe_merge_pairs> load_merge_pairs_file(std::string const& filena
   return std::make_unique<bpe_merge_pairs>(std::move(input_column), stream, mr);
 }
 
+std::unique_ptr<bpe_merge_pairs> load_merge_pairs(cudf::strings_column_view const& merge_pairs,
+                                                  rmm::cuda_stream_view stream,
+                                                  rmm::mr::device_memory_resource* mr)
+{
+  CUDF_EXPECTS(!merge_pairs.is_empty(), "Merge pairs must not be empty");
+  CUDF_EXPECTS(!merge_pairs.has_nulls(), "Merge pairs may not contain nulls");
```
PointKernel, line 162: Just to make sure, are docs updated after adding those checks?

cpp/src/text/bpe/byte_pair_encoding.cuh
```diff
@@ -30,85 +31,165 @@
...
+  }
+  // used by find
+  __device__ bool operator()(cudf::size_type lhs, cudf::string_view const& rhs) const noexcept
+  {
+    auto const left = d_strings.element<cudf::string_view>(lhs);
+    return left == rhs;
+  }
+};
+
+using mp_probe_scheme = cuco::experimental::linear_probing<1, mp_hasher>;
```
PointKernel, line 161: nit: this is the same as `mp_probe_scheme`. maybe unify them?

PointKernel (approved):
Several non-blocking questions. Otherwise LGTM

cpp/include/nvtext/byte_pair_encoding.hpp
```diff
@@ -36,7 +36,7 @@ namespace nvtext {
  */
 struct bpe_merge_pairs {
   struct bpe_merge_pairs_impl;
-  std::unique_ptr<bpe_merge_pairs_impl> impl{};  ///< Implementation of the BPE merge pairs table.
+  bpe_merge_pairs_impl* impl{};  ///< Implementation of the BPE merge pairs table.
```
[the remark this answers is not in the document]
davidwendt: This is to setup for integrating with Cython in the next PR.
It was technically a raw pointer previously -- just wrapped in a unique-ptr.

cpp/src/text/bpe/load_merge_pairs.cu
davidwendt: Yes, the are documented here: https://github.com/rapidsai/cudf/blob/1640c16898da96880e31eb1a5233091b18ee35a3/cpp/include/nvtext/byte_pair_encoding.hpp#L122

cpp/src/text/bpe/byte_pair_encoding.cuh
davidwendt: Do you mean `bpe_probe_scheme`? 
One uses `mp_hasher` and the other uses `bpe_hasher` which are very different.
Maybe I'm missing something.

PointKernel: Ah, you are right, I thought that they were both linear probing with CG size = 1 and didn't pay attention to the hasher part.

PR #14330, opened 2023-10-25
Upgrade `arrow` to `13`

galipremsagar:
## Description
This PR upgrades `arrow` to `13`.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #14339, opened 2023-10-27
Upgrade wheels to use arrow 13

vyasr:
## Description

In #14330 we upgraded to Arrow 13. However, we only did so for conda packages. For wheels, we couldn't do the same because pyarrow 13 started supporting two manylinux versions, both 2.17 and 2.28. This results in ABI compatibility issues because cudf wheels bundle an identical libarrow and was previously configured to compile with the CXX11 ABI turned off for compatibility with the  libarrow in the 2.17 wheels. To address this, this PR modifies the CMake logic to only set that flag conditionally based on the glibc version on the host system.

## Checklist


vyasr:
Before we merge this, we need https://github.com/rapidsai/ci-imgs/pull/87 merged and we need to make a corresponding PR to https://github.com/rapidsai/shared-action-workflows/ that enables building 2.28 wheels in addition to 2.17 wheels. We then need to update the workflows in this repo to trigger the extra 2.28 build.

vyasr:
https://github.com/rapidsai/ci-imgs/pull/87 and https://github.com/rapidsai/ci-imgs/pull/88 are merged. We should be able to merge this once https://github.com/rapidsai/shared-workflows/pull/150 is merged (and once https://github.com/rapidsai/cudf/pull/14344 is merged to unblock cudf CI).

galipremsagar (approved):

robertmaynard (approved):

ci/test_wheel_cudf.sh
```diff
@@ -3,8 +3,20 @@
 
 set -eou pipefail
 
+# Set the manylinux version to pull the newer ABI wheels when testing on Ubuntu
+# 20.04 or later. Note that this logic is specific to the current set of
+# citestwheel images and would need to change if the images change.
+manylinux="manylinux_2_17"
+if command -v lsb_release >/dev/null 2>&1 ; then
```
bdice, line 10: As before, I think we can derive this programmatically by parsing output of `ldd --version` or some other means.

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -62,10 +62,26 @@ function(find_libarrow_in_python_wheel PYARROW_VERSION)
   # to builds of GTest. Similarly, benchmarks will not work without updating GBench (and possibly
```
bdice, line 62: Can we update the comment above this? It states "pyarrow's `libarrow.so` is compiled for manylinux2014" which is now only true for a particular build.

dependencies.yaml
```diff
@@ -279,7 +279,7 @@ dependencies:
       - output_types: [requirements, pyproject]
         packages:
           # Allow runtime version to float up to minor version
-          - pyarrow==12.*
+          - pyarrow==13.*
```
bdice, line 282: Merge this with the `conda` dependency list above. Look at the previous PR to see how this was split up, and merge it back together by reverting those changes to the structure.

dependencies.yaml
```diff
@@ -251,7 +251,7 @@ dependencies:
         packages:
           # Hard pin the patch version used during the build. This must be kept
           # in sync with the version pinned in get_arrow.cmake.
-          - pyarrow==12.0.1.*
+          - pyarrow==13.0.0.*
```
bdice, line 254: Merge this with the `conda` dependency list above. Look at the previous PR to see how this was split up, and merge it back together by reverting those changes to the structure.

raydouglass (approved):

ci/build_wheel_cudf.sh
```diff
@@ -9,8 +9,12 @@ export SKBUILD_CONFIGURE_OPTIONS="-DCUDF_BUILD_WHEELS=ON -DDETECT_CONDA_ENV=OFF"
 
 ./ci/build_wheel.sh cudf ${package_dir}
 
+manylinux="manylinux_2_17"
 mkdir -p ${package_dir}/final_dist
+if command -v dnf >/dev/null 2>&1 ; then
```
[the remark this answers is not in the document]
raydouglass: I agree. In hindsight, this should be updated.

raydouglass (changes requested):

ci/build_wheel_cudf.sh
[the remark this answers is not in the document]
vyasr: Done.

cpp/cmake/thirdparty/get_arrow.cmake
vyasr: Yup.

dependencies.yaml
vyasr: Done.

bdice (changes requested):

ci/build_wheel_cudf.sh
```diff
@@ -9,8 +9,9 @@ export SKBUILD_CONFIGURE_OPTIONS="-DCUDF_BUILD_WHEELS=ON -DDETECT_CONDA_ENV=OFF"
 
 ./ci/build_wheel.sh cudf ${package_dir}
 
-mkdir -p ${package_dir}/final_dist
+# Set the manylinux version used for uploading the wheels
+manylinux="manylinux_$(ldd --version | head -1 | grep -o "[0-9]\.[0-9]\+" | sed 's/\./_/g')"
```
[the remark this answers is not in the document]
vyasr: Actually for the build script this is even simpler. We can just use [`AUDITWHEEL_POLICY`, which is always set in these images](https://github.com/rapidsai/ci-imgs/blob/main/ci-wheel.Dockerfile#L120).

ci/test_wheel_cudf.sh
```diff
@@ -3,8 +3,11 @@
 
 set -eou pipefail
 
+# Set the manylinux version used for downloading the wheels
+manylinux="manylinux_$(ldd --version | head -1 | grep -o "[0-9]\.[0-9]\+" | sed 's/\./_/g')"
```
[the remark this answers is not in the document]
vyasr: Updated.

ci/build_wheel_cudf.sh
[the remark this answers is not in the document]
bdice: Brilliant!

bdice (approved):

raydouglass (approved):

PR #14341, opened 2023-10-27
Remove unnecessary section

vyasr:
## Description

Currently all the group pages in the documentation contained a "Detailed Description" section, but that section is always empty because we don't include additional information on each group. Removing that section should help clean things up.

Alternatively, we could write up a longer description for each group.

## Checklist

### DIFF
--- a/cpp/doxygen/DoxygenLayout.xml
+++ b/cpp/doxygen/DoxygenLayout.xml
@@ -191,7 +191,6 @@
       <friends title=""/>
       <membergroups visible="yes"/>
     </memberdecl>
-    <detaileddescription title=""/>
     <memberdef>
       <pagedocs/>
       <inlineclasses title=""/>


vyasr:
Here's an example of the empty section in the hashing docs:
image

If we want to instead fill it in properly, we would need to add descriptions below each `defgroup` in https://github.com/rapidsai/cudf/blob/branch-23.12/cpp/include/doxygen_groups.h.

vyasr:
This came up in https://github.com/rapidsai/rmm/pull/1324. If we choose to fill out the docs for each group then we may want to reorder the layout so the summary comes above the ToC of members of the group.

vyasr:
At least in rmm we've decided to try and fill out the documentation for each group, so I'm closing this for now. If we go in the opposite direction for libcudf we can easily reopen this PR.

davidwendt:
> At least in rmm we've decided to try and fill out the documentation for each group, so I'm closing this for now. If we go in the opposite direction for libcudf we can easily reopen this PR.

I will take a stab at documenting the sections.

PR #14342, opened 2023-10-28
Expose streams in public unary APIs

vyasr:
## Description

Contributes to #925 

## Checklist


cpp/include/cudf/unary.hpp
```diff
@@ -65,41 +66,47 @@ enum class unary_operator : int32_t {
  *
  * @param input A `column_view` as input
  * @param op operation to perform
+ * @param stream CUDA stream used for device memory operations and kernel launches.
```
ttnghia, line 69: The period is not removed after copying :)
(not just here but all around this file)

harrism (approved):

cpp/include/cudf/unary.hpp
vyasr: Was that something that we wanted to do? I can do it, I didn't remember.

ttnghia (approved):

PR #14345, opened 2023-10-30
Fix overflow check in `cudf::merge`

divyegala:
## Description
https://github.com/rapidsai/cudf/pull/14250 added a check to ensure `cudf::merge` throws when the total number of merged rows exceed `cudf::size_type` limit, however @bdice pointed out that the check was not correct because the accumulation was still occurring in `cudf::size_type`. This PR computes the accumulation in `std::size_t`.

## Checklist


bdice (approved):

mythrocks (approved):

ttnghia (approved):
The issue keeps showing up many times and this is just fixing the leaf, not the root of the problem. Probably we should consider a better approach: changing `column(_view)::size()` to return `int64_t` instead.

PR #14348, opened 2023-10-31
cudf jit parser now supports .pragma instructions with quotes

robertmaynard:
## Description
During testing of CUDA 12.3 it was found that python user defined functions could generate .pragma instructions that have quotes, so as "nounroll" and "used_bytes_mask N".

The jit parser now makes sure to escape quotes that are part of a `.pragma` instruction so that it can be properly passed to nvrtc/jitify as a CUDA header.

## Checklist


cpp/tests/CMakeLists.txt
```diff
@@ -617,6 +617,11 @@ ConfigureTest(
 # * bin tests ----------------------------------------------------------------------------------
 ConfigureTest(LABEL_BINS_TEST labeling/label_bins_tests.cpp)
 
+# ##################################################################################################
+# * jit tests ----------------------------------------------------------------------------------
+ConfigureTest(JIT_PARSER_TEST jit/parse_ptx_function.cpp)
+target_include_directories(JIT_PARSER_TEST PRIVATE "$<BUILD_INTERFACE:${CUDF_SOURCE_DIR}/src>")
```
vyasr, line 623: What do we need this include for? What header is getting pulled?

vyasr (approved):
One question about the CMake, otherwise looks good. I'm curious, how did you generate the tests? I assume you didn't write the PTX from scratch. I'm wondering if it would be helpful to include a compilation line in the test file demonstrating how to generate that PTX to make it easier to add future tests if needed.

PointKernel (approved):

cpp/tests/CMakeLists.txt
robertmaynard: The `<jit/parser.hpp>` file includes other headers that are private to cudf. So I needed this include to test the JIT parser in isolation.

robertmaynard:
> One question about the CMake, otherwise looks good. I'm curious, how did you generate the tests? I assume you didn't write the PTX from scratch. I'm wondering if it would be helpful to include a compilation line in the test file demonstrating how to generate that PTX to make it easier to add future tests if needed.


I generated the PTX by dumping existing PTX code from libcudf with `cuobjdump`. From that I selected subsection of existing functions that had relevant bits to test ( differerent pragmas ) and merged it with an empty function signature. Since we only need to parse the PTX I wasn't worried about the correctness of the function, only that it contained valid instructions.

PR #14350, opened 2023-11-01
Expose streams in ORC reader and writer APIs

shrshi:
## Description

This PR contributes to https://github.com/rapidsai/cudf/issues/13744.
-Added stream parameters to public APIs
```
cudf::io::read_orc
cudf::io::write_orc
cudf::io::read_orc_metadata
cudf::io::read_parsed_orc_statistics
```
-Added stream gtests

## Checklist


cpp/tests/streams/io/orc_test.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <string>
+#include <vector>
+
+auto const temp_env = static_cast<cudf::test::TempDirTestEnvironment*>(
+  ::testing::AddGlobalTestEnvironment(new cudf::test::TempDirTestEnvironment));
+
+class ORCTest : public cudf::test::BaseFixture {};
+
+template <typename T>
+inline auto random_values(size_t size)
```
davidwendt, line 42: Generating random values seems overkill for stream testing since the values themselves will not influence how the stream is passed through and the results are not checked.
Perhaps just hardcode a set of values instead?

ttnghia: Even with random generated input, this is not what we are doing. We already have `generate_input_tables` for on-GPU random data generation.

vuule: +1
Assuming that the goal is to call all kernels that ORC reader/writer ever call, we can just pass all zeroes. 
Another benefit from non-random values: strings column with all equal, non-empty strings would exercise the dictionary encode/decode path, increasing the test code coverage.

shrshi: Thank you for the feedback! I've modified the tests to take in constant columns, and that gave rise to stream pool errors in the dictionary encoder calls. Missed this case in the previous tests!

cpp/tests/streams/io/orc_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+    return numeric::decimal128{ones[i], numeric::scale_type{-12}};
+  });
+
+  cudf::test::fixed_width_column_wrapper<bool> col0(zeros.begin(), zeros.end());
+  cudf::test::fixed_width_column_wrapper<int8_t> col1(zeros.begin(), zeros.end());
+  cudf::test::fixed_width_column_wrapper<int16_t> col2(zeros.begin(), zeros.end());
+  cudf::test::fixed_width_column_wrapper<int32_t> col3(zeros.begin(), zeros.end());
+  cudf::test::fixed_width_column_wrapper<float> col4(zeros.begin(), zeros.end());
+  cudf::test::fixed_width_column_wrapper<double> col5(zeros.begin(), zeros.end());
+  cudf::test::fixed_width_column_wrapper<numeric::decimal128> col6(col6_data, col6_data + num_rows);
```
ttnghia, line 60: If a variable is temporary which is used only to initialize other variable, put it into a lambda.
```suggestion
  cudf::test::fixed_width_column_wrapper<numeric::decimal128> col6 = [] {
     auto col6_data = cudf::detail::make_counting_transform_iterator(0, [&](auto i) {
    return numeric::decimal128{ones[i], numeric::scale_type{12}};
  });
      return cudf::test::fixed_width_column_wrapper<numeric::decimal128>(col6_data, col6_data + num_rows);
  }();
```

ttnghia: Similar applies for `col7`.

cpp/tests/streams/io/orc_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+  tab_metadata.column_metadata[1].set_name("int8s");
+  tab_metadata.column_metadata[2].set_name("int16s");
+  tab_metadata.column_metadata[3].set_name("int32s");
+  tab_metadata.column_metadata[4].set_name("floats");
+  tab_metadata.column_metadata[5].set_name("doubles");
+  tab_metadata.column_metadata[6].set_name("decimal_pos_scale");
+  tab_metadata.column_metadata[7].set_name("decimal_neg_scale");
+  tab_metadata.column_metadata[8].set_name("lists");
+  tab_metadata.column_metadata[9].set_name("structs");
+  tab_metadata.column_metadata[10].set_name("strings");
```
ttnghia, line 85: This is not needed for stream test, isn't it?

cpp/tests/streams/io/orc_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+  cudf::test::lists_column_wrapper<int64_t> col8{
+    {1, 1}, {1, 1, 1}, {}, {1}, {1, 1, 1, 1}, {1, 1, 1, 1, 1}, {}, {1, -1}, {}, {-1, -1}};
+
+  cudf::test::fixed_width_column_wrapper<int32_t> child_col(ones.begin(), ones.end());
+  cudf::test::structs_column_wrapper col9{child_col};
+
+  std::vector<std::string> col10_data(num_rows, "rapids");
+  cudf::test::strings_column_wrapper col10(col10_data.begin(), col10_data.end());
+
+  cudf::table_view tab({col0, col1, col2, col3, col4, col5, col6, col7, col8, col9, col10});
```
ttnghia, line 125: This is repeated exactly as in the previous test. How about extracting it into a free function?

cpp/tests/utilities/identify_stream_usage.cpp
```diff
@@ -72,7 +73,8 @@ bool stream_is_invalid(cudaStream_t stream)
 {
 #ifdef STREAM_MODE_TESTING
   // In this mode the _only_ valid stream is the one returned by cudf::test::get_default_stream.
-  return (stream != cudf::test::get_default_stream().value());
+  return (stream == cudf::get_default_stream().value() || stream == cudaStreamDefault ||
+          stream == cudaStreamLegacy || stream == cudaStreamPerThread);
```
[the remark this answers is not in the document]
shrshi: This change was originally added to handle the test stream in stream pool. Now that testing stream pool has been implemented ([PR 14437](https://github.com/rapidsai/cudf/pull/14437)), we can get rid of this relaxed check.

cpp/tests/streams/io/orc_test.cpp
shrshi: Thanks, addressed!

cpp/tests/streams/io/orc_test.cpp
shrshi: No, we don't need metadata for stream tests. I've removed it.

cpp/tests/streams/io/orc_test.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+template <typename... UniqPtrs>
+std::vector<std::unique_ptr<cudf::column>> make_uniqueptrs_vector(UniqPtrs&&... uniqptrs)
+{
+  std::vector<std::unique_ptr<cudf::column>> ptrsvec;
+  (ptrsvec.push_back(std::forward<UniqPtrs>(uniqptrs)), ...);
+  return ptrsvec;
+}
+
+cudf::table construct_table()
```
ttnghia, line 49: Typically we put `table`/`column` into a `unique_ptr` so we can have resource managed.

shrshi: Since the function returns a temporary object, wouldn't the compiler perform RVO? The table would not be unnecessarily copied and `tab` object in each of the tests will be owning.

vuule: RVO definitely kicks in here, but having a `unique_ptr` makes it harder to accidentally copy the table once it's created. IMO it's not required here (test code), but it wouldn't hurt.

cpp/tests/streams/io/orc_test.cpp
```diff
@@ -0,0 +1,127 @@
...
+  }();
+  cudf::test::fixed_width_column_wrapper<numeric::decimal128> col7 = [&ones, num_rows] {
+    auto col7_data = cudf::detail::make_counting_transform_iterator(0, [&](auto i) {
+      return numeric::decimal128{ones[i], numeric::scale_type{-12}};
+    });
+    return cudf::test::fixed_width_column_wrapper<numeric::decimal128>(col7_data,
+                                                                       col7_data + num_rows);
+  }();
+
+  cudf::test::lists_column_wrapper<int64_t> col8{
```
vuule, line 77: Just one thing in terms of coverage - reader does some additional work when list/struct columns have nulls. The parent null mask is pushed down to make sure child ORC rows are read to correct output rows.
So having some nulls in the list and struct columns would complete the coverage, AFAICT.

vuule (changes requested):
Just one coverage improvement requested, looks good otherwise.

cpp/tests/streams/io/orc_test.cpp
shrshi: Good idea, I've added the validity mask for the columns. Thanks!

vuule (approved):
Thank you for addressing the suggestions, looks great!

cpp/tests/streams/io/orc_test.cpp
```diff
@@ -0,0 +1,135 @@
...
+  (ptrsvec.push_back(std::forward<UniqPtrs>(uniqptrs)), ...);
+  return ptrsvec;
+}
+
+cudf::table construct_table()
+{
+  constexpr auto num_rows = 10;
+
+  std::vector<size_t> zeros(num_rows, 0);
+  std::vector<size_t> ones(num_rows, 1);
```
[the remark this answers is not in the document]
shrshi: TIL, thanks for the suggestion!

ttnghia (approved):

Issue #11463, opened 2022-08-04
[FEA] Add `stream` parameter for groupby operations (aggregate, scan, shift, etc.)


ttnghia:
Currently, groupby operations do not have a `stream` parameter. As such, we can't pass in the current `stream` value but the operations will use the default stream instead.

We should allow passing in `stream` parameter in groupby aggregation.

davidwendt:
Closed by https://github.com/rapidsai/cudf/pull/17324

PR #14354, opened 2023-11-02
Replace default stream for scalars and column factories usages (because of defaulted arguments)

karthikeyann:
## Description
This PR contributes to https://github.com/rapidsai/cudf/issues/13744

Adds missing stream for scalars, column factories
Uses right `set_null_mask` for moving null masks instead of copy. (due to defaulted stream argument).

## Checklist


cpp/src/groupby/sort/aggregate.cpp
```diff
@@ -571,9 +571,11 @@ void aggregate_result_functor::operator()<aggregation::MERGE_HISTOGRAM>(aggregat
  * @param column_1 The second column
  * @return tuple with new null mask (if null masks of input differ) and new column views
  */
-auto column_view_with_common_nulls(column_view const& column_0, column_view const& column_1)
+auto column_view_with_common_nulls(column_view const& column_0,
```
ttnghia, line 574: Need to update docs above this line.

cpp/src/io/csv/writer_impl.cu
```diff
@@ -367,10 +371,10 @@ void write_chunked(data_sink* out_sink,
 
   CUDF_EXPECTS(str_column_view.size() > 0, "Unexpected empty strings column.");
 
-  cudf::string_scalar newline{options.get_line_terminator()};
+  cudf::string_scalar newline{options.get_line_terminator(), true, stream};
   auto p_str_col_w_nl = cudf::strings::detail::join_strings(str_column_view,
                                                             newline,
-                                                            string_scalar("", false),
+                                                            string_scalar("", false, stream),
```
ttnghia, line 377: `newline` is defined separately while the 3rd parameter is defined inline. Can you make them consistent please?

cpp/src/io/csv/writer_impl.cu
```diff
@@ -472,14 +476,15 @@ void write_csv(data_sink* out_sink,
...
-                                                    delimiter_str,
-                                                    options.get_na_rep(),
-                                                    strings::separator_on_nulls::YES,
-                                                    stream,
-                                                    rmm::mr::get_current_device_resource());
-        cudf::string_scalar narep{options.get_na_rep()};
+          return cudf::strings::detail::concatenate(
+            str_table_view,
+            cudf::string_scalar(delimiter_str, true, stream),
+            narep,
```
ttnghia, line 484: One parameter is inline defined while the other is not. Please make them consistent.

karthikeyann: I don't think, it's not a matter of being consistent. That temporary scalar is not required after this usage. So, it's not a named parameter. its scope end after function call.
Here narep is used below this line.

cpp/src/io/csv/writer_impl.cu
karthikeyann: I don't think, it's not a matter of being consistent. That temporary scalar is not required after this usage. So, it's not a named parameter. its scope end after function call.
newline is used below.

cpp/src/groupby/sort/aggregate.cpp
karthikeyann: updated.

bdice (commented):
I started a review earlier and didn't submit it -- sorry. Flushing comments for now, and I'll try to get back to completing the review soon.

cpp/src/groupby/groupby.cu
```diff
@@ -107,16 +107,21 @@ struct empty_column_constructor {
...
 
     if constexpr (k == aggregation::Kind::COLLECT_LIST || k == aggregation::Kind::COLLECT_SET) {
-      return make_lists_column(
-        0, make_empty_column(type_to_id<size_type>()), empty_like(values), 0, {});
+      return make_lists_column(0,
+                               make_empty_column(type_to_id<size_type>()),
+                               empty_like(values),
+                               0,
+                               {},
+                               cudf::get_default_stream());
```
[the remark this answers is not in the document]
vuule: is the stream relevant in this case? Is any allocation actually happening when creating an empty column?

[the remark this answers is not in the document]
ttnghia: I've filed an issue a while ago for it: https://github.com/rapidsai/cudf/issues/11463.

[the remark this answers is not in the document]
ttnghia: So before that issue is closed, we don't have any input stream to put into line 115 above and have to use the default stream anyway.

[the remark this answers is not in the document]
karthikeyann: groupby needs a bigger cleanup; I can exclude it as part of this PR. It should be a separate PR.

cpp/src/io/csv/writer_impl.cu
```diff
@@ -161,7 +161,11 @@ struct column_to_strings_fn {
     column_view const& column) const
```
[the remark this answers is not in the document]
karthikeyann: The 2nd note can be removed. stream is exposed now. so it's not true anymore.

cpp/src/groupby/groupby.cu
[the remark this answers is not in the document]
karthikeyann: Removed groupby from this PR.

cpp/src/lists/dremel.cu
```diff
@@ -127,7 +127,8 @@ dremel_data get_encoding(column_view h_col,
   }
   std::unique_ptr<column> empty_list_offset_col;
   if (has_empty_list_offsets) {
-    empty_list_offset_col = make_fixed_width_column(data_type(type_id::INT32), 1);
+    empty_list_offset_col =
+      make_fixed_width_column(data_type(type_id::INT32), 1, mask_state::UNALLOCATED, stream);
```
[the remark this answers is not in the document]
karthikeyann: done.

cpp/src/unary/cast_ops.cu
```diff
@@ -194,7 +194,7 @@ std::unique_ptr<column> rescale(column_view input,
       auto const scalar  = make_fixed_point_scalar<T>(0, scale_type{scale}, stream);
       auto output_column = make_column_from_scalar(*scalar, input.size(), stream, mr);
       if (input.nullable()) {
-        auto const null_mask = detail::copy_bitmask(input, stream, mr);
+        auto null_mask = detail::copy_bitmask(input, stream, mr);
         output_column->set_null_mask(std::move(null_mask), input.null_count());
```
vuule, line 198: oof. I thought this woudn't compile. This seems pretty hard to find in general.

cpp/src/table/table.cpp
```diff
@@ -29,7 +29,7 @@ table::table(table const& other) : _num_rows{other.num_rows()}
   CUDF_FUNC_RANGE();
   _columns.reserve(other._columns.size());
   for (auto const& c : other._columns) {
-    _columns.emplace_back(std::make_unique<column>(*c));
+    _columns.emplace_back(std::make_unique<column>(*c, cudf::get_default_stream()));
```
vuule, line 32: Is this constructor ever used internally?

cpp/src/scalar/scalar.cpp
```diff
@@ -421,7 +421,7 @@ duration_scalar<T>::duration_scalar(duration_scalar<T> const& other,
 template <typename T>
 typename duration_scalar<T>::rep_type duration_scalar<T>::count()
 {
-  return this->value().count();
+  return this->value(cudf::get_default_stream()).count();
```
vuule, line 424: are we planning to add the stream parameter to these at some point?

vuule (approved):
Looks good.
Posted a few follow-up questions.

bdice (changes requested):
We should not add new calls to `cudf::get_default_stream()` in internal implementations. @vuule just raised this here: https://github.com/rapidsai/cudf/pull/14354/files#r1393097787

It's the same problem as mentioned here: https://github.com/rapidsai/cudf/pull/14354#discussion_r1380645497

vuule:
> We should not add new calls to `cudf::get_default_stream()` in internal implementations. @vuule just raised this here: https://github.com/rapidsai/cudf/pull/14354/files#r1393097787
> 
> It's the same problem as mentioned here: [#14354 (comment)](https://github.com/rapidsai/cudf/pull/14354#discussion_r1380645497)
> 
> Is there a way we can fix this in order, so that those calls have the stream passed through by their caller?

I'm okay with explicitly passing the default stream in cases where we don't have the user stream available (yet). It makes the default stream use very visible so we'll find these places easily once we get to updating those APIs to take a stream. IMO this is a good change, even though it does not really change the stream.

vyasr:
I would recommend just adding a stream parameter with a default value and then forwarding that to the column constructor. Is there any reason we can't do that? It shouldn't break any existing APIs. While we're at it, we probably want to add a default mr parameter too.

cpp/src/unary/cast_ops.cu
ttnghia: `std::move` doesn't move anything. It just tries to convert the parameter into `rvalue`. So for this case it failed to convert and the nul mask was just copied.

karthikeyann:
As @vuule mentioned, default stream is added to make its use visible and hence allows us to replace it in future PRs.
@vyasr Adding default arguments (stream/mr) will not break existing APIs. I am also concerned about their future usage ignoring these defaults, which will again need to be fixed by similar cleanup-PRs (if they are not caught by existing stream tests). Is there any alternative or a tradeoff? Ideally, we should prefer to make it hard to use the APIs wrongly, and easy to use it in intended way.

I thought about `explicit` keyword for copy constructor of scalar (I think, it's similar for column as well). I feared if it might cause our code be too verbose. But I compiled libcudf with `explicit` for scalar; only scalar tests need to be updated. Also, this might be a breaking change, as users need to update their code to call the copy constructor explicitly. 

Moving this PR for 24.02.

vyasr:
I don't think we have a general solution for this problem for any APIs (not just copy constructors) other than improving the coverage of our stream tests (which in the long run we want to do by modifying all existing tests to be stream tests, but that's not going to happen until all our APIs support streams). I agree that ideally we would make it as hard as possible to use these APIs the wrong way, but as long as public APIs are going to have a default stream parameter I don't see any alternative to adding stream tests to ensure that we don't accidentally rely on that default internally. 

Copy constructors for all of our types are a bit trickier than other APIs, mainly because 1) we can't have a detail equivalent like we do normally, and 2) we need to prevent implicit invocation. I am in favor of making all of our copy constructors for owning types explicit in a future release to prevent implicit invocation. If we do that, then the copy constructor problem reduces to the same problem that we have for all other APIs, which is that explicitly incorrect invocations relying on the default stream parameter will only be caught if that code path is covered in a stream test. I think that's the best that we can do for now but am open to suggestions if anyone has other ideas.

cpp/src/io/json/write_json.cu
vuule, line 1: Addressed in https://github.com/rapidsai/cudf/pull/14444
Apologies, forgot to check this PR before opening mine.
FWIW, my PR also avoid redundant creation of these string_scalars

cpp/src/io/csv/writer_impl.cu
```diff
@@ -161,7 +161,11 @@ struct column_to_strings_fn {
...
   {
     return cudf::strings::detail::from_booleans(
-      column, options_.get_true_value(), options_.get_false_value(), stream_, mr_);
+      column,
+      cudf::string_scalar(options_.get_true_value(), true, stream_),
+      cudf::string_scalar(options_.get_false_value(), true, stream_),
```
vuule: Would you mind making a change similar to the one in https://github.com/rapidsai/cudf/pull/14444?

cpp/src/io/csv/reader_impl.cu
```diff
@@ -868,8 +868,8 @@ table_with_metadata read_csv(cudf::io::datasource* source,
         // quotechars in quoted fields results in reduction to a single quotechar
         // TODO: Would be much more efficient to perform this operation in-place
         // during the conversion stage
-        std::string const quotechar(1, parse_opts.quotechar);
-        std::string const dblquotechar(2, parse_opts.quotechar);
+        cudf::string_scalar const quotechar(std::string(1, parse_opts.quotechar), true, stream);
+        cudf::string_scalar const dblquotechar(std::string(2, parse_opts.quotechar), true, stream);
```
vuule, line 872: I opened https://github.com/rapidsai/cudf/pull/14443 for this one, let me know how you want to proceed.

wence-:
None of the Python changes look like they belong in this PR. Is it because you were previously targeted to 23.12 and have some changes from there that are not yet in 24.02 due to the forward-merge no coming across yet?

cpp/src/unary/cast_ops.cu
karthikeyann: in this case, it called the overloaded function `set_null_mask` with `lvaue` rmm::buffer, and defaulted stream. So inside that `set_null_mask`, it copied the buffer.

cpp/src/table/table.cpp
karthikeyann: Good catch. `table(table)` constructor is not internally used. 
It's only used in `cpp/benchmarks/stream_compaction/unique.cpp` (which could be replaced too).

Should we consider removing it?

cpp/src/io/csv/reader_impl.cu
karthikeyann: https://github.com/rapidsai/cudf/pull/14443 looks good, and can be merged.

karthikeyann:
@wence- merge done. python changes are gone.

cpp/src/io/csv/reader_impl.cu
vuule: I pulled the `string_scalar` out of the loop so we don't create them for each column.

wence-:
> @wence- pulled 24.02. python changes are gone.

Thanks! Removed the now unnecessary codeowner review requests.

cpp/src/strings/convert/convert_floats.cu
```diff
@@ -97,7 +97,8 @@ std::unique_ptr<column> to_floats(strings_column_view const& input,
                                   rmm::mr::device_memory_resource* mr)
 {
   size_type strings_count = input.size();
-  if (strings_count == 0) return make_numeric_column(output_type, 0);
+  if (strings_count == 0)
+    return make_numeric_column(output_type, 0, mask_state::UNALLOCATED, stream);
```
bdice, line 101:
```suggestion
  if (strings_count == 0) {
    return make_numeric_column(output_type, 0, mask_state::UNALLOCATED, stream);
  }
```

cpp/src/strings/convert/convert_integers.cu
```diff
@@ -274,7 +274,8 @@ std::unique_ptr<column> to_integers(strings_column_view const& input,
                                     rmm::mr::device_memory_resource* mr)
 {
   size_type strings_count = input.size();
-  if (strings_count == 0) return make_numeric_column(output_type, 0);
+  if (strings_count == 0)
+    return make_numeric_column(output_type, 0, mask_state::UNALLOCATED, stream);
```
bdice, line 278:
```suggestion
  if (strings_count == 0) {
    return make_numeric_column(output_type, 0, mask_state::UNALLOCATED, stream);
  }
```

cpp/src/strings/convert/convert_ipv4.cu
```diff
@@ -77,7 +77,8 @@ std::unique_ptr<column> ipv4_to_integers(strings_column_view const& input,
                                          rmm::mr::device_memory_resource* mr)
 {
   size_type strings_count = input.size();
-  if (strings_count == 0) return make_numeric_column(data_type{type_id::INT64}, 0);
+  if (strings_count == 0)
+    return make_numeric_column(data_type{type_id::INT64}, 0, mask_state::UNALLOCATED, stream);
```
bdice, line 81:
```suggestion
  if (strings_count == 0) {
    return make_numeric_column(data_type{type_id::INT64}, 0, mask_state::UNALLOCATED, stream);
  }
```

cpp/src/table/table.cpp
bdice: I think I want to see similarities between the design of columns and tables, so that probably leans towards keeping this. However, I could be swayed if there is an explicit API (free function?) that allows for table or column copying. If we decide to remove, I think we'd want an issue documenting this and to deprecate it before removing.

vuule: My concern with having copy constructors for tables/columns is that we risk accidental copies as they can happen implicitly. On the other hand, we usually access these through unique_ptr, which would prevent copies.
I wonder if marking the copy ctors as explicit suffices here.

karthikeyann: added explicity and fixed the copies in benchmark

ttnghia: Should we add a `stream` parameter for this constructor? Always using the default stream may be an issue.

cpp/src/scalar/scalar.cpp
```diff
@@ -441,7 +441,7 @@ template class duration_scalar<duration_ns>;
 template <typename T>
 typename timestamp_scalar<T>::rep_type timestamp_scalar<T>::ticks_since_epoch()
 {
-  return this->value().time_since_epoch().count();
+  return this->value(cudf::get_default_stream()).time_since_epoch().count();
```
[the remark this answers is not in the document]
karthikeyann: updated to be consistent with duration scalar.

bdice (approved):

vyasr (approved):

PR #14360, opened 2023-11-03
Parquet sub-rowgroup reading.

### LINKED ISSUE #14270: [FEA] Proposal for sub-rowgroup reading in the parquet reader.
The Spark RAPIDS plugin team has customers with parquet files that have proven tricky to load with the existing reader.  There have been several categories of problems:

- Output sizes that are larger than the 32 bit size limits of cuDF.
- Files that are large enough such that their compressed and decompressed intermediate data does not fit on the GPU all at once.
- Files where individual row groups are large enough such that their compressed and decompressed intermediate data does not fit on the GPU all at once.

Up to this point, we have handled the first two items.  The first pass at the chunked reader takes care of output sizes by breaking decoded output up into "chunks".   The second item is handled with more functionality in the chunked reader that divides the input stage up into "passes" that attempt to limit how much memory is used for the temporary decompression data.

However we have run into situations where this is not enough.  We have at least one case where there is a file containing a single row group with a single column in it.  The compressed size is `~120 MB`, but the decompressed size is `~19 GB`.   This causes out of memory issues immediately.  

So the solution is to change things such that our input passes happen at the sub-rowgroup level.  A quick overview of the way things are arranged in Parquet:

```
Column A                          Column B                      ...
   Row Group 1, Chunk A             Row Group 1, Chunk B
      Page 0                            Page 0
      Page 1                            Page 1                    
      ...                               ...
```

The file is broken up into "row groups", which are the format's mechanism for allowing the user to select what rows to load out of the file.  Within each row group, each column contains a single "chunk".   Each chunk is the unit of IO, containing page header information and the compressed data for the pages themselves.   Each chunk contains a variable number of pages, and the pages themselves are individually compressed/decompressed.

So while we have to read whole rowgroup chunks in one shot, we do not have to decompress them in one shot. 

At the input level, the reader currently does the following:

```
L = user requested limit on temporary memory usage for io and decompression
while(rowgroups left to read){
   R = select_more_row_groups_that_will_fit_in_memory(L);
   D = load_and_decompress_rowgroups(R);
   O = decode_decompressed_data(D);   
}
```
Additionally, the last step may produce more than one "chunk" of actual output but that detail is left out here.  The issue we face is when R contains just 1 rowgroup that is too large by itself to process.  So the proposal here is to add another layer which processes groups of pages below the rowgroups themselves.

```
L = user requested limit on temporary memory usage for io and decompression
while(rowgroups left to read){
   R = select_more_row_groups(L);
   while(there_are_pages_left_to_decode(R)){
      P = load_and_decompress_more_pages(R, L);
      O = decode_decompressed_data(P);   
   }
}
```
At the high level this is a reasonably straightforward change.   The code in the reader's  `load_and_decompress_data` https://github.com/rapidsai/cudf/blob/301dce10382a57cfb6cfff6481d15b4ba0936b28/cpp/src/io/parquet/reader_impl_preprocess.cu#L751, `reader_and_decompress_column_chunks` https://github.com/rapidsai/cudf/blob/301dce10382a57cfb6cfff6481d15b4ba0936b28/cpp/src/io/parquet/reader_impl_preprocess.cu#L682 and `decompress_page_data` https://github.com/rapidsai/cudf/blob/301dce10382a57cfb6cfff6481d15b4ba0936b28/cpp/src/io/parquet/reader_impl_preprocess.cu#L348  functions needs to be broken up and arranged such that it is done iteratively as part of the overall chunking/input pass loop.

There are a couple of fuzzy details though.  At the point where we call `select_more_row_groups(L)` we will not know the number of pages or their individual compressed and decompressed sizes.  All we will know is the total compressed and decompressed size for an entire column chunk.  In our bad case here, this is `120 MB` and `19 GB`, respectively.  We expect to see limit values (L) of maybe 2 GB or 4 GB.  It is generally safe to expect that individual pages are roughly `1 MB` of uncompressed data (For our example data, the `19 GB` of decompressed data is contained in `15,000` pages so it holds roughly true).  So how do we choose the granularity of chunks to load?  

Option 1 :  Only consider compressed size and fill the limit L up to some hardcoded ratio of compressed/decompressed data size.  Say, `25%` of L.  So if the user has specified a `4 GB` limit, we will read up to `1 GB` of compressed data and use the remaining `3 GB` for decompression.  In our example case, we have `120 MB` of compressed data, which would leave > `3.8 GB` of space for decompressing and processing the pages in batches.  But if we had a different file structure where we had 100 of these extremely large chunks, we might buffer up ~6 of them (120 MB x 6) before starting the decompression passes. 

Option 2 : Consider (compressed + decompressed size).  Read until the sum goes over the used specified limit L.  The idea here would be that when we have huge disparities in compressed/decompressed size, we will prefer to read as little as we can and leave maximum space for batches of decompression. 

Option 3 :  ?

For our case, there would be no difference between options 1 and 2 since a single chunk explodes everything.  But for other cases, option 1 would bias things towards doing larger IO reads and the cost of having less decompression space to work with.  And option 2 would bias things towards maximizing decompression space for extreme cases.  

It is worth mentioning that depending on the compression codec used, nvcomp itself may require a very large amount of temporary memory.  For example using ZSTD, it may require as much as 2.5x the size of the decompressed data as purely temporary memory.  Whereas SNAPPY requires no scratch space at all.   And unfortunately, we will not know the compression types (they can be mixed within the file) until we have actually performed the reads and gotten the page headers.

----------------------------------------------
Note that all of these chunked reader behaviors are optional and not default.  The default behavior is to just attempt to read everything at once, requiring no user tweaking.   Any implementation here would leave that as is and would be done in such a way that there should be no performance impact when "no limit" is chosen.

nvdbaranec:
closes #14270

Implementation of sub-rowgroup reading of Parquet files.  This PR implements an additional layer on top of the existing chunking system.  Currently, the reader takes two parameters:  `input_pass_read_limit` which specifies a limit on temporary memory usage when reading and decompressing file data;  and `output_pass_read_limit` which specifies a limit on how large an output chunk (a table) can be.

Currently when the user specifies a limit via `input_pass_read_limit`, the reader will perform multiple `passes` over the file at row-group granularity.  That is, it will control how many row groups it will read at once to conform to the specified limit.

However, there are cases where this is not sufficient.  So this PR changes things so that we now have `subpasses` below the top level `passes`.  It works as follows:

- We read a set of input chunks based on the `input_pass_read_limit` but we do not decompress them immediately. This constitutes a `pass`.
- Within each pass of compressed data, we progressively decompress batches of pages as `subpasses`.
- Within each `subpass` we apply the output limit to produce `chunks`.

So the overall structure of the reader is:  (read) `pass` -> (decompress) `subpass` -> (decode) `chunk`

Major sections of code changes:

- Previously the incoming page data in the file was unsorted. To handle this we later on produced a `page_index` that could be applied to the array to get them in schema-sorted order.  This was getting very unwieldy so I just sort the pages up front now and the `page_index` array has gone away.

- There are now two sets of pages to be aware of in the code.  Within each `pass_intermediate_data` there is the set of all pages within the current set of loaded row groups.  And then within the `subpass_intermediate_data` struct there is a separate array of pages representing the current batch of decompressed data we are processing.  To keep the confusion down I changed a good amount of code to always reference it's array though it's associated struct.  Ie,  `pass.pages` or `subpass.pages`. In addition, I removed the `page_info` from `ColumnChunkDesc` to help prevent the kernels from getting confused. `ColumnChunkDesc` now only has a `dict_page` field which is constant across all subpasses.

- The primary entry point for the chunking mechanism is in `handle_chunking`. Here we iterate through passes, subpasses and output chunks.  Successive subpasses are computed and preprocessed through here. 

- The volume of diffs you'll see in `reader_impl_chunking.cu` is a little deceptive.  A lot of this is just functions (or pieces of functions) that have been moved over from either `reader_impl_preprocess.cu` or `reader_impl_helpers.cpp`.   The most relevant actual changes are in: ` handle_chunking`, `compute_input_passes`, `compute_next_subpass`, and `compute_chunks_for_subpass`.

Note on tests:   I renamed `parquet_chunked_reader_tests.cpp` to `parquet_chunked_reader_test.cu` as I needed to use thrust. The only actual changes in the file are the addition of the `ParquetChunkedReaderInputLimitConstrainedTest` and `ParquetChunkedReaderInputLimitTest` test suites at the bottom.


GregoryKimball:
Hello @etseidl we think that this work #14360 will have some conflicts with your decoder addition in #14101. Our plan is to complete the work on #14101 first, and then resolve conflicts to this PR. In the meantime, would you please take a look at @nvdbaranec 's work here?

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -30,21 +30,22 @@ namespace cudf::io::parquet::detail {
 
 void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
 {
-  auto& chunks              = _pass_itm_data->chunks;
-  auto& pages               = _pass_itm_data->pages_info;
-  auto& page_nesting        = _pass_itm_data->page_nesting_info;
-  auto& page_nesting_decode = _pass_itm_data->page_nesting_decode_info;
+  auto& pass    = *_pass_itm_data;
+  auto& subpass = *pass.subpass;
```
etseidl, line 34:
```suggestion
  auto& subpass = *pass.subpass;
  auto& chunks  = pass.chunks;
  auto& pages   = subpass.pages;
```
Would cut down on a lot of the diffs in this function, but if you want to emphasize the "pass" aspect then nevermind :)

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -1403,30 +1156,30 @@ void reader::impl::allocate_columns(size_t skip_rows, size_t num_rows, bool uses
...
   rmm::device_uvector<size_t> d_col_sizes(col_sizes.size(), _stream);
 
   // use page_index to fetch page string sizes in the proper order
-  auto val_iter = thrust::make_transform_iterator(
-    page_index.begin(), page_to_string_size{pages.device_ptr(), chunks.device_ptr()});
+  auto val_iter = thrust::make_transform_iterator(subpass.pages.d_begin(),
+                                                  page_to_string_size{pass.chunks.d_begin()});
 
   // do scan by key to calculate string offsets for each page
   thrust::exclusive_scan_by_key(rmm::exec_policy(_stream),
```
etseidl, line 1172: The other conflict with #14101 is here, where now `rmm::exec_policy_nosync` is used (and for the `reduce_by_key` below as well).

etseidl:
It's a lot to digest, but looks great so far. I have a few questions. First, will this help with skip_rows? I'm thinking of the predicate case where an index gives you a range of rows to read from the middle of a rowgroup. Can this work be modified to (or does it already handle) process just the pages needed to satisfy the predicate along with any needed dictionary pages?

The second is if the size statistics from #14000 are available, would you still use this mechanism but feed in the stats, or would it be better to have an entirely different path for stats driven chunked reading?

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -1280,11 +1280,11 @@ inline __device__ bool setupLocalPageInfo(page_state_s* const s,
           if (((s->col.data_type & 7) == BYTE_ARRAY) && (s->col.str_dict_index)) {
             // String dictionary: use index
             s->dict_base = reinterpret_cast<uint8_t const*>(s->col.str_dict_index);
-            s->dict_size = s->col.page_info[0].num_input_values * sizeof(string_index_pair);
+            s->dict_size = s->col.dict_page->num_input_values * sizeof(string_index_pair);
           } else {
             s->dict_base =
-              s->col.page_info[0].page_data;  // dictionary is always stored in the first page
-            s->dict_size = s->col.page_info[0].uncompressed_page_size;
+              s->col.dict_page->page_data;  // dictionary is always stored in the first page
```
ttnghia, line 1286:
```suggestion
              s->col.dict_page->page_data;
```

ttnghia: Also join lines 1285-1286.

cpp/src/io/parquet/page_hdr.cu
```diff
@@ -382,11 +384,10 @@ __global__ void __launch_bounds__(128)
...
       bs->page.kernel_mask         = 0;
     }
-    num_values     = bs->ck.num_values;
-    page_info      = bs->ck.page_info;
-    num_dict_pages = bs->ck.num_dict_pages;
-    max_num_pages  = (page_info) ? bs->ck.max_num_pages : 0;
-    values_found   = 0;
+    num_values    = bs->ck.num_values;
+    page_info     = chunk_pages ? chunk_pages[chunk].pages : nullptr;
+    max_num_pages = (page_info) ? bs->ck.max_num_pages : 0;
```
ttnghia, line 389:
```suggestion
    max_num_pages = page_info ? bs->ck.max_num_pages : 0;
```

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -538,10 +538,10 @@ __global__ void __launch_bounds__(preprocess_block_size) gpuComputePageStringSiz
         if (col.str_dict_index) {
           // String dictionary: use index
           dict_base = reinterpret_cast<const uint8_t*>(col.str_dict_index);
-          dict_size = col.page_info[0].num_input_values * sizeof(string_index_pair);
+          dict_size = col.dict_page->num_input_values * sizeof(string_index_pair);
         } else {
-          dict_base = col.page_info[0].page_data;  // dictionary is always stored in the first page
-          dict_size = col.page_info[0].uncompressed_page_size;
+          dict_base = col.dict_page->page_data;  // dictionary is always stored in the first page
```
ttnghia, line 543:
```suggestion
          dict_base = col.dict_page->page_data;
```

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -375,6 +389,13 @@ struct ColumnChunkDesc {
   int32_t src_col_schema{};  // my schema index in the file
 };
 
+/**
+ * @brief A utility structure for use in decoding page headers.
+ */
+struct chunk_page_info {
+  PageInfo* pages;
+};
```
ttnghia, line 397: Is this redundant? So the struct contains a pointer, and we will use pointer of pointer?

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -520,11 +541,13 @@ constexpr bool is_string_col(ColumnChunkDesc const& chunk)
  * @brief Launches kernel for parsing the page headers in the column chunks
  *
  * @param[in] chunks List of column chunks
+ * @param[in] chunk_pages List of pages associated with the chunks, in chunk-sorted order
  * @param[in] num_chunks Number of column chunks
  * @param[out] error_code Error code for kernel failures
  * @param[in] stream CUDA stream to use
  */
 void DecodePageHeaders(ColumnChunkDesc* chunks,
+                       chunk_page_info* chunk_pages,
```
ttnghia, line 550: Pointer of pointer? Can we just use `PageInfo*`?

vuule (commented):
first subpass ;)

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -311,231 +333,165 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
...
+  stream.synchronize();
 
-  // Count the exact number of compressed pages
-  size_t num_comp_pages    = 0;
-  size_t total_decomp_size = 0;
+  // validate page encodings
+  CUDF_EXPECTS(std::all_of(pass.pages.begin(),
+                           pass.pages.end(),
+                           [](auto const& page) { return is_supported_encoding(page.encoding); }),
+               "Unsupported page encoding detected");
```
etseidl, line 425:
```suggestion
```
This was brought up by @vuule in a different issue, but these lines were removed in #14237. The `std::all_of` will actually never return `false` because of the test up at line 329.

vuule (commented):
almost finished the first pass.
Bunch of tiny suggestion, nothing pivotal.

vuule (commented):
almost finished the first pass.
Bunch of tiny suggestion, nothing pivotal.

GregoryKimball:
Hello @etseidl would you please share any new feedback you might have about this work?

cpp/src/io/parquet/reader_impl.cpp
nvdbaranec: Yeah I want to be very explicit about where various data is coming from, pass or subpass.  Although in reality, the only thing that ever really changes is the list of pages (chunks are constant across an entire pass).

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -331,60 +341,16 @@ void reader::impl::prepare_data(int64_t skip_rows,
...
-        CUDF_EXPECTS(_pass_itm_data->output_chunk_read_info.size() == 1,
-                     "Reading the whole file should yield only one chunk.");
-      }
-    }
-
-    _pass_preprocessed = true;
-  }
+  // handle any chunking work (ratcheting through the subpasses and chunks within
+  // our current pass)
+  if (_file_itm_data.num_passes() > 0) { handle_chunking(uses_custom_row_bounds); }
```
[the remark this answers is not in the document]
nvdbaranec: The other case is if the user has somehow managed to select no row groups.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -733,22 +690,23 @@ std::pair<bool, std::vector<std::future<void>>> reader::impl::read_and_decompres
...
 }
 
-void reader::impl::load_and_decompress_data()
+void reader::impl::load_compressed_data()
 {
+  auto& pass = *_pass_itm_data;
+
   // This function should never be called if `num_rows == 0`.
-  CUDF_EXPECTS(_pass_itm_data->num_rows > 0, "Number of reading rows must not be zero.");
+  // CUDF_EXPECTS(_pass_itm_data->num_rows > 0, "Number of reading rows must not be zero.");
```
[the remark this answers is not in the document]
nvdbaranec: Good catch.  Debugging remnant.

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -138,20 +140,91 @@ class reader::impl {
...
    * @return pair of boolean indicating if compressed chunks were found and a vector of futures for
    * read completion
    */
-  std::pair<bool, std::vector<std::future<void>>> read_and_decompress_column_chunks();
+  std::pair<bool, std::vector<std::future<void>>> read_column_chunks();
+
+  /**
+   * @brief Load compressed data and page information for the current pass.
+   */
+  void load_compressed_data();
```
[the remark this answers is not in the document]
nvdbaranec: I'll rename to read_compressed_data() just to be consistent.  To me "read" is more suggestive of "actually perform IO" where "load" is a little more hazy.

cpp/src/io/parquet/reader_impl_chunking.hpp
```diff
@@ -30,58 +30,101 @@ struct file_intermediate_data {
...
-  cudf::detail::hostdevice_vector<PageNestingDecodeInfo> page_nesting_decode_info{};
+  cudf::detail::hostdevice_vector<PageInfo> pages{};
 
-  rmm::device_uvector<int32_t> page_keys{0, rmm::cuda_stream_default};
-  rmm::device_uvector<int32_t> page_index{0, rmm::cuda_stream_default};
-  rmm::device_uvector<string_index_pair> str_dict_index{0, rmm::cuda_stream_default};
+  // offsets to each group of input pages (by column/schema, indexed by _input_columns.size())
+  // so if we had 2 columns/schemas, with page keys
+  //
+  // 1 1 1 1 1 2 2 2
```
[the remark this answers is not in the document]
nvdbaranec: Correct.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -593,11 +547,11 @@ void reader::impl::allocate_nesting_info()
     auto cur_schema = _metadata->get_schema(schema_idx);
     int cur_depth   = max_depth - 1;
```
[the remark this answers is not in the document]
nvdbaranec: Added the check.  Worth pointing out that for historical reasons, `page_nesting_info` is sometimes over-allocated. We alllocate the worst case of max input column depth (ie, depth in the parquet schema tree) and max output column depth (depth in the output cudf columns) for convenience.  `page_nesting_decode_info` needs to remain that way, but `page_nesting_info` could be reduced.   I think that's better done as a followup though, since it's tangential to this PR.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -311,231 +333,165 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
...
-      page_count += page_stride;
-    }
-  };
+  // compute offsets to each group of input pages.
+  // page_keys:   1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3
+  //
+  // result:      0,          4,          8
+  rmm::device_uvector<size_type> page_counts(pass.pages.size() + 1, stream);
+  auto page_keys             = make_page_key_iterator(pass.pages);
+  auto const page_counts_end = thrust::reduce_by_key(rmm::exec_policy(stream),
```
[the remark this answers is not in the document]
nvdbaranec: I think it makes sense to leave this one as exec_policy() since it implicitly syncs because of the return value.  The alternative would be:
```
result = thrust::reduce_by_key(rmm::exec_policy_nosync(stream), ...)
stream.synchronize()
```
which would seem odd to me since we're synchronizing after being handed a result.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -692,13 +647,15 @@ std::pair<bool, std::vector<std::future<void>>> reader::impl::read_and_decompres
 
   // Initialize column chunk information
   size_t total_decompressed_size = 0;
-  auto remaining_rows            = num_rows;
+  // TODO: make this respect the pass-wide skip_rows/num_rows instead of the file-wide
+  // skip_rows/num_rows
```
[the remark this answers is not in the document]
nvdbaranec: Yes.  The `remaining_rows` check assumed we would always be iterating over all possible rows the user has requested.  But with passes, this is not necessarily the case.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -938,18 +832,16 @@ struct start_offset_output_iterator {
 
   constexpr void operator=(start_offset_output_iterator const& other)
```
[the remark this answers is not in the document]
nvdbaranec: I'm not quite sure.  This is old code that has been moved.

cpp/src/io/parquet/reader_impl_chunking.hpp
```diff
@@ -30,58 +30,101 @@ struct file_intermediate_data {
...
 };
 
 /**
- * @brief Struct to identify the range for each chunk of rows during a chunked reading pass.
+ * @brief Struct to identify a range of rows.
  */
-struct chunk_read_info {
+struct row_range {
+  size_t skip_rows;
+  size_t num_rows;
```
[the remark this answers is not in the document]
nvdbaranec: "No rows".  The more I see `skip_rows`/`num_rows` the more I wish we used the more idiomatic `start_row`/`end_row` since we are always working on absolute indices any way (skip_rows is always an absolute value within the selected row groups, regardless of pass, subpass, output chunk, etc).  

It's probably worth a followup when someone has a bunch of free time :)

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +379,715 @@ template <typename T = uint8_t>
...
+std::pair<size_t, size_t> get_row_group_size(RowGroup const& rg)
+{
+  auto compressed_size_iter = thrust::make_transform_iterator(
+    rg.columns.begin(), [](ColumnChunk const& c) { return c.meta_data.total_compressed_size; });
+
+  // the trick is that total temp space needed is tricky to know
+  auto const compressed_size =
+    std::reduce(compressed_size_iter, compressed_size_iter + rg.columns.size());
+  auto const total_size = compressed_size + rg.total_byte_size;
+  return {compressed_size, total_size};
```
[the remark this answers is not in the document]
nvdbaranec: Right.  This is a remnant from early on where I was experimenting with how to determine how many compressed bytes should be loaded.  The state it's in now is pretty solid so this can be simplified.

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -253,11 +278,13 @@ void reader::impl::decode_page_data(size_t skip_rows, size_t num_rows)
     }
   }
 
+  _stream.synchronize();
```
etseidl, line 281: I don't think this sync is necessary.  The `error_code.value()` call will sync all the D2H copies above, and the H2D copies can wait for the sync at the end of the function.

cpp/src/io/utilities/column_buffer.cpp
```diff
@@ -145,6 +148,30 @@ string_policy column_buffer_base<string_policy>::empty_like(string_policy const&
   return new_buff;
 }
 
+template <typename string_policy>
+std::string type_to_name(column_buffer_base<string_policy> const& buffer, bool include_nesting)
```
etseidl, line 152: Is `include_nesting` still needed?

cpp/src/io/utilities/column_buffer.hpp
```diff
@@ -253,6 +253,10 @@ std::unique_ptr<column> empty_like(column_buffer_base<string_policy>& buffer,
                                    rmm::cuda_stream_view stream,
                                    rmm::mr::device_memory_resource* mr);
 
+template <class string_policy>
```
etseidl, line 256: docstring or is this temporary?

etseidl (commented):
A few small nits...still working up the courage to slog through reader_impl_chunking :sweat_smile:

cpp/src/io/parquet/reader_impl_preprocess.cu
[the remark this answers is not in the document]
nvdbaranec: So it turns out there is a bug here, but it's very old.  Verified with mainline 24.02.  In some cases (certain schema arragements), the 0th element can remain uninitialized, which has a mild side effect:  The size estimation for output chunks (but only output chunks, not the input stuff in this PR) will end up a little small for list columns.

I'm going to revert this change for now and do a proper fix as a followup so I don't accidentally blow up this PR :)

https://github.com/rapidsai/cudf/issues/14634

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -412,6 +428,15 @@ struct ColumnChunkDesc {
 
   int32_t src_col_index{};   // my input column index
   int32_t src_col_schema{};  // my schema index in the file
+
+  float list_bytes_per_row_est{};  // for LIST columns, an estimate on number of bytes per row
```
ttnghia, line 432: `float` type?

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -412,6 +428,15 @@ struct ColumnChunkDesc {
...
+};
+
+/**
+ * @brief A utility structure for use in decoding page headers.
+ */
+struct chunk_page_info {
+  PageInfo* pages;
```
ttnghia, line 439: This is just a type aliasing right?

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -138,20 +140,102 @@ class reader::impl {
...
+  void setup_next_pass(bool uses_custom_row_bounds);
+
+  /**
+   * @brief Setup step for the next decompression subpass.
+   *
+   * @param uses_custom_row_bounds Whether or not num_rows and skip_rows represents user-specific
+   *        bounds
+   *
+   * A 'subpass' is defined as a subset of pages within a pass that are
+   * decompressed a decoded as a batch. Subpasses may be further subdivided
```
ttnghia, line 184:
```suggestion
   * decompressed and decoded as a batch. Subpasses may be further subdivided
```

cpp/src/io/parquet/reader_impl.hpp
```diff
@@ -138,20 +140,102 @@ class reader::impl {
...
+   *        bounds
+   * @param filter Optional AST expression to filter output rows
+   * @return The output table along with columns' metadata
+   */
+  table_with_metadata read_chunk_internal(
+    bool uses_custom_row_bounds,
+    std::optional<std::reference_wrapper<ast::expression const>> filter);
+
+  // utility functions
+ private:
```
ttnghia, line 204: Do the functions below need to be member functions? I typically recommend to have just free functions if possible.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -19,45 +19,55 @@
 
 #include <cudf/detail/iterator.cuh>
 #include <cudf/detail/utilities/integer_utils.hpp>
+#include <cudf/detail/utilities/vector_factories.hpp>
 
+#include <io/comp/nvcomp_adapter.hpp>
+
```
ttnghia, line 25:
```suggestion
```

cpp/src/io/parquet/reader_impl_chunking.hpp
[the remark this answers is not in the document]
ttnghia: I support this. Let's call `start_row` and `end_row`.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -19,45 +19,55 @@
...
 
 namespace {
 
-struct cumulative_row_info {
-  size_t row_count;   // cumulative row count
-  size_t size_bytes;  // cumulative size in bytes
-  int key;            // schema index
+struct split_info {
+  row_range rows;
+  int64_t split_pos;
```
ttnghia, line 46: Should this be consistent with `row_range` by using `size_t` instead of `int64_t`?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -19,45 +19,55 @@
...
+// do a subpass decode. if the difference between the user specified limit and
+// the actual memory used for compressed/temp data is > than this value, we will still use
+// at least this many additional bytes.
+constexpr size_t minimum_subpass_expected_size = 200 * 1024 * 1024;
+
 #if defined(CHUNKING_DEBUG)
 void print_cumulative_page_info(cudf::detail::hostdevice_vector<PageInfo>& pages,
-                                rmm::device_uvector<int32_t> const& page_index,
-                                rmm::device_uvector<cumulative_row_info> const& c_info,
+                                cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks,
```
ttnghia, line 57: Assume that this is just a print function, the input should be `const`. And also somebody (including me) will yell at this not being `span`.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -114,13 +130,13 @@ void print_cumulative_row_info(host_span<cumulative_row_info const> sizes,
 #endif  // CHUNKING_DEBUG
 
 /**
- * @brief Functor which reduces two cumulative_row_info structs of the same key.
+ * @brief Functor which reduces two cumulative_page_info structs of the same key.
  */
-struct cumulative_row_sum {
-  cumulative_row_info operator()
-    __device__(cumulative_row_info const& a, cumulative_row_info const& b) const
+struct cumulative_page_sum {
```
ttnghia, line 135: Should it be `page_prefix_sum`? Because `cummulative_` is essentially `prefix_`.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -178,32 +194,57 @@ __device__ size_t row_size_functor::operator()<string_view>(size_t num_rows, boo
  *
  * Sums across all nesting levels.
  */
-struct get_cumulative_row_info {
-  PageInfo const* const pages;
-
-  __device__ cumulative_row_info operator()(size_type index)
+struct get_page_output_size {
+  __device__ cumulative_page_info operator()(PageInfo const& page)
```
ttnghia, line 198: `const`?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -178,32 +194,57 @@ __device__ size_t row_size_functor::operator()<string_view>(size_t num_rows, boo
...
     // total nested size, not counting string data
-    auto iter =
-      cudf::detail::make_counting_transform_iterator(0, [page, index] __device__(size_type i) {
-        auto const& pni = page.nesting[i];
-        return cudf::type_dispatcher(
-          data_type{pni.type}, row_size_functor{}, pni.size, pni.nullable);
-      });
-
-    size_t const row_count = static_cast<size_t>(page.nesting[0].size);
+    auto iter = cudf::detail::make_counting_transform_iterator(0, [page] __device__(size_type i) {
```
ttnghia, line 205: `__device__` lambda must be wrapped in `proclaim_return_type` now.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -178,32 +194,57 @@ __device__ size_t row_size_functor::operator()<string_view>(size_t num_rows, boo
...
       thrust::reduce(thrust::seq, iter, iter + page.num_output_nesting_levels) + page.str_bytes,
       page.src_col_schema};
   }
 };
 
+/**
+ * @brief Functor which sets the (uncompressed) size of a page.
+ */
+struct get_page_input_size {
+  __device__ cumulative_page_info operator()(PageInfo const& page)
```
ttnghia, line 220:
```suggestion
  __device__ cumulative_page_info operator()(PageInfo const& page) const
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -178,32 +194,57 @@ __device__ size_t row_size_functor::operator()<string_view>(size_t num_rows, boo
...
+    return {0, static_cast<size_t>(page.uncompressed_page_size), page.src_col_schema};
+  }
+};
+
+/**
+ * @brief Functor which sets the absolute row index of a page in a cumulative_page_info struct
+ */
+struct set_row_index {
+  device_span<const ColumnChunkDesc> chunks;
+  device_span<const PageInfo> pages;
```
ttnghia, line 236: East const.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -219,84 +260,91 @@ struct get_cumulative_row_info {
...
  */
-struct row_total_size {
-  cumulative_row_info const* c_info;
+struct page_total_size {
+  cumulative_page_info const* c_info;
   size_type const* key_offsets;
   size_t num_keys;
 
-  __device__ cumulative_row_info operator()(cumulative_row_info const& i)
+  __device__ cumulative_page_info operator()(cumulative_page_info const& i)
```
ttnghia, line 268:
```suggestion
  __device__ cumulative_page_info operator()(cumulative_page_info const& i) const
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -219,84 +260,91 @@ struct get_cumulative_row_info {
...
   {
     // sum sizes for each input column at this row
     size_t sum = 0;
     for (int idx = 0; idx < num_keys; idx++) {
       auto const start = key_offsets[idx];
       auto const end   = key_offsets[idx + 1];
       auto iter        = cudf::detail::make_counting_transform_iterator(
-        0, [&] __device__(size_type i) { return c_info[i].row_count; });
+        0, [&] __device__(size_type i) { return c_info[i].row_index; });
```
ttnghia, line 276: `proclaim_return_type` for device lambda.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -219,84 +260,91 @@ struct get_cumulative_row_info {
...
   }
 };
 
 /**
- * @brief Given a vector of cumulative {row_count, byte_size} pairs and a chunk read
- * limit, determine the set of splits.
+ * @brief Functor which returns the compressed data size for a chunk
+ */
+struct get_chunk_compressed_size {
+  __device__ size_t operator()(ColumnChunkDesc const& chunk) { return chunk.compressed_size; }
```
ttnghia, line 289:
```suggestion
  __device__ size_t operator()(ColumnChunkDesc const& chunk) const { return chunk.compressed_size; }
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -219,84 +260,91 @@ struct get_cumulative_row_info {
...
+ * @brief Find the first entry in the aggreggated_info that corresponds to the specified row
  *
- * @param sizes Vector of cumulative {row_count, byte_size} pairs
- * @param num_rows Total number of rows to read
- * @param chunk_read_limit Limit on total number of bytes to be returned per read, for all columns
  */
-std::vector<chunk_read_info> find_splits(std::vector<cumulative_row_info> const& sizes,
-                                         size_t num_rows,
-                                         size_t chunk_read_limit)
+size_t find_start_index(std::vector<cumulative_page_info> const& aggregated_info, size_t start_row)
```
ttnghia, line 296: `host_span`?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -219,84 +260,91 @@ struct get_cumulative_row_info {
...
+                        size_t cur_cumulative_size,
+                        std::vector<cumulative_page_info> const& sizes,
+                        size_t size_limit)
+{
+  auto start = thrust::make_transform_iterator(sizes.begin(), [&](cumulative_page_info const& i) {
+    return i.size_bytes - cur_cumulative_size;
+  });
+  auto end   = start + sizes.size();
+
+  int64_t split_pos = thrust::lower_bound(thrust::seq, start + cur_pos, end, size_limit) - start;
```
ttnghia, line 331: `auto const` for all.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -219,84 +260,91 @@ struct get_cumulative_row_info {
...
+ * @brief Given a current position and row index, find the next split based on the
+ * specified size limit
+ *
+ * @returns The inclusive index within `sizes` where the next split should happen
+ *
+ */
+int64_t find_next_split(int64_t cur_pos,
+                        size_t cur_row_index,
+                        size_t cur_cumulative_size,
+                        std::vector<cumulative_page_info> const& sizes,
```
ttnghia, line 323: `host_span`?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +389,826 @@ template <typename T = uint8_t>
 }
 
 struct row_count_compare {
-  __device__ bool operator()(cumulative_row_info const& a, cumulative_row_info const& b)
+  __device__ bool operator()(cumulative_page_info const& a, cumulative_page_info const& b)
```
ttnghia, line 392:
```suggestion
  __device__ bool operator()(cumulative_page_info const& a, cumulative_page_info const& b) const
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +389,826 @@ template <typename T = uint8_t>
...
+/**
+ * @brief For a set of cumulative_page_info data, adjust the size_bytes field
+ * such that it reflects the worst case for all pages that span the same rows.
+ *
+ * By doing this, we can now look at row X and know the total
+ * byte cost for all pages that span row X, not just the cost up to row X itself.
+ *
+ * This function is asynchronous. Call stream.synchronize() before using the
+ * results.
+ *
```
ttnghia, line 423:
```suggestion
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +389,826 @@ template <typename T = uint8_t>
...
+               c_info_sorted.end(),
+               row_count_compare{});
+
+  // page keys grouped by split.
+  rmm::device_uvector<int32_t> page_keys_by_split{c_info.size(), stream};
+  thrust::transform(rmm::exec_policy_nosync(stream),
+                    c_info_sorted.begin(),
+                    c_info_sorted.end(),
+                    page_keys_by_split.begin(),
+                    [] __device__(cumulative_page_info const& c) { return c.key; });
```
ttnghia, line 443: `proclaim_return_type`.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +389,826 @@ template <typename T = uint8_t>
...
+}
+
+struct page_span {
+  size_t start, end;
+};
+
+struct get_page_row_index {
+  device_span<const cumulative_page_info> c_info;
+
+  __device__ size_t operator()(size_t i) { return c_info[i].row_index; }
```
ttnghia, line 486:
```suggestion
  __device__ size_t operator()(size_t i) const { return c_info[i].row_index; }
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +389,826 @@ template <typename T = uint8_t>
...
+                size_t _end_row)
+    : page_offsets(_page_offsets),
+      page_row_index(_page_row_index),
+      start_row(_start_row),
+      end_row(_end_row)
   {
-    return a.row_count < b.row_count;
+  }
+
+  __device__ page_span operator()(size_t column_index)
```
ttnghia, line 511:
```suggestion
  __device__ page_span operator()(size_t column_index) const
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +389,826 @@ template <typename T = uint8_t>
...
+        break;
+      default: CUDF_FAIL("Unexpected decompression dispatch"); break;
+    }
+    start_pos += codec.num_pages;
+  }
+
+  CUDF_EXPECTS(thrust::all_of(rmm::exec_policy(stream),
+                              comp_res.begin(),
+                              comp_res.end(),
+                              [] __device__(auto const& res) {
```
ttnghia, line 810: `proclaim_return_type`.

ttnghia (commented):
Another review pass. Still not done yet.

cpp/src/io/parquet/reader_impl.cpp
nvdbaranec: Hmm.  The error_code.value() doing the sync is a little bit too subtle for me.   Your average reader isn't going to recognize this easily.  I think the right thing to do here is to just move this particular synchronize right after the error_code.value(). 

Even though it won't technically be doing anything, I like the explicitness.

cpp/src/io/utilities/column_buffer.cpp
nvdbaranec: Nope. Good catch.

cpp/src/io/parquet/parquet_gpu.hpp
nvdbaranec: It could be a size_t, but since this is just an estimate I figured leaving it as a float couldn't hurt.

cpp/src/io/parquet/reader_impl_chunking.cu
nvdbaranec: This is actually needed.

cpp/src/io/parquet/reader_impl_chunking.cu
nvdbaranec: It could be, but I'm using `cumulative` elsewhere in here. In particular:  `cumulative_page_info` to represent cumulative row counts and byte sizes for groups of pages.  'prefix sum' doesn't seem as clear to me.

cpp/src/io/utilities/column_buffer.hpp
nvdbaranec: Done

cpp/src/io/parquet/parquet_gpu.hpp
nvdbaranec: This an array of arrays.  Ie, every element of in chunk_pages[] is a separate array of pages.  That is, each chunk has an array of pages relevant to it.  It's (the PageInfo*) not stored as a span because the size of the arrays is stored in the chunks themselves:

https://github.com/rapidsai/cudf/blob/b01b06e4a17be55e009e1eabde09c07f767bacba/cpp/src/io/parquet/page_hdr.cu#L397

https://github.com/rapidsai/cudf/blob/b01b06e4a17be55e009e1eabde09c07f767bacba/cpp/src/io/parquet/page_hdr.cu#L458

cpp/src/io/parquet/parquet_gpu.hpp
nvdbaranec: See https://github.com/rapidsai/cudf/pull/14360#discussion_r1459833187

cpp/src/io/parquet/parquet_gpu.hpp
nvdbaranec: See https://github.com/rapidsai/cudf/pull/14360#discussion_r1459833187

cpp/src/io/parquet/reader_impl_chunking.cu
nvdbaranec: split_pos needs to handle the possibility of being negative when being used.

vyasr (approved):
CMake approval.

cpp/src/io/comp/nvcomp_adapter.hpp
```diff
@@ -111,6 +111,11 @@ void batched_decompress(compression_type compression,
                         size_t max_total_uncomp_size,
                         rmm::cuda_stream_view stream);
 
+size_t batched_decompress_temp_size(compression_type compression,
```
vuule, line 114: please add docs

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +399,962 @@ template <typename T = uint8_t>
...
+
+/**
+ * @brief Functor which returns total scratch space required based on computed decompression_info
+ * data.
+ *
+ */
+struct get_decomp_scratch {
+  size_t operator()(decompression_info const& di)
+  {
+    cudf::io::nvcomp::compression_type nvcomp_codec = cudf::io::nvcomp::compression_type::INVALID;
```
vuule, line 975: I don't think we need to add INVALID to the nvcomp adapter enum.
A helper function
`nvcomp_decomp_scratch(cudf::io::nvcomp::compression_type, decompression_info const&);`
that we call in the switch would eliminate the code after the switch with minimal duplication.

cpp/tests/io/parquet_chunked_reader_test.cu
```diff
@@ -953,64 +953,283 @@ TEST_F(ParquetChunkedReaderTest, TestChunkedReadNullCount)
...
 
-  {
-    // 12 chunks of 200k rows each, plus 1 final chunk of 100k rows.
-    auto const [result, num_chunks] = chunked_read(filepath, 0, 8002301);
-    EXPECT_EQ(num_chunks, 13);
-    CUDF_TEST_EXPECT_COLUMNS_EQUAL(expected, result->get_column(0));
-  }
+struct ParquetChunkedReaderInputLimitConstrainedTest : public cudf::test::BaseFixture {};
+
+TEST_F(ParquetChunkedReaderInputLimitConstrainedTest, SingleFixedWidthColumn)
```
vuule, line 1012: Supercalifragilisticexpialidocious has been real quiet since the new test name dropped :)

cpp/tests/io/parquet_chunked_reader_test.cu
```diff
@@ -953,64 +953,283 @@ TEST_F(ParquetChunkedReaderTest, TestChunkedReadNullCount)
...
+{
+  auto base_path          = temp_env->get_temp_filepath("single_col_fixed_width");
+  constexpr auto num_rows = 1'000'000;
+  auto iter1 = cudf::detail::make_counting_transform_iterator(0, [](int i) { return 15; });
```
vuule, line 1016: `thrust::make_constant_iterator(15)` should be a simpler option to generate the input here

cpp/tests/io/parquet_chunked_reader_test.cu
```diff
@@ -953,64 +953,283 @@ TEST_F(ParquetChunkedReaderTest, TestChunkedReadNullCount)
...
-    EXPECT_EQ(num_chunks, 25);
-    CUDF_TEST_EXPECT_COLUMNS_EQUAL(expected, result->get_column(0));
-  }
+void input_limit_test_read(std::string const& base_path,
+                           cudf::table_view const& t,
+                           size_t output_limit,
+                           size_t input_limit,
+                           int const expected_chunk_counts[4])
+{
+  std::vector<std::string> file_suffix{"_a.parquet", "_b.parquet", "_c.parquet", "_d.parquet"};
```
vuule, line 998: I feel like it would be simpler to generate the full file names in a helper function and pass the vector of names to input_limit_test_read/input_limit_test_write, rather then assembling the names in multiple places.

cpp/tests/io/parquet_chunked_reader_test.cu
```diff
@@ -953,64 +953,283 @@ TEST_F(ParquetChunkedReaderTest, TestChunkedReadNullCount)
...
+  cudf::test::fixed_width_column_wrapper<double> col1(iter1, iter1 + num_rows);
+  auto tbl = cudf::table_view{{col1}};
+
+  input_limit_test_write(base_path, tbl);
+
+  // semi-reasonable limit
+  constexpr int expected_a[] = {1, 17, 4, 1};
+  input_limit_test_read(base_path, tbl, 0, 2 * 1024 * 1024, expected_a);
+  // an unreasonable limit
+  constexpr int expected_b[] = {1, 50, 50, 1};
```
vuule, line 1026: how come we only get a single chunk in some of these cases?

cpp/tests/io/parquet_chunked_reader_test.cu
```diff
@@ -953,64 +953,283 @@ TEST_F(ParquetChunkedReaderTest, TestChunkedReadNullCount)
...
+};
+
+template <typename T>
+struct value_gen {
+  __device__ T operator()(int i) { return i % 1024; }
+};
+TEST_F(ParquetChunkedReaderInputLimitTest, List)
+{
+  auto base_path          = temp_env->get_temp_filepath("list");
+  constexpr int num_rows  = 50'000'000;
```
vuule, line 1073: holy moly! reading 200M ints 12 times!
How long does this one take?

vuule (commented):
flushing comments so far

cpp/tests/io/parquet_chunked_reader_test.cu
nvdbaranec: So in the initial case (array entry 0) the data is uncompressed - so no need for temporary space. There's only 1 row group, so the entire blob of data is ~8MB. We have no choice but to load that.  But once we do, there's no decompression to do so we can return the whole thing in 1 pass.

And the last case includes a dictionary, which obliterates the data size even before compression.  The whole row group is just a bit over 1KB.

cpp/tests/io/parquet_chunked_reader_test.cu
nvdbaranec: About 5 seconds on an A5000.

vuule: much better than I expected :+1:

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +399,965 @@ template <typename T = uint8_t>
...
+  pass.subpass  = std::make_unique<subpass_intermediate_data>();
+  auto& subpass = *pass.subpass;
+
+  auto const num_columns = _input_columns.size();
+
+  // if the user has passed a very small value (under the hardcoded minimum_subpass_expected_size),
+  // respect it.
+  auto const min_subpass_size = _input_pass_read_limit < minimum_subpass_expected_size
+                                  ? _input_pass_read_limit
+                                  : minimum_subpass_expected_size;
```
vuule, line 1224: std::min?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -19,45 +19,60 @@
...
-  int key;            // schema index
+struct split_info {
+  row_range rows;
+  int64_t split_pos;
 };
 
+// the minimum amount of memory we can safely expect to be enough to
+// do a subpass decode. if the difference between the user specified limit and
+// the actual memory used for compressed/temp data is > than this value, we will still use
+// at least this many additional bytes.
```
vuule, line 54: I don't undestand this sentence. Can you please give an example?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +399,965 @@ template <typename T = uint8_t>
...
+      pass.decomp_dict_data.size() +
+      thrust::reduce(rmm::exec_policy(_stream), chunk_iter, chunk_iter + pass.chunks.size());
+
+    // since there is only ever 1 dictionary per chunk (the first page), do it at the
+    // pass level.
+    build_string_dict_indices();
+
+    // if we are doing subpass reading, generate more accurate num_row estimates for list columns.
+    // this helps us to generate more accurate subpass splits.
+    if (_input_pass_read_limit != 0) { generate_list_column_row_count_estimates(); }
```
vuule, line 1190: isn't `_output_chunk_read_limit != 0 ` the condition to check if we have subpasses?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +399,965 @@ template <typename T = uint8_t>
...
+    : pass.base_mem_size + min_subpass_size >= _input_pass_read_limit
+      ? min_subpass_size
+      : _input_pass_read_limit - pass.base_mem_size;
+
+  auto [page_indices, total_pages, total_expected_size] =
+    [&]() -> std::tuple<std::vector<page_span>, size_t, size_t> {
+    // special case:  if we contain no compressed data, or if we have no input limit, we can always
+    // just do 1 subpass since what we already have loaded is all the temporary memory we will ever
+    // use.
+    if (!pass.has_compressed_data || _input_pass_read_limit == 0) {
```
vuule, line 1242: don't we have a single pass if _input_pass_read_limit == 0?

vuule (commented):
another flush, mostly questions in this batch

cpp/src/io/parquet/reader_impl_chunking.cu
nvdbaranec: What it's saying is:  
- Let's say the user has passed in a limit of 1GB
- We have somehow consumed 900 MB of that before we even start decompressing
- In theory we should only use 100 MB more memory, based on the user's limit.
- But we want to at least work with 200 MB at a time (to keep the GPU busy, etc), so it's just going to use 200 instead of the 100.

The counter-example is: 

- Let's say the user has passed in a limit of 1GB
- We have consumed 100 MB prior to decompression.
- We've got 800 MB left, which is more than 200, so we will decompress in 800 MB batches.

nvdbaranec: I can add the example:
```
// Example:
// - user has specified 1 GB limit
// - we have read in 900 MB of compressed data
// - that leaves us 100 MB of space for decompression batches
// - to keep the gpu busy, we really don't want to do less than 200 MB at a time so we're just going to use 200 MB of space  even if that goes past the user-specified limit.
```

vuule: the example is very clear, would not mind it

cpp/src/io/parquet/reader_impl.cpp
```diff
@@ -553,10 +539,9 @@ bool reader::impl::has_next()
                {} /*row_group_indices, empty means read all row groups*/,
                std::nullopt /*filter*/);
 
-  size_t const num_input_passes = std::max(
-    int64_t{0}, static_cast<int64_t>(_file_itm_data.input_pass_row_group_offsets.size()) - 1);
-  return (_pass_itm_data->current_output_chunk < _pass_itm_data->output_chunk_read_info.size()) ||
-         (_current_input_pass < num_input_passes);
+  // current_input_pass will only be incremented to be == num_passes after
+  // the last chunk in the last subpass in the last pass has been returned
+  return has_more_work();
```
ttnghia, line 544: Nit: I feel has "more work" is a bit vague. Does that mean has more data to process?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -19,45 +19,67 @@
...
-                                rmm::device_uvector<cumulative_row_info> const& c_info,
+void print_cumulative_page_info(device_span<const PageInfo> d_pages,
+                                device_span<const ColumnChunkDesc> d_chunks,
+                                device_span<const cumulative_page_info> d_c_info,
                                 rmm::cuda_stream_view stream)
 {
-  pages.device_to_host_sync(stream);
+  std::vector<PageInfo> pages              = cudf::detail::make_std_vector_sync(d_pages, stream);
+  std::vector<ColumnChunkDesc> chunks      = cudf::detail::make_std_vector_sync(d_chunks, stream);
+  std::vector<cumulative_page_info> c_info = cudf::detail::make_std_vector_sync(d_c_info, stream);
```
ttnghia, line 76: All sync is probably not very good, but it's fine for debugging purpose :man_shrugging:

cpp/src/io/parquet/reader_impl.cpp
nvdbaranec: Right. Work can mean a number of things:  actual IO, decompression, decoding, etc. It does generally mean "more data to process and return"

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +406,963 @@ template <typename T = uint8_t>
...
+  rmm::device_uvector<page_span> page_bounds(num_columns, stream);
+  auto iter = thrust::make_counting_iterator(size_t{0});
+  auto page_row_index =
+    cudf::detail::make_counting_transform_iterator(0, get_page_row_index{c_info});
+  thrust::transform(rmm::exec_policy_nosync(stream),
+                    iter,
+                    iter + num_columns,
+                    page_bounds.begin(),
+                    get_page_span{page_offsets, page_row_index, start_row, end_row});
+  auto h_page_bounds = cudf::detail::make_std_vector_sync(page_bounds, stream);
```
ttnghia, line 609: So we transfer to the host for further processing. Can we do further processing on device?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +406,963 @@ template <typename T = uint8_t>
...
+      copy_in, stream, rmm::mr::get_current_device_resource());
+    auto const d_copy_out = cudf::detail::make_device_uvector_async(
+      copy_out, stream, rmm::mr::get_current_device_resource());
+
+    gpu_copy_uncompressed_blocks(d_copy_in, d_copy_out, stream);
+  }
+
+  pages.host_to_device_async(stream);
+
+  stream.synchronize();
```
ttnghia, line 846:
```suggestion
  pages.host_to_device_sync(stream);
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +406,963 @@ template <typename T = uint8_t>
...
+               "Error during decompression");
+
+  // now copy the uncompressed V2 def and rep level data
+  if (not copy_in.empty()) {
+    auto const d_copy_in = cudf::detail::make_device_uvector_async(
+      copy_in, stream, rmm::mr::get_current_device_resource());
+    auto const d_copy_out = cudf::detail::make_device_uvector_async(
+      copy_out, stream, rmm::mr::get_current_device_resource());
+
+    gpu_copy_uncompressed_blocks(d_copy_in, d_copy_out, stream);
```
ttnghia, line 841: Does this function sync?

cpp/src/io/parquet/reader_impl_chunking.cu
nvdbaranec: I don't like hiding synchronizations like this.  In the future someone might add more asynchronous work and not see the `_sync()` function happening.  I think it's much cleaner to always terminate a series of async operations with an explicit `stream.synchronize()` as a clear flag for the reader.

cpp/src/io/parquet/reader_impl_chunking.cu
nvdbaranec: It does not.

cpp/src/io/parquet/reader_impl_chunking.cu
nvdbaranec: Actually, I confused this with the other comment https://github.com/rapidsai/cudf/pull/14360#discussion_r1464087196

The output from here is a series of spans used to copy pages around using std::copy().   This could in theory be done on the gpu, but it's a multi-buffer copy so it would require a little bit of thought.  I think this would be better suited as a followup.

cpp/src/io/parquet/reader_impl_chunking.cu
ttnghia: But `d_copy_in, d_copy_out` are destroyed after this line. If it doesn't sync then....?

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +406,963 @@ template <typename T = uint8_t>
...
+};
+
+/**
+ * @brief Functor which retrieves per-page decompression information.
+ *
+ */
+struct get_decomp_info {
+  device_span<const ColumnChunkDesc> chunks;
+
+  __device__ decompression_info operator()(PageInfo const& p)
```
ttnghia, line 951:
```suggestion
  __device__ decompression_info operator()(PageInfo const& p) const
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +406,963 @@ template <typename T = uint8_t>
...
+            static_cast<size_t>(p.uncompressed_page_size)};
+  }
+};
+
+/**
+ * @brief Functor which accumulates per-page decompression information.
+ *
+ */
+struct decomp_sum {
+  __device__ decompression_info operator()(decompression_info const& a, decompression_info const& b)
```
ttnghia, line 965:
```suggestion
  __device__ decompression_info operator()(decompression_info const& a, decompression_info const& b) const
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +406,963 @@ template <typename T = uint8_t>
...
+  }
+};
+
+/**
+ * @brief Functor which returns total scratch space required based on computed decompression_info
+ * data.
+ *
+ */
+struct get_decomp_scratch {
+  size_t operator()(decompression_info const& di)
```
ttnghia, line 980:
```suggestion
  size_t operator()(decompression_info const& di) const
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -19,45 +19,67 @@
...
+// data vs uncompressed data.
+constexpr float input_limit_compression_reserve = 0.3f;
+
 #if defined(CHUNKING_DEBUG)
-void print_cumulative_page_info(cudf::detail::hostdevice_vector<PageInfo>& pages,
-                                rmm::device_uvector<int32_t> const& page_index,
-                                rmm::device_uvector<cumulative_row_info> const& c_info,
+void print_cumulative_page_info(device_span<const PageInfo> d_pages,
+                                device_span<const ColumnChunkDesc> d_chunks,
+                                device_span<const cumulative_page_info> d_c_info,
```
ttnghia, line 71:
```suggestion
                                device_span<ColumnChunkDesc const> d_chunks,
                                device_span<cumulative_page_info const> d_c_info,
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -178,32 +206,59 @@ __device__ size_t row_size_functor::operator()<string_view>(size_t num_rows, boo
...
+ */
+struct get_page_input_size {
+  __device__ cumulative_page_info operator()(PageInfo const& page) const
+  {
+    // we treat dictionary page sizes as 0 for subpasses because we have already paid the price for
+    // them at the pass level.
+    if (page.flags & PAGEINFO_FLAGS_DICTIONARY) {
+      return cumulative_page_info{0, 0, page.src_col_schema};
+    }
+    return {0, static_cast<size_t>(page.uncompressed_page_size), page.src_col_schema};
```
ttnghia, line 241:
```suggestion
      return {0, 0, page.src_col_schema};
    }
    return {0, static_cast<size_t>(page.uncompressed_page_size), page.src_col_schema};
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +406,968 @@ template <typename T = uint8_t>
...
+  RowIndexIter page_row_index;
+  size_t const start_row;
+  size_t const end_row;
+
+  get_page_span(device_span<const size_type> _page_offsets,
+                RowIndexIter _page_row_index,
+                size_t _start_row,
+                size_t _end_row)
+    : page_offsets(_page_offsets),
+      page_row_index(_page_row_index),
```
vuule, line 522: probably completely irrelevant, but we can at least avoid one copy
```suggestion
      page_row_index(std::move(_page_row_index)),
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +406,968 @@ template <typename T = uint8_t>
...
+    }
+    // otherwise we have to look forward and choose a batch of pages
+
+    // as subpasses get decoded, the initial estimates we have for list row counts
+    // get updated with accurate data, so regenerate cumulative size info and row
+    // indices
+    rmm::device_uvector<cumulative_page_info> c_info(pass.pages.size(), _stream);
+    auto page_keys = make_page_key_iterator(pass.pages);
+    auto page_size = thrust::make_transform_iterator(pass.pages.d_begin(), get_page_input_size{});
+    thrust::inclusive_scan_by_key(rmm::exec_policy(_stream),
```
vuule, line 1271:
```suggestion
    thrust::inclusive_scan_by_key(rmm::exec_policy_nosync(_stream),
```

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -341,14 +406,968 @@ template <typename T = uint8_t>
...
+ * - For some reason, the sum total of the number of rows over all pages for a given column
+ *   is != N
+ *
+ * @param pages All pages to be decoded
+ * @param chunks Chunk data
+ * @param expected_row_count Expected row count, if applicable
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ */
+void detect_malformed_pages(cudf::detail::hostdevice_vector<PageInfo> const& pages,
+                            cudf::detail::hostdevice_vector<ColumnChunkDesc> const& chunks,
```
vuule, line 894:
```suggestion
void detect_malformed_pages(device_span<PageInfo const> pages,
                            device_span<ColumnChunkDesc const> chunks,
```

vuule (commented):
still not done, but getting close

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -415,185 +1442,101 @@ void reader::impl::compute_input_passes()
...
+      return row_group.num_rows;
+    });
+    _file_itm_data.input_pass_start_row_count.push_back(
+      std::reduce(rg_row_count, rg_row_count + row_groups_info.size()));
     return;
   }
 
   // generate passes. make sure to account for the case where a single row group doesn't fit within
   //
   std::size_t const read_limit =
```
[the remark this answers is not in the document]
nvdbaranec: Well. it's really just the "read limit". Ie, how many bytes we're going to read off disk.  99% of the time that corresponds to compressed data though.   `comp_read_limit` works.

cpp/src/io/parquet/reader_impl_chunking.cu
nvdbaranec: On an iterator? What's the motivation there? I haven't seen that before.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -380,6 +1399,13 @@ void reader::impl::create_global_chunk_info()
                         schema.converted_type,
                         schema.type_length);
 
+      // for lists, estimate the number of bytes per row. this is used by the subpass reader to
+      // determine where to split the decompression boundaries
+      float const list_bytes_per_row_est =
+        schema.max_repetition_level > 0 ? static_cast<float>(col_meta.total_uncompressed_size) /
+                                            static_cast<float>(row_group.num_rows)
```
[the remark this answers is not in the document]
nvdbaranec: Good question. I've never seen an empty row group, but I suppose it's theoretically possible.  I'll just treat that as 0 bytes per row.

cpp/src/io/parquet/reader_impl_chunking.cu
vuule: Iterators can be heavier than a simple pointer, depending on what the lambda captures. I don't think this is important, I just saw the double copy and chose not to discard the thought.

vuule: actually, yeah, I missed the point here. resolving.

cpp/src/io/parquet/reader_impl_chunking.cu
```diff
@@ -340,15 +409,967 @@ template <typename T = uint8_t>
...
+  // add to the cumulative_page_info data
+  rmm::device_uvector<size_t> d_temp_cost = cudf::detail::make_device_uvector_async(
+    temp_cost, stream, rmm::mr::get_current_device_resource());
+  auto iter = thrust::make_counting_iterator(size_t{0});
+  thrust::for_each(rmm::exec_policy(stream),
+                   iter,
+                   iter + pages.size(),
+                   [temp_cost = d_temp_cost.begin(), c_info = c_info.begin()] __device__(size_t i) {
+                     c_info[i].size_bytes += temp_cost[i];
+                   });
```
ttnghia, line 1064: Again, `d_temp_cost` is destroyed after this so we probably need to sync.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -386,231 +409,251 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
...
+    thrust::stable_sort_by_key(rmm::exec_policy_nosync(stream),
+                               page_keys.begin(),
+                               page_keys.end(),
+                               sort_indices.begin(),
+                               thrust::less<int>());
+    pass.pages = cudf::detail::hostdevice_vector<PageInfo>(
+      unsorted_pages.size(), unsorted_pages.size(), stream);
+    thrust::transform(rmm::exec_policy_nosync(stream),
+                      sort_indices.begin(),
+                      sort_indices.end(),
```
ttnghia, line 469: Same stream-order deallocation issue.

cpp/src/io/parquet/reader_impl_chunking.cu
ttnghia: Since we have local host buffer, we should really sync.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -386,231 +409,251 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
...
-  return decomp_pages;
+void reader::impl::build_string_dict_indices()
+{
+  auto& pass = *_pass_itm_data;
+
+  // compute number of indices per chunk and a summed total
+  rmm::device_uvector<size_t> str_dict_index_count(pass.chunks.size() + 1, _stream);
+  thrust::fill(
+    rmm::exec_policy(_stream), str_dict_index_count.begin(), str_dict_index_count.end(), 0);
+  thrust::for_each(rmm::exec_policy(_stream),
```
vuule, line 599:
```suggestion
  thrust::for_each(rmm::exec_policy_nosync(_stream),
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -386,231 +409,251 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
...
+                   pass.pages.end(),
+                   set_str_dict_index_count{str_dict_index_count, pass.chunks});
+
+  size_t const total_str_dict_indexes = thrust::reduce(
+    rmm::exec_policy(_stream), str_dict_index_count.begin(), str_dict_index_count.end());
+  if (total_str_dict_indexes == 0) { return; }
+
+  // convert to offsets
+  rmm::device_uvector<size_t>& str_dict_index_offsets = str_dict_index_count;
+  thrust::exclusive_scan(rmm::exec_policy(_stream),
```
vuule, line 610:
```suggestion
  thrust::exclusive_scan(rmm::exec_policy_nosync(_stream),
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -386,231 +409,251 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
...
+                         str_dict_index_offsets.begin(),
+                         0);
+
+  // allocate and distribute pointers
+  pass.str_dict_index = cudf::detail::make_zeroed_device_uvector_async<string_index_pair>(
+    total_str_dict_indexes, _stream, rmm::mr::get_current_device_resource());
+
+  auto iter = thrust::make_counting_iterator(0);
+  thrust::for_each(
+    rmm::exec_policy(_stream),
```
vuule, line 622:
```suggestion
    rmm::exec_policy_nosync(_stream),
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -767,13 +804,15 @@ std::pair<bool, std::vector<std::future<void>>> reader::impl::read_and_decompres
...
+  // TODO: make this respect the pass-wide skip_rows/num_rows instead of the file-wide
+  // skip_rows/num_rows
+  // auto remaining_rows            = num_rows;
   std::vector<std::future<void>> read_chunk_tasks;
   size_type chunk_count = 0;
   for (auto const& rg : row_groups_info) {
     auto const& row_group       = _metadata->get_row_group(rg.index, rg.source_index);
     auto const row_group_source = rg.source_index;
-    auto const row_group_rows   = std::min<int>(remaining_rows, row_group.num_rows);
+    // auto const row_group_rows   = std::min<int>(remaining_rows, row_group.num_rows);
```
vuule, line 815: should this commented out code be removed?

vuule (commented):
the last couple of nits.
thank you for leaving extensive comments, this would be unapproachable otherwise.

cpp/src/io/parquet/reader_impl_preprocess.cu
nvdbaranec: Yep.

ttnghia (approved):

vuule (approved):

PR #14363, opened 2023-11-06
Change `nullable()` to `has_nulls()` in `cudf::detail::gather`

divyegala:
## Description
In https://github.com/rapidsai/cudf/pull/13795, we found out that `nullable()` causes severe perf degradation for the nested-type case when the input is read from file via `cudf::io::read_json`. This is because the JSON reader adds a null mask for columns that don't have NULLs. This change is a no-overhead replacement that checks the actual null count instead of checking if a null mask is present.

This PR also solves a bug in quantile/median groupby where NULLs were being [set](https://github.com/rapidsai/cudf/blob/8deb3dd7573000e7d87f18a9e2bbe39cf2932e10/cpp/src/groupby/sort/group_quantiles.cu#L73) but the null count was not updated.

## Checklist


bdice:
If I have a column with an allocated null mask but no null values, and call `gather`, does this change whether the output column has an allocated null mask? If a null mask is allocated for the input, libcudf should preserve that property on the output even if there are no null values in the column. This is problematic for round-tripping data, as nullability of a column is a part of the schema/metadata even if there are no nulls declared.

divyegala:
@bdice good question, I'll try it out

divyegala:
Closing in favor of https://github.com/rapidsai/cudf/issues/14366

cpp/include/cudf/detail/gather.cuh
```diff
@@ -678,9 +679,24 @@ std::unique_ptr<table> gather(table_view const& source_table,
...
+      std::any_of(
+        source_table.begin(), source_table.end(), [](auto const& col) { return col.has_nulls(); });
+    if (has_nulls) {
+      auto const op = bounds_policy == out_of_bounds_policy::NULLIFY
+                        ? gather_bitmask_op::NULLIFY
+                        : gather_bitmask_op::DONT_CHECK;
+      gather_bitmask(source_table, gather_map_begin, destination_columns, op, stream, mr);
+    } else {
+      for (size_type i = 0; i < source_table.num_columns(); ++i) {
+        if (source_table.column(i).nullable() and not source_table.column(i).has_nulls()) {
```
bdice, line 693: I think in this `else` branch, you already know that `col.has_nulls()` is false for all columns, by the `std::any_of(...)`. Therefore, you should only have to check `col.nullable()` to know if you need an all-valid mask.
```suggestion
        if (source_table.column(i).nullable()) {
```

bdice (commented):
@divyegala I can check the failures you saw in CI after it runs.

cpp/include/cudf/detail/gather.cuh
divyegala: Ah right, good point. I will update this once we have finished this CI run's cpp tests

cpp/include/cudf/detail/gather.cuh
```diff
@@ -678,9 +679,24 @@ std::unique_ptr<table> gather(table_view const& source_table,
                           return col.nullable();
                         });
   if (nullable) {
-    auto const op = bounds_policy == out_of_bounds_policy::NULLIFY ? gather_bitmask_op::NULLIFY
-                                                                   : gather_bitmask_op::DONT_CHECK;
-    gather_bitmask(source_table, gather_map_begin, destination_columns, op, stream, mr);
+    auto const has_nulls =
+      bounds_policy == out_of_bounds_policy::NULLIFY ||
+      std::any_of(
+        source_table.begin(), source_table.end(), [](auto const& col) { return col.has_nulls(); });
```
vyasr, line 685: This actually already exists as [`cudf::has_nulls`](https://github.com/rapidsai/cudf/blob/branch-23.12/cpp/include/cudf/table/table_view.hpp#L318). However, I suspect that the reason that tests are failing is that what you really mean to check is whether there are nested nulls, which also exists as [`cudf::has_nested_nulls](https://github.com/rapidsai/cudf/blob/branch-23.12/cpp/include/cudf/table/table_view.hpp#L329)? That may at least account for the struct cases.

divyegala: That did solve `STRUCT_TEST` failures but these tests still fail for me locally:
```[  FAILED  ] JoinTest.Repro_StructsWithoutNullsPushedDown```

```[  FAILED  ] 24 tests, listed below:
[  FAILED  ] groupby_median_test/0.zero_valid_values, where TypeParam = signed char
[  FAILED  ] groupby_median_test/0.null_keys_and_values, where TypeParam = signed char
[  FAILED  ] groupby_median_test/1.zero_valid_values, where TypeParam = short
[  FAILED  ] groupby_median_test/1.null_keys_and_values, where TypeParam = short
[  FAILED  ] groupby_median_test/2.zero_valid_values, where TypeParam = int
[  FAILED  ] groupby_median_test/2.null_keys_and_values, where TypeParam = int
[  FAILED  ] groupby_median_test/3.zero_valid_values, where TypeParam = long
[  FAILED  ] groupby_median_test/3.null_keys_and_values, where TypeParam = long
[  FAILED  ] groupby_median_test/4.zero_valid_values, where TypeParam = float
[  FAILED  ] groupby_median_test/4.null_keys_and_values, where TypeParam = float
[  FAILED  ] groupby_median_test/5.zero_valid_values, where TypeParam = double
[  FAILED  ] groupby_median_test/5.null_keys_and_values, where TypeParam = double
[  FAILED  ] groupby_quantile_test/0.zero_valid_values, where TypeParam = signed char
[  FAILED  ] groupby_quantile_test/0.null_keys_and_values, where TypeParam = signed char
[  FAILED  ] groupby_quantile_test/1.zero_valid_values, where TypeParam = short
[  FAILED  ] groupby_quantile_test/1.null_keys_and_values, where TypeParam = short
[  FAILED  ] groupby_quantile_test/2.zero_valid_values, where TypeParam = int
[  FAILED  ] groupby_quantile_test/2.null_keys_and_values, where TypeParam = int
[  FAILED  ] groupby_quantile_test/3.zero_valid_values, where TypeParam = long
[  FAILED  ] groupby_quantile_test/3.null_keys_and_values, where TypeParam = long
[  FAILED  ] groupby_quantile_test/4.zero_valid_values, where TypeParam = float
[  FAILED  ] groupby_quantile_test/4.null_keys_and_values, where TypeParam = float
[  FAILED  ] groupby_quantile_test/5.zero_valid_values, where TypeParam = double
[  FAILED  ] groupby_quantile_test/5.null_keys_and_values, where TypeParam = double```

File: cpp/src/groupby/sort/group_quantiles.cu [hunk 3]
```diff
@@ -68,11 +69,13 @@ struct calculate_quantile_fn {
     thrust::for_each_n(thrust::seq,
                        thrust::make_counting_iterator(0),
                        num_quantiles,
-                       [d_result = d_result, segment_size, offset](size_type j) {
-                         if (segment_size == 0)
+                       [d_result = d_result, segment_size, offset, this](size_type j) {
+                         if (segment_size == 0) {
                            d_result.set_null(offset + j);
-                         else
+                           atomicAdd(this->null_count, 1);
```
bdice, line 75: Generally I would avoid atomics and compute this count in a separate kernel that works on the bitmask. I vaguely remember seeing anecdotal performance results on a similar type of thing, and I think the benchmarks favored a separate kernel instead of relying on atomics if there is any sizeable number of nulls. The case of all-null columns is too common to disregard here, in my opinion, and would have poor performance with every thread hammering the atomics.

vyasr: I agree that the alternative is likely faster for a meaningful number of nulls, but would prefer to see real benchmarks on how many nulls are required before making any changes here. This isn't the only place where we have this pattern. If we're going to change something we should scope this out more broadly.

bdice: I'm fine with that approach, if we're doing this same thing elsewhere. I haven't inventoried the libcudf code base to see other places where this is being done.

cpp/include/cudf/table/table_view.hpp
```diff
@@ -336,6 +336,22 @@ inline bool has_nested_nulls(table_view const& input)
   });
 }
 
+/**
+ * @brief Returns True if the table is nullable in any of its columns hierarchy
+ *
+ * @param input The table to check for nullables
+ * @return True if the table has nullables in any of its columns hierarchy, false otherwise
+ */
+inline bool has_nested_nullables(table_view const& input)
```
bdice, line 345: It's weird to name nullable columns "nullables" in my opinion. Can we rename to "nullable columns"?
```suggestion
/**
 * @brief Returns True if the table has a nullable column at any level of the column hierarchy
 *
 * @param input The table to check for nullable columns
 * @return True if the table has nullable columns at any level of the column hierarchy, false otherwise
 */
inline bool has_nested_nullable_columns(table_view const& input)
```

bdice (approved):
Approving with one name change suggestion.

vyasr (approved):
Not sure if my approval should count given that I've made significant modifications to this PR, but I did review it substantially and think the logic is sound.

cpp/include/cudf/table/table_view.hpp
```diff
@@ -336,6 +336,22 @@ inline bool has_nested_nulls(table_view const& input)
   });
 }
 
+/**
+ * @brief Returns True if the table has a nullable column at any level of the column hierarchy
+ *
+ * @param input The table to check for nullable columns
+ * @return True if the table has nullable columns at any level of the column hierarchy, false otherwise
+ */
+inline bool has_nested_nullable_columns(table_view const& input)
```
davidwendt, line 345: Same with this implementation. This could be moved to https://github.com/rapidsai/cudf/blob/branch-23.12/cpp/src/table/table.cpp
This applies to the other non-templated `inline` functions in this header as well so I would be ok if this was done in a follow up PR -- in 24.02 too.

davidwendt (changes requested):

cpp/include/cudf/detail/null_mask.hpp
```diff
@@ -259,6 +260,22 @@ cudf::size_type inplace_bitmask_and(device_span<bitmask_type> dest_mask,
...
+/**
+ * @brief Recursively set valid null masks for all children.
+ *
+ * This function applies all valid null masks to the output column if input column satisfies
+ * `nullable() == true` condition
+ *
+ * @param input input column
+ * @param output output column
+ * @param stream
+ * @param mr
```
bdice, line 272: Update these to have proper docstrings.

bdice (approved):
Approving up to some docstrings additions.

cpp/src/groupby/sort/group_quantiles.cu
```diff
@@ -49,6 +49,7 @@ struct calculate_quantile_fn {
   double const* d_quantiles;
   size_type num_quantiles;
   interpolation interpolation;
+  size_type* null_count;
```
ttnghia, line 52: This seems to be an unrelated change to this PR so ideally it should be in a separate PR. But I'm fine to keep this here but please clarify that in the PR description.

divyegala: Done

davidwendt (approved):

cpp/include/cudf/detail/gather.cuh
```diff
@@ -674,13 +674,20 @@ std::unique_ptr<table> gather(table_view const& source_table,
...
-                        std::any_of(source_table.begin(), source_table.end(), [](auto const& col) {
-                          return col.nullable();
-                        });
+                        cudf::has_nested_nullable_columns(source_table);
   if (nullable) {
-    auto const op = bounds_policy == out_of_bounds_policy::NULLIFY ? gather_bitmask_op::NULLIFY
-                                                                   : gather_bitmask_op::DONT_CHECK;
-    gather_bitmask(source_table, gather_map_begin, destination_columns, op, stream, mr);
+    auto const has_nulls =
+      bounds_policy == out_of_bounds_policy::NULLIFY || cudf::has_nested_nulls(source_table);
```
ttnghia, line 680: Should this uses `&&` instead? Because if we indeed don't have any nulls here then we don't need to call `gather_bitmask`.

ttnghia: Probably I misunderstood the usage of this variable. So this variable should be called `need_new_bitmask` or so. It should not be `has_nulls`.

divyegala: But we need to call `gather_bitmask` if `out_of_bounds_policy::NULLIFY`. `gather_bitmask` will help nullify any OOB accesses.

ttnghia: Okay, so `||` is indeed needed, but please rename that variable.

divyegala: Done

ttnghia (approved):

cpp/include/cudf/detail/gather.cuh
```diff
@@ -673,14 +673,20 @@ std::unique_ptr<table> gather(table_view const& source_table,
...
-                          return col.nullable();
-                        });
-  if (nullable) {
-    auto const op = bounds_policy == out_of_bounds_policy::NULLIFY ? gather_bitmask_op::NULLIFY
-                                                                   : gather_bitmask_op::DONT_CHECK;
-    gather_bitmask(source_table, gather_map_begin, destination_columns, op, stream, mr);
+  auto needs_new_bitmask = bounds_policy == out_of_bounds_policy::NULLIFY ||
+                           cudf::has_nested_nullable_columns(source_table);
+  if (needs_new_bitmask) {
+    needs_new_bitmask = needs_new_bitmask || cudf::has_nested_nulls(source_table);
```
ttnghia, line 679:
```suggestion
  auto const needs_new_bitmask = bounds_policy == out_of_bounds_policy::NULLIFY ||
                           cudf::has_nested_nullable_columns(source_table);
  if (needs_new_bitmask) {
    auto const has_nulls = cudf::has_nested_nulls(source_table);
```

divyegala: This is incorrect. We still need to check for `bounds_policy == out_of_bounds_policy::NULLIFY` in the inner condition

ttnghia: Sounds good.

bdice, line 679: Non-blocking suggestion:
```suggestion
    needs_new_bitmask |= cudf::has_nested_nulls(source_table);
```

cpp/include/cudf/table/table_view.hpp
bdice: Let's make an issue for this.

cpp/include/cudf/detail/gather.cuh
divyegala: We'll have to rerun CI again. Happy to make this change if you feel strongly about it, otherwise let's pass on it

cpp/include/cudf/table/table_view.hpp
divyegala: https://github.com/rapidsai/cudf/issues/14431

Issue #14325, opened 2023-10-25
[BUG] ORC writer produce wrong timestamp metrics which causes spark not to do predicate push down


thirtiseven:
**Describe the bug**
PR #13848 added minimum/maximum and minimumNanos/maximumNanos for ORC Writer timestamp statistics. It was intended to fix #13899 that Spark does not do predicate push down for gpu generated timestamp files. However the predicate push down test is still fails after above PR was merged, see https://github.com/NVIDIA/spark-rapids/issues/9075.

When trying to see the meta of related files with orc-tools, it throws `Exception in thread "main" java.lang.IllegalArgumentException: nanos > 999999999 or < 0`. And the `min` `max` are also mismatched with cpu-generated file with same data. I think it cause Spark to fail to do pushdown.

**Steps/Code to reproduce bug**

spark-shell with spark-rapids:
```
scala> import java.sql.{Date, Timestamp}
import java.sql.{Date, Timestamp}

scala> val timeString = "2015-08-20 14:57:00"
timeString: String = 2015-08-20 14:57:00

scala> val data = (0 until 10).map { i =>
     |           val milliseconds = Timestamp.valueOf(timeString).getTime + i * 3600
     |           Tuple1(new Timestamp(milliseconds))
     |         }
data: scala.collection.immutable.IndexedSeq[(java.sql.Timestamp,)] = Vector((2015-08-20 14:57:00.0,), (2015-08-20 14:57:03.6,), (2015-08-20 14:57:07.2,), (2015-08-20 14:57:10.8,), (2015-08-20 14:57:14.4,), (2015-08-20 14:57:18.0,), (2015-08-20 14:57:21.6,), (2015-08-20 14:57:25.2,), (2015-08-20 14:57:28.8,), (2015-08-20 14:57:32.4,))

scala> val df = spark.createDataFrame(data).toDF("a")
df: org.apache.spark.sql.DataFrame = [a: timestamp]

scala> df.write.orc("ORC_PPD_GPU")
```
orc-tools:
```
java -jar orc-tools-1.9.1-uber.jar meta ORC_PPD_GPU/
[main] WARN org.apache.hadoop.util.NativeCodeLoader - Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
log4j:WARN No appenders could be found for logger (org.apache.htrace.core.Tracer).
log4j:WARN Please initialize the log4j system properly.
[...]

PR #14367, opened 2023-11-07
Fix and disable encoding for nanosecond statistics in ORC writer

vuule:
## Description
Issue https://github.com/rapidsai/cudf/issues/14325

Use uint when reading/writing nano stats because nanoseconds have int32 encoding (different from both unit32 and sint32, _obviously_), which does not use zigzag. 
sint32 uses zigzag, and unit32 does not allow negative numbers, so we can use uint since we'll never have negative nanoseconds.

Also disabled the nanoseconds because it should only be written after ORC-135; we don't write the version so readers get confused if nanoseconds are there. Planning to re-enable once we start writing the version.

## Checklist


cpp/src/io/orc/orc.hpp
```diff
@@ -41,6 +41,9 @@ static constexpr uint32_t block_header_size = 3;
 // Seconds from January 1st, 1970 to January 1st, 2015
 static constexpr int64_t orc_utc_epoch = 1420070400;
 
+static constexpr int32_t DEFAULT_MIN_NANOS = 0;
+static constexpr int32_t DEFAULT_MAX_NANOS = 999999;
```
vyasr, line 45: Could we document why these are the min and max values? These aren't powers of 2 so I assume it's not integer size based, is it limited by some spec?

cpp/src/io/orc/stats_enc.cu
```diff
@@ -27,6 +27,8 @@ namespace cudf::io::orc::gpu {
 
 using strings::detail::fixed_point_string_size;
 
+constexpr bool enable_nanosecond_statistics = false;
```
vyasr, line 30: Could we link https://github.com/rapidsai/cudf/issues/14325 here as this flag is what needs to be be flipped to add the currently excluded statistic?

vyasr (approved):
A couple of requests for comments, otherwise the change looks fine insofar as it implements exactly what the description promises. Based on the discussion in #14325 it looks like this fix also stops the crash, which is the primary goal for 23.12

cpp/src/io/orc/orc.cpp
```diff
@@ -182,6 +182,9 @@ void ProtobufReader::read(timestamp_statistics& s, size_t maxlen)
                        field_reader(5, s.minimum_nanos),
                        field_reader(6, s.maximum_nanos));
   function_builder(s, maxlen, op);
+  // adjust nanoseconds because someone had a bright idea to store nanoseconds + 1 in the protobuf
+  if (s.minimum_nanos.has_value()) { --s.minimum_nanos.value(); }
+  if (s.maximum_nanos.has_value()) { --s.maximum_nanos.value(); }
```
ttnghia, line 187: Since we are using unsigned int, will this underflow?

cpp/src/io/orc/orc.hpp
ttnghia: Why don't we upgrade the type of these constants to `uint32_t`?

cpp/src/io/orc/orc.cpp
vuule: The nanoseconds are stored as `value+1`, so zero is not a valid content here. We found out the hard way about this :D

ttnghia: Can you add a comment to clarify that please? So we won't forget about it.

cpp/src/io/orc/orc.hpp
vuule: These do come from specs.
Nanosecond timestamp statistics are stored at millisecond precision + the leftover nanoseconds. This is why the max is 999999, one more and you have a full millisecond.

I can derive the max from `chrono`. Let me know.

Should we use unsigned here? AFAIK we should avoid any uints for any arithmetic.

vyasr: I'm fine using the signed ints, and I don't feel the need to use chrono. Just document that we store as ms + ns so 1e6 ns is the largest you'll ever need.

ttnghia: But will we have issue with sign vs unsign comparison?

vuule: The only place these are used is to compare against `min_ns_remainder`/`max_ns_remainder`, and these are signed.

ttnghia: Great.

ttnghia (approved):

cpp/src/io/orc/orc.cpp
ttnghia: Oh, approved but forgot about this :smile:

vuule: no worries, I haven't ;)

PR #14380, opened 2023-11-09
Ensure slow private attrs are maybe proxies

mroeschke:
## Description
Expected pandas test failures:

>  tests/indexing/test_indexing.py

Due to this PR, it appears an `assert something._values is something_else` fails more after this PR since `._values` wraps objects in an proxy object now (a known failure mode)

> tests/series/indexing/test_setitem.py

Runs into the issue where a test set up calls `proxy._values[key] = something` using a pandas helper function that isn't proxying correctly

## Checklist


python/cudf/cudf_pandas_tests/test_cudf_pandas.py
```diff
@@ -1078,6 +1078,13 @@ def test_dataframe_query():
     tm.assert_equal(actual, expected)
 
 
+def test_private_method_result_wrapped():
+    xoffset = xpd.offsets.Day()
+    dt = datetime.datetime(2020, 1, 1)
+    result = xoffset._apply(dt)
+    assert type(result) is xpd.Timestamp
```
bdice, line 1085: Is this semantically equivalent for this case?
```suggestion
    assert isinstance(result, xpd.Timestamp)
```

bdice (approved):
One suggestion, otherwise LGTM.

bdice:
@mroeschke Can you comment on the change in pandas test coverage / pass rate or any metrics like performance of particular code paths that change as a result of this PR?

python/cudf/cudf_pandas_tests/test_cudf_pandas.py
mroeschke: Yes I believe so (I am not sure why I didnt use this in the first place)

mroeschke:
> @mroeschke Can you comment on the change in pandas test coverage / pass rate or any metrics like performance of particular code paths that change as a result of this PR?

IIRC the net pandas test pass rate will decrease with this PR (down 1-2%), primarily due to the reasons in https://github.com/rapidsai/cudf/pull/14380#issue-1984622727

However, this PR will help proxy the result of `SlowObject._private_method_or_attribute` which before would always be a pandas object and not a proxy object.

bdice:
> IIRC the net pandas test pass rate will decrease with this PR (down 1-2%), primarily due to the reasons in [#14380 (comment)](https://github.com/rapidsai/cudf/pull/14380#issue-1984622727)

Aha! I did not understand that comment until now. Thank you for clarifying what that meant.

galipremsagar (approved):

PR #14381, opened 2023-11-09
Preserve DataFrame(columns=).columns dtype during empty-like construction

mroeschke:
## Description
`.column` used to always return `pd.Index([], dtype=object)` even if an empty-dtyped columns was passed into the DataFrame constructor e.g. `DatetimeIndex([])`. Needed to preserved some information about what column dtype was passed in so we can return a correctly type Index

## Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -734,6 +734,10 @@ def __init__(
                 rangeindex = isinstance(
                     columns, (range, pd.RangeIndex, cudf.RangeIndex)
                 )
+                if len(columns) == 0:
+                    empty_dtype = getattr(columns, "dtype", None)
+                else:
+                    empty_dtype = None
```
wence-, line 740: If the columns are non-empty and have a dtype, should we respect that? It feels like there are still times in which we might lose the dtype, or do I have it wrong?

wence- (approved):

python/cudf/cudf/core/dataframe.py
mroeschke: Good point. Currently a column passed in as `int8` would be returned as `int64`. I'll push an update to avoid this upcasting

mroeschke:
In a followup, I have a feeling `ColumnAccessor` should receive the user's passed `column` argument so we can centralize the logic whether a `MultiIndex/RangeIndex/non-default dtype Index` was passed

wence- (approved):

PR #14392, opened 2023-11-09
Limit DELTA_BINARY_PACKED encoder to the same number of bits as the physical type being encoded

etseidl:
## Description
The current implementation of the DELTA_BINARY_PACKED encoder can in certain circumstances use more bits for encoding than are present in the data. Specifically, for INT32 data, if the range of values is large enough, the encoder will use up to 33 bits. While not a violation of the Parquet specification, this does cause problems with certain parquet readers (see [this](https://github.com/apache/arrow/issues/20374) arrow-cpp issue, for instance). libcudf and parquet-mr have no issue with reading data encoded with 33 bits, but in the interest of a) greater interoperability and b) smaller files, this PR changes the DELTA_BINARY_PACKED encoder to use no more bits than are present in the physical type being encoded (32 for INT32, 64 for INT64).

The actual change made here is to perform all the delta computations using 32-bit integers for INT32 data, and 64-bit integers for INT64 data. The prior encoder used 64-bit integers exclusively. To deal with possible overflow, all math is done on unsigned integers to get defined wrapping behavior (overflow with signed integers is UB), with results cast back to signed afterwards. This is in line with the approach taken by parquet-mr, arrow-cpp, and arrow-rs.

## Checklist


GregoryKimball:
Thank you @etseidl for contributing this. (Part of #13501, to target 24.02 release)

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -128,9 +124,15 @@ inline __device__ void bitpack_mini_block(
 // Object used to turn a stream of integers into a DELTA_BINARY_PACKED stream. This takes as input
 // 128 values with validity at a time, saving them until there are enough values for a block
 // to be written.
-// T is the input data type (either zigzag128_t or uleb128_t).
+// T is the input data type (either int32_t or int64_t).
 template <typename T>
 class delta_binary_packer {
+ public:
+  using U            = std::make_unsigned_t<T>;
```
vuule, line 131: ❤️

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -164,9 +166,9 @@ class delta_binary_packer {
   }
 
   // Signed subtraction with defined wrapping behavior.
-  inline __device__ zigzag128_t subtract(zigzag128_t a, zigzag128_t b)
+  inline __device__ T subtract(T a, T b)
   {
-    return static_cast<zigzag128_t>(static_cast<uleb128_t>(a) - static_cast<uleb128_t>(b));
+    return static_cast<T>(static_cast<U>(a) - static_cast<U>(b));
```
vuule, line 171: oh my, I don't think I understood this correctly before.
So if `a` is smaller than `b`: unsigned result "underflows" and becomes ~max_uint.
And then what happens when it's cast back to signed?

etseidl: The point of all this is to get defined behavior when we overflow or underflow. Sadly in C++ signed over/underflow has undefined behavior, but unsigned overflow is well defined. So when we're calculating deltas, we'll switch to the unsigned domain to get the defined behavior, and then cast back to signed. 1 - 2 is 0xffffffff whether signed or unsigned, and we get -1. INT_MIN - 1 is undefined in signed, but 0x7fffffff in unsigned (although every compiler I've tried yields 0x7fffffff for signed anyway).

vuule: diabolical.

vuule (approved):
Looks good, if a bit odd :D
Can you please add a note in the description about the 2s complement math going on with this change?

cpp/src/io/parquet/delta_enc.cuh
```diff
@@ -225,27 +227,26 @@ class delta_binary_packer {
...
 
     // Get max normalized delta for each warp, and use that to determine how many bits to use
     // for the bitpacking of this warp.
-    zigzag128_t const warp_max =
-      delta::warp_reduce(_warp_tmp[warp_id]).Reduce(norm_delta, cub::Max());
+    U const warp_max = warp_reduce(_warp_tmp[warp_id]).Reduce(norm_delta, cub::Max());
     __syncwarp();
 
-    if (lane_id == 0) { _mb_bits[warp_id] = sizeof(zigzag128_t) * 8 - __clzll(warp_max); }
+    if (lane_id == 0) { _mb_bits[warp_id] = sizeof(long long) * 8 - __clzll(warp_max); }
```
PointKernel, line 249: Wanted to suggest `sizeof(int64_t)` instead of `sizeof(long long)` since it's considered [a good C++ style](https://softwareengineering.stackexchange.com/questions/317670/does-a-long-ban-make-sense/317672#317672). Then I realized `long long` is more consistent with the use of `__clzll`. Never mind then.

PointKernel (approved):
LGTM

cpp/src/io/parquet/delta_enc.cuh
etseidl: Exactly why I chose `long long`. Previously I was using the wrong type, and that bit me when I made the change to `T/U` typing :sweat_smile:

hyperbolic2346 (approved):
Thank you for doing this!

vuule:
@ttnghia should we wait on your review before merging? I see that you self-requested a review.

ttnghia:
I'm fine to merge it now. If it is still here then I can take a look later tonight.

vuule:
> Don't we need a python review?

Well, NOW we do! :D 

Thanks for the reminder (I don't see github requiring the python review). I'll ping folks tomorrow :)

bdice (approved):
Python approval. 🐍 ✅

PR #14411, opened 2023-11-14
Ensure nvbench initializes nvml context when built statically

robertmaynard:
## Description
Port https://github.com/NVIDIA/nvbench/pull/148 to cudf so that nvbench benchmarks work now that we always use a static version of nvbench.

## Checklist


bdice (commented):
@robertmaynard Do you know why this nvbench we run in our CI tests doesn't seem to catch the problem? https://github.com/rapidsai/cudf/blob/b446a6f187241e765c925da1053ece2679313a06/ci/test_cpp.sh#L37-L41

robertmaynard:
> @robertmaynard Do you know why this nvbench we run in our CI tests doesn't seem to catch the problem?
> 
> https://github.com/rapidsai/cudf/blob/b446a6f187241e765c925da1053ece2679313a06/ci/test_cpp.sh#L37-L41

I presume that our CI build env doesn't have nvml, so therefore nvbench doesn't enable the feature and we avoid the issue.

bdice (approved):

PR #14421, opened 2023-11-15
REF: Remove instances of pd.core

mroeschke:
## Description
`pandas.core` is technically private and methods could be moved at any time. Avoiding places in the codepace where they could be avoided

## Checklist


bdice (approved):
LGTM. Let's target new PRs to 24.02 since we're in burndown, unless you think this is essential for 23.12.

mroeschke:
Ah OK. Not essential for 23.12 so will retarget for 24.02

python/cudf/cudf/core/index.py
```diff
@@ -334,13 +334,16 @@ def __contains__(self, item):
...
         try:
-            item = pd.core.dtypes.common.ensure_python_int(item)
-        except TypeError:
+            int_item = int(item)
+            assert int_item == item
+        except (TypeError, AssertionError):
             return False
-        if not item % 1 == 0:
+        if not int_item % 1 == 0:
             return False
```
wence-, line 345: **suggestion:** ‏
```suggestion
```

This is an unreachable branch since $\\{x: x \in \mathbb{Z} \land x \mod 1 \ne 0\\} = \emptyset$

python/cudf/cudf/core/index.py
```diff
@@ -334,13 +334,16 @@ def __contains__(self, item):
...
-        return item in range(self._start, self._stop, self._step)
+        return int_item in range(self._start, self._stop, self._step)
```
wence-, line 346:
```suggestion
        return int_item in self._range
```

python/cudf/cudf/core/index.py
```diff
@@ -334,13 +334,16 @@ def __contains__(self, item):
...
             return False
+        elif isinstance(item, bool):
+            return False
         try:
-            item = pd.core.dtypes.common.ensure_python_int(item)
-        except TypeError:
+            int_item = int(item)
+            assert int_item == item
+        except (TypeError, AssertionError):
             return False
```
wence-, line 343: Please don't use exceptions for control-flow (not least `python -O` will remove the `assert` line):

I think, once the boolean case has been handled, the whole body of the function can be:
```
try:
    int_item = int(item)
    return int_item == item and int_item in self._range
except (ValueError, OverflowError):
    return False
```

The `ValueError` handles `nan`, the `OverflowError` handles `inf`.

So the body could be:
```
if isinstance(item, bool) or not isinstance(item, tuple(np.sctypes["int"] + np.sctypes["float"] + [int, float])
    return False
try:
    int_item = int(item)
    return int_item == item and int_item in self._range
except (ValueError, OverflowError):
    return False
```

python/cudf/cudf/tests/test_rolling.py
```diff
@@ -536,10 +536,9 @@ def get_window_bounds(
     "indexer",
     [
         pd.api.indexers.FixedForwardWindowIndexer(window_size=2),
-        pd.core.window.expanding.ExpandingIndexer(),
-        pd.core.window.indexers.FixedWindowIndexer(window_size=3)
-        if PANDAS_LT_140
-        else pd.core.indexers.objects.FixedWindowIndexer(window_size=3),
+        pd.api.indexers.VariableOffsetWindowIndexer(
+            index=pd.date_range("2020", periods=5), offset=pd.offsets.BDay(1)
+        ),
```
wence-, line 541: **question:** ‏This _seems_ like it is testing something different? What am I missing? (I don't know a lot about the rolling indexers)

wence- (changes requested):

python/cudf/cudf/core/dataframe.py
```diff
@@ -2310,13 +2310,14 @@ def to_dict(
...
-            into_c = pd.core.common.standardize_mapping(into)
-            return into_c((k, v) for k, v in self.items())
-
-        return self.to_pandas().to_dict(orient=orient, into=into)
+            # Ensure values are cudf.Series
+            converted = ((k, Series(v)) for k, v in result.items())
+            if isinstance(into, defaultdict):
+                return type(result)(into.default_factory, converted)
+            return type(result)(converted)
+        return result
```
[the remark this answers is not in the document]
mroeschke: > now produces both a DtoH and HtoD copy.

Oof thanks for the catch! Any tips on generally knowing when a DtoH or HtoD occurs?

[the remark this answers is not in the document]
wence-: Any time you call `to_pandas` or `to_arrow` or `.values_host` or similar to get a host-side version of the object. Then in the other direction, any time you have a host-side object.

python/cudf/cudf/tests/test_rolling.py
mroeschke: This is testing _slightly_ different but using a more public API.

Before I presume this was meant to test that rolling was compatible with rolling window indexer objects, and this test was testing private indexer objects (used in `df.expanding` for example). I switched this to test the public indexer objects

wence- (approved):

PR #14428, opened 2023-11-16
Match pandas join ordering obligations in pandas-compatible mode

### LINKED ISSUE #14001: [BUG] DataFrame.join(Series) does not maintain order of Series values in mode.pandas_compatible
**Describe the bug**
DataFrame.join(Series) does not maintain order of Series values

```python
In [12]: import cudf

In [13]: df = cudf.DataFrame([1, 2])

In [16]: ser = cudf.Series(list("abcdef"), index=[0, 0, 0, 1, 1, 1], name="var1")

In [17]: df.join(ser)
Out[17]: 
   0 var1
0  1    a
0  1    c
0  1    b
1  2    d
1  2    e
1  2    f

In [18]: df.to_pandas().join(ser.to_pandas())
Out[18]: 
   0 var1
0  1    a
0  1    b
0  1    c
1  2    d
1  2    e
1  2    f
```

**Expected behavior**
I would expect the order of the joined values to be maintained.

   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**

**Additional context**

wence-:
## Description

If we pass sort=True to merges we are on the hook to sort the result in order with respect to the key columns. If those key columns have repeated values there is still some space for ambiguity. Currently we get a result back whose order (for the repeated key values) is determined by the gather map that libcudf returns for the join. This does not come with any ordering guarantees.

When sort=False, pandas has join-type dependent ordering guarantees which we also do not match. To fix this, in pandas-compatible mode only, reorder the gather maps according to the order of the input keys. When sort=False this means that our result matches pandas ordering. When sort=True, it ensures that (if we use a stable sort) the tie-break for equal sort keys is the input dataframe order.

While we're here, switch from argsort + gather to sort_by_key when sorting results.

- Closes #14001

## Checklist


wence-:
Note that there is still one case where we don't match pandas, but I think that is a pandas bug: https://github.com/pandas-dev/pandas/issues/55992

python/cudf/cudf/core/join/join.py
```diff
@@ -160,6 +199,55 @@ def __init__(
...
+        lengths = [len(left_cols[0]), len(right_cols[0])]
+        # Only nullify those maps that need it.
+        nullify = [
+            self.how not in {"inner", "left"},
+            self.how not in {"inner", "right"},
+        ]
+        # To reorder maps so that they are in order of the input
+        # tables, we gather from iota on both right and left, and then
+        # sort the gather maps with those two columns as key.
+        key_order = list(
```
shwina, line 232: Is `key_order` always of length `2`?

wence-: Yes

shwina (approved):
LGTM!

bdice:
Can you share benchmarks?

Do we still need to propose moving this logic to libcudf, if the performance is reduced? It sounds like @shwina is approving but I thought we were wanting to confirm performance impact first.

shwina:
I think we still want to move to libcudf in the medium term. The perf impact measured here seems _OK_ for the short term: https://github.com/rapidsai/cudf-private/pull/111#issuecomment-1779677674

@beckernick would you agree?

bdice:
Great. I'm good with merging this in the short term and then revisiting moving this logic to libcudf. Ideally that could land in 24.02.

bdice (approved):
Nice. I like the tests -- very thorough!

PR #14432, opened 2023-11-16
cudf.pandas: Fix accessor registration for proxy types

shwina:
## Description

Fixes how we handle accessor registration. Currently, our `register_datafrane_accessor` accessor is a proxy function which will simply register an accessor on the fast (`cudf.DataFrame`) or slow (`pandas.DataFrame`) type, instead of on the proxy `cudf.pandas._wrappers.pandas.DataFame`) type.

To register accessors with the actual proxy type, this PR achieves this by passing the actual proxy type to the internal pandas function (`_register_accessor`). That's not an _ideal_ solution but appears to work. We might want to consider vendoring this function instead. Happy to hear @mroeschke's thoughts here.

## Checklist


mroeschke:
> That's not an ideal solution but appears to work. We might want to consider vendoring this function instead.

Curious to know why you think this might not be the ideal solution. It seems like with this solution a user's accessor code would be able to be accelerated since it will see a proxy object which is what we want?

python/cudf/cudf_pandas_tests/test_cudf_pandas.py
```diff
@@ -1226,3 +1226,21 @@ def test_concat_fast():
...
+
+
+def test_register_accessor():
+    @xpd.api.extensions.register_dataframe_accessor("xyz")
+    class XYZ:
+        def __init__(self, obj):
+            self._obj = obj
+
+        @property
+        def foo(self):
```
mroeschke, line 1238: Would be good to have this test a function that should go through the fast path and a function that should go through like the slow path like

```
def should_be_fast(self):
    return self._obj.dropna()

def should_be_slow(self):
    return self._obj.apply(lambda x: 5)

...

assert df.xyz.should_be_fast ==
assert df.xyz.should_be_slow ==
```

vyasr:
Closing, #14534 makes this obsolete.

PR #14437, opened 2023-11-16
Testing stream pool implementation

shrshi:
## Description
The goal of this PR is to create a global pool containing only the cudf test stream which is to be used all stream tests that invoke `fork_streams` in their execution path. The stream pool is constructed by `create_global_cuda_stream_pool()` in `identify_stream_usage.cpp`, and overrides the function implementation in `utilities/stream_pool.cpp` when preloaded. 

The test checks for only `cudaLaunchKernel` being invoked with the wrong stream.

## Checklist


cpp/tests/streams/pool_test.cu
```diff
@@ -0,0 +1,32 @@
...
+#include <cudf_test/base_fixture.hpp>
+#include <cudf_test/default_stream.hpp>
+
+class PoolTest : public cudf::test::BaseFixture {};
+
+__global__ void do_nothing_kernel() {}
+
+TEST_F(PoolTest, ForkStreams)
+{
+  auto streams = cudf::detail::fork_streams(cudf::test::get_default_stream(), 4);
```
harrism, line 29: **❓ question:** ‏ Why 4 streams when only 2 are used?

cpp/tests/streams/pool_test.cu
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+#include <cudf/detail/utilities/stream_pool.hpp>
+#include <rmm/cuda_stream_view.hpp>
+
+#include <cudf_test/base_fixture.hpp>
+#include <cudf_test/default_stream.hpp>
+
+class PoolTest : public cudf::test::BaseFixture {};
```
harrism, line 23: **💡 suggestion:** ‏ Since "Pool" is overloaded, perhaps name this more specifically.
```suggestion
class StreamPoolTest : public cudf::test::BaseFixture {};
```

cpp/tests/utilities/identify_stream_usage.cpp
```diff
@@ -66,6 +67,34 @@ rmm::cuda_stream_view const get_default_stream()
 }  // namespace test
```
harrism, line 67: **🚩 issue:** ‏ The comment above refers to `STREAM_MODE_TESTING=ON` and `STREAM_MODE_TESTING=OFF`, but the code above and the new code you added both use `#ifdef STREAM_MODE_TESTING`. In other words, the code doesn't consider the value but only whether or not `STREAM_MODE_TESTING` is defined. I think the comment above should be rewritten to reflect reality.

cpp/tests/utilities/identify_stream_usage.cpp
```diff
@@ -66,6 +67,34 @@ rmm::cuda_stream_view const get_default_stream()
...
+
+  std::vector<rmm::cuda_stream_view> get_streams(std::size_t count) override
+  {
+    return std::vector<rmm::cuda_stream_view>(count, cudf::test::get_default_stream());
+  }
+
+  std::size_t get_stream_pool_size() const override { return 1UL; }
+};
+
+cuda_stream_pool* create_global_cuda_stream_pool() { return new test_cuda_stream_pool(); }
```
harrism, line 93: **🚩 issue:** ‏ When is this function called rather than the one implemented in `stream_pool.cpp`? Is the other one undefined if `STREAM_MODE_TESTING` is defined? Can you clarify the documentation? It would actually be nice to have them in the same file if this is indeed how it works. Otherwise I worry about hard-to-diagnose multiple definition linker errors. It's a little confusing that the class `test_cuda_stream_pool` is defined in a file called `identify_stream_usage.cpp`.

harrism (changes requested):
The code looks fine, but I have a little bit of confusion that I would like to clear up. Thanks!

cpp/tests/utilities/identify_stream_usage.cpp
vyasr: This is my fault. When I originally wrote that comment I had assigned values, but I ended up just writing the CMake to conditionally define the variable. I agree that updating the comment would be good.

cpp/tests/utilities/identify_stream_usage.cpp
```diff
@@ -66,6 +67,34 @@ rmm::cuda_stream_view const get_default_stream()
...
+namespace detail {
+
+/**
+ * @brief Implementation of `cuda_stream_pool` that always returns the
+ * `cudf::test::get_default_stream()`
+ */
+class test_cuda_stream_pool : public cuda_stream_pool {
+ public:
+  rmm::cuda_stream_view get_stream() override { return cudf::test::get_default_stream(); }
+  rmm::cuda_stream_view get_stream(stream_id_type stream_id) override
```
vyasr, line 80: Can we remove the variable name since it's never used (or if that causes unforeseen problems, mark it as `[[ maybe_unused ]]`)?
```suggestion
  rmm::cuda_stream_view get_stream(stream_id_type) override
```

cpp/tests/utilities/identify_stream_usage.cpp
```diff
@@ -66,6 +67,34 @@ rmm::cuda_stream_view const get_default_stream()
...
+  {
+    return cudf::test::get_default_stream();
+  }
+
+  std::vector<rmm::cuda_stream_view> get_streams(std::size_t count) override
+  {
+    return std::vector<rmm::cuda_stream_view>(count, cudf::test::get_default_stream());
+  }
+
+  std::size_t get_stream_pool_size() const override { return 1UL; }
```
vyasr, line 90: Hmm I'm not sure how exactly this function is meant to be used. Is it coupled with calls to `get_streams`? If so, then really the size is infinite since we can request an arbitrary size subset of the pool and get that many copies of the same stream. Is returning 1 here consistent with those semantics?

cpp/tests/utilities/identify_stream_usage.cpp
vyasr: Everything in this file needs to be in this file. It is compiled and linked into a separate library that is then preloaded (i.e. via `LD_PRELOAD`) in certain contexts, which allows functions defined here to override the ones defined in libcudf proper. There's more information at the bottom of the testing dev guide page: https://docs.rapids.ai/api/libcudf/stable/md_developer_guide_testing.

On that note, It would be nice to update the dev guide to mention the existence of the stream pool as well, maybe just a single sentence noting that pools are also overridden by this testing stream pool.

vyasr (changes requested):

cpp/tests/utilities/identify_stream_usage.cpp
harrism: OK, maybe a file-level comment explaining what this is. If this is a separate library, perhaps putting it in `cpp/tests/utilities/` is not the best?

cpp/tests/utilities/identify_stream_usage.cpp
shrshi: The comment above the `get_streams` declaration in `include/cudf/detail/utilities/stream_pool.hpp` suggests that `get_streams(int n)` returns $n$ unique streams from the pool if $n \leq$ `get_stream_pool_size()`. I returned 1 here since the test stream pool has one unique stream.

cpp/include/cudf/detail/utilities/stream_pool.hpp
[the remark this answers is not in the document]
shrshi: Yes, I have moved the parent class declaration from `stream_pool.cpp` to the header file so that `test_cuda_stream_pool` in `identify_stream_usage.cpp` can include it.

cpp/tests/utilities/identify_stream_usage.cpp
shrshi: Updated the comments for the stream mode testing magic :)

cpp/include/cudf/detail/utilities/stream_pool.hpp
```diff
@@ -25,6 +25,62 @@
...
+   *
+   * This function is thread safe with respect to other calls to the same function.
+   *
+   * @param count The number of stream views to return.
+   * @return Vector containing `count` stream views.
+   */
+  virtual std::vector<rmm::cuda_stream_view> get_streams(std::size_t count) = 0;
+
+  /**
+   * @brief Get the number of stream objects in the pool.
```
vyasr, line 70:
```suggestion
   * @brief Get the number of unique stream objects in the pool.
```

cpp/tests/utilities/identify_stream_usage.cpp
vyasr: I see, yes that makes sense. I made a small suggestion there to make the uniqueness explicit, otherwise LGTM.

cpp/tests/utilities/identify_stream_usage.cpp
vyasr: Yeah we could certainly work on improving that. I can look into this for a follow-up PR.

vyasr (approved):

cpp/include/cudf/detail/utilities/stream_pool.hpp
vuule: FWIW, they are not unique in the debug pool

cpp/include/cudf/detail/utilities/stream_pool.hpp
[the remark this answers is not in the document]
vuule: Notes from an offline discussion - `cuda_stream_pool` was "hidden" in the source file on purpose; we want the pool usage to be limited to `fork_streams`/`join_streams`. The change here exposes more of the stream pool than we'd like. 
However, we don't see a better solution, since the current approach at least does not require additional APIs in libcudf.

cpp/tests/streams/pool_test.cu
```diff
@@ -0,0 +1,32 @@
...
+#include <rmm/cuda_stream_view.hpp>
+
+#include <cudf_test/base_fixture.hpp>
+#include <cudf_test/default_stream.hpp>
+
+class StreamPoolTest : public cudf::test::BaseFixture {};
+
+__global__ void do_nothing_kernel() {}
+
+TEST_F(StreamPoolTest, ForkStreams)
```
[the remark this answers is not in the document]
shrshi: Yes, the test verifies if `cudaLaunchKernel` is being invoked with `test::get_default_stream` for all streams in the pool.

cpp/tests/streams/pool_test.cu
```diff
@@ -0,0 +1,32 @@
...
+{
+  auto streams = cudf::detail::fork_streams(cudf::test::get_default_stream(), 2);
+  do_nothing_kernel<<<1, 32, 0, streams[0].value()>>>();
+  do_nothing_kernel<<<1, 32, 0, streams[1].value()>>>();
```
vuule, line 31: Inspired by "for all streams" in the comment above :)
```suggestion
  for (auto& stream : streams){
    do_nothing_kernel<<<1, 32, 0, stream.value()>>>();
  }
```

cpp/include/cudf/detail/utilities/stream_pool.hpp
```diff
@@ -25,6 +25,62 @@
...
+  virtual std::vector<rmm::cuda_stream_view> get_streams(std::size_t count) = 0;
+
+  /**
+   * @brief Get the number of unique stream objects in the pool.
+   *
+   * This function is thread safe with respect to other calls to the same function.
+   *
+   * @return the number of stream objects in the pool
+   */
+  virtual std::size_t get_stream_pool_size() const = 0;
```
ttnghia, line 76: The name is a bit redundant. We know that this is a stream pool.
```suggestion
  virtual std::size_t get_pool_size() const = 0;
```
or just `size()`.

cpp/tests/streams/pool_test.cu
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cudf/detail/utilities/stream_pool.hpp>
+#include <rmm/cuda_stream_view.hpp>
```
ttnghia, line 18:
```suggestion

#include <rmm/cuda_stream_view.hpp>
```

cpp/tests/streams/pool_test.cu
```diff
@@ -0,0 +1,33 @@
...
+
+#include <cudf_test/base_fixture.hpp>
+#include <cudf_test/default_stream.hpp>
```
ttnghia, line 21: Group of `cudf_test` headers should be on top of the file.

cpp/tests/streams/pool_test.cu
```diff
@@ -0,0 +1,33 @@
...
+
+class StreamPoolTest : public cudf::test::BaseFixture {};
+
+__global__ void do_nothing_kernel() {}
```
ttnghia, line 25: I am not sure if this kernel without any code will be optimized out so the for loop below will never be executed?

cpp/include/cudf/detail/utilities/stream_pool.hpp
etseidl: `get_pool_size()` is consistent with `rmm::cuda_stream_pool`. Surprised this wasn't caught before.

cpp/tests/streams/pool_test.cu
shrshi: Thanks for pointing this out! I printed the GPU trace from the nsys profiler to check this and it shows that the `do_nothing_kernel` is being executed.

vuule (approved):

cpp/include/cudf/detail/utilities/stream_pool.hpp
vuule: yeah, should be `get_pool_size`. Not a change for this PR, IMO.

ttnghia: Okay, but then please address it in a follow up work.

ttnghia (approved):

cpp/include/cudf/detail/utilities/stream_pool.hpp
shrshi: Thank you for pointing this out! I'll work on this change in a follow-up PR.

harrism: Why isn't this a change for this PR, if all the code in this section is new in this PR?

harrism (approved):
Nice work.

PR #14438, opened 2023-11-17
REF: Remove **kwargs from to_pandas, raise if nullable is not implemented

mroeschke:
## Description
* Remove unnecessary **kwargs from signature
* Typing and improve error message if nullable is not implemented

## Checklist


python/cudf/cudf/core/_base_index.py
```diff
@@ -850,7 +850,7 @@ def notna(self):
         """
         raise NotImplementedError
 
-    def to_pandas(self, nullable=False):
+    def to_pandas(self, nullable: bool = False):
```
wence-, line 853: **suggestion:** I think we should make all of these arguments keyword-only
```
def to_pandas(self, *, nullable: bool = False):
  ...
```

Rationale: API safety in the case when one can pass more than just `nullable` as an argument. WDYT?

python/cudf/cudf/core/column/column.py
```diff
@@ -203,17 +203,19 @@ def __repr__(self):
...
         """
         # This default implementation does not handle nulls in any meaningful
-        # way, but must consume the parameter to avoid passing it to PyArrow
-        # (which does not recognize it).
-        kwargs.pop("nullable", None)
-        pd_series = self.to_arrow().to_pandas(**kwargs)
+        # way
+        if nullable:
+            raise NotImplementedError(f"{nullable=} is not implemented.")
+        pd_series = self.to_arrow().to_pandas()
```
wence-, line 218: **note:** In theory this removes some ability to pass kwargs into the `to_pandas` call, but I doubt anyone was actually doing that.

python/cudf/cudf/core/_base_index.py
mroeschke: Agreed. Does that mean I should label this PR as `breaking`?

bdice: This is probably an API break, yes.

python/cudf/cudf/io/json.py
```diff
@@ -228,7 +235,14 @@ def to_json(
...
         warnings.warn("Using CPU via Pandas to write JSON dataset")
-        pd_value = cudf_val.to_pandas(nullable=True)
+        if isinstance(cudf_val, cudf.DataFrame):
+            pd_data = {
+                col: maybe_return_nullable_pd_obj(series)
+                for col, series in cudf_val.items()
+            }
+            pd_value = pd.DataFrame(pd_data)
+        else:
+            pd_value = maybe_return_nullable_pd_obj(cudf_val)
```
wence-, line 245: This is just handling that `nullable=True` was previously _silently_ ignored for `to_pandas` calls that didn't support it, and now it is loud?

wence-:
One test `test_list_to_pandas_nullable_true` still needs handling but otherwise looks good to go.

PR #14443, opened 2023-11-17
Fix default stream use in the CSV reader

vuule:
## Description
String scalars, used to replace double quotes, are implicitly created from `std::string`s, using the default stream. In addition, these scalars are created for each string column.

This PR changes how these objects are created to avoid redundant H2D copies and the use of the default stream.

## Checklist


vuule:
CC @shrshi

shrshi:
Looks good to me. Thank you!

shrshi (approved):

karthikeyann (approved):

PR #14453, opened 2023-11-20
Include encode type in the error message when unsupported Parquet encoding is detected

ZelboK:
## Description
Per #14209 this will list out unsupported encodings that were found.

## Checklist


ZelboK:
> @ZelboK Could you run `pre-commit`? See instructions in the Contributing Guide. I think this PR is going to fail C++ style checks from clang-format in its current state. https://github.com/rapidsai/cudf/blob/branch-23.12/CONTRIBUTING.md#using-pre-commit-hooks

Will do, thanks for mentioning!

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -82,6 +82,18 @@ constexpr void set_error(int32_t error, int32_t* error_code)
   }
 }
 
+/**
+ * @brief Enum for the unsupported encodings
+ *
+ * These values are used as bitmasks, so they must be powers of 2.
+ */
+enum class unsupported_encoding_error : int8_t {
```
etseidl, line 90: Rather than using this enum, you can use the `encoding_to_mask()` function defined below, coupled with `is_supported_encoding()` above. Then as encodings are added, there a fewer things to change.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -305,8 +365,15 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
   DecodePageHeaders(chunks.device_ptr(), chunks.size(), error_code.data(), stream);
 
   if (error_code.value() != 0) {
-    // TODO(ets): if an unsupported encoding was detected, do extra work to figure out which one
-    CUDF_FAIL("Parquet header parsing failed with code(s)" + error_code.str());
+    int32_t mask              = static_cast<int32_t>(decode_error::UNSUPPORTED_ENCODING);
+    bool unsupported_encoding = error_code.value() & mask;
+    if (unsupported_encoding) {
```
etseidl, line 370:
```suggestion
    if (BitAnd(error_code.value(), decode_error::UNSUPPORTED_ENCODING) != 0) {
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -280,6 +280,66 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+        ? static_cast<int8_t>(unsupported_encoding_error::GROUP_VAR_INT)
+      : page.encoding == Encoding::BIT_PACKED
+        ? static_cast<int8_t>(unsupported_encoding_error::BIT_PACKED)
+      : page.encoding == Encoding::DELTA_LENGTH_BYTE_ARRAY
+        ? static_cast<int8_t>(unsupported_encoding_error::DELTA_LENGTH_BYTE_ARRAY)
+      : page.encoding == Encoding::BYTE_STREAM_SPLIT
+        ? static_cast<int8_t>(unsupported_encoding_error::BYTE_STREAM_SPLIT)
+        : 0x0;
+    return encoding_value;
+  };
```
etseidl, line 333:
```suggestion
  auto to_mask = [] __device__(auto const& page) {
    return is_supported_encoding(page.encoding) ? 0U : encoding_to_mask(page.encoding);
  };
```
Then just change the template param for the `thrust::bit_or`.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -280,6 +280,66 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
   return total_pages;
 }
 
+/**
+ * @brief Helper function to convert an encoding bitmask to a readable string
+ *
+ * @param bitmask Bitmask of found unsupported encodings
+ * @returns Human readable string with unsupported encodings
+ */
+std::string encoding_bitmask_to_str(int8_t encoding_bitmask)
```
etseidl, line 289: You might want to loop over the set bits in `encoding_bitmask`, otherwise this won't catch new encodings that might be added to the Parquet spec in the future.

etseidl (changes requested):
Thanks for this. Just a few suggestions :)

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -280,6 +280,63 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+    };  
+    int8_t unsupported = thrust::transform_reduce(
+        rmm::exec_policy(stream),
+        pages.begin(),
+        pages.end(),
+        to_mask,
+        static_cast<int8_t>(0),
+        thrust::bit_or<int8_t>()
+    );
+    return encoding_bitmask_to_str(unsupported);
```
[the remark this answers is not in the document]
etseidl: @ZelboK the python tests have a [test](https://github.com/rapidsai/cudf/blob/723c565f7a03e3e9a842526cd4cc94bcf6f582e5/python/cudf/cudf/tests/test_parquet.py#L2942) for detecting the unsupported encoding, but that won't check the output string. You could run the test with the `-s` option to capture stdout and visually verify it's working as intended.

cpp/src/io/parquet/reader_impl_preprocess.cu
ZelboK: Sounds good! Tysm for your prompt replies :), will push changes shortly. I wasn't aware of the utility lying around so this is really helpful. Makes code a lot cleaner

etseidl: > I wasn't aware of the utility lying around

If I had a nickel for every time some kind soul pointed out a utility function I had missed... :moneybag:

cpp/src/io/parquet/reader_impl_preprocess.cu
ZelboK: So this will give an SFINAE error likely because `error_code` is not a scoped enumeration. I can add a new overload to deal with this, but I don't have enough context on the codebase to know whether or not this is okay to do. Could you comment? @etseidl

ZelboK: So `BitAnd` won't work because it'll return a SFINAE error since the kernel error does not return a scoped enumeration. I can add another overload, but the comments in `BitAnd` clearly say it's only for scoped enumerations so I hesitate to do so. 

I imagine the kernel_error can also be modified, too, but again I don't have enough context on the codebase to know which is appropriate. (Or maybe there's a way to map it into an enumeration that im unaware of?)

ZelboK:
Question: I can't use `BitAnd` because the newly added `kernel_error` class's `value()` does not return a scoped enumeration. `BitAnd` uses SFINAE to only allow for scoped enums to use it.
I can:
1) Add new overload for `BitAnd`
2) Modify `kernel_error` to use scoped enumerations? (Might entail a lot more effort)
3) Maybe im forgetting some simple way to map the types properly. I'm still kinda new to C++ so i apologize.
4) Just not use BitAnd  and do a regular bitwise operation.

I hesitate doing 1, or 2, because I don't have enough context for the codebase to know the friction it may cause. I imagine there's some reason `BitAnd` is constrained on scoped enums, and I don't have a good visual on how much work is required for 2.

cpp/src/io/parquet/reader_impl_preprocess.cu
etseidl: Ah, probably should have made error_code an unsigned int :( You could just cast `error_code.value()` to a `uint32_t` for now. Might be worth revisiting the `kernel_error` design later.

ZelboK: SG! Thank you, will push changes shortly.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -280,6 +280,70 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+    case Encoding::GROUP_VAR_INT: return "GROUP_VAR_INT";
+    case Encoding::PLAIN_DICTIONARY: return "PLAIN_DICTIONARY";
+    case Encoding::RLE: return "RLE";
+    case Encoding::BIT_PACKED: return "BIT_PACKED";
+    case Encoding::DELTA_BINARY_PACKED: return "DELTA_BINARY_PACKED";
+    case Encoding::DELTA_LENGTH_BYTE_ARRAY: return "DELTA_LENGTH_BYTE_ARRAY";
+    case Encoding::DELTA_BYTE_ARRAY: return "DELTA_BYTE_ARRAY";
+    case Encoding::RLE_DICTIONARY: return "RLE_DICTIONARY";
+    case Encoding::BYTE_STREAM_SPLIT: return "BYTE_STREAM_SPLIT";
+    case Encoding::NUM_ENCODINGS: return "NUM_ENCODINGS";
```
etseidl, line 302:
```suggestion
    case Encoding::NUM_ENCODINGS:
```
You should just let this last case fall through, since `NUM_ENCODINGS` is an enum hack to get at the upper bound, not an actual valid encoding type. :wink:

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -280,6 +280,70 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+    default: return "UNKNOWN";
```
etseidl, line 303: Could this be something like `"UNKNOWN(" + std::to_string(encoding) + ")" (probably needs a cast to int)?

etseidl (commented):
Looking good, thanks for the quick changes :) A few more minor suggestions.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -280,6 +280,70 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+/**
+ * @brief Helper function to convert an encoding bitmask to a readable string
+ *
+ * @param bitmask Bitmask of found unsupported encodings
+ * @returns Human readable string with unsupported encodings
+ */
+std::string encoding_bitmask_to_str(int32_t encoding_bitmask)
+{
+  std::string result;
+  for (int32_t i = 0; i < 32; ++i) {
```
etseidl, line 316: Lost a comment :( I was wondering if this could use `bitset<uint32_t>`. Then the loop could use `bitset.size()` as the upper bound, and you could use `bitset.test(i)` to check bit positions. Removes one magic number :)

ZelboK: Sounds good! I'm a scala programmer by day so I might miss some things as I rarely get to program at a low level 😅. Will change now.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -280,6 +280,70 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+ * @brief Create a readable string for the user that will list out all unsupported encodings found.
+ *
+ * @param pages List of page information
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @returns Human readable string with unsupported encodings
+ */
+std::string list_unsupported_encodings(const cudf::detail::hostdevice_vector<PageInfo>& pages,
+                                       rmm::cuda_stream_view stream)
+{
+  auto to_mask = [] __device__(const PageInfo& page) -> int8_t {
```
etseidl, line 335:
```suggestion
  auto to_mask = [] __device__(const PageInfo& page) {
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -280,6 +280,70 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+    return is_supported_encoding(page.encoding) ? 0U : encoding_to_mask(page.encoding);
+  };
+  int32_t unsupported = thrust::transform_reduce(rmm::exec_policy(stream),
+                                                 pages.begin(),
+                                                 pages.end(),
+                                                 to_mask,
+                                                 static_cast<int32_t>(0),
+                                                 thrust::bit_or<int32_t>());
```
etseidl, line 343:
```suggestion
                                                 0U),
                                                 thrust::bit_or<uint32_t>());
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -280,6 +280,70 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @returns Human readable string with unsupported encodings
+ */
+std::string list_unsupported_encodings(const cudf::detail::hostdevice_vector<PageInfo>& pages,
+                                       rmm::cuda_stream_view stream)
+{
+  auto to_mask = [] __device__(const PageInfo& page) -> int8_t {
+    return is_supported_encoding(page.encoding) ? 0U : encoding_to_mask(page.encoding);
+  };
+  int32_t unsupported = thrust::transform_reduce(rmm::exec_policy(stream),
```
etseidl, line 338:
```suggestion
  uint32_t const unsupported = thrust::transform_reduce(rmm::exec_policy(stream),
```
I too wish C++ had Scala's `val` vs `var` :sweat_smile:

etseidl (commented):
Sorry, a few more suggestions.

cpp/src/io/parquet/reader_impl_preprocess.cu
ZelboK: Ahhh my bad. I context switch too much so the habits from both languages don't translate well😂. All good. I appreciate the review! :)

GregoryKimball:
Thank you @ZelboK and @etseidl for iterating on this!

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -305,8 +365,14 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
   DecodePageHeaders(chunks.device_ptr(), chunks.size(), error_code.data(), stream);
 
   if (error_code.value() != 0) {
-    // TODO(ets): if an unsupported encoding was detected, do extra work to figure out which one
-    CUDF_FAIL("Parquet header parsing failed with code(s)" + error_code.str());
+    uint32_t error = static_cast<uint32_t>(error_code.value());
+    if (BitAnd(error, decode_error::UNSUPPORTED_ENCODING) != 0) {
```
etseidl, line 369:
```suggestion
    if (BitAnd(error_code.value(), decode_error::UNSUPPORTED_ENCODING) != 0) {
```
Now that #14464 has merged, this should work now.

GregoryKimball:
@ZelboK is this work complete?

ZelboK:
@GregoryKimball Ah right. Let me push this up and resolve the conversations. I still want to test it via the pytests like @etseidl suggested. I was about to do this but I recall I had some issue with my environment on setting them up. I should have time later today or tomorrow to try that again. But the code/work is complete, I just want to do a sanity check.

etseidl:
This looks good from my perspective. I think it's ready for review. Thanks @ZelboK

ZelboK:
Unfortunately running into problems setting up cudf python with the devcontainers even. Maybe someone else can pull in the branch(that has cudf python setup properly) and can verify against the output to see if it is working?

When I first pushed this PR upstream, I did verify beforehand it was working by "artificially" introducing an error by modifying the code in `gpuDecodePageData` in `page_data.cu` (hopefully I'm remembering the names properly) and looking at the output in the unit tests.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -282,6 +283,67 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+ * @returns Human readable string with unsupported encodings
+ */
+std::string list_unsupported_encodings(const cudf::detail::hostdevice_vector<PageInfo>& pages,
+                                       rmm::cuda_stream_view stream)
+{
+  auto to_mask = [] __device__(const PageInfo& page) {
+    return is_supported_encoding(page.encoding) ? 0U : encoding_to_mask(page.encoding);
+  };
+  uint32_t const unsupported = thrust::transform_reduce(
+    rmm::exec_policy(stream), pages.begin(), pages.end(), to_mask, 0U, thrust::bit_or<uint32_t>());
```
etseidl, line 343:
```suggestion
    rmm::exec_policy(stream), pages.d_begin(), pages.d_end(), to_mask, 0U, thrust::bit_or<uint32_t>());
```

etseidl (commented):
I was not getting the list of bad encodings. One problem was using the device iterators, but another issue is that `count_page_headers` will also receive the same error code, but the pages vector doesn't exist yet so we can't use that to get the set of bad encodings. I'd suggest testing for `UNSUPPORTED_ENCODING` in the error handling for `count_page_headers()` and not FAIL if that flag is set.

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -282,6 +283,67 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+ * @returns Human readable string with unsupported encodings
+ */
+std::string encoding_bitmask_to_str(int32_t encoding_bitmask)
+{
+  std::bitset<32> bits(encoding_bitmask);
+  std::string result;
+
+  for (size_t i = 0; i < bits.size(); ++i) {
+    if (bits.test(i)) {
+      Encoding current = static_cast<Encoding>(i);
```
ttnghia, line 323:
```suggestion
      auto const current = static_cast<Encoding>(i);
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -282,6 +283,67 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+ * @brief Create a readable string for the user that will list out all unsupported encodings found.
+ *
+ * @param pages List of page information
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @returns Human readable string with unsupported encodings
+ */
+std::string list_unsupported_encodings(const cudf::detail::hostdevice_vector<PageInfo>& pages,
+                                       rmm::cuda_stream_view stream)
+{
+  auto to_mask = [] __device__(const PageInfo& page) {
```
ttnghia, line 339: Need to wrap this in `cuda::proclaim_return_type`. See https://github.com/rapidsai/cudf/pull/14577

ttnghia, line 339:
```suggestion
  auto to_mask = [] __device__(auto const& page) {
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -282,6 +283,67 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+  return result;
+}
+/**
+ * @brief Create a readable string for the user that will list out all unsupported encodings found.
+ *
+ * @param pages List of page information
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @returns Human readable string with unsupported encodings
+ */
+std::string list_unsupported_encodings(const cudf::detail::hostdevice_vector<PageInfo>& pages,
```
ttnghia, line 336:
```suggestion
std::string list_unsupported_encodings(cudf::detail::hostdevice_vector<PageInfo> const& pages,
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -307,8 +369,13 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
   DecodePageHeaders(chunks.device_ptr(), chunks.size(), error_code.data(), stream);
 
   if (error_code.value() != 0) {
-    // TODO(ets): if an unsupported encoding was detected, do extra work to figure out which one
-    CUDF_FAIL("Parquet header parsing failed with code(s)" + error_code.str());
+    if (BitAnd(error_code.value(), decode_error::UNSUPPORTED_ENCODING) != 0) {
+      auto unsupported =
```
ttnghia, line 373:
```suggestion
      auto const unsupported_str =
```

ZelboK:
> I was not getting the list of bad encodings. One problem was using the device iterators, but another issue is that `count_page_headers` will also receive the same error code, but the pages vector doesn't exist yet so we can't use that to get the set of bad encodings. I'd suggest testing for `UNSUPPORTED_ENCODING` in the error handling for `count_page_headers()` and not FAIL if that flag is set.

Is there any way I can test this behavior without having to rely on cudf python? Unfortunately I've run into a few blockers trying to get it set up. I'd like to finish this PR but it wouldn't be too efficient for me to offload the testing part to you like this. 

I should be able to create unsupported encodings in a unit test I hope?

etseidl:
> Is there any way I can test this behavior without having to rely on cudf python?

You could write a small C++ prog that calls `cudf::io::read_parquet()`. Compiling it might be an adventure :sweat_smile:

ZelboK:
> > Is there any way I can test this behavior without having to rely on cudf python?
> 
> You could write a small C++ prog that calls `cudf::io::read_parquet()`. Compiling it might be an adventure 😅

Hm. Curious to know - when it comes to cuIO will I often need to run things from cudf python to test out functionality? If so that would justify me putting more effort into getting cudf python to work. I just have a lot more motivation for getting more experience with GPU programming than I do with tooling so I put it off lol. I'll spend a couple hours trying to figure this out

(btw this is the problem i get when trying to run `pytest`, forgive me, I don't program in python often 😅 ). Builds just fine, tests jut don't start up for me. 

```(cudf_dev) ksm@Kashimo:~/work/improving/cudf/python/cudf$ pytest -v
======================== test session starts ========================
platform linux -- Python 3.10.13, pytest-7.4.3, pluggy-1.3.0 -- /home/ksm/anaconda3/envs/cudf_dev/bin/python3.10
cachedir: .pytest_cache
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase(PosixPath('/home/ksm/work/improving/cudf/python/cudf/.hypothesis/examples'))
rootdir: /home/ksm/work/improving/cudf/python/cudf
plugins: benchmark-4.0.0, cases-3.8.1, anyio-4.1.0, cov-4.1.0, hypothesis-6.92.0, xdist-3.5.0
collected 0 items / 1 error                                         

============================== ERRORS ===============================
___________________ ERROR collecting test session ___________________
/home/ksm/anaconda3/envs/cudf_dev/lib/python3.10/site-packages/_pytest/config/__init__.py:641: in _importconftest
    mod = import_path(conftestpath, mode=importmode, root=rootpath)
/home/ksm/anaconda3/envs/cudf_dev/lib/python3.10/site-packages/_pytest/pathlib.py:567: in import_path
    importlib.import_module(module_name)
/home/ksm/anaconda3/envs/cudf_dev/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1050: in _gcd_import
    ???
<frozen importlib._bootstrap>:1027: in _find_and_load
    ???
<frozen importlib._bootstrap>:1006: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:688: in _load_unlocked
    ???
/home/ksm/anaconda3/envs/cudf_dev/lib/python3.10/site-packages/_pytest/assertion/rewrite.py:186: in exec_module
    exec(co, module.__dict__)
benchmarks/conftest.py:59: in <module>
    from config import cudf  # noqa: W0611, E402, F401
benchmarks/common/config.py:46: in <module>
    import cudf  # noqa: W0611, F401
cudf/__init__.py:9: in <module>
    _setup_numba()
cudf/utils/_numba.py:122: in _setup_numba
    ptx_toolkit_version = _get_cuda_version_from_ptx_file(CC_60_PTX_FILE)
cudf/utils/_numba.py:157: in _get_cuda_version_from_ptx_file
    with open(path) as ptx_file:
E   FileNotFoundError: [Errno 2] No such file or directory: '/home/ksm/work/improving/cudf/python/cudf/cudf/utils/../core/udf/shim_60.ptx'
====================== short test summary info ======================
ERROR  - FileNotFoundError: [Errno 2] No such file or directory: '/home/k...
!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!
========================= 1 error in 0.77s ==========================
(cudf_dev) ksm@Kashimo:~/work/improving/cudf/python/cudf$
```

DanialJavady96:
> I was not getting the list of bad encodings. One problem was using the device iterators, but another issue is that `count_page_headers` will also receive the same error code, but the pages vector doesn't exist yet so we can't use that to get the set of bad encodings. I'd suggest testing for `UNSUPPORTED_ENCODING` in the error handling for `count_page_headers()` and not FAIL if that flag is set.

Might be misunderstanding, can you clarify a few things?
Just wondering, `count_page_headers()` without the pages vector will not be able to give anything past a general message no? Are you saying to relocate the logic in `decode_page_headers` over in `count_page_headers()`? I can use `CUDF_LOG_ERROR` to give a general message to the user. 

Or are you saying that this logic
```
   if (error_code.value() != 0) {
    CUDF_FAIL("Parquet header parsing failed with code(s) " + error_code.str());
   }
  ```
 is impacting the logic in `decode_page_headers()` because that shouldn't be the case

**Comment by vuule** [comment]
I _think_ @etseidl 's idea is to completely ignore UNSUPPORTED_ENCODING error in count_page_headers so we can get to print the unsupported encoding types after DecodePageHeaders, where it will be reported again.

File: cpp/src/io/parquet/reader_impl_preprocess.cu [hunk 18]
```diff
@@ -307,8 +369,13 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
   DecodePageHeaders(chunks.device_ptr(), chunks.size(), error_code.data(), stream);
 
   if (error_code.value() != 0) {
-    // TODO(ets): if an unsupported encoding was detected, do extra work to figure out which one
-    CUDF_FAIL("Parquet header parsing failed with code(s)" + error_code.str());
+    if (BitAnd(error_code.value(), decode_error::UNSUPPORTED_ENCODING) != 0) {
+      auto const unsupported_str =
+        ". With unsupported encodings found: " + list_unsupported_encodings(pages, stream);
+      CUDF_FAIL("Parquet header parsing failed with code(s) " + error_code.str() + unsupported);
```
vuule, line 375:
```suggestion
      CUDF_FAIL("Parquet header parsing failed with code(s) " + error_code.str() + unsupported_str);
```

cpp/src/io/parquet/reader_impl_preprocess.cu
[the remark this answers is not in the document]
DanialJavady96: > For line 275:
> 
> ```
>   if (error_code.value() != 0 and
>       error_code.value() != static_cast<uint32_t>(decode_error::UNSUPPORTED_ENCODING)) {
> ```
> 
> With this change I'm getting the error printed out in `test_parquet_reader_unsupported_page_encoding`. I think we want to check equality here (rather than BitAnd) so we don't ignore other errors in `DecodePageHeaders`, if multiple happen at once.

Strange. For some reason even when I have `count_page_headers()` ignore `UNSUPPORTED_ENCODINGS`, it still does not print. The way it's set up should definitely print it though. Passing the unsupported encodings to `CUDF_FAIL` does not seem to print anything, but if I use `CUDF_LOG_ERROR` it will.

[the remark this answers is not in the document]
etseidl: Oh, sorry, I thought you weren't using the python test. I think the error is being eaten somewhere because the failure is expected. I only see the error if I read the test file with c++.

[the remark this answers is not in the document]
DanialJavady96: > Oh, sorry, I thought you weren't using the python test. I think the error is being eaten somewhere because the failure is expected. I only see the error if I read the test file with c++.

Ah sorry about that, I actually managed to fix the python tests last night, edited my comment yesterday instead of a new one. I apologize for overloading you with too much information 😂. So in that case this work should be complete, correct?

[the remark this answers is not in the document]
DanialJavady96: Also, sorry just a heads up I was asked to contribute to open source using my companys account this week over my personal one. That might've been confusing. 

Also curious to know how you read the test file with C++?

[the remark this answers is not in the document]
vuule: I deleted line 2942 in the Python Parquet test file to see the error output

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -271,8 +272,10 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
   DecodePageHeaders(chunks.device_ptr(), chunks.size(), error_code.data(), stream);
   chunks.device_to_host_sync(stream);
 
-  if (error_code.value() != 0) {
-    CUDF_FAIL("Parquet header parsing failed with code(s) " + error_code.str());
+  if (error_code.value() != 0 and
```
etseidl, line 275: Might want to add a comment here to explain why we're ignoring unsupported encoding errors. Maybe tag this PR in the comment too.

etseidl (commented):
One more small request :sweat_smile: 

Then I think it's good to go.

cpp/src/io/parquet/reader_impl_preprocess.cu
DanialJavady96: Good point, will push that

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -282,6 +289,71 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+  }
+}
+
+/**
+ * @brief Helper function to convert an encoding bitmask to a readable string
+ *
+ * @param bitmask Bitmask of found unsupported encodings
+ * @returns Human readable string with unsupported encodings
+ */
+std::string encoding_bitmask_to_str(int32_t encoding_bitmask)
```
vuule, line 322:
```suggestion
[[nodiscard]] std::string encoding_bitmask_to_str(uint32_t encoding_bitmask)
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -282,6 +289,71 @@ void generate_depth_remappings(std::map<int, std::pair<std::vector<int>, std::ve
...
+  return result;
+}
+/**
+ * @brief Create a readable string for the user that will list out all unsupported encodings found.
+ *
+ * @param pages List of page information
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @returns Human readable string with unsupported encodings
+ */
+std::string list_unsupported_encodings(cudf::detail::hostdevice_vector<PageInfo> const& pages,
```
vuule, line 342: No need to pass `hostdevice_vector`, only the device buffer is used.
```suggestion
[[nodiscard]] std::string list_unsupported_encodings(device_span<PageInfo const> pages,
```

cpp/src/io/parquet/reader_impl_preprocess.cu
```diff
@@ -307,8 +379,13 @@ int decode_page_headers(cudf::detail::hostdevice_vector<ColumnChunkDesc>& chunks
   DecodePageHeaders(chunks.device_ptr(), chunks.size(), error_code.data(), stream);
 
   if (error_code.value() != 0) {
-    // TODO(ets): if an unsupported encoding was detected, do extra work to figure out which one
-    CUDF_FAIL("Parquet header parsing failed with code(s)" + error_code.str());
+    if (BitAnd(error_code.value(), decode_error::UNSUPPORTED_ENCODING) == 0x4) {
```
vuule, line 382: I don't think we need to compare with exact (hardcoded value) here
```suggestion
    if (BitAnd(error_code.value(), decode_error::UNSUPPORTED_ENCODING) != 0) {
```

vuule (changes requested):
Thank you for the patch, looks good!
Few small suggestions

cpp/src/io/parquet/reader_impl_preprocess.cu
DanialJavady96: RIP, I did this to test something and forgot to change it back. Mb

ttnghia (approved):

vuule (approved):

etseidl (approved):
All my nits have been picked :smile:. LGTM

PR #14458, opened 2023-11-20
Include writer code and writerVersion in ORC files

vuule:
## Description
Closes https://github.com/rapidsai/cudf/issues/14325
Changes some of the metadata written to ORC file:

- Include the (cuDF) writer code (5).
- include writerVersion, with the value of 7; This value means that bugs up to ORC-517 are fixed. This version (6+ required) allows us to write the nanosecond statistics.

This change can have unexpected impact, depending on how readers use these fields.

## Checklist


cpp/src/io/orc/orc.hpp
```diff
@@ -40,6 +40,9 @@ namespace orc {
 static constexpr uint32_t block_header_size = 3;
 // Seconds from January 1st, 1970 to January 1st, 2015
 static constexpr int64_t orc_utc_epoch = 1420070400;
+// Each ORC writer implementation should write its code to the file footer
+// the codes are specified in the ORC specification
+static constexpr int32_t cudf_writer_code = 5;
```
davidwendt, line 45: We are ORC writer number 5? 
Is there a link to these codes?

harrism:
I think someone more familiar with ORC should review. Removing my assignment.

revans2 (approved):
This looks good to me.

cpp/src/io/orc/writer_impl.cu
```diff
@@ -2702,7 +2703,8 @@ void writer::impl::close()
   ps.footerLength         = pbw.size();
   ps.compression          = _compression_kind;
   ps.compressionBlockSize = _compression_blocksize;
-  ps.version              = {0, 12};
+  ps.version              = {0, 12};  // Hive 0.12
+  ps.writerVersion        = 7;        // ORC-517 fixed
```
hyperbolic2346, line 2707: should this `7` be defined above like `MAGIC` with a comment about it? Do we change the writer version every major release or major changes? Is this something we never change?

vuule: My understanding is that we would increase this version when we implement newer features and fix notable bugs, see https://github.com/apache/orc/blob/main/proto/orc_proto.proto#L412-L422

I'm not sure if this should be a constant. I'll make the change if you think that's better.

hyperbolic2346: If it is a constant it is easier to add a multi-line comment explaining that major changes should increment it. I'd also be ok with a comment about the version. At the extreme, I would have a changelist in the comments
```
// version 1: Initial implementation
// version 2: added nested types
// version 3: fixed bug xyz
```

vuule: Included the list of versions, let me know what you think

cpp/src/io/orc/orc.hpp
```diff
@@ -40,6 +40,28 @@ namespace orc {
...
 static constexpr int64_t orc_utc_epoch = 1420070400;
+// ORC datasets start with a 3 byte header
+static constexpr char const* MAGIC = "ORC";
+
+// Each ORC writer implementation should write its code to the file footer
+// the codes are specified in the ORC specification
+static constexpr int32_t cudf_writer_code = 5;
+// Each ORC writer implementation should write its version to the PostScript
+// The version values are based on the ORC Java writer bug fixes and features
+// From https://github.com/apache/orc/blob/main/proto/orc_proto.proto:
```
ttnghia, line 51: The link cannot be accessed.

cpp/src/io/orc/orc.hpp
```diff
@@ -48,12 +70,13 @@ static constexpr int32_t DEFAULT_MIN_NANOS = 0;
...
-  uint32_t compressionBlockSize{};     // the maximum size of each compression chunk
-  std::vector<uint32_t> version;       // the version of the writer [major, minor]
-  uint64_t metadataLength = 0;         // the length of the metadata section in bytes
-  std::string magic       = "";        // the fixed string "ORC"
+  uint64_t footerLength       = 0;        // the length of the footer section in bytes
+  CompressionKind compression = NONE;     // the kind of generic compression used
+  uint32_t compressionBlockSize{};        // the maximum size of each compression chunk
+  std::vector<uint32_t> version;          // the version of the file format [major, minor]
+  uint64_t metadataLength = 0;            // the length of the metadata section in bytes
+  std::optional<uint32_t> writerVersion;  // The version of the writer that wrote the file
```
ttnghia, line 78: Why this is an optional? Should it be a permanent field?

cpp/src/io/orc/orc.hpp
vuule: yeah, they removed the proto module in a recent PR :(
I'll track down where the information is now.

cpp/src/io/orc/orc.hpp
vuule: it's based on the protobuf specs: https://github.com/apache/orc/blob/branch-1.9/proto/orc_proto.proto#L439
We might always write the version now, but we should be able to read files without this field.

cpp/src/io/orc/orc.hpp
vuule: fixed the link
what timing!

hyperbolic2346 (approved):

PR #14470, opened 2023-11-22
Avoid DataFrame conversion in `MultiIndex.from_pandas`

mroeschke:
## Description

This uncovered a bug where `MultiIndex.levels` could be a list of Series instead of a list of Index.

## Checklist


python/cudf/cudf/core/multiindex.py
```diff
@@ -481,7 +476,8 @@ def copy(
 
         mi = MultiIndex._from_data(self._data.copy(deep=deep))
         if self._levels is not None:
-            mi._levels = [s.copy(deep) for s in self._levels]
+            # TODO: why is deep=True not being respected here
+            mi._levels = [idx.copy(deep) for idx in self._levels]
```
mroeschke, line 480: @wence- could use some guidance here regarding the failing test below: `test_multiindex_copy_deep` for the `deep=True` case

Before `self._levels` was a list of `Series`, so it was performing a `Series.copy(True)` so the pointer checking in the test was passing. With this PR, now `self._levels` is a list of `Index`, and a `Index.copy(True)` is not passing the pointer checking in the test.

`deep=True` looks to be passed down to `Column.copy` and friends so I am not sure why the pointer in this case is not changing

wence-: I'll take a look, thanks

wence-:
```suggestion
            mi._levels = [idx.copy(deep=deep) for idx in self._levels]
```

`idx.copy` takes a bunch more arguments (same as pandas) than just `deep=True/False`.

python/cudf/cudf/core/multiindex.py
```diff
@@ -178,18 +179,12 @@ def __init__(
...
-                    index=range(-1, len(levels[i])),
-                )
-            else:
-                level = cudf.DataFrame({column_name: levels[i]})
-
-            source_data[column_name] = libcudf.copying.gather(
-                [level._data[column_name]], col
-            )[0]
+        for (column_name, col), level in zip(codes._data.items(), levels):
+            result_col = level._column.take(col)
```
wence-, line 183: On line 175 we already do partial bounds-checking. The treatment of `-1` is AFAICT undocumented pandas behaviour. That is `codes` do not do normal indexing but instead `-1` is for "missing value".

But let us move all of the validation into one place:
```
import cudf._lib as libcudf
from cudf._lib.types import size_type_dtype
import numpy as np

lo, hi = libcudf.reduce.minmax(code)
if lo.value < -1 or hi.value > len(level) - 1:
    raise ValueError(...)
if lo.value == -1:
    # Now we can gather and insert null automatically
    code[code == -1] = np.iinfo(size_type_dtype).min
result_col, = libcudf.copying.gather([level], code, nullify=True)
```

python/cudf/cudf/core/multiindex.py
```diff
@@ -158,7 +159,7 @@ def __init__(
                     "codes and is inconsistent!"
```
wence-, line 159: FWIW, I am not sure that `codes` can be `np.int64` it should probably be size_type_dtype. But let us not fix that here.

python/cudf/cudf/core/multiindex.py
mroeschke: 🤦 thanks for pointing out my oversight!

python/cudf/cudf/core/multiindex.py
mroeschke: Nice find. Was able to consolidate the codes validate in this loop

python/cudf/cudf/tests/test_multiindex.py
```diff
@@ -2079,3 +2079,20 @@ def test_multiindex_eq_other_multiindex():
...
+    [
+        cudf.MultiIndex.from_product([[0, 1], [1, 0]]),
+        cudf.MultiIndex.from_tuples([(0, 1), (0, 0), (1, 1), (1, 0)]),
+        # TODO: Implement from_arrays like pandas
+        # cudf.MultiIndex.from_arrays([0, 0, 1, 1], [1, 0, 1, 0]),
+        cudf.MultiIndex(
+            levels=[[0, 1], [0, 1]], codes=[[0, 0, 1, 1], [1, 0, 1, 0]]
+        ),
+    ],
+)
```
wence-, line 2095: **issue** This parameterisation now constructs cudf objects at test collection time. Although we are not good at obeying these rules we, we prefer not to do this (see https://docs.rapids.ai/api/cudf/nightly/developer_guide/testing/#parametrization-custom-fixtures-and-pytest-mark-parametrize)

So preference is either: 
- hide these values behind a lambda `lambda: ...` which you just call inside the test 
- or (preferable) use a fixture, which is a little more setup work but can be reused:

```
@pytest.fixture(params=["product", "tuples", pytest.param("arrays", marks=pytest.mark.xfail(reason="from_arrays not yet implemented")), "levels-codes"])
def midx(request);
    if request.param == ...:
        return ...
    elif ...
```

Not also the pre-emptive marking of the arrays case as `xfail` rather than just commenting out with TODO.

mroeschke: Ah I didn't know about that guidance. I converted this into a fixture instead

python/cudf/cudf/tests/test_dropna.py
```diff
@@ -256,19 +256,21 @@ def test_dropna_index(data, dtype):
...
-        # pandas-gh44792. Pandas infers the dtypes as (int64, int64), though
-        # int64 doesn't really store null/nans. The dtype propagates to the
-        # result of dropna. cuDF infers the dtypes as (float, float), which
-        # differs from pandas.
-        assert_eq(expect, got)
+    if how == "all" and "data0" in request.node.callspec.id:
+        request.applymarker(
+            pytest.mark.xfail(
+                reason="pandas NA value np.nan results in float type. "
+                "cuDF correctly retains int type"
```
wence-, line 270:
```suggestion
                "cuDF correctly retains int type "
                "(https://github.com/pandas-dev/pandas/issues/44792)"
```

wence- (approved):
Minor suggestion to retain the pandas issue reference in the xfail, otherwise looks great, thanks!

galipremsagar:
@mroeschke can we drop the `REF: ` prefix from the PR title? A while ago we decided not to have prefixes in the PR titles as they will be added to the commit history.

python/cudf/cudf/core/multiindex.py
```diff
@@ -170,26 +173,24 @@ def __init__(
...
+        for (column_name, code), level in zip(codes._data.items(), levels):
+            if len(code):
+                lo, hi = libcudf.reduce.minmax(code)
+                if lo.value < -1 or hi.value > len(level) - 1:
+                    raise ValueError(
+                        f"Codes must be -1 <= codes <= {len(level) - 1}"
+                    )
+                if lo.value == -1:
+                    # Now we can gather and insert null automatically
+                    code[code == -1] = np.iinfo(size_type_dtype).min
```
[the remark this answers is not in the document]
wence-: https://github.com/rapidsai/cudf/issues/14511

Issue #14200, opened 2023-09-26
[BUG] Unexpected errors for groupby agg on empty dask-cudf dataframe


charlesbluca:
**Describe the bug**
I'm encountering unexpected `DataErrors` and `RuntimeErrors` when attempting groupby aggregations on an empty dask-cudf dataframe that I do not encounter with 23.08; it's not immediately obvious if this change was intentional, as these generally seem like groupby aggs that should be somewhat trivial.

**Steps/Code to reproduce bug**
```python
import cudf
import dask_cudf

df = cudf.DataFrame(
    columns=['a', 'b', 'c'],
    dtype={
        'a': 'int64',
        'b': 'datetime64[ns]',
        'c': 'object'})

ddf = dask_cudf.from_cudf(df, npartitions=1)

ddf.groupby(['a']).agg({'b': 'count'}).compute()  # RuntimeError: CUDF failure at: /opt/conda/conda-bld/work/cpp/src/groupby/groupby.cu:166: Invalid type/aggregation combination.
ddf.groupby(['a']).agg({'c': 'count'}).compute()  # DataError: All requested aggregations are unsupported
```

**Expected behavior**
I would expect the same behavior that occurred using dask-cudf 23.08, which in this case returned the respective empty dataframe results of the above groupby aggs.

**Environment overview (please complete the following information)**
 - Environment location: bare metal
 - Method of cuDF install: conda

**Environment details**
<details><summary>Click here to see environment details</summary><pre>
     
     **git***
print_env.sh: 11: [: true: unexpected operator
     Not inside a git repository
     
     ***OS Information***
     DGX_NAME="DGX Server"
     DGX_PRETTY_NAME="NVIDIA DGX Server"
     DGX_SWBUILD_DATE="2023-03-27-13-31-04"
     DGX_SWBUILD_VERSION="5.5.0"
     DGX_COMMIT_ID="b2e06e0"
     DGX_PLATFORM="DGX Server for DGX-1"
     DGX_SERIAL_NUMBER="QTFCOU8220020"
     DISTRIB_ID=Ubuntu
     DISTRIB_RELEASE=20.04
     DISTRIB_CODENAME=focal
     DISTRIB_DESCRIPTION="Ubuntu 20.04.6 LTS"
     NAME="Ubuntu"
     VERSION="20.04.6 LTS (Focal Fossa)"
     ID=ubuntu
     ID_LIKE=debian
     PRETTY_NAME="Ubuntu 20.04.6 LTS"
[...]

PR #14473, opened 2023-11-22
Correct dtype of count aggregations on empty dataframes

wence-:
## Description
A count aggregation should always return an int64 column, even if the grouped dataframe is empty. Previously we did not do this because the short-circuiting for empty inputs was hit before handling the count case. Fix this by reordering the conditions.

- Closes https://github.com/rapidsai/cudf/issues/14200

## Checklist


bdice (approved):
Thanks for the fix.

galipremsagar (approved):

Issue #14471, opened 2023-11-22
[BUG] `pd.Timestamp` objects are not picklable when `cudf.pandas` is active


shwina:
https://github.com/python/cpython/issues/111306 blocks us from proxying `pd.Timestamp`. This means `pd.Timestamp` is the "true" pandas `Timestamp` type, and pickling "true" pandas types is broken when `cudf.pandas` is active:

```python
In [1]: %load_ext cudf.pandas

In [2]: import pandas as pd

In [3]: import pickle

In [4]: pickle.dumps(pd.Timestamp("2001-01-01"))
---------------------------------------------------------------------------
PicklingError                             Traceback (most recent call last)
Cell In[4], line 1
----> 1 pickle.dumps(pd.Timestamp("2001-01-01"))

PicklingError: Can't pickle <built-in function _unpickle_timestamp>: it's not the same object as pandas._libs.tslibs.timestamps._unpickle_timestamp
```

PR #14474, opened 2023-11-22
Enable `pd.Timestamp` objects to be picklable when `cudf.pandas` is active

shwina:
Closes https://github.com/rapidsai/cudf/issues/14471.

This PR makes `Timestamp` objects picklable by registering a custom reducer for `pd.Timestamp` objects when `cudf.pandas` is active.


python/cudf/cudf/pandas/_wrappers/pandas.py
```diff
@@ -1304,3 +1305,50 @@ def _df_query_method(self, *args, local_dict=None, global_dict=None, **kwargs):
...
+    with disable_module_accelerator():
+        # args can contain objects that are unpicklable
+        # when the module accelerator is disabled
+        # (freq is of a proxy type):
+        pickled_args = pickle.dumps(ts.__reduce__())
+
+    return _unpickle_timestamp, (pickled_args,)
+
+
+def _unpickle_timedelta(pickled_args):
```
bdice, line 1332: How is `_unpickle_timedelta` different from `_unpickle_timestamp`? Similarly for `_reduce_*`. They look identical to me. Do we need this duplication or can we simplify things?

vyasr: Agreed, this looks like it can be deduplicated.

vyasr (approved):
Approving assuming Bradley's question about duplication is addressed.

PR #14476, opened 2023-11-22
Move creation of env.yaml outside the current directory

davidwendt:
## Description
Moves the creation of the `env.yaml` file in build scripts to a temporary directory instead of the current directory.

During the CI builds, a `env.yaml` file is created for use with a mamba call later in the script. This is a problem when trying to reproduce CI locally when the launched docker container tries to create this file locally as root. If the user does not have root access (or equivalent permissions), the file is not created and the process aborts.

The same error occurs when creating the `test_results` directory when the `RAPIDS_TEST_DIR` is not set. In this case the script tries to create the `test_results` in the current working directory.

## Checklist


vyasr:
Could we port this change to other repos as well? We have a new tool [rapids-reviser](https://github.com/rapidsai/rapids-reviser) that can help automate that change, @bdice could probably show you how to set this up (although scripting this change is slightly nontrivial relative to our usual since this isn't just a single sed replacement but also a new line addition).

ci/build_docs.sh
```diff
@@ -6,12 +6,14 @@ set -euo pipefail
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
+ENV_YAML_DIR="$(mktemp -d)"
+
 rapids-dependency-file-generator \
   --output conda \
   --file_key docs \
-  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee env.yaml
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee ${ENV_YAML_DIR}/env.yaml
```
raydouglass, line 14: While unlikely in regular usage, someone could mess with `TMPDIR` and add a space to the directory, so just quote usage of the `ENV_YAML_DIR`.

This should be done everywhere.

```suggestion
  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee "${ENV_YAML_DIR}/env.yaml"
```

ci/build_docs.sh
```diff
@@ -6,12 +6,14 @@ set -euo pipefail
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
+ENV_YAML_DIR="$(mktemp -d)"
```
raydouglass, line 9: Not strictly necessary since these scripts almost always run in a container, but cleaning up temp directories isn't a bad idea.

```suggestion
ENV_YAML_DIR="$(mktemp -d)"
trap 'rm -rf -- "$ENV_YAML_DIR"' EXIT
```

ajschmidt8: Aren't temp directories cleared on boot automatically? I'd vote to omit this extra line to keep things as succinct as possible.

ajschmidt8 (approved):
thanks David

vyasr:
I'm fine with getting this merged ASAP, but could we open an issue to track https://github.com/rapidsai/cudf/pull/14476#issuecomment-1823235867?

PR #14477, opened 2023-11-22
Enable doxygen XML generation and fix issues

vyasr:
## Description

Doxygen-generated XML is parsed by Breathe, so this is a prerequisite for #13846.

## Checklist

### DIFF
--- a/cpp/include/cudf_test/column_wrapper.hpp
+++ b/cpp/include/cudf_test/column_wrapper.hpp
@@ -177,6 +177,9 @@ rmm::device_buffer make_elements(InputIterator begin, InputIterator end)
     elements.data(), size * sizeof(ElementTo), cudf::test::get_default_stream()};
 }
 
+// The two signatures below are identical to the above overload apart from
+// SFINAE so doxygen sees it as a duplicate.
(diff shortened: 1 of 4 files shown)


cpp/include/cudf_test/column_wrapper.hpp
```diff
@@ -177,6 +177,9 @@ rmm::device_buffer make_elements(InputIterator begin, InputIterator end)
     elements.data(), size * sizeof(ElementTo), cudf::test::get_default_stream()};
 }
 
+// The two signatures below are identical to the above overload apart from
+// SFINAE so doxygen sees it as a duplicate.
+//! @cond Doxygen_Suppress
```
karthikeyann, line 182: will `@copydoc` work without warnings/errors?
Add an extra line below in addition to copydoc mentioning that this is specialization for that type (fixed_point, non-fixed_point).

vyasr: No, that didn't work for me. The problem is that the function signature is the same so doxygen sees two identical functions. It doesn't seem to understand the template specialization via SFINAE, only explicitly template specialization.

PR #14478, opened 2023-11-22
Improve memory footprint of isin by using contains

### LINKED ISSUE #14298: [BUG] OOM with isin when lhs/rhs contain repeats / spill fails
**Describe the bug**
When using `.isin` on a comparatively small set of `cudf.Series`, OOM occurs. This is because the `cudf.Series` contain repeats, which explodes memory in `.isin`.

The minimal reproducer is contained in https://gitlab-master.nvidia.com/tcomer/isin_bug so that you can download the .csv files containing the relatively small serieses with numerous repeats.

**Expected behavior**
`.isin` should not OOM when comparing a 2.2m column with a 275k column.

Cudf installed via conda:

`micromamba install -c rapidsai -c conda-forge -c nvidia cudf`

```
(ins)(sisclip-viz) tcomer@tcomer-NVIDIA:~/miniconda3/envs/sisclip-viz/lib/python3.10/site-packages/cudf$ micromamba list | grep cudf
  cudf                           23.08.00      cuda11_py310_230809_g8150d38e08_0  rapidsai   
  dask-cudf                      23.08.00      cuda11_py310_230809_g8150d38e08_0  rapidsai   
  libcudf                        23.08.00      cuda11_230809_g8150d38e08_0        rapidsai  
```

wence-:
## Description

Previously, isin was implemented using an inner join between the column we are searching (the haystack) and the values we are searching for (the needles). This had a large memory footprint when there were repeated needles (since that blows up the cardinality of the merge).

To fix this, note that we don't need to do a merge at all, since libcudf provides a primitive (contains) to search for many needles in a haystack. The only thing we must bear in mind is that left.isin(right) is asking for the locations in left that match an entry in right, whereas contains(haystack, needles) provides a bool mask that selects needles that are in the haystack. To get the behaviour we want, we therefore need to do contains(right, left) and treat the values to search for as the haystack.

As well as having a much better memory footprint, this hash-based approach search is significantly faster than the previous merge-based one.

While we are here, lower the memory footprint of MultiIndex.isin by using a left-semi join (the implementation is separate from the isin implementation on columns and looks a little more complicated to unpick).

- Closes #14298

## Checklist

### DIFF
--- a/python/cudf/cudf/core/column/column.py
+++ b/python/cudf/cudf/core/column/column.py
@@ -917,9 +917,19 @@ class ColumnBase(Column, Serializable, BinaryOperand, Reducible):
         to determine what values of `rhs` exist in `self`.
         """
         # We've already matched dtypes by now
+        # self.isin(other) asks "which values of self are in other"
+        # contains(haystack, needles) asks "which needles are in haystack"
+        # hence this argument ordering.
         result = libcudf.search.contains(rhs, self)
-        if result.null_count:
-            return result.fillna(False)
+        if self.null_count > 0:
+            # If one of the needles is null, then the result contains
+            # nulls, these nulls should be replaced by whether or not the
+            # haystack contains a null.
+            # TODO: this is unnecessary if we resolve
+            # https://github.com/rapidsai/cudf/issues/14515 by
+            # providing a mode in which cudf::contains does not mask
+            # the result.
+            result = result.fillna(rhs.null_count > 0, dtype=bool)
         return result
 
     def as_mask(self) -> Buffer:
--- a/python/cudf/cudf/core/column/column.py
+++ b/python/cudf/cudf/core/column/column.py
@@ -916,13 +916,11 @@ class ColumnBase(Column, Serializable, BinaryOperand, Reducible):
         Helper function for `isin` which merges `self` & `rhs`
         to determine what values of `rhs` exist in `self`.
         """
-        ldf = cudf.DataFrame({"x": self, "orig_order": arange(len(self))})
-        rdf = cudf.DataFrame(
-            {"x": rhs, "bool": full(len(rhs), True, dtype="bool")}
-        )
-        res = ldf.merge(rdf, on="x", how="left").sort_values(by="orig_order")
-        res = res.drop_duplicates(subset="orig_order", ignore_index=True)
-        return res._data["bool"].fillna(False)
+        # We've already matched dtypes by now
+        result = libcudf.search.contains(rhs, self)
+        if result.null_count:
+            return result.fillna(False)
+        return result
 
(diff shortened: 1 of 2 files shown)


bdice (approved):

galipremsagar (approved):

wence-:
> As well as having a much better memory footprint, this hash-based approach search is significantly faster than the previous merge-based one

Using the data from #14298:

Old:
```
In [1]: %timeit result = haystack.isin(needles.unique())
6.08 ms ± 24.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
New
```
# No need for unique now either
In [1]: %timeit result = haystack.isin(needles)
123 µs ± 273 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Happy thanksgiving @thomcom

wence-:
Hmph, broken somehow...

wence-:
The problem is that `contains` preserves the null mask of the needles being searched for. I fix this by dropping the returned null mask on the floor, but marking as do-not-merge so that we can discuss if this is the best approach.

python/cudf/cudf/core/column/column.py
```diff
@@ -916,13 +916,17 @@ def _obtain_isin_result(self, rhs: ColumnBase) -> ColumnBase:
...
+        # We've already matched dtypes by now
+        result = libcudf.search.contains(rhs, self)
+        # libcudf contains runs the search with nulls comparing equal
+        # and then copying the bitmask from the needles to the result.
+        # In cudf, we want nulls in both needle and haystack to
+        # produce True and nulls in only one to produce False. If we
+        # drop the bitmask of the result on the floor, it already
+        # contains the values we want. We can do this unilaterally
+        # without checking the null count of the result.
+        result.set_base_mask(None)
```
wence-, line 928: This happens to be correct because `cudf::contains` writes all the output and then adds a null mask. But that is not in the contract for a column with a null mask, the masked entries are allowed to contain anything. We can do it "above-board" by an appropriate `fillna` call here. Or we could augment the libcudf API to allow control of what happens when nulls are in the needles.

galipremsagar: I don't know if the libcudf API is being used by others elsewhere, but I'm okay having the fillna call here.

wence-: Yeah, though it's another kernel call + memory allocation. I'll do that and open a feature req.

wence-: https://github.com/rapidsai/cudf/issues/14515

wence-: OK, did the above-board thing for now.

PR #3303, opened 2019-11-05
[REVIEW] Define and implement new stream compaction APIs `copy_if`, `drop_nulls`, `apply_boolean_mask`, `drop_duplicate` and `unique_count`.

rgsl888prabhu:
Define and implement new stream compaction APIs `copy_if`, `drop_nulls`, `apply_boolean_mask`, `drop_duplicate` and `unique_count`.

NOTE: All the APIs supports string

close #2948 
close #3415

PR #14493, opened 2023-11-24
Implement `IndexedFrame.duplicated` with `distinct_indices` + `scatter`

### LINKED ISSUE #14486: [PERF/ENH]: `DataFrame.duplicated` does unnecessary inner merge
`DataFrame.duplicated` returns a `Series` that marks, for each row, whether it is a duplicate of another row in the dataframe.

The core is implemented like so:

```
        df._data["index"] = cudf.core.column.arange(
            0, len(self), dtype=size_type_dtype
        )
        new_df = df.drop_duplicates(subset=subset, keep=keep)
        idx = df.merge(new_df, how="inner")["index"]
        s = cudf.Series._from_data(
            {
                None: cudf.core.column.full(
                    size=len(self), fill_value=True, dtype="bool"
                )
            },
            index=self.index,
        )
        s.iloc[idx] = False
```

So we first add a column to the dataframe that will track which rows are not duplicates, call `drop_duplicates` and then perform a merge and grab that same "index" column. Then we construct a new column and set the values of the index column to `False`.

But `df.merge(new_df, how="inner")["index"] == new_df["index"]`. By definition, the rows in `new_df` will exist in `df` and the ones that will do so are exactly those that `drop_duplicates` returns so there's no need to do the merge at all.

Finally, we can just construct the result by with `libcudf.copying.scatter` sending a column of `False` values through `idx` into a column of `True` values (which is what the construction of `s` is doing, but with higher overhead).

wence-:
## Description

To obtain the duplicate rows in a dataframe we previously performed a drop-duplicates with a carrier column of row indices and then set entries in a boolean column to False for those row indices that remained. Furthermore, we were performing an unnecessary merge after the drop-duplicates call to obtain the row indices.

Note that the carrier column provides exactly the information that is computed internally in `libcudf` by `cudf::detail::get_distinct_indices` (called as part of `cudf::distinct`). We therefore promote `get_distinct_indices` to a public function (as `cudf::distinct_indices`) and replace the (unnecessary) merge plus `iloc`-based setting of the result with a call to `libcudf.copying.scatter`.

This provides a reasonable speedup (around 1.5x) for `duplicated()` on `Series`, and significantly improves performance of `duplicated()` on `DataFrames`, especially when providing a `subset` argument. Previously we would pay the cost in drop-duplicates of moving all columns of the distinct rows to the output table, even though we only actually needed the carrier "indices" column. Now we just obtain those indices directly, `duplicated()` scales only with the number of "active" columns. In some simple benchmarking this is between two and five times faster for tables with 10% distinct rows depending on the number of passive additional columns.

- Closes #14486

## Checklist

### DIFF
--- a/cpp/include/cudf/detail/stream_compaction.hpp
+++ b/cpp/include/cudf/detail/stream_compaction.hpp
@@ -99,17 +99,8 @@ std::unique_ptr<table> stable_distinct(table_view const& input,
                                        rmm::mr::device_memory_resource* mr);
 
 /**
- * @brief Create a column of indices of all distinct rows in the input table.
+ * @copydoc cudf::distinct_indices
  *
- * Given an `input` table_view, an output vector of all row indices of the distinct rows is
- * generated. If there are duplicate rows, which index is kept depends on the `keep` parameter.
- *
- * @param input The input table
- * @param keep Get index of any, first, last, or none of the found duplicates
- * @param nulls_equal Flag to specify whether null elements should be considered as equal
- * @param nans_equal Flag to specify whether NaN elements should be considered as equal
- * @param stream CUDA stream used for device memory operations and kernel launches
- * @param mr Device memory resource used to allocate the returned vector
  * @return A device_uvector containing the result indices
  */
 rmm::device_uvector<size_type> get_distinct_indices(table_view const& input,
--- a/cpp/include/cudf/detail/stream_compaction.hpp
+++ b/cpp/include/cudf/detail/stream_compaction.hpp
@@ -103,12 +103,12 @@ std::unique_ptr<table> stable_distinct(table_view const& input,
  *
  * @return A device_uvector containing the result indices
  */
-rmm::device_uvector<size_type> get_distinct_indices(table_view const& input,
-                                                    duplicate_keep_option keep,
-                                                    null_equality nulls_equal,
-                                                    nan_equality nans_equal,
-                                                    rmm::cuda_stream_view stream,
-                                                    rmm::mr::device_memory_resource* mr);
+rmm::device_uvector<size_type> distinct_indices(table_view const& input,
+                                                duplicate_keep_option keep,
+                                                null_equality nulls_equal,
+                                                nan_equality nans_equal,
+                                                rmm::cuda_stream_view stream,
+                                                rmm::mr::device_memory_resource* mr);
 
 /**
  * @copydoc cudf::unique_count(column_view const&, null_policy, nan_policy)
--- a/cpp/include/cudf/stream_compaction.hpp
+++ b/cpp/include/cudf/stream_compaction.hpp
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <cudf/types.hpp>
+#include <cudf/utilities/default_stream.hpp>
 
 #include <rmm/mr/device/per_device_resource.hpp>
 
@@ -275,6 +276,28 @@ std::unique_ptr<table> distinct(
   duplicate_keep_option keep          = duplicate_keep_option::KEEP_ANY,
   null_equality nulls_equal           = null_equality::EQUAL,
   nan_equality nans_equal             = nan_equality::ALL_EQUAL,
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+
+/**
+ * @brief Create a column of indices of all distinct rows in the input table.
+ *
+ * Given an `input` table_view, an output vector of all row indices of the distinct rows is
+ * generated. If there are duplicate rows, which index is kept depends on the `keep` parameter.
+ *
+ * @param input The input table
+ * @param keep Get index of any, first, last, or none of the found duplicates
+ * @param nulls_equal Flag to specify whether null elements should be considered as equal
+ * @param nans_equal Flag to specify whether NaN elements should be considered as equal
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned vector
+ * @return Column containing the result indices
+ */
+std::unique_ptr<column> distinct_indices(
+  table_view const& input,
+  duplicate_keep_option keep          = duplicate_keep_option::KEEP_ANY,
+  null_equality nulls_equal           = null_equality::EQUAL,
+  nan_equality nans_equal             = nan_equality::ALL_EQUAL,
+  rmm::cuda_stream_view stream        = cudf::get_default_stream(),
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
 /**
--- a/python/cudf/cudf/_lib/cpp/stream_compaction.pxd
+++ b/python/cudf/cudf/_lib/cpp/stream_compaction.pxd
@@ -6,10 +6,12 @@ from libcpp.vector cimport vector
 
 from cudf._lib.types import cudf_to_np_types, np_to_cudf_types
 
+from cudf._lib.cpp.column.column cimport column
 from cudf._lib.cpp.column.column_view cimport column_view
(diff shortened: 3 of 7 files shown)


cpp/include/cudf/stream_compaction.hpp
```diff
@@ -277,6 +277,28 @@ std::unique_ptr<table> distinct(
...
+ *
+ * @param input The input table
+ * @param keep Get index of any, first, last, or none of the found duplicates
+ * @param nulls_equal Flag to specify whether null elements should be considered as equal
+ * @param nans_equal Flag to specify whether NaN elements should be considered as equal
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ * @param mr Device memory resource used to allocate the returned vector
+ * @return Column containing the result indices
+ */
+std::unique_ptr<column> distinct_indices(
```
wence-, line 294: I decided not to call it `cudf::get_distinct_indices` because AFAICT nothing is that returns a result is called `get_XXX`.

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -2090,26 +2089,19 @@ def duplicated(self, subset=None, keep="first"):
...
-            df = self.copy(deep=False)
-        df._data["index"] = cudf.core.column.arange(
-            0, len(self), dtype=size_type_dtype
+            columns = [self._data[n] for n in subset]
+        distinct = libcudf.stream_compaction.distinct_indices(
+            columns, keep=keep
         )
-
-        new_df = df.drop_duplicates(subset=subset, keep=keep)
-        idx = df.merge(new_df, how="inner")["index"]
```
wence-, line 2102: This merge was a no-op.

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -2090,26 +2089,19 @@ def duplicated(self, subset=None, keep="first"):
...
         else:
-            df = self.copy(deep=False)
-        df._data["index"] = cudf.core.column.arange(
-            0, len(self), dtype=size_type_dtype
+            columns = [self._data[n] for n in subset]
+        distinct = libcudf.stream_compaction.distinct_indices(
+            columns, keep=keep
         )
-
-        new_df = df.drop_duplicates(subset=subset, keep=keep)
```
wence-, line 2101: This did too much work (we only need the indices column for further progress which is what the new API provides).

wence-:
Note that one could also imagine a `distinct_by_key(table_view values, table_view keys)`, but the choice in these functions (which is deliberate policy, https://github.com/rapidsai/cudf/pull/3303#discussion_r345363610) is to just allow specifying keys by subsetting a table. However, if one only needs a subset of the table as output to continue, this will move more data in the gather phase than necessary: there's no way with the current `distinct(table_view values, std::vector<size_type> key_columns)` to say "use these columns as keys, but only gather these other columns".

cpp/include/cudf/detail/stream_compaction.hpp
```diff
@@ -99,17 +99,8 @@ std::unique_ptr<table> stable_distinct(table_view const& input,
...
- *
- * @param input The input table
- * @param keep Get index of any, first, last, or none of the found duplicates
- * @param nulls_equal Flag to specify whether null elements should be considered as equal
- * @param nans_equal Flag to specify whether NaN elements should be considered as equal
- * @param stream CUDA stream used for device memory operations and kernel launches
- * @param mr Device memory resource used to allocate the returned vector
  * @return A device_uvector containing the result indices
  */
 rmm::device_uvector<size_type> get_distinct_indices(table_view const& input,
```
bdice, line 106: Can we rename this to `distinct_indices`? It's a detail method so we should be safe to refactor that internally to match the public API that is introduced in this PR.

cpp/include/cudf/stream_compaction.hpp
bdice: Good call. Let's rename the detail API to match, though (see other comment).

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -2090,26 +2089,19 @@ def duplicated(self, subset=None, keep="first"):
...
-            {
-                None: cudf.core.column.full(
-                    size=len(self), fill_value=True, dtype="bool"
-                )
-            },
-            index=self.index,
+        (result,) = libcudf.copying.scatter(
+            [full(len(distinct), False, dtype=bool)],
+            distinct,
+            [full(len(self), True, dtype=bool)],
```
bdice, line 2101: Is it possible to scatter (or "set") a **scalar** `True` value over a column of `False` instead of scattering from column-of-true to column-of-false?

wence-: I can probably scatter a scalar False to a column of True. To scatter a True to a column of False would require one invert the mask first

bdice: I didn't check which direction it's supposed to go, I just wanted to avoid allocating two full columns. Please verify my claim. :)

wence-: Implemented via scatter-scalar.

cpp/include/cudf/detail/stream_compaction.hpp
wence-: Done.

python/cudf/cudf/_lib/stream_compaction.pyx
```diff
@@ -160,6 +163,68 @@ def drop_duplicates(list columns,
...
 
+@acquire_spill_lock()
+def distinct_indices(
+    list columns,
+    object keep="first",
+    bool nulls_equal=True,
+    bool nans_equal=True,
+):
+    """
+    Return indices of the distinct rows in a table
```
bdice, line 174:
```suggestion
    Return indices of the distinct rows in a table.
```

python/cudf/cudf/_lib/stream_compaction.pyx
```diff
@@ -160,6 +163,68 @@ def drop_duplicates(list columns,
...
+    cdef duplicate_keep_option cpp_keep_option
+
+    if keep == 'first':
+        cpp_keep_option = duplicate_keep_option.KEEP_FIRST
+    elif keep == 'last':
+        cpp_keep_option = duplicate_keep_option.KEEP_LAST
+    elif keep is False:
+        cpp_keep_option = duplicate_keep_option.KEEP_NONE
+    else:
+        raise ValueError('keep must be either "first", "last" or False')
```
bdice, line 201:
```suggestion
        raise ValueError('keep must be either "first", "last", or False')
```

bdice (approved):

python/cudf/cudf/_lib/cpp/stream_compaction.pxd
```diff
@@ -45,3 +47,10 @@ cdef extern from "cudf/stream_compaction.hpp" namespace "cudf" \
         duplicate_keep_option keep,
         null_equality nulls_equal,
     ) except +
+
+    cdef unique_ptr[column] distinct_indices(
+        table_view input,
+        duplicate_keep_option keep,
+        null_equality nulls_equal,
+        nan_equality nans_equal,
+    )
```
karthikeyann, line 56: `except +` missing. is this the intended behaviour?

karthikeyann: Related docs:
https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#exceptions
https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#add-public-attributes

wence-: It is not, thanks!
```suggestion
    ) except +
```

Issue #14023, opened 2023-08-31
[FEA] Stop working around Cython issues


vyasr:
**Is your feature request related to a problem? Please describe.**
cudf Cython code currently contains various workarounds for Cython limitations. With the Cython 3 release, many of these limitations have been lifted. Cython development is quite active right now, and now that we have migrated to Cython 3 we can also track new minor/patch version updates easily if new issues are fixed. This is a tracking issue of the various workarounds in cudf that we should look into removing as they are resolved upstream.

Due to the ongoing pylibcudf effort, in general it will not make sense to fix these issues in legacy cudf Cython code. We should instead endeavor to fix these issues in new pylibcudf code that will replace legacy Cython over time.

**Describe the solution you'd like**

- Enum class support
  - Added in Cython 3, but there are various workarounds currently in place. Once https://github.com/cython/cython/issues/5637 and https://github.com/cython/cython/issues/5609 are merged and released, we can update cudf to remove workarounds introduced in https://github.com/rapidsai/cudf/pull/13931.
- reference_wrapper
  - Currently we have [our own declarations](https://github.com/rapidsai/cudf/blob/branch-23.10/python/cudf/cudf/_lib/cpp/libcpp/functional.pxd). There's [an open PR](https://github.com/cython/cython/pull/5671) to add this to Cython, once it's merged we can remove our version
- Use of const in templates
  - There are issues in Cython parsing const inside templates. https://github.com/cython/cython/issues/3885 was resolved by https://github.com/cython/cython/pull/3886, but https://github.com/cython/cython/issues/4180 is still open. Based on my experimentation within cudf there are still some contexts where this is not resolved correctly. The current workaround inside cudf is to ctypedef a combined type (e.g. `ctypedef const scalar constscalar`). Once the bugs are fixed we should be able to get rid of the extraneous typedefs.
- optional
  - Support for std::optional was added in https://github.com/cython/cython/pull/3294 so we can use that anytime now
- make_unique
  - The fix we need for make_unique was completed in https://github.com/cython/cython/pull/5569, so we can use that as well

vyasr:
reference_wrapper, optional, and make_unique are handled in #14382

vyasr:
The outstanding work here is blocked on Cython improvements
- https://github.com/cython/cython/issues/4180 
- https://github.com/cython/cython/issues/5609

[...]

PR #14508, opened 2023-11-27
Implement more copying APIs in pylibcudf

vyasr:
## Description

This PR implements a handful of libcudf copying APIs in pylibcudf and rewrites cudf Cython internals to use those functions where appropriate. This PR only implements those methods that can be done without any additional changes to the pylibcudf Column class. I'll follow up with other PRs for changes that require those changes.

## Checklist


cpp/include/cudf/copying.hpp
```diff
@@ -211,7 +211,7 @@ std::unique_ptr<table> scatter(
 /**
  * @brief Indicates when to allocate a mask, based on an existing mask.
  */
-enum class mask_allocation_policy {
+enum class mask_allocation_policy : int32_t {
```
vuule, line 214: why not a smaller type?

vyasr: No particular reason other than that's what our other enums use. I think everything has probably just copied `type_id`: https://github.com/rapidsai/cudf/blob/branch-24.02/cpp/include/cudf/types.hpp#L187

vuule (approved):
C++ looks good :D

python/cudf/cudf/_lib/pylibcudf/copying.pyx
```diff
@@ -55,3 +66,198 @@ cpdef Table gather(
...
+                    target_table.view(),
+                )
+            )
+    else:
+        raise ValueError("source must be a Table or list[Scalar]")
+
+    return Table.from_libcudf(move(c_result))
+
+
+cpdef object empty_like(object input):
```
shwina, line 118: Is it worth having an `empty_column_like()` and `empty_table_like()`?

python/cudf/cudf/_lib/pylibcudf/copying.pyx
```diff
@@ -1,23 +1,34 @@
...
 from cudf._lib.cpp.copying import \
     out_of_bounds_policy as OutOfBoundsPolicy  # no-cython-lint
 
-from cudf._lib.cpp.table.table cimport table
-
 from .column cimport Column
 from .table cimport Table
 
+# workaround for https://github.com/cython/cython/issues/3885
+ctypedef const scalar constscalar
```
wence-, line 30: That bug is now fixed, so plausibly we don't need this any more. But I think you need this for cython/cython#4180, so **suggestion** update the issue link?

python/cudf/cudf/_lib/pylibcudf/copying.pyx
```diff
@@ -55,3 +66,198 @@ cpdef Table gather(
...
+                    (<Table> source).view(),
+                    scatter_map.view(),
+                    target_table.view(),
+                )
+            )
+    elif isinstance(source, list):  # TODO: is list too restrictive?
+        for slr in source:
+            source_scalars.push_back(
+                reference_wrapper[constscalar](dereference(slr.c_obj))
+            )
```
wence-, line 102: **suggestion** This pattern already repeats twice, let's pull it into a utility function.

python/cudf/cudf/_lib/pylibcudf/copying.pxd
```diff
@@ -13,3 +15,23 @@ cpdef Table gather(
...
+
+cpdef Table boolean_mask_scatter(object input, Table target, Column boolean_mask)
+
+cpdef Column copy_range(
+    Column input_column,
+    Column target_column,
+    size_type input_begin,
+    size_type input_end,
+    size_type target_begin,
+)
```
wence-, line 37: **Suggestion/nitpick** What order should write these? I would suggest one of:

1. Alphabetical
2. The same order as the C++ header file
3. The same order as the doxygen documentation of the matching header file (which I think is the same as (2)

Ideally we would also name the arguments the same, but that is very persnickity.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
vyasr: I debated this as well. I could go either way as long as whatever we decide here is made consistent policy across pylibcudf. The current approach matches the libcudf API and makes it simpler for the user at the cost of having to deal with typing internally and introducing a bit of overhead as a result. The alternative would be strongly typed but slightly more verbose functions. Do you think the latter is better?

python/cudf/cudf/_lib/copying.pyx
```diff
@@ -598,45 +467,33 @@ def boolean_mask_scatter(list input_, list target_columns,
         return []
 
     if isinstance(input_[0], Column):
-        return _boolean_mask_scatter_columns(
-            input_,
-            target_columns,
-            boolean_mask
-        )
+        plc_input = pylibcudf.Table([col.to_pylibcudf(mode="read") for col in input_])
```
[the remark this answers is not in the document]
vyasr: It's not reserved per se. It's a built-in function as opposed to a keyword like `in`, so you can shadow it. It's only really an issue if you actually needed that function, but I don't know if it might confuse some linters (I haven't observed that in a long time though) and it can lead to odd syntax highlighting in editors. I can change it if you'd like, no strong preference.

python/cudf/cudf/_lib/cpp/copying.pxd
```diff
@@ -36,118 +36,118 @@ cdef extern from "cudf/copying.hpp" namespace "cudf" nogil:
     ) except +
 
     cdef unique_ptr[table] scatter (
-        table_view source_table,
-        column_view scatter_map,
-        table_view target_table,
+        const table_view source_table,
```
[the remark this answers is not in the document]
vyasr: Correct, unfortunately not supported.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
vyasr: Ah good catch, yes I documented that in https://github.com/rapidsai/cudf/issues/14023 but didn't update this link. Can do.

python/cudf/cudf/_lib/pylibcudf/copying.pxd
vyasr: I meant to do that, thanks for the reminder. I think matching the header makes sense.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
shwina: I think:

- Separate APIs for `Column`/`Table`
- No need for separate APIs for other kinds of overloads in C++

vyasr: I'll also include scalars alongside column and tables, but otherwise yes agreed.

python/cudf/cudf/_lib/copying.pyx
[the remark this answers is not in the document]
vyasr: Also no longer relevant as above.

python/cudf/cudf/_lib/cpp/copying.pxd
```diff
@@ -36,118 +36,118 @@ cdef extern from "cudf/copying.hpp" namespace "cudf" nogil:
...
+        const scalar& lhs,
+        const column_view rhs,
+        const column_view boolean_mask
     ) except +
 
     cdef unique_ptr[column] copy_if_else (
-        column_view lhs,
-        scalar rhs,
-        column_view boolean_mask
+        const column_view lhs,
```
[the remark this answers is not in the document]
vyasr: Good call, I'll fix that.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
vyasr: Done.

python/cudf/cudf/_lib/pylibcudf/copying.pxd
vyasr: Done.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
vyasr: Hmm this choice ended up being a bit more loaded than I would have liked given the functions that accept either a table or a vector of scalars. Let me know what you think of the current set of functions that I've defined.

vyasr:
I've addressed all the comments but still need to track down the dask test failure. It doesn't actually look dask-related, but rather a case that's better covered by our dask tests than our other tests.

vyasr:
Whoops sorry @vuule I clicked on the wrong button and re-requested you unintentionally!

wence- (approved):
Minor (nonblocking) suggestions, LGTM!

python/cudf/cudf/_lib/pylibcudf/copying.pyx
```diff
@@ -1,23 +1,51 @@
...
+    if not isinstance(source, list) or not isinstance(source[0], Scalar):
+        raise ValueError("source must be a list[Scalar]")
+
+    cdef vector[reference_wrapper[const scalar]] c_scalars
+    c_scalars.reserve(len(source))
+    cdef Scalar slr
+    for slr in source:
+        c_scalars.push_back(
+            reference_wrapper[constscalar](dereference(slr.c_obj))
+        )
```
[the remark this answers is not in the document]
vyasr: I did not know that syntax existed! Cool, yes that works better.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
```diff
@@ -55,3 +83,187 @@ cpdef Table gather(
...
+        c_table_result = move(
+            cpp_copying.empty_like(
+                (<Table> input).view(),
+            )
+        )
+    return Table.from_libcudf(move(c_table_result))
+
+
+cpdef Column allocate_like(
+    Column input_column, mask_allocation_policy policy, size=None
```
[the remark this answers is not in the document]
vyasr: Hmm this raises a good question of in what ways we want to deviate from libcudf APIs. Keyword-only arguments could be nice, but should we have systematic rules on when we use them?

python/cudf/cudf/_lib/types.pyx
```diff
@@ -270,9 +271,13 @@ cpdef dtype_to_pylibcudf_type(dtype):
         else:
             tid = pylibcudf.TypeId.DECIMAL32
         return pylibcudf.DataType(tid, -dtype.scale)
-    return pylibcudf.DataType(
-        SUPPORTED_NUMPY_TO_PYLIBCUDF_TYPES[np.dtype(dtype)]
-    )
+
+    # libcudf types don't support localization so convert to the base type
```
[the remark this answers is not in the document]
vyasr: It is, but I don't think it matters. At least, it's equivalent to what we were doing previously. The reason `dtype_to_data_type` doesn't have this is because we do implicit preprocessing by calling [`_get_base_dtype`](https://github.com/rapidsai/cudf/blob/branch-24.02/python/cudf/cudf/_lib/column.pyx#L338), which [does the same thing](https://github.com/rapidsai/cudf/blob/branch-24.02/python/cudf/cudf/utils/dtypes.py#L748).

python/cudf/cudf/_lib/pylibcudf/copying.pyx
shwina: Hmm, I agree I don't love it. If this were pure Python maybe `overload`s would be the right choice typing wise.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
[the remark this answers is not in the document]
wence-: My preference is that optional arguments should be keyword-only, which is perhaps an easy systematic rule.

My rationale for this is that multiple defaulted keyword-like (but also positional) arguments are easy to get wrong when using an API. For example, `allocate_like` here takes an overriding `size`, but what if it also took an overriding `dtype`. With `allocate_like(input, policy, size=None, dtype=None)` I have to remember which way round to specify things if I want to override both, with `allocate_like(input, policy, *, size=None, dtype=None)` I don't, and can't get it wrong.

Equally, the use of the keyword name makes the (often untyped) argument self-documenting. It's obvious at the call-site that we're passing a mask_allocation_policy in as the policy argument (since that's an enum that we have to get a value out of), but for `size` it could come from anywhere.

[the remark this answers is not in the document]
vyasr: I agree, and @shwina also agreed offline. Unfortunately, Cython doesn't support keyword-only arguments for cdef functions, and therefore also not for cpdef functions.

python/cudf/cudf/_lib/pylibcudf/copying.pyx
vyasr: We discussed naming offline. The state of the PR as of https://github.com/rapidsai/cudf/pull/14508/commits/b0ae7f0ff7ca31c35b84e0e87723450f16bc1b69 is our final choice of names.

vyasr:
Approved offline with the current set of changes, going to merge now.

PR #14526, opened 2023-11-29
Drop llvm16 from cuda118-conda devcontainer image

charlesbluca:
## Description
Looks like the old image used for cuda11.8-conda no longer exists:

```
[-26553 ms] Error response from daemon: No such image: rapidsai/devcontainers:24.02-cpp-llvm16-cuda11.8-mambaforge-ubuntu22.04
```

On the suggestion of @trxcllnt, trying out `24.02-cpp-cuda11.8-mambaforge-ubuntu22.04` instead.

## Checklist

### DIFF
--- a/.devcontainer/cuda11.8-conda/devcontainer.json
+++ b/.devcontainer/cuda11.8-conda/devcontainer.json
@@ -5,7 +5,7 @@
     "args": {
       "CUDA": "11.8",
       "PYTHON_PACKAGE_MANAGER": "conda",
-      "BASE": "rapidsai/devcontainers:24.02-cpp-llvm16-cuda11.8-mambaforge-ubuntu22.04"
+      "BASE": "rapidsai/devcontainers:24.02-cpp-cuda11.8-mambaforge-ubuntu22.04"
     }
   },
   "hostRequirements": {"gpu": "optional"},


bdice:
@charlesbluca This won't be tested by CI -- only CUDA 12 images are tested in CI. If you can verify that this works locally, I will approve and merge it.

bdice (commented):
Is a similar change needed here for CUDA 11.8 + pip? https://github.com/charlesbluca/cudf/blob/75cebe9382c69994f89953f004c732b302c9fd6e/.devcontainer/cuda11.8-pip/devcontainer.json#L8

charlesbluca:
Yup, good catch @bdice - can make the change there too and will follow up after verifying the devcontainer builds for `cuda11.8-conda|pip`

cwharris:
We want to keep llvm in the pip env containers. We want to drop it for conda. We'll get the relevant bits from conda as conda often requires we do so. Meanwhile pip is fine with system-wide deps in most cases. I think the changes look good as is.

cwharris (approved):
LGTM

charlesbluca:
Forgot to follow up - can verify that the changes here unblocked the `cuda11.8-conda` devcontainer build on my end, so if things look good this should be safe to merge once tests are green 🙂

bdice (approved):

bdice:
Thanks for the context @cwharris, I see why that makes sense to use system LLVM for pip devcontainers but wouldn’t have thought to do it otherwise.

PR #14529, opened 2023-11-29
Add a test for issue 12773

### LINKED ISSUE #12773: [BUG] can't merge dask-cudf dataframes when index is non-numeric?
**Describe the bug**

Consider
```python
import cudf
import dask_cudf as dd

df1 = cudf.DataFrame({"a": ["a", "b"], "b": [1, 2]})
df2 = cudf.DataFrame({"a": ["a", "c"], "b": [2, 3]})

ddf1 = dd.from_cudf(df1, npartitions=2).set_index("a")
ddf2 = dd.from_cudf(df2, npartitions=2).set_index("a")

union = ddf1.merge(ddf2, left_index=True, right_index=True, how="left").compute()
```

This produces

```pytb
File ~/cudf/python/cudf/cudf/core/column/string.py:5475, in StringColumn.as_numerical_column(self, dtype, **kwargs)
   5473 elif out_dtype.kind == "f":
   5474     if not libstrings.is_float(string_col).all():
-> 5475         raise ValueError(
   5476             "Could not convert strings to float "
   5477             "type due to presence of non-floating values."
   5478         )
   5480 result_col = _str_to_numeric_typecast_functions[out_dtype](string_col)
   5481 return result_col

ValueError: Could not convert strings to float type due to presence of
non-floating values.
```

This should work. I think something strange is going on, effectively
the error occurs because dask is trying to do

```python
df3 = df1.set_index("a")

df4 = df3.assign(foo=cudf.Series([1, 2]))
```

And this fails because the assignment column doesn't have the same
index as the dataframe.

Whose bug is this? Everything "works" with pandas-backed dataframes.

vyasr:
## Description

Closes #12773.

## Checklist


vyasr:
CC @wence-

galipremsagar (approved):

python/dask_cudf/dask_cudf/tests/test_join.py
```diff
@@ -359,3 +359,14 @@ def test_single_partition():
...
+
+def test_issue_12773():
+    # https://github.com/rapidsai/cudf/issues/12773
+    df1 = cudf.DataFrame({"a": ["a", "b"], "b": [1, 2]})
+    df2 = cudf.DataFrame({"a": ["a", "c"], "b": [2, 3]})
+
+    ddf1 = dgd.from_cudf(df1, npartitions=2).set_index("a")
+    ddf2 = dgd.from_cudf(df2, npartitions=2).set_index("a")
+
+    ddf1.merge(ddf2, left_index=True, right_index=True, how="left").compute()
```
wence-, line 372: Can we check for correctness please?

wence- (changes requested):
Thanks! Would prefer to check the output value as well.

python/dask_cudf/dask_cudf/tests/test_join.py
vyasr: Yup, done. Sorry for the delay.

wence- (approved):
Thanks!

PR #14534, opened 2023-11-30
Eagerly populate the class dict for cudf.pandas proxy types

shwina:
Rather than dynamically looking up class attributes (and methods), this PR makes it so that we eagerly populate the class with all known methods and attributes (by inspecting the "slow" class).

This solves a number of problems:

- it makes `getattr` trivially inexpensive (no dynamic `__getattr__` for each attribute access)
- it ensures the _same_ object is returned every time you do, e.g., `DataFrame.max`
- it makes tab completion fast because the attributes don't have to be computed each time
- it no longer exposes attributes that are specific to cuDF - for example `Series.list`
- it allows subclassing of proxy types to work better. For example, derived types can now call `super().` to access attributes of base types


python/cudf/cudf/pandas/fast_slow_proxy.py
```diff
@@ -466,15 +436,6 @@ def __dir__(self):
         except AttributeError:
             return type.__dir__(self)
 
-    def __getattr__(self, name: str) -> Any:
```
shwina, line 469: Note - there's no longer dynamic attribute lookup via `__getattr__`.

python/cudf/cudf/pandas/_wrappers/pandas.py
```diff
@@ -93,14 +93,16 @@ class _AccessorAttr:
     """
 
     def __init__(self, typ):
-        self.__typ = typ
+        self._typ = typ
+
+    def __set_name__(self, owner, name):
```
bdice, line 98: TIL a new part of the descriptors object model!

https://docs.python.org/3/howto/descriptor.html#closing-thoughts

python/cudf/cudf/pandas/fast_slow_proxy.py
```diff
@@ -857,17 +775,141 @@ class _FunctionProxy(_CallableProxyMixin):
...
+        fast: Callable | _Unusable,
+        slow: Callable,
+        *,
+        assigned=None,
+        updated=None,
+    ):
         self._fsproxy_fast = fast
         self._fsproxy_slow = slow
-        functools.update_wrapper(self, slow)
+        assigned = (
```
bdice, line 788: These just reassign None when the value is None, but in a slightly nonobvious way. Let’s simplify to
```python
if assigned is None:
    assigned = functools.WRAPPER_ASSIGNMENTS
```
and similarly for `updated`.

python/cudf/cudf/pandas/fast_slow_proxy.py
```diff
@@ -857,17 +775,141 @@ class _FunctionProxy(_CallableProxyMixin):
...
+            fast,
+            slow,
+            updated=functools.WRAPPER_UPDATES,
+            assigned=(
+                tuple(filter(lambda x: x != "__name__", _WRAPPER_ASSIGNMENTS))
+            ),
+        )
+
+    def __dir__(self):
+        return self._fsproxy_slow.__dir__()
```
bdice, line 895: Generally use the “normal” ways of accessing the object model even if it’s a part of defining the object model. Same for things like `__hash__` and `__eq__` and `__repr__`. (If that makes sense…)
```suggestion
        return dir(self._fsproxy_slow)
```

bdice (approved):

python/cudf/cudf/pandas/fast_slow_proxy.py
```diff
@@ -217,6 +220,14 @@ def _fsproxy_state(self) -> _State:
         elif v is not _DELETE:
             cls_dict[k] = v
 
+    for slow_name in dir(slow_type):
+        if slow_name in cls_dict or slow_name.startswith("__"):
```
mroeschke, line 224: For arithmetic dunder methods like `__add__`, wouldn't we want to create a `_FastSlowAttribute` for that?

shwina: So those are defined in the base class `_FastSlowProxy` now concretely (so it's already in the class dictionary).

shwina:
> Generally looks good to me. Any impact on performance or the pass rate of the pandas test suite?

w.r.t benchmarks (and assuming we can unblock them) - would an improvement (or at least no degradation) in the time it takes to run the pandas unit tests be a good benchmark? (@bdice)

python/cudf/cudf/pandas/_wrappers/pandas.py
```diff
@@ -744,7 +749,7 @@ def _get_eval_locals_and_globals(level, local_dict=None, global_dict=None):
     return local_dict, global_dict
 
 
-@register_proxy_func(pd.eval)
+@register_proxy_func(pd.core.computation.eval.eval)
```
mroeschke, line 752: Curious, does this _need_ to be the exact location of `eval` instead of using the public `pd.eval`?

Similar to the register functions below e.g. `pd.api.extensions.register_dataframe_accessor`?

shwina: I think that used to be the case when we first wrote it, but no longer so

mroeschke: Ah OK. Would be nice if this could be avoided but not a big deal

galipremsagar:
@vyasr It looks like there are hangs, I'm currently debugging them locally. The run is currently taking more than 2 hours. I'll push the changes to this PR once I have them ready.

vyasr:
Got it, thanks for the update @galipremsagar! I was clearing out my backlog and thought this PR might be ready.

python/cudf/cudf/pandas/fast_slow_proxy.py
```diff
@@ -546,140 +517,283 @@ def __dir__(self):
...
             return
-        return _FastSlowAttribute("__setattr__").__get__(self)(name, value)
+        return _FastSlowAttribute("__setattr__").__get__(self, type(self))(
+            name, value
+        )
 
     def __add__(self, other):
-        return _fast_slow_function_call(operator.add, self, other)[0]
+        return _fast_slow_function_call(
+            lambda x, y: x.__add__(y)
```
wence-, line 530: *question*: these changes are because `operator.add` might do `y.__radd__(x)`, which we don't want?

python/cudf/cudf/pandas/fast_slow_proxy.py
```diff
@@ -855,17 +969,170 @@ class _FunctionProxy(_CallableProxyMixin):
...
+                fast_attr = getattr(
+                    owner._fsproxy_fast, self._name, _Unusable()
+                )
+
+            # if self._name in {"_data", "_mask", "storage", "css", "ctx"}:
+            #     return _maybe_wrap_result(
+            #             getattr(instance._fsproxy_slow, self._name),
+            #             None,  # type: ignore
+            #         )
+            # else:
```
wence-, line 1060: Debugging?

python/cudf/cudf/pandas/fast_slow_proxy.py
```diff
@@ -855,17 +969,170 @@ class _FunctionProxy(_CallableProxyMixin):
...
+            try:
+                slow_attr = getattr(owner._fsproxy_slow, self._name)
+                # if is_bound_method(slow_attr) and instance is not None:
+                #     slow_attr = getattr(slow_attr, "__func__", slow_attr)
```
wence-, line 1064: Debugging?

wence- (commented):
*question*: It seems like this branch is also pulling in a bunch of unrelated (?) changes in cudf-proper. Is that deliberate in the sense these are bug-fixes for issues that are exposed by these changes?

galipremsagar:
> _question_: It seems like this branch is also pulling in a bunch of unrelated (?) changes in cudf-proper. Is that deliberate in the sense these are bug-fixes for issues that are exposed by these changes?

Yes, those are all the bug-fixes for the issues that are exposed by these changes, failures can be found here: https://github.com/rapidsai/cudf/actions/runs/8916717301/attempts/2#summary-24512387065

wence-:
> > _question_: It seems like this branch is also pulling in a bunch of unrelated (?) changes in cudf-proper. Is that deliberate in the sense these are bug-fixes for issues that are exposed by these changes?
> 
> Yes, those are all the bug-fixes for the issues that are exposed by these changes, failures can be found here: https://github.com/rapidsai/cudf/actions/runs/8916717301/attempts/2#summary-24512387065
> 
> The initial set of changes caused 12% drop in pass-rate, with the current bug fixes we are down to 2% decrease (still WIP)

Aha, thanks!

python/cudf/cudf/pandas/_wrappers/pandas.py
```diff
@@ -1171,6 +1187,135 @@ def _df_query_method(self, *args, local_dict=None, global_dict=None, **kwargs):
...
+    "GoodFriday",
+    _Unusable,
+    pd_GoodFriday,
+    fast_to_slow=_Unusable(),
+    slow_to_fast=_Unusable(),
+    additional_attributes={"__hash__": _FastSlowAttribute("__hash__")},
+)
+
+EasterMonday = make_final_proxy_type(
+    "GoodFriday",
```
wence-, line 1310:
```suggestion
    "EasterMonday",
```

python/cudf/cudf/pandas/fast_slow_proxy.py
galipremsagar: Yes, I forked out these changes to a separate PR: https://github.com/rapidsai/cudf/pull/15684

ci/cudf_pandas_scripts/pandas-tests/run.sh
```diff
@@ -22,6 +22,7 @@ bash python/cudf/cudf/pandas/scripts/run-pandas-tests.sh \
   -n 10 \
   --tb=no \
   -m "not slow" \
+  --durations=0 \
```
galipremsagar, line 25:
```suggestion
```

galipremsagar (approved):

galipremsagar:
@wence- I've reverted the vast majority of the changes I made to this PR and this is now ready for review. Since I've made a couple of minor fixes to this pr I'd like to get it reviewed by someone other than me before merging.

python/cudf/cudf/pandas/fast_slow_proxy.py
galipremsagar: I'm not sure why, but using `dir` instead of `.__dir__()` is causing 500 more failures consistently, in the interest of time I haven't dug into those failures.

python/cudf/cudf/pandas/fast_slow_proxy.py
```diff
@@ -790,17 +717,162 @@ class _FunctionProxy(_CallableProxyMixin):
...
+
+class _FastSlowAttribute:
     """
-    Methods of fast-slow proxies are of type _MethodProxy.
+    A descriptor type used to define attributes of fast-slow proxies.
     """
 
+    _attr: Any
+
+    def __init__(self, name: str, private=False):
```
mroeschke, line 785:
```suggestion
    def __init__(self, name: str, *, private: bool = False):
```

Would be nice to make `private` keyword only

python/cudf/cudf/pandas/_wrappers/numpy.py
```diff
@@ -122,6 +123,7 @@ def wrap_ndarray(cls, arr: cupy.ndarray | numpy.ndarray, constructor):
         "__iter__": custom_iter,
         # Special wrapping to handle scalar values
         "_fsproxy_wrap": classmethod(wrap_ndarray),
+        "base": _FastSlowAttribute("base", True),
```
mroeschke, line 126:
```suggestion
        "base": _FastSlowAttribute("base", private=True),
```

python/cudf/cudf/pandas/fast_slow_proxy.py
galipremsagar: Done 👍

python/cudf/cudf/pandas/fast_slow_proxy.py
```diff
@@ -790,17 +717,162 @@ class _FunctionProxy(_CallableProxyMixin):
...
+                    owner._fsproxy_fast, self._name, _Unusable()
+                )
+
+            try:
+                slow_attr = getattr(owner._fsproxy_slow, self._name)
+            except AttributeError as e:
+                if instance is not None:
+                    return _maybe_wrap_result(
+                        getattr(instance._fsproxy_slow, self._name),
+                        None,  # type: ignore
```
mroeschke, line 809: Would be good to adjust the `func` typing in _maybe_wrap_result as `func: Callable | None` and raise an exception in `is_final_type` and `is_intermediate_type` branches `if func is None`

python/cudf/cudf/pandas/_wrappers/pandas.py
```diff
@@ -273,6 +277,9 @@ def Index__new__(cls, *args, **kwargs):
         "__new__": Index__new__,
         "_constructor": _FastSlowAttribute("_constructor"),
         "__array_ufunc__": _FastSlowAttribute("__array_ufunc__"),
+        "_accessors": set(),
+        "_data": _FastSlowAttribute("_data", private=True),
```
mroeschke, line 281: So these attributes we are wanting the fast path to have a chance to evaluate and not automatically be `_Unusable()`?

galipremsagar: yes

python/cudf/cudf/pandas/scripts/run-pandas-tests.sh
```diff
@@ -138,7 +138,7 @@ and not test_eof_states"
 # TODO: Remove "not db" once a postgres & mysql container is set up on the CI
 PANDAS_CI="1" timeout 30m python -m pytest -p cudf.pandas \
     -v -m "not single_cpu and not db" \
-    -k "$TEST_THAT_NEED_MOTO_SERVER and $TEST_THAT_CRASH_PYTEST_WORKERS" \
+    -k "$TEST_THAT_NEED_MOTO_SERVER and $TEST_THAT_CRASH_PYTEST_WORKERS and not test_groupby_raises_category_on_category and not test_constructor_no_pandas_array and not test_is_monotonic_na and not test_index_contains and not test_index_contains and not test_frame_op_subclass_nonclass_constructor and not test_round_trip_current" \
```
mroeschke, line 141: So these are still confirmed failing with now eagerly populating the class dict?

python/cudf/cudf/pandas/_wrappers/pandas.py
mroeschke: Ah so in this case since it's `private` we won't try to operate on e.g. `cudf.Index._data` which is a `ColumnAccessor` and instead reach for `cudf.Index._data` which is a numpy ndarray which we can wrap.

This part of the eager population I'm not too thrilled about. These private variables are not guaranteed to be stable. It would be nice if any slow attribute was attempted to return a proxy object so we didn't have to specify attributes like this but I suppose that can be looked at in a follow up

python/cudf/cudf/pandas/scripts/run-pandas-tests.sh
galipremsagar: yes, some of these segfaults and a few take quite a long time. The plan is to tackle each of them separately as follow-ups.

python/cudf/cudf/pandas/_wrappers/pandas.py
galipremsagar: Yeah, there are still such cases we can address in follow-ups. At this point the scope of this PR has kept on expanding, that's why I haven't addressed additional issues in this PR.

mroeschke (approved):
My typing request and concern about having to define private attributes in our proxies can be addressed in a follow up

python/cudf/cudf/pandas/fast_slow_proxy.py
galipremsagar: This doesn't seem to be straight-forward, while testing this change I noticed `4k` additional failures. Maybe we can revisit this in another PR.

PR #14535, opened 2023-11-30
Move non-templated inline function definitions from table_view.hpp to table_view.cpp

### LINKED ISSUE #14431: Move non-templated inline function definitions from `table_view.hpp` to source file
Reference comment: https://github.com/rapidsai/cudf/pull/14363#discussion_r1394244316_

It is possibly worth an effort to sweep through the codebase to find other spots that may benefit from such refactoring as well.

davidwendt:
## Description
Moves some `inline` functions from the `table_view.hpp` to the `table_view.cpp` to help reduce dependency bloat.
Also reference #14531: I'd like to minimize changes to the highly inclusive `table_view.hpp`.

Closes #14431 

## Checklist

### DIFF
--- a/cpp/include/cudf/table/table_view.hpp
+++ b/cpp/include/cudf/table/table_view.hpp
@@ -302,10 +302,7 @@ class mutable_table_view : public detail::table_view_base<mutable_column_view> {
  * @param view The table to check for nullability
  * @return True if any of the columns in the table is nullable, false otherwise
  */
-inline bool nullable(table_view const& view)
-{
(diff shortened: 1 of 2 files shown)


vyasr (approved):

hyperbolic2346:
Are we concerned about the performance implications of removing the ability to inline these functions? I only ask because that is the assumed reason they were there in the first place.

davidwendt:
I'm not concerned about inlining CPU code. These in particular should be negligible.
They were likely created before the cpp file existed.

karthikeyann (approved):
LGTM 👍

hyperbolic2346 (approved):

Issue #10004, opened 2022-01-10
[FEA][JSON reader] to support parsing with single quotes


wbo4958:
This is part of FEA of https://github.com/NVIDIA/spark-rapids/issues/9
We have a JSON file

``` json
{'name': 'Reynold Xin'}
```

CUDF can't parse this file because of 
``` console
ai.rapids.cudf.CudfException: cuDF failure at: /home/bobwang/work.d/nvspark/cudf/cpp/src/io/json/reader_impl.cu:609: Error determining column names.
```
Spark can parse it by default.

We expect there is a configure `allowSingleQuotes` to control this behavior.

revans2:
By default Spark allows single quotes by default so this is a blocker for us to enable JSON parsing in Spark by default.

GregoryKimball:
After some discussion it seems we could support this by adding new states for single-quoted field names and single-quoted string values. We would have to increase the size of the state machine which would bring a performance penalty. I'll leave this in backlog for now.

revans2:
I think this is done now. @GregoryKimball @andygrove do you both agree?

GregoryKimball:
Closed by #14729. Thank you @shrshi!

PR #14545, opened 2023-12-02
JSON quote normalization

shrshi:
## Description
The goal of this PR is to address [PR 10004](https://github.com/rapidsai/cudf/issues/10004) by supporting parsing of JSON files containing single quotes for field/value strings. 

## Checklist


cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,328 @@
...
+  {
+    switch (symbol) {
+      case '\"': return static_cast<uint32_t>(dfa_symbol_group_id::DOUBLE_QUOTE_CHAR);
+      case '\'': return static_cast<uint32_t>(dfa_symbol_group_id::SINGLE_QUOTE_CHAR);
+      case '\\': return static_cast<uint32_t>(dfa_symbol_group_id::ESCAPE_CHAR);
+      case '\n': return static_cast<uint32_t>(dfa_symbol_group_id::NEWLINE_CHAR);
+      default: return static_cast<uint32_t>(dfa_symbol_group_id::OTHER_SYMBOLS);
+    }
+  };
+};
```
elstehle, line 77: https://github.com/rapidsai/cudf/pull/14561 was merged, so we should be good to revert the workaround.
```suggestion
// The i-th string representing all the characters of a symbol group
std::array<std::vector<SymbolT>, NUM_SYMBOL_GROUPS - 1> const qna_sgs{{{'\"'}, {'\''}, {'\\'}, {'\n'}}};
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,328 @@
...
+    // -------- TRANSLATION TABLE ------------
+    // state | read_symbol -> output_symbols
+    // DQS   | *           -> *
+    // DEC   | *           -> *
+    // OOS   | '           -> "
+    // OOS   | *           -> *
+    // SQS   | "           -> \"
+    // SQS   | '           -> "
+    // SQS   | *           -> *
+    // SEC   | *           -> *
```
elstehle, line 108: My idea was that this would reflect _all_ transitions, and the below `SPECIAL CASES` would exclude the `<state>, * -> *` cases from this table. 

Currently this part of the table is not representative of the latest changes, where we now have `SQS | \\ -> <nop>`. Should we update this part of the table or just get rid of this part of the table? Do you think the special cases part is sufficient for future maintainers?

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,328 @@
...
+                   static_cast<SymbolOffsetT>(d_input.size()),
+                   output_gpu.device_ptr(),
+                   out_indexes_gpu.device_ptr(),
+                   output_gpu_size.device_ptr(),
+                   start_state,
+                   stream.value());
+
+  // Async copy results from device to host
+  output_gpu.device_to_host_async(stream.view());
+  out_indexes_gpu.device_to_host_async(stream.view());
```
elstehle, line 227:
```suggestion
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,328 @@
...
+  constexpr std::size_t single_item = 1;
+  cudf::detail::hostdevice_vector<SymbolT> output_gpu(input.size() * 2, stream_view);
+  cudf::detail::hostdevice_vector<SymbolOffsetT> output_gpu_size(single_item, stream_view);
+  cudf::detail::hostdevice_vector<SymbolOffsetT> out_indexes_gpu(input.size(), stream_view);
+
+  // Allocate device-side temporary storage & run algorithm
+  parser.Transduce(d_input.data(),
+                   static_cast<SymbolOffsetT>(d_input.size()),
+                   output_gpu.device_ptr(),
+                   out_indexes_gpu.device_ptr(),
```
elstehle, line 220: We are not interested in the indexes, since we simply transform the input. Please make sure to include `#include thrust/iterator/discard_iterator.h>`.
```suggestion
                   thrust::make_discard_iterator(),
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,328 @@
...
+    cudf::io::fst::detail::make_translation_functor(TransduceToNormalizedQuotes{}),
+    stream);
+
+  auto d_input_scalar = cudf::make_string_scalar(input);
+  auto& d_input       = static_cast<cudf::scalar_type_t<std::string>&>(*d_input_scalar);
+
+  // Prepare input & output buffers
+  constexpr std::size_t single_item = 1;
+  cudf::detail::hostdevice_vector<SymbolT> output_gpu(input.size() * 2, stream_view);
+  cudf::detail::hostdevice_vector<SymbolOffsetT> output_gpu_size(single_item, stream_view);
```
elstehle, line 213:
```suggestion
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,328 @@
...
+std::array<std::array<dfa_states, NUM_SYMBOL_GROUPS>, TT_NUM_STATES> const qna_state_tt{{
+  /* IN_STATE      "       '       \       \n    OTHER  */
+  /* TT_OOS */ {{TT_DQS, TT_SQS, TT_OOS, TT_OOS, TT_OOS}},
+  /* TT_DQS */ {{TT_OOS, TT_DQS, TT_DEC, TT_OOS, TT_DQS}},
+  /* TT_SQS */ {{TT_SQS, TT_OOS, TT_SEC, TT_OOS, TT_SQS}},
+  /* TT_DEC */ {{TT_DQS, TT_DQS, TT_DQS, TT_OOS, TT_DQS}},
+  /* TT_SEC */ {{TT_SQS, TT_SQS, TT_SQS, TT_OOS, TT_SQS}},
+}};
+// The DFA's starting state
+constexpr char start_state = static_cast<char>(TT_OOS);
```
elstehle, line 88: nit:
```suggestion
constexpr auto start_state = static_cast<char>(TT_OOS);
```

elstehle (commented):
Looks good so far! Left a few minor comments. 

I assume the next step is to integrate this preprocessing stage into the reader depending on a new reader option?

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,328 @@
...
+    else if (state_id == static_cast<StateT>(dfa_states::TT_SEC)) {
+      return (relative_offset == 0) ? '\\' : read_symbol;
+    }
+    // In all other cases we simply output the input symbol
+    /*
+    else if (!((match_id == static_cast<SymbolGroupT>(dfa_symbol_group_id::ESCAPE_CHAR)) &&
+            ((state_id == static_cast<StateT>(dfa_states::TT_SQS))))) {
+      return read_symbol;
+    }
+    */
```
hyperbolic2346, line 153:
```suggestion
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,328 @@
...
+  std::string output = R"({"NUMBER":100"0,"STRING":"SOMETHING"})";
+  run_test(input, output);
+}
+TEST_F(FstTest, GroundTruth_QuoteNormalization_Invalid6)
+{
+  std::string input  = R"({'a':'\\''})";
+  std::string output = R"({"a":"\\""})";
+  run_test(input, output);
+}
+
```
hyperbolic2346, line 327: Are we worried about malformed data like `}'a': 'b'{`?

hyperbolic2346 (changes requested):
This is looking good to me. Some comments and questions.

cpp/tests/io/fst/quote_normalization_test.cu
shrshi: I think we should retain the non-special transitions for completeness - we can construct the DFA by just reading the transition table. How about dividing the table into non-special and special transitions? We can also make the input symbols used explicit. For instance, `OOS | <s> \in Sigma\{'} -> <s>` in non-special and `OOS | ' -> "` in special.

cpp/tests/io/fst/quote_normalization_test.cu
shrshi: Yes, the input to the FST can be invalid JSONs as well. Thank you for suggesting the above input - I've added it to the tests.

cpp/tests/io/fst/quote_normalization_test.cu
shrshi: I've retained the `output_gpu_size` to compare against the expected output size. Should I remove this?

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,315 @@
...
+    return 1;
+  }
+};
+
+}  // namespace
+
+// Base test fixture for tests
+struct FstTest : public cudf::test::BaseFixture {};
+
+void run_test(std::string& input, std::string& output)
```
vuule, line 178: What is the plan for this PR in terms of code location? Here, in tests?

shrshi: We can place this code in `io/fst` and have the user read their JSON with a pre-process reader option, or have a separate API that reads this file from device buffer and runs this FST as a pre-processing step.

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,315 @@
...
+  NUM_SYMBOL_GROUPS   ///< Total number of symbol groups
+};
+// Aliases for readability of the transition table
+constexpr auto TT_OOS            = dfa_states::TT_OOS;
+constexpr auto TT_DQS            = dfa_states::TT_DQS;
+constexpr auto TT_SQS            = dfa_states::TT_SQS;
+constexpr auto TT_DEC            = dfa_states::TT_DEC;
+constexpr auto TT_SEC            = dfa_states::TT_SEC;
+constexpr auto TT_NUM_STATES     = static_cast<char>(dfa_states::TT_NUM_STATES);
+constexpr auto NUM_SYMBOL_GROUPS = static_cast<uint32_t>(dfa_symbol_group_id::NUM_SYMBOL_GROUPS);
```
ttnghia, line 61: Do these symbols, or any of them, exist in some of the cudf headers? If so, why don't include the header but redefine them?

cpp/tests/io/fst/quote_normalization_test.cu
ttnghia: Sorry I don't understand. Everything here is in a test file, so there is no new feature or API was added. What does this PR actually offer?

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,315 @@
...
+#include <string>
+#include <vector>
+
+namespace {
+
+// Type used to represent the atomic symbol type used within the finite-state machine
+using SymbolT = char;
+using StateT  = char;
+// Type sufficiently large to index symbols within the input and output (may be unsigned)
+using SymbolOffsetT = uint32_t;
```
ttnghia, line 44: As I know, these types all exist but they are scattered in various places. It is better to reorganize them then include their declaration instead of redefining them, which is kind of error-prone.

cpp/tests/io/fst/quote_normalization_test.cu
shrshi: The plan is to integrate this pre-processing FST into libcudf in a follow-up PR to enable parsing the single-quote variant of JSON. This PR constructs the normalizing FST and checks if it can correctly handle valid and invalid JSON cases.

ttnghia: If this is just a proof-of-concept that is implemented directly into the unit tests then I don't recommend merging it, as the new unit tests do not actually test any existing API. Only when we have a new API implemented (in the follow up PR) then we merge its corresponding unit tests.

vuule: We discussed this and decided it would be beneficial to merge this POC before holidays and then build up the API next year.
IMO the unit tests are pretty useful even with POC because of the `run_test` abstraction, which is the sole entry point. We can easily update the test util code to call the new API from `run_test`, without even changing each test.

cpp/tests/io/fst/quote_normalization_test.cu
ttnghia: Okay, since this is a POC, which will be reimplemented later, I'm fine to redefine the type aliasing above, but please add comments (with `TODO`) to remove them later in the follow up work.

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,317 @@
...
+    // Whether this transition translates to the escape sequence: \"
+    const bool outputs_escape_sequence =
+      (state_id == static_cast<StateT>(dfa_states::TT_SQS)) &&
+      (match_id == static_cast<SymbolGroupT>(dfa_symbol_group_id::DOUBLE_QUOTE_CHAR));
+    // Case when a double quote needs to be replaced by the escape sequence: \"
+    if (outputs_escape_sequence) {
+      return (relative_offset == 0) ? '\\' : '"';
+    }
+    // Case when a single quote needs to be replaced by a double quote
+    else if ((match_id == static_cast<SymbolGroupT>(dfa_symbol_group_id::SINGLE_QUOTE_CHAR)) &&
```
ttnghia, line 116: Since in each `if()` statement has a `return` statement, don't need to have the `else` keyword.
```suggestion
    if ((match_id == static_cast<SymbolGroupT>(dfa_symbol_group_id::SINGLE_QUOTE_CHAR)) &&
```
This also applies similarly for other `if()` below.

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,317 @@
...
+             ((state_id == static_cast<StateT>(dfa_states::TT_SEC)))) {
+      return '\'';
+    }
+    // Case when an escaped symbol <s> that is not a single-quote needs to be replaced with \<s>
+    else if (state_id == static_cast<StateT>(dfa_states::TT_SEC)) {
+      return (relative_offset == 0) ? '\\' : read_symbol;
+    }
+    // In all other cases we simply output the input symbol
+    else
```
ttnghia, line 139:
```suggestion
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,317 @@
...
+      return read_symbol;
+  }
+  /**
```
ttnghia, line 142: Add a blank line after each function.

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,317 @@
...
+  constexpr CUDF_HOST_DEVICE int32_t operator()(StateT const state_id,
+                                                SymbolGroupT const match_id,
+                                                SymbolT const read_symbol) const
+  {
+    // Whether this transition translates to the escape sequence: \"
+    const bool sqs_outputs_escape_sequence =
+      (state_id == static_cast<StateT>(dfa_states::TT_SQS)) &&
+      (match_id == static_cast<SymbolGroupT>(dfa_symbol_group_id::DOUBLE_QUOTE_CHAR));
+    // Number of characters to output on this transition
+    if (sqs_outputs_escape_sequence) return 2;
```
ttnghia, line 158: Humnn, I thought that `clang-format` should automatically add the braces like this:
```suggestion
    if (sqs_outputs_escape_sequence) { return 2; }
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,317 @@
...
+    // Whether this transition translates to the escape sequence \<s> or unescaped '
+    const bool sec_outputs_escape_sequence =
+      (state_id == static_cast<StateT>(dfa_states::TT_SEC)) &&
+      (match_id != static_cast<SymbolGroupT>(dfa_symbol_group_id::SINGLE_QUOTE_CHAR));
+    // Number of characters to output on this transition
+    if (sec_outputs_escape_sequence) return 2;
```
ttnghia, line 164:
```suggestion
    if (sec_outputs_escape_sequence) { return 2; }
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,317 @@
...
+    // Whether this transition translates to no output <nop>
+    const bool sqs_outputs_nop =
+      (state_id == static_cast<StateT>(dfa_states::TT_SQS)) &&
+      (match_id == static_cast<SymbolGroupT>(dfa_symbol_group_id::ESCAPE_CHAR));
+    // Number of characters to output on this transition
+    if (sqs_outputs_nop) return 0;
```
ttnghia, line 170:
```suggestion
    if (sqs_outputs_nop) { return 0; }
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,317 @@
...
+  /* IN_STATE      "       '       \       \n    OTHER  */
+  /* TT_OOS */ {{TT_DQS, TT_SQS, TT_OOS, TT_OOS, TT_OOS}},
+  /* TT_DQS */ {{TT_OOS, TT_DQS, TT_DEC, TT_OOS, TT_DQS}},
+  /* TT_SQS */ {{TT_SQS, TT_OOS, TT_SEC, TT_OOS, TT_SQS}},
+  /* TT_DEC */ {{TT_DQS, TT_DQS, TT_DQS, TT_OOS, TT_DQS}},
+  /* TT_SEC */ {{TT_SQS, TT_SQS, TT_SQS, TT_OOS, TT_SQS}},
+}};
+// The DFA's starting state
+constexpr char start_state = static_cast<char>(TT_OOS);
+struct TransduceToNormalizedQuotes {
```
ttnghia, line 78:
```suggestion

struct TransduceToNormalizedQuotes {
```

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,317 @@
...
+  CUDF_TEST_EXPECT_VECTOR_EQUAL(output_gpu, output, output.size());
+}
+
+TEST_F(FstTest, GroundTruth_QuoteNormalization1)
+{
+  std::string input  = R"({"A":'TEST"'})";
+  std::string output = R"({"A":"TEST\""})";
+  run_test(input, output);
+}
+TEST_F(FstTest, GroundTruth_QuoteNormalization2)
```
ttnghia, line 231: There should be one blank line before/after each `TEST_F`. I'm not sure why `clang-format` doesn't enforce this.

ttnghia (approved):

hyperbolic2346 (approved):

cpp/tests/io/fst/quote_normalization_test.cu
elstehle: That makes sense; we should keep it. I think I wanted to suggest to remove allocating memory for the output indexes and I may have just hit the wrong line here.

cpp/tests/io/fst/quote_normalization_test.cu
```diff
@@ -0,0 +1,335 @@
...
+  output_gpu_size.device_to_host_async(stream_view);
+
+  // Make sure results have been copied back to host
+  stream.synchronize();
+
+  // Verify results
+  std::cout << "Expected output: " << output << std::endl << "Computed output: ";
+  for (size_t i = 0; i < output_gpu_size[0]; i++)
+    std::cout << output_gpu[i];
+  std::cout << std::endl;
```
elstehle, line 224: suggestion: I think we want to remove printing to stdout to not pollute test logs?

elstehle (approved):

PR #14553, opened 2023-12-04
Enforce `Index.to_frame` deprecations

galipremsagar:
## Description
This PR enforces deprecations of `Index.to_frame` and updates pytests related to this API.

## Checklist

### DIFF
--- a/python/cudf/cudf/core/_base_index.py
+++ b/python/cudf/cudf/core/_base_index.py
@@ -796,12 +796,6 @@ class BaseIndex(Serializable):
 
         return super().fillna(value=value)
 
-    def _get_level_names(self):
-        """
-        Return a name or list of names with None replaced by the level number.
-        """
-        return 0 if self.name is None else self.name
-
     def to_frame(self, index=True, name=no_default):
         """Create a DataFrame with a column containing this Index
 
@@ -852,7 +846,7 @@ class BaseIndex(Serializable):
         """
 
         if name is no_default:
-            col_name = self._get_level_names()
+            col_name = 0 if self.name is None else self.name
         else:
             col_name = name
 
--- a/python/cudf/cudf/core/_base_index.py
+++ b/python/cudf/cudf/core/_base_index.py
@@ -796,6 +796,12 @@ class BaseIndex(Serializable):
 
         return super().fillna(value=value)
 
+    def _get_level_names(self):
+        """
+        Return a name or list of names with None replaced by the level number.
+        """
+        return 0 if self.name is None else self.name
+
     def to_frame(self, index=True, name=no_default):
         """Create a DataFrame with a column containing this Index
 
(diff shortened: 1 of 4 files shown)


python/cudf/cudf/core/_base_index.py
```diff
@@ -796,6 +796,12 @@ def fillna(self, value, downcast=None):
 
         return super().fillna(value=value)
 
+    def _get_level_names(self):
+        """
+        Return a name or list of names with None replaced by the level number.
+        """
+        return 0 if self.name is None else self.name
```
wence-, line 803: This docstring is a "generic" one, but this function never returns a list.

python/cudf/cudf/core/multiindex.py
```diff
@@ -970,6 +969,16 @@ def __getitem__(self, index):
...
 
+    @_cudf_nvtx_annotate
+    def _get_level_names(self):
+        """
+        Return a name or list of names with None replaced by the level number.
+        """
+        return [
+            level if name is None else name
+            for level, name in enumerate(self.names)
+        ]
```
wence-, line 980: And this one always returns a list.

Would it be more natural to always return a list, and in the non-multiindex case just have a singleton list that the caller destructures?

python/cudf/cudf/core/_base_index.py
```diff
@@ -844,22 +850,12 @@ def to_frame(self, index=True, name=no_default):
...
-                FutureWarning,
-            )
-            name = no_default
-        if name is not no_default:
-            col_name = name
-        elif self.name is None:
-            col_name = 0
+
+        if name is no_default:
+            col_name = self._get_level_names()
```
wence-, line 855:
```suggestion
            (col_name,) = self._get_level_names()
```

Like that.

python/cudf/cudf/core/multiindex.py
mroeschke: Yeah IMO if these functions are only needed in `to_frame` and Index vs MultiIndex have their own implementation, I think hardcoding this in the function is sufficient

python/cudf/cudf/core/_base_index.py
galipremsagar: Removed this part of code

python/cudf/cudf/core/multiindex.py
galipremsagar: Went ahead and removed these single use functions.

python/cudf/cudf/core/_base_index.py
galipremsagar: Dropped this method altogether.

wence-:
This is pandas-2, so I'm not expecting clean CI, right?

wence- (approved):

galipremsagar:
> This is pandas-2, so I'm not expecting clean CI, right?

Yup, That's right.

PR #14558, opened 2023-12-04
Use cudf_test temp_directory class for nvtext::subword_tokenize gbenchmark

davidwendt:
## Description
Changes the creation of a temporary subword hash file to use the `temp_directory` class from `cudf_test/file_utilities.hpp`.
This is part of an overall effort to consolidate and document libcudf environment variables.

## Checklist

### DIFF
--- a/cpp/benchmarks/text/subword.cpp
+++ b/cpp/benchmarks/text/subword.cpp
@@ -29,10 +29,9 @@
 #include <iostream>
 #include <vector>
 
-temp_directory const subword_tmpdir{"cudf_gbench"};
-
 static std::string create_hash_vocab_file()
 {
+  static temp_directory const subword_tmpdir{"cudf_gbench"};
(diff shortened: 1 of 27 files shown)


cpp/benchmarks/text/subword.cpp
```diff
@@ -18,19 +18,22 @@
...
 #include <cudf/strings/strings_column_view.hpp>
+
 #include <nvtext/subword_tokenize.hpp>
 
 #include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <vector>
 
+temp_directory const subword_tmpdir{"cudf_gbench"};
```
harrism, line 32: question: Does this need to be global? 

suggestion: Seems this is only used in one static function, so maybe move the variable inside that function?

harrism (approved):
One minor suggestion.

hyperbolic2346 (approved):

PR #14565, opened 2023-12-04
Simplify Python CMake

vyasr:
## Description

This PR makes a number of changes to streamline the CMake code in the Python build. There are additional potential improvements that may be possible but I'd prefer to wait on until after the scikit-build-core migration because I suspect that there may be subtle differences (particularly around the install rules). The major improvements here are:
- nvcomp installation no longer needs as much special casing thanks to https://github.com/rapidsai/rapids-cmake/pull/496 adding missing targets
- get_arrow.cmake now determines how to find arrow by checking for 1) the presence of Python, 2) the presence of pyarrow, and 3) the presence of libarrow.so.* inside the pyarrow directory. This approach works for both pip and conda packages as well as pure C++ builds and removes the need for manual determination of where to look by the builder. The assumption baked in is that the developer has installed the desired pyarrow package when building.
- The `CUDF_BUILD_WHEELS` variable is now removed. All Python builds that don't find the C++ library all go down the same path now. This is particularly relevant for doing something like a local `pip install`. This is the desired behavior because if you're building the Python package in that kind of environment you always want the same behaviors. The different case is when the Python build finds the C++ build (e.g. in a conda environment where libcudf is a separate package).
- The logic for linking to pyarrow headers is now centralized in a single function since it needs to be used by every Cython target.

## Checklist


python/cudf/CMakeLists.txt
```diff
@@ -74,42 +65,32 @@ include(rapids-cython)
...
+  # control where they are installed to. Since there are multiple subpackages of cudf._lib that
+  # require access to libcudf, we place the library and all its dependent artifacts in the cudf
+  # directory as a single source of truth and modify the other rpaths appropriately.
   set(cython_lib_dir cudf)
-  install(TARGETS cudf DESTINATION ${cython_lib_dir})
+  include(cmake/Modules/WheelHelpers.cmake)
+  # TODO: This install is currently overzealous. We should only install the libraries that are
+  # downloaded by CPM during the build, not libraries that were found on the system.  However, in
+  # practice right this would only be a problem is if libcudf was not found but some of the
+  # dependencies were, and we have no real use cases where that happens.
```
vyasr, line 84: I don't want to try and fix this right now. I would like to manage the scikit-build-core transition first, then revisit the best way for us to handle install rules. Some anecdotal experiments suggest that the configure/build paths might be a bit different in scikit-build-core, so I don't want to do redundant work.

raydouglass (approved):

brandon-b-miller (approved):

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -34,20 +34,34 @@ function(find_libarrow_in_python_wheel PYARROW_VERSION)
   # version number soname, just `${MAJOR_VERSION}00`
   set(PYARROW_LIB "libarrow.so.${PYARROW_SO_VER}00")
 
-  find_package(Python REQUIRED)
```
robertmaynard, line 37: We should check to make sure `Python_EXECUTABLE` is defined and error out otherwise. In theory, a single scope could have included `get_arrow.cmake` and a different scope calls `find_and_configure_arrow` so we end up with an invalid `execute_process`

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -34,20 +34,34 @@ function(find_libarrow_in_python_wheel PYARROW_VERSION)
...
   execute_process(
```
robertmaynard, line 37: Why do we need this execute_process here? This looks the same as the initial_code_block that will be called on line 67 once we find Arrow?

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -34,20 +34,34 @@ function(find_libarrow_in_python_wheel PYARROW_VERSION)
...
     Arrow NO_CONFIG
     VERSION "${PYARROW_VERSION}"
     LIBRARY_NAMES "${PYARROW_LIB}"
     BUILD_EXPORT_SET cudf-exports
     INSTALL_EXPORT_SET cudf-exports
-    HEADER_NAMES arrow/python/arrow_to_pandas.h
+    HEADER_NAMES arrow/python/arrow_to_pandas.h INITIAL_CODE_BLOCK initial_code_block
   )
 
   find_package(Arrow ${PYARROW_VERSION} MODULE REQUIRED GLOBAL)
```
robertmaynard, line 67: Nitpick: can you fix the uppercase commands between lines 79:98

cpp/cmake/thirdparty/get_arrow.cmake
```diff
@@ -432,7 +446,35 @@ if(NOT DEFINED CUDF_VERSION_Arrow)
   )
 endif()
 
+# If the pyarrow package contains libarrow we want to link against it directly instead of searching
```
robertmaynard, line 449: This logic should be explicitly opt-in.
If you are building libcudf on a machine with a system installed pyarrow you would always use that with no way to opt out.

robertmaynard (changes requested):

cpp/cmake/thirdparty/get_arrow.cmake
vyasr: Ugh you're right, but this is unfortunate. It does mean we continue to have to specify nonzero information from the CLI when invoking certain (wheel) builds, but I guess that's not the worst.

robertmaynard (approved):

PR #14566, opened 2023-12-04
Switch to static libcudart builds

vyasr:
This PR changes the default for C++ builds to statically link to libcudart.


vyasr:
One obvious issue arising here, the stream checking preload won't work with a static cudart.

davidwendt:
Why is this being changed? Can you add more detail to the description?

vyasr:
Closing, will require some more planning.

PR #14576, opened 2023-12-05
Update to CCCL 2.2.0.

bdice:
## Description
This PR updates cuDF to CCCL 2.2.0. Do not merge until all of RAPIDS is ready to update.

Depends on #14577.

Replaces #13222.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #14577, opened 2023-12-05
Use cuda::proclaim_return_type on device lambdas.

bdice:
## Description
This PR makes cudf compatible with Thrust 2 by adding `cuda::proclaim_return_type`, but does not upgrade the version of Thrust/CCCL just yet. Currently we use libcudacxx 2.1.0, which makes `cuda::proclaim_return_type` available, but we still use Thrust 1.17.2 which doesn't require device lambdas to have proclaimed return types. This diff is separated out from #14576 and reduces the diff we must carry in #14576 -- which should contain only packaging changes / version updates for the CCCL 2 migration. This PR is **nonbreaking**, while #14576 will be **breaking**.

## Checklist


cpp/include/cudf/detail/utilities/cast_functor.cuh
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * @brief Functor that casts another functor's result to a specified type.
+ *
+ * CUB 2.0.0 reductions require that the binary operator returns the same type
+ * as the initial value type, so we wrap binary operators with this when used
+ * by CUB.
+ */
+template <typename ResultType, typename F>
+struct cast_functor_fn {
```
ttnghia, line 42: Why do we have this new file?

ttnghia (approved):

cpp/src/copying/copy.cu
```diff
@@ -35,6 +35,8 @@
 #include <thrust/distance.h>
 #include <thrust/iterator/counting_iterator.h>
 
+#include <cuda/functional>
```
davidwendt, line 38: Doesn't look like this header is needed.

cpp/src/io/json/nested_json_gpu.cu
```diff
@@ -45,6 +45,8 @@
 #include <thrust/transform.h>
 #include <thrust/tuple.h>
 
+#include <cuda/functional>
+
```
davidwendt, line 49:
```suggestion
```

cpp/src/quantiles/tdigest/tdigest_aggregation.cu
```diff
@@ -623,17 +630,18 @@ std::unique_ptr<column> build_output_column(size_type num_rows,
...
-                    thrust::make_counting_iterator(0) + num_rows,
-                    sizes.begin(),
-                    [offsets = offsets->view().begin<size_type>()] __device__(size_type i) {
-                      return offsets[i + 1] - offsets[i];
-                    });
+  auto offsets_iter = offsets->view().begin<size_type>();
+  thrust::adjacent_difference(rmm::exec_policy(stream),
+                              offsets->view().begin<size_type>() + 1,
+                              offsets->view().end<size_type>(),
+                              sizes.begin());
```
davidwendt, line 637: This rewrite assumes that the first offset value is always 0 while the original transform code does not.

davidwendt (approved):

vuule (commented):
never finished the review :(
Wanted to asked about this comment anyway.

cpp/include/cudf/column/column_view.hpp
```diff
@@ -478,7 +478,10 @@ class mutable_column_view : public detail::column_view_base {
  public:
   mutable_column_view() = default;
 
-  ~mutable_column_view() = default;
+  ~mutable_column_view(){
+    // Needed so that the first instance of the implicit destructor for any TU isn't 'constructed'
+    // from a host+device function marking the implicit version also as host+device
```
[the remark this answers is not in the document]
bdice: That's a comment we acquired with help from @robertmaynard. I was suffering from inexplicable compilation problems and this fixed it.

cpp/include/cudf/column/column_view.hpp
```diff
@@ -478,7 +478,10 @@ class mutable_column_view : public detail::column_view_base {
...
+  };
```
ttnghia, line 484: This will trigger IDE warning.

```suggestion
  }
```

PR #14559, opened 2023-12-04
Add column sanitization checks in `CUDF_TEST_EXPECT_COLUMN_*` macros

SurajAralihalli:
This PR addresses Issue #[12786](https://github.com/rapidsai/cudf/issues/12786)

The listed functions have been modified to incorporate a column sanitization check; otherwise, they will raise a `std::invalid_argument` error.
- `expect_column_properties_equal`
- `expect_column_properties_equivalent`
- `expect_columns_equal`
- `expect_columns_equivalent`

## Checklist
- [X] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #14578, opened 2023-12-05
Fix unsanitized nulls produced by libcudf dictionary decode

davidwendt:
## Description
Fixes `cudf::dictionary::decode` logic to produced sanitized null entries for compound column types.

Reference: #14559 -- fixes many of the errors found here concerning dictionary column gtests.

## Checklist

### DIFF
--- a/cpp/src/dictionary/decode.cu
+++ b/cpp/src/dictionary/decode.cu
@@ -34,7 +34,7 @@ struct indices_handler_fn {
   cudf::detail::input_indexalator const d_iterator;
   column_device_view const d_indices;
   size_type oob_index;  // out-of-bounds index identifies nulls
-  __device__ size_type operator()(size_type idx)
+  __device__ size_type operator()(size_type idx) const
   {
     return d_indices.is_null(idx) ? oob_index : d_iterator[idx];
   }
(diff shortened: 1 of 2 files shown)


cpp/src/dictionary/decode.cu
```diff
@@ -30,6 +29,18 @@
 namespace cudf {
 namespace dictionary {
 namespace detail {
+namespace {
+struct indices_handler_fn {
+  cudf::detail::input_indexalator const d_iterator;
+  column_device_view const d_indices;
+  size_type oob_index;  // out-of-bounds index identifies nulls
+  __device__ size_type operator()(size_type idx)
```
ttnghia, line 37:
```suggestion
  __device__ size_type operator()(size_type idx) const
```

ttnghia (approved):

cpp/src/dictionary/decode.cu
```diff
@@ -30,6 +29,18 @@
 namespace cudf {
 namespace dictionary {
 namespace detail {
+namespace {
+struct indices_handler_fn {
```
mythrocks, line 33: Thanks for fixing this.  Would there be value in naming this functor a "nullifying_handler" or some such?

mythrocks (approved):
LGTM!

cpp/src/dictionary/decode.cu
davidwendt: The actually nullifying happens far away in the gather function. This function simply returns indices.

PR #14582, opened 2023-12-06
Use exceptions instead of return values to handle errors in `CompactProtocolReader`

vuule:
## Description
Align the coding style with the rest of libcudf and use exceptions to report errors instead of returning a bool. 
This simplifies the control flow in many functions. In addition, the return value used top be ignored at many call sites.

## Checklist


etseidl (commented):
Love this.  Next up snake_case names! 😄 

Here are a few unasked for comments.

cpp/src/io/parquet/compact_protocol_reader.cpp
```diff
@@ -61,16 +63,15 @@ class parquet_field_list : public parquet_field {
   }
 
  public:
-  inline bool operator()(CompactProtocolReader* cpr, int field_type)
+  inline void operator()(CompactProtocolReader* cpr, int field_type)
   {
-    if (field_type != ST_FLD_LIST) { return true; }
+    CUDF_EXPECTS(field_type == ST_FLD_LIST, "expected list field");
```
vuule, line 68: Note to reviewers - returning true signals failure in these functions.

cpp/src/io/parquet/compact_protocol_reader.cpp
```diff
@@ -463,68 +443,64 @@ bool CompactProtocolReader::skip_struct_field(int t, int depth)
...
   constexpr int index = std::tuple_size<std::tuple<Operator...>>::value - 1;
   int field           = 0;
   while (true) {
     int const current_byte = cpr->getb();
     if (!current_byte) { break; }
-    int const field_delta    = current_byte >> 4;
-    int const field_type     = current_byte & 0xf;
-    field                    = field_delta ? field + field_delta : cpr->get_i16();
-    bool const exit_function = FunctionSwitchImpl<index>::run(cpr, field_type, field, op);
-    if (exit_function) { return false; }
```
vuule, line 522: This is where the logic inverts and false signals failure. This logic is(was) used from here to the top-level `read` functions.

harrism (approved):

ttnghia (approved):

cpp/src/io/parquet/compact_protocol_reader.cpp
robertmaynard: We should state what kind of field we are expecting, be it integers or strings.

vuule: I don't follow, we do state in the error string that we expect a list.

cpp/src/io/parquet/compact_protocol_reader.cpp
```diff
@@ -127,12 +127,24 @@ class parquet_field_int : public parquet_field {
 
   T& val;
 
+  std::string type_string()
```
ttnghia, line 130:
```suggestion
  std::string type_string() const
```

ttnghia: It's better to use switch:
```
switch (EXPECTED_TYPE) {
 case ....: return ...;
}
```

cpp/src/io/parquet/compact_protocol_reader.cpp
```diff
@@ -127,17 +127,29 @@ class parquet_field_int : public parquet_field {
...
+  {
+    if constexpr (EXPECTED_TYPE == ST_FLD_BYTE) {
+      return "byte";
+    } else if constexpr (EXPECTED_TYPE == ST_FLD_I32) {
+      return "int32";
+    } else if constexpr (EXPECTED_TYPE == ST_FLD_I64) {
+      return "int64";
+    }
+    return "unknown(EXPECTED_TYPE=" + std::to_string(EXPECTED_TYPE) + ")";
```
etseidl, line 139: could this be an "else" and then the function made constexpr?

vuule: I don't know if it makes sense to make the function constexpr when it returns a `string`. A previous version returned a `char const*`, but CUDF_EXPECTS didn't like that.

etseidl: Guess that's a C++20 thing 😅. Someday...

cpp/src/io/parquet/compact_protocol_reader.cpp
robertmaynard: If the correct field is a list of strings, when provided a list of integers we error saying "expected list field", which is the wrong message. We should state we expected a list of strings ( even better also state what we got )

robertmaynard: I thought I left my original comment on line 70, where we do the check on the `T` type of the list, which is what should have the better diagnostic.

cpp/src/io/parquet/compact_protocol_reader.cpp
vuule: changed to a switch when I generalized the function so it can be reused

cpp/src/io/parquet/compact_protocol_reader.cpp
vuule: Centralized the field type checks, now all of them should print detailed info about the failing check.

cpp/src/io/parquet/compact_protocol_reader.cpp
```diff
@@ -40,37 +42,57 @@ class parquet_field {
...
+    case ST_FLD_STRUCT: return "struct";
+    case ST_FLD_LIST: return "list";
+    case ST_FLD_SET: return "set";
+    default: return "unknown(" + std::to_string(type) + ")";
+  }
+}
+
+void assert_field_type(int type, FieldType expected)
+{
+  CUDF_EXPECTS(type == expected,
```
ttnghia, line 65: Oh no, this is comparing `int` vs `enum`.

cpp/src/io/parquet/compact_protocol_reader.cpp
```diff
@@ -40,37 +42,57 @@ class parquet_field {
...
-  {
-  }
+  parquet_field_list(int f, std::vector<T>& v) : parquet_field(f), val(v) {}
 
  public:
-  inline bool operator()(CompactProtocolReader* cpr, int field_type)
+  inline void operator()(CompactProtocolReader* cpr, int field_type)
   {
-    if (field_type != ST_FLD_LIST) { return true; }
+    assert_field_type(field_type, ST_FLD_LIST);
```
ttnghia, line 89: Probably we are wrong from the beginning. Why not `field_type` an enum value, instead of `int`?

ttnghia: In addition, should the caller handle the input type instead of checking it here?

robertmaynard (approved):

cpp/src/io/parquet/compact_protocol_reader.cpp
vuule: We always compared `int`s and `FieldType` in this code, e.g. `if (field_type != ST_FLD_LIST) { return true; }`. I think it's fine, given that we throw if the values are not equal - no int values that are not valid `FieldType` values will get past this check. We have to sanitize the input somewhere :man_shrugging:

cpp/src/io/parquet/compact_protocol_reader.cpp
vuule: `field_type` is obtained by the caller by parsing the encoded input, so it could be any value if the input file is garbage. We cannot assume it will be a valid FieldType.

As for the validation on the caller side - the caller only sees these parquet_field_xyz objects as functors and does not know the actual field type that is being read. We could move validation to the caller if `parquet_field` classes somehow expose the expected field type, but I don't think this is a better option.

cpp/src/io/parquet/compact_protocol_reader.cpp
ttnghia: We definitely need to refactor it, changing into `enum class` for `FieldType`.

vuule: I'm not following. We have to compare the int input with a FieldType value somewhere. Using an enum class helps with the scope of the value names, but we still need the int <-> enum comparison, just with a `static_cast`.

ttnghia: Then we will do an explicit `static_cast` before comparison. I think it is better than this implicit comparison.

vuule: Just checked, `FieldType` is not exactly part of the spec, we have some creative freedom around it. 
If you don't mind, let's make it an `enum class` in a separate PR. It's a wider change than the error handling this PR is about.

PR #14586, opened 2023-12-06
Fix unsanitized nulls from strings segmented-reduce

davidwendt:
## Description
Fixes the string specialization logic in `cudf::segmented_reduce` to not produce unsanitized null entries. The functor used to build a gather map for argmin/argmax was corrected to handle include/exclude nulls correctly.

Reference: #14559 

## Checklist

### DIFF
--- a/cpp/src/reductions/segmented/simple.cuh
+++ b/cpp/src/reductions/segmented/simple.cuh
@@ -24,7 +24,6 @@
 #include <cudf/detail/gather.hpp>
 #include <cudf/detail/unary.hpp>
 #include <cudf/detail/utilities/cuda.cuh>
-#include <cudf/detail/utilities/element_argminmax.cuh>
 #include <cudf/detail/valid_if.cuh>
 #include <cudf/reduction/detail/segmented_reduction.cuh>
 #include <cudf/types.hpp>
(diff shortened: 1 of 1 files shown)


cpp/src/reductions/segmented/simple.cuh
```diff
@@ -114,6 +113,23 @@ std::unique_ptr<column> simple_segmented_reduction(
...
+template <typename T>
+struct reduce_argminmax_fn {
+  column_device_view const d_col;  // column data
+  bool const arg_min;              // true if argmin, otherwise argmax
+  null_policy null_handler;        // include or exclude nulls
+
+  __device__ inline auto operator()(size_type const& lhs_idx, size_type const& rhs_idx) const
+  {
+    if (lhs_idx < 0 || lhs_idx >= d_col.size()) { return rhs_idx; }  // CUB segmented reduce
+    if (rhs_idx < 0 || rhs_idx >= d_col.size()) { return lhs_idx; }  // calls with OOB indices
```
bdice, line 125: I really struggle to understand comments that are split across lines since it looks like two separate sentences, neither of which have enough context. I would prefer to have one comment line, before the relevant code.
```suggestion
    // CUB segmented reduce calls with out-of-bounds indices
    if (lhs_idx < 0 || lhs_idx >= d_col.size()) { return rhs_idx; }  
    if (rhs_idx < 0 || rhs_idx >= d_col.size()) { return lhs_idx; }
```

bdice (approved):

PR #14588, opened 2023-12-06
Remove WORKSPACE env var from cudf_test temp_directory class

davidwendt:
## Description
Removes the `WORKSPACE` env var check from the `cudf_test` `temp_directory` class. This env var is not being used. This is part of a larger effort to consolidate and document libcudf env vars.

## Checklist

### DIFF
--- a/cpp/include/cudf_test/file_utilities.hpp
+++ b/cpp/include/cudf_test/file_utilities.hpp
@@ -16,14 +16,13 @@
 
 #pragma once
 
+#include <cudf/utilities/error.hpp>
+
 #include <cstdio>
 #include <cstdlib>
 #include <filesystem>
-#include <string>
-
 #include <ftw.h>
-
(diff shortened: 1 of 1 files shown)


ttnghia (approved):

vuule:
I actually used this when I wanted to save the generated files in IO benchmarks.
I guess we can always hard code a non-temp-dir value for experiments, but this was nice.

vuule:
from https://en.cppreference.com/w/cpp/filesystem/temp_directory_path:

> On POSIX systems, the path may be the one specified in the environment variables TMPDIR, TMP, TEMP, TEMPDIR, and, if none of them are specified, the path "/tmp" is returned.

So that should work fine as well. I'll give it a try.

vuule:
`TMPDIR` works fine. So this was just reinventing the wheel :)

vuule (approved):

PR #14590, opened 2023-12-06
Add DELTA_LENGTH_BYTE_ARRAY encoder and decoder for Parquet

etseidl:
## Description
Part of #13501. This adds the ability to read and write Parquet pages with DELTA_LENGTH_BYTE_ARRAY encoding.

## Checklist


etseidl:
Unlike past PRs, in this PR I've done both the encoder and decoder. One reason is that the changes here are smaller than in previous delta work. A better reason is that by adding both, I can test the reader and writer at the same time without having to rely on python tests.

That said, this PR does expand the existing python tests anyway...perhaps too much. Reviewers, please let me know if the delta tests in `test_parquet.py` are a little overboard.

A note on the encoding itself. DELTA_LENGTH_BYTE_ARRAY is much like PLAIN encoding for strings.  The difference is that the lengths are encoded first, and then the string data. The theory is that separating the two will lead to better compression. The other difference is that the lengths are DELTA_BINARY_PACKED encoded, so there should be a space savings.

Because the string data is in one big chunk per page, the processing is much simpler than with DELTA_BYTE_ARRAY or even PLAIN.  On the decode side, all we need to do is decode the length data into the offsets child column, and then do an exclusive scan on this to turn lengths into offsets. The character data can just be parallel memcpy'd straight into the chars child column. On the encoding side, it's a similar story. First encode the length data, and then memcpy the chars data into the page buffer. Of course, skip_rows complicates things some, but not too much.

cpp/tests/io/parquet_test.cpp
```diff
@@ -7091,4 +7091,203 @@ TEST_F(ParquetReaderTest, RepeatedNoAnnotations)
...
+
+// make a random list<string> column, with random string lengths of 0..max_string_len,
+// and up to max_vals_per_row strings in each list.
+std::unique_ptr<cudf::column> make_parquet_string_list_col(int num_rows,
+                                                           int max_vals_per_row,
+                                                           int max_string_len,
+                                                           bool include_validity)
+{
+  constexpr auto seed = 21337;
+  static std::mt19937 engine{seed};
```
vuule, line 7171: because we use the same seed in every function call, I think we'll end up with a lot of similarities between columns (e.g. pretty sure all null masks are the same). We also cannot have a global sequence of seeds as changes in tests/test order will lead to data changes.
I think these tests need a seed that will change in a deterministic way for every column. This value would be passed when creating columns and modified for the next column. In benchmark data gen, we use another rendom engine to generate seeds, but a simple increment might be perfectly fine here as well.

etseidl: Cleaned this up some.

python/cudf/cudf/tests/test_parquet.py
```diff
@@ -1352,8 +1352,13 @@ def test_delta_binary(nrows, add_nulls, dtype, tmpdir):
 
 @pytest.mark.parametrize("nrows", delta_num_rows())
 @pytest.mark.parametrize("add_nulls", [True, False])
-@pytest.mark.parametrize("str_encoding", ["DELTA_BYTE_ARRAY"])
-def test_delta_byte_array_roundtrip(nrows, add_nulls, str_encoding, tmpdir):
+@pytest.mark.parametrize("string_len", [12, 48, 96, 128])
```
etseidl, line 1355: Should this instead be four columns in the dataframe? I'm trying to test the different string copy mechanisms. Same question for nulls.

vuule: This seems better to me, easier to see what failed from the parameter values. Unless you specifically want to test multiple copy mechanisms running at the same time.
Although this may be a question for the actual Python reviewer(s) :)

vuule (commented):
partial review, I guess
had to flush a review comment to reply :\

python/cudf/cudf/tests/test_parquet.py
etseidl: This is finer grained, for sure, but leads to 8X the invocations of this test.  Pretty soon this library will be cuParquet plus some other stuff :rofl:

cpp/src/io/parquet/parquet_gpu.hpp
```diff
@@ -474,9 +480,10 @@ constexpr uint32_t encoding_to_mask(Encoding encoding)
  * Used to control which encode kernels to run.
  */
 enum class encode_kernel_mask {
-  PLAIN        = (1 << 0),  // Run plain encoding kernel
-  DICTIONARY   = (1 << 1),  // Run dictionary encoding kernel
-  DELTA_BINARY = (1 << 2)   // Run DELTA_BINARY_PACKED encoding kernel
+  PLAIN           = (1 << 0),  // Run plain encoding kernel
+  DICTIONARY      = (1 << 1),  // Run dictionary encoding kernel
+  DELTA_BINARY    = (1 << 2),  // Run DELTA_BINARY_PACKED encoding kernel
+  DELTA_LENGTH_BA = (1 << 3),  // Run DELTA_LENGTH_BYTE_ARRAY encoding kernel
```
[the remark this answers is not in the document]
etseidl: It will make room. Delta byte array is Klingon!

python/cudf/cudf/tests/test_parquet.py
vuule: > Pretty soon this library will be cuParquet plus some other stuff 🤣

And that's only the beginning. One day RAPIDS will stand for Reading All Parquet Inputs for Data Science.

cpp/src/io/parquet/delta_binary.cuh
```diff
@@ -101,7 +101,7 @@ struct delta_binary_decoder {
   uint8_t const* cur_mb_start;   // pointer to the start of the current mini-block data
   uint8_t const* cur_bitwidths;  // pointer to the bitwidth array in the block
 
-  uleb128_t value[delta_rolling_buf_size];  // circular buffer of delta values
+  zigzag128_t value[delta_rolling_buf_size];  // circular buffer of delta values
```
ttnghia, line 104: Will this be a breaking change?

cpp/src/io/parquet/delta_binary.cuh
```diff
@@ -297,6 +297,51 @@ struct delta_binary_decoder {
     }
   }
 
+  // Decodes and skips values until the block containing the value after `skip` is reached.
+  // Keeps a running sum of the values and returns that upon exit. called by all threads in a
```
ttnghia, line 301:
```suggestion
  // Keeps a running sum of the values and returns that upon exit. Called by all threads in a
```

cpp/src/io/parquet/page_delta_decode.cu
```diff
@@ -625,4 +777,30 @@ void __host__ DecodeDeltaByteArray(cudf::detail::hostdevice_vector<PageInfo>& pa
   }
 }
 
+/**
+ * @copydoc cudf::io::parquet::gpu::DecodeDeltaByteArray
+ */
+void __host__
```
ttnghia, line 783: Why do we need to mark it `__host__`?
```suggestion
void __host__
```

cpp/src/io/parquet/delta_binary.cuh
etseidl: should not be user visible...results are cast elsewhere anyway

cpp/src/io/parquet/page_delta_decode.cu
etseidl: It's an idiosyncrasy I picked up from the original DecodePageData. I guess to be explicit about what's host code and what's device code when they're mixed?

vuule: Yeah, a lot of older code has this. I don't include `__host__` in new code but I don't feel strongly about it. We should probably decide on the guideline and replace-all in one fell swoop.

vuule (commented):
partial review, need to continue fresh tomorrow
few nitpicks so far

cpp/src/io/parquet/page_delta_decode.cu
etseidl: I count 7 instances of its use in the parquet code, 3 of which were added by me :sweat_smile: I think we can get rid of them all as part of this PR. Failing that I can at least get rid of the ones in page_delta_decode.cu

cpp/src/io/parquet/delta_binary.cuh
```diff
@@ -297,6 +297,51 @@ struct delta_binary_decoder {
...
+  // thread block. Result is only valid on thread 0.
+  // This is intended for use only by the DELTA_LENGTH_BYTE_ARRAY decoder.
+  inline __device__ size_t skip_values_and_sum(int skip)
```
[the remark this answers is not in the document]
etseidl: I'll give it a whirl.  I think this is leftover from experiments in decoding using all warps that didn't pan out.

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1956,6 +1966,116 @@ __global__ void __launch_bounds__(block_size, 8)
...
+  }
+
+  // TODO: test for overflow (string_len > 2GB)
+  // string_len is only valid on thread 0
+  auto const string_len = block_reduce(temp_storage.reduce_storage).Sum(len);
+  if (t == 0) { string_data_len = string_len; }
+  __syncthreads();
+
+  // finish off the delta block and get the pointer to the end of the delta block
+  uint8_t* const output_ptr = reinterpret_cast<uint8_t*>(packer.flush());
```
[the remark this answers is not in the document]
etseidl: Derp. Failed attempt to cast away const :sob:. Didn't change back after changing the` flush` return type.

vuule (commented):
full pass!

cpp/src/io/parquet/page_enc.cu
```diff
@@ -1956,6 +1966,116 @@ __global__ void __launch_bounds__(block_size, 8)
...
+  if (t == 0) {
+    first_string = nullptr;
+    packer.init(s->cur, s->page.num_valid, reinterpret_cast<int32_t*>(delta_shared), &temp_storage);
+
+    // if there are valid values, find a pointer to the first valid string
+    if (s->page.num_valid != 0) {
+      for (uint32_t idx = 0; idx < s->page.num_leaf_values; idx++) {
+        size_type const idx_in_col = s->page_start_val + idx;
+        if (s->col.leaf_column->is_valid(idx_in_col)) {
+          // TODO: do we need to account for list(uint8) vs string?
```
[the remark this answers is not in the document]
etseidl: this TODO refers to what's going on around L1591, where if it's not a string, we use `byte_array_view`. Didn't know if just using the `element<string_view>` would lead to incorrect behavior if we're dealing with a byte array that doesn't have a string logical type.

cpp/src/io/parquet/page_delta_decode.cu
```diff
@@ -625,4 +779,29 @@ void __host__ DecodeDeltaByteArray(cudf::detail::hostdevice_vector<PageInfo>& pa
   }
 }
 
+/**
+ * @copydoc cudf::io::parquet::gpu::DecodeDeltaByteArray
+ */
+void DecodeDeltaLengthByteArray(cudf::detail::hostdevice_vector<PageInfo>& pages,
+                                cudf::detail::hostdevice_vector<ColumnChunkDesc> const& chunks,
```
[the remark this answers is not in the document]
etseidl: ok, we can change all these in between `CompactProtocolReader` refactors :smile:

cpp/src/io/parquet/page_enc.cu
```diff
@@ -696,8 +697,14 @@ __global__ void __launch_bounds__(128)
...
           if (is_use_delta) {
-            page_size =
-              max(page_size, delta_data_len(physical_type, type_id, page_g.num_leaf_values));
+            auto delta_len = delta_data_len(physical_type, type_id, page_g.num_leaf_values);
+            // for byte array, delta_len will be just the length data. page_size is the plain
+            // encoded size (i.e. num_values * sizeof(size_type) + char_data_len), so add that
+            // to delta_len but subtract the first term.
+            if (physical_type == BYTE_ARRAY) {
+              delta_len += page_size - page_g.num_leaf_values * sizeof(size_type);
+            }
```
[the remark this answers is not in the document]
etseidl: done

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -697,6 +699,100 @@ __global__ void __launch_bounds__(delta_preproc_block_size) gpuComputeDeltaPageS
...
+  bool const is_bounds_pg = is_bounds_page(s, min_row, num_rows, has_repetition);
+
+  // for DELTA_LENGTH_BYTE_ARRAY, string size is page_data_size - size_of_delta_binary_block.
+  // so all we need to do is skip the encoded string size info and then do pointer arithmetic,
+  // if this isn't a bounds page.
+  if (not is_bounds_pg) {
+    if (t == 0) {
+      auto const* string_start = string_lengths.find_end_of_block(s->data_start, s->data_end);
+      size_t len               = static_cast<size_t>(s->data_end - string_start);
+      // TODO check for overflow
```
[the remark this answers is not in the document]
etseidl: Might be easier to do this on on the host side since we bring the sizes back. Can do a quick check before allocating the char data for the column buffer.

cpp/src/io/parquet/page_enc.cu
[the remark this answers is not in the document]
etseidl: went ahead and changed it, and modified the strings as binary test to write delta too

cpp/src/io/parquet/parquet_gpu.hpp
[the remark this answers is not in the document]
vuule: This was a mildly serious question since as the enums are ATM we'll eventually end up with different order in `decode_kernel_mask` and `encode_kernel_mask`.
But I realized that these are internal and we can modify the assigned values at any point.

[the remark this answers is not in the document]
etseidl: I'm not sure I follow...DELTA_BYTE_ARRAY will be bit 4. The only other possible encoding right now would be BYTE_STREAM_SPLIT which is for floats (but there's a proposal to extend it to FIXED_LEN_BYTE_ARRAY), and that would probably be implemented in the PLAIN encoder anyway. If the mask is 32 bits, that leaves plenty of room for other flags, like NO_NULLS, NO_REPETITION, etc which would be more in line with @nvdbaranec's original vision for the smaller special purpose kernels.

vuule (approved):
:fire: :fire:

isVoid (approved):
One small nitpick.

ttnghia (approved):

PR #14593, opened 2023-12-07
Support `freq` in DatetimeIndex

shwina:
When a `DatetimeIndex` has a fixed frequency offset, pandas defaults to it having a `.freq` attribute. Because we don't support that, we raise in pandas compatible mode.

Thus, working with datetimes is practically impossible in pandas compatible mode because so many datetime operations involve setting a datetime column as an index (resample, groupby).

This PR adds rudimentary support for the `freq` attribute.


python/cudf/cudf/core/index.py
```diff
@@ -2159,6 +2160,19 @@ def __init__(
 
         super().__init__(data, **kwargs)
 
+        if self._freq is not None:
+            unique_vals = self[1:] - self[:-1]
+            if len(unique_vals) != 1 or unique_vals[0] != self._freq:
```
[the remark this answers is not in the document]
shwina: Ah yeah, sorry, I should have done something like this:

```python
deltas = self[1:] - self[:-1]

# all deltas should be the same and equal to `self._freq`:
if len(deltas.unique()) != 1  or deltas[0] != self._freq:
    ... 
```

python/cudf/cudf/core/index.py
```diff
@@ -2142,6 +2141,8 @@ def __init__(
         if yearfirst is not False:
             raise NotImplementedError("yearfirst == True is not yet supported")
 
+        self._freq = _validate_freq(freq)
```
galipremsagar, line 2144: While looking on adding `freq` support before, I found some APIs manipulate `freq`(to new values) and return new results. (_I vaguely remember..but I think that happens in binops?_) Should we add a TODO comment here that this is not fully functional yet and `freq` support needs to be added in rest of the code-base?

shwina: Yes, although maybe the default behaviour could be for `DatetimeIndex` to infer `freq` from its values. Then this should just work.

Also, we should probably only do that in compatibility mode for perf reasons.

python/cudf/cudf/tests/test_datetime.py
```diff
@@ -1586,7 +1624,29 @@ def test_date_range_start_end_freq(request, start, end, freq):
...
+                    start
+                    in {
+                        "2000-02-13 08:41:06",
+                        "1996-11-21 04:05:30",
+                        "1970-01-01 00:00:00",
+                        "1831-05-08 15:23:21",
+                    }
+                )
+            ),
+            reason="Nanosecond offsets being dropped by pandas, which is "
```
bdice, line 1646: Is this better solved by fixing the condition on the parameter, which should be "pandas < 2.0"?

https://github.com/shwina/cudf/blob/ed3ba3ff17cf686d1e6e38f01073d27b1be64799/python/cudf/cudf/tests/test_datetime.py#L1512

python/cudf/cudf/core/index.py
```diff
@@ -1433,9 +1433,10 @@ def __repr__(self):
         if self.name is not None:
             lines[-1] = lines[-1] + ", name='%s'" % self.name
```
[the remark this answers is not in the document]
galipremsagar: We cannot use `+=` for native strings.

python/cudf/cudf/core/tools/datetimes.py
```diff
@@ -463,13 +463,19 @@ class DateOffset:
     }
 
     _CODES_TO_UNITS = {
+        "N": "nanoseconds",
```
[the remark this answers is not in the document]
galipremsagar: There were a bunch of failing tests without these changes, adding these units passed the cudf pytests. 

There is only slight increase in pandas-pytest failures:

```
# This PR:
= 12094 failed, 174794 passed, 3850 skipped, 3314 xfailed, 8 xpassed, 21406 warnings, 102 errors in 1516.39s (0:25:16) =

# `branch-24.02`:
= 11607 failed, 175286 passed, 3849 skipped, 3312 xfailed, 11 xpassed, 21414 warnings, 97 errors in 1493.35s (0:24:53) =
```

galipremsagar (approved):

python/cudf/cudf/core/index.py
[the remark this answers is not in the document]
galipremsagar: > self.diff().unique()

Switched to this for now.

python/cudf/cudf/core/index.py
```diff
@@ -1432,9 +1433,14 @@ def __repr__(self):
...
```
bdice, line 1434:
```suggestion
            lines[-1] += f", name='{self.name}'"
```

python/cudf/cudf/core/index.py
```diff
@@ -1432,9 +1433,14 @@ def __repr__(self):
...
         if "length" in tmp_meta:
-            lines[-1] = lines[-1] + ", length=%d)" % len(self)
-        else:
-            lines[-1] = lines[-1] + ")"
+            lines[-1] = lines[-1] + ", length=%d" % len(self)
```
bdice, line 1436:
```suggestion
            lines[-1] += f", length={len(self)}"
```

python/cudf/cudf/core/index.py
```diff
@@ -1432,9 +1433,14 @@ def __repr__(self):
...
+        if (
+            "freq" in tmp_meta
+            and isinstance(self, DatetimeIndex)
+            and self._freq is not None
+        ):
+            lines[-1] = lines[-1] + f", freq={self._freq}"
```
bdice, line 1442:
```suggestion
            lines[-1] += f", freq={self._freq}"
```

python/cudf/cudf/core/index.py
```diff
@@ -1432,9 +1433,14 @@ def __repr__(self):
...
+        lines[-1] = lines[-1] + ")"
```
bdice, line 1443:
```suggestion
        lines[-1] += ")"
```

python/cudf/cudf/core/tools/datetimes.py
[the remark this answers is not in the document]
bdice: Sounds good, thanks for checking.

python/cudf/cudf/tests/test_datetime.py
bdice: Okay. We can clean it up later.

bdice (approved):
Approving with a few final comments.

python/cudf/cudf/core/index.py
```diff
@@ -2158,6 +2164,27 @@ def __init__(
 
         super().__init__(data, **kwargs)
 
+        if self._freq is not None:
+            unique_vals = self.diff().unique()
+            if len(unique_vals) != 1 or unique_vals[0] != self._freq:
```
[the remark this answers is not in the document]
wence-: > Can an Index be empty but have a freq?

Yes.

python/cudf/cudf/core/index.py
```diff
@@ -2158,6 +2164,27 @@ def __init__(
...
+                raise ValueError()
```
wence-, line 2170: Please add a useful error message here.

python/cudf/cudf/core/index.py
```diff
@@ -2158,6 +2164,27 @@ def __init__(
...
+
+    @_cudf_nvtx_annotate
+    def _copy_type_metadata(
+        self: DatetimeIndex, other: DatetimeIndex, *, override_dtypes=None
+    ) -> GenericIndex:
+        super()._copy_type_metadata(other, override_dtypes=override_dtypes)
+        self._freq = other._freq
```
wence-, line 2177: Unlike `_from_data` and `__init__`, it appears this one doesn't validate that the provided frequency is valid?

python/cudf/cudf/core/index.py
```diff
@@ -2158,6 +2164,27 @@ def __init__(
...
+        return self
+
+    @classmethod
+    def _from_data(
+        cls, data: MutableMapping, name: Any = no_default, freq: Any = None
+    ):
+        result = super()._from_data(data, name)
+        result._freq = _validate_freq(freq)
```
wence-, line 2185: This validates that `freq` is viable as a `DateOffset`, but not that `freq` is correct for `data`?

python/cudf/cudf/core/index.py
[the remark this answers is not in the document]
galipremsagar: Yes, added more validation here.

python/cudf/cudf/core/index.py
```diff
@@ -2158,6 +2164,27 @@ def __init__(
 
         super().__init__(data, **kwargs)
 
+        if self._freq is not None:
+            unique_vals = self.diff().unique()
```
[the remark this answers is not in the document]
galipremsagar: Yes, changes it

python/cudf/cudf/core/index.py
galipremsagar: Done

python/cudf/cudf/core/index.py
galipremsagar: I added validation in `from_data` too. I haven't made this a method of `DatetimeIndex` because we aren't utilizing any attributes of an existing object(`self`), but don't feel strongly against adding it as a class method. Let me know if either of you feel we need it as class method.

bdice (approved):
Fix the repr -- then this is good from my side.

PR #14595, opened 2023-12-07
Fix synchronization issue when writing string columns with dictionary to ORC

vuule:
## Description
Changes in https://github.com/rapidsai/cudf/pull/14295 introduced a synchronization issue in `build_dictionaries`. After stripe_dicts are initialized on the host, we copy them to the device and then launch kernels that read the dicts (device copy). However, after these kernels we deallocate buffers that are not longer needed and clear the dicts' views to these buffers on the host. The problem is that, without synchronization after the H2D copy, the host modification can be done before the H2D copy is performed, and we run the kernels with the altered state.
This PR adds a sync point to make sure the copy is done before host-side modification.

## Checklist

### DIFF
--- a/cpp/include/cudf/io/orc.hpp
+++ b/cpp/include/cudf/io/orc.hpp
@@ -451,7 +451,7 @@ class orc_writer_options {
   // Optional compression statistics
   std::shared_ptr<writer_compression_statistics> _compression_stats;
   // Specify whether string dictionaries should be alphabetically sorted
-  bool _enable_dictionary_sort = true;
+  bool _enable_dictionary_sort = false;
 
   friend orc_writer_options_builder;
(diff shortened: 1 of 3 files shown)


GregoryKimball:
@abellina Would you please confirm that disabling the ORC dict sorting addresses the query error? I believe this confirmation is prerequisite to merging.

ttnghia (approved):

jlowe:
@abellina and I verified that disabling the dictionary sorting did not resolve the issue.  We're currently digging into other parts of the associated PR that were not specific to the sorted dictionary path.

abellina (approved):

bdice (approved):
This looks good to me.

Issue #14597, opened 2023-12-08
[BUG] Random Parquet CI failures


etseidl:
Beginning with #14000, it was noticed that the `ParquetWriterTest.ManyFragments` unit test would fail at random times. Rerunning the failing test would clear the error. But after that PR was merged, the error began to appear in other PRs as that change was merged in. Running the test through racecheck yielded several reported problems.
```
========= Error: Race reported between Write access at 0x1d10 in /home/seidl2/src/cudf_branches/cudf_strings/cpp/build/_deps/thrust-src/dependencies/cub/cub/block/specializations/block_reduce_warp_reductions.cuh:146:unsigned int cub::CUB_101702_860_NS::BlockReduceWarpReductions<unsigned int, (int)512, (int)1, (int)1, (int)860>::ApplyWarpAggregates<(bool)1, cub::CUB_101702_860_NS::Sum>(T2, unsigned int, int)
=========     and Read access at 0x1b90 in /home/seidl2/src/cudf_branches/cudf_strings/cpp/build/_deps/thrust-src/dependencies/cub/cub/block/specializations/block_reduce_warp_reductions.cuh:117:unsigned int cub::CUB_101702_860_NS::BlockReduceWarpReductions<unsigned int, (int)512, (int)1, (int)1, (int)860>::ApplyWarpAggregates<(bool)1, cub::CUB_101702_860_NS::Sum, (int)1>(T2, unsigned int, int, cub::CUB_101702_860_NS::Int2Type<T3>) [240 hazards]
=========     and Read access at 0x1ba0 in /home/seidl2/src/cudf_branches/cudf_strings/cpp/build/_deps/thrust-src/dependencies/cub/cub/block/specializations/block_reduce_warp_reductions.cuh:117:unsigned int cub::CUB_101702_860_NS::BlockReduceWarpReductions<unsigned int, (int)512, (int)1, (int)1, (int)860>::ApplyWarpAggregates<(bool)1, cub::CUB_101702_860_NS::Sum, (int)2>(T2, unsigned int, int, cub::CUB_101702_860_NS::Int2Type<T3>) [320 hazards]
[...]

PR #14598, opened 2023-12-08
Fix race detected in Parquet writer

etseidl:
## Description
While investigating #14597 it was found that there was a race introduced by #14000. Adding a sync between invocations of a block reduce resolves the error. 

## Checklist

### DIFF
--- a/cpp/src/io/parquet/page_enc.cu
+++ b/cpp/src/io/parquet/page_enc.cu
@@ -205,7 +205,7 @@ void __device__ calculate_frag_size(frag_init_state_s* const s, int t)
       }
     }
   }
-  __syncthreads();
+
   auto const total_len = block_reduce(reduce_storage).Sum(len);
   __syncthreads();
   auto const total_valid = block_reduce(reduce_storage).Sum(num_valid);
--- a/cpp/src/io/parquet/page_enc.cu
+++ b/cpp/src/io/parquet/page_enc.cu
@@ -206,7 +206,8 @@ void __device__ calculate_frag_size(frag_init_state_s* const s, int t)
     }
   }
   __syncthreads();
-  auto const total_len   = block_reduce(reduce_storage).Sum(len);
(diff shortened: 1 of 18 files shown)


PointKernel (approved):
LGTM

cpp/src/io/parquet/page_enc.cu
```diff
@@ -206,7 +206,8 @@ void __device__ calculate_frag_size(frag_init_state_s* const s, int t)
     }
   }
   __syncthreads();
-  auto const total_len   = block_reduce(reduce_storage).Sum(len);
+  auto const total_len = block_reduce(reduce_storage).Sum(len);
+  __syncthreads();
```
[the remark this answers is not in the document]
etseidl: I was worried too, but dug into the Sum code and saw that there were syncthreads inside and convinced myself it was ok :(

ttnghia, line 210: So we really need two sync like this?

vuule (approved):
Thank you for looking into this issue promptly!
Hopefully that's the last of the `ManyFragments` failures.

cpp/src/io/parquet/page_enc.cu
vuule: hm, maybe we should remove it if racecheck does not complain

etseidl: I'll try removing it and see

vuule: > I'll try removing it and see
trying this on my end as well

> What is racecheck?
racecheck tool in compute-sanitizer

etseidl: remove?

vuule: no racecheck errors, and, looking at the code, no shared data is modified before the reductions. IMHO we should be good to remove the first sync.

PR #14599, opened 2023-12-08
Tighten up type promotion in join key matching

### LINKED ISSUE #14594: [FEA] Tighten up promotion when merging with non-equal key column dtypes

To date, cudf has attempted to match pandas semantics when matching join keys in a merge. libcudf does not perform merges between mismatching table dtypes. Consequently, the first step of a merge in cudf is to determine a "common" dtype for each pair of columns used as keys in the merge.

The pandas rules are mostly (though not completely since there is some under the table work that happens in the join algorithm) encoded in https://github.com/pandas-dev/pandas/blob/f7c73a5f1aaf0724598e60c0cc5732604ec842a8/pandas/core/reshape/merge.py#L1340

There are a few problems when trying to match these in cudf:

- not all column types in pandas can be represented in cudf (we do not have an `object` column for example)
- it is difficult to unambiguously determine the type promotion rules since they are not written down anywhere
    - for example, promotion rules for categorical columns differ depending on whether the categorical is the left or right key.

Moreover, there are other, correctness, problems. The current type promotion rules admit lossy conversions that can result in false positive matches in merges.

Example:
```
left = cudf.DataFrame({"key": [1, 2, 2**53]})
right = cudf.DataFrame({"key": [2**53 + 1, 10]})
right["key"] = right.key.astype("uint64")
left.merge(right, on="key", how="inner")
#            key
# 0  9.007199e+15
left
#                key
# 0                 1
# 1                 2
# 2  9007199254740992
right
#                key
# 0  9007199254740993
# 1                10
```

Pandas is also susceptible to this, but produces a different wrong result.

I would like to tighten up the rules in cudf, so that it is impossible for the user to get a "surprising" result without some explicit intervention on their behalf. We would also try and match pandas more closely where that is possible, but my preference is to be correct in a subset of cases over dubiously correct in a larger set.

There are, I think, three levels of things we could do:

1. Push the burden of dtype matching completely on to the user: complain (raise) if merge keys do not match dtypes _exactly_
2. Promote keys whose dtypes allow so safely (without needing to inspect values), and raise for cases where that is not possible. The user can still perform the merge by intentionally casting to matching types. But then they must know that it is safe.
3. Try and match pandas promotions as closely as possible and accept that there might be false positives.

I would like to go for (2). (1) is easiest; (3) is difficult, probably a moving target and can result in false positives without the user explicitly "requesting" them.

With cudf-pandas (2), I think, skates the line between ease of use and correctness reasonably well. We can run as much on the GPU as possible and raise (possibly providing a warning in pandas-compat mode) with fallback to CPU. When using cudf directly, users will hopefully be willing to accept a few more edge cases in the name of consistency. 

Concretely this would mean:

- No casting for strings
- No casting for lists
- No casting for structs
- Categoricals:
    - if both columns are categorical and match, no casting
    - if both columns are categorical and _do not_ match, raise[^1]
    - if one column is categorical, unwrap, and go round again[^2]
 - No casting for decimals
 - No casting for datetimes[^3]
 - For numeric types, use a type promotion lattice that has lossless least upper bounds for all types[^4]

For numeric types, that means that we would only promote pairs of types where there exists a wider type whose values are uniquely and unambiguously mapped onto from the narrower types.

For example `(int32, uint32) -> int64` would be allowed, but merging a pair `(int32, uint64)` would raise (since there is no signed 128bit int that we could use). Similarly, we would safely be able to promote `(intX, floatY)` pairs (and similarly with `uintX`) as long as the integer type is 32 or fewer bits wide[^5].

[^1]: I could also be convinced to unwrap and go round again, but that would lose information about the categorical nature of the inputs
[^2]: Pandas behaviour in this case depends on whether the left or right key is categorical (and which merge type it is): it casts the non-categorical to object, and the categorical to its underlying dtype, then imperfectly goes through its matching process again
[^3]: I haven't looked at what pandas does here, but I guess the other thing one could do is promote when one can losslessly convert
[^4]: See, for example https://jax.readthedocs.io/en/latest/jep/9407-type-promotion.html though I disagree with their approach of selecting a "weak" float64 as the least upper bound for `(int64, uint64)`
[^5]: Merging between float and int columns is kind of weird, so I could also be convinced to raise when merging between mismatching numeric kinds.

wence-:
## Description
This is much tighter than the current scheme, but makes sure that we never accidentally promote in situations where we might produce false positives in the subsequent join.

The assumption is that most joins occur on columns with matching dtypes. Where it is definitely safe to promote numeric types, we do so, but never between kinds. For almost all other column types we require exact dtype matching and now raise. The only exception is if exactly one of the two columns is categorical in which case we  decategorize and go round again.

- Closes #14594 

## Checklist

### DIFF
--- a/python/cudf/cudf/core/dtype_matching.py
+++ b/python/cudf/cudf/core/dtype_matching.py
@@ -1,0 +1,259 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+from __future__ import annotations
+
+import functools
+import itertools
+from typing import Dict, List, Set, Tuple, TypeVar, cast
+
+import numpy as np
+from pandas.api.types import is_numeric_dtype
+
+import cudf.core.column as column
+from cudf.core.dtypes import is_categorical_dtype
+
+T = TypeVar("T")
+
+
+def toposort(graph: Dict[T, Set[T]]) -> List[T]:
+    """
+    Return nodes in a DAG in a (non-unique) topological order.
+
+    The order may not be unique.
+
+    Raises RuntimeError if the provided graph is not a DAG.
+    """
+    # Kahn's algorithm, https://dl.acm.org/doi/10.1145/368996.369025
+    # Copy because we'll modify the graph in place.
+    graph = {k: set(v) for k, v in graph.items()}
+    invgraph: Dict[T, Set[T]] = {k: set() for k in graph}
+    for k, v in graph.items():
+        for v_ in v:
+            invgraph[v_].add(k)
+    order = []
+    predecessors = set(k for k, v in invgraph.items() if not v)
+    while predecessors:
+        order.append(n := predecessors.pop())
+        edges = graph[n]
+        while edges:
+            invgraph[m := edges.pop()].remove(n)
+            if not invgraph[m]:
+                predecessors.add(m)
+    if any(v for v in graph.values()):
+        raise RuntimeError("Provided lattice structure is not a DAG")
+    return order
+
+
+def ancestors(graph: Dict[T, Set[T]]) -> Dict[T, Set[T]]:
+    """
+    Return the graph of the ancestors of a DAG
+
+    Arrows in the graph should point from child to parent.
+
+    Raises RuntimeError if the provided graph is not a DAG.
+    """
+    graph = {k: set(v) for k, v in graph.items()}
+    order = toposort(graph)
+    for k in reversed(order):
+        graph[k] |= set(
+            itertools.chain.from_iterable(graph[v_] for v_ in graph[k])
+        )
+    return graph
+
+
+def least_common_ancestors(
+    a: T, b: T, *, ancestors: Dict[T, Set[T]]
+) -> Set[T]:
+    r"""
+    Return the least common ancestors of a and b
+
+    Arrows in the graph should point from child to parent
+
+    For example, given the graph
+
+    C  D
+    ^  ^
+    |\/|
+    |/\|
+    A  B
+
+    LCA(A, B) = (C, D)
+    LCA(A, C) = (C)
+    LCA(A, D) = (D)
+    LCA(B, C) = (C)
+    LCA(B, D) = (D)
+    LCA(C, D) = ()
+    """
+    # This is not very sophisticated but the graphs are tiny.
+    # It's basically the algorithm described in Section 4.3 of Bender,
+    # Farach-Colton, Pemmasani, Skiena, and Sumazin, J. Algorithms
+    # (2005) https://doi.org/10.1016/j.jalgor.2005.08.001
+
+    # Candidate subgraph is intersection of the closure of the nodes
+    # closure is ancestors + self.
+    subnodes = (ancestors[a] | {a}) & (ancestors[b] | {b})
+
+    # least common ancestors are all nodes in the subgraph that are
+    # roots (i.e. have no edges pointing in to them)
+    return subnodes - set(
+        itertools.chain.from_iterable(ancestors[node] for node in subnodes)
+    )
+
+
+def promotion_table(graph: Dict[T, Set[T]]) -> Dict[Tuple[T, T], T]:
+    """
+    Return a type promotion table for all pairs of types in the
+    provided partial semi-lattice.
+
+    Raises RuntimeError if the input is not a DAG
+
+    Raises ValueError if any pair does not have a unique least upper
+    bound.
+
+    The promotion for each pair in the input is the unique least upper
+    bound of the pair in the input. The absence of a pair in the
+    return value means that no unique upper bound exists, so there is
+    no safe promotion route for that pair.
+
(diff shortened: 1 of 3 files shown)


python/cudf/cudf/core/dtype_matching.py
```diff
@@ -0,0 +1,259 @@
...
+    # Direct promotion routes for float types
+    # +-----+    +-----+    +-----+
+    # | f16 |--->| f32 |--->| f64 |
+    # +-----+    +-----+    +-----+
+    float_lattice: Dict[np.dtype, Set[np.dtype]] = {
+        np.dtype("float16"): {np.dtype("float32")},
+        np.dtype("float32"): {np.dtype("float64")},
+        np.dtype("float64"): set(),
+    }
+    return {**promotion_table(int_lattice), **promotion_table(float_lattice)}
```
wence-, line 173: The results here are small enough that we could just write out all the pairs, but this code is generic to handle any partial semi-lattice of promotion rules.

python/cudf/cudf/core/join/join.py
```diff
@@ -255,20 +255,9 @@ def perform_merge(self) -> cudf.DataFrame:
...
             left_join_cols.append(lcol_casted)
             right_join_cols.append(rcol_casted)
-
-            # Categorical dtypes must be cast back from the underlying codes
-            # type that was returned by _match_join_keys.
-            if (
-                self.how == "inner"
-                and isinstance(lcol.dtype, cudf.CategoricalDtype)
-                and isinstance(rcol.dtype, cudf.CategoricalDtype)
-            ):
```
wence-, line 268: Pandas never did this, AFAICT. For an inner merge if the _left_ column was categorical, it would decat the left column and convert the right column to object (if not categorical), and vice versa. The resulting key column in the output would be from the left column. So if you were merging on (cat, int) you would get an `int` key column, and on (int, cat) you get an `object` key column:

```
n [2]: right = pd.DataFrame({"key": pd.Series([1, 2], dtype="category"), "value": [2, 1]})

In [3]: left = pd.DataFrame({"key": [1, 3], "value": [1, 2]})

In [4]: left.merge(right, on="key", how="inner").dtypes
Out[4]: 
key        object
value_x     int64
value_y     int64
dtype: object

In [5]: left.merge(right, on="key", how="inner")
Out[5]: 
  key  value_x  value_y
0   1        1        2

In [6]: right.merge(left, on="key", how="inner").dtypes
Out[6]: 
key        int64
value_x    int64
value_y    int64
dtype: object

In [7]: left.merge(right, on="key", how="inner")
Out[7]: 
  key  value_x  value_y
0   1        1        2

# If both are cat, then we decat and merge.
In [8]: left = pd.DataFrame({"key": pd.Series([1, 3], dtype="category"), "value": [1, 2]})

In [9]: left.merge(right, on="key", how="inner")
Out[9]: 
   key  value_x  value_y
0    1        1        2

In [10]: left.merge(right, on="key", how="inner").dtypes
Out[10]: 
key        int64
value_x    int64
value_y    int64
dtype: object

```

python/cudf/cudf/core/dtype_matching.py
```diff
@@ -0,0 +1,159 @@
...
+from typing import Tuple, cast
+
+import numpy as np
+
+import cudf.core.column as column
+from cudf.core.dtypes import is_categorical_dtype
+
+
+@functools.cache
+def common_numeric_type(ltype: np.dtype, rtype: np.dtype) -> np.dtype | None:
```
wence-, line 14: TODO: Write a test for this that asserts that the full promotion table generated by this scheme is a partial semi-lattice. That way we assure ourselves that promotion is associative.

python/cudf/cudf/core/dtype_matching.py
```diff
@@ -0,0 +1,259 @@
...
+    ltype = left.dtype
+    rtype = right.dtype
+
+    if ltype == rtype:
+        return left, right
+    left_is_cat = is_categorical_dtype(left.dtype)
+    right_is_cat = is_categorical_dtype(right.dtype)
+
+    if left_is_cat and right_is_cat:
+        raise ValueError("Cannot merge on non-matching categorical types")
```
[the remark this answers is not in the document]
wence-: I now do this.

vyasr:
@wence- just double-checking, this is intentionally in draft, right? Not expecting a review?

wence-:
Yes, I need to finish working through adapting the tests

vyasr:
Closing this for now. The `_match_join_keys` logic has evolved significantly since this PR was opened (categorical handling, boolean-dtype fixes, decimal types, string types, pandas 3 compat), and the area is still actively changing. The goals from #14594 remain valid but are better addressed once the join matching scheme stabilizes. The key insight (tighten numeric cross-kind promotion without disturbing `how`-dependent categorical behavior) is preserved in the issue for future work.

PR #14601, opened 2023-12-08
Fix DataFrame.reindex removing column name

mroeschke:
## Description
Fixes the following behavior

```python
In [1]: import cudf
    
In [2]: import pandas as pd

In [3]: df = cudf.DataFrame([1], columns=cudf.Index([1], name="foo"))

In [4]: df
Out[4]: 
foo  1
0    1

In [5]: df = df.reindex(index=[0, 1])

In [6]: df
Out[6]: 
      1  <-- foo missing
0     1
1  <NA>
```

## Checklist

### DIFF
--- a/python/cudf/cudf/core/indexed_frame.py
+++ b/python/cudf/cudf/core/indexed_frame.py
@@ -2661,13 +2661,17 @@ class IndexedFrame(Frame):
             )
             for name in names
         }
+        if column_names is None:
+            level_names = self._data.level_names
+        elif isinstance(column_names, pd.Index):
+            level_names = tuple(column_names.names)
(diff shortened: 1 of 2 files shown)


bdice (approved):
The same pattern appears here. Is it also problematic? https://github.com/rapidsai/cudf/blob/8b23436e0fc48706d21629003dcf144fb02045fb/python/cudf/cudf/core/dataframe.py#L746-L748

mroeschke:
> The same pattern appears here. Is it also problematic? 

I think this one is OK. Since this is in `__init__`, for this `data=None` branch there is no data column name to preserve

PR #14602, opened 2023-12-08
Address potential race conditions in Parquet reader

etseidl:
## Description
Related to #14597. Fixes reported errors by racecheck.

## Checklist


vuule (approved):

vuule:
There is another racecheck error in decode, might be somewhat related. Decided with @etseidl to investigate remaining errors before merging this fix.

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -691,6 +696,10 @@ __device__ void gpuUpdateValidityOffsetsAndRowIndices(int32_t target_input_value
 
   PageNestingDecodeInfo* nesting_info_base = s->nesting_info;
 
+  // need this to ensure input_value_count is read by all threads before s->input_value_count
+  // is modified below (just in case input_value count >= target_input_value_count).
+  __syncwarp();
```
etseidl, line 701: This one I'm not so sure about needing. In the worst case, thead 0 sets the local var, skips the loop (and the syncwarp within it) and then overwrites the shared value before other threads read it. But in that case it will just overwrite with the same value.

vuule: do we actually need to update s->nz_count, s->input_value_count and s->input_row_count if we never enter the loop?

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -848,7 +853,7 @@ __global__ void __launch_bounds__(decode_block_size)
       target_pos = min(s->nz_count, src_pos + decode_block_size - out_thread0);
       if (out_thread0 > 32) { target_pos = min(target_pos, s->dict_pos); }
     }
-    // TODO(ets): see if this sync can be removed
+    // this needs to be here to prevent warp 1/2 modifying src_pos before all threads have read it
```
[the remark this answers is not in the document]
etseidl: yes.

cpp/src/io/parquet/page_decode.cuh
etseidl: I'm thinking no...they shouldn't have changed if the loop wasn't entered. But I'll admit this is one of the parts of the parquet code that I understand the least.

vuule: If that's the case, we should be able to return early if initially `input_value_count >= target_input_value_count`, right?
That would simplify the logic and prevent the tool from reporting the race condition.
CC @nvdbaranec

etseidl: I made the change and verified that racecheck is happy

ttnghia (approved):

cpp/src/io/parquet/delta_binary.cuh
```diff
@@ -219,6 +219,8 @@ struct delta_binary_decoder {
 
     // need to account for the first value from header on first pass
     if (current_value_idx == 0) {
+      // make sure all threads access current_value_idx above before incrementing
+      __syncwarp();
```
nvdbaranec, line 223: This sync seems unnecessary.  If t0 is the only thread that ever modifies `current_value_idx` there's never a case where it can be out of sync with itself.  And having all the other threads wait on the existing __syncwarp() takes care of this particular modification.

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -243,6 +243,9 @@ __device__ cuda::std::pair<int, int> gpuDecodeDictionaryIndices(page_state_s* s,
   int pos            = s->dict_pos;
   int str_len        = 0;
 
+  // ensure all threads read s->dict_pos before returning
+  __syncwarp();
```
nvdbaranec, line 247: Not sure about this one.  The return value from this function is explicitly stated to only be valid on thread 0. Looking at all the call sites, it's always thread 0 that actually does any work with the value.

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -357,6 +360,9 @@ inline __device__ int gpuDecodeRleBooleans(page_state_s* s, state_buf* sb, int t
   uint8_t const* end = s->data_end;
   int64_t pos        = s->dict_pos;
 
+  // ensure all threads read s->dict_pos before returning
```
nvdbaranec, line 363: Same comment as the one in `gpuDecodeDictionaryIndices`

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -78,8 +78,10 @@ __device__ thrust::pair<int, int> page_bounds(page_state_s* const s,
 
   // can skip all this if we know there are no nulls
   if (max_def == 0 && !is_bounds_pg) {
-    s->page.num_valids = s->num_input_values;
-    s->page.num_nulls  = 0;
+    if (t == 0) {
```
nvdbaranec, line 81: Good find. This was technically UB the old way.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -294,7 +296,6 @@ __device__ thrust::pair<int, int> page_bounds(page_state_s* const s,
       pp->num_nulls  = null_count;
       pp->num_valids = pp->num_input_values - null_count;
     }
-    __syncthreads();
```
nvdbaranec, line 297: This seems dangerous to remove.  Aren't all threads except 0 in danger of using the wrong `pp->num_nulls` value?

cpp/src/io/parquet/delta_binary.cuh
etseidl: but can thread 0 modify `current_value_idx` before thread 31 does the test for `current_value_idx == 0`? If that happens, and `value_count = 1`, then thread 0 could return and thread 31 continue, no?

cpp/src/io/parquet/page_decode.cuh
etseidl: Yeah, this is kind of like the one in `gpuUpdateValidityOffsetsAndRowIndices`, except here the assignment back to s->dict_pos is done after this call returns.  If the loop is entered, then all threads will hit the syncwarp there. It's only an issue if `pos >= target_pos`. Given this has worked without problems for quite some time, I can get rid of this and the one in `gpuDecodeRleBooleans`.

cpp/src/io/parquet/page_string_decode.cu
etseidl: This is another only-valid-on-thread-0 result. I originally added _more_ syncthreads before all the other returns, but @vuule pointed out that once this function returns, all that happens is thread 0 takes the return values and copies them to global memory (along with 2 shared mem fields) and then returns. The other threads simply return ignored garbage and exit.

cpp/src/io/parquet/delta_binary.cuh
vuule: All threads read `current_value_idx` in line 218 and 221. For example, if `current_value_idx == 0` and `value_count == 1`, thread 0 can bump `current_value_idx` to 1 before other threads read `current_value_idx` in line 218 or 221. This would cause some threads to take the incorrect path.

nvdbaranec: I don't think that can happen.  Threads in a warp don't run independently of each other.  Everyone will evaluate `if(current_value_idx == 0)` at the same time (except exited threads). But once we have the branch `if (lane_id == 0) { current_value_idx++; }` that write to shared could be incomplete by the time everyone does the final if.

vuule: > Everyone will evaluate if(current_value_idx == 0) at the same time

I don't think that's guaranteed on recent archs.

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -243,6 +243,8 @@ __device__ cuda::std::pair<int, int> gpuDecodeDictionaryIndices(page_state_s* s,
   int pos            = s->dict_pos;
   int str_len        = 0;
 
+  // NOTE: racecheck warns about a RAW involving s->dict_pos, which is likely a false positive
```
vuule, line 246: Something along these lines?
```suggestion
  // NOTE: racecheck warns about a RAW involving s->dict_pos, which is likely a false positive because the only path that does not include a sync will lead to s->dict_pos being overwritten with the same value
```

cpp/src/io/parquet/delta_binary.cuh
nvdbaranec: After some internal discussions I think this is fine.  It certainly doesn't hurt anything.

nvdbaranec (approved):

PR #14605, opened 2023-12-09
Fix DataFrame.reindex when column reindexing to MultiIndex/RangeIndex

mroeschke:
## Description

1. `reindex(columns=cudf.MultiIndex)` will raise since `set(cudf.MultiIndex)` is being called
2. `reindex(columns=range/RangeIndex)` would not necessarily retain a RangeIndex result

Discovered during refactoring `DataFrame.__init__` and further convinces me that the `column` should be passed to `ColumnAccessor` eventually

## Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -2744,11 +2744,14 @@ def reindex(
...
-        df = (
-            self
-            if columns is None
-            else self[list(set(self._column_names) & set(columns))]
-        )
+        if columns is None:
+            df = self
+        else:
+            columns = as_index(columns)
+            intersection = self._data.to_pandas_index().intersection(
```
mroeschke, line 2751: Happy to have suggestion if there's a way to do this without dispatching to pandas. This was to ensure the MultiIndex intersection case worked correctly

galipremsagar (approved):
Seems reasonable to me

PR #14607, opened 2023-12-10
Remove `cuda::proclaim_return_type` from nested lambda

ttnghia:
This removes `cuda::proclaim_return_type` from a device lambda because that lambda is going to be nested inside another device lambda, which is in turn enclosed by `cuda::proclaim_return_type`. 

This PR is to fix a compile issue that we encountered:
```
/usr/local/cuda/include/cuda/std/detail/libcxx/include/__functional/invoke.h(402): error: 
calling a __device__ function("cudf::tdigest::detail::_NV_ANON_NAMESPACE::build_output_column(int,   
  ::std::unique_ptr<   ::cudf::column,     ::std::default_delete<   ::cudf::column> >  &&,     ::std::unique_ptr<   ::cudf::column,     ::std::default_delete<   ::cudf::column> >  &&,     ::std::unique_ptr<   ::cudf::column,     ::std::default_delete<   ::cudf::column> >  &&,     ::std::unique_ptr<   ::cudf::column,     ::std::default_delete<   ::cudf::column> >  &&,     ::std::unique_ptr<   ::cudf::column,     ::std::default_delete<   ::cudf::column> >  &&, bool,  ::rmm::cuda_stream_view,  ::rmm::mr::device_memory_resource *)
::[lambda(int) (instance 2)]::operator ()(int) const") from a __host__ __device__ function("__invoke") is not allowed
```

Note: The issue is reproducible only in our build environment: ARM architecture, cuda 12 + rockylinux8.

Closes https://github.com/rapidsai/cudf/issues/14610.

### DIFF
--- a/cpp/src/quantiles/tdigest/tdigest_aggregation.cu
+++ b/cpp/src/quantiles/tdigest/tdigest_aggregation.cu
@@ -588,14 +588,13 @@ std::unique_ptr<column> build_output_column(size_type num_rows,
     return weights[i] == 0;
   };
   // whether or not this particular tdigest is a stub
-  auto is_stub_digest = cuda::proclaim_return_type<size_type>(
-    [offsets = offsets->view().begin<size_type>(), is_stub_weight] __device__(size_type i) {
-      return is_stub_weight(offsets[i]) ? 1 : 0;
-    });
+  auto is_stub_digest = [offsets = offsets->view().begin<size_type>(), is_stub_weight] __device__(
+                          size_type i) { return is_stub_weight(offsets[i]) ? 1 : 0; };
 
   size_type const num_stubs = [&]() {
     if (!has_nulls) { return 0; }
-    auto iter = cudf::detail::make_counting_transform_iterator(0, is_stub_digest);
+    auto iter = cudf::detail::make_counting_transform_iterator(
+      0, cuda::proclaim_return_type<size_type>(is_stub_digest));
     return thrust::reduce(rmm::exec_policy(stream), iter, iter + num_rows);
   }();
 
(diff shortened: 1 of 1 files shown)


cpp/src/quantiles/tdigest/tdigest_aggregation.cu
```diff
@@ -588,14 +588,13 @@ std::unique_ptr<column> build_output_column(size_type num_rows,
     return weights[i] == 0;
   };
   // whether or not this particular tdigest is a stub
-  auto is_stub_digest = cuda::proclaim_return_type<size_type>(
-    [offsets = offsets->view().begin<size_type>(), is_stub_weight] __device__(size_type i) {
-      return is_stub_weight(offsets[i]) ? 1 : 0;
-    });
+  auto is_stub_digest = [offsets = offsets->view().begin<size_type>(), is_stub_weight] __device__(
```
ttnghia, line 591: This lambda is going to be used in line 643, inside another lambda which is enclosed in `proclaim_return_type`.

miscco: I would love a comment for that so the reader does not have to puzzle the pieces together

miscco (approved):

bdice:
> Perhaps @bdice should verify this does not break his CCCL PR. #14576

I’ll verify this as soon as possible.

karthikeyann (approved):

nvdbaranec (approved):

bdice (approved):
I confirmed that #14576 builds with this change. I noted in #14610 that we should file an issue with CCCL that documents the compiler versions / etc. used to trigger this error, with a minimal reproduction (especially since RAPIDS CI doesn't complain). The `proclaim_return_type` wrapper should not break normal device lambda compilation.

PR #14612, opened 2023-12-11
Create strings-specific make_offsets_child_column for multiple offset types

davidwendt:
## Description
Creates `cudf::strings::detail::make_offsets_child_column` for specific use with strings offsets to enable int64 offsets.
This also introduces the `LIBCUDF_LARGE_STRINGS_THRESHOLD` environment variable which can be used at runtime to produce int64 offsets for an arbitrary number of character bytes. This utility is not used in this PR but will be employed in a future PR.

## Checklist


cpp/include/cudf/strings/detail/strings_children.cuh
```diff
@@ -122,6 +122,62 @@ auto make_strings_children(SizeAndExecuteFunction size_and_exec_fn,
...
+  auto d_offsets    = offsets_view.template data<int32_t>();
+
+  // The number of offsets is count+1 so to build the offsets from the sizes
+  // using exclusive-scan technically requires count+1 input values even though
+  // the final input value is never used.
+  // The input iterator is wrapped here to allow the 'last value' to be safely read.
+  auto map_fn =
+    cuda::proclaim_return_type<size_type>([begin, count] __device__(size_type idx) -> size_type {
+      return idx < count ? static_cast<size_type>(begin[idx]) : size_type{0};
+    });
```
mythrocks, line 161: :+1:

cpp/include/cudf/strings/detail/strings_children.cuh
```diff
@@ -122,6 +122,62 @@ auto make_strings_children(SizeAndExecuteFunction size_and_exec_fn,
...
+ * @return Offsets column and total elements
+ */
+template <typename InputIterator>
+std::pair<std::unique_ptr<column>, int64_t> make_offsets_child_column(
+  InputIterator begin,
+  InputIterator end,
+  rmm::cuda_stream_view stream,
+  rmm::mr::device_memory_resource* mr)
+{
+  auto const count = static_cast<size_type>(std::distance(begin, end));
```
mythrocks, line 148: Nitpick: The `count` here is the number of input string rows.  But it seems to be used as `count + 1` (i.e. as the `num_offsets`).  

Would there be value in defining `num_offsets = std::distance(begin, end) + 1` instead of `count`, and using that in place of `count+1`? 

I ask because `sizes_to_offsets(input_itr, input_iter + count + 1, ... )` is surprising on first glance.

mythrocks (approved):
LGTM. Minor nitpick.

cpp/include/cudf/strings/detail/strings_children.cuh
davidwendt: So I still need to pass `count` on line 159. I don't see a need to define both local variables.

ttnghia, line 148: Do we need to check if `std::distance(begin, end)` is within the 32 bit integer limit?

cpp/include/cudf/strings/detail/strings_children.cuh
```diff
@@ -122,6 +122,62 @@ auto make_strings_children(SizeAndExecuteFunction size_and_exec_fn,
...
+  auto offsets_column =
+    make_numeric_column(data_type{type_id::INT32}, count + 1, mask_state::UNALLOCATED, stream, mr);
+  auto offsets_view = offsets_column->mutable_view();
+  auto d_offsets    = offsets_view.template data<int32_t>();
```
ttnghia, line 152: They can be combined.
```
auto const d_offsets = offsets_column->mutable_view().template data<int32_t>();
```

cpp/include/cudf/strings/detail/strings_children.cuh
ttnghia: IMO `count` here sounds confusing because it doesn't tell if this is a string count of offset count. So maybe `string_count` is better here?

cpp/src/strings/utilities.cu
```diff
@@ -129,6 +129,13 @@ special_case_mapping const* get_special_case_mapping_table()
   });
 }
 
+int64_t get_offset64_threshold()
+{
+  auto const threshold  = std::getenv("LIBCUDF_LARGE_STRINGS_THRESHOLD");
+  std::size_t const rtn = threshold != nullptr ? std::atol(threshold) : 0;
+  return rtn > 0 ? rtn : std::numeric_limits<int32_t>::max();
```
karthikeyann, line 136: It's possible to set `LIBCUDF_LARGE_STRINGS_THRESHOLD` to be greater than `std::numeric_limits<int32_t>::max()`.
Is there a reason behind choosing return type as `int64_t`?
For the purpose of this function, is `int32_t` enough?

cpp/include/cudf/strings/detail/strings_children.cuh
```diff
@@ -122,6 +122,62 @@ auto make_strings_children(SizeAndExecuteFunction size_and_exec_fn,
...
+    cuda::proclaim_return_type<size_type>([begin, count] __device__(size_type idx) -> size_type {
+      return idx < count ? static_cast<size_type>(begin[idx]) : size_type{0};
+    });
+  auto input_itr = cudf::detail::make_counting_transform_iterator(0, map_fn);
+  // Use the sizes-to-offsets iterator to compute the total number of elements
+  auto const total_elements = sizes_to_offsets(input_itr, input_itr + count + 1, d_offsets, stream);
+
+  // TODO: replace exception with if-statement when enabling creating INT64 offsets
+  CUDF_EXPECTS(total_elements <= static_cast<int64_t>(std::numeric_limits<size_type>::max()),
+               "Size of output exceeds the column size limit",
```
karthikeyann, line 168:
```suggestion
               "Size of output exceeds the character size limit",
```

cpp/src/strings/utilities.cu
davidwendt: It will always be compared to an int64 variable.
Is there a reason to not support int64?

karthikeyann: Maximum value `LIBCUDF_LARGE_STRINGS_THRESHOLD` that's valid is `std::numeric_limits<int32_t>::max()` for string column purpose. It's representable with `int32_t`. `int64_t` is not necessary, unless there is a reason besides string column 32bit/64bit offsets choice.

What happens if a user sets `LIBCUDF_LARGE_STRINGS_THRESHOLD`  > int32 max? what's the expected behaviour? We should document valid range and behaviour of this environment variable in this function documentation.

ttnghia (approved):

karthikeyann (approved):
Nice work!  🚀  This will help a lot during testing long strings developments.

PR #14614, opened 2023-12-12
Consolidate DataFrame.__init__ logic to prepare data before calling super

mroeschke:
## Description
I noticed that `DataFrame.__init__` essentially has the following pattern

```python
super().__init__()

if condition:
    self._data = this
elif condition
    self._data = that

self._data.attribute = other
```

I find this pattern fairly brittle and leads to diverging paths for validation and coercion logic. This refactor essentials creates a `ColumnAccessor` from the inputs first and then passes that to `super`, then does post processing that all the branches can share.

This refactor does not touch when `data` is a `list`

Fixes the following bugs:

* Ensure `DataFrame(dict)` with tuple keys fill with NA instead of empty string like pandas
* Ensure `DataFrame(DataFrame(...), index=, column=)` reindexes like pandas
* Ensure `DataFrame(dict)` with only scalar values raises like pandas

## Checklist


python/cudf/cudf/core/dataframe.py
```diff
@@ -6482,11 +6519,15 @@ def select_dtypes(self, include=None, exclude=None):
...
-                df._insert(len(df._data), k, col)
+        if inclusion:
+            for k, col in self._data.items():
+                infered_type = cudf_dtype_from_pydata_dtype(col.dtype)
+                if infered_type in inclusion:
+                    df._insert(len(df._data), k, col)
+        else:
+            df._data.rangeindex = self._data.rangeindex
+            df._data.multiindex = self._data.multiindex
+            df._data.label_dtype = self._data.label_dtype
```
wence-, line 6530: **question** Why not transfer this information for the `inclusion` case? Is it because that destroys the information?

python/cudf/cudf/core/dataframe.py
```diff
@@ -6482,11 +6519,15 @@ def select_dtypes(self, include=None, exclude=None):
...
-        for k, col in self._data.items():
-            infered_type = cudf_dtype_from_pydata_dtype(col.dtype)
-            if infered_type in inclusion:
-                df._insert(len(df._data), k, col)
+        if inclusion:
+            for k, col in self._data.items():
+                infered_type = cudf_dtype_from_pydata_dtype(col.dtype)
+                if infered_type in inclusion:
+                    df._insert(len(df._data), k, col)
+        else:
```
wence-, line 6527: **suggestion (non-blocking, because it would be a bigger redo)**. Why is this function not implemented as:

1. determine columns to keep
2. `return self.loc[:, columns_to_keep]`

Then we could promote the handling of `self._data.rangeindex` and friends to `loc` (where it should presumably also be handled).

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1908,6 +1914,10 @@ def _slice(self, arg: slice, keep_index: bool = True) -> Self:
                 ),
                 keep_index=keep_index,
             )
+            result._data.rangeindex = col_was_rangeindex
+            result._data.multiindex = col_was_multiindex
+            result._data.label_dtype = col_label_dtype
+            return result
```
wence-, line 1920: **suggestion** Is `frame._slice` the only place where we need to care about carrying over this information? It seems like it might be necessary generally. Hence, should we move this to `IndexedFrame._gather`?

python/cudf/cudf/core/indexed_frame.py
```diff
@@ -1923,6 +1933,10 @@ def _slice(self, arg: slice, keep_index: bool = True) -> Self:
             None if has_range_index or not keep_index else self._index.names,
         )
 
+        result._data.rangeindex = col_was_rangeindex
+        result._data.multiindex = col_was_multiindex
+        result._data.label_dtype = col_label_dtype
```
wence-, line 1938: Similarly here, should `_from_columns_like_self` handle this transfer of information?

python/cudf/cudf/tests/test_dataframe.py
```diff
@@ -7627,7 +7636,11 @@ def test_dataframe_append_dataframe(df, other, sort, ignore_index):
     if expected.shape != df.shape:
         assert_eq(expected.fillna(-1), actual.fillna(-1), check_dtype=False)
     else:
-        assert_eq(expected, actual, check_index_type=not gdf.empty)
+        # pandas returns Index[object] while this should be an empty RangeIndex
+        # for empty df/other
```
wence-, line 7640: **question** are these pandas bugs that we should mark somehow?

python/cudf/cudf/core/dataframe.py
```diff
@@ -758,77 +777,121 @@ def __init__(
...
+            label_dtype=col_dtype,
+        )
+        super().__init__(col_accessor, index=index)
+        self._check_data_index_length_match()
+        if columns_from_data is not None:
+            # TODO: This there a better way to do this?
+            columns_from_data = as_index(columns_from_data)
+            reindexed = self.reindex(
+                columns=columns_from_data.to_pandas(), copy=False
+            )
```
wence-, line 875: **question** What are you trying to do conceptually that you would like a better way for?

python/cudf/cudf/core/dataframe.py
```diff
@@ -665,38 +665,47 @@ class DataFrame(IndexedFrame, Serializable, GetAttrGetItemMixin):
     def __init__(
         self, data=None, index=None, columns=None, dtype=None, nan_as_null=True
     ):
-        super().__init__()
+        col_is_rangeindex = False
```
wence-, line 668: **suggestion/discussion point** Even after this heroic refactoring to make things clearer, this `__init__` method is still _very long_. I haven't yet reviewed everything in detail because I found it quite hard to follow when things are preprocessing to deliver information to a later part of the function compared with preprocessing to produce the final result.

Hence, would it make sense to write the different cases as free functions (or `@staticmethod`s) so that we have something that then looks like:

```
if case_a:
   preprocessed_args = handle_case_a(...)
elif case_b:
   preprocessed_args = handle_case_b(...)

# or whatever
super().__init__(preprocessed_args)
```

WDYT?

mroeschke:
Sorry for the notification noise. I'll reopen this PR to reset

Issue #14160, opened 2023-09-21
Errors and deprecation warnings from numba 0.58


bdice:
The release of numba 0.58 has surfaced several errors and deprecation warnings. Below is an example:

```
numba.core.errors.TypingError: Internal error at resolving type of attribute "sum" of "group".
...
numba.core.errors.NumbaPendingDeprecationWarning: Failed in cuda mode pipeline (step: nopython frontend)
Code using Numba extension API maybe depending on 'old_style' error-capturing, which is deprecated and will be replaced by 'new_style' in a future release. See details at https://numba.readthedocs.io/en/latest/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors
```

See details for a full log.

<details><summary>Details</summary>
<p>

```
____________________ test_groupby_apply_return_df[<lambda>] ____________________
[gw0] linux -- Python 3.10.13 /pyenv/versions/3.10.13/bin/python

typingctx = <numba.cuda.target.CUDATypingContext object at 0x7f72be9a3340>
targetctx = <numba.cuda.target.CUDATargetContext object at 0x7f7152cdb100>
interp = <numba.core.ir.FunctionIR object at 0x7f7152b50a60>
args = (Record([('a', {'type': Group(int64, int64), 'offset': 0, 'alignment': None, 'title': None, }), ('b', {'type': Group(int64, int64), 'offset': 24, 'alignment': None, 'title': None, })], 48, True),)
return_type = None, locals = {}, raise_errors = True

    def type_inference_stage(typingctx, targetctx, interp, args, return_type,
                             locals={}, raise_errors=True):
        if len(args) != interp.arg_count:
            raise TypeError("Mismatch number of argument types")
        warnings = errors.WarningsFixer(errors.NumbaWarning)
    
        infer = typeinfer.TypeInferer(typingctx, interp, warnings)
        callstack_ctx = typingctx.callstack.register(targetctx.target, infer,
                                                     interp.func_id, args)
        # Setup two contexts: 1) callstack setup/teardown 2) flush warnings
        with callstack_ctx, warnings:
            # Seed argument types
            for index, (name, ty) in enumerate(zip(interp.arg_names, args)):
                infer.seed_argument(name, index, ty)
    
            # Seed return type
            if return_type is not None:
                infer.seed_return(return_type)
    
            # Seed local types
            for k, v in locals.items():
                infer.seed_type(k, v)
    
            infer.build_constraint()
            # return errors in case of partial typing
[...]

PR #14616, opened 2023-12-12
Don't constrain `numba<0.58`

brandon-b-miller:
This PR removes the constraint for numba 0.58 in our dependencies. 

Closes https://github.com/rapidsai/cudf/issues/14160


bdice (approved):

bdice:
Should we test with development builds of numba 0.59 to ensure we don't see breakage from that release? I'm happy either way, just wanted to pose the question. cc: @gmarkall

galipremsagar (approved):

raydouglass (approved):

jakirkham:
> Should we test with development builds of numba 0.59 to ensure we don't see breakage from that release?

Could try adding their dev channel `numba/label/ci` and seeing what happens. Would make sure it is installing the latest `numba`. Have noticed that doesn't always work out-of-the-box with the dev channel

dependencies.yaml
```diff
@@ -184,6 +184,7 @@ files:
     includes:
       - test_python_common
 channels:
+  - numba/label/ci
```
jakirkham, line 187: It looks like this only affects the environment files (so offline testing). We may need a different approach if we want to test on CI

brandon-b-miller: Can you explain what you mean by offline testing? I see the following in the `conda-python-cudf-tests` CI jobs with the latest commit:

```
  + numba                                   0.59.0dev0  np1.21py3.10hfbc5805_g7c1ea796a_596  numba/label/ci          4MB
```

conda/environments/all_cuda-120_arch-x86_64.yaml
```diff
@@ -1,6 +1,7 @@
 # This file is generated by `rapids-dependency-file-generator`.
 # To make changes, edit ../../dependencies.yaml and run `rapids-dependency-file-generator`.
 channels:
+- numba/label/ci
```
jakirkham, line 4: The channel only seems to show up environment files, which IIUC are only used for offline testing and development

For CI testing think we need this channel in CI files that `dependencies.yaml` doesn't affect AFAIK

dependencies.yaml
jakirkham: Yeah it seems this is only used in environment files that are AFAIK not used in CI

ajschmidt8: `dependencies.yaml` is used by CI:

- https://github.com/rapidsai/cudf/blob/cee642916cfc3b8df73e819bf3bc50f1b9fc684f/ci/test_cpp_common.sh#L12-L15

You can see the environment it generates for CI tests in the logs below:

- https://github.com/rapidsai/cudf/actions/runs/7211137353/job/19646850155?pr=14616#step:7:27

jakirkham: It's not used in [the build though]( https://github.com/rapidsai/cudf/actions/runs/7211137353/job/19646852176#step:7:498 )

Also it appears [the tests don't use it either]( https://github.com/rapidsai/cudf/actions/runs/7211137353/job/19647545108#step:7:949 )

jakirkham:
For clarity don't think the test actually worked: https://github.com/rapidsai/cudf/pull/14616#discussion_r1427200453

dependencies.yaml
brandon-b-miller: @ajschmidt8 could we enforce numba 0.59 for testing by pinning the dependency [here](https://github.com/rapidsai/cudf/blob/branch-24.02/dependencies.yaml#L596C3-L596C19) perhaps? Maybe along with adding the extra channel back somewhere?

jakirkham: Maybe we could fold these into one command

https://github.com/rapidsai/cudf/blob/cee642916cfc3b8df73e819bf3bc50f1b9fc684f/ci/test_python_common.sh#L19

https://github.com/rapidsai/cudf/blob/cee642916cfc3b8df73e819bf3bc50f1b9fc684f/ci/test_python_common.sh#L37-L40

brandon-b-miller: Tried installing it as a step at the end - will see what happens.

brandon-b-miller: With https://github.com/rapidsai/cudf/pull/14616/commits/5ffac4d6ab66b12c0ff3a24252b536ae23b4a78a and https://github.com/rapidsai/cudf/pull/14616/commits/8e67d620fa8e0748482e74b5f48c5c0eb77bf671 I think I managed to get it to install numba 0.59 just before the cuDF tests run. @jakirkham let me know if this seems like this covers us for now.

jakirkham: Thanks Brandon! 🙏

Looking at [the Conda cuDF Python CI test]( https://github.com/rapidsai/cudf/actions/runs/7225028551/job/19688893979?pr=14616#step:7:944 ), it seems to oddly be getting an older version from `conda-forge`

```
  - numba                 0.58.1  py310h7dc5dd1_0                      conda-forge             Cached
  + numba                 0.57.1  py310h0f6aa51_0                      conda-forge                4MB
```

jakirkham: Ah sorry was looking at the latest commit. Should I be looking at the checks on the commit right before?

If I do, I see the following in the [Conda cuDF Python CI test]( https://github.com/rapidsai/cudf/actions/runs/7223085574/job/19682887349#step:7:1079 ):

```
  Package        Version  Build                                Channel              Size
──────────────────────────────────────────────────────────────────────────────────────────
  Install:
──────────────────────────────────────────────────────────────────────────────────────────

  + zlib          1.2.13  hd590300_5                           conda-forge          93kB
  + numba     0.59.0dev0  np1.21py3.10ha54aa7d_gfaaf5f891_609  numba/label/ci        4MB

  Upgrade:
──────────────────────────────────────────────────────────────────────────────────────────

  - llvmlite      0.40.1  py310h1b8f574_0                      conda-forge        Cached
  + llvmlite  0.42.0dev0  py310he1b5a44_74                     numba/label/ci       35MB
```

This looks good to me

Thanks again for working on this Brandon! 🙏

PR #14623, opened 2023-12-13
Add (implicit) handling for torch tensors in is_scalar

wence-:
## Description
PyTorch tensors advertise that they support the number API, and hence answer "True" to the question pd.api.types.is_scalar(torch_tensor). This trips up some of our data ingest, since in as_index we check if the input is a scalar (and raise) before handing off to as_column. To handle this, if we get True back from pandas' is_scalar call, additionally check that the object has an empty shape attribute (if it exists).

See also:

- https://github.com/pytorch/pytorch/issues/99646
- https://github.com/pandas-dev/pandas/issues/52701

## Checklist

### DIFF
--- a/python/cudf/cudf/api/types.py
+++ b/python/cudf/cudf/api/types.py
@@ -135,7 +135,17 @@ def is_scalar(val):
             cudf._lib.scalar.DeviceScalar,
             cudf.core.tools.datetimes.DateOffset,
         ),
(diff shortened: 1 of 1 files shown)


wence-:
cc @shwina @puririshi98

shwina (approved):
Ugh but LGTM

PR #14637, opened 2023-12-14
Fix invalid memory access in Parquet reader

### LINKED ISSUE #14633: [BUG] Memcheck error cudf::io::parquet::detail::<unnamed>::gpuComputeStringPageBounds<unsigned char>()
**Describe the bug**
Nightly memcheck build is reporting out-of-bounds read error in `cudf::io::parquet::detail::<unnamed>::gpuComputeStringPageBounds<unsigned char>()` when running gtest `ParquetChunkedReaderTest.TestChunkedReadWithString`

```
[ RUN      ] ParquetChunkedReaderTest.TestChunkedReadWithString
========= Invalid __global__ read of size 8 bytes
=========     at 0x3a50 in void cudf::io::parquet::detail::<unnamed>::gpuComputeStringPageBounds<unsigned char>(cudf::io::parquet::detail::PageInfo *, cudf::device_span<const cudf::io::parquet::detail::ColumnChunkDesc, (unsigned long)18446744073709551615>, unsigned long, unsigned long)
=========     by thread (0,0,0) in block (5,0,0)
=========     Address 0x7f04727bcc08 is out of bounds
=========     and is 489 bytes after the nearest allocation at 0x7f04727bca00 of size 32 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x3344e0]
=========                in /usr/lib/x86_64-linux-gnu/libcuda.so.1
=========     Host Frame: [0x1488c]
=========                in /conda/envs/rapids/lib/libcudart.so.11.0
=========     Host Frame:cudaLaunchKernel [0x6c318]
=========                in /conda/envs/rapids/lib/libcudart.so.11.0
=========     Host Frame:cudf::io::parquet::detail::ComputePageStringSizes(cudf::detail::hostdevice_vector<cudf::io::parquet::detail::PageInfo>&, cudf::detail::hostdevice_vector<cudf::io::parquet::detail::ColumnChunkDesc> const&, rmm::device_uvector<unsigned char>&, unsigned long, unsigned long, int, unsigned int, rmm::cuda_stream_view) [0x14fcd2d]
=========                in /conda/envs/rapids/lib/libcudf.so
=========     Host Frame:cudf::io::parquet::detail::reader::impl::decode_page_data(unsigned long, unsigned long) [0x151fe2a]
=========                in /conda/envs/rapids/lib/libcudf.so
=========     Host Frame:cudf::io::parquet::detail::reader::impl::read_chunk_internal(bool, std::optional<std::reference_wrapper<cudf::ast::expression const> >) [0x15234d8]
=========                in /conda/envs/rapids/lib/libcudf.so
=========     Host Frame:cudf::io::parquet::detail::reader::impl::read_chunk() [0x1524292]
=========                in /conda/envs/rapids/lib/libcudf.so
=========     Host Frame:cudf::io::parquet::detail::chunked_reader::read_chunk() const [0x1513303]
=========                in /conda/envs/rapids/lib/libcudf.so
=========     Host Frame:cudf::io::chunked_parquet_reader::read_chunk() const [0x1400f06]
=========                in /conda/envs/rapids/lib/libcudf.so
=========     Host Frame:(anonymous namespace)::chunked_read(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long, unsigned long) [clone .constprop.0] [0x2d1545]
=========                in /cudf/cpp/build/gtests/PARQUET_TEST
=========     Host Frame:ParquetChunkedReaderTest_TestChunkedReadWithString_Test::TestBody() [0x2d6311]
=========                in /cudf/cpp/build/gtests/PARQUET_TEST

```

The following can be used to reproduce the error.
```
compute-sanitizer --tool memcheck gtests/PARQUET_TEST --gtest_filter=ParquetChunkedReaderTest.TestChunkedReadWithString --rmm_mode=cuda
```

This error may have been introduced sometime in 23.12 but was masked by the memcheck failures due to #14440 which was only recently resolved in 24.02.
@etseidl @nvdbaranec @vuule

etseidl:
## Description
Fixes #14633

When reading files in multiple passes, some pointer fields in `ColumnChunkDesc` that point to transient memory are not cleared out at the end of each pass. This can lead to trying to dereference deallocated memory during Parquet reader string preprocessing.

## Checklist

### DIFF
--- a/cpp/src/io/parquet/page_decode.cuh
+++ b/cpp/src/io/parquet/page_decode.cuh
@@ -243,6 +243,10 @@ __device__ cuda::std::pair<int, int> gpuDecodeDictionaryIndices(page_state_s* s,
   int pos            = s->dict_pos;
   int str_len        = 0;
 
+  // NOTE: racecheck warns about a RAW involving s->dict_pos, which is likely a false positive
+  // because the only path that does not include a sync will lead to s->dict_pos being overwritten
+  // with the same value
+
   while (pos < target_pos) {
     int is_literal, batch_len;
     if (!t) {
@@ -356,6 +360,10 @@ inline __device__ int gpuDecodeRleBooleans(page_state_s* s, state_buf* sb, int t
 {
   uint8_t const* end = s->data_end;
   int64_t pos        = s->dict_pos;
+
+  // NOTE: racecheck warns about a RAW involving s->dict_pos, which is likely a false positive
+  // because the only path that does not include a sync will lead to s->dict_pos being overwritten
+  // with the same value
 
   while (pos < target_pos) {
     int is_literal, batch_len;
@@ -549,6 +557,9 @@ __device__ void gpuDecodeStream(
     batch_coded_count += batch_len;
     value_count += batch_len;
   }
+  // issue #14597
+  // racecheck reported race between reads at the start of this function and the writes below
+  __syncwarp();
 
   // update the stream info
   if (!t) {
@@ -681,6 +692,9 @@ __device__ void gpuUpdateValidityOffsetsAndRowIndices(int32_t target_input_value
                                                       level_t const* const def,
                                                       int t)
 {
+  // exit early if there's no work to do
+  if (s->input_value_count >= target_input_value_count) { return; }
+
   // max nesting depth of the column
   int const max_depth       = s->col.max_nesting_depth;
   bool const has_repetition = s->col.max_level[level_type::REPETITION] > 0;
--- a/cpp/src/io/parquet/page_decode.cuh
+++ b/cpp/src/io/parquet/page_decode.cuh
@@ -1014,6 +1014,12 @@ struct mask_filter {
   }
 };
 
+enum class page_processing_stage {
+  PREPROCESS,
+  STRING_BOUNDS,
+  DECODE,
+};
+
 /**
  * @brief Sets up block-local page state information from the global pages.
  *
@@ -1036,8 +1042,7 @@ inline __device__ bool setupLocalPageInfo(page_state_s* const s,
                                           size_t min_row,
                                           size_t num_rows,
                                           Filter filter,
-                                          bool is_decode_step,
-                                          bool is_bounds_step)
+                                          page_processing_stage stage)
 {
   int t = threadIdx.x;
 
@@ -1128,7 +1133,8 @@ inline __device__ bool setupLocalPageInfo(page_state_s* const s,
   //
   // NOTE: this check needs to be done after the null counts have been zeroed out
   bool const has_repetition = s->col.max_level[level_type::REPETITION] > 0;
-  if ((is_decode_step || is_bounds_step) && s->num_rows == 0 &&
+  if ((stage == page_processing_stage::STRING_BOUNDS || stage == page_processing_stage::DECODE) &&
+      s->num_rows == 0 &&
       !(has_repetition && (is_bounds_page(s, min_row, num_rows, has_repetition) ||
                            is_page_contained(s, min_row, num_rows)))) {
     return false;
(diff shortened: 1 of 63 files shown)


etseidl:
@nvdbaranec this PR addresses the issue, but is not the most elegant fix. I tried to find a location to zero out the pointers where I could piggyback on an existing H2D copy of the chunks, but didn't have much luck. Could you please take a look and suggest a better fix?

etseidl:
An alternative that doesn't require copying the chunks array to the device is to zero out the pointers at the beginning of `gpuComputeStringPageBounds`.

Also, the read is somewhat benign since the result of the dereference is never used. This _shouldn't_ impact users if it went out in 23.12.

Ran the full PARQUET_TEST through compute-sanitizer after the change and got no additional errors reported.

etseidl:
Yet another option may be to add another boolean argument to `setupLocalPageInfo`.  This would allow passing `is_decode_step` as `false` but still get the early return for non-overlapping pages. Then we'd avoid the bad pointer dereference altogether.

cpp/src/io/parquet/page_string_decode.cu
```diff
@@ -616,7 +616,9 @@ __global__ void __launch_bounds__(preprocess_block_size) gpuComputeStringPageBou
 
   // setup page info
   auto const mask = BitOr(decode_kernel_mask::STRING, decode_kernel_mask::DELTA_BYTE_ARRAY);
-  if (!setupLocalPageInfo(s, pp, chunks, min_row, num_rows, mask_filter{mask}, true)) { return; }
```
vuule, line 619: Why does this PR change `is_decode_step` value in some of these calls?

etseidl: The string preprocessing was passing that as true, leading the setup call to believe the output buffers were valid and thus accessing invalid memory. With the new flag true and the old flag false, we get the behavior that was originally desired, but can now skip the bad pointer arithmetic.

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -1126,7 +1128,7 @@ inline __device__ bool setupLocalPageInfo(page_state_s* const s,
   //
   // NOTE: this check needs to be done after the null counts have been zeroed out
   bool const has_repetition = s->col.max_level[level_type::REPETITION] > 0;
-  if (is_decode_step && s->num_rows == 0 &&
+  if ((is_decode_step or is_bounds_step) && s->num_rows == 0 &&
```
ttnghia, line 1131: We are using `&&` later in the same line.
```suggestion
  if ((is_decode_step || is_bounds_step) && s->num_rows == 0 &&
```

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -1387,7 +1389,7 @@ inline __device__ bool setupLocalPageInfo(page_state_s* const s,
 
       // if we're in the decoding step, jump directly to the first
       // value we care about
-      if (is_decode_step) {
+      if (is_decode_step or is_bounds_step) {
```
ttnghia, line 1392: Let's use symbol for consistency.
```suggestion
      if (is_decode_step || is_bounds_step) {
```

ttnghia (approved):

cpp/src/io/parquet/page_decode.cuh
```diff
@@ -1387,7 +1389,7 @@ inline __device__ bool setupLocalPageInfo(page_state_s* const s,
 
       // if we're in the decoding step, jump directly to the first
       // value we care about
```
vuule, line 1391: I think this comment needs to be updated.

vuule (approved):
Two adjacent bool parameters are a bit sketchy, but clean up can be a separate PR

cpp/src/io/parquet/page_decode.cuh
etseidl: Could change this to move the `is_bonds_step` test inside the `else` block...I think all that's necessary is to just not zero out those values.

etseidl:
> Two adjacent bool parameters are a bit sketchy, but clean up can be a separate PR

Well, it's a tri-state: preprocessing (false false), string bounds (false true), decode (true false). Could add an enum...

ttnghia:
> Well, it's a tri-state: preprocessing (false false), string bounds (false true), decode (true false). Could add an enum...

Yeah an enum (with comment) should be much clearer, instead of just `true/false`.

vuule:
The change to enum is great. No need to decipher the steps.

PR #14641, opened 2023-12-15
Use instance over is_foo_dtype

mroeschke:
## Description
Similar to https://github.com/rapidsai/cudf/pull/14638, use isinstance when we know we are checking a dtype instance

## Checklist


python/cudf/cudf/_lib/groupby.pyx
```diff
@@ -73,6 +73,21 @@ _DECIMAL_AGGS = {
 ctypedef const scalar constscalar
 
 
+cdef get_valid_aggregation(object dtype):
```
vyasr, line 76: Rather than a cdef function, I wonder if it would be worth using a `functools.singledispatch` here. We're not benefiting from the `cdef` because everything is a pure Python type, and the `isinstance` cascade isn't free, especially if groupby is run many times.

vyasr (approved):
One suggested optimization, but looks good.

python/cudf/cudf/_lib/groupby.pyx
mroeschke: Good suggestion, singledispatch is a good alternative so I'll use that

PR #14642, opened 2023-12-15
Convert `FieldType` to scoped enum

vuule:
## Description
Switch to scoped enum (`enum class`); they are better because, well, values now have a scope.
Another benefit in this case - values are now named consistently with compact protocol.
De-duplicated some code, now that more static_casts are required and duplication stands out more.

## Checklist


cpp/src/io/parquet/compact_protocol_writer.hpp
```diff
@@ -78,7 +78,14 @@ class CompactProtocolFieldWriter {
 
   uint32_t put_int(int64_t v);
 
-  void put_field_header(int f, int cur, int t);
+  template <typename T>
+  void put_packed_type_byte(T high_bits, FieldType t)
+  {
+    uint8_t const clamped_high_bits = std::min(std::max(high_bits, t{0}), T{0xf});
```
etseidl, line 84:
```suggestion
    uint8_t const clamped_high_bits = std::min(std::max(high_bits, T{0}), T{0xf});
```
?

cpp/src/io/parquet/parquet_common.hpp
```diff
@@ -140,19 +140,20 @@ enum BoundaryOrder {
...
-  ST_FLD_DOUBLE = 7,
-  ST_FLD_BINARY = 8,
-  ST_FLD_LIST   = 9,
-  ST_FLD_SET    = 10,
-  ST_FLD_MAP    = 11,
-  ST_FLD_STRUCT = 12,
+enum class FieldType : uint8_t {
+  BOOLEAN_TRUE  = 1,
+  BOOLEAN_FALSE = 2,
+  BYTE          = 3,
```
etseidl, line 146: Thrift spec actually uses `I8` here.

cpp/src/io/parquet/compact_protocol_writer.hpp
vuule: I'm, uh, using a case insensitive compiler locally

cpp/src/io/parquet/compact_protocol_reader.cpp
```diff
@@ -45,28 +45,36 @@ class parquet_field {
...
+  CUDF_EXPECTS(type == static_cast<int>(expected),
                "expected " + field_type_string(expected) + " field, got " +
                  field_type_string(static_cast<FieldType>(type)) + " field instead");
 }
 
+void assert_bool_field_type(int type)
+{
+  CUDF_EXPECTS(type == static_cast<int>(FieldType::BOOLEAN_TRUE) ||
+                 type == static_cast<int>(FieldType::BOOLEAN_FALSE),
+               "expected bool field, got " + field_type_string(static_cast<FieldType>(type)) +
```
etseidl, line 74:
```suggestion
  auto const field_type = static_cast<FieldType>(type);
  CUDF_EXPECTS(field_type == FieldType::BOOLEAN_TRUE ||
                 field_type == FieldType::BOOLEAN_FALSE,
               "expected bool field, got " + field_type_string(field_type) +
```
Fewer casts

vuule: In your version `field_type` could be an invalid `FieldType` value. I wanted to avoid having potentially invalid `FieldType` vars by always casting in the other direction. But now I see that I'm passing `type` to `field_type_string` either way, so might as well avoid some casts.

ttnghia (approved):

cpp/src/io/parquet/compact_protocol_reader.cpp
```diff
@@ -439,24 +450,25 @@ class parquet_field_optional : public parquet_field {
...
+  switch (t_enum) {
+    case FieldType::BOOLEAN_TRUE:
+    case FieldType::BOOLEAN_FALSE: break;
+    case FieldType::I16:
+    case FieldType::I32:
+    case FieldType::I64: get_u64(); break;
+    case FieldType::I8: skip_bytes(1); break;
+    case FieldType::DOUBLE: skip_bytes(8); break;
+    case FieldType::BINARY: skip_bytes(get_u32()); break;
+    case FieldType::LIST: [[fallthrough]];
```
mythrocks, line 463: Is there a recommendation on when to use `[[fallthrough]]`?  I'm not sure it's used consistently here.
Would it be wrong to simply remove the `[[fallthrough]]` here?  We don't seem to have used it above.

mythrocks (approved):
LGTM!

cpp/src/io/parquet/compact_protocol_reader.cpp
etseidl: The wording I've always seen is something to the effect that multiple cases for a block do not require `[[falthrough]]`. It's really intended for the case:
```c++
  case A:
    { // do some stuff }
    [[fallthrough]]; // let compiler know the lack of a `break` is intentional
  case B:
    { // do more stuff }
    break;
```

vuule: lol, first time reading this guideline. Makes sense.
I tend to use `[[fallthrough]]` for multiple cases as well. I thought compiler might warn for any missing breaks (eventually, or with some flags).

vuule: Fixed, thanks for pointing this out. My fallthrough tyranny has ended.

vyasr (approved):

PR #14651, opened 2023-12-19
Refactor numpy array input in as_column

mroeschke:
## Description
Simplifies the numpy array input logic to `as_column` to be

```
if object/string dtype like:
    # parse with pandas with inference
elif numeric-like dtype or datelike with nat:
    # parse with pyarrow (due to np.nan/np.nat/nan_is_null handling)
else:
    # create column from buffer
```

## Checklist


python/cudf/cudf/core/column/column.py
```diff
@@ -2215,18 +2215,42 @@ def as_column(
...
+                arbitrary = arbitrary.astype(new_type)
+
+            is_nat = np.isnat(arbitrary)
+            if (nan_as_null is None or nan_as_null) and np.isnat(
+                arbitrary
+            ).any():
+                return as_column(
+                    pa.array(arbitrary),
+                    dtype=dtype,
+                    nan_as_null=nan_as_null,
```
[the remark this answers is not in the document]
mroeschke: In the pyarrow route, we want to convert NaT to NA which pyarrow does nicely be default (`nan_as_null=True`)

In the buffer route, we want to consider NaT as NA in the mask but maintain NaT as a value (`nan_as_null=False`)

I'll leave a comment about this

vyasr:
Is this intended to still be in draft?

mroeschke:
> Is this intended to still be in draft?

Thanks for checking in on this. Yes, I think there are some failing tests I am still working through

python/cudf/cudf/core/column/column.py
```diff
@@ -2032,117 +2037,74 @@ def as_column(
...
+                return as_column(
+                    cupy.asarray(arbitrary, dtype=arbitrary[0].dtype),
+                    nan_as_null=nan_as_null,
+                    dtype=dtype,
+                    length=length,
+                )
+            if pd.isna(arbitrary).any():
+                arbitrary = pa.array(arbitrary)
+            else:
+                arbitrary = pd.Series(arbitrary)
```
vyasr, line 2064: What happens if we just always use pyarrow here? Do we lose something from non-nullable data that pandas captures better?

python/cudf/cudf/core/column/column.py
```diff
@@ -2032,117 +2037,74 @@ def as_column(
...
+            mask = None
+            if is_nat.any():
+                if nan_as_null is None or nan_as_null:
+                    # Convert NaT to NA, which pyarrow does by default
+                    return as_column(
+                        pa.array(arbitrary),
+                        dtype=dtype,
+                        nan_as_null=nan_as_null,
+                    )
+                else:
```
vyasr, line 2096: Superfluous else since we're returning above.

python/cudf/cudf/core/column/column.py
mroeschke: There are cases where pandas will infer `dtype=object` into a non-object type that would help match pandas behavior e.g. 

```python
In [1]: import cudf, pandas as pd, numpy as np

# e.g a `to_numpy()` round-trip
In [2]: pd.Series(np.array([pd.Timestamp(2020, 1, 1)], dtype=object))
Out[2]: 
0   2020-01-01
dtype: datetime64[ns]
```

I'll add a comment about that here

python/cudf/cudf/core/column/column.py
mroeschke: Good point. Removed

wence- (approved):
I think this one looks good too, thanks for all the ongoing cleanup work.

python/cudf/cudf/core/column/column.py
```diff
@@ -2032,117 +2037,74 @@ def as_column(
...
-            shape
-            and shape[0]
-            and hasattr(arbitrary[0], "__cuda_array_interface__")
-        ):
+        if arbitrary.dtype.kind in "OSU":
+            # Handle case that `arbitrary` elements are cupy arrays
+            if len(arbitrary) > 0 and hasattr(
+                arbitrary[0], "__cuda_array_interface__"
+            ):
+                return as_column(
```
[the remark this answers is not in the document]
mroeschke: Correct. I had added this since I thought there was a unit test that would hit this branch. I am not longer encountering that after running the test suite again so I'll actually remove it for now

PR #14653, opened 2023-12-19
Update CONTRIBUTING for pyproject-only builds

vyasr:
## Description

Now that we no longer have setup.py, builds must be handled by a front-end like `pip install`.

## Checklist


DanialJavady96:
Wanted to comment from an outsiders POV, the wording in the docs say

`First, build the libcudf C++ library following the steps above` but you will need to build cudf & libcudf, not just libcudf right? Maybe "build the libcudf C++ library and cudf library" would be helpful

vyasr:
That's a good point, there's really no reason not to prefer build.sh here. I'll update again.

mroeschke (approved):
LGTM (also great to have an option to install in editable mode too!)

mroeschke:
Noting that when I build with `--pydevelop`, I get this error when importing cudf

```python
In [1]: import cudf
FileNotFoundError                         Traceback (most recent call last)
Cell In[1], line 1
----> 1 import cudf

File ~/python/cudf/cudf/__init__.py:9
      6 from cudf.utils._numba import _setup_numba
      7 from cudf.utils.gpu_utils import validate_setup
----> 9 _setup_numba()
     10 validate_setup()
     12 import cupy

File ~/python/cudf/cudf/utils/_numba.py:122, in _setup_numba()
    120 if versions != NO_DRIVER:
    121     driver_version, runtime_version = versions
--> 122     ptx_toolkit_version = _get_cuda_version_from_ptx_file(CC_60_PTX_FILE)
    124     # MVC is required whenever any PTX is newer than the driver
    125     # This could be the shipped PTX file or the PTX emitted by
    126     # the version of NVVM on the user system, the latter aligning
    127     # with the runtime version
    128     if (driver_version < ptx_toolkit_version) or (
    129         driver_version < runtime_version
    130     ):

File ~/python/cudf/cudf/utils/_numba.py:157, in _get_cuda_version_from_ptx_file(path)
    137 def _get_cuda_version_from_ptx_file(path):
    138     """
    139     https://docs.nvidia.com/cuda/parallel-thread-execution/
    140     Each PTX module must begin with a .version
   (...)
    155 
    156     """
--> 157     with open(path) as ptx_file:
    158         for line in ptx_file:
    159             if line.startswith(".version"):

FileNotFoundError: [Errno 2] No such file or directory: '.../python/cudf/cudf/utils/../core/udf/shim_60.ptx'
```

PR #14656, opened 2023-12-20
Unconditionally export the CCCL path

vyasr:
## Description

We need to unconditionally export the CCCL path because the condition can no longer be evaluated since the export is happening before the call to `rapids_cpm_cccl` as of #14655. Since cudf should in practice always be getting its own local copy of CCCL, it should be safe to have this export unconditionally; we will never use an externally provided copy of CCCL.

## Checklist

### DIFF
--- a/cpp/cmake/thirdparty/get_cccl.cmake
+++ b/cpp/cmake/thirdparty/get_cccl.cmake
@@ -31,7 +31,6 @@ function(find_and_configure_cccl)
   rapids_export_find_package_root(
     INSTALL CCCL [=[${CMAKE_CURRENT_LIST_DIR}/../../../include/libcudf/lib/rapids/cmake/cccl]=]
     EXPORT_SET cudf-exports
-    CONDITION CCCL_SOURCE_DIR
   )
 
   # Find or install CCCL with our custom set of patches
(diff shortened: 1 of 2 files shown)


vyasr:
I downloaded these artifacts into a cuspatial devcontainer and verified that it fixed the build issues observed in rapidsai/cuspatial#1315.

cpp/cmake/thirdparty/get_cccl.cmake
```diff
@@ -23,15 +23,14 @@ function(find_and_configure_cccl)
 
   # Make sure we install cccl into the `include/libcudf` subdirectory instead of the default
   include(GNUInstallDirs)
-  set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/libcudf")
-  set(CMAKE_INSTALL_LIBDIR "${CMAKE_INSTALL_INCLUDEDIR}/lib")
+  set(CMAKE_INSTALL_LIBDIR "${CMAKE_INSTALL_INCLUDEDIR}/libcudf/lib")
```
robertmaynard, line 26: These shouldn't be necessary and indicate the root issue is in rapids-cmake/cpm/cccl.cmake

robertmaynard (changes requested):

cpp/cmake/thirdparty/get_cccl.cmake
robertmaynard: Root fix at: https://github.com/rapidsai/rapids-cmake/pull/513

vyasr: Updated this PR accordingly.

robertmaynard (approved):

Issue #12311, opened 2022-12-05
[FEA] Throw more descriptive error when file is not found in CUIO.


cwharris:
**Is your feature request related to a problem? Please describe.**
CUIO file utilities throw an unhelpful error when file is not found.
`Cannot query file size`
```
./build/morpheus/_lib/tests/test_libmorpheus.x --gtest_filter=TestMatxUtil.ReduceMax2dColMajor
Note: Google Test filter = TestMatxUtil.ReduceMax2dColMajor
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from TestMatxUtil
[ RUN      ] TestMatxUtil.ReduceMax2dColMajor
unknown file: Failure
C++ exception with description "cuDF failure at: /workspace/.conda-bld/work/cpp/src/io/utilities/file_io_utilities.cpp:34: Cannot query file size" thrown in the test body.
[  FAILED  ] TestMatxUtil.ReduceMax2dColMajor (13 ms)
[----------] 1 test from TestMatxUtil (13 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (13 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] TestMatxUtil.ReduceMax2dColMajor

 1 FAILED TEST
```

**Describe the solution you'd like**
Throw a more descriptive error when file is not found.

**Context**
This error was throw when attempting to open a csv file which did not exist.

GregoryKimball:
What file format were you trying to read? For text files the errors can be a bit weird because they accept string data as well as string filenames.

vuule:
This should happen with any reader, as long as they read from a file. The error is thrown from a utility class used in the file datasource.
I looked into this a bit, and it seems like we cannot be certain about the cause of the error, even with errno. If it helps, we can expand the error message to include something like "(possibly incorrect file name)".

Issue #9564, opened 2021-10-29
[FEA] Clarify non-existent subdirectory error messages for file writers


randerzander:
With latest cudf nightlies, attempting to write a file to a non-existend subdirectory fails as expected (matching Pandas), but the error is unclear:
```
RuntimeError: cuDF failure at: ../src/io/utilities/data_sink.cpp:36: Cannot open output file
```

Pandas's is a bit better:
```
IOError: [Errno 2] No such file or directory: './dir/name.csv'
```

It would be helpful for users to know if the write fails because the subdirectory didn't exist, as opposed to some other permissions or worse (if writing to a distributed fs) error.

rnyak:
I get same error `RuntimeError: cuDF failure at: ....: Cannot open output file` when the subdirectory doesn't exist but the error msg is not informative.. it is hard to tell what's the issue. I agree that It would be helpful for users to know that this is because the subdirectory doesn't exist.

PR #14662, opened 2023-12-20
Custom error messages for IO with nonexistent files

vuule:
## Description
Closes https://github.com/rapidsai/cudf/issues/12311; closes https://github.com/rapidsai/cudf/issues/9564
We return a somewhat cryptic error when opening a file that does not exist: "Cannot query file size".

With this change, we report whether the file exists, or, if the file does exist, what the errno value is after `open`.
Also added a check for the output files' directory in `file_sink`. This check is now also included in `file_wrapper`, just in case initialization order changes at some point.
Now we should always correctly report missing output file directory and missing input files.

## Checklist


cpp/src/io/utilities/data_sink.cpp
```diff
@@ -34,7 +36,13 @@ class file_sink : public data_sink {
...
   {
     _output_stream.open(filepath, std::ios::out | std::ios::binary | std::ios::trunc);
-    CUDF_EXPECTS(_output_stream.is_open(), "Cannot open output file");
+    if (!_output_stream.is_open()) {
+      auto const dir_path = std::filesystem::path(filepath).parent_path();
+      if (not std::filesystem::exists(dir_path)) {
+        CUDF_FAIL("Cannot open output file; directory does not exist");
+      }
+      CUDF_FAIL("Cannot open output file");
+    }
```
karthikeyann, line 45: would it help to make this code also similar to `cpp/src/io/utilities/file_io_utilities.cpp` ? That seems more detailed.

vuule: added errno. I originally excluded it because I didn't know if `ofstream::open` sets it.

ttnghia: Should we move this into `file_io_utilities`? As it should contain all file IO.

mythrocks: TIL `std::strerror`

I had to look up where the allocation happens for the string returned by `std::strerror()`. I'm not convinced it is threadsafe, given that it's allowed to reuse the same allocated space to return strings across threads.

Should we consider switching this to `std::strerror_s()`?

cpp/src/io/utilities/file_io_utilities.cpp
```diff
@@ -21,30 +21,43 @@
...
-  struct stat st;
-  CUDF_EXPECTS(fstat(file_descriptor, &st) != -1, "Cannot query file size");
-  return static_cast<size_t>(st.st_size);
+  auto const fd = open(filepath.c_str(), flags, mode);
+  if (fd != -1) { return fd; }
+
+  // save errno because it may be overwritten by subsequent calls
+  auto const err = errno;
+
+  if (auto const path = std::filesystem::path(filepath); flags & O_CREAT) {
```
mythrocks, line 40: Noice.

mythrocks (commented):
LGTM, save for the potential thread-safety issue.

cpp/src/io/utilities/data_sink.cpp
vuule: great catch! I'll look into `strerror_s`.

vuule: > Should we move this into file_io_utilities?

I don't think `file_sink` belongs in utilities, but I'll try to reuse this error checking and place the common code in `file_io_utilities`.

vuule: I should have just put a mutex in here. `strerror_s` portability is horrendous.
Our compiler does not support it, so I have to use `strerror_r`. Aaand, this one is problematic as well, see https://www.club.cc.cmu.edu/~cmccabe/blog_strerror.html

ttnghia (approved):

karthikeyann (approved):
TIL `[[noreturn]]`

PR #14663, opened 2023-12-20
Split parquet test into multiple files

etseidl:
## Description
Compiling the parquet C++ tests takes a long time.  This work attempts to address this by splitting the (mostly) monolithic `parquet_test.cpp` into multiple compilation units. On my workstation (10 core i7) I was able to reduce the compile time for PARQUET_TEST from 48s down to 23s.

This PR also splits the testing `base_fixture.hpp` file into three parts to cut down on unnecessary rmm includes. This change resulted in a time savings of around 20% when compiling the entire test suite locally.

## Checklist


etseidl:
Latest changes get the compile time down to 19s.

etseidl:
A bit of scope creep...while trying to figure out where compile time was going, I found that including `base_fixture.hpp` actually added a significant chunk of time. I split that file into two, moving parts associated with defining `main` to a new file `testing_main.hpp`. This reduced the overall compile time for the full test suite from around 10.5 min to just over 7.5 min.

vuule:
I'll measure the build time locally (tomorrow) to help confirm that this change does reduce the total build time.

vuule:
The new quote_normalization_test file needs to be updated, got an error building locally.

cpp/tests/io/parquet_test.cpp
vuule, line 1: why not remove this file?

etseidl: There used to be a few misc tests still in `parquet_test.cpp`, but I moved them elsewhere to leave that file responsible just for generating `main()`.

vuule: I'd be okay with moving main() gen to another test file and removing parquet_test.cpp. Pretty sure we do this for other multi-file test sets.
Could we even place main() gen in `parquet_common`?

etseidl: Yes, but then whichever file it's moved to will take that much longer to compile, and likely become the long pole in the tent. For instance, moving `main` to `parquet_common.cpp` added 2 sec to the compile time for `PARQUET_TEST`. How about renaming to `parquet_test_main.cpp`?

Then again, I guess it depends on what to optimize...one fewer file to compile might make the overall time for the entire test suite decrease.  Guess I'll test that too.

vuule: I did not realize this had measurable impact on compilation time :sweat_smile: 
If we do keep this file just for main(), might be a good idea to add a comment about it so I don't remove the file in a few months, when I forget this thread :)

etseidl: Overall compile time was a wash, so I'll add the note to `parquet_test.cpp`.

vuule (approved):
90 files. 8K lines. All solid :D

PR #14664, opened 2023-12-20
Add row conversion code from spark-rapids-jni

ttnghia:
This temporarily moves the row conversion code from spark-rapids-jni into libcudf. It is necessary to have the row conversion code compiled in a static library to overcome a CCCL issue that triggers invalid memory access when calling to `thrust::in(ex)clusive_scan` (https://github.com/NVIDIA/spark-rapids-jni/issues/1567).

In the future, when we have CCCL updated to fix the issue (1567), we may need to move the code back into spark-rapids-jni.


hyperbolic2346 (approved):

cpp/include/cudf/row_conversion.hpp
vyasr, line 1: Looks like this is just copied/moved from java/src/main/native/src/row_conversion.hpp

cpp/tests/transform/row_conversion.cpp
vyasr, line 1: Copied from https://github.com/NVIDIA/spark-rapids-jni/blob/branch-24.02/src/main/cpp/tests/row_conversion.cpp

cpp/src/transform/row_conversion.cu
vyasr, line 1: The style diff here is because the code in `java/` was not formatted with clang-format.

vyasr (approved):

mythrocks (approved):
Let's update the description to mention that this is a workaround for the failure in `spark-rapids-jni`.

cpp/src/transform/row_conversion.cu
mythrocks: That accounts for the style diffs.  There were material differences in the `spark-rapids-jni` version of `row_conversion.cu` that's not the easiest to spot in this diff.  This remains a temp fix, of course, until the CCCL issue is resolved.

cpp/src/transform/row_conversion.cu
```diff
@@ -61,37 +60,40 @@
...
 using cuda::aligned_size_t;
 #else
-template <std::size_t> using aligned_size_t = size_t; // Local stub for cuda::aligned_size_t.
-#endif // ASYNC_MEMCPY_SUPPORTED
+template <std::size_t>
+using aligned_size_t = size_t;  // Local stub for cuda::aligned_size_t.
+#endif  // ASYNC_MEMCPY_SUPPORTED
 
 namespace cudf {
-namespace jni {
```
mythrocks, line 94: I admire your thoroughness in this, @ttnghia.

hyperbolic2346:
Seems there was a cmake file missed

PR #14665, opened 2023-12-21
Reorganize ORC reader into multiple files and perform some small fixes to cuIO code

ttnghia:
This refactors the ORC reader, moving ORC code around to facilitate the upcoming support for chunked reading of the input files.

No new functionality/implementation is added in this PR. Only the existing code is moving around, except that some small issues of the related ORC/cuIO code are also fixed.


cpp/src/io/utilities/hostdevice_vector.hpp
```diff
@@ -204,8 +204,10 @@ class hostdevice_vector {
 template <typename T>
 class hostdevice_2dvector {
  public:
+  hostdevice_2dvector() : hostdevice_2dvector(0, 0, cudf::get_default_stream()) {}
```
ttnghia, line 207: This is added so the class now has a default constructor.

cpp/src/io/utilities/hostdevice_vector.hpp
```diff
@@ -204,8 +204,10 @@ class hostdevice_vector {
...
+
   hostdevice_2dvector(size_t rows, size_t columns, rmm::cuda_stream_view stream)
-    : _size{rows, columns}, _data{rows * columns, stream}
+    : _data{rows * columns, stream}, _size{rows, columns}
```
ttnghia, line 210: Fix initialization order.

robertmaynard (approved):
Approving CMake changes

cpp/src/io/orc/reader_impl.hpp
```diff
@@ -30,15 +27,14 @@
 #include <rmm/cuda_stream_view.hpp>
 
 #include <memory>
-#include <string>
-#include <utility>
+#include <optional>
 #include <vector>
 
 namespace cudf::io::orc::detail {
+using namespace cudf::io::detail;
```
vuule, line 34: I don't know if we should be doing this, especially in a header file.

ttnghia: This is due to the changing of namespace that was just merged in, and to minimize code change. Of course I can remove this and add more prefix into the variables.

ttnghia: Indeed, there was only one variable that needed to be added prefix. Thanks for rasing this up.

vuule: thank you for looking into this!

cpp/src/io/orc/reader_impl.hpp
```diff
@@ -71,18 +66,50 @@ class reader::impl {
...
+  void prepare_data(uint64_t skip_rows,
+                    std::optional<size_type> const& num_rows_opt,
+                    std::vector<std::vector<size_type>> const& stripes);
+
+  /**
+   * @brief Populate the output table metadata from the file metadata.
+   *
+   * @return Columns' metadata to output along with the table read from file
+   */
+  table_metadata populate_metadata();
```
vuule, line 85: The name doesn't quite fit the singature. `make_table_metadata` would be a batter option IMO.

cpp/src/io/orc/reader_impl_helpers.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * @brief Keeps track of orc mapping and child column details.
+ */
+struct reader_column_meta {
+  // Mapping between column id in orc to processing order.
+  std::vector<std::vector<size_type>> orc_col_map;
+
+  // Number of rows in child columns.
+  std::vector<uint32_t> num_child_rows;
```
vuule, line 42: A lot of `uint32_t`s used instead of `cudf::size_type` in these files. Not a suggestion for this PR, just a note for potential future clean up.

cpp/src/io/orc/reader_impl.cu
```diff
@@ -989,12 +68,22 @@ table_with_metadata reader::impl::read(uint64_t skip_rows,
...
+  // Save the output table metadata into `_output_metadata` for reuse next time.
+  _output_metadata = std::make_unique<table_metadata>(out_metadata);
+
+  return out_metadata;
+}
+
+table_with_metadata reader::impl::read_chunk_internal()
+{
+  // There are no columns in the table
+  if (_selected_columns.num_levels() == 0) { return {std::make_unique<table>(), table_metadata{}}; }
```
vuule, line 80: Another one mostly tangential to the PR: we should probably still return user data in the table_metadata, even if no columns are selected/exist.
Edit: should be an easy fix, we just need to call `populate_metadata` before this line and return `std::move(out_metadata)`.

cpp/src/io/orc/reader_impl.hpp
```diff
@@ -71,18 +66,50 @@ class reader::impl {
                            std::vector<std::vector<size_type>> const& stripes);
 
  private:
+  /**
+   * @brief Perform the necessary data preprocessing for parsing file later on.
```
vuule, line 70: we perform decoding in this function, I don't see how it's preparing for further processing. The only work we do after this function is to create columns out of the decoded buffers.

vuule (commented):
flushing the partial review, got some large scope questions

cpp/src/io/orc/reader_impl.hpp
ttnghia: Tried to match it with the similar function in parquet reader. How about `make_output_metadata`?

cpp/src/io/orc/reader_impl.hpp
ttnghia: Yes this function should be everything here. I was planning for the chunked reader thus the docs is ahead of time. Let me change it back.

cpp/src/io/orc/reader_impl.hpp
vuule: Anything that makes sense with factory-ness of the function is fine by me

cpp/src/io/orc/reader_impl.cu
ttnghia: Humnn, this seems to be wrong:
```
FAILED python/cudf/cudf/tests/test_orc.py::test_orc_columns_and_index_param[columns1-True-None] - IndexError: list index out of range
FAILED python/cudf/cudf/tests/test_orc.py::test_orc_columns_and_index_param[columns1-True-index_obj1] - IndexError: list index out of range
FAILED python/cudf/cudf/tests/test_orc.py::test_orc_columns_and_index_param[columns1-True-index_obj2] - IndexError: list index out of range
FAILED python/cudf/cudf/tests/test_orc.py::test_orc_columns_and_index_param[columns1-None-None] - AssertionError: DataFrame are different

DataFrame shape mismatch
[left]:  (0, 0)
[right]: (3, 0)
FAILED python/cudf/cudf/tests/test_orc.py::test_orc_columns_and_index_param[columns1-None-index_obj1] - IndexError: list index out of range
FAILED python/cudf/cudf/tests/test_orc.py::test_orc_columns_and_index_param[columns1-None-index_obj2] - IndexError: list index out of range
```

cpp/src/io/orc/reader_impl_preprocess.cu
vuule, line 1: Another one where consistency with Parquet does not quite fit. Maybe reader_impl_decode.cu?

cpp/src/io/orc/reader_impl.cu
vuule, line 1: could this be a .cpp file now?

cpp/src/io/orc/reader_impl.cu
vuule: Looks like Python uses something from the the table_metadata that isn't correct for this case.
Let's leave this one for another PR :)

cpp/src/io/orc/reader_impl.cu
ttnghia: I tried that. Unfortunately, the file includes cuco's map through the header chain. I tried to refactor that out without success.

vuule: sad

cpp/src/io/orc/reader_impl_preprocess.cu
ttnghia: But it does not just decoding. There may be file read then decompressing then decoding, so probably "preprocess" is still good.

vuule: IMO that's not "preprocess", that's "process" :D
The only thing done after this is to create tables from the data buffers, which remain unchanged.

ttnghia: I can change it to "process" now but will change back to "preprocess" once I add an additional "compute chunk" step between it and the output step. What do you think?

vuule: sounds good, thank you

vuule (approved):

cpp/src/io/orc/aggregate_orc_metadata.cpp
```diff
@@ -220,27 +219,25 @@ aggregate_orc_metadata::select_stripes(
   }
 
   // Read each stripe's stripefooter metadata
-  if (not selected_stripes_mapping.empty()) {
```
bdice, line 223: How do we know if it's safe to remove this `if`? (I am assuming the other changes in this section are from the change in indentation. If I've missed logical changes, please point them out.)

cpp/src/io/orc/reader_impl_chunking.hpp
```diff
@@ -0,0 +1,42 @@
+/*
```
ttnghia, line 1: The new files contain extracted code from `reader_impl.hpp/cu`.

harrism:
@ttnghia I think this needs a more descriptive PR title.

ttnghia:
> @ttnghia I think this needs a more descriptive PR title.

I've changed the title a little bit. Please feel free to recommend any better one.

harrism:
> > @ttnghia I think this needs a more descriptive PR title.
> 
> I've changed the title a little bit. Please feel free to recommend any better one.

I guess I meant the "refactor ORC reader" part is a bit vague and is likely to occur again.

ttnghia:
@harrism how about now?

cpp/src/io/orc/reader_impl_helpers.hpp
```diff
@@ -0,0 +1,153 @@
...
+    case orc::BYTE: return type_id::INT8;
+    case orc::SHORT: return type_id::INT16;
+    case orc::INT: return type_id::INT32;
+    case orc::LONG: return type_id::INT64;
+    case orc::FLOAT: return type_id::FLOAT32;
+    case orc::DOUBLE: return type_id::FLOAT64;
+    case orc::STRING:
+    case orc::BINARY:
+    case orc::VARCHAR:
+    case orc::CHAR:
```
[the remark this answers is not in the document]
bdice: I can see how `[[fallthrough]]` might feel overused. I'm usually in favor of using it but I will defer to you on choosing whether to use it here.

cpp/src/io/orc/aggregate_orc_metadata.cpp
bdice: Ah. I thought there was more to consider when I quickly reviewed this the first time. I agree this is fine.

bdice (approved):

PR #14666, opened 2023-12-23
Fix column parameter handling in `read_orc`

galipremsagar:
## Description
When `columns=[]` for `read_orc`, pandas actually only drops the column and preserves the `index` while reading an orc file. Fixing the `cudf` behavior to match the same.

```
= 198 failed, 101241 passed, 2091 skipped, 954 xfailed, 312 xpassed in 1098.81s (0:18:18) =
```

This PR:
```
= 185 failed, 101254 passed, 2091 skipped, 954 xfailed, 312 xpassed in 1088.47s (0:18:08) =
```
## Checklist


python/cudf/cudf/_lib/orc.pyx
```diff
@@ -129,14 +129,20 @@ cpdef read_orc(object filepaths_or_buffers,
 
     data, index = data_from_unique_ptr(
         move(c_result.tbl),
-        col_names if columns is None else names,
+        col_names if columns is None
+        or (isinstance(columns, list) and len(columns) == 0) else names,
         actual_index_names
     )
```
isVoid, line 135: Would the following be clearer in logic?

Use `if columns is not None and (isinstance(columns, list) and len(columns) == 0)` as the main branching condition:

```python
# shortcut to return
if columns is not None and (isinstance(columns, list) and len(columns) == 0):
        index = cudf.RangeIndex(0, len(data[list(data.keys())[0]]))
        data = {}
        return index, data
        
data, index = data_from_unique_ptr(...) # unchanged
...
```

galipremsagar: We could do that if we knew how many rows are in the entire orc file. Unfortunately, we don't have that data from orc file metadata, that's why we need to read the data and then fetch the number of rows with `len(data[list(data.keys())[0]])`

isVoid: Ok. The num row information should exist in `c_result`. But there's no easy way extracting that information without `data_from_unique_ptr`. I'm a bit rusted on cython+owning pointer tricks in cuDF, but would the following code work?

```python
if columns is not None and (isinstance(columns, list) and len(columns) == 0):
        cdef size_type nrow = c_result.tbl[0].view().nrows()
        return cudf.RangeIndex(nrow), {}
```

This is also non-blocking, so feel free to move on.

galipremsagar: Getting the following error:
```
              nrows = c_result.tbl[0].view().num_rows()
                                        ^
      
      /nvme/0/pgali/cudf/python/cudf/cudf/_lib/orc.pyx:134:34: Indexing 'unique_ptr[table]' not supported for index type 'long'
      ninja: build stopped: subcommand failed.
```

isVoid: Ah, so `c_result.tbl` is a `unique_ptr<table>` type. with `.get` method you can retrieve the underlying raw pointer.

```python
    cdef size_type nrows = c_result.tbl.get()[0].view().num_rows()
    if columns is not None and (isinstance(columns, list) and len(columns) == 0):
        return cudf.RangeIndex(nrows), {}
```

galipremsagar: Incorporated this suggestion 👍 Thanks @isVoid for helping with this issue.

mroeschke (approved):

PR #14668, opened 2023-12-25
Deprecate setting of incompatible dtypes to an existing column

galipremsagar:
## Description
This PR deprecates the setting of a value that is not of same dtype as that of a column.

```
= 198 failed, 101241 passed, 2091 skipped, 954 xfailed, 312 xpassed in 1098.81s (0:18:18) =
```

This PR:
```
= 176 failed, 101263 passed, 2091 skipped, 954 xfailed, 312 xpassed in 1096.08s (0:18:16) =
```

## Checklist

### DIFF
--- a/python/cudf/cudf/core/series.py
+++ b/python/cudf/cudf/core/series.py
@@ -284,6 +284,15 @@ class _SeriesIlocIndexer(_FrameIndexer):
             to_dtype = np.result_type(value.dtype, self._frame._column.dtype)
             value = value.astype(to_dtype)
             if to_dtype != self._frame._column.dtype:
+                # Do not remove until pandas-3.0 support is added.
+                warnings.warn(
+                    f"Setting an item of incompatible dtype is deprecated "
+                    "and will raise in a future error of pandas. "
+                    f"Value '{value}' has dtype incompatible with "
+                    f"{self._frame._column.dtype}, "
+                    "please explicitly cast to a compatible dtype first.",
+                    FutureWarning,
+                )
                 self._frame._column._mimic_inplace(
                     self._frame._column.astype(to_dtype), inplace=True
                 )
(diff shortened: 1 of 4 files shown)


python/cudf/cudf/core/series.py
```diff
@@ -284,6 +284,15 @@ def __setitem__(self, key, value):
             to_dtype = np.result_type(value.dtype, self._frame._column.dtype)
             value = value.astype(to_dtype)
             if to_dtype != self._frame._column.dtype:
+                # Do not remove until pandas-3.0 support is added.
+                warnings.warn(
```
mroeschke, line 288: Would cudf allow

```
ser = pd.Series([1, 2])
ser[0] = np.nan
```
?

In pandas, this is still allowed to cast int to float since nan is a missing value

python/cudf/cudf/core/series.py
```diff
@@ -284,6 +284,15 @@ def __setitem__(self, key, value):
...
+                    f"Setting an item of incompatible dtype is deprecated "
```
mroeschke, line 289: Additionally, what would happen when you have

```
ser = pd.Series()
ser[0] = 1
```

In pandas this casts object to int which is OK.

galipremsagar: We error when we try to set a non-string data to `object` dtype series. (Since cudf has no true object dtype, make perfect sense for pandas to cast in this case).

python/cudf/cudf/core/series.py
galipremsagar: This isn't possible in cudf, we don't allow setting `np.nan` to a non-floating column.

python/cudf/cudf/core/series.py
mroeschke: True makes sense. I think it's OK that cudf is more strict here than pandas

mroeschke (approved):

Issue #14684, opened 2023-12-29
[BUG] `str.character_ngrams` produces <NA> with strings < ngram length


Vortexx2:
**Describe the bug**
The `str.character_ngrams` function produces token `<NA>` for strings which are lesser than the provided `n` (shown in image for the case of bigrams).
![result output](https://github.com/rapidsai/cudf/assets/68988130/946aeebb-6be3-4719-91e7-25eb9e2c0091)

I have debugged this and as far as I understand it, it is being caused by an empty list returned by the `libstrings.generate_character_ngrams` function. This causes <NA> to be a part of the result when it is exploded in the problematic function.
This issue causes several bugs in downstream tasks (like when using cuml for `CountVectorizer` etc).

**Steps/Code to reproduce bug**
Minimum code required to reproduce the bug:
```
import cudf
str_series = cudf.Series(['1744', '4'])
str_series.str.character_ngrams(2)
```

**Expected behavior**
<NA> should not be a part of the output. This causes several downstream tasks to fail because <NA> is not a valid token in the actual input string series.

**Environment overview (please complete the following information)**
 - Environment location: Cloud GCP
 - Method of cuDF install: pip
 
**Environment details**
```
**git***
     Not inside a git repository
     
     ***OS Information***
     PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
     NAME="Debian GNU/Linux"
     VERSION_ID="11"
     VERSION="11 (bullseye)"
     VERSION_CODENAME=bullseye
     ID=debian
     HOME_URL="https://www.debian.org/"
     SUPPORT_URL="https://www.debian.org/support"
     BUG_REPORT_URL="https://bugs.debian.org/"
     Linux janmey-gpu-c2 5.10.0-26-cloud-amd64 #1 SMP Debian 5.10.197-1 (2023-09-29) x86_64 GNU/Linux
     
     ***GPU Information***
     Fri Dec 29 10:21:54 2023
     +-----------------------------------------------------------------------------+
     | NVIDIA-SMI 525.105.17   Driver Version: 525.105.17   CUDA Version: 12.0     |
     |-------------------------------+----------------------+----------------------+
     | GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
     | Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
     |                               |                      |               MIG M. |
     |===============================+======================+======================|
[...]

PR #14685, opened 2023-12-29
Fix issue with below limit strings in ngram calculation

Vortexx2:
When strings provided were below `n` characters, the ngram function returns an empty list. This was previously exploded, without filtering out the empty lists, causing the <NA> token to occur erroneously. Now, the empty lists should be filtered out.

## Description
Changes Proposed:
Should close the issue [BUG] #14684.
Addressed an issue in the ngram function where input strings below a specified length (n) resulted in the function returning an empty list. Previously, this empty list was not filtered out during further processing, leading to the inadvertent occurrence of the <NA> token.

Solution:
Modified the code to include a filtering step that removes empty lists generated by the ngram function when input strings are below the specified length. This ensures that the <NA> token is no longer erroneously introduced.

Impact:
This enhancement improves the accuracy and reliability of the ngram function by handling edge cases where input strings are shorter than the specified length. The filtering of empty lists prevents unintended consequences, specifically the occurrence of the <NA> token in such scenarios.

## Checklist
- [✅ ] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cudf/blob/HEAD/CONTRIBUTING.md).
- [ ❌ ] New or existing tests cover these changes.
- [✅ ] The documentation is up to date with these changes.


python/cudf/cudf/core/column/string.py
```diff
@@ -4843,6 +4843,7 @@ def character_ngrams(
         result = self._return_or_inplace(lc, retain_index=True)
 
         if isinstance(result, cudf.Series) and not as_list:
+            result = result[result.list.len() > 0] # before exploding, removes those lists which have 0 length
```
davidwendt, line 4846: Seems this fix could go in the calling code (cuml?). 
Just pass `as_list=True` and do the filter step and explode outside this function.

davidwendt (changes requested):
Just wondering if this fix is an appropriate, breaking change to cuDF behavior.

python/cudf/cudf/core/column/string.py
Vortexx2: This fix should be included in the cuDF repo due to it being a bug in the `character_ngrams` function itself. sklearn does not have this underlying issue with the `sklearn.feature_extraction.text.CountVectorizer` class that causes NA tokens to be returned on empty strings for example.

davidwendt: Ok. Could you add a pytest for this?

Vortexx2: Sure

davidwendt: Any progress on this @Vortexx2 ?

davidwendt:
Closing this in favor of #15371

PR #14686, opened 2023-12-29
Fix constructing Series/Index from arrow array and dtype

mroeschke:
## Description
Previously, passing a `dtype=` argument to the constructors was ignored when passing arrow array objects

## Checklist

### DIFF
--- a/python/cudf/cudf/core/column/column.py
+++ b/python/cudf/cudf/core/column/column.py
@@ -2036,6 +2036,8 @@ def as_column(
                 new_dtype = "str"
 
             col = col.astype(new_dtype)
+        elif dtype is not None:
+            col = col.astype(dtype)
 
         return col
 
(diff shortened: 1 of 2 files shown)


python/cudf/cudf/core/column/column.py
```diff
@@ -2140,15 +2151,9 @@ def as_column(
                 arbitrary,
                 from_pandas=True,
             )
-            if isinstance(pyarrow_array.type, pa.Decimal128Type):
-                pyarrow_type = cudf.Decimal128Dtype.from_arrow(
-                    pyarrow_array.type
-                )
-            else:
-                pyarrow_type = arbitrary.dtype
```
vyasr, line 2148: Do we not still need this logic if the provided dtype was None?

mroeschke: I don't think so. if `dtype=None`, the dtype inference will be done by pyarrow and will be picked up by `ColumnBase`.  Plus we don't want to override the user's passed `dtype` if it wasn't `None`

vyasr (approved):

PR #14687, opened 2023-12-29
Fix `nan_as_null` not being respected when passing cudf object

mroeschke:
## Description
Fixes the following

```python
In [1]: import cudf

In [2]: import pyarrow as pa

In [5]: cudf.Series(cudf.Series(pa.array([float("nan"), None]), nan_as_null=False), nan_as_null=True)
Out[5]: 
0     NaN <-- not converted to <NA>
1    <NA>
dtype: float64
```

Needs https://github.com/rapidsai/cudf/pull/14450 first

## Checklist


python/cudf/cudf/core/column/column.py
```diff
@@ -1939,20 +1939,18 @@ def as_column(
...
+            column = arbitrary._column
+        elif isinstance(arbitrary, cudf.BaseIndex):
+            column = arbitrary._values
         else:
-            return arbitrary
-
-    elif isinstance(arbitrary, cudf.Series):
-        data = arbitrary._column
+            column = arbitrary
+        if column.dtype.kind == "f" and (nan_as_null is None or nan_as_null):
```
isVoid, line 1949: Usually I tend to prefer some parameter handling at the start of the function, such as:
```python
if nan_as_null is None:
    nan_as_null = True
```

isVoid (approved):
A small suggestion.

wence- (approved):

python/cudf/cudf/core/column/column.py
rjzamora, line 1949: @mroeschke - [The CI failure](https://github.com/rapidsai/cudf/actions/runs/7465250252/job/20314620832?pr=14687) is a result of us hitting this check when [we call compute() in the dask-cudf test](https://github.com/rapidsai/cudf/blob/6a23775db29dc4b38820994297c94201c9287aaf/python/dask_cudf/dask_cudf/tests/test_core.py#L62).

When a dask collection is "computed", the partitions need to be concatenated together. It seems that calling `cudf.concat` uses the [cudf.Series._concat](https://github.com/rapidsai/cudf/blob/6a23775db29dc4b38820994297c94201c9287aaf/python/cudf/cudf/core/series.py#L1647), which in turn calls this `as_column` function with a default `nan_as_null=True` setting. This means that dask-cudf correctly uses the user-provided `nan_as_null=False` option when each partition is converted from pandas to cudf, but then the nans are lost when the partitions are concatenated together in `compute`.

One possible fix *may* be to modify [this line](https://github.com/rapidsai/cudf/blob/6a23775db29dc4b38820994297c94201c9287aaf/python/cudf/cudf/core/series.py#L1707) to something like:

```python
 return cls(data=col, index=index, name=name, nan_as_null=False)
```

However, I'm not immediately sure if there are other corner cases that this doesn't catch.

rjzamora: Also, I suppose the best way to test this outside of dask is to add a `cudf.concat` test that checks that existing nan values are preserved. In the case that there are both nan and null values, then we would probably need to raise an error or "promote" everything to null.

mroeschke: Thanks for the insight! The `Series._concat` piece is what I was missing. I am going to see what breaks if `nan_as_null` is set to `False`. In theory, I would imagine nans and nulls should not be mangled together when merged together

mroeschke:
Looks like there's too many moving parts here when having `nan_as_null` applying toward cudf objects. Maybe it's just easier to ignore this keyword for them so closing

PR #14689, opened 2023-12-30
Use as_column instead of arange for range like inputs

mroeschke:
## Description

1. Allows range-like inputs in `as_column` to short circuit and not materialize when creating columns
2. Avoids diverging column construction logic between `column.arange` and `column.as_column` 

## Checklist


python/cudf/cudf/core/column/column.py
```diff
@@ -1938,13 +1936,28 @@ def as_column(
     * Objects exposing ``__array_interface__``(e.g., numpy arrays)
     * pyarrow array
     * pandas.Categorical objects
+    * range objects
     """
-    if isinstance(arbitrary, ColumnBase):
+    if isinstance(arbitrary, (range, pd.RangeIndex, cudf.RangeIndex)):
+        # Handle cudf.RangeIndex, before cudf.BaseIndex to avoid
```
vyasr, line 1942: I'm confused, if we're calling `sequence` here then we are materializing a column. How does adding this stage avoid the materialization? Is there an intermediate materialization that would happen otherwise?

vyasr (commented):
The arange removal looks good, just trying to understand one thing before approving.

python/cudf/cudf/core/column/column.py
mroeschke: Yeah that's fair. Before we were materializing by essentially evaluating `list(range)` as opposed to now we materialize by passing the `stat, step, length` attributes. I don't think this different is very important to future readers so I'll remove

vyasr (approved):

PR #14696, opened 2024-01-02
Change `io::detail::orc` namespace into `io::orc::detail`

ttnghia:
This PR simply changes the (sub) namespace `io::detail::orc` namespace into `io::orc::detail`. From now, the implementation of ORC reader and writer will be part of the `detail` namespace under the `io::orc` namespace.

### DIFF
--- a/cpp/include/cudf/io/orc.hpp
+++ b/cpp/include/cudf/io/orc.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2023, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/cpp/include/cudf/io/orc.hpp
+++ b/cpp/include/cudf/io/orc.hpp
@@ -1304,7 +1304,7 @@ class orc_chunked_writer {
   void close();
 
   /// Unique pointer to impl writer class
-  std::unique_ptr<cudf::io::detail::orc::writer> writer;
+  std::unique_ptr<cudf::io::orc::detail::writer> writer;
 };
 
 /** @} */  // end of group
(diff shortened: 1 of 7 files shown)


cpp/include/cudf/io/orc.hpp
```diff
@@ -1304,7 +1304,7 @@ class orc_chunked_writer {
   void close();
 
   /// Unique pointer to impl writer class
-  std::unique_ptr<cudf::io::detail::orc::writer> writer;
+  std::unique_ptr<cudf::io::orc::detail::writer> writer;
```
vuule, line 1307: How about this?
```suggestion
  std::unique_ptr<orc::detail::writer> writer;
```

cpp/src/io/orc/reader_impl.hpp
```diff
@@ -34,7 +34,7 @@
 #include <utility>
 #include <vector>
 
-namespace cudf::io::detail::orc {
+namespace cudf::io::orc::detail {
 using namespace cudf::io::orc;
```
vuule, line 38: We should not need this anymore, since now `cudf::io::orc::detail` is nested in `cudf::io::orc`.

hyperbolic2346 (approved):
❤️

mythrocks (approved):
Ah, this is much better.

cpp/src/io/orc/writer_impl.hpp
vuule, line 1: L357 and L358 have fully qualified namespace - not needed any more.

cpp/src/io/functions.cpp
```diff
@@ -270,7 +270,7 @@ void write_csv(csv_writer_options const& options,
     mr);
 }
 
-namespace detail_orc = cudf::io::detail::orc;
+namespace orc_detail = cudf::io::orc::detail;
```
vuule, line 273: this alias only saves a single character - let's remove and use `orc::detail`

cpp/src/io/functions.cpp
vuule, line 1: line 325 does not need `cudf::io` namespace

vuule (commented):
Found a few more redundancies
looks great otherwise

PR #14697, opened 2024-01-02
Use make_strings_children for fill() specialization logic

davidwendt:
## Description
Refactors the `cudf::strings::detail::fill()` function to use `make_strings_children`. 
This refactors some duplicated logic and will help to enable large strings support.

## Checklist

### DIFF
--- a/cpp/src/strings/filling/fill.cu
+++ b/cpp/src/strings/filling/fill.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2023, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


vuule (approved):

vuule:
What is the difference in behavior with https://github.com/rapidsai/cudf/pull/14697/commits/3de1ee5b3b3421c34f0b5539f2ba7fc622b2d821?
Looks like it's just calling copy_string with an empty string, I must be missing something.

davidwendt:
> What is the difference in behavior with [3de1ee5](https://github.com/rapidsai/cudf/commit/3de1ee5b3b3421c34f0b5539f2ba7fc622b2d821)? Looks like it's just calling copy_string with an empty string, I must be missing something.

Just a subtle improvement. The current commit only calls `element()` when necessary.
The previous commit called it for every row and threw it away if it was not needed.
This is more inline with original code as well.

vuule:
> > What is the difference in behavior with [3de1ee5](https://github.com/rapidsai/cudf/commit/3de1ee5b3b3421c34f0b5539f2ba7fc622b2d821)? Looks like it's just calling copy_string with an empty string, I must be missing something.
> 
> Just a subtle improvement. The current commit only calls `element()` when necessary. The previous commit called it for every row and threw it away if it was not needed. This is more inline with original code as well.

oh, that was not the fix for tests? I misunderstood and spent too much time staring at the diff. I'll take a look at the other new commit as well.

ttnghia (approved):

PR #14700, opened 2024-01-03
Use offsetalator in gather_chars

davidwendt:
## Description
Update `cudf::strings::detail::gather_chars()` utility to use the offsetalator iterators. 
The offsetalator enables the utility to use int32 or int64 offsets in the future with no additional changes.
This utility is mainly used by a `make_strings_column` factory and is already optimized for columns with both long and short strings.
The `input_offsetalator` is also updated to include an additional optional offset parameter for cases where the parent strings column has been sliced and requires building the iterator starting from a non-zero index within the offsets child column.

## Checklist

### DIFF
--- a/cpp/include/cudf/strings/detail/gather.cuh
+++ b/cpp/include/cudf/strings/detail/gather.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2023, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


cpp/include/cudf/strings/detail/gather.cuh
```diff
@@ -109,28 +110,25 @@ __global__ void gather_chars_fn_string_parallel(StringIterator strings_begin,
     // between `[out_start_aligned, out_end_aligned)` will be copied using uint4.
     // `out_start + 4` and `out_end - 4` are used instead of `out_start` and `out_end` to avoid
     // `load_uint4` reading beyond string boundaries.
-    int32_t out_start_aligned =
+    int64_t const out_start_aligned =
```
PointKernel, line 113: Shall we use an alias to avoid using raw `int64_t` all over the place? e.g.
```cpp
  using offset_type = typename cudf::detail::input_offsetalator::value_type;
  offset_type const out_start_aligned = ...
```

davidwendt: I'd rather not do this at this time. Since offset-types can be either int32 or int64, I do not want to use `offset_type` as a name. I'd prefer using `int64` for the time being.

PointKernel: Explicit type is preferred based on offline discussions.

PointKernel (approved):
LGTM

Issue #14039, opened 2023-09-06
[BUG] cudf.to_datetime with arg with `Z` drops UTC offset incorrectly


mroeschke:
**Describe the bug**
cudf.to_datetime with arg with `Z` drops UTC offset incorrectly

**Steps/Code to reproduce bug**
```python
In [1]: import cudf

In [2]: cudf.to_datetime(["2019-01-01T00:00:00.000Z"])
Out[2]: DatetimeIndex(['2019-01-01'], dtype='datetime64[ns]')
```

**Expected behavior**
Since cudf does not support timezones currently, I would expect this to raise a `NotImplementedError`

It appears there was a commit that tried to address this, https://github.com/rapidsai/cudf/commit/95082103d26d327bbada92a0ed9e87aff978e23b, but it looks to simply ignore "Z"

**Environment overview (please complete the following information)**
 - Environment location: Bare-metal
 - Method of cuDF install: conda
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**
Please run and paste the output of the `cudf/print_env.sh` script here, to gather any other relevant environment details

**Additional context**

cwharris:
> It appears there was a commit that tried to address this, https://github.com/rapidsai/cudf/commit/95082103d26d327bbada92a0ed9e87aff978e23b, but it looks to simply ignore "Z"

This is exactly how it should work. Adding Z does not change the timezone. All time data should be assumed to be in UTC unless otherwise specified by a timezone offset. Throwing an exception is not appropriate and is a breaking change that is effecting [nv-morpheus/morpheus](https://github.com/nv-morpheus/morpheus) as we try to update to newer versions of CUDF.

bdice:
> All time data should be assumed to be in UTC unless otherwise specified by a timezone offset.

Pandas differentiates between "naive" datetimes and timezone-aware datetimes. In pandas, parsing a `Z` will return a `datetime64[ns, UTC]` type instead of a `datetime64[ns]` type. cudf needs to align with pandas behavior.

We are temporarily reverting this change of behavior in #14701 when cuDF is used with pandas compatibility mode turned off. See my proposal for a deprecation step followed by a behavior change in cuDF here: https://github.com/rapidsai/cudf/pull/14701#pullrequestreview-1803039887

PR #14701, opened 2024-01-03
Allow Z in datetime string parsing in non pandas compat mode

mroeschke:
## Description
xref https://github.com/rapidsai/cudf/issues/14039#issuecomment-1874672202

## Checklist

### DIFF
--- a/python/cudf/cudf/tests/test_datetime.py
+++ b/python/cudf/cudf/tests/test_datetime.py
@@ -2312,12 +2312,12 @@ def test_format_timezone_not_implemented(code):
 
 
 @pytest.mark.parametrize("tz", ["UTC-3", "+01:00"])
-def test_utcoffset_not_implemented(tz):
+def test_utc_offset_not_implemented(tz):
     with pytest.raises(NotImplementedError):
         cudf.to_datetime([f"2020-01-01 00:00:00{tz}"])
 
 
-def test_Z_utcoffset():
+def test_Z_utc_offset():
     with cudf.option_context("mode.pandas_compatible", True):
         with pytest.raises(NotImplementedError):
             cudf.to_datetime(["2020-01-01 00:00:00Z"])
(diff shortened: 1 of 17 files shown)


python/cudf/cudf/tests/test_datetime.py
```diff
@@ -1306,8 +1306,11 @@ def test_datetime_infer_format(data, timezone, dtype):
 
         assert_eq(expected, actual)
     else:
-        with pytest.raises(NotImplementedError):
-            sr.astype(dtype)
+        with cudf.option_context("mode.pandas_compatible", True):
+            with pytest.raises(NotImplementedError):
+                sr.astype(dtype)
+
+        # pandas doesn't allow parsing "Z" to naive type
```
bdice, line 1313: Please move this comment above the `sr.astype(dtype)` call to make it clear that it's the cause of the error.

python/cudf/cudf/tests/test_datetime.py
```diff
@@ -2308,12 +2311,22 @@ def test_format_timezone_not_implemented(code):
         )
 
 
-@pytest.mark.parametrize("tz", ["Z", "UTC-3", "+01:00"])
-def test_no_format_timezone_not_implemented(tz):
+@pytest.mark.parametrize("tz", ["UTC-3", "+01:00"])
+def test_utcoffset_not_implemented(tz):
```
bdice, line 2315:
```suggestion
def test_utc_offset_not_implemented(tz):
```

python/cudf/cudf/tests/test_datetime.py
```diff
@@ -2308,12 +2311,22 @@ def test_format_timezone_not_implemented(code):
...
     with pytest.raises(NotImplementedError):
         cudf.to_datetime([f"2020-01-01 00:00:00{tz}"])
 
 
+def test_Z_utcoffset():
```
bdice, line 2320:
```suggestion
def test_Z_utc_offset():
```

bdice (commented):
A couple small comments. We need to follow up with a deprecation for the behavior that doesn't align with pandas, and eventually make a breaking change for that since it's a bug that cudf and pandas disagree on the behavior. We'll need the Morpheus team (@cwharris) to update their code accordingly.

cwharris (approved):

python/cudf/cudf/tests/test_datetime.py
mroeschke: Sure thing. Moved

bdice (approved):

PR #14702, opened 2024-01-03
Skip numba test that fails on ARM

brandon-b-miller:
Nightly jobs have been [failing](https://github.com/rapidsai/cudf/actions/runs/7382855293/job/20083184931) with a numba segfault.

This appears to be a longstanding issue with numba on aarch64 fixed by https://github.com/numba/llvmlite/pull/1009. Technically, the issue exists already in our tests, but it appears that changes from numba 0.58 make the conditions for the issue to occur much more likely, hence the failures occurring after removing the numba 0.58 version constraint recently. The issue should be fixed in numba 0.59.

For now however we should skip things so that nightlies can be fixed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ exclude = [
     # TODO: Remove this in a follow-up where we fix __all__.
     "__init__.py",
 ]
-line-length = 83
+line-length = 88
 
 [tool.ruff.per-file-ignores]
 # Lots of pytest implicitly injected attributes in conftest-patch.py
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ exclude = [
     # TODO: Remove this in a follow-up where we fix __all__.
     "__init__.py",
 ]
-line-length = 79
+line-length = 83
 
 [tool.ruff.per-file-ignores]
 # Lots of pytest implicitly injected attributes in conftest-patch.py
--- a/python/cudf/cudf_pandas_tests/test_cudf_pandas.py
+++ b/python/cudf/cudf_pandas_tests/test_cudf_pandas.py
@@ -668,7 +668,9 @@ def test_rolling_win_type():
     tm.assert_equal(result, expected)
 
 
-@pytest.mark.skip(reason="Requires Numba 0.59 to fix segfaults on ARM. See https://github.com/numba/llvmlite/pull/1009")
+@pytest.mark.skip(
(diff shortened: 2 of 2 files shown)


python/cudf/cudf_pandas_tests/test_cudf_pandas.py
```diff
@@ -668,6 +668,7 @@ def test_rolling_win_type():
     tm.assert_equal(result, expected)
 
 
+@pytest.mark.skip(reason="https://github.com/numba/llvmlite/pull/1009")
```
bdice, line 671:
```suggestion
@pytest.mark.skip(reason="Requires Numba 0.59 to fix segfaults on ARM. See https://github.com/numba/llvmlite/pull/1009")
```

python/cudf/cudf_pandas_tests/test_cudf_pandas.py
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023-2024, NVIDIA CORPORATION & AFFILIATES.
```
bdice, line 1: This is failing style checks for a really gross reason. This is too long of a line. There's one other case where I could see this, but I don't love the solution:

https://github.com/rapidsai/cudf/blob/af65d52c7d4ca41606482926bdcc001644b7d108/python/cudf/cudf/core/resample.py#L1-L3

brandon-b-miller: can the copyright section be exempted from check somehow? It's not really a "comment" in the normal sense I'd say.

bdice: You could `# noqa` but that’s even noisier and not better imo. I am inclined to say that if we require SPDX compatible license headers, we have to increase the line length limits.

cc: @shwina for input

brandon-b-miller: I updated this to 83.

pyproject.toml
```diff
@@ -66,7 +66,7 @@ exclude = [
     # TODO: Remove this in a follow-up where we fix __all__.
     "__init__.py",
 ]
-line-length = 79
+line-length = 83
```
brandon-b-miller, line 69:
```suggestion
line-length = 88
```

bdice (approved):

PR #14705, opened 2024-01-03
Ensure DataFrame column types are preserved during serialization

mroeschke:
## Description
xref https://github.com/rapidsai/ucxx/issues/149

Ensures column subclass e.g `RangeIndex` and data types e.g. `int8` are preserved during serialization.

I think this should be backwards compatible since we're just adding keys to the the serialized `header` dict and the deserialization checks if those new keys exist

## Checklist


pentschev (approved):
Thanks @mroeschke , I can confirm this resolves the issue as reported in https://github.com/rapidsai/ucxx/issues/149 .

charlesbluca (approved):

